PR #40, opened 2016-02-03
Action manager/redux channel

k88hudson:
This implements an action manager, which can be used both by firefox and the content side, and a `ReduxChannel`.


src/actions/action-manager.js
```diff
@@ -0,0 +1,43 @@
+// This has to be relative so the firefox add-on side can read the path
+const ActionManager = require("../lib/ActionManager");
+
+const am = new ActionManager([
+  'TOP_FRECENT_SITES_REQUEST',
+  'TOP_FRECENT_SITES_RESPONSE',
+  'RECEIVE_PLACES_CHANGES',
+  'BOOKMARKS_REQUEST',
+  'BOOKMARKS_RESPONSE'
```
marcoscaceres: nit: missing trailing comma

src/actions/action-manager.js
```diff
@@ -0,0 +1,43 @@
...
+]);
+
+function Response(type, data, options = {}) {
+  const action = {type, data};
+  if (options.error) action.error = true;
```
marcoscaceres: Nit: can I kindly ask that we adhere to JSCS please?

src/lib/ActionManager.js
```diff
@@ -0,0 +1,49 @@
+// This is from redux
+function compose(...funcs) {
+ return (...args) => {
+   if (funcs.length === 0) {
+     return args[0];
```
marcoscaceres: how odd, this would just return `undefined`, no?

src/lib/ActionManager.js
```diff
@@ -0,0 +1,49 @@
...
+   }
+
+   const last = funcs[funcs.length - 1];
+   const rest = funcs.slice(0, -1);
+
+   return rest.reduceRight((composed, f) => f(composed), last(...args));
```
marcoscaceres: Funky composer. Like it.

src/lib/ActionManager.js
```diff
@@ -0,0 +1,49 @@
...
+ };
+}
+
+module.exports = class ActionManager {
+  constructor(types) {
+    this._types = new Set(types);
+    this.validators = [this.validateStandardForm, this.validateType];
+    this.actions = {};
+    this.defineActions({
+      Action: action => action
```
marcoscaceres: should that be `action => action()`? Or calling the action returns the action? (i.e., `action()()`)?

k88hudson: in this case it's just a "base" action that returns an object, but has validations applied

yamscripts.yml
```diff
@@ -26,6 +26,9 @@ scripts:
     jpm: jpm test
     karma: karma start
 
+# tdd: Run tests continuously
+  tdd: karma start --no-single-run
```
rlr: hot :fire:

rlr:
:shipit: r+!

README.md
```diff
@@ -0,0 +1,98 @@
...
+    // error should be thrown
+    expect: "SOMETHING_RESPONSE"
+
+    // Optional. A custom timeout for an 'expect' type action.
+    timeout: 10000
+
+    // This indicates to the ReduxChannel middleware that
+    // the action should broadcast via message passing to the other
+    // side of the channel.
+    broadcast: "content-to-addon"
```
oyiptong: This will be confusing. I recommend renaming this property to `send`.

In the addon code, we use `send` to mean one-to-one communication and `broadcast` to mean one-to-many.
From content, sending a message may or may not result in a `broadcast`, if it changes state for every other tab as well.

README.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+{
+  // The type should be upper-case.
+  // Request-type actions should end in _REQUEST
+  // Response-type actions should end in _RESPONSE
+  // Required.
+  type: "SOMETHING_REQUEST",
+
+  // Only for response-type actions.
```
oyiptong: It might also be for requests. e.g. for search completion, where you'd send the keystrokes so far

README.md
```diff
@@ -0,0 +1,98 @@
...
+  // If the response is successful, it should be an array OR object.
+  // If the response is an error, it should be an Error.
+  // Optional.
+  data: {},
+
+  // Only for request-type actions. Should be an object.
```
oyiptong: Oh, I suppose that's what you intend to send the query params from my question above?

src/lib/ReduxChannel.js
```diff
@@ -0,0 +1,86 @@
...
+    const event = new CustomEvent(this.options.outgoing, {detail: JSON.stringify(action)});
+    this.options.target.dispatchEvent(event);
+  }
+
+  connectStore(store) {
+    this.on(action => store.dispatch(action));
+  }
+
+  get middleware() {
+    return store => next => function (action) {
```
oyiptong: You javascript hipster. What does this do?

marcoscaceres: currying: returns a function, which returns a function, but the last "function" is bound to the outer "this".

lib/main.js
```diff
@@ -36,19 +38,16 @@ ActivityStreams.prototype = {
   /**
    * Send a message to a worker
    */
-  send(actionName, msg, worker) {
-    worker.port.emit('addon-to-content', {
-      type: actionName,
-      data: msg
-    });
+  send(action, worker) {
+    worker.port.emit('addon-to-content', action);
```
oyiptong: unless i'm not getting something, that's broken. It won't work.

marcoscaceres: I think doing reviews at 1am is broken :)

PR #47, opened 2016-02-04
Addon Message Passing is testable

oyiptong:
Makes messaging passing to content testable


oyiptong:
Btw, no need to review httpd, it is an import from the addon sdk's repo

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource:///modules/NewTabURL.jsm");
+
+XPCOMUtils.defineLazyGetter(this, "EventEmitter", function () {
+  const {EventEmitter} = Cu.import("resource://devtools/shared/event-emitter.js", {});
+  return EventEmitter;
+});
+
+let ActivityStreams = function (options) {
+  console.debug(`ActivityStreams.init`);
```
marcoscaceres: Q: I guess we should file a bug for us to remove these at some point at the end of the project.

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource:///modules/NewTabURL.jsm");
+
+XPCOMUtils.defineLazyGetter(this, "EventEmitter", function () {
+  const {EventEmitter} = Cu.import("resource://devtools/shared/event-emitter.js", {});
+  return EventEmitter;
+});
+
+let ActivityStreams = function (options) {
```
marcoscaceres: You can probably just use default param here: 
```JS
const ActivityStreams = function (options = {}) {
```

marcoscaceres: Note, I updated the default param to have the default URL for pageURL

marcoscaceres: If the call to data.url() results in a string, maybe we can just inline the string?

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+    PlacesProvider.links.off("linkChanged", this._handlePlacesChanges);
+    PlacesProvider.links.off("manyLinksChanged", this._handlePlacesChanges);
+
+    this.off("REQUEST_TOP_FRECENT_SITES", this._respondToPlacesRequests);
+  },
+
+  /**
+   * Sets up communications with the pages
+   */
+  _setupPageMod() {
```
marcoscaceres: Maybe call this _createPageMod(), and have it return the PageMod() instace instead. That way, all `this.` assignments happen in one place (in the constructor).

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+    this.pagemod = new PageMod({
+      include: [this.pageURL],
+      contentScriptFile: data.url("content-bridge.js"),
+      contentScriptWhen: "start",
+      attachTo: "top",
+      onAttach: worker => {
+        if (!this.workers.has(worker)) {
```
marcoscaceres: Note: What `this` is ambiguous here because the fat arrow + this being an object + this being inside another function (i.e., there are 3 possible `this.` meanings). It might be good to just add a comment clarifying which `this` you are trying to act on - and just double check you are actually operating on the correct one.

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+  return EventEmitter;
+});
+
+let ActivityStreams = function (options) {
+  console.debug(`ActivityStreams.init`);
+
+  this.options = options ? options : {};
+  this.pageURL = this.options.pageURL ? options.pageURL : data.url("content/activity-streams.html");
+
+  this.workers = new Set();
```
marcoscaceres: Maybe this needs to be a WeakSet, because we don't "own/create" the workers - they are handed to us `onAttach`.

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+        worker.port.on("content-to-addon", msg => {
+          if (!msg.type) {
+            console.warn(`ActivityStreams.dispatch error: unknown message type`);
+            return;
+          }
+          if (msg.type === "pagehide") {
+            console.debug(`ActivityStreams.pagemod unloading worker`);
+            this.workers.delete(worker);
+          }
+          this.emit(msg.type, {msg:msg, worker:worker});
```
marcoscaceres: Nit: `this.emit(msg.type, {msg, worker})`;

lib/ActivityStreams.js
oyiptong: We need to iterate through them when we broadcast. A WeakSet is too weak

lib/ActivityStreams.js
oyiptong: The URL is dynamic, generated in part with the addon's id

lib/ActivityStreams.js
oyiptong: Meh. It's a debug log. It doesn't do anything in a production build

lib/ActivityStreams.js
oyiptong: RE: default params.

I keep forgetting JS is not a toy language :-p

lib/ActivityStreams.js
marcoscaceres: Nit, as the constructor is not referenced as variable anywhere, you can just write: 
JS```
function ActivityStreams(options = {pageURL: data.url("content/activity-streams.html"}) {
```

File: lib/ActivityStreams.js (hunk 4 above)
Comment (oyiptong) [inline lib/ActivityStreams.js]: Unsure of what you mean. Maybe I'm a JS brute? I think this is the object's instance. Perhaps I'm wrong and it's the prototype.

File: lib/ActivityStreams.js (hunk 1 above)
Comment (marcoscaceres) [inline lib/ActivityStreams.js]: > Meh. It's a debug log. It doesn't do anything in a production build

Ah, sorry. I thought it still spat out the messages. Awesome if they are ignored in production builds.

File: lib/ActivityStreams.js (hunk 5 above)
Comment (marcoscaceres) [inline lib/ActivityStreams.js]: Curse you WeakSet!

File: lib/ActivityStreams.js [hunk 7]
```diff
@@ -0,0 +1,144 @@
...
+};
+
+ActivityStreams.prototype = {
+
+  pageMod: null,
+
+  /**
+   * Send a message to a worker
+   */
+  send(actionName, msg, worker) {
```
marcoscaceres: .send() seems like indirection to me. That is, we are basically receiving 3 arguments and adding "addon-to-content". An alternative would be to just expose "addon-to-content" as a static and let the caller just call:

```JS
worker.port.emit(ActivityStreams.SEND, {type, data});
```

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+  },
+
+
+  /**
+   * Broadcast places changes to pages
+   */
+  _handlePlacesChanges(eventName, data) {
+    this.broadcast("RECEIVE_PLACES_CHANGES", {
+      "event": eventName,
+      "data": data
```
marcoscaceres: Nit:  just `data`

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+  },
+
+  /**
+   * Unload the application
+   */
+  unload(reason) { // eslint-disable-line no-unused-vars
+    console.debug(`ActivityStreams.unload on ${reason}`);
+
+    if(reason == "shutdown") {
+      // handle browser shutdown
```
marcoscaceres: Maybe add a console.warn("Unimplemented") here?

lib/ActivityStreams.js
marcoscaceres: An alternative might be to make this _private, so you can keep using it as a utility function.

lib/ActivityStreams.js
```diff
@@ -0,0 +1,144 @@
...
+    });
+  },
+
+  /**
+   * Unload the application
+   */
+  unload(reason) { // eslint-disable-line no-unused-vars
+    console.debug(`ActivityStreams.unload on ${reason}`);
+
+    if(reason == "shutdown") {
```
marcoscaceres: Q: if there might be other reason in the future, this might be better as switch - then you can handle all cases explicitly?
```JS
switch (reason){
case "shutdown":
  console.warn("Not implemented");
  break;
case "uninstall":
case "upgrade":
case "downgrade":
 this.workers.clear();
}
```

Or just:
```JS
switch (reason){
case "shutdown":
  console.warn("Not implemented");
  break;
default:
 this.workers.clear();
}
```

lib/ActivityStreams.js
marcoscaceres: Also, in case of a shutdown, won't it make sense to also call this.workers.clear()? ... that is, in case something cancels the shutdown later.

lib/main.js
```diff
@@ -1,156 +1,28 @@
...
-
-Cu.import("resource://gre/modules/XPCOMUtils.jsm");
-Cu.import("resource:///modules/NewTabURL.jsm");
+const {storage} = require("sdk/simple-storage");
+const {uuid} = require("sdk/util/uuid");
+const {PlacesProvider} = require("lib/PlacesProvider");
+const {ActivityStreams} = require("lib/ActivityStreams");
 
-// Generate a UUID for this client, if we don't have one yet.
+// Generate a UUID for this client, if we don"t have one yet.
```
marcoscaceres: Nit: typo `"`

lib/main.js
```diff
@@ -1,156 +1,28 @@
...
-            app.workers.delete(worker);
-          }
-          app.dispatch(msg, worker);
-        });
-      }
-    });
+  main(options) {
+    PlacesProvider.links.init();
+    let app = new ActivityStreams(options);
+    exports.app = app;
```
marcoscaceres: Nit: you can probably just assign `new ActivityStreams()` directly to `exports.app`.

test-content/components/Header.test.js
```diff
@@ -1,6 +1,8 @@
+/* globals beforeEach, afterEach, describe, it */
+
 const assert = require("chai").assert;
 const Header = require("components/Header/Header");
-const React = require("react");
+const React = require("react"); // eslint-disable-line no-unused-vars
```
marcoscaceres: Q: what does eslinit complain about here? It's not like this is doing anything different from the lines above/below?

test/lib/utils.js
```diff
@@ -0,0 +1,54 @@
...
+const {stack: Cs} = components;
+
+function do_get_file(path, allowNonexistent) {
+  try {
+    let lf = Cc["@mozilla.org/file/directory_service;1"]
+      .getService(Ci.nsIProperties)
+      .get("CurWorkD", Ci.nsILocalFile);
+
+    let bits = path.split("/");
+    for (let i = 0; i < bits.length; i++) {
```
marcoscaceres: Maybe:
```JS
  bits.filter(bit => bit)
    .forEach(bit => {
      if (bit !== "..") {
        return lf.append(bit);
      }
      lf = lf.parent;
    });
```

test/lib/utils.js
```diff
@@ -0,0 +1,54 @@
...
+          lf = lf.parent;
+        else
+          lf.append(bits[i]);
+      }
+    }
+
+    if (!allowNonexistent && !lf.exists()) {
+      // Not using do_throw(): caller will continue.
+      let stack = Cs.caller;
+      console.error(`[${stack.name} : ${stack.lineNumber}] ${lf.path} does not exist`);
```
marcoscaceres: Q: Why not use "new Error(`${lf.path} does not exist`)" here - which will give you the full stack? Or is stack.name and stack.lineNumber going to be different from this line? If so, you might still want to use `new Error()` here, otherwise, what's going to appear in the console going to be a bit confusing.

test/lib/utils.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    return lf;
+  }
+  catch (ex) {
+    do_throw(ex.toString(), Cs.caller);
```
marcoscaceres: Q: what includes do_throw()?

test/lib/utils.js
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+
+    if (!allowNonexistent && !lf.exists()) {
+      // Not using do_throw(): caller will continue.
+      let stack = Cs.caller;
+      console.error(`[${stack.name} : ${stack.lineNumber}] ${lf.path} does not exist`);
+    }
+
+    return lf;
```
marcoscaceres: Returning early seems crappy. IMHO, we should return the lf; at the end of the function, and return the null after the do_throw()

marcoscaceres: updated comment above.

test/lib/utils.js
marcoscaceres: Ah! found it!!! :)

test/lib/utils.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+  catch (ex) {
+    do_throw(ex.toString(), Cs.caller);
+  }
+
+  return null;
+}
+
+function do_throw(error, stack) {
```
marcoscaceres: Nit: rename doThrow(). Underscores in function names is not very JS (and will upset JSCS).

test/resources/dummy-activitystreams.html
```diff
@@ -0,0 +1,31 @@
+<!doctype html>
+<html lang=en-us>
+  <head>
+    <meta charset="utf8">
+  </head>
+  <body>
+      <p>Activity Streams!</p>
```
marcoscaceres: Nit: whitespace. Any reason this is using so much? It then switches back to 2 spaces in the js.

test/test-httpd.js
```diff
@@ -0,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*eslint-disable*/
```
marcoscaceres: Any reason why es-lint is disabled?

test/test-httpd.js
```diff
@@ -0,0 +1,75 @@
...
+  loader.sandbox("./lib/httpd").DEBUG = true;
+
+exports.testBasicHTTPServer = function(assert, done) {
+  // Use the profile directory for the temporary file as that will be deleted
+  // when tests are complete
+  let basePath = pathFor("ProfD");
+  let filePath = file.join(basePath, 'test-httpd.txt');
+  let content = "This is the HTTPD test file.\n";
+  let fileStream = file.open(filePath, 'w');
+  fileStream.write(content);
```
marcoscaceres: heh, synchronous IO FTW!!! :smile:

test/test-httpd.js
```diff
@@ -0,0 +1,75 @@
...
+      srv.stop(done);
+    }
+  }).get();
+};
+
+exports.testAutomaticPortSelection = function (assert, done) {
+  const srv = httpd.startServerAsync(-1);
+
+  const port = srv.identity.primaryPort;
+  assert.ok(0 <= port && port <= 65535);
```
marcoscaceres: Nit: it's easier to reason that `port >= 0 && port <= 65535`

test-content/components/Header.test.js
oyiptong: What is seemingly an import time side effect or transpiled code requirement

lib/ActivityStreams.js
oyiptong: Good point

test/test-httpd.js
marcoscaceres: More seriously, I'm confused about why you write the file here, but then you don't actually read the file "onComplete" below (instead, `response.text` just get's compared to `content`). Is there voodoo magic happening that the httpd server is reading that file from disk when serving?

test/test-httpd.js
oyiptong: This is lifted from the addon sdk. I treat it like I would a library import

lib/ActivityStreams.js
marcoscaceres: I honestly don't know from reading the code :crying_cat_face:, which is why I asked. It would be even more confusing if it's the prototype... as that adds a 4th potential of ambiguity of `this`... JS, I :heart: you, I :broken_heart: you.

marcoscaceres: @k88hudson, can you maybe comment here? Could use a second opinion.

test/test-httpd.js
oyiptong: Look at the initialization. It serves the directory given by basePath.

test/test-httpd.js
```diff
@@ -0,0 +1,75 @@
...
+  fileStream.close();
+
+  let srv = httpd.startServerAsync(port, basePath);
+
+  // Request this very file.
+  let Request = require('sdk/request').Request;
+  Request({
+    url: "http://localhost:" + port + "/test-httpd.txt",
+    onComplete: function (response) {
+      assert.equal(response.text, content);
```
marcoscaceres: As I mentioned above, I would expect the correct test to read the disk's content for /test-httpd.txt and then compare it.

lib/ActivityStreams.js
oyiptong: It's a matter of aesthetics. I like keeping it at send. This is consistent with broadcast and on the request hooks.

test/test-httpd.js
oyiptong: Check the servers I it code. The basePath argument gives the server a default directory to serve files from

lib/ActivityStreams.js
marcoscaceres: Ok, np. Is the expectation that people using ActivityStreams always use `send()` with their related workers? My only worry was that there is two ways of sending messages to the worker.

lib/ActivityStreams.js
rlr: I seriously doubt that this `this` is the `ActivityStreams` object. You can prove that by `console.log(this)`? 

you could do something like:
````
_setupPageMod() {
    let me = this;
    ...
    .... me.workers.has(worker)...
    ...
```

test/lib/utils.js
oyiptong: We use eslint for this project. And probably for all other projects as well. Gecko is not on eslint, makes things easy.

The naming is such because this code has been ported over from xpcshell, for better or for worse.

Since this has been taken with a few modifications, I think it would be fine to make it our own.

lib/ActivityStreams.js
rlr: ooooh. that fat arrow... interesting... it probably is fine https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#Arrow_functions

oyiptong: Well. Lifecycle management works, according to the tests, so I'm thinking this is the object's instance. I could verify.

The point of using the fat arrow is to make the scoping sane.

oyiptong: Yes, indeed @rlr, that is why I have abandoned var and use => wherever I can. I hope we get a version for generators too!

marcoscaceres: @rlr yeah, it's no doubt it's fine (as the tests work fine) - but it's just a bit confusing for the reader. I don't think we should add a closure - but just a code comment saying what `this` refers to. That should be sufficient.

rlr: awesomesauce. learn something new every day. #getoffmylawn #closuresaresweet :laughing:

test/test-message-passing.js
```diff
@@ -0,0 +1,79 @@
...
+    pingPromise = new Promise(resolve => {
+      app.once("PING", resolve);
+    });
+    tabs.open(url);
+    yield pingPromise;
+
+    let broadcastResponsePromise = new Promise(resolve => {
+      let count = 0;
+      app.on("PINGPONG_ALL", function response() {
+        count += 1;
```
marcoscaceres: nit: count++;

test/test-message-passing.js
```diff
@@ -0,0 +1,79 @@
...
+        if (count == 2) {
```
marcoscaceres: nit: ===

marcoscaceres:
Finished review. This is shaping up super nicely, @oyiptong!

lib/ActivityStreams.js
oyiptong: You mean to move the pagemod setup to the initialization?

lib/ActivityStreams.js
oyiptong: Yep. The expectation is that they are not aware of the 'port' implementation.

The abstraction holds if we move to webchannel or some other means of communication.

test/lib/utils.js
oyiptong: I guess the original code wanted to print an error and continue

oyiptong: misleading when it says !allowNonexistent

oyiptong: It does not print the current line number. Rather, it prints the caller's line number.

test/resources/dummy-activitystreams.html
oyiptong: that's just my editor. let's keep it consistent

lib/ActivityStreams.js
marcoscaceres: No, it's fine to keep the setup code here (i.e., construction of the object); but just return the newly constructed object.  

Ideally, initialization of all of the "this.whateves = " should happen in one place (the constructor), or else it can  get hard to track when properties are added to an object.

oyiptong:
is that an r+, @marcoscaceres ?

marcoscaceres:
R+, bar a couple of nits. Fix any you see fit.

lib/ActivityStreams.js
k88hudson: Personally I don't find this *too* confusing (in terms of the context), although there is a lot going on in this function in general so you could refactor it into separate functions etc.

yamscripts.yml
```diff
@@ -37,5 +37,5 @@ scripts:
 
 # This is just to make sure we don't make commits with failing tests
 # or uncompiled yamscripts.yml. Run automatically with husky.
-  precommit: =>test:lint && =>yamscripts
+  # precommit: =>test:lint && =>yamscripts
```
k88hudson: why comment this out?

k88hudson:
R+ for .jpmignore

lib/ActivityStreams.js
oyiptong: I vetoed on this one. We don't need custom pagemod creation yet. We can take a look in round 2 in case that breaks.

make `setupPageMod` into `createPageMod` doesn't work unless we pass the paramers the function needs, so the function signature would look like `_createPageMod(options.pageURL, this.workers, this)`, which make this a PageMod factory, where param 1 is the url to attach to, this.workers becomes the `Set` to manage workers in and `this` is the EventEmitter to use.

Seemed a bit overkill for now.

PR #52, opened 2016-02-05
fix travis

k88hudson:


k88hudson:
@oyiptong R?

k88hudson:
The reason the tests aren't passing are that jpm is looking in the test directory, which is fixed in your patch

oyiptong:
r+ with additional blacklists in jpmignore for:

configs, dev-prefs.json

PR #53, opened 2016-02-05
chore(npm): Add cpx dependency, add bundle task to ci

k88hudson:
Forgot a dependency here. Added bundle to travis so we make sure we don't break it.

@oyiptong or @rlr R?


rlr:
:+1: :ship: it

PR #54, opened 2016-02-05
fix(add-on): Fix ActivityStreams option defaulting, main.js exports

k88hudson:
I'm not sure why, but for some reason the `main` function never gets run when you replace `exports` completely; 

Also fixes defaulting of individual options for `ActivtiyStreams`

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -16,9 +16,14 @@ XPCOMUtils.defineLazyGetter(this, "EventEmitter", function () {
   return EventEmitter;
 });
 
-let ActivityStreams = function (options = {pageURL: data.url("content/activity-streams.html")}) {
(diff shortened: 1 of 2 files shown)


k88hudson:
R? @oyiptong

oyiptong:
Ah yeah, I remember now that replacing the exports object doesn't work. It looks like the fancy defaults don't work huh?

oyiptong:
nit: can you please change the commit message to `fix(addon)` so we're consistent?
r+

k88hudson:
@oyptiptong you can do it if you specify them inline like
```
function ({url="asdasd.com", whatever="blah"} = {}) {..}
```

but that pattern doesn't really work if you want to use the options object without destructuring it... at least as far as i can think of :(

Issue #45, opened 2016-02-04
Create "top sites" component


k88hudson:
![image](https://cloud.githubusercontent.com/assets/1455535/12825540/d193378a-cb44-11e5-9ab9-1c17d1fdc833.png)

rlr:
initial version landed ^^

PR #57, opened 2016-02-05
feat(content) #45 top sites ui component

rlr:
* I copied then tweaked a bunch of the styles from RNT
* Should I name it Tiles instead of TopSites?
* This is probably missing some needed media queries but it's hard to know until the rest of the components are laid out

r? @k88hudson

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,6 +15,7 @@
     "computed-property-spacing": [2, "never"],
     "array-bracket-spacing": [2, "never"],
     "space-before-function-paren": [2, {"anonymous": "always", "named": "never"}],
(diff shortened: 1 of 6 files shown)


rlr:
^^ the eslint check dislikes the test code... How do I fix that?

k88hudson:
Awesome! Can you add back the content-test to eslint and add the following to `"rules"`:

```
    "react/jsx-uses-react": 1
```

and the following to `"env"`:

```
"mocha": true
```

k88hudson:
R+ after that

rlr:
done :beers:

PR #63, opened 2016-02-08
Implement ActivityFeed component

k88hudson:
This implements the ActivityFeed component + tests

### DIFF
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -1,0 +1,42 @@
+const React = require("react");
+
+const ActivityFeedItem = React.createClass({
+  render() {
+    const site = this.props;
+    return (<li className="feed-item">
+      <div className="feed-icon" style={{backgroundImage: `url(${site.image})`}} ref="icon" />
+      <div className="feed-details">
+        <div className="feed-description">
+          <h4 className="feed-title" ref="title">{site.title}</h4>
+          <a className="feed-link" href={site.url} ref="link">{site.url}</a>
+        </div>
+        <div className="feed-stats">
+          <div>1:26pm</div>
+          <div>...</div>
+        </div>
+      </div>
+    </li>);
+  }
+});
+
+ActivityFeedItem.propTypes = {
+  url: React.PropTypes.string.isRequired,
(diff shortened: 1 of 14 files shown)


k88hudson:
@rlr R?

content-src/lib/shim.js
```diff
@@ -1,7 +1,48 @@
 const fakeFrecent = [
   {
     title: "Facebook",
-    leadImage: "https://www.facebook.com/images/fb_icon_325x325.png",
```
rlr: was `leadImage` a typo? I have `url(${site.leadImage || site.image})` in the TopSites ^^ component. I can remove if not needed.

content-src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -0,0 +1,42 @@
+const React = require("react");
+
+const ActivityFeedItem = React.createClass({
```
rlr: I probably should've broken up the TopSites component like this too (TopSitesItem). I'll revisit later :+1:

content-src/components/Base/Base.js
```diff
@@ -21,17 +22,25 @@ const Main = React.createClass({
...
-      <pre>
-      {JSON.stringify(props.Sites.frecent, null, 2)}
-      {JSON.stringify(props.Sites.changes, null, 2)}
-      </pre>
     </div>);
   }
 });
 
+
+
```
rlr: nit: 3 blank lines?

rlr:
it looks awwwwsome! :fire: R+

content-src/lib/shim.js
k88hudson: yeah i think it was just like, me not using the same schema. 

In the future i think this will need to change anyway, we will need a way to specify different resolutions for images, favicons, etc. Keeping it `image` is probably OK for now but we'll need to update things once we lock that down. Maybe let's discuss this week

content-src/components/Base/Base.js
k88hudson: oops!

content-src/components/ActivityFeed/ActivityFeed.js
k88hudson: yeah, it makes it a bit easier to test stuff :)

PR #58, opened 2016-02-05
test(addon): #20 Port PlacesProvider tests from Firefox.

emtwo:

PR #68, opened 2016-02-08
Fix broken tests and error messages on CI

oyiptong:


oyiptong:
ok, there seems to be a concurrency problem in this test... perhaps in PlacesUtils

oyiptong:
There were two problems:
1. concurrency issue with places tests
2. the app was not unloading itself in the message passing tests

1 is solved by forcing the db driver to do a sql query and at the same time doing a validation of the expected data.

r? @rlr, @k88hudson or @marcoscaceres

oyiptong:
looks like there are still some problems with the places db connection.

hmm... we must wait until the db has flushed to disk somehow...

oyiptong:
managed to get the error to be intermittent. There must be some race condition from the time things are written to disk/committed and observers are notified.

oyiptong:
https://travis-ci.org/mozilla/activity-streams/builds/108093661

oyiptong:
to be clear, this is a problem with the addVisits function, and the underlying `asyncHistory` implementation.

Basically, the promise returns, but the changes haven't been flushed to disk yet.

oyiptong:
Code for the UpdatePlaces method: https://dxr.mozilla.org/mozilla-central/rev/584870f1cbc5d060a57e147ce249f736956e2b62/toolkit/components/places/History.cpp#2849

In theory, it should invoke the completion callback: https://dxr.mozilla.org/mozilla-central/rev/584870f1cbc5d060a57e147ce249f736956e2b62/toolkit/components/places/History.cpp#2981

The notifications, however, may be sent before the transaction is committed:
https://dxr.mozilla.org/mozilla-central/rev/584870f1cbc5d060a57e147ce249f736956e2b62/toolkit/components/places/History.cpp#988

oyiptong:
Here is a summary of things that made things better:
1. use Task.jsm in tests explicitly in test-PlacesProvider.js, which doesn't make much sense, because that should happen: https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/test.js#L54
2. wait on all notifications for places inserts (3 per place) as well as the insert to complete (in theory, just the insert completion should suffice... but it didn't). As it turns out, notifications are not a good indicator of places being found on disk, because they are sent before the transaction is committed
3. promiseDBConnection and addVisit run concurrently due to using different connections, so, one must be mindful of that
4. in addVisits, moved notification promise declaration before of insert promise declaration to avoid a race condition. Doing the reverse may lead to missing some messages

It may turn out that points 1-2 are just slowing down the system so that SQLite has the time to flush to disk... who knows.

Points 3-4 are good things to know from a concurrency standpoint regardless of the findings in this issue

oyiptong:
ok ready for review.

r? @rlr, @k88hudson or @marcoscaceres

oyiptong:
thanks @mak77 for the introspective in the Places code

test/lib/PlacesTestUtils.js
```diff
@@ -24,22 +25,59 @@ const PlacesTestUtils = Object.freeze({
...
+    let urlSet = new Set(urlList);
+    let urlCount = {};
+    for (let url of urlList) {
+      urlCount[url] = 0;
+    }
+    // there are 3 notifications per place inserted
+    let totalCount = urlList.length * 3;
+    let notifCount = 0;
+
+    let notifPromise = new Promise(resolve => {
```
k88hudson: can you maybe rename this to something or add a comment about needing to wait for notifications before doing the insert, it's a bit unclear just by looking at it immediately why it's needed

k88hudson:
Ok, looks good to me! :+1:

test/lib/PlacesTestUtils.js
oyiptong: wait, the code is not self-documenting? ;-)

oyiptong:
hmm i *did* run the tests 50 times in CI. seems there's still a race condition somewhere.

oyiptong:
Solving the problem in a crude way: doing a SQL query before and after the inserts. Polling the DB until the rows are inserted.

oyiptong:
Replaced the notification checks by the table polling. It's a crude tool, but it works.

oyiptong:
going to run the test 100 times using the --times JPM command

oyiptong:
> 5000 of 5000 tests passed.
> All tests passed!

Note that I ran it 100 times before as well.

oyiptong:
doing another 100 times run

oyiptong:
> 5000 of 5000 tests passed.
> All tests passed!

two more test 100 runs passed:

https://travis-ci.org/mozilla/activity-streams/builds/108162336
https://travis-ci.org/mozilla/activity-streams/builds/108162315

oyiptong:
revision 037e71a seems to be working ok, let's try to experiment

test/lib/PlacesTestUtils.js
```diff
@@ -24,22 +26,69 @@ const PlacesTestUtils = Object.freeze({
...
-      } else if (Array.isArray(placeInfo)) {
-        places = places.concat(placeInfo);
-      } else {
-        places.push(placeInfo);
+    let places = [];
+
+    if (placeInfo instanceof Ci.nsIURI) {
+      places.push({uri: placeInfo});
+    } else if (Array.isArray(placeInfo)) {
+      places = places.concat(placeInfo);
```
marcoscaceres: Nit, I think here you want to: 
```JS
places.push(...placesInfo);
```
Otherwise, you will trash the original array object with a new one (could lead to confusion later).

test/lib/PlacesTestUtils.js
```diff
@@ -24,22 +26,69 @@ const PlacesTestUtils = Object.freeze({
...
+    } else {
+      places.push(placeInfo);
+    }
+
+    // Keep track of the number of rows for final validation
+    function* getPlacesRows() {
```
marcoscaceres: This doesn't look right to me... if you want this to run async, you will need to also wrap all this in Task.spawn:

```JS
let initialNumRows = yield Task.spawn(function* getPlacesRows() {
  const conn = yield PlacesUtils.promiseDBConnection();
  const rows = yield conn.execute("SELECT id, url, title, frecency FROM moz_places");
  return rows;
}); 
```

test/lib/PlacesTestUtils.js
```diff
@@ -11,6 +12,7 @@ XPCOMUtils.defineLazyModuleGetter(this, "PlacesUtils",
   "resource://gre/modules/PlacesUtils.jsm");
 
 const PlacesTestUtils = Object.freeze({
```
marcoscaceres: q: why Object.freeze?

test/lib/PlacesTestUtils.js
marcoscaceres: Or maybe the following, as you use getPlacesRows() below with `waitUntil()`:
```JS
function* getPlacesRows() {
  const conn = yield PlacesUtils.promiseDBConnection();
  const rows = yield conn.execute("SELECT id, url, title, frecency FROM moz_places");
  return rows;
}
yield Task.spawn(getPlacesRows())
```

test/lib/PlacesTestUtils.js
```diff
@@ -69,7 +118,28 @@ const PlacesTestUtils = Object.freeze({
...
+      let rows = yield getPlacesRows();
+      if ((rows.length - initialNumRows) === urlList.length) {
+        let rowChecked = 0;
+        for (let row of rows) {
+          if (urlSet.has(row.url) && row.frecency > -1 && row.title) {
+            // check if all have completed
+            rowChecked++;
+          }
+        }
+        if (rowChecked === urlList.legnth) {
```
marcoscaceres: Bug: typo: legnth

test/lib/PlacesTestUtils.js
```diff
@@ -69,7 +118,28 @@ const PlacesTestUtils = Object.freeze({
         }
       );
     });
-    return (yield promise);
+
+    yield notifPromise;
+    yield insertPromise;
+
+    // poll every 10ms until history items appear on disk
+    yield waitUntil(function*() {
```
marcoscaceres: This is a little convoluted, IMHO. You have a function that seems to return to different types. Could maybe be simplified to:

```JS
  yield waitUntil(function*() {
    let rows = yield getPlacesRows();
    if ((rows.length - initialNumRows) !== urlList.length) {
      return false;
    }
    let rowChecked = 0;
    rows.filter(
      row => urlSet.has(row.url) && row.frecency > -1 && row.title
    ).forEach(
      () => rowChecked++
    )
    return rowChecked === urlList.length;
  }, 10);
```

test/lib/PlacesTestUtils.js
marcoscaceres: heh, forgot about :bug: ... and :beetle: ... MUST USE MOAR EEEMOOOJJIII!! :+1:

test/lib/PlacesTestUtils.js
```diff
@@ -24,22 +26,69 @@ const PlacesTestUtils = Object.freeze({
    *            [optional] visitDate: visit date in microseconds from the epoch
    *            [optional] referrer: nsIURI of the referrer for this visit
    *          }
-   *
-   * @return {Promise}
-   *            resolves When all visits have been added successfully.
-   *            rejects JavaScript exception.
    */
   addVisits: Task.async(function*(placeInfo) {
```
marcoscaceres: Q: does it matter if a consumer sends in undefined or null as palceInfo? Is it worth checking for that?

test/lib/PlacesTestUtils.js
marcoscaceres: Fixed up code above.

test/lib/PlacesTestUtils.js
```diff
@@ -24,22 +26,69 @@ const PlacesTestUtils = Object.freeze({
...
+      let rows = yield conn.execute("SELECT id, url, title, frecency FROM moz_places");
+      return rows;
+    }
+    let initialNumRows = yield getPlacesRows().length;
+
+    // setup listeners for history events to validate db inserts and to ensure
+    // that eventual queries will lead to expected results
+    let urlList = places.map(place => place.uri.spec);
+    let urlSet = new Set(urlList);
+    let urlCount = {};
```
marcoscaceres: Nit: this really ought to be a map... ain't you happy Marcos is back? :wink:

marcoscaceres: If you wanted to use a map:  
```JS
let urlCount = Array.from(urlSet).reduce((map, next) => map.set(next, 0) , new Map());
```

test/lib/PlacesTestUtils.js
```diff
@@ -24,22 +26,69 @@ const PlacesTestUtils = Object.freeze({
...
+    for (let url of urlList) {
+      urlCount[url] = 0;
+    }
+    // there are 3 notifications per place inserted
+    let totalCount = urlList.length * 3;
+    let notifCount = 0;
+
+    let notifPromise = new Promise(resolve => {
+      function collectNotifs(aURI) {
+        if (urlSet.has(aURI.spec)) {
```
marcoscaceres: Q: if `urlSet.has(aURI.spec)` doesn't have the URL, don't we already have bigger issues (i.e., something is screwing the testing environment with either a frecency change or title change)? Maybe we should report that problem and stop?

test/lib/PlacesTestUtils.js
oyiptong: This is a straight port of https://dxr.mozilla.org/mozilla-central/rev/2dfb45d74f42d2a0010696f5fd47c7a7da94cedb/toolkit/components/places/tests/PlacesTestUtils.jsm

was done in #58

test/lib/PlacesTestUtils.js
oyiptong: i'm currently using this PR as a means to push jobs on travis. sorry about the bug.

I haven't solved the concurrency issue yet, so this is still a WIP, not review-ready yet as i thought unfortunately.

oyiptong: it's weird, but it kind of makes me wish for try servers

test/lib/PlacesTestUtils.js
oyiptong: The original code in PlacesTestUtils doesn't. We shouldn't change it if we want to maintain compatibitlity

test/lib/PlacesTestUtils.js
oyiptong: i'm unsure at this point if we care. I'll reflag for review once I fix the concurrency issue

test/lib/PlacesTestUtils.js
oyiptong: Good idea!

test/lib/PlacesTestUtils.js
oyiptong: doesn't this get flattened out anyway when running addVisits in a task?

marcoscaceres: I'm not sure it will. I would urge you to try it quickly. It may fix the concurrency issue. If I'm wrong, I owe you a beer.

oyiptong: The original code was definitely wrong. I got a chance to look at it with @mzhilyaev and it looks like i should've used `yield*` instead of `yield`

oyiptong: also, wrapping the query in a Task seems to cause less failures, though there are occasional occurrences i think, but i may be dreaming

oyiptong: it may be me who owes you a beer

marcoscaceres: Ooh, what is "yield*" do? I've never seen that before?

oyiptong: It delegates to another generator. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*

Though, if we make heavy use of generators in that way, we will need to adopt some function naming conventions so we can tell generators from regular functions

test/lib/PlacesTestUtils.js
oyiptong: I stole your code snippet. I sure am glad Marcos is back :-)

test/lib/PlacesTestUtils.js
oyiptong: I like the early return statement and the other statement types, but I think that a loop with a conditional increment is more legible (and efficient).

oyiptong:
Ok. build history is shaping up. I think there is still some kind of race condition but it happens much more rarely now. The build history is very green

oyiptong:
final r? i hope

oyiptong:
The key insight is that we need the `getPlacesRows` query to verify addVisits. `handleCompletion` being called isn't enough.

The connection we use for inserting data in sqlite and the connection we use for querying are concurrent.

test/lib/PlacesTestUtils.js
marcoscaceres: Ah, ok. Thanks!

test/lib/PlacesTestUtils.js
marcoscaceres: Ok, but I'm still not sure we want to use `yield*` here. It's going to be mega confusing (not just because of the naming, but because `yield*` injects itself mid-step creating a weird "Inception"... yes, like in the movie Inception!). 

Using Task.spawn() is cleaner IMHO, because it let's you know straight away that:

 1. it returns a promise, on which the other generator can yield (see my code again). 
 1. it immediately runs the generator.
 1. it doesn't require the generator itself to be called, or for it to even have a name - it is just passed as an argument and Task.spawn() does the right thing.  

Re: the naming... We should try to named things like "function* generateFoos()" or "function* asyncFooRunner()"... to make it more clear what a generator does.

test/lib/PlacesTestUtils.js
marcoscaceres: Re: efficiency, that assumption is incorrect. I encourage you to watch: 
https://www.youtube.com/watch?v=65-RbBwZQdU

(I'll try to find the exact time where he shows that the VM can optimize declarative calls, like `.filter()` better than imperative for loops)

marcoscaceres: I also realized that we can write `rowChecked` as:

```JS
let rowChecked = rows.filter(row => urlSet.has(row.url) && row.frecency > -1 && row.title).length;
```

6a68:
@oyiptong Hey, have you considered using a synchronous DB connection for unit testing? It might make the tests simpler, and it seems fine to hog the main thread if we're just running some tests.

If you passed a DB connection into the PlacesProvider constructor, instead of having it get its own connection from `PlacesUtils.promiseDBConnection`, then the test harness could pass it a synchronous connection, and I bet the tests would be simpler and more reliable. I guess you'd have to wrap the synchronous connection in an object with an `executeCached` function, but that seems easy enough.

Does this help at all? Happy to chat about it tomorrow, I'm guessing you're enjoying a few (or many) drinks by this point in the afternoon ;-)

test/lib/PlacesTestUtils.js
marcoscaceres: (discussed with @oyiptong already... adding for record keeping)

Just one more note on `yield*`: calling it will run the generator, but, AFAICT, those DB promises will not being handled be handled by any promise-resolving machinery (because you need something like Task.spawn/async to handle passing the resolved value back into the generator).

marcoscaceres:
R+ from me.

oyiptong:
Hey @6a68, thanks for the suggestion. The code that adds visits has been lifted from gecko, and I'm unsure why we should be having errors at all; looking at the C++ code, the `handleComplete` message should only return when the transaction committed.

@mak77 also took a look and at a glance didn't see anything wrong.

Sync addVisits would diminish the amount of checking we need to do in the test, but since we're polling every 10ms for results with a promiseDBConnection, and only returning when we do, *in theory*, we should be safe, right? ;-) it looks like just not yet.

Either way, good idea. I'll revisit later with sync addVisits. We have a milestone due next week, and I need to be focusing on other things for now.

oyiptong:
Ah, @marcoscaceres wants to give it a shot today. Perhaps we can incorporate @6a68's idea of implementing sync addVisits? I still have no clue why this is failing for us here :-/

marcoscaceres:
Filed issue #93 for the intermittent failures.

Issue #35, opened 2016-02-02
Create "spotlight" component


k88hudson:
![image](https://cloud.githubusercontent.com/assets/1455535/12763708/ec210636-c9c3-11e5-85e2-ad37d3817aa2.png)

Should include lead image, title, and description/short summary

tspurway:
@nchapman to donate some rules for this

nchapman:
Here are the rules that Bryan and I came up with:

1. Up to 1 synced tab that was active in the last hour
2. Up to 2 bookmarks with images from the past 24 hours
3. Fill the rest with recent history items that have images

If no data is available, hide the spotlight section for now.

rlr:
Landed the react component 1d8ad808ca9a561f4bd44de4f6a202f26e04a56b

We'll need a new issue for the "backend logic" where we'll apply the rules.

PR #72, opened 2016-02-09
feat(content): #35 Implement Spotlight component

rlr:


content-src/components/Base/Base.js
```diff
@@ -5,6 +5,10 @@ const {actions} = require("actions/action-manager");
 const Header = require("components/Header/Header");
 const TopSites = require("components/TopSites/TopSites");
 const ActivityFeed = require("components/ActivityFeed/ActivityFeed");
+const Spotlight = require("components/Spotlight/Spotlight");
+
+// TODO: replace this with the appropriate actions/reducers/shim.
+const fakeSpotlightItems = require("lib/shim").data.fakeSpotlightItems;
```
rlr: I wasn't sure if we were going to need new actions or reuse the existing ones so just did this for now...

rlr:
not sure why travis isn't running on this PR

k88hudson:
I think it might be backed up at the moment

content-src/components/Spotlight/Spotlight.scss
```diff
@@ -0,0 +1,56 @@
+.spotlight {
```
k88hudson: The overflow-y: auto on .new-tab-wrapper is cutting off the bottom of the shadows on these, can you just remove that rule from `NewTabPage/NewTabPage.scss`? We don't want it to scroll anyway i think

content-src/components/Spotlight/Spotlight.scss
```diff
@@ -0,0 +1,56 @@
...
+    a {
+      color: $dark-grey;
+    }
+  }
+
+  .spotlight-info {
+    padding: 10px;
+  }
+
+  .spotlight-description {
```
k88hudson: i think this would be better at font size 14px

rlr: btw, is there a way to have like `text-overflow: ellipses;` work on multiline text like this? So we could have max 3 lines or something

k88hudson: unfortunately not :(

k88hudson:
R+ after those changes and a rebase. Note that the tests are failing now due to unrelated reasons (olivier is fixing currently)

PR #78, opened 2016-02-10
chore(react): Convert React.createClass to extends React.Component

pdehaan:
Following guidelines from http://www.newmediacampaigns.com/blog/refactoring-react-components-to-es6-classes:

> Starting with React 0.13, defining components using ES6 classes is encouraged.

I haven't investigated yet to see if some/most of these could be converted to a [Stateless Function](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions).

Also attempted to replace all the `require()` and `module.exports` with the more ES2015-esque `import` and `export`, but oh boy that went over poorly locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
-node_modules
-npm-debug.log
-.DS_Store
 *.sw[po]
 *.xpi
-logs/
-dist/
+.DS_Store
+config.yml
 data/content/
+dist/
 firefox/
+logs/
+node_modules
+npm-debug.log
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -1,6 +1,14 @@
 const React = require("react");
 
-const ActivityFeedItem = React.createClass({
+class ActivityFeedItem extends React.Component {
+  static get propTypes() {
+    return {
+      image: React.PropTypes.string, // TODO: Have fallback for this, change to icon
+      title: React.PropTypes.string.isRequired,
+      url: React.PropTypes.string.isRequired
+    };
+  }
+
   render() {
     const site = this.props;
     return (<li className="feed-item">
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -1,9 +1,10 @@
-node_modules
-npm-debug.log
-.DS_Store
 *.sw[po]
 *.xpi
-logs/
-dist/
+.DS_Store
+config.yml
```
pdehaan: Added this "config.yml" bit. The rest of the noise was A-Z sorting for "easier" readability.

content-src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -17,26 +25,22 @@ const ActivityFeedItem = React.createClass({
...
+  static get propTypes() {
+    return {
+      sites: React.PropTypes.array.isRequired
+    };
+  }
 
-const ActivityFeed = React.createClass({
   render() {
-    const {props} = this;
+    const props = this.props;
```
pdehaan: Tweaked this for consistency with the rest of the codebase. I think this was the only instance i saw where we destructured.

k88hudson:
The only reason I don't like using the es6 React classes is that (1) they don't *really* add anything spectacularly better than React.createClass and (2) they don't support mixins, which are occasionally needed to use libraries or for certain use cases, and it's kind of a pain to add decorators just for that reason... although I have to admit, since 0.14 those cases happen MUCH less often. :stuck_out_tongue_winking_eye: 

As for `import`, I definitely went back and forth on this for awhile, but I generally prefer to stick to require for now since import is not actually natively supported anywhere yet, and babel compiles import down to require anyway 2. There are also cases where dealing with the weback require context is needed (such as bulk importing from a whole folder), and I like to keep things consistent.

pdehaan:
Sounds good to me. Closing this PR for now until we decide React.Component and classes are a worthy improvement.

PR #80, opened 2016-02-10
style(content): Minor style improvements, font-awesome

k88hudson:
Adds font-awesome (for now) and some other little style improvements

### DIFF
--- a/content-src/static/activity-streams.html
+++ b/content-src/static/activity-streams.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf8">
     <link rel="stylesheet" href="http://localhost:1963/main.css" />
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
   </head>
   <body>
     <div id="root"></div>
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ logs/
 dist/
 data/content/
 firefox/
+config.yml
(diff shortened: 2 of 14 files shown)


k88hudson:
R? @rlr

rlr:
One error fail looks related:

```
FAILED TESTS:

  Header

    ✖ "before each" hook for "should not have an img element if no user image is provided"

      Firefox 47.0.0 (Linux 0.0.0)

    props.currentRoute is undefined
```

rlr:
Otherwise it looks awesome. r+ with test fix

content-src/static/activity-streams.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf8">
     <link rel="stylesheet" href="http://localhost:1963/main.css" />
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
```
pdehaan: We may be able to use a tool like https://icomoon.io/app/#/select/font to build a font-set with the 2-3 icons that we need instead of grabbing the full font family (even though I think the above file is only about 7.4KB minimized).

Ref: https://github.com/FortAwesome/Font-Awesome/wiki/Customize-Font-Awesome

k88hudson: Definitely, I think we'll actually not want to use font-awesome at all, this is just a super hacky solution for this first milestone / before we get icon assets

k88hudson: Filed https://github.com/mozilla/activity-streams/issues/83 so we don't forget

Issue #87, opened 2016-02-10
Replace nav on home with link to Timeline view


pdehaan:
We added some font-awesome icons in #80, but looks like there isn't an icon for the Home page, only the TImeline page (and the "All" and "Bookmarks" sections). This seems a bit inconsistent and causes an odd `undefined` to appear in the CSS.

<img width="404" alt="nightly" src="https://cloud.githubusercontent.com/assets/557895/12965907/cfb52fd0-d00e-11e5-9420-68ad08dddb44.png">
**Figure 1:** Home page.

---

<img width="435" alt="nightly_and_unravel_review__a_heartwarming_mix_of_challenge_and_charm___the_verge" src="https://cloud.githubusercontent.com/assets/557895/12965914/e2efd3ac-d00e-11e5-8288-4f55b826cc46.png">
**Figure 2:** Timeline page (with icon and child page icons).

---

<img width="661" alt="nightly" src="https://cloud.githubusercontent.com/assets/557895/12965935/0e95a9e6-d00f-11e5-9243-5f492e7dc2ed.png">
**Figure 3:** Home page `undefined` CSS class for font-awesome icon.

pdehaan:
Huh... We could add an `icon` property to the `<IndexRoute ...>` in [/content-src/components/Routes/Routes.js:11](https://github.com/mozilla/activity-streams/blob/9e9e50ac099d1e1437ded431f8f8a7bd63fad277/content-src/components/Routes/Routes.js#L11), but it looks like the "fa-home" icon is a few pixels narrower than the `fa-timeline` icon, so the UI still shifts slightly.

k88hudson:
The designs actually didn't spec a home icon, although the latest iteration removes the top bar altogether, so this will need some clarification.

Although... rendering undefined into the className is definitely :-1:,  probably about time we added `classnames`

pdehaan:
I did a cheap fix and defaulted a missing/undefined icon to an empty string in https://github.com/mozilla/activity-streams/pull/92#discussion-diff-52556840. :shrug:

k88hudson:
Instead of showing the top nav on `Home` we are going to have a button that links to Activity Streams

@bryanbell to determine icon

k88hudson:
Closing since new UI no longer has a header.

PR #92, opened 2016-02-11
chore(content): Convert key to numeric instead of string values to shorten generated ids

### LINKED ISSUE #91: Evaluate `key` attributes on all react components
I was just devTool's-in around trying to determine why there was no hover-state on the Spotlight item's title or a clickable image, when I noticed that we're using `key={site.url}` for `<SpotlightItem>` and therefore our reactid's are absurdly long.

We may want to reel that back slightly and use a numerical index instead.
I traced it back to [/content-src/components/Spotlight/Spotlight.js:37](https://github.com/mozilla/activity-streams/blob/9e9e50ac099d1e1437ded431f8f8a7bd63fad277/content-src/components/Spotlight/Spotlight.js#L37).

Changing it from this:
```js
{props.sites.map(site => <SpotlightItem key={site.url} {...site} />)}
```

... to this:
```js
{props.sites.map((site, idx) => <SpotlightItem key={idx} {...site} />)}
```

Seems to help out greatly. This is a super-nit, but I know that if I don't file it now while I see it, I'll probably forget forever. 

---

nightly_and_activity-streams_spotlight_js_at_1d8ad808ca9a561f4bd44de4f6a202f26e04a56b_ _mozilla_activity-streams_and_activity-streams_routes_js_at_9e9e50ac099d1e1437ded431f8f8a7bd63fad277_ _mozilla_activity-streams_and_theintern_intern__a_n
**Figure 1:** Before: `data-reactid=".0.1.0.0.1.1.$http=2//www=1cnn=1com/videos/tv/2016/02/09/man-throws-alligator-in-wendys-wptv-dnt=1cnn"`

---

nightly_and_spotlight_js_ _activity-streams_and_activity-streams_spotlight_js_at_9e9e50ac099d1e1437ded431f8f8a7bd63fad277_ _mozilla_activity-streams
**Figure 2:** After: `data-reactid=".0.1.0.0.1.1.$0"`

pdehaan:
Fixes #91

### DIFF
--- a/content-src/components/Header/Header.js
+++ b/content-src/components/Header/Header.js
@@ -10,7 +10,7 @@ const Header = React.createClass({
     return (<header className="head">
       <section className="nav" onClick={() => this.setState({showDropdown: !this.state.showDropdown})}>
         <h1>
-          <span hidden={!props.currentRoute.icon} className={`icon fa ${props.currentRoute.icon}`} />
+          <span hidden={!props.currentRoute.icon} className={`icon fa ${props.currentRoute.icon || ""}`} />
           <span>{props.currentRoute.title}</span>
           <span className="arrow fa fa-chevron-down" />
         </h1>
--- a/content-src/components/TopSites/TopSites.js
+++ b/content-src/components/TopSites/TopSites.js
@@ -6,7 +6,7 @@ const TopSites = React.createClass({
     return (<section className="top-sites">
       <h3 className="section-title">Top Sites</h3>
       <div className="tiles-wrapper">
-        {props.sites.map((site) => (<a key={site.url} className="tile" href={site.url}>
+        {props.sites.map((site, idx) => (<a key={idx} className="tile" href={site.url}>
           <div className="tile-img-container">
             {(site.leadImage || site.image) && <div className="tile-img"
               style={{backgroundImage: `url(${site.leadImage || site.image})`}} />}
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -29,7 +29,7 @@ const ActivityFeed = React.createClass({
   render() {
     const {props} = this;
     return (<ul className="activity-feed">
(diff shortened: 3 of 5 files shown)


content-src/components/Header/Header.js
```diff
@@ -10,7 +10,7 @@ const Header = React.createClass({
     return (<header className="head">
       <section className="nav" onClick={() => this.setState({showDropdown: !this.state.showDropdown})}>
         <h1>
-          <span hidden={!props.currentRoute.icon} className={`icon fa ${props.currentRoute.icon}`} />
+          <span hidden={!props.currentRoute.icon} className={`icon fa ${props.currentRoute.icon || ""}`} />
```
pdehaan: This fixes the `undefined` className issue in #87 by defaulting to an empty string.

marcoscaceres: :+1:

content-src/components/TopSites/TopSites.js
```diff
@@ -6,7 +6,7 @@ const TopSites = React.createClass({
     return (<section className="top-sites">
       <h3 className="section-title">Top Sites</h3>
       <div className="tiles-wrapper">
-        {props.sites.map((site) => (<a key={site.url} className="tile" href={site.url}>
+        {props.sites.map((site, idx) => (<a key={idx} className="tile" href={site.url}>
```
marcoscaceres: Q: just wondering, what's the key used for?

pdehaan: Actually, using `idx` may not be good enough since using a numerical index may not be unique enough:

> The situation gets more complicated when the children are shuffled around (as in search results) or if new components are added onto the front of the list (as in streams). In these cases where the identity and state of each child must be maintained across render passes, you can uniquely identify each child by assigning it a `key`:
> &mdash; http://facebook.github.io/react/docs/multiple-components.html#dynamic-children

pdehaan:
I'm actually just going to close this. I'm not convinced that using the array index is unique enough here. And looking at the returned data, it seems like the `url` may be the best/only unique thing we can use, since I didn't see any UUIDs or GUIDs returned.

Issue #84, opened 2016-02-10
npm run bundle should generate html without links to localhost


k88hudson:
When we generate a build for production, the html page should not link to `localhost:1963`; it should link to `main.css` and `main.js` instead.

We need to replace the static index file with a generate-html page instead, which can be something like: https://github.com/mozilla/remote-newtab/blob/master/bin/generate-html.js

rlr:
this landed yesterday :rocket:

PR #96, opened 2016-02-11
chore(scripts) #84 script to generate html

rlr:
Do we want to create a `tests-node` setup like RNT?

r? @k88hudson


rlr:
I am pretty sure that failed test is the intermittent issue being worked on:

```
45 of 46 tests passed.

There were test failures...

The following tests failed:

  ./test/test-PlacesProvider.test_Links_getTopFrecentSites_Order: failure
```

bin/generate-html.js
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env node
+"use strict";
+const defaults = {
+  baseUrl: "./"
```
pdehaan: why not just a default of `""`?

bin/generate-html.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+function template(rawOptions) {
+  const options = Object.assign({}, defaults, rawOptions || {});
+  return `
+<!doctype html>
+<html lang=en-us>
+  <head>
+    <meta charset="utf8">
```
pdehaan: > Error: Bad value `utf8` for attribute `charset` on element meta: `utf8` is not a preferred encoding name. The preferred label for this encoding is `utf-8`.
> &mdash; via https://html5.validator.nu/

Also, not sure if we want to add a dummy `<title>` tag:

> Error: Element `head` is missing a required instance of child element `title`.
From line 7, column 3; to line 7, column 9

yamscripts.yml
```diff
@@ -10,9 +10,10 @@ scripts:
 
 # bundle: Bundle assets for production
   bundle:
-    static: cpx "content-src/static/**/*" data/content -C
+    static: cpx "content-src/static/**/!(activity-streams.html)" data/content -C
```
k88hudson: you shouldn't need this if you're deleting the file

yamscripts.yml
```diff
@@ -20,6 +21,7 @@ scripts:
     static: =>bundle:static -- -w
     webpack: webpack -w
     css: =>bundle:css && =>bundle:css -- -w
+    html: =>bundle:html -- -b http://localhost:1963/
```
k88hudson: i think it's actually a bit clearer to have `--baseUrl` despite being longer

bin/generate-html.js
k88hudson: :+1:

bin/generate-html.js
k88hudson: :+1:  for `""`

k88hudson:
@rlr let's not add `node-tests` for now since we're pretty tight on time; we can rely on the `npm bundle` task failing on travis to tell us if something is severely wrong. Feel free to file a  "build script tests" with the "Nice but not critical" tag if someone wants to grab it/has an extra few minutes.

yamscripts.yml
rlr: the reason I had to add this is that the file was getting deleted by cpx because of the `-C` argument. There is sort of a race condition and I was losing it. Any other ideas?

bin/generate-html.js
```diff
@@ -0,0 +1,33 @@
+#! /usr/bin/env node
+"use strict";
+const defaults = {
+  baseUrl: "./"
+};
+
+function template(rawOptions) {
+  const options = Object.assign({}, defaults, rawOptions || {});
```
pdehaan: Not sure if you want to just default the `rawOptions` on L7 to have a default value of `{}` and then you probably wouldn't need the `|| {}` here.

rlr:
Feedback applied ^^.

yamscripts.yml
k88hudson: Ah! got it. I would remove the `-C` and add a `clean` task instead that runs synchronously

PR #102, opened 2016-02-11
chore(travis): Temporarily disable jpm tests due to #92

k88hudson:
I'd like to disable our `jpm` tests temporarily and rely on running them locally until we resolve the race condition issues/travis issues, so we don't have unrelated failures on other PRs in the mean time.

Are you ok with this @rlr @oyiptong @emtwo @mzhilyaev?


oyiptong:
I'm of two minds about this. Perhaps we can re-trigger if there's a failure for now?
We lose a pretty valuable test if we remove the JPM tests.

Let's see if it gets more annoying.

Issue #65, opened 2016-02-08
Places data API: recent links


oyiptong:
Obtains recent links. Similar to top frecent sites, except ordered by `last_visit_date`.

Unlike top frecent sites, it returns a list of most recent links and does not group by `rev_host`

oyiptong:
ActivityStreams needs to listen to a `TOP_RECENT_LINKS_REQUEST' message, and respond with a `TOP_RECENT_LINKS_RESPONSE`.

The implementation of the query should be in PlacesProvider.js, the listening of the message and the formulation of a response in ActivityStreams.js.

An entry for the message types should be made in: https://github.com/mozilla/activity-streams/blob/master/content-src/actions/action-manager.js

oyiptong:
Also, we also need to extract the favicons and include it in the query

PR #111, opened 2016-02-12
feat(recentlinks): #65 recent links query and emitters

mzhilyaev:
fixed reviewer comments and rebased on top bookmarks_API implementation

### DIFF
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -170,6 +170,39 @@ Links.prototype = {
   },
 
   /**
+   * Gets the top recent links
+   *
+   * @param {Integer} options
+   *                  limit: Maximum number of results to return. Max 100.
+   *
+   * @returns {Promise} Returns a promise with the array of links as payload.
+   */
+  getRecentLinks: Task.async(function*(options={}) {
+
+    let {limit} = options;
+    if (!limit || limit.options > HISTORY_RESULTS_LIMIT) {
+      limit = HISTORY_RESULTS_LIMIT;
+    }
+
+    let sqlQuery =  `SELECT moz_places.url as url, moz_favicons.data as favicon, mime_type as mimeType,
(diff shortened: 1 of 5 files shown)


lib/PlacesProvider.js
```diff
@@ -170,6 +170,43 @@ Links.prototype = {
...
+   * @param {Integer} options
+   *                  limit: Maximum number of results to return. Max 100.
+   *
+   * @returns {Promise} Returns a promise with the array of links as payload.
+   */
+  getRecentLinks: Task.async(function*(options) {
+
+    if (!options) {
+      options = {};
+    }
```
pdehaan: Why not just do `function *(options = {}) {` above instead of the extra check?

test/test-PlacesProvider.js
```diff
@@ -100,6 +100,53 @@ exports.test_Links_getTopFrecentSites_Order = function*(assert) {
...
+  };
+  yield PlacesTestUtils.addVisits(visits, faviconData);
+  yield PlacesTestUtils.addFavicons(faviconData);
+
+  links = yield provider.getRecentLinks();
+  assert.equal(links.length, visits.length, "number of links added is the same as obtain by getRecentLinks");
+  assert.equal(links[0].url, "https://mozilla2.com/1", "Expected 1-st link");
+  assert.equal(links[1].url, "https://mozilla1.com/0", "Expected 1-st link");
+  assert.equal(links[2].url, "https://mozilla3.com/2", "Expected 1-st link");
+  assert.equal(links[3].url, "https://mozilla4.com/3", "Expected 1-st link");
```
pdehaan: Are all these "Expected ** link" strings correct? Seems confusing if a test fails you'd never exactly know what line. Not sure if it should be "Expected 1st link", "Expected 2nd link", etc.

mzhilyaev:
cancelling again to merge against master now

PR #116, opened 2016-02-12
feat(recentlinks): #65 recent links query and emitters

mzhilyaev:


lib/ActivityStreams.js
```diff
@@ -66,6 +66,11 @@ ActivityStreams.prototype = {
           this.send(am.actions.Response("RECENT_BOOKMARKS_RESPONSE", links), worker);
         });
         break;
+      case am.type("RECENT_LIKS_REQUEST"):
```
pdehaan: s/LIKS/LINKS/

We should almost revisit the "should every `switch` have a `default` case to catch unexpected things.
I have no idea how/if something like this would be catchable via any linter.
Although we could go the old Adobe Flex route of having constants for event names and stuff. So instead of static strings like "RECENT_LINKS_REQUEST" we'd export a const like `am.responses.RECENT_LINKS_RESPONSE = "RECENT_LINKS_RESPONSE";` which could be whatever string value we want. That way any linter would catch typos.

oyiptong: am.type should validate that typo (or rather, invalidate it)

oyiptong: but it obviously didn't. @k88hudson what was your intention behind the action manager?

oyiptong: We should make it throw

test/test-PlacesProvider.js
```diff
@@ -100,6 +100,53 @@ exports.test_Links_getTopFrecentSites_Order = function*(assert) {
...
+exports.test_Links_getRecentLinks = function*(assert) {
+  let provider = PlacesProvider.links;
+  let {
+    TRANSITION_TYPED,
+    TRANSITION_LINK
+  } = PlacesUtils.history;
+
+  function timeDaysAgo(numDays) {
+    let now = new Date();
+    return now.getTime() - (numDays * 24 * 60 * 60 * 1000);
```
pdehaan: Since we're now doing this in more than 1x place, not sure if we want to make this magic math into a const like `const DAYS_IN_MS = 24 * 60 * 60 * 1000;` and then convert this to:

```js
return now.getTime() - (numDays * DAYS_IN_MS);
```

lib/ActivityStreams.js
oyiptong: Actually, this code is not being executed, because we aren't receiving messages yet from content

test/test-PlacesProvider.js
oyiptong: I agree

test/test-PlacesProvider.js
```diff
@@ -100,6 +100,53 @@ exports.test_Links_getTopFrecentSites_Order = function*(assert) {
...
+  };
+  yield PlacesTestUtils.addVisits(visits, faviconData);
+  yield PlacesTestUtils.addFavicons(faviconData);
+
+  links = yield provider.getRecentLinks();
+  assert.equal(links.length, visits.length, "number of links added is the same as obtain by getRecentLinks");
+  assert.equal(links[0].url, "https://mozilla2.com/1", "Expected 1-st link");
+  assert.equal(links[1].url, "https://mozilla1.com/0", "Expected 2-st link");
+  assert.equal(links[2].url, "https://mozilla3.com/2", "Expected 3-st link");
+  assert.equal(links[3].url, "https://mozilla4.com/3", "Expected 4-st link");
```
pdehaan: Nit: "2st", "3st" and "4st" aren't things.
Not sure if these should be "2nd", "3rd", and "4th".

(disclaimer: I'm a jerk)

mzhilyaev: thirst is a thing

content-src/components/Base/Base.js
```diff
@@ -9,6 +9,9 @@ const Base = React.createClass({
     // This should work!
     this.props.dispatch(actions.RequestTopFrecent());
 
+    // This should work!
+    this.props.dispatch(actions.RequestLinks());
```
oyiptong: RequestRecentLinks

pdehaan:
**Protip:** I know everybody knows this, but if you put "Closes #65" in your original comment above, GitHub will auto-close the linked issue when this gets merged.

It will literally save you _seconds_ a day! :tada:

oyiptong:
r+

oyiptong:
@pdehaan you're awesome

pdehaan:
@oyiptong Don't worry. I get worse.
Fact: @nchapman only keeps me around because I occasionally have whimsical GIFs.

---

content-src/components/Base/Base.js
oyiptong: Let's keep it

oyiptong:
lol

Issue #112, opened 2016-02-12
Add more ESLint rules


pdehaan:
Ref: https://github.com/mozilla/activity-streams/pull/110#discussion_r52712168

This seems like a good start: https://github.com/walmartlabs/eslint-config-defaults

I also extended "defaults" and "airbnb/es6", but then had to fiddle a bunch of individual rules to get things passing and a good set of rules.

Just filing this to get some feedback before I enter the merge hell that is touching large number of files with adding spaces or extra parens (protip: `eslint . --fix` is your friend).

```js
{
  // When adding items to this file please check for effects on sub-directories.
  "plugins": [
    "mozilla",
    "react"
  ],
  "extends": [
    "eslint:recommended",
    "defaults",
    "defaults/configurations/airbnb/es6"
  ],
  "rules": {
    "mozilla/components-imports": 1,
    "mozilla/import-globals-from": 1,
    "mozilla/this-top-level-scope": 1,

    "react/jsx-uses-react": 1,

    "array-bracket-spacing": [2, "never"],
    "arrow-body-style": 0,
    "arrow-parens": [2, "always"],
    "camelcase": 0,
    "comma-dangle": 0,
    "computed-property-spacing": [2, "never"],
    "default-case": 0, // TODO: Remove this.
    "eqeqeq": [2, "smart"],
    "func-names": 0,
    "generator-star-spacing": [2, {"before": true, "after": false}],
    "max-len": 0,
    "new-cap": 0,
    "no-console": 1,
    "no-param-reassign": 0,
    "no-shadow": 0, // TODO: Set to warning or error?
    "no-trailing-spaces": [2, {"skipBlankLines": false}],
    "no-undef": 2,
    "no-unused-vars": [2, {"vars": "all", "args": "none"}],
    "no-use-before-define": [2, "nofunc"],
    "no-var": 1,
    "object-curly-spacing": [2, "never"],
    "prefer-arrow-callback": 0,
    "prefer-const": 0, // TODO: Remove this.
    "quote-props": 0,
    "quotes": [2, "double"],
    "semi": [2, "always"],
    "space-before-function-paren": [2, {"anonymous": "never", "named": "never"}],
    "strict": 0
  },
  "ecmaFeatures": {
    "jsx": true
  },
  "env": {
    "browser": true,
    "es6": true,
    "mocha": true,
    "node": true,
  },
  "globals": {
    "StopIteration": true,
    "__CONFIG__": true
  }
}
```

[...]

Issue #60, opened 2016-02-05
Add React-specific rules to eslint


k88hudson:
It would be nice to have a few more:

https://github.com/yannickcr/eslint-plugin-react

PR #118, opened 2016-02-15
chore(lint): Add ESLint strict and react-strict presets

pdehaan:
Oy vey!

Per #60 and #112, I added the [**eslint-config-strict**](http://npm.im/eslint-config-strict) and [**eslint-config-strict-react**](http://npm.im/eslint-config-strict-react) presets.
Most of the .eslintrc noise is me relaxing the strict rules to make everything pass.

There be lots of noise in the diff, but the general gist of the rules is:

1. params in arrow callbacks should be wrapped in parens. ie:
  ```js
(foo) => {}
```

2. params in template strings should have some spaces between the `{` and `}`. ie:
  ```js
console.log(`hello ${ name }`);
```

3. We prefer string templates instead of concatenation.

---

As always, please feel free to close this for now as it's probably a pretty disruptive diff.


content-src/components/Header/Header.js
```diff
@@ -24,8 +24,7 @@ const Header = React.createClass({
         {props.userName && <span>
           {props.userName}
         </span>}
-        {props.userImage && <img alt=""
-          src={props.userImage} />}
+        {props.userImage && <img alt="" src={props.userImage} />}
```
pdehaan: ESLint really didn't like having the `` open on one line and and closing on another, and I couldn't figure out how it wanted me to position that closing `/>`. So I cheated and made it a single line since it was short.

content-src/components/TimelinePage/TimelinePage.js
```diff
@@ -17,13 +17,11 @@ const TimelinePage = React.createClass({
...
-              <a className={item.active ? "active" : ""}>
-                <span className={`fa ${item.icon}`} /> {item.title}
-              </a>
-            </li>);
-          })}
+          {navItems.map((item) => (<li key={item.title}>
+            <a className={item.active ? "active" : ""}>
+              <span className={`fa ${ item.icon }`} /> {item.title}
+            </a>
+          </li>))}
```
pdehaan: I forget what rule it was, but ESLint was complaining that we were using a `return` statement as the only statement in an arrow function and it preferred this marginally less verbose syntax. :shrug:

content-src/components/TopSites/TopSites.js
```diff
@@ -8,8 +8,8 @@ const TopSites = React.createClass({
       <div className="tiles-wrapper">
         {props.sites.map((site) => (<a key={site.url} className="tile" href={site.url}>
           <div className="tile-img-container">
-            {(site.leadImage || site.image) && <div className="tile-img"
-              style={{backgroundImage: `url(${site.leadImage || site.image})`}} />}
+            {(site.leadImage || site.image) &&
+              <div className="tile-img" style={{backgroundImage: `url(${ site.leadImage || site.image })`}} />}
```
pdehaan: Same as above, I couldn't figure out exactly how ESLint wanted me to format the `/>` since the `<div>` opened on one line and closed on another, so I put all on one line.

content-src/lib/ReduxChannel.js
```diff
@@ -42,16 +42,17 @@ class Channel {
...
   get middleware() {
-    return store => next => function(action) {
+    return (store) => (next) => function(action) {
       const meta = action.meta || {};
       const timeouts = this.timeouts;
 
       // Send action to the next step in the middleware
-      next(action);
+      // TODO: "Expected return with your callback function"
+      next(action); // eslint-disable-line callback-return
```
pdehaan: This was an interesting one. ESLint seems to think that since we're calling `next()` we should have a return statement here (which sounds solid if we're talking about middleware, callbacks, or promises). I'm not familiar with the code to know if this is a bug or not, so I just disabled the rule here and added a "TODO".

package.json
```diff
@@ -39,7 +39,10 @@
     "chai": "^3.5.0",
     "chai-as-promised": "^5.2.0",
     "cpx": "^1.2.1",
-    "eslint": "^1.10.3",
+    "eslint": "2.0.0",
```
pdehaan: Also bumps to **eslint@2**, because why not.

pdehaan:
I did a bunch of this work in #218.
It should now be easier to add React specific rules or other strict rules in the future when we want stricter rules.

Closing.

PR #122, opened 2016-02-16
fix(content): Only show sites in Spotlight if it has atleast one image

### LINKED ISSUE #121: Spotlight should only show sites with images
We should filter out sites without lead images

k88hudson:
This should fix an existing bug where Spotlights don't show up at all if the `images` prop is missing

Fixes #121


rlr:
LGTM R+

PR #124, opened 2016-02-16
fix(addon): #123 Fix include path to include hashes

### LINKED ISSUE #123: Add-on messaging doesn't work after refresh
After refreshing the page, messages no longer work with the following error in the console

```
Error: Couldn't find the worker to receive this message.
The script may not be initialized yet, or may already have been unloaded.
```

k88hudson:
This fixes a bug where if you reload the page, new workers don't get attached, and also properly removes workers when the tab is reloaded/closed

Fixes #123


k88hudson:
@oyiptong @emtwo R?

test/test-setup.js
```diff
@@ -0,0 +1,100 @@
...
+const PORT = 8099;
+const path = "/dummy-activitystreams.html";
+const url = `http://localhost:${PORT}${path}`;
+const otherUrl = `http://localhost:${PORT}/dummy-other.html`;
+
+let srv;
+let app;
+let openTabs;
+
+function beforeEach() {
```
oyiptong: Use the addon sdk's before and after test methods

test/test-setup.js
```diff
@@ -0,0 +1,100 @@
...
+  app = new ActivityStreams({pageURL: url});
+  openTabs = [];
+
+  function onOpen(tab) {
+    openTabs.push(tab);
+  }
+  tabs.on("open", onOpen);
+}
+
+function afterEachPromise() {
```
oyiptong: Ditto

test/test-setup.js
```diff
@@ -0,0 +1,100 @@
...
+  beforeEach();
+  yield new Promise(resolve => {
+    let worker;
+    function onReady(tab) {
+      if (tab.url === url) {
+        assert.equal(app.workers.size, 1);
+        worker = getFirstWorker();
+        worker.port.on("content-to-addon", msg => {
+          if (msg.type === "pagehide") {
+            assert.equal(app.workers.size, 0, "Worker should be unloaded when a new url is loaded in the tab");
```
oyiptong: There might be a race condition here, depending on when the listener is registered. Might be better to listen to a tab event if we can then check the worker if posdible

test/test-setup.js
```diff
@@ -0,0 +1,100 @@
...
+  beforeEach();
+  yield new Promise(resolve => {
+    function onReady(tab) {
+      if (tab.url === url) {
+        assert.equal(app.workers.size, 1, "Worker is loaded");
+      }
+      resolve();
+    }
+    tabs.open({url, onReady});
+  });
```
oyiptong: Maybe a better test would be to have no tabs, test the number of workers (zero), open a new tab, test (one worker).

Same applies for other tests

test/test-setup.js
oyiptong: Hmm actually, I take that back. The activity streams object attaches the worker listener on init, and this listener should come second in the queue. One quick check of the port implementation would confirm.

The test then checks if page hide is sent or hangs, which is good

lib/ActivityStreams.js
```diff
@@ -129,11 +129,17 @@ ActivityStreams.prototype = {
     // `this` here refers to the object instance
     this.workers = new Set();
     this._pagemod = new PageMod({
-      include: [this.options.pageURL],
+      include: [this.options.pageURL + "*"],
```
oyiptong: :+1 finding this!

lib/ActivityStreams.js
```diff
@@ -129,11 +129,17 @@ ActivityStreams.prototype = {
...
       contentScriptFile: data.url("content-bridge.js"),
       contentScriptWhen: "start",
-      attachTo: "top",
+      attachTo: ["existing", "top"],
       onAttach: worker => {
+
+        // This detaches workers on reload or closing the tab
+        worker.on("detach", () => {
```
oyiptong: +1 :-)

test/test-setup.js
```diff
@@ -0,0 +1,89 @@
...
+    if (tab.url === url) {
+      assert.equal(app.workers.size, 1, "Worker is loaded");
+    }
+    done();
+  }
+  assert.equal(app.workers.size, 0, "app.workers start at 0");
+  tabs.open({url, onReady});
+};
+
+exports["test removing worker on url change"] = function(assert, done) {
```
oyiptong: why not generator style?

k88hudson: there's not really a point to having a generator now since it's a single async function, it's less code/easier to read this way

k88hudson: i'm happy to use it if you need yield more than one promise, but yeah, it's just not needed in this case

k88hudson:
I noticed the add-on button wasn't getting destroyed, so I added that in as well with a separate commit. @oyiptong r?

oyiptong:
r+ after CI is fixed

pdehaan:
@k88hudson @oyiptong Is this ready to merge?

oyiptong:
I think @k88hudson is working on this and should merge whenever she's ready

Issue #28, opened 2016-02-02
Engagement data collection


oyiptong:
Implement code in add-on to send messages to data-pipeline.

Maybe start with a few messages add-on side, we can do content in a follow-up ticket

oyiptong:
* Number of newtab opens per user

oyiptong:
* scroll depth

k88hudson:
@oyiptong @emtwo to clarify

oyiptong:
following a meeting with @nchapman, @tspurway jotted down some notes in #27.

The pings should be as follows:

When a user opens a newtab page, we make note of the time (start a counter).

We then capture an exit condition:

1. a user clicks on a link
2. a user enters something in the url bar and hits enter
3. a user closes the tab

In case #1, when that happens, we want to record:

1. the maximum scroll depth reached
2. the y-position of the link in px
3. the type of link they clicked (top sites, bookmark, etc)
4. the time they spent on the page

in case #2 we only want to record an impression, but that they used the URL bar

in case #3 we only want to record an impression, but that they closed it

There may be more things to record, or more use cases, but this should give a good idea

PR #127, opened 2016-02-17
feature(addon): #28 Tab open time data collection.

emtwo:
r? @oyiptong @k88hudson

### DIFF
--- a/lib/TabTracker.js
+++ b/lib/TabTracker.js
@@ -1,0 +1,103 @@
+const tabs = require("sdk/tabs");
+const ss = require("sdk/simple-storage");
+
+function TabTracker(trackableURLs) {
+  if (!ss.storage.tabData) {
+    ss.storage.tabData = {};
+  }
+  for (let trackableURL of trackableURLs) {
+    if (!ss.storage.tabData[trackableURL]) {
+      ss.storage.tabData[trackableURL] = {};
+    }
+  }
+  this._trackableURLs = trackableURLs;
+  this.onOpen = this.onOpen.bind(this);
+  tabs.on("open", this.onOpen);
+}
+
+TabTracker.prototype = {
+  _openTabs: [],
+
+  uninit: function() {
+    for (let tab of this._openTabs) {
+      tab.removeListener("ready", this.logReady);
+      tab.removeListener("activate", this.logActivate);
+      tab.removeListener("deactivate", this.logDeactivate);
+      tab.removeListener("close", this.logDeactivate);
+    }
+    tabs.removeListener("open", this.onOpen);
+  },
+
+  isActivityStreamsURL: function(URL) {
+    return this._trackableURLs.indexOf(URL) !== -1;
+  },
+
+  getLastActivation: function(tab) {
+    if (!ss.storage.tabData[tab.url] || !ss.storage.tabData[tab.url][tab.id]) {
+      return null;
+    }
+    let activations = ss.storage.tabData[tab.url][tab.id].activations;
+    return activations[activations.length - 1];
+  },
+
+  navigateAwayFromPage: function(tab) {
+    if (!ss.storage.tabData[tab.url][tab.id]) {
+      // We're navigating away from an activity streams page that
+      // didn't even load yet. Let's say it's been active for 0 seconds.
+      ss.storage.tabData[tab.url][tab.id] = {activations: [{startTime: Date.now(), totalTime: 0}]};
+      return;
+    }
+    let lastElement = this.getLastActivation(tab);
+    if (lastElement && lastElement.startTime) {
+      lastElement.totalTime = (Date.now() - lastElement.startTime) / 1000;
+    }
+  },
+
+  logReady: function(tab) {
+    if (this.isActivityStreamsURL(tab.url)) {
+      if (!ss.storage.tabData[tab.url][tab.id]) {
+        ss.storage.tabData[tab.url][tab.id] = {activations: []};
+      } else if (!this.getLastActivation(tab).totalTime) {
+        // The page content has been reloaded but a total time wasn't set.
+        // This is due to a page refresh. Let's set the total time now.
+        this.navigateAwayFromPage(tab);
+      }
+      this.logActivate(tab);
+      return;
+    }
+    // We loaded a URL other than activity streams. If the last open URL
+    // in this tab was activity streams (totalTime hasn't been set yet),
+    // then update its state.
+    let lastElement = this.getLastActivation(tab);
+    if (lastElement && !lastElement.totalTime) {
+      this.navigateAwayFromPage(tab);
+    }
+  },
+
+  logActivate: function(tab) {
+    if (this.isActivityStreamsURL(tab.url)) {
+      ss.storage.tabData[tab.url][tab.id].activations.push({startTime: Date.now()});
+    }
+  },
+
+  logDeactivate: function(tab) {
+    if (this.isActivityStreamsURL(tab.url)) {
+      this.navigateAwayFromPage(tab);
+    }
+  },
+
+  onOpen: function(tab) {
+    this._openTabs.push(tab);
+
+    this.logReady = this.logReady.bind(this);
+    this.logActivate = this.logActivate.bind(this);
+    this.logDeactivate = this.logDeactivate.bind(this);
+
+    tab.on("ready", this.logReady);
+    tab.on("activate", this.logActivate);
+    tab.on("deactivate", this.logDeactivate);
+    tab.on("close", this.logDeactivate);
+  },
+};
+
+exports.TabTracker = TabTracker;
--- a/lib/main.js
+++ b/lib/main.js
@@ -15,6 +15,7 @@ Object.assign(exports, {
 
     // options.loadReason can be install/enable/startup/upgrade/downgrade
     PlacesProvider.links.init();
+    options.telemetry = false;
     exports.app = new ActivityStreams(options);
   },
 
--- a/test/test-TabTracker.js
+++ b/test/test-TabTracker.js
@@ -1,0 +1,123 @@
+"use strict";
+
+const {before, after} = require("sdk/test/utils");
+const tabs = require("sdk/tabs");
+const ss = require("sdk/simple-storage");
+const {setTimeout} = require("sdk/timers");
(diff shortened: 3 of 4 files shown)


lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+    // We loaded a URL other than our newtab. If this tab had our newtab
+    // open before, then update its state.
+    if (ss.storage.tabData[tab.id]) {
+      this.navigateAwayFromNewtab(tab);
+    }
+  },
+
+  logActivate: function(tab) {
+    if (tab.url == NEWTAB_URL) {
```
pdehaan: nit: `===`
(same w/ lines 63 and 69 below)

test/test-TabTracker.js
```diff
@@ -0,0 +1,113 @@
+"use strict";
+
+const {before, after} = require("sdk/test/utils");
+const tabs = require("sdk/tabs");
+const ss = require("sdk/simple-storage");
+const {setTimeout} = require("sdk/timers");
+const {TabTracker} = require("lib/TabTracker");
+
+const NEWTAB_URL = "resource://activity-streams/data/content/activity-streams.html#/";
```
pdehaan: Nit: this is duplicated in this file and the previous file. Not sure if it is worth exporting from one file and sharing, or trying to shim into webpack config or something.

test/test-TabTracker.js
```diff
@@ -0,0 +1,113 @@
...
+  assert.ok(secondsOpen >= 5, "The tab should have stayed open for at least 5 seconds");
+};
+
+exports.test_TabTracker_reactivating = function*(assert) {
+  let openTabs = [];
+
+  let tabsOpenedPromise = new Promise(resolve => {
+    let onOpen = function(tab) {
+      openTabs.push(tab);
+      if (openTabs.length == 2) {
```
pdehaan: nit: `===`

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
+const tabs = require("sdk/tabs");
+const ss = require("sdk/simple-storage");
+
+const NEWTAB_URL = "resource://activity-streams/data/content/activity-streams.html#/";
```
oyiptong: Do not hardcode the url here. Please obtain it from `ActivityStreams`

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+
+const TabTracker = {
+  _openTabs: [],
+
+  init: function() {
```
oyiptong: If you need an init function, have you considered making TabTracker an object?

i.e.
```javascript
function TabTracker(options = {}) {
 ...
};

TabTracker.prototype = {
 ...
}
```
That would also obviate the need to do the function binding: `this.onOpen = this.onOpen.bind(this);`

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+    this.onOpen = this.onOpen.bind(this);
```
oyiptong: I have to look at what `this` is bound to, but this is used elsewhere in the code. Is that even necessary?

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+    tabs.on("open", this.onOpen);
+    if (!ss.storage.tabData) {
+      ss.storage.tabData = {};
+    }
+  },
+
+  uninit: function() {
+    for (let i in this._openTabs) {
+      let tab = this._openTabs[i];
+      tab.removeListener("pageshow", this.logShow);
```
oyiptong: The listener won't actually be removed. This is a bug elsewhere in the code as well.

if a listener was registered with e.g. `this.logShow.bind(this)`, when `removeListener` is invoked, `this.logShow === this.logShow.bind(this) === false `, i.e. it won't get removed

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+      // We're navigating away from a newtab that didn't even load yet.
+      // Let's say it's been active for 0 seconds.
+      ss.storage.tabData[tab.id] = {activations: [{startTime: Date.now(), totalTime: 0}]};
+      return;
+    }
+    let activations = ss.storage.tabData[tab.id].activations;
+    let lastElement = activations[activations.length - 1];
+    if (lastElement.startTime && !lastElement.totalTime) {
+      lastElement.totalTime = (Date.now() - lastElement.startTime) / 1000;
+    }
```
oyiptong: we might want to `Cu.reportError` or `console.warn` in case `totalTime` is not falsey. It means this callback has been called multiple times

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+  },
+
+  logShow: function(tab) {
```
oyiptong: nit: logPageShow ?

lib/TabTracker.js
```diff
@@ -0,0 +1,89 @@
...
+    if (tab.url == NEWTAB_URL) {
+      ss.storage.tabData[tab.id] = {
+        activations: [{startTime: Date.now()}]
```
oyiptong: won't this clobber the past activations?

in `navigateAwayFromNewtab`, line 36, you get the lastElement. I assume you made the activations key an array so you could buffer activations in case the data hasn't been sent out.

But this code makes it so that if you reload the activitystreams page, the activations array gets clobbered.

lib/TabTracker.js
oyiptong: Also, it's not newtab, but rather, activity streams page

oyiptong: Actually, let me refine my comment:

we have 2 urls we care about, the "activity streams" page and the "newtab" page. We want to track engagement on both separately.

@k88hudson can advise how the URLs differ

lib/TabTracker.js
oyiptong: Ah, upon reload, the `activate` event is triggered instead?

lib/TabTracker.js
oyiptong: We also want to send out an observer notification, so that an eventual data uploader can respond to a finished interaction and send the data out.

test/test-TabTracker.js
oyiptong: I think it should be exported from an `ActivityStreams` instance, as parameter (getter?)

oyiptong: Also, will the default `ActivityStreams` object be initialized in the test?

test/test-TabTracker.js
```diff
@@ -0,0 +1,113 @@
...
+
+exports.test_TabTracker_open_close_tab = function*(assert) {
+  let tabClosedPromise = new Promise(resolve => {
+    let onOpen = function(tab) {
+      setTimeout(function() {
+        tab.close(() => {
+          tabs.removeListener("pageshow", onOpen);
+          resolve();
+        });
+      }, 5000);
```
oyiptong: Can we wait until the document has loaded instead of a `setTimeout` of 5 seconds? That makes testing very slow

test/test-message-passing.js
```diff
@@ -4,11 +4,13 @@ const test = require("sdk/test");
 const tabs = require("sdk/tabs");
 const {ActivityStreams} = require("lib/ActivityStreams");
 const httpd = require("./lib/httpd");
+const {TabTracker} = require("lib/TabTracker");
```
oyiptong: what does `TabTracker` init do in this test?

lib/TabTracker.js
emtwo: Indeed, that's why I have the following line of code when init-ing,  to avoid that bug:
`this.logShow = this.logShow.bind(this);`

It may not be necessary if I use a TabTracker object.

oyiptong: Ah! nice workaround

lib/main.js
```diff
@@ -15,6 +16,7 @@ Object.assign(exports, {
 
     // options.loadReason can be install/enable/startup/upgrade/downgrade
     PlacesProvider.links.init();
+    TabTracker.init();
```
oyiptong: Can we do that init in ActivityStreams with an optional parameter?
Something like {telemetry: true}, where telemetry is enabled via a pref

oyiptong: addon-pref that is

test/test-TabTracker.js
oyiptong: Let's leave the pref to another bug. For now, a telemetry parameter passed to `Activitystreams` that doesn't init TabTracker at all is great.

lib/main.js
oyiptong: Let's leave the pref to another bug. For now, a telemetry parameter passed to Activitystreams that doesn't init TabTracker at all is great.

test/test-TabTracker.js
emtwo: I was actually waiting 5 seconds to test that `totalTime` is set to a non-zero value. But waiting for page load and immediately closing might still end up with 0 `totalTime`. How about 10ms instead of 5000?

oyiptong: 10ms is great!

oyiptong: how about DOMContentLoaded? will it be zero if you can do that?

oyiptong: To be clear, I think 10ms is fine. Was just curious about a DOMContentLoaded event and the result of totalTime

lib/TabTracker.js
emtwo: It's not necessarily an error if `navigateAwayFromNewtab` gets called more than once for a given tab. Loading a new URL twice in a tab that once had activity streams open will trigger`navigateAwayFromNewtab` ... I think instead I'll avoid`navigateAwayFromNewtab` from being triggered more than once and add a comment.

oyiptong:
r+

lib/main.js
```diff
@@ -15,6 +15,7 @@ Object.assign(exports, {
 
     // options.loadReason can be install/enable/startup/upgrade/downgrade
     PlacesProvider.links.init();
+    options.telemetry = true;
```
oyiptong: Nit: Can you please turn it off by default?
We don't have means to send the data yet, which means they will just accumulate in simple storage.
Let's turn it on by default once we can send the data.

This code is still useful because it is probably foundational for all other telemetry.

PR #142, opened 2016-02-18
feat(addon-ui) #94 hide url in awesomebar

### LINKED ISSUE #94: Implement correct url in awesome bar
Opening a new tab shows the resource URL and puts the cursor insertion point is at the end of the resource URL which makes it hard to type things into the Awesome Bar:

screen recording 2016-02-10 at 07 25 pm

rlr:
R? @oyiptong 

Fixes #94

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -1,8 +1,8 @@
-/* globals NewTabURL, EventEmitter, XPCOMUtils */
+/* globals NewTabURL, EventEmitter, XPCOMUtils, Services */
 
 "use strict";
 
-const {Cu} = require("chrome");
+const {Ci,Cu} = require("chrome");
 const {data} = require("sdk/self");
 const {PageMod} = require("sdk/page-mod");
 const {ActionButton} = require("sdk/ui/button/action");
@@ -10,6 +10,7 @@ const tabs = require("sdk/tabs");
 const {PlacesProvider} = require("lib/PlacesProvider");
 const am = require("content-src/actions/action-manager");
 
+Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource:///modules/NewTabURL.jsm");
 
@@ -30,6 +31,7 @@ function ActivityStreams(options = {}) {
   this._setupPageMod();
   this._setupListeners();
   this._setupButton();
+  this._hideAppURLs();
   NewTabURL.override(this.options.pageURL);
 }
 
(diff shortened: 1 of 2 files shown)


lib/ActivityStreams.js
```diff
@@ -18,6 +18,14 @@ XPCOMUtils.defineLazyGetter(this, "EventEmitter", function() {
   return EventEmitter;
 });
 
+XPCOMUtils.defineLazyServiceGetter(this, "windowWatcher",
```
oyiptong: Actually, it might be better to do:

```javascript
```

You can access the window watcher by calling `Services.ww` and `Services.wm`.
https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Services.jsm

lib/ActivityStreams.js
```diff
@@ -30,6 +38,7 @@ function ActivityStreams(options = {}) {
   this._setupPageMod();
   this._setupListeners();
   this._setupButton();
+  this._addToInitialPages();
```
oyiptong: `addToHiddenURLs` might be more descriptive

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
    */
   _setupButton() {
-    ActionButton({
+    this._button = ActionButton({
       id: "activity-streams-link",
       label: "Timeline",
       icon: data.url("content/img/list-icon.svg"),
       // TODO: Fix url below to about:timeline when that gets implemented.
-      onClick: () => tabs.open("resource://activity-streams/data/content/activity-streams.html#timeline")
+      onClick: () => tabs.open(`${this.options.pageURL}#/timeline`)
```
oyiptong: :+1:

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
+      baseUrl,
+      `${baseUrl}#/`,
+      `${baseUrl}#/timeline`,
+    ];
+  },
+
+  /**
+   * Add the initial pages to the passed in window.
+   */
+  _addInitialPages(window) {
```
oyiptong: as above, addToHiddenURLs might be more descriptive

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
+
+  /**
+   * Watches for new windows and adds our urls to gInitialPages.
+   */
+  _windowObserver: {
+    observe: (chromeWindow, topic) => {
+      if (topic === "domwindowopened") {
+        let window = chromeWindow;
+        window.QueryInterface(Ci.nsIDOMWindow).addEventListener("DOMContentLoaded", {
+          handleEvent: () => {
```
oyiptong: Perhaps we can only add the url to the hidden list if the current URL is an appURL

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
       // TODO: Fix url below to about:timeline when that gets implemented.
-      onClick: () => tabs.open("resource://activity-streams/data/content/activity-streams.html#timeline")
+      onClick: () => tabs.open(`${this.options.pageURL}#/timeline`)
     });
   },
 
   /**
+   * The URLs for the app.
+   */
+  get appURLs() {
```
oyiptong: If this is getting called on potentially every new dom window opened, we should cache this.
We should also use a `Set`

oyiptong: Actually, you don't test for inclusion, an Array is fine

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
+    ];
+  },
+
+  /**
+   * Add the initial pages to the passed in window.
+   */
+  _addInitialPages(window) {
+    if (window.gInitialPages) {
+      for (let url of this.appURLs) {
+        window.gInitialPages.push(url);
```
oyiptong: How about `window.gInitialPages.push(...this.appURLs)`?

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
+        }, false);
+      }
+    }
+  },
+
+  /**
+   * Adds our urls to the gInitialPages of each open window and sets up
+   * an observer for new windows that are opened.
+   */
+  _addToInitialPages() {
```
oyiptong: how about `_hideAppURLs`?

lib/ActivityStreams.js
rlr: Not sure I follow? We need to have it set on every open window in case they open a new tab.

lib/ActivityStreams.js
rlr: so fancy

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
+  _addInitialPages(window) {
+    if (window.gInitialPages) {
+      for (let url of this.appURLs) {
+        window.gInitialPages.push(url);
+      }
+    }
+  },
+
+  /**
+   * Watches for new windows and adds our urls to gInitialPages.
```
oyiptong: > to the list of hidden awesomebar URLs

lib/ActivityStreams.js
```diff
@@ -164,16 +173,69 @@ ActivityStreams.prototype = {
...
+          handleEvent: () => {
+            this._addInitialPages(window);
+          }
+        }, false);
+      }
+    }
+  },
+
+  /**
+   * Adds our urls to the gInitialPages of each open window and sets up
```
oyiptong: > Adds our urls to the list of hidden awesomebar URLs of each open window...

lib/ActivityStreams.js
oyiptong: I'm learning from @marcoscaceres ;-)

oyiptong:
r+ with above mentioned changes

lib/ActivityStreams.js
oyiptong: Ah, good point

rlr:
^^ I think that covers everything

PR #153, opened 2016-02-20
chore(npm): remove shrinkwrap

k88hudson:
Since this is a bit of a pain/seems to go out of sync a lot and is causing more issues than it is preventing them, let's remove shrinkwrap for now and rely on locked semver for dependencies. We can restore if we need it.


pdehaan:
LGTM. Did you want to remove all the semver range operators in the package.json file as well?
I spotted ~45 of them, which may cause slightly breaking behavior between builds now that shrinkwrap isn't locking version numbers.

pdehaan:
Merging. We can lock devDeps in a future PR if we run out of literally all other valuable work and get that desparate.

PR #155, opened 2016-02-20
fix(content): Make sure components do not throw if some data is missing

k88hudson:
The reasoning behind this is that sometimes we get weirdly formatted history items, which ideally we will filter out, but at least we don't want to crash the whole parent component.

### DIFF
--- a/content-test/components/ActivityFeed.test.js
+++ b/content-test/components/ActivityFeed.test.js
@@ -37,6 +37,12 @@ describe("ActivityFeedItem", function() {
     el = ReactDOM.findDOMNode(instance);
   });
 
+  it("should not throw if missing props", () => {
+    assert.doesNotThrow(() => {
+      TestUtils.renderIntoDocument(<ActivityFeedItem />);
+    });
+  });
+
   describe("valid sites", () => {
     it("should create the element", () => {
       assert.ok(el);
--- a/yamscripts.yml
+++ b/yamscripts.yml
@@ -41,7 +41,7 @@ scripts:
     # test:jpm: Run jpm tests
     jpm: jpm test -b beta
     # test:karma: Run content tests only
-    karma: karma start
+    karma: karma start --browsers FirefoxNightly
 
 # tdd: Run content tests continuously
   tdd: karma start --no-single-run
--- a/content-src/components/Header/Header.js
+++ b/content-src/components/Header/Header.js
@@ -7,16 +7,17 @@ const Header = React.createClass({
   },
   render() {
     const props = this.props;
+    const currentRoute = props.currentRoute || {};
     return (<header className="head">
       <section className="nav" onClick={() => this.setState({showDropdown: !this.state.showDropdown})}>
         <h1>
-          <span hidden={!props.currentRoute.icon} className={`icon fa ${props.currentRoute.icon}`} />
-          <span>{props.currentRoute.title}</span>
+          <span hidden={!currentRoute.icon} className={`icon fa ${currentRoute.icon}`} />
+          <span>{currentRoute.title}</span>
           <span className="arrow fa fa-chevron-down" />
         </h1>
         <ul className="nav-picker" hidden={!this.state.showDropdown}>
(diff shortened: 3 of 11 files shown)


content-test/components/ActivityFeed.test.js
```diff
@@ -37,6 +37,12 @@ describe("ActivityFeedItem", function() {
     el = ReactDOM.findDOMNode(instance);
   });
 
+  it("should not throw if on missing props", () => {
```
pdehaan: nit: not sure if you want "should not throw **if** missing props" or "should not throw **on** missing props".
Having both "**if on**" sounds a bit awkward (same w/ component-test/components/{SiteIcon,TopSites}.test.js below too)

k88hudson: oops brain fail

pdehaan:
I think this needs a rebase to pick up the latest FF 45/Beta fixes.

k88hudson:
@pdehaan I just realized karma is going to be a problem if you have stable firefox installed, since it will point to stable firefox

k88hudson:
@pdehaan R? I added nightly back in for local test running so they don't fail.. as for the beta path, i'm going to file a ticket about it

yamscripts.yml
```diff
@@ -41,7 +41,7 @@ scripts:
     # test:jpm: Run jpm tests
     jpm: jpm test -b beta
     # test:karma: Run content tests only
-    karma: karma start
+    karma: karma start --browsers FirefoxNightly
```
pdehaan: So this is just overriding the `browsers` set in the [karma.conf.js:7](https://github.com/mozilla/activity-streams/blob/fix-empty-case/karma.conf.js#L7)?

pdehaan:
I'm getting this locally:
```
SUMMARY:
✔ 64 tests completed
✖ 1 test failed

FAILED TESTS:
  ActivityFeedItem
    ✖ should not throw if on missing props
      Firefox 47.0.0 (Mac OS X 10.11.0)
    expected [Function] to not throw an error but 'TypeError: url is undefined' was thrown
    AssertionError@/Users/pdehaan/dev/github/activity-streams/content-test/index.js:354:19 <- webpack:///~/assertion-error/index.js:73:0
```

pdehaan:
Travis-CI is passing, but I'm seeing the following warnings in the log:

> - ERROR: 'Warning: Failed propType: Required prop `url` was not specified in `ActivityFeedItem`.' ([source](https://travis-ci.org/mozilla/activity-streams/builds/110851472#L657))
>
> - ERROR: 'Warning: Failed propType: Required prop `sites[0].url` was not specified in `TopSites`.' ([source](https://travis-ci.org/mozilla/activity-streams/builds/110851472#L709))
> 
> - ERROR: 'Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `TopSites`. See https://fb.me/react-warning-keys for more information.' ([source](https://travis-ci.org/mozilla/activity-streams/builds/110851472#L710))

k88hudson:
Those warnings are actually expected, since we're making sure the components don't throw (just show errors)

k88hudson:
@rlr what do you think?

oyiptong:
Exceptions, even when expected, cause log entries to occur, at least in my experience. It be because we set the log to verbose mode.

oyiptong:
As far as I can see, the tests actually not throwing anything, so I don't know why the errors are showing up

rlr:
I see. Should we make those props have `isRequired=False` since we gracefully handle their absence? Or do we want the warnings in the console? I kind of lean towards just making it all optional since that's how we are handling it.

k88hudson:
@rlr @oyiptong They are just validation errors from React; I think we do want to keep them, because it means there is something screwed up in our data source i.e. we're not filtering something properly and we probably want to know about it. However, I figured it's better just to warn rather than throw, which will destroy the whole component

rlr:
sounds good. definitely warn over throw, especially on the released version :+1:

oyiptong:
:+1:

PR #172, opened 2016-02-22
chore(content): #157 Use the dev server endpoint for embedly

### LINKED ISSUE #157: Switch embedly URL over to use dev-server
The new endpoint is: `http://embedly-proxy.dev.mozaws.net/extract`, we should use this instead of localhost.

It would be good to add this to `default-config.yml` and leave it configurable.

k88hudson:
Fixes #157


k88hudson:
@oyiptong / @rlr / @jaredkerim R?

oyiptong:
need to fix npm scripts. they still want to run the embedly server

oyiptong:
screen shot 2016-02-22 at 4 41 52 pm

oyiptong:
Code looks fine. Request goes to the embedly server and the correct response is obtained.

Is it expected that the embedly content doesn't show up on screen?

oyiptong:
it works now!

oyiptong:
plus the places links are gone too... hmm weird

oyiptong:
r+

pdehaan:
What's odd is that I'm seeing a Flash of the shim data, and then it all disappears and my data gets loaded in, but it looks like everything gets wiped because I knocked over our Embedly proxy server (with HTTP 500, 400, and 414 statuses):

new_tab

oyiptong:
look at you. breaking our server on hour one ;-)

pdehaan:
look at you. passing 5000 URLs in a query string like there are no laws.
http___embedly-proxy_dev_mozaws_net_extract_urls_http___embedly-proxy_dev_mozaws_net_extract_urls_http_3a_2f_2fideatown_dev_3a8000_2f_urls_http_3a_2f_2flocalhost_3a1963_2factivity-streams_html_urls_http_3a_2f_2fidea-town-dev_elasticbeanstal

But yeah, the HTTP 414 seems to be "Request-URI Too Large".
I'm not able to reproduce the HTTP 500 at the moment, but it may have been due to some weird "places:" links or links to XPIs or other random URLs that I look at.

pdehaan:
Hopefully the Embedly proxy server has some logging so we can dig in and see what maybe caused the 500s. /cc @jaredkerim

jaredkerim:
@pdehaan Please file any tickets for the embedly proxy here: https://github.com/mozilla/embedly-proxy and include query data (in this case the full url hitting the server) so I can reproduce the error.  Thanks!

pdehaan:
@jaredkerim Sure thing! Peter's Private Browsing Data as a Service.
Although this particular use case may just be a temporary blip until we start POSTing the data instead of passing everything via the query string.

k88hudson:
I changed the limit on history items to be 25 instead, which should be a lot nicer

k88hudson:
@oyiptong ok, is this good to merge?

oyiptong:
Lgtm

PR #181, opened 2016-02-23
fix(content): #177 Handle filtering of URLs and embedly data more gracefully

### LINKED ISSUE #177: Better filtering of site data
My current output looks like this:
new_tab_and_new_tab

You can see in the developer tools output that I'm getting some errors back from the embedly proxy where we get an `error_code` and `error_message` (mainly because there is no way for a proxy to resolve my "localhost" and "ideatown.dev" testing domains.

new_tab_and_new_tab

We probably want to filter out any results from the embedly proxy when we encounter a [response](http://docs.embed.ly/docs/extract#response-attributes) with the following properties:
- `error_code`
- `error_message`

For the latter two, it seems like embedly will return a `type` of "error" which is probably easiest to filter on.
For more information, see http://docs.embed.ly/docs/extract#error-codes.

k88hudson:
Fixes #177 by filtering out error responses. I also added some URL filtering for long URLs (>100 characters), localhost, and non-http[s] protocols

### DIFF
--- a/content-src/lib/filters.js
+++ b/content-src/lib/filters.js
@@ -1,0 +1,44 @@
+const urlParse = require("url-parse");
+
+const TEMP_MAX_LENGTH = 100;
+const ALLOWED_PROTOCOLS = new Set([
+  "http:",
+  "https:"
+]);
+const DISALLOWED_HOSTS = new Set([
+  "localhost",
+  "127.0.0.1",
+  "0.0.0.0"
+]);
+
+function createFilter(definition) {
+  return function(item) {
+    let result = true;
+    const url = item && item.url && urlParse(item.url) || {};
+    definition.forEach(test => {
+      if (!test(item, url)) {
+        result = false;
+      }
+    });
+    return result;
+  };
+}
+
+const URL_FILTERS = [
+  (item) => !!item.url,
+  // This is temporary, until we can use POST-style requests
+  // see https://github.com/mozilla/embedly-proxy/issues/1
+  (item) => item.url && item.url.length < TEMP_MAX_LENGTH,
+  (item, url) => ALLOWED_PROTOCOLS.has(url.protocol),
+  (item, url) => !DISALLOWED_HOSTS.has(url.hostname)
+];
+
+const DATA_FILTERS = [
+  item => !(item.error_code || item.error_message)
+];
+
+module.exports = {
+  createFilter,
+  urlFilter: createFilter(URL_FILTERS),
+  siteFilter: createFilter(DATA_FILTERS)
+};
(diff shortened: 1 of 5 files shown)


k88hudson:
@oyiptong @pdehaan or @rlr R?

content-src/lib/filters.js
```diff
@@ -0,0 +1,28 @@
...
+        result = false;
+      }
+    });
+    return result;
+  };
+}
+
+const URL_FILTERS = [
+  item => !!item.url,
+  item => item.url && item.url.length < 100,
```
pdehaan: Why an arbitrary URL length limit?

For example, why wouldn't I want to see URLs like: `https://github.com/k88hudson/activity-streams/commit/eceda0e47e9f4a810bddf089e08c7b04dea1688f` (don't actually fact check the length, it's only like 94 chars, but you get the point).

Also, bonus points for making the 100 magic number a `const` at the top of the file, like `MAX_URL_LEN` or whatever.

content-src/lib/filters.js
```diff
@@ -0,0 +1,28 @@
...
+  item => /^https?:/i.test(item.url),
+  item => !(/^https?:\/\/localhost/i.test(item.url))
```
pdehaan: Not sure if we need to add a `\/` at the end of localhost here. Something tells me that this would incorrectly catch on domains like "http://localhost-is-cool.com"

Also, do we want to try and filter out loopback URLs like "127.0.0.1" and "0.0.0.0"?

content-src/lib/filters.js
k88hudson: This is kind of a stop-gap until we implement the POST-style requests. I'll make a note

content-src/lib/filters.js
k88hudson: :+1:

pdehaan: Actually, the loopback URLs would probably be handled by `DATA_FILTERS` below.

k88hudson:
@pdehaan ok, what do you think about this?

pdehaan:
Looks great!
But I'm still seeing the one edge case of I have a hacked site in my /etc/hosts for ideatown.dev which is showing up in my Top Sites (but it should have returned an error from the embedly proxy, I would assume).

new_tab_and_nightly_and_nightly

k88hudson:
weird! hmm...

k88hudson:
Aha, found the issue!

k88hudson:
@pdehaan ok, can you see if this fixes it?

pdehaan:
### FAN-TAS-TIC!

Looks good to me. I'm seeing some other issues w/ duplicate child keys, but I think it's probably unrelated and something we can address in a future PR. :+1:

(Note that I only have 5 Top Sites and 2 Spotlight sites)
new_tab

k88hudson:
Yes, we should definitely sort out that de-duping

PR #186, opened 2016-02-23
Closes  #165 - Add bookmark status to Recent Links API query

### LINKED ISSUE #165: Add bookmark status to Recent Links API query
In the same fashion as #164, we want to return the bookmark `id` and `guid` for a recent link if it has been bookmarked.

This requires a `LEFT JOIN moz_bookmarks b ON p.id = b.fk` in the query as well as new columns in the expected return column list

mzhilyaev:
Fixes #165

### DIFF
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -196,15 +196,16 @@ Links.prototype = {
 
     let sqlQuery =  `SELECT moz_places.url as url,
                             moz_favicons.data as favicon,
-                            mime_type as mimeType,
+                            moz_favicons.mime_type as mimeType,
                             moz_places.title,
-                            frecency,
-                            last_visit_date as lastVisitDate,
+                            moz_places.frecency,
+                            moz_places.last_visit_date as lastVisitDate,
                             "history"  as type,
-                            moz_bookmarks.id as bookmark
+                            moz_bookmarks.guid as bookmarkGuid,
+                            moz_bookmarks.dateAdded / 1000 as bookmarkDateCreated
                      FROM moz_places
                      LEFT JOIN moz_favicons
-                     ON favicon_id = moz_favicons.id
+                     ON moz_places.favicon_id = moz_favicons.id
                      LEFT JOIN moz_bookmarks
                      ON moz_places.id = moz_bookmarks.fk
                      WHERE hidden = 0 AND last_visit_date NOTNULL
@@ -212,7 +213,8 @@ Links.prototype = {
                      LIMIT :limit`;
 
     let links = yield this.executePlacesQuery(sqlQuery, {
-                  columns: ["url", "favicon", "mimeType", "title", "lastVisitDate", "frecency", "type", "bookmark"],
+                  columns: ["url", "favicon", "mimeType", "title", "lastVisitDate",
+                            "frecency", "type", "bookmarkGuid", "bookmarkDateCreated"],
                   params: {limit: limit}
                 });
 
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -194,17 +194,25 @@ Links.prototype = {
       limit = HISTORY_RESULTS_LIMIT;
     }
 
-    let sqlQuery =  `SELECT moz_places.url as url, moz_favicons.data as favicon, mime_type as mimeType,
-                            title, frecency, last_visit_date as lastVisitDate,  "history"  as type
+    let sqlQuery =  `SELECT moz_places.url as url,
+                            moz_favicons.data as favicon,
+                            mime_type as mimeType,
+                            moz_places.title,
+                            frecency,
+                            last_visit_date as lastVisitDate,
+                            "history"  as type,
+                            moz_bookmarks.id as bookmark
                      FROM moz_places
                      LEFT JOIN moz_favicons
                      ON favicon_id = moz_favicons.id
+                     LEFT JOIN moz_bookmarks
+                     ON moz_places.id = moz_bookmarks.fk
                      WHERE hidden = 0 AND last_visit_date NOTNULL
                      ORDER BY lastVisitDate DESC, frecency DESC, url
                      LIMIT :limit`;
 
     let links = yield this.executePlacesQuery(sqlQuery, {
-                  columns: ["url", "favicon", "mimeType", "title", "lastVisitDate", "frecency", "type"],
+                  columns: ["url", "favicon", "mimeType", "title", "lastVisitDate", "frecency", "type", "bookmark"],
                   params: {limit: limit}
                 });
 
--- a/test/test-PlacesProvider.js
+++ b/test/test-PlacesProvider.js
@@ -138,7 +138,7 @@ exports.test_Links_getRecentLinks = function*(assert) {
   yield PlacesTestUtils.addFavicons(faviconData);
 
   // insert a bookmark
-  yield Bookmarks.insert({url: "https://mozilla1.com/0", parentGuid: "root________", type: Bookmarks.TYPE_BOOKMARK});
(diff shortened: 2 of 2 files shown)


lib/PlacesProvider.js
```diff
@@ -194,17 +194,25 @@ Links.prototype = {
...
-    let sqlQuery =  `SELECT moz_places.url as url, moz_favicons.data as favicon, mime_type as mimeType,
-                            title, frecency, last_visit_date as lastVisitDate,  "history"  as type
+    let sqlQuery =  `SELECT moz_places.url as url,
+                            moz_favicons.data as favicon,
+                            mime_type as mimeType,
+                            moz_places.title,
+                            frecency,
+                            last_visit_date as lastVisitDate,
+                            "history"  as type,
+                            moz_bookmarks.id as bookmark
```
oyiptong: Can you see what parameters we'd need if we wanted to unbookmark?

mzhilyaev: we can definitely  add API to remove a bookmark, but i suggest opening a different bug for it

oyiptong: I don't mean to say we need to implement it here. Just to see if bookmarks.id would be enough to implement that in the future

k88hudson:
Awesome! Looks good to me

lib/PlacesProvider.js
oyiptong: @mzhilyaev here, i looked it up:

https://dxr.mozilla.org/mozilla-central/rev/5b2baa5e9356644a7ed0b73e422eaff62e159ffb/toolkit/components/places/Bookmarks.jsm#381

We need the bookmark GUID instead of theID.

lib/PlacesProvider.js
```diff
@@ -194,17 +194,25 @@ Links.prototype = {
       limit = HISTORY_RESULTS_LIMIT;
     }
 
-    let sqlQuery =  `SELECT moz_places.url as url, moz_favicons.data as favicon, mime_type as mimeType,
-                            title, frecency, last_visit_date as lastVisitDate,  "history"  as type
+    let sqlQuery =  `SELECT moz_places.url as url,
+                            moz_favicons.data as favicon,
+                            mime_type as mimeType,
+                            moz_places.title,
+                            frecency,
```
oyiptong: nit: inconsistent use of column declaration

consider doing this instead:
```sql
SELECT moz_places.url as url,
               moz_favicons.data as favicon,
               moz_favicons.mime_type as mimeType,
               moz_places.title,
               moz_places.frecency,
               moz_places.last_visit_date as lastVisitDate,
               moz_bookmarks.guid as bookmarkGUID,
               "history"  as type
FROM moz_places
LEFT JOIN moz_favicons
ON moz_places.favicon_id = moz_favicons.id
LEFT JOIN moz_bookmarks
ON moz_places.id = moz_bookmarks.fk
WHERE hidden = 0 AND last_visit_date NOTNULL
ORDER BY lastVisitDate DESC, frecency DESC, url
LIMIT :limit;
```

The query becomes much more readable.

test/test-PlacesProvider.js
```diff
@@ -137,6 +137,9 @@ exports.test_Links_getRecentLinks = function*(assert) {
   yield PlacesTestUtils.addVisits(visits, faviconData);
   yield PlacesTestUtils.addFavicons(faviconData);
 
+  // insert a bookmark
+  yield Bookmarks.insert({url: "https://mozilla1.com/0", parentGuid: "root________", type: Bookmarks.TYPE_BOOKMARK});
```
oyiptong: could obtain the object returned here

test/test-PlacesProvider.js
```diff
@@ -144,6 +147,8 @@ exports.test_Links_getRecentLinks = function*(assert) {
   assert.equal(links[2].url, "https://mozilla3.com/2", "Expected 3-rd link");
   assert.equal(links[3].url, "https://mozilla4.com/3", "Expected 4-th link");
   assert.equal(faviconData[links[2].url], links[2].favicon, "favicon data is stored as expected");
+  assert.ok(links[1].bookmark != null, "expected non-null bookmark for second link");
```
oyiptong: ... and assert the value here

oyiptong:
feedback+

k88hudson:
Actually, it might be a good idea to include the `date_created` for the bookmark as well, so we know when it was bookmarked. What do you think @oyiptong @mzhilyaev?

oyiptong:
since this is a history API, we should probably call the column `bookmarkCreatedAt`.

Are you thinking about using the date for sorting?

oyiptong:
Or are you thinking about another kind of ranking algorithm?

I'd say for the sake of speed, we should send the raw data to content.

However, if we create a new kind of ranking, based on bookmark status and/or the bookmark's creation date (or last visit date), we should probably do it add-on side.

k88hudson:
@oyiptong not for sorting in this case, just for displaying whether or not it was bookmarked recently, or whether it was just a visit to an old bookmark if it's in a spotlight component

+1 to naming it `bookmarkDateCreated`

oyiptong:
Awesome

lib/PlacesProvider.js
```diff
@@ -194,17 +194,27 @@ Links.prototype = {
...
-                            title, frecency, last_visit_date as lastVisitDate,  "history"  as type
+    let sqlQuery =  `SELECT moz_places.url as url,
+                            moz_favicons.data as favicon,
+                            moz_favicons.mime_type as mimeType,
+                            moz_places.title,
+                            moz_places.frecency,
+                            moz_places.last_visit_date as lastVisitDate,
+                            "history"  as type,
+                            moz_bookmarks.guid as bookmarkGuid,
+                            moz_bookmarks.dateAdded / 1000 as bookmarkDateCreated
```
oyiptong: why are we dividing by 1000 here?

Looking at my places DB, both `last_visit_date` and `dateAdded` seem to have the same unit of time for me.

```python
In [1]: places = 1454547050246805

In [2]: bookmark = 1327245403927000

In [3]: def num_digits(num):
   ...:     count = 0
   ...:     while num > 0:
   ...:         num /= 10
   ...:         count += 1
   ...:     return count
   ...:
In [4]: num_digits(places)
Out[4]: 16

In [5]: num_digits(bookmark)
Out[5]: 16
```

oyiptong: Returning millis is probably what we want and we should be consistent we're consistent.

I should say, we need to divide by 1000 above as well.

mzhilyaev: Agreed, and we also need to change the tests to insert microseconds, which i did in the this PR https://github.com/mozilla/activity-streams/pull/203.

oyiptong: @k88hudson do you convert from micro to millis for `lastVisitDate` right now?

oyiptong: Most excellent, sir

rlr: Yes. I am dividing by 1000 in my last still open pull request: https://github.com/mozilla/activity-streams/pull/197/files#diff-f4ab583e7c603c5a0141b261c46d2367R21

Let me know if you are gonna change it.

k88hudson: We should probably file a ticket for some  basic integration tests/making sure what the client is expecting is what it's going to get

k88hudson: not for this ticket/now obviously but would be nice to have

oyiptong:
r+ with the condition that the timestamp issue being fixed shortly after

mzhilyaev:
All date related fixes and unification will be done in https://github.com/mozilla/activity-streams/issues/204

PR #203, opened 2016-02-25
Closes #164 - Time-constrained top frecent links API

### LINKED ISSUE #164: Time-constrained top frecent links API
We need a new API to power the "top activity" rubric instead of the recent bookmark API.

Here's a cartoon sketch of a query, using postgres operators for datetime. Probably needs work.

```sql
SELECT p.url, p.title, p.frecency,
               p.last_visit_date as lastVisitDate,
                b.id as bookmarkId,
                b.guid as bookmarkGuid,
                f.data as favicon,
                f.mime_type as mimeType,
                "history" as type
FROM moz_places p
LEFT JOIN moz_bookmark m
ON m.fk = p.id
LEFT JOIN moz_favicons f
ON p.favicon_id = f.id
WHERE p.last_visit_date > current_date - interval '72' hour
AND  p.hidden = 0 AND p.last_visit_date NOTNULL
ORDER BY frecency DESC, lastVisitDate DESC, url;
```

It is a history type of query, with bookmarkGuid and id as a marker of whether or not the link has been bookmarked.

The context surrounding this task is explained in #158.

mzhilyaev:
Note a change to timeDaysAgo() - it will return microseconds instead of milliseconds.
This is done because FX keeps last_visit_date in microseconds

Fixes #164

### DIFF
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -30,6 +30,9 @@ XPCOMUtils.defineLazyGetter(this, "gPrincipal", function() {
 // The maximum number of results PlacesProvider retrieves from history.
 const HISTORY_RESULTS_LIMIT = 25;
 
+// time interval for for frecent links query in milliseconds (72 hours).
+const FRECENT_RESULTS_TIME_LIMIT = 72 * 60 * 60 * 1000;
+
 /**
  * Singleton that checks if a given link should be displayed on about:newtab
  * or if we should rather not do it for security reasons. URIs that inherit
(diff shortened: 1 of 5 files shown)


lib/PlacesProvider.js
```diff
@@ -223,6 +245,32 @@ Links.prototype = {
   }),
 
   /**
+   * Gets the top recent links
+   *
+   * @param {Integer} options
+   *                  limit: Maximum number of results to return. Max 100.
```
oyiptong: Max 25 by default, right?

lib/PlacesProvider.js
```diff
@@ -180,20 +183,38 @@ Links.prototype = {
   },
 
   /**
-   * Gets the top recent links
+   * Gets links from history based on supplied options
    *
-   * @param {Integer} options
-   *                  limit: Maximum number of results to return. Max 100.
+   * @param {Object} options
+   *        {Integer} limit: Maximum number of results to return. Max 100.
```
oyiptong: Max seems to be 25 now

oyiptong:
r+ with nits

lib/PlacesProvider.js
k88hudson: Yeah, this is temporary til we get post-style requests with embedly

mzhilyaev:
Closing this PR as it needs to be rebased on top of master

PR #211, opened 2016-02-25
fix(content): #210 show real bookmark ui in ActivityFeed

### LINKED ISSUE #210: Add real bookmarks UI on ActivityFeed
Stars should show up for bookmarked items only

k88hudson:
I need to update this after https://github.com/mozilla/activity-streams/pull/208 lands

Fixes #210

### DIFF
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -2,6 +2,7 @@ const React = require("react");
 const SiteIcon = require("components/SiteIcon/SiteIcon");
 const {prettyUrl} = require("lib/utils");
 const moment = require("moment");
+const classNames = require("classnames");
 
 const DEFAULT_LENGTH = 3;
 const ICON_SIZE = 40;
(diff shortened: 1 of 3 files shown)


pdehaan:
Looks fantastic! It seems to enable/disable correctly if I remove the bookmark.
Only oddity I noticed is that when you mouse-over the bookmark star icon, it switches to an i-beam cursor like it's trying to select text. Not sure which cursor is best in this case though (presumably just the standard non-text cursor &mdash; `cursor: default;` on the [`&.bookmark::after`](https://github.com/mozilla/activity-streams/pull/211/files#diff-74c812b8c086428c4833cff0b44c4ed7R55) selector): https://css-tricks.com/almanac/properties/c/cursor/
But that may not be important if we're moving towards the ability to click on the star/circle to bookmark or unbookmark items.

new_tab

k88hudson:
Ok, all rebased!

k88hudson:
R? @rlr or @pdehaan?

pdehaan:
:+1:

PR #213, opened 2016-02-25
fix(addon): #209 Save tab data in memory instead of simple storage and send notification on tab session completion.

### LINKED ISSUE #209: Tab tracking should be done in memory, not simple storage

emtwo:
Fixes #209

### DIFF
--- a/lib/TabTracker.js
+++ b/lib/TabTracker.js
@@ -1,14 +1,15 @@
+/* globals Services */
+
 const tabs = require("sdk/tabs");
-const ss = require("sdk/simple-storage");
+const {Cu} = require("chrome");
+
+Cu.import("resource://gre/modules/Services.jsm");
 
 function TabTracker(trackableURLs) {
-  if (!ss.storage.tabData) {
-    ss.storage.tabData = {};
-  }
+  this._tabData = {};
+
   for (let trackableURL of trackableURLs) {
-    if (!ss.storage.tabData[trackableURL]) {
-      ss.storage.tabData[trackableURL] = {};
-    }
+    this._tabData[trackableURL] = {};
   }
   this._trackableURLs = trackableURLs;
   this.onOpen = this.onOpen.bind(this);
@@ -17,6 +18,10 @@ function TabTracker(trackableURLs) {
 
 TabTracker.prototype = {
   _openTabs: [],
+
+  get tabData() {
+    return this._tabData;
+  },
 
   uninit: function() {
     for (let tab of this._openTabs) {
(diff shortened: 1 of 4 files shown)


lib/TabTracker.js
```diff
@@ -33,30 +34,32 @@ TabTracker.prototype = {
...
   },
 
   navigateAwayFromPage: function(tab) {
-    if (!ss.storage.tabData[tab.url][tab.id]) {
+    if (!this._tabData[tab.url][tab.id]) {
       // We're navigating away from an activity streams page that
       // didn't even load yet. Let's say it's been active for 0 seconds.
-      ss.storage.tabData[tab.url][tab.id] = {activations: [{startTime: Date.now(), totalTime: 0}]};
+      this._tabData[tab.url][tab.id] = {activations: [{startTime: Date.now(), totalTime: 0}]};
+      DataSender.send(DataSender.TAB_SESSION_COMPLETE, this._tabData);
```
oyiptong: Can you please send the tabData as part of an observer notification?

This gives us flexibility as far as how the data ends up in the data sender.

lib/TabTracker.js
```diff
@@ -33,30 +34,32 @@ TabTracker.prototype = {
...
       return;
     }
     let lastElement = this.getLastActivation(tab);
     if (lastElement && lastElement.startTime) {
       lastElement.totalTime = (Date.now() - lastElement.startTime) / 1000;
     }
+    DataSender.send(DataSender.TAB_SESSION_COMPLETE, this._tabData);
```
oyiptong: as above

lib/DataSender.js
```diff
@@ -0,0 +1,14 @@
+const DataSender = {
```
oyiptong: A more evocative name would be:

`TelemetrySender`

lib/TabTracker.js
```diff
@@ -33,30 +34,32 @@ TabTracker.prototype = {
   },
 
   getLastActivation: function(tab) {
-    if (!ss.storage.tabData[tab.url] || !ss.storage.tabData[tab.url][tab.id]) {
+    if (!this._tabData[tab.url] || !this._tabData[tab.url][tab.id]) {
```
oyiptong: Since we will only ever have one active tab at any given time, it makes sense to make `tab.id` a property of `_tabData` as a sibling of `tab.url`.

lib/TabTracker.js
```diff
@@ -33,30 +34,32 @@ TabTracker.prototype = {
...
     }
+    DataSender.send(DataSender.TAB_SESSION_COMPLETE, this._tabData);
   },
 
   logReady: function(tab) {
     if (this.isActivityStreamsURL(tab.url)) {
-      if (!ss.storage.tabData[tab.url][tab.id]) {
-        ss.storage.tabData[tab.url][tab.id] = {activations: []};
+      if (!this._tabData[tab.url][tab.id]) {
+        this._tabData[tab.url][tab.id] = {activations: []};
```
oyiptong: `activations` may not be useful anymore, since we'll only have one

lib/TabTracker.js
```diff
@@ -76,7 +79,7 @@ TabTracker.prototype = {
 
   logActivate: function(tab) {
     if (this.isActivityStreamsURL(tab.url)) {
-      ss.storage.tabData[tab.url][tab.id].activations.push({startTime: Date.now()});
+      this._tabData[tab.url][tab.id].activations.push({startTime: Date.now()});
```
oyiptong: `startTime` should probably be a first-level object to tabData as well

lib/TabTracker.js
oyiptong: We should still do sanity checks for `url` and `id` whenever an event occurs.

lib/DataSender.js
oyiptong: Or were you thinking of implementing the `Services.obs.notifyObservers` function call here?

oyiptong:
feeback+

lib/TabTracker.js
emtwo: I'm going to change this entire structure to align with what we talked about today.This pull request was just for switching to using memory rather than simple storage and for notifications.

oyiptong: Ah, ok, thanks for the heads up. Will you change the structure in another PR?

emtwo: yes, I just separated them so I can have something review-able by eod and I'll get to the structure change in another PR.

emtwo:
@oyiptong I made the notification changes and updated the module name to Telemetry sender. I'd like to address your other comments in a separate PR if that's ok with you?

oyiptong:
r+ after there is at least one test of the observer notifications

PR #217, opened 2016-02-26
feat(embedly): Remove basic auth and query string data for embedly requests

### LINKED ISSUE #196: Remove basic auth and query string for embedly requests
I'm a little bit concerned about us leaking really sensitive stuff in urls the embedly requests, so I'm wondering if we should see how good the results are if we remove basic auth/query strings (just for embedly). We can still show the full URL locally, just not send it out over the wire.

It could be that this is *too* restrictive, but let's play with this/try something out.

pdehaan:
r? @k88hudson 

Fixes #196

### DIFF
--- a/content-src/lib/utils.js
+++ b/content-src/lib/utils.js
@@ -18,11 +18,19 @@ module.exports = {
     return url.replace(/^((https?:)?\/\/)?(www\.)?/i, "").toLowerCase();
   },
 
-  sanitizeUrl(url) {
+  sanitizeUrl(site) {
+    if (!site) {
+      return "";
+    }
+
     const ALLOWED_QUERY_PARAMS = new Set(["id", "p", "q", "query", "s", "search", "sitesearch"]);
     const REMOVE_KEYS = ["auth", "password", "username"];
-    const parsedUrl = urlParse(url, true);
+    const parsedUrl = (typeof site === "string") ? urlParse(site, true) : site.parsedUrl;
     const safeQueryParams = {};
+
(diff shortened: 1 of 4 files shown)


pdehaan:
Here's a screenshot of my local data. The full, unedited URL is still displayed in the activity stream, but you can see the embedly proxy is only sent the pre-approved list of query params.

new_tab

content-src/lib/utils.js
```diff
@@ -1,16 +1,53 @@
+var urlParse = require("url-parse");
```
k88hudson: `const` please! we should add an eslint rule for that

content-src/lib/utils.js
```diff
@@ -1,16 +1,53 @@
...
     }
     return url.replace(/^((https?:)?\/\/)?(www\.)?/i, "").toLowerCase();
+  },
+
+  sanitizeUrl(site) {
+    if (!site) {
+      return "";
+    }
+
+    const ALLOWED_QUERY_PARAMS = new Set(["id", "p", "q", "query", "s", "search", "sitesearch"]);
```
k88hudson: should we store these at the top of the file perhaps?

k88hudson:
Looks good to me! Just a couple of nits

content-src/lib/utils.js
pdehaan: I'll add the ESLint rule in https://github.com/mozilla/activity-streams/pull/218#discussion_r54442320.

PR #218, opened 2016-02-26
feat(lint): Update to eslint@2 and add more ESLint rules

### LINKED ISSUE #112: Add more ESLint rules

pdehaan:
Output is a bit more obnoxious now with about 19 warnings currently (but 10 of those come from /lib/PlacesProvider.js).

This doesn't solve the other bug of adding more React-specific rules, but I'll attempt that in another PR once I get my courage up.

Fixes #112


test/test-TabTracker.js
```diff
@@ -93,7 +93,7 @@ exports.test_TabTracker_reactivating = function*(assert) {
   let tabClosedPromise = new Promise(resolve => {
     for (let i in openTabs) {
       openTabs[i].close(() => {
-        if (i == openTabs.length - 1) {
+        if (Number(i) === openTabs.length - 1) {
```
pdehaan: Turns out, `i` was a String here, so `==` was doing the casting thing. I had to explicitly wrap in `Number()` otherwise tests were a failing.

package.json
```diff
@@ -97,6 +97,7 @@
     "start:server": "live-server data/content --port=1963 --no-browser",
     "firefox": "jpm run -b beta",
     "test": "npm-run-all test:*",
+    "pretest": "npm run bundle",
```
pdehaan: Turns out, tests were failing if you ran `npm run clean` and then `npm test` because required assets weren't generated. So I added a `pre` hook here to make sure `bundle` gets called. Perhaps this was too much and slows down builds/tests though.

lib/PlacesProvider.js
```diff
@@ -475,9 +475,9 @@ Links.prototype = {
 /**
  * Singleton that serves as the default link provider for the grid.
  */
-const gLinks = new Links();
+const gLinks = new Links(); // eslint-disable-line no-use-before-define
```
pdehaan: I don't know how to solve this. Currently this little line is causing about 10 warnings from ESLint:
```
/Users/pdehaan/dev/github/activity-streams/lib/PlacesProvider.js
   87:7   warning  'gLinks' was used before it was defined  no-use-before-define
   93:7   warning  'gLinks' was used before it was defined  no-use-before-define
  101:9   warning  'gLinks' was used before it was defined  no-use-before-define
  113:7   warning  'gLinks' was used before it was defined  no-use-before-define
  117:7   warning  'gLinks' was used before it was defined  no-use-before-define
  134:9   warning  'gLinks' was used before it was defined  no-use-before-define
  135:11  warning  'gLinks' was used before it was defined  no-use-before-define
  142:9   warning  'gLinks' was used before it was defined  no-use-before-define
  148:9   warning  'gLinks' was used before it was defined  no-use-before-define
  149:11  warning  'gLinks' was used before it was defined  no-use-before-define
```

I couldn't find the optimal place to put this magic little line where `Links` was defined and tests would still pass. And I didn't want to solve it by adding 10 `// eslint-disable-line no-use-before-define` hacks all over the place.

content-src/lib/ActionManager.js
```diff
@@ -41,7 +41,7 @@ class ActionManager {
       throw new Error("Looks like your action definition does not return an object.");
     }
     if (!action.type) {
-      throw new Error(`You must define a type for an action.`);
+      throw new Error("You must define a type for an action.");
```
pdehaan: Turns out ESLint doesn't like using the backtick strings if you don't have any nested variables. So clever.

.eslintrc
```diff
@@ -8,34 +23,48 @@
...
+    "no-shadow": 1, // TODO: Remove this?
     "no-trailing-spaces": [2, {"skipBlankLines": false}],
     "no-undef": 2,
-    "no-unused-vars": 2,
+    "no-underscore-dangle": 0,
+    "no-use-before-define": [1, {"functions": false, "classes": false}],
+    "no-unused-vars": [2, {"vars": "all", "args": "none"}],
+    "no-warning-comments": 1,
     "object-curly-spacing": [2, "never"],
+    "prefer-reflect": 0,
```
pdehaan: Add `"prefer-const": 1,` here.

pdehaan: Turns out adding `prefer-const: 1` throws 204 warnings, so it will be mildly disruptive.

pdehaan: Add [`"no-var", 2`](http://eslint.org/docs/rules/no-var). I think that's the rule we want too. It'll fail linting if somebody uses `var`. Whereas [`prefer-const`](http://eslint.org/docs/rules/prefer-const) controls whether we should use `const` for unchanging values.

.eslintrc
```diff
@@ -8,34 +23,50 @@
...
+    "no-shadow": 1, // TODO: Remove this?
     "no-trailing-spaces": [2, {"skipBlankLines": false}],
     "no-undef": 2,
-    "no-unused-vars": 2,
+    "no-underscore-dangle": 0,
+    "no-use-before-define": [1, {"functions": false, "classes": false}],
+    "no-unused-vars": [2, {"vars": "all", "args": "none"}],
+    "no-warning-comments": 1,
     "object-curly-spacing": [2, "never"],
+    "no-var": 2,
```
pdehaan: Ref: https://github.com/mozilla/activity-streams/pull/217#discussion_r54427153

lib/PlacesProvider.js
k88hudson: I don't really feel like  `no-use-before-define` is necessary anymore since `const`/`let` throw `ReferenceError`s if they're defined after usage, unlike `var`, so I wouldn't be opposed to leaving out this rule

package.json
k88hudson: :+1:

lib/PlacesProvider.js
pdehaan: Deleted `no-use-before-define` from .eslintrc (and inline above).

k88hudson:
Ok, looks good to me! R+

PR #227, opened 2016-02-29
Alpha fixes

k88hudson:
Just a couple of little touch-ups before we ship alpha

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -29,7 +29,7 @@ const NewTabPage = React.createClass({
           </section>
 
           <section>
-            <GroupedActivityFeed title="Top Activity" sites={props.History.rows} length={20} />
+            <GroupedActivityFeed title="Top Activity" sites={props.History.rows} length={6} />
           </section>
         </div>
       </div>
--- a/content-src/components/Search/Search.js
+++ b/content-src/components/Search/Search.js
@@ -6,14 +6,20 @@ const Search = React.createClass({
       onSearch: function() {}
     };
   },
+  doSearch(value) {
+    if (!value) {
+      return;
+    }
+    this.props.onSearch(value);
+  },
   render() {
-    return (<div className="search-wrapper">
+    return (<form className="search-wrapper">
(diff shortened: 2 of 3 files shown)


k88hudson:
@rlr @oyiptong or @pdehaan R?

content-src/components/Search/Search.js
```diff
@@ -7,13 +7,13 @@ const Search = React.createClass({
     };
   },
   render() {
-    return (<div className="search-wrapper">
+    return (<form className="search-wrapper">
       <span className="search-label fa fa-search" />
-      <input ref="input" placeholder="Search" required />
+      <input ref="input" type="search" placeholder="Search" required />
```
pdehaan: I'm a bit confused here. Even though we specify `required`, I can leave the field blank and click Enter or press the arrow icon and I get a brief tooltip pop up saying "Please fill out this field." and then the form immediately submits to Yahoo!.

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -29,7 +29,7 @@ const NewTabPage = React.createClass({
           </section>
 
           <section>
-            <GroupedActivityFeed title="Top Activity" sites={props.History.rows} length={20} />
+            <GroupedActivityFeed title="Top Activity" sites={props.History.rows} length={6} />
```
pdehaan: Was this to match a UX spec or prevent v-scrolling or something, ~~or just a speed thing so we don't need to send as many URLs to embedly proxy~~ (nope, I'm still sending 11 `urls=` to embedly proxy, we're just only showing 6 of them)?

k88hudson: We use all the URLs on the Timeline page, just not on NewTab

content-src/components/Search/Search.js
k88hudson: ah good call, ill add real disabling

pdehaan:
Cool, thanks for the clarification.

R+ from me, with that one disabled form still submitting nit thing.

k88hudson:
@pdehaan thanks!

Issue #219, opened 2016-02-29
Deployment script for addons


oyiptong:
We need a deployment script that will:

1. package the addon
2. sign the addon
3. set the upload URL
4. upload to s3

pdehaan:
Related: https://github.com/mozilla/testpilot/issues/406 and https://github.com/mozilla/testpilot/issues/92

oyiptong:
very relevant indeed

this deployment script will be temporary until mozilla/testpilot#406 is complete

lmorchard:
Also, we should share notes. Whatever ends up working for you is probably the kind of thing we want to do for Test Pilot itself - and hopefully we make it as convenient as possible!

PR #230, opened 2016-02-29
chore(deploy): #219 s3 deployment scripts for addon

oyiptong:
This change includes a Fabric deployment script that uploads to the following URLs:

* https://moz-activity-streams.s3.amazonaws.com/dist/activity_streams_experiment-[VERSION_NUMBER]-fx.xpi
* https://moz-activity-streams.s3.amazonaws.com/dist/activity-streams-latest.xpi
* https://moz-activity-streams.s3.amazonaws.com/dist/update.rdf

The update URL is checked by the addon manager. This file points to `activity-streams-latest.xpi`.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -1,0 +1,127 @@
+import time
+import hashlib
+import json
+from operator import itemgetter
+from distutils.util import strtobool
+import boto
+from fabric.api import local, env
+from boto.s3.key import Key
+from pathlib import Path
+
+env.bucket_name = "moz-activity-streams"
+S3 = boto.connect_s3()
+
+
+def to_bool(value):
+    if not isinstance(value, bool):
+        return strtobool(value)
+    else:
+        return value
+
+
+def get_s3_headers():
+    return {
+        'Cache-Control': 'public, max-age={}'.format(24*60*60),
+        'Content-Disposition': 'inline',
+    }
+
+
+def package(signing_key, signing_password):
+    local("npm install")
+    local("npm run package")
+    local("./node_modules/jpm/bin/jpm sign --api-key {} --api-secret {}"
+          .format(signing_key, signing_password))
+    local("mv activity_streams_experiment-*.xpi dist/")
+    local("mv \@activity-streams-*.update.rdf dist/update.rdf")
+    local("rm dist/activity-streams-*.xpi")
+
+
+def get_packages():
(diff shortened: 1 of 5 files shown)


oyiptong:
This also includes a version bump to 1.0.0

r? @rlr, @k88hudson

oyiptong:
Note:

* this skips an upload if MD5 checksums match, from S3 to local
* this requires an AMO api key and secret, which is left off the repository
* the command will fail if the signature process fails

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "activity-streams-experiment",
   "name": "activity-streams",
-  "version": "0.0.1",
```
k88hudson: Maybe we should update this with `npm version`, rather than in this patch? That way we can tag/do the package.json update at the same time

fabfile.py
```diff
@@ -0,0 +1,127 @@
...
+    update_manifest.name = "dist/update.rdf"
+    update_manifest.set_contents_from_filename(
+        "./dist/update.rdf", headers=headers)
+    update_manifest.set_acl("public-read")
+
+    return (k, latest, update_manifest)
+
+
+def deploy(run_package=True, signing_key=None, signing_password=None):
+    if not (signing_key is not None and signing_password is not None):
```
k88hudson: do we maybe want to throw if we don't have this, since we always want to sign?

oyiptong:
Deployment is run after creating a virtualenv and instantiating it.

e.g.
```sh
# create a virtualenv
$ virtualenv ./activity-streams-env
# activate the virtualenv
$ . ./activity-streams/env/bin/activate
# install dependencies
(activity-streams-env) $ pip install -r ./requirements.txt
```

to run the deployment script:
```sh
(activity-streams-env) $ fab deploy
```
Running the deployment script requires a virtualenv instantiated, or the dependencies existing in the current python installation.

This only runs on Python 2.7

package.json
oyiptong: what's `npm version`?

k88hudson:
Ok looks good (although I haven't tested it locally, probably fine just to merge and test it out)

fabfile.py
oyiptong: It will throw if we don't have those. It will read them from a config file if not presented.

k88hudson:
R+

PR #231, opened 2016-02-29
fix(addon): make addon run on Firefox 45

oyiptong:
r? @pdehaan

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "author": "Mozilla",
   "engines": {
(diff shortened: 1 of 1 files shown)


pdehaan:
@oyiptong Clicking the "New Tab" icon now seems to set the page's address to:
<resource://activity-streams/data/content/activity-streams.html#/>

oyiptong:
That's a regression tackled elsewhere, not introduced in this bug

pdehaan:
R+

Issue #206, opened 2016-02-25
Removing the add-on doesn't restore the new tab page


k88hudson:
STR:

1. Install add-on, add-on is working
2. Remove add-on
3. Visit new tab page

Actual: new tab is blank
Expected: new tab page should be old new tab page

rlr:
I tried adding `NewTabURL.reset();` to the `unload()` of `ActivityStreams` but no luck.

rlr:
After restarting the browser the normal newtab page is back though. Seems like possibly a bug in the reset code?

rlr:
landed

PR #233, opened 2016-03-01
fix(addon): #206 reset the NewTabURL on unload

rlr:
Turns out the `ActivityStreams.unload()` was never being called.

r? @oyiptong @k88hudson


oyiptong:
r+

PR #235, opened 2016-03-01
Closes #198 - Add some sort of pagination to Recent Links API query

### LINKED ISSUE #198: Add some sort of pagination to Recent Links API query
We need a way to go further back in history. We also probably need a way to get "future" activity after the last time we queried. @k88hudson suggested we could just have a AFTER/BEFORE=&lt;timestamp&gt; parameter.

mzhilyaev:
Simply add beforeDate in milliseconds to  RequestRecentLinks() as in

For example to get recent links 3 days or older, do call below
```
RequestRecentLinks({beforeDate: Date.now - 3*24*60*60*1000})
```
@rlr @k88hudson  let me know if this is something that you expected

Fixes #198

### DIFF
--- a/content-src/actions/action-manager.js
+++ b/content-src/actions/action-manager.js
@@ -50,9 +50,6 @@ function RequestExpect(type, expect, options = {}) {
   if (options.timeout) {
     action.meta.timeout = options.timeout;
   }
-  if (options.query) {
-    action.query = options.query;
-  }
   if (options.data) {
     action.data = options.data;
   }
--- a/content-src/actions/action-manager.js
+++ b/content-src/actions/action-manager.js
@@ -53,6 +53,9 @@ function RequestExpect(type, expect, options = {}) {
   if (options.query) {
     action.query = options.query;
   }
+  if (options.data) {
+    action.data = options.data;
+  }
   return action;
 }
 
@@ -64,8 +67,8 @@ function RequestBookmarks() {
   return RequestExpect("RECENT_BOOKMARKS_REQUEST", "RECENT_BOOKMARKS_RESPONSE");
 }
 
-function RequestRecentLinks() {
-  return RequestExpect("RECENT_LINKS_REQUEST", "RECENT_LINKS_RESPONSE");
+function RequestRecentLinks(data) {
+  return RequestExpect("RECENT_LINKS_REQUEST", "RECENT_LINKS_RESPONSE", {data: data});
 }
 
 function RequestFrecentLinks() {
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -191,14 +191,15 @@ Links.prototype = {
    *
    * @param {Object} options
    *        {Integer} limit: Maximum number of results to return. Max 20
-   *        {Milliseconds} limitDate: Only return links fresher than limitDate
+   *        {Milliseconds} beforeDate: Only return links older than beforeDate
+   *        {Milliseconds} afterDate: Only return links fresher than afterDate
    *        {String} order: if order=="frecency", orders by frecency, otherwise by recency
    *
    * @returns {Promise} Returns a promise with the array of links as payload.
    */
   _getHistoryLinks: Task.async(function*(options={}) {
 
-    let {limit, limitDate, order} = options;
+    let {limit, afterDate, beforeDate, order} = options;
     if (!limit || limit.options > HISTORY_RESULTS_LIMIT) {
       limit = HISTORY_RESULTS_LIMIT;
     }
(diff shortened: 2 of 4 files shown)


oyiptong:
Xvfb strikes again.

```
JPM [error] Xlib:  extension "RANDR" missing on display ":99.0".
```

test/lib/utils.js
```diff
@@ -48,6 +48,11 @@ function doThrow(error, stack) {
   throw(new Error(`Error at ${filename}`));
 }
 
+function doDump(object, trailer) {
```
oyiptong: is this ever used?

test/test-PlacesProvider.js
```diff
@@ -158,6 +158,13 @@ exports.test_Links_getRecentLinks = function*(assert) {
   assert.equal(links[1].bookmarkDateCreated, new Date(bookmark.dateAdded).getTime(), "expected bookmark date for second link");
   assert.ok(isVisitDateOK(links[1].lastVisitDate), "visit date within expected range");
   assert.ok(!(links[0].bookmark || links[2].bookmark || links[3].bookmark), "other bookmarks are empty");
+
+  // test beforeDate functionality
+  let beforeDate = timeDaysAgo(1) / 1000 - 1000;
+  links = yield provider.getRecentLinks({beforeDate: beforeDate});
```
oyiptong: we need tests for `afterDate`

lib/PlacesProvider.js
```diff
@@ -187,14 +187,15 @@ Links.prototype = {
    *
    * @param {Object} options
    *        {Integer} limit: Maximum number of results to return. Max 20
-   *        {Milliseconds} limitDate: Only return links fresher than limitDate
+   *        {Milliseconds} beforeDate: Only return links older than beforeDate
+   *        {Milliseconds} afterDate: Only return links fresher than afterDate
```
oyiptong: I'm unsure we will need `afterDate`

content-src/actions/action-manager.js
```diff
@@ -64,8 +67,8 @@ function RequestBookmarks() {
   return RequestExpect("RECENT_BOOKMARKS_REQUEST", "RECENT_BOOKMARKS_RESPONSE");
 }
 
-function RequestRecentLinks() {
-  return RequestExpect("RECENT_LINKS_REQUEST", "RECENT_LINKS_RESPONSE");
+function RequestRecentLinks(data) {
+  return RequestExpect("RECENT_LINKS_REQUEST", "RECENT_LINKS_RESPONSE", {data: data});
```
k88hudson: this should be `query`; `data` is for responses

Also, you can do `{query}` instead of `{query: query}`

content-src/actions/action-manager.js
```diff
@@ -53,6 +53,9 @@ function RequestExpect(type, expect, options = {}) {
   if (options.query) {
     action.query = options.query;
   }
+  if (options.data) {
```
k88hudson: This can be removed (since there is already a check for `options.query`)

lib/PlacesProvider.js
mzhilyaev: afterDate is used by PlacesProvider.getFrecentLinks() - recall we wanted to get last 72 hours worth of links ordered by frecency

content-src/actions/action-manager.js
mzhilyaev: Kate, note that what you suggest is different from  how Notify is implemented.
https://github.com/mozilla/activity-streams/blob/master/content-src/actions/action-manager.js#L42

Also note that FX wants msg.data not msg.query as in:
https://github.com/mozilla/activity-streams/blob/master/lib/ActivityStreams.js#L80

I am fine either way, but we need to consistent.
Let me know which way you would like to clean the code, I can replace msg.data and msg.query in action handling.

k88hudson: The idea with `query` v.s. `data` is that `query` is supposed to be for "request" type actions, data fir responses (kind of like query strings v.s. payload) ... but you know what, now that we're discussing it, it's obviously confusing :stuck_out_tongue_winking_eye: 

So... let's just use `data`! :dancer:

content-src/actions/action-manager.js
```diff
@@ -53,6 +53,9 @@ function RequestExpect(type, expect, options = {}) {
   if (options.query) {
```
k88hudson: Can you actually delete the `options.query`as well, now that we've decided not the use it?

content-src/actions/action-manager.js
pdehaan: > So... let's just use `data`!

@k88hudson Do we need to update the [README](https://github.com/mozilla/activity-streams#conventions-for-actions) if we're getting rid of `query`?

k88hudson: Ah, good call. Forgot about that!

pdehaan: Thanks. Will file a new bug so we can do it before 1.0 (or whenever).

Issue #237, opened 2016-03-01
Failed propType: Invalid prop `length` of type `string` supplied to `GroupedActivityFeed`, expected `number`.


pdehaan:
I'm seeing a warning/error in my DevTools console when I click the "Activity Streams" icon in the Firefox toolbar.

![new_tab](https://cloud.githubusercontent.com/assets/557895/13436945/92730cb8-df96-11e5-88d7-a7bba1575ca9.png)

> **Warning: Failed propType: Invalid prop `length` of type `string` supplied to `GroupedActivityFeed`, expected `number`. Check the render method of `TimelinePage`.**

```jsx
<GroupedActivityFeed title="Just now" sites={props.History.rows} length="20" />
```

&mdash; via [./content-src/components/TimelinePage/TimelinePage.js:29](https://github.com/mozilla/activity-streams/blob/88d5e697e8e7108ea1bd4d5b61c3f9f33afe6fd6/content-src/components/TimelinePage/TimelinePage.js#L29)

I'm guessing it should be `{20}` like it is on [./content-src/components/NewTabPage/NewTabPage.js:32](https://github.com/mozilla/activity-streams/blob/d41acda2524569ed528c165dd55a7d0768002935/content-src/components/NewTabPage/NewTabPage.js#L32):

```jsx
<GroupedActivityFeed title="Top Activity" sites={props.History.rows} length={6} />
```

PR incoming.

---

**UPDATE:** Fixed in PR #238

PR #239, opened 2016-03-01
feat(content): #225 Bookmarks filter on Timeline

### LINKED ISSUE #171: Refactor Header component
Right now the `Header` component is inside the `Base` component because it used to exist in both the `NewTabPage` and `TimelinePage`, however, it is now only required in the `TimelinePage`.

In https://github.com/mozilla/activity-streams/pull/170 we just added some code to hide it, but we should refactor to move the `Header` component inside `TimelinePage` instead.

### LINKED ISSUE #225: Bookmarks filter/page
We don't actually have a filter for bookmarks that works on the Timeline view; it needs to be implemented. It should use the `RECENT_BOOKMARKS` data

k88hudson:
This adds sub-routing to the Timeline view, and a bookmark filter for the Timeline layout.

For now this is basically the same layout as "All", but it may change in the future.

I also added the `Header` component to the `TimelinePage` page instead of `Base`.

Fixes #225 
Fixes #171

### DIFF
--- a/content-src/components/TimelinePage/TimelineBookmarks.js
+++ b/content-src/components/TimelinePage/TimelineBookmarks.js
@@ -1,0 +1,17 @@
+const React = require("react");
+const {connect} = require("react-redux");
+const {GroupedActivityFeed} = require("components/ActivityFeed/ActivityFeed");
+const Spotlight = require("components/Spotlight/Spotlight");
+
+const TimelineBookmarks = React.createClass({
+  render() {
+    const props = this.props;
+    return (<div className="wrapper">
+      <Spotlight sites={props.Bookmarks.rows} />
+      <GroupedActivityFeed title="Just now" sites={props.Bookmarks.rows} length={20} />
+    </div>);
+  }
+});
+
+module.exports = connect(({History, Bookmarks}) => ({History, Bookmarks}))(TimelineBookmarks);
+module.exports.TimelineBookmarks = TimelineBookmarks;
--- a/content-src/components/TimelinePage/TimelineHistory.js
+++ b/content-src/components/TimelinePage/TimelineHistory.js
@@ -1,0 +1,17 @@
+const React = require("react");
+const {connect} = require("react-redux");
+const {GroupedActivityFeed} = require("components/ActivityFeed/ActivityFeed");
+const Spotlight = require("components/Spotlight/Spotlight");
+
+const TimelineHistory = React.createClass({
+  render() {
+    const {History} = this.props;
+    return (<div className="wrapper">
+      <Spotlight sites={History.rows} />
+      <GroupedActivityFeed title="Just now" sites={History.rows} length={20} />
+    </div>);
+  }
+});
+
+module.exports = connect(({History}) => ({History}))(TimelineHistory);
+module.exports.TimelineHistory = TimelineHistory;
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -223,6 +223,7 @@ ActivityStreams.prototype = {
         baseUrl,
         `${baseUrl}#/`,
         `${baseUrl}#/timeline`,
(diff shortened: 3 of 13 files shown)


lib/ActivityStreams.js
```diff
@@ -223,6 +223,7 @@ ActivityStreams.prototype = {
         baseUrl,
         `${baseUrl}#/`,
         `${baseUrl}#/timeline`,
+        `${baseUrl}#/timeline/bookmarks`
```
k88hudson: This doesn't actually seem to be working but i'm assuming rlr is fixing this in https://github.com/mozilla/activity-streams/issues/223

content-src/components/TimelinePage/TimelineBookmarks.js
```diff
@@ -0,0 +1,17 @@
...
+const {connect} = require("react-redux");
+const {GroupedActivityFeed} = require("components/ActivityFeed/ActivityFeed");
+const Spotlight = require("components/Spotlight/Spotlight");
+
+const TimelineBookmarks = React.createClass({
+  render() {
+    const props = this.props;
+    return (<div className="wrapper">
+      <Spotlight sites={props.Bookmarks.rows} />
+      <GroupedActivityFeed title="Just now" sites={props.Bookmarks.rows} length="20" />
```
pdehaan: Should this be `length={20}`? (see #237)

content-src/components/TimelinePage/TimelineHistory.js
```diff
@@ -0,0 +1,17 @@
...
+const {connect} = require("react-redux");
+const {GroupedActivityFeed} = require("components/ActivityFeed/ActivityFeed");
+const Spotlight = require("components/Spotlight/Spotlight");
+
+const TimelineHistory = React.createClass({
+  render() {
+    const {History} = this.props;
+    return (<div className="wrapper">
+      <Spotlight sites={History.rows} />
+      <GroupedActivityFeed title="Just now" sites={History.rows} length="20" />
```
pdehaan: Should this be `length={20}`?

content-src/components/TimelinePage/TimelineBookmarks.js
k88hudson: :+1:

pdehaan:
@k88hudson Does this also fix #171?

k88hudson:
@pdehaan ah yes, it does!

pdehaan:
LGTM locally, and the code looks good &mdash; but I'll let @oyiptong and @rlr take a look before pressing the Big Green Button(tm).

lib/PlacesProvider.js
```diff
@@ -349,7 +350,7 @@ Links.prototype = {
                     LIMIT :limit`;
 
     let links = yield this.executePlacesQuery(sqlQuery, {
-                  columns: ["bookmarkId", "bookmarkTitle", "bookmarkGuid", "url", "title", "lastVisitDate", "frecency", "type", "lastModified", "favicon", "mimeType"],
+                  columns: ["bookmarkId", "bookmarkTitle", "bookmarkGuid", "bookmarkDateCreated", "url", "title", "lastVisitDate", "frecency", "type", "lastModified", "favicon", "mimeType"],
```
oyiptong: let's add a test in test-PlacesProvider, just so we have some regression safety net

k88hudson:
@oyiptong ok, tests updated

oyiptong:
i'll defer to @rlr for the react component code, r+ on the query side of things

rlr:
looking!

lib/ActivityStreams.js
rlr: :+1: I'll make sure that one works too.

rlr:
Nice cleanup with header. New timeline component LGTM. r=me

PR #244, opened 2016-03-01
fix(addon): #243 execute search with correct engine

### LINKED ISSUE #150: Implement Search for the newtab page
We need to implement search in the newtab page of Activity Streams.

We can port over the implementation of the search bar from the remote-newtab implementation, both chrome-side and content-side.

### LINKED ISSUE #243: search form field always uses yahoo
1. install xpi, 
2. set browser search engine to duckDuckGo
3. open new tab
4. type in search box and hit return

actual: uses yahoo

expected: use my firefox search engine choice: duckDuckGo

sarracini:
r? @k88hudson @oyiptong 

Fixes #243 
Fixes #150

### DIFF
--- a/content-src/reducers/reducers.js
+++ b/content-src/reducers/reducers.js
@@ -21,8 +21,35 @@ function setRowsOrError(type) {
   };
 }
 
+function setSearchState(type) {
+  return (prevState = {currentEngine: {}, error: false}, action) => {
+    const state = {};
+    switch (action.type) {
+      case am.type(type):
+        if (action.error) {
+          state.currentEngine = {};
+          state.error = action.data;
+        } else {
+          if (!action.data || !action.data.currentEngine) {
+            state.currentEngine = {};
+            state.error = true;
+          } else {
+            state.error = false;
+            state.currentEngine = action.data.currentEngine;
+          }
+        }
+        break;
+      // TODO: Handle changes
+      default:
+        return prevState;
+    }
+    return Object.assign({}, prevState, state);
+  };
+}
+
 module.exports = {
   TopSites: setRowsOrError("TOP_FRECENT_SITES_RESPONSE"),
   History: setRowsOrError("RECENT_LINKS_RESPONSE"),
-  Bookmarks: setRowsOrError("RECENT_BOOKMARKS_RESPONSE")
+  Bookmarks: setRowsOrError("RECENT_BOOKMARKS_RESPONSE"),
+  Search: setSearchState("SEARCH_STATE_RESPONSE")
 };
--- a/lib/SearchProvider.js
+++ b/lib/SearchProvider.js
@@ -1,0 +1,69 @@
+/* global XPCOMUtils, ContentSearch, Task, Services, EventEmitter */
+
+"use strict";
+const {Ci, Cu} = require("chrome");
+
+const CURRENT_ENGINE = "browser-search-engine-modified";
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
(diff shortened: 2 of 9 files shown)


pdehaan:
> The command "nvm use 5.5.0" failed and exited with 1 during .

That's a new one... We should also possibly consider bumping Travis-CI up to a newer [Node 5.x](https://nodejs.org/en/blog/vulnerability/february-2016-security-releases/) build.

oyiptong:
travis is quite the finicky CI service

sarracini:
this just needed a rebase, my bad

lib/SearchProvider.js
```diff
@@ -0,0 +1,69 @@
...
+  return EventEmitter;
+});
+
+let NewTabSearchProvider = function NewTabSearchProvider() {
+  EventEmitter.decorate(this);
+};
+
+NewTabSearchProvider.prototype = {
+
+  observe(subject, topic, data) { // jshint unused:false
```
oyiptong: Can we have tests for the `SearchProvider` please?

oyiptong: Can we have tests for the `SearchProvider` please?

pdehaan: Also, we use ESLint (instead of JSHint), so the `jshint unused` is, in fact, unused.

oyiptong:
The addon code looks good. We just need some tests.

I'll defer to @k88hudson for the content code

content-src/reducers/reducers.js
```diff
@@ -21,8 +21,34 @@ function setRowsOrError(type) {
   };
 }
 
+function setSearchState(type) {
```
k88hudson: Since this reducer is unique to `Search` you can just have this be a single function

content-src/reducers/reducers.js
```diff
@@ -21,8 +21,34 @@ function setRowsOrError(type) {
...
+  return (prevState = {currentEngine: {}, error: false}, action) => {
+    const state = {};
+    switch (action.type) {
+      case am.type(type):
+        if (action.error) {
+          state.currentEngine = {};
+          state.error = action.data;
+        } else {
+          state.error = false;
+          state.currentEngine = {
```
k88hudson: I think it would be a good idea to have a check here for `action.data` and `action.data.currentEngine` or a try/catch in case something goes unexpectedly wrong.

Also, you're probably safe to pass `action.data.currentEngine`, since actions are plain objects

content-src/reducers/reducers.js
sarracini: @k88hudson i thought it might be useful to keep the reducer here now, for future search reducer stuff?

oyiptong:
While as it is, the `NewTabSearchProvider` is just an interface to `ContentSearch.jsm`, we should have at least some sanity checks for regression purposes.

The tests don't need to be elaborate, as in, we don't need to test `ContentSearch`'s functionality, however, we want to know if something changed in `ContentSearch`, especially if our current code is not in m-c.

oyiptong:
r+

sarracini:
before sending down a message from the addon to content, we have to make sure the worker is there. Specifically when getting the search state, the function is asynchronous and the worker might either not be initialized yet or might have disappeared already and so the message won't go through.

Yay race conditions!

pdehaan:
@sarracini Can we remove `// TODO: Replace with real search api via addon` from [/content-src/components/NewTabPage/NewTabPage.js:11](https://github.com/mozilla/activity-streams/blob/38b400d6330776eecfaf2fc53a6b591d7ad3c3d2/content-src/components/NewTabPage/NewTabPage.js#L11), or is that still a valid TODO that needs to be TODONE?

Issue #223, opened 2016-02-29
opening new tab should have url selected


edmoz:
1. open a new tab
2. start typing

actual: when i type - it appends it to the end of the current url as such:
resource://activity-streams/data/content/activity-streams.html#/<MY_INPUT_GOES_HERE>

this is a regression from current Release Firefox - this is my normal flow for opening new URLs

expected: the URL should be selected or delete the current URL upon text input.

k88hudson:
See https://github.com/mozilla/activity-streams/issues/236 for more details

edmoz:
word is that there's an upstream change that broke us...

oyiptong:
@rlr says this broke our url bar clearing logic: https://github.com/mozilla/gecko-dev/commit/96797e42b2ed2ead220c6fbb7008f05cb6331bd9

rlr:
landed!

k88hudson:
Nice work @rlr!

PR #254, opened 2016-03-02
fix(addon): #223 hide app URLs in all cases

rlr:
r? @oyiptong @k88hudson

### DIFF
--- a/content-src/actions/action-manager.js
+++ b/content-src/actions/action-manager.js
@@ -16,7 +16,8 @@ const am = new ActionManager([
   "NOTIFY_PERFORM_SEARCH",
   "RECEIVE_CURRENT_ENGINE",
   "SEARCH_STATE_REQUEST",
-  "SEARCH_STATE_RESPONSE"
+  "SEARCH_STATE_RESPONSE",
+  "NOTIFY_ROUTE_CHANGE",
 ]);
 
 // This is a a set of actions that have sites in them,
@@ -88,6 +89,10 @@ function NotifyPerformSearch(data) {
   return Notify("NOTIFY_PERFORM_SEARCH", data);
 }
 
+function NotifyRouteChange(data) {
+  return Notify("NOTIFY_ROUTE_CHANGE", data);
+}
+
 am.defineActions({
   Notify,
   Response,
@@ -99,6 +104,7 @@ am.defineActions({
   RequestSearchState,
   NotifyHistoryDelete,
   NotifyPerformSearch,
+  NotifyRouteChange,
 });
 
 module.exports = am;
--- a/content-src/components/Routes/Routes.js
+++ b/content-src/components/Routes/Routes.js
@@ -1,21 +1,26 @@
 const React = require("react");
 const {Router, Route, IndexRoute} = require("react-router");
 const {createHashHistory} = require("history");
+const {connect} = require("react-redux");
+const {actions} = require("actions/action-manager");
 
 const history = createHashHistory({queryKey: false});
 
 const Routes = React.createClass({
+  onChange(data) {
+    this.props.dispatch(actions.NotifyRouteChange(data));
+  },
   render() {
     return (<Router history={history}>
       <Route path="/" component={require("components/Base/Base")}>
-        <IndexRoute title="Home" component={require("components/NewTabPage/NewTabPage")} />
+        <IndexRoute title="Home" component={require("components/NewTabPage/NewTabPage")} onEnter={data => this.onChange(data)} />
         <Route title="Activity Stream" path="timeline" component={require("components/TimelinePage/TimelinePage")}>
-          <IndexRoute title="History" component={require("components/TimelinePage/TimelineHistory")} />
-          <Route title="History" path="bookmarks" component={require("components/TimelinePage/TimelineBookmarks")} />
+          <IndexRoute title="History" component={require("components/TimelinePage/TimelineHistory")} onEnter={data => this.onChange(data)} />
+          <Route title="History" path="bookmarks" component={require("components/TimelinePage/TimelineBookmarks")} onEnter={data => this.onChange(data)} />
         </Route>
       </Route>
     </Router>);
   }
 });
 
-module.exports = Routes;
+module.exports = connect(() => ({}))(Routes);
--- a/content-src/components/Routes/Routes.js
+++ b/content-src/components/Routes/Routes.js
@@ -7,16 +7,21 @@ const {actions} = require("actions/action-manager");
 const history = createHashHistory({queryKey: false});
 
 const Routes = React.createClass({
-  onChange(data) {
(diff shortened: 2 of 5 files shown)


content-src/actions/action-manager.js
```diff
@@ -12,7 +12,8 @@ const am = new ActionManager([
   "RECENT_LINKS_RESPONSE",
   "FRECENT_LINKS_REQUEST",
   "FRECENT_LINKS_RESPONSE",
-  "NOTIFY_HISTORY_DELETE"
+  "NOTIFY_HISTORY_DELETE",
+  "NOTIFY_ROUTE_CHANGE",
```
rlr: Added a new redux action to notify the addon of a route change.

lib/ActivityStreams.js
```diff
@@ -33,11 +33,11 @@ function ActivityStreams(options = {}) {
   this._setupPageMod();
   this._setupListeners();
   this._setupButton();
-  this._hideAppURLs();
-  NewTabURL.override(this.options.pageURL);
+  NewTabURL.override(`${this.options.pageURL}#/`);
   if (this.options.telemetry) {
     this._tabTracker = new TabTracker(this.appURLs);
   }
+  this._appURLHider = new AppURLHider(this.appURLs);
```
rlr: I moved all the URL hiding code into a separate `AppURLHider` object.

lib/AppURLHider.js
```diff
@@ -0,0 +1,104 @@
...
+
+function AppURLHider(appURLs) {
+  this._appURLs = appURLs;
+
+  this.maybeHideURL = this.maybeHideURL.bind(this);
+  tabs.on("open", this.maybeHideURL);
+  tabs.on("ready", this.maybeHideURL);
+  tabs.on("load", this.maybeHideURL);
+  tabs.on("activate", this.maybeHideURL);
+  tabs.on("pageshow", this.maybeHideURL);
```
rlr: That might be overkill of the events to listen to. lol

lib/AppURLHider.js
```diff
@@ -0,0 +1,104 @@
...
+    if (this.isAppURL(tab.url)) {
+      this.hideURL();
+    }
+  },
+
+  /**
+   * Add the app urls to the hidden pages of the passed in window.
+   */
+  _addHiddenURLsTo(window) {
+    if (window.gInitialPages && !window.gInitialPages.includes(this._appURLs[0])) {
```
rlr: I kept this because without it you can see the URL flash in then out as the tab loads.

lib/ActivityStreams.js
```diff
@@ -33,11 +33,11 @@ function ActivityStreams(options = {}) {
   this._setupPageMod();
   this._setupListeners();
   this._setupButton();
-  this._hideAppURLs();
-  NewTabURL.override(this.options.pageURL);
+  NewTabURL.override(`${this.options.pageURL}#/`);
```
oyiptong: This might not be as synchronous as you'd expect. You might want to block on obtaining the `newtab-url-changed` observer notification.

This notification is sent from https://dxr.mozilla.org/mozilla-central/source/browser/components/newtab/aboutNewTabService.js#232

oyiptong: Actually, since this is in the parent process, I'm probably wrong, and this is probably synchronous.

lib/AppURLHider.js
oyiptong: let's find one and stick with it ;-)

lib/AppURLHider.js
```diff
@@ -0,0 +1,104 @@
...
+      };
+    }
+    return this._cachedWindowObserver;
+  },
+
+  /**
+   * Adds app urls to the list of hidden awesomebar URLs of each open window and
+   * sets up an observer for new windows that are opened.
+   */
+  _hideAppURLs() {
```
oyiptong: We should have a sanity test.

One of the tests would consist of loading an appURL and asserting that the the url bar value is the empty string.
Another test could test for '#' url changes.

oyiptong: This will have the benefit of alerting us whenever something changes from under our feet, for regression purposes.

lib/AppURLHider.js
rlr: It might just be `activate` because the rest should be handled by `gInitialPages` and the route change handler. I'll play with it.

lib/AppURLHider.js
rlr: We had decent tests but they didn't test hash changing since we were loading a dummy activity-streams.html... I think that's all we have to add to the tests.

content-src/components/Routes/Routes.js
```diff
@@ -1,21 +1,26 @@
...
 
 const Routes = React.createClass({
+  onChange(data) {
+    this.props.dispatch(actions.NotifyRouteChange(data));
+  },
   render() {
     return (<Router history={history}>
       <Route path="/" component={require("components/Base/Base")}>
-        <IndexRoute title="Home" component={require("components/NewTabPage/NewTabPage")} />
+        <IndexRoute title="Home" component={require("components/NewTabPage/NewTabPage")} onEnter={data => this.onChange(data)} />
```
k88hudson: Instead of this, add a listener to history (on line 7) 

```
history.listen(location => {
  this.props.dispatch....
})

```

https://github.com/mjackson/history#basic-usage

k88hudson: Another thing we might want to consider is not using url-based routing at all (i.e. don't use browserHistory, store route state in an object instead) if we're never going to be storing things in history or showing the URL anyway

rlr: That would fix the issue too and this patch wouldn't be required. I didn't know it was possible.

rlr:
feedback applied and test added ^^

test/test-awesomebar-empty.js
```diff
@@ -64,4 +65,32 @@ exports["test awesomebar is empty for all app urls in new windows too"] = functi
...
+  let app = new ActivityStreams({pageURL: url});
+
+  yield new Promise(resolve => tabs.open({
+    url: url,
+    onReady: (tab) => {
+      let browserWindow = windowMediator.getMostRecentWindow("navigator:browser");
+      assert.equal(browserWindow.gURLBar.value, "");
+
+      // change the hash and verify url bar is still empty.
+      tab.url = url + "#/timeline";
```
rlr: hmm. not sure this is testing what I want.

rlr:
ok, I like that test better. r?

pdehaan:
@rlr Looks like this needs a rebase.

rlr:
thanks @pdehaan! rebased!

pdehaan:
Works great locally for me!
r? @oyiptong @k88hudson

oyiptong:
r+

PR #256, opened 2016-03-03
fix(entropy): Grab the first available image instead of highest entropy

### LINKED ISSUE #192: Entropy isn't a great metric for how relevant an image is
I've been playing with `npm run once` since it gives me a nice clean profile each time, and I'm [starting to notice](https://github.com/mozilla/activity-streams/pull/170#issuecomment-187414998) that it isn't always choosing the most relevant photo, just the [most complex](http://docs.embed.ly/docs/extract#response-attributes).

> "Image entropy can be roughly thought of as how "busy" an image is. Image entropy can be useful in programmatically choosing the type of image to display. For instance, if an API user wants to display photographic type images, but not logos, they can ignore images with lower entropy."
>
> &mdash; via http://docs.embed.ly/docs/extract#response-attributes (see `images` attribute docs)

I've been browsing www.ign.com reviews and am anecdotally noticing that the first image in the `images[]` array is usually the most accurate. In my case, the image with the highest entropy that gets chosen, is a random avatar on the page, rather than the massive header image for the article.

new_tab

pdehaan:
r? @k88hudson /cc @nchapman 

Fixes #192

### DIFF
--- a/content-src/components/Spotlight/Spotlight.js
+++ b/content-src/components/Spotlight/Spotlight.js
@@ -27,9 +27,7 @@ function getBestImage(images) {
     return null;
   }
 
-  return filteredImages.reduce((prev, next) => {
-    return next.entropy > prev.entropy ? next : prev;
-  }) || null;
+  return filteredImages[0];
 }
 
 const SpotlightItem = React.createClass({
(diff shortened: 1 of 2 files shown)


k88hudson:
R+

PR #269, opened 2016-03-05
Adding a couple of data attributes

emtwo:


lib/TabTracker.js
```diff
@@ -4,10 +4,13 @@ const tabs = require("sdk/tabs");
 const {Cu} = require("chrome");
 
 Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/ClientID.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
```
emtwo: Nit: don't need these two imports

lib/TabTracker.js
```diff
@@ -81,10 +88,22 @@ TabTracker.prototype = {
...
+    return function(tab) {
+      if (this.isActivityStreamsURL(tab.url)) {
+        this.navigateAwayFromPage(tab, reason);
+      }
+    }.bind(this);
+  },
+
+  logClose: function(reason) {
+    return function(tab) {
+      if (tabs.activeTab.url != tab.url) {
```
pdehaan: Via Travis-CI lint warnings:
```
/home/travis/build/mozilla/activity-streams/lib/TabTracker.js
  101:30  warning  Expected '!==' and instead saw '!='  eqeqeq
```

Not sure if we should promote that from a lint warning to a failure.

lib/main.js
```diff
@@ -18,7 +20,11 @@ Object.assign(exports, {
     PlacesProvider.links.init();
     SearchProvider.search.init();
     options.telemetry = false;
-    app = new ActivityStreams(options);
+
+    Task.spawn(function*() {
+      options.clientID = yield ClientID.getClientID();
```
oyiptong: I'll confirm if that's the client ID we want. This will yield us the telemetry client id. IIRC we wanted an FxA client ID/

Also, please remove mention of `storage.clientUUID` in lib/main.js, it won't be needed if we use another client id.

lib/TabTracker.js
oyiptong: :+1:

lib/TabTracker.js
```diff
@@ -96,8 +115,8 @@ TabTracker.prototype = {
 
     tab.on("ready", this.logReady);
     tab.on("activate", this.logActivate);
-    tab.on("deactivate", this.logDeactivate);
-    tab.on("close", this.logDeactivate);
+    tab.on("deactivate", this.logDeactivate("unfocus"));
```
oyiptong: good use of currying :+1:

lib/main.js
oyiptong: @vladikoff suggested to use `identity.fxaccounts.lastSignedInUserHash`. He will confirm it.

We can use `ClientID.getClientID()` as a fallback in case the user is not signed in to FxA.

oyiptong: Actually, scratch that. Telemetry Client ID is good enough.

@emtwo, please remove the older clientUUID declaration in the same file

oyiptong:
r+ with nits fixed

lib/main.js
vladikoff: Hey, thanks for asking about the FxA stuff :)

We had a bit of discussion about this with some of the FxA team. The suggestion would be to **not** use the fxa uid as a primary identifier for this. Create your own or use the telemetry one. 

Also do not use `identity.fxaccounts.lastSignedInUserHash` in other metrics fields, this pref is used for a specific account flow and might not always be there. If you want to include the fxa details in your metrics feilds then fetch it using this component: http://mxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#462

You can also get the device id using: http://mxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#562

Let me know if you need help with those!

PR #273, opened 2016-03-07
fix(content) #185 convert protocol relative URLs to https

### LINKED ISSUE #185: Certain favicons not iconing if protocol not specified
This is a funny one, but I can't convince myself on how edge case it could potentially be.

I'm seeing a weird issue w/ certain Top Sites favicons while using the shim data (specifically "YouTube" and "Reddit"). As close as I can diagnose this, it is because they are using protocol-less versions for their icon in ./content-src/lib/fake-data.js:

- [`youtube.com => icons[0].url`](https://github.com/mozilla/activity-streams/blob/6081566c5d379c29f3e5486e886ba359c8db0733/content-src/lib/fake-data.js#L82) &mdash; "//s.ytimg.com/yts/img/yt_1200-vfl4C3T0K.png"
- [`reddit.com => icons[0].url`](https://github.com/mozilla/activity-streams/blob/6081566c5d379c29f3e5486e886ba359c8db0733/content-src/lib/fake-data.js#L755) &mdash; "//www.redditstatic.com/icon-touch.png"

These never seemed to manifest themselves when I was testing on localhost:1963, but seem to be happening when I test via resource:// or just via new-tab, so I'm guessing that a non-protocol syntax may not translate well in certain situations. I can't seem to find any of my current sites which show the same behavior, so not sure how common it is for favicons to have a non-explicit HTTP or HTTPS prefix, but not sure if we want to try prefixing with an "https:" if no protocol was specified.

---

new_tab

rlr:
r? @k88hudson @pdehaan 

Fixes #185

### DIFF
--- a/content-src/components/SiteIcon/SiteIcon.js
+++ b/content-src/components/SiteIcon/SiteIcon.js
@@ -8,13 +8,18 @@ const FAVICON_SIZE = 16;
 
 function getSiteImage(site, width) {
   const {favicon_url, icons} = site;
+  let url = null;
   if (icons && icons.length) {
-    return icons[0].url;
+    url = icons[0].url;
   } else if (favicon_url && width / FAVICON_SIZE <= MIN_RESOLUTION) {
-    return favicon_url;
-  } else {
-    return null;
+    url = favicon_url;
   }
+
+  if (url && url.startsWith("//")) {
+    url = `https:${url}`;
+  }
+
+  return url;
 }
 
 const SiteIconImage = React.createClass({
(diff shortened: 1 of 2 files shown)


pdehaan:
LGTM! Not sure if we'll eventually want to use this elsewhere and it's worth extracting into [/content-src/lib/utils.js](https://github.com/mozilla/activity-streams/blob/master/content-src/lib/utils.js) (especially since it already looks like I'm doing something similar in [`sanitizeUrl()`](https://github.com/mozilla/activity-streams/blob/603206245c82b14f459b989a30ec9cde6fa571e8/content-src/lib/utils.js#L47-L50)).

I'll let @k88hudson give final review and press the button.

k88hudson:
R+

PR #278, opened 2016-03-08
Closes #19 - Instrument add-on for performance

### LINKED ISSUE #19: Instrument add-on for performance
We should have a way of determining the impact of the add-on on performance/load-time etc.

mzhilyaev:
Fixes #19

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -13,6 +13,11 @@ const NewTabPage = React.createClass({
   onSearch(value) {
     this.props.dispatch(actions.NotifyPerformSearch(value));
   },
+  componentDidUpdate() {
+    if (this.props.isReady) {
+      this.props.dispatch(actions.NotifyTelemetry("NEWTAB_RENDER"));
+    }
+  },
   render() {
     const props = this.props;
     return (<main className="new-tab">
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -11,6 +11,7 @@ const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {TabTracker} = require("lib/TabTracker");
 const {TelemetrySender} = require("lib/TelemetrySender");
+const {PerfMeter} = require("lib/PerfMeter");
 const {AppURLHider} = require("lib/AppURLHider");
 const am = require("content-src/actions/action-manager");
 
@@ -45,6 +46,7 @@ function ActivityStreams(options = {}) {
     this._tabTracker = new TabTracker(this.appURLs, options.clientID);
   }
   this._appURLHider = new AppURLHider(this.appURLs);
+  this._perfMeter = new PerfMeter(this.appURLs);
 }
 
 ActivityStreams.prototype = {
@@ -59,6 +61,7 @@ ActivityStreams.prototype = {
     // if the function is async, the worker might not be there yet, or might have already disappeared
     try {
       worker.port.emit("addon-to-content", action);
+      this._perfMeter.log(worker.tab, action.type);
     } catch (err) {
       this.workers.delete(worker);
       Cu.reportError(err);
@@ -239,6 +242,7 @@ ActivityStreams.prototype = {
           if (msg.type === "pagehide") {
             this._removeWorker(worker);
           }
+          this._perfMeter.log(worker.tab, msg.type, msg.data);
           this.emit(msg.type, {msg, worker});
         });
       },
@@ -252,12 +256,16 @@ ActivityStreams.prototype = {
    * Adds a worker and calls callback if defined
    */
   _addWorker(worker) {
+    this._perfMeter.log(worker.tab, "WORKER_ATTACHED");
     this.workers.add(worker);
     if (this.options.onAddWorker) {
       this.options.onAddWorker();
     }
   },
 
+  /**
+   * Removes a worker and calls callback if defined
+   */
   _removeWorker(worker) {
     this.workers.delete(worker);
     if (this.options.onRemoveWorker) {
@@ -265,9 +273,6 @@ ActivityStreams.prototype = {
     }
   },
 
-  /**
-   * Removes a worker and calls callback if defined
-   */
   _setupButton() {
     this._button = ActionButton({
       id: "activity-streams-link",
@@ -297,6 +302,10 @@ ActivityStreams.prototype = {
     return this._tabTracker.tabData;
   },
 
+  get performanceData() {
+    return this._perfMeter.events;
+  },
+
   /**
    * Unload the application
    */
@@ -314,6 +323,7 @@ ActivityStreams.prototype = {
           this._telemetrySender.uninit();
         }
         this._appURLHider.uninit();
+        this._perfMeter.uninit();
     }
   }
 };
--- a/lib/PerfMeter.js
+++ b/lib/PerfMeter.js
@@ -1,0 +1,147 @@
+/* globals Services */
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/Services.jsm");
+
+const tabs = require("sdk/tabs");
+const simplePrefs = require("sdk/simple-prefs");
+
+const VALID_TELEMETRY_TAGS = new Set([
+  "TAB_READY",
+  "WORKER_ATTACHED",
+  "TOP_FRECENT_SITES_REQUEST",
+  "TOP_FRECENT_SITES_RESPONSE",
+  "RECENT_BOOKMARKS_REQUEST",
+  "RECENT_BOOKMARKS_RESPONSE",
+  "RECENT_LINKS_REQUEST",
+  "RECENT_LINKS_RESPONSE",
+  "FRECENT_LINKS_REQUEST",
+  "FRECENT_LINKS_RESPONSE",
+  "SEARCH_STATE_REQUEST",
+  "SEARCH_STATE_RESPONSE",
+  "NOTIFY_PERFORMANCE",
+]);
+
+function PerfMeter(trackableURLs) {
+  this._trackableURLs = trackableURLs;
+  this._tabs = {};
+  this.onOpen = this.onOpen.bind(this);
+  this.onReady = this.onReady.bind(this);
+  this.onClose = this.onClose.bind(this);
+  this.onPrefChange = this.onPrefChange.bind(this);
+  tabs.on("open", this.onOpen);
+  simplePrefs.on("performance.log", this.onPrefChange);
+  this._active = simplePrefs.prefs["performance.log"];
+}
+
+PerfMeter.prototype = {
+
+  get events() {
+    return this._tabs;
+  },
+
+  uninit() {
+    tabs.removeListener("open", this.onOpen);
+    simplePrefs.removeListener("performance.log", this.onPrefChange);
+    this.clearTabs();
+  },
+
+  clearTabs() {
+    // remove tab listeners
+    for (let id of Object.keys(this._tabs)) {
+      this._tabs[id].tab.removeListener("ready", this.onReady);
+      this._tabs[id].tab.removeListener("close", this.onClose);
+    }
+    this._tabs = {};
+  },
+
+  onPrefChange() {
+    this._active = simplePrefs.prefs["performance.log"];
+    Services.obs.notifyObservers(null, "performance-pref-changed", null);
+  },
+
+  isActivityStreamsURL(URL) {
+    return this._trackableURLs.indexOf(URL) !== -1;
+  },
+
+  onReady(tab) {
+    if (this.isActivityStreamsURL(tab.url)) {
+      this.log(tab, "TAB_READY");
+    } else {
+      // not an activity stream tab, get rid of it
+      delete this._tabs[tab.id];
+    }
+    tab.removeListener("ready", this.onReady);
+  },
+
+  onClose(tab) {
+    delete this._tabs[tab.id];
+  },
+
+  onOpen(tab) {
+    let item = {tag: "TAB_OPEN", start: 0};
+    this._tabs[tab.id] = {
+      tab: tab,
+      openAt: Date.now(),
+      events: [item],
+      requests: new Map(),
+      workerWasAttached: false,
+    };
+    tab.on("ready", this.onReady);
+    tab.on("close", this.onClose);
+    this.displayItem(tab.id, item);
+  },
+
+  log(tab, tag, data) {
+    if (this.isActivityStreamsURL(tab.url) && VALID_TELEMETRY_TAGS.has(tag)) {
+      let tabData = this._tabs[tab.id];
+
+      // when tab is reloaded, the worker will be re-attached to the tab and another
+      // WORKER_ATTACHED event will be sent. In which case we re-initialize tabData
+      if (tag == "WORKER_ATTACHED") {
+        if (tabData.workerWasAttached) {
+          // tab is reloaded - re-initialize it. Since reload does not generate TAB_OPEN
+          // and TAB_READY events, we introduce articficial ones starting at 0
+          tabData.events = [
+            {tag: "TAB_RELOAD", start: 0},
+            {tag: "TAB_READY", start: 0},
+          ];
+          tabData.requests = new Map();
+          tabData.openAt = Date.now();
+        }
+        tabData.workerWasAttached = true;
+      }
+
+      let item = {
+        tag: tag,
+        start: Date.now() - tabData.openAt,
+        data: data,
+      };
+
+      // handle requests/response pairs
+      if (tag.endsWith("_REQUEST")) {
+        tabData.requests.set(tag, item.start);
+      } else if (tag.endsWith("_RESPONSE")) {
+        let request = tag.replace("_RESPONSE", "_REQUEST");
+        item.delta = item.start - tabData.requests.get(request);
+      }
+
+      tabData.events.push(item);
+
+      // display the event onto console
+      this.displayItem(tab.id, item);
+
+      // check for last event
+      if (data === "NEWTAB_RENDER") {
+        Services.obs.notifyObservers(null, "performance-log-complete", JSON.stringify({tabId: tab.id, events: tabData.events}));
+      }
+    }
+  },
+
+  displayItem(tabId, item) {
+    if (this._active) {
+      console.info(`${tabId} ${item.start} ${item.tag}${item.data ? ":" + item.data : ""} ${item.delta || ""}`);
+    }
+  },
+};
+
+exports.PerfMeter = PerfMeter;
--- a/package.json
+++ b/package.json
@@ -117,5 +117,13 @@
     "help": "yamscripts help",
     "yamscripts": "yamscripts compile",
     "__": "# NOTE: THESE SCRIPTS ARE COMPILED!!! EDIT yamscripts.yml instead!!!"
-  }
+  },
+  "preferences": [
+    {
+      "name": "performance.log",
+      "title": "Enable Performance Logging",
+      "type": "bool",
+      "value": false
+    }
+  ]
 }
--- a/test/test-PerfMeter.js
+++ b/test/test-PerfMeter.js
@@ -1,0 +1,220 @@
+/* globals Services */
+
+"use strict";
+const simplePrefs = require("sdk/simple-prefs");
+
+const {before, after} = require("sdk/test/utils");
+const tabs = require("sdk/tabs");
+const {ActivityStreams} = require("lib/ActivityStreams");
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/Services.jsm");
+
+let app;
+
+function openTestTab(openUrl, notifyEvent) {
+  return new Promise(resolve => {
+    let tabData = {};
+    function onShow(tab) {
+      tabData.tab = tab;
+      if (!notifyEvent) {
+        resolve(tabData);
+      }
+    }
+
+    function onNotify(subject, topic, data) {
+      Services.obs.removeObserver(onNotify, notifyEvent);
+      tabData.notifyData = JSON.parse(data);
+      resolve(tabData);
+    }
+
+    if (notifyEvent) {
+      Services.obs.addObserver(onNotify, notifyEvent);
+    }
+
+    // open the page in timeout, to make it fire after we return
+    tabs.open({url: openUrl, onPageShow: onShow});
+  });
+}
+
+function prefSetter(value) {
+  return new Promise(resolve => {
+    function onPrefChange() {
+      Services.obs.removeObserver(onPrefChange, "performance-pref-changed");
+      resolve();
+    }
+    Services.obs.addObserver(onPrefChange, "performance-pref-changed");
+    simplePrefs.prefs["performance.log"] = value;
+  });
+}
+
+function verifyPerfEvents(eventsArray, assert) {
+  let exectedEvents = new Set([
+    "TAB_OPEN",
+    "TAB_RELOAD",
+    "WORKER_ATTACHED",
+    "TAB_READY",
+    "TOP_FRECENT_SITES_REQUEST",
+    "RECENT_LINKS_REQUEST",
+    "FRECENT_LINKS_REQUEST",
+    "RECENT_BOOKMARKS_REQUEST",
+    "SEARCH_STATE_REQUEST",
+    "NOTIFY_PERFORMANCE",
+    "TOP_FRECENT_SITES_RESPONSE",
+    "RECENT_LINKS_RESPONSE",
+    "FRECENT_LINKS_RESPONSE",
+    "RECENT_BOOKMARKS_RESPONSE",
+    "SEARCH_STATE_RESPONSE",
+    "NOTIFY_PERFORMANCE",
+  ]);
+
+  let expectedData = new Set([
+    "BASE_MOUNTED",
+    "NEWTAB_RENDER"
+  ]);
+
+  let seenEvents = new Set();
+  for (let event of eventsArray) {
+    assert.ok(exectedEvents.has(event.tag), "Received expected event " + event.tag);
+    assert.ok(!seenEvents.has(event.tag + event.data), "Event tag and data pair is unique");
+    if (event.tag === "NOTIFY_PERFORMANCE") {
+      assert.ok(expectedData.has(event.data) || event.data.startsWith("DOC_READY_STATE"),
+                "Correct data associated with content notification");
+    }
+    if (event.tag.indexOf("_RESPONSE") > -1) {
+      // all responses must have delta computed
+      assert.ok(Number.isInteger(event.delta), "Responses have time delta attached");
+    }
+    seenEvents.add(event.tag + event.data);
+  }
+  // the first start must be 0
+  assert.equal(eventsArray[0].start, 0, "First event always start at 0");
+}
+
+exports.test_PerfMeter_init = function(assert) {
+  assert.deepEqual(app.performanceData, {}, "no performance data");
+};
+
+exports.test_PerfMeter_events = function*(assert) {
+  let tabData = yield openTestTab(app.appURLs[1], "performance-log-complete");
+  let notifyData = tabData.notifyData;
+
+  let performanceData = app.performanceData;
+  assert.ok(performanceData[tabData.tab.id], "Entry with expected tab id exists");
+
+  let events = performanceData[tabData.tab.id].events;
+  // did events come as expected
+  verifyPerfEvents(events, assert);
+  // check that "performance-log-complete" notification returns same tabId
+  assert.equal(notifyData.tabId, tabData.tab.id, "Tab ids are identical");
+  // check that "performance-log-complete" notification events are included in app events
+  function toString(item) {return item.start + item.tag + item.data;}
+  let eventsSet = new Set(events.map(item => toString(item)));
+  notifyData.events.forEach(item => {
+    assert.ok(eventsSet.has(toString(item)), "Matched item " + item.tag);
+  });
+
+  // test reload
+  yield new Promise(resolve => {
+    function onNotify(subject, topic, data) {
+      Services.obs.removeObserver(onNotify, "performance-log-complete");
+      resolve();
+    }
+    Services.obs.addObserver(onNotify, "performance-log-complete");
+    tabData.tab.reload();
+  });
+
+  events = app.performanceData[tabData.tab.id].events;
+  // note that verifyPerfEvents checks for duplicate events and will fail
+  // if the tab structure doesn't re-initialize after reload
+  verifyPerfEvents(events, assert);
+  // verify the first three expected events on reload
+  assert.ok(events[0].tag === "TAB_RELOAD" && events[0].start == 0, "Expected TAB_RELOAD");
+  assert.ok(events[1].tag === "TAB_READY" && events[0].start == 0, "Expected TAB_READY");
+  assert.ok(events[2].tag === "WORKER_ATTACHED" && events[0].start == 0, "Expected WORKER_ATTACHED");
+
+  tabData.tab.close();
+};
+
+exports.test_PerfMeter_pref = function*(assert) {
+  // We are always tracking events regadless of performance.log settings
+  // However, if performance.log is on, we also logging the events to console
+  let tabData = yield openTestTab(app.appURLs[1], "performance-log-complete");
+
+  let events = tabData.notifyData.events;
+  assert.equal(events[events.length - 1].tag, "NOTIFY_PERFORMANCE", "Expected last event");
+  tabData.tab.close();
+
+  assert.equal(simplePrefs.prefs["performance.log"], true, "performance.log is on");
+  assert.ok(app._perfMeter._active, "PerfMeter active flag is set");
+
+  yield prefSetter(false);
+  assert.equal(simplePrefs.prefs["performance.log"], false, "performance.log is off");
+  assert.ok(!app._perfMeter._active, "PerfMeter active flag is not set");
+
+  // fire a new tab and ensure perfMeter still collects events
+  tabData = yield openTestTab(app.appURLs[1], "performance-log-complete");
+
+  events = tabData.notifyData.events;
+  assert.equal(events[events.length - 1].tag, "NOTIFY_PERFORMANCE", "Expected last event");
+  tabData.tab.close();
+
+  // set pref back on, and start getting events again
+  yield prefSetter(true);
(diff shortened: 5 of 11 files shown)


lib/TelemetryTracker.js
```diff
@@ -0,0 +1,120 @@
...
+  },
+
+  log: function(tab, tag, data) {
+    if (this._active && this.isActivityStreamsURL(tab.url) && VALID_TELEMETRY_TAGS.has(tag)) {
+      if (!this._events) {
+        this._events = [];
+        this._events.push({tag: "TAB_OPEN", start: this._lastTabOpen});
+      }
+      this._events.push({tag: tag, start: Date.now(), data: data});
+      if (data == "NEWTAB_RENDER") {
```
pdehaan:
```
91:16  warning  Expected '===' and instead saw '=='  eqeqeq
```

lib/TelemetryTracker.js
```diff
@@ -0,0 +1,120 @@
...
+      // subtract first start time from all subsequent times
+      item.start -= firstStart;
+      if (item.tag.endsWith("_RESPONSE")) {
+        // if response, record the time delta
+        let request = item.tag.replace("_RESPONSE", "_REQUEST");
+        item.delta = item.start - tagTimes.get(request);
+      }
+      // record start time for the event
+      tagTimes.set(item.tag, item.start);
+      console.log(item.start + " " + item.tag + (item.data ? ":" + item.data : "") + " " + (item.delta || ""));
```
pdehaan: Does this `console.log()` print to the user's devTools console, or is this on the server?

ESLint is displaying the following warning:
```
111:7   warning  Unexpected console statement         no-console
```

test/test-TelemetryTracker.js
```diff
@@ -0,0 +1,149 @@
...
+      Services.obs.addObserver(observer, notifyEvent);
+    }
+    tabs.on("pageshow", onShow);
+  });
+  tabs.open(ACTIVITY_STREAMS_URL);
+  return tabData;
+}
+
+exports.test_TelemetryTracker_init = function(assert) {
+  assert.ok(app.telemetryEvents == null, "no telemetry events");
```
pdehaan:
```
48:33  warning  Expected '===' and instead saw '=='  eqeqeq
```

lib/TelemetryTracker.js
```diff
@@ -0,0 +1,120 @@
...
+  "RECENT_LINKS_REQUEST",
+  "RECENT_LINKS_RESPONSE",
+  "FRECENT_LINKS_REQUEST",
+  "FRECENT_LINKS_RESPONSE",
+  "SEARCH_STATE_REQUEST",
+  "SEARCH_STATE_RESPONSE",
+  "NOTIFY_TELEMETRY",
+]);
+
+function TelemetryTracker(trackableURLs) {
```
oyiptong: A better name is `PerformanceTracker`, some data will be sent to our telemetry servers some not

test/test-TelemetryTracker.js
```diff
@@ -0,0 +1,149 @@
...
+
+  let tabData = openTestTab("telemetry-complete");
+  yield tabData.showPromise;
+
+  let events = app.telemetryEvents;
+  for (let i = 0; i < exectedEvents.length; i++) {
+    assert.equal(events[i].tag, exectedEvents[i], "Received expected event " + exectedEvents[i]);
+    if (events[i].tag == "NOTIFY_TELEMETRY") {
+      // must have base-mount or newtab render data bit attached
+      assert.ok(events[i].data == "BASE_MOUNTED" || events[i].data == "NEWTAB_RENDER", "Correct data associated with content notification");
```
pdehaan:
```
  74:23  warning  Expected '===' and instead saw '=='  eqeqeq
  76:32  warning  Expected '===' and instead saw '=='  eqeqeq
  76:68  warning  Expected '===' and instead saw '=='  eqeqeq
```

mzhilyaev:
Following the debug session with @oyiptong , I am logging all the things that went wrong with tracker and things to fix:

-  Events should be logged as they appear
-  For all events but requests, there should be a time delta between it and the previous one
-  It's OK to print out all tab requests upon tab close
-  it appears that multiple NEWTAB_RENDER may be fired for same tab, hence printing report on the first one may not always work
-  if a user opens a bunch of tabs, events from different tabs get mixed, hence we need to keep track which event comes from which tab, and log tabid next to the event.
- TelemetryTracker should be renamed to PerfTracker
- console.log does not work right away on existing profiles - the need to figure why and how to set logging prefs to see it
- WORKER_ATTACH event does not seem to be added to the array, but is logged
- EXTRA events:
   -  emebdly calls if not too hard to do
   -  document. state (like ready state, dom loaded and the like).  Look at content-bridge to implement custom events there, as it gets loaded first before Redux

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -6,12 +6,19 @@ const {GroupedActivityFeed} = require("components/ActivityFeed/ActivityFeed");
...
+let REPORT_RENDER_EVENT = true;
 
 const NewTabPage = React.createClass({
   // TODO: Replace with real search api via addon
   onSearch(value) {
     this.props.dispatch(actions.NotifyPerformSearch(value));
   },
+  componentWillUpdate(nextProps, nextState) {
+    if (nextProps.isReady && REPORT_RENDER_EVENT) {
+      this.props.dispatch(actions.NotifyTelemetry("NEWTAB_RENDER"));
```
oyiptong: When will this trigger multiple times?

lib/TelemetryTracker.js
oyiptong: This should print to the terminal output (STDOUT if running jpm, otherwise devtools).

The eslint rule for console statements should be removed; debug logs are a useful tool.

lib/main.js
```diff
@@ -17,7 +17,7 @@ Object.assign(exports, {
     // options.loadReason can be install/enable/startup/upgrade/downgrade
     PlacesProvider.links.init();
     SearchProvider.search.init();
-    options.telemetry = false;
+    options.telemetry = true;
```
oyiptong: the default should be false for now

oyiptong: we need an additional pref for performance tracking. maybe we can call it 'profiling mode'?

package.json
```diff
@@ -117,5 +117,13 @@
     "help": "yamscripts help",
     "yamscripts": "yamscripts compile",
     "__": "# NOTE: THESE SCRIPTS ARE COMPILED!!! EDIT yamscripts.yml instead!!!"
-  }
+  },
+  "preferences": [
+    {
+      "name": "telemetryLog",
+      "title": "Enable Telemetry",
```
oyiptong: Performance Profiling

oyiptong:
Also, the performance profiling needs to account for the 'reload' use-case

lib/ActivityStreams.js
```diff
@@ -59,6 +61,7 @@ ActivityStreams.prototype = {
     // if the function is async, the worker might not be there yet, or might have already disappeared
     try {
       worker.port.emit("addon-to-content", action);
+      this._logPerformance(action.type, worker);
```
oyiptong: Why do we need the indirection? Can we call `this._perfMeter.log(worker.tab, action.type)` directly?

lib/ActivityStreams.js
```diff
@@ -239,6 +242,7 @@ ActivityStreams.prototype = {
           if (msg.type === "pagehide") {
             this._removeWorker(worker);
           }
+          this._logPerformance(msg.type, worker, msg.data);
```
oyiptong: same as above

lib/ActivityStreams.js
```diff
@@ -252,12 +256,16 @@ ActivityStreams.prototype = {
    * Adds a worker and calls callback if defined
    */
   _addWorker(worker) {
+    this._logPerformance("WORKER_ATTACHED", worker);
```
oyiptong: same as above

lib/ActivityStreams.js
```diff
@@ -266,8 +274,12 @@ ActivityStreams.prototype = {
   },
 
   /**
-   * Removes a worker and calls callback if defined
+   * Logs telemetry event
    */
+  _logPerformance(tag, worker, data) {
```
oyiptong: Seems like unnecessary indirection

lib/PerfMeter.js
```diff
@@ -0,0 +1,143 @@
...
+  "FRECENT_LINKS_REQUEST",
+  "FRECENT_LINKS_RESPONSE",
+  "SEARCH_STATE_REQUEST",
+  "SEARCH_STATE_RESPONSE",
+  "NOTIFY_PERFORMANCE",
+]);
+
+function PerfMeter(trackableURLs) {
+  this._trackableURLs = trackableURLs;
+  this.init();
```
oyiptong: the contents of init aren't too big, why not put those here?

lib/ActivityStreams.js
```diff
@@ -45,6 +46,7 @@ function ActivityStreams(options = {}) {
     this._tabTracker = new TabTracker(this.appURLs);
   }
   this._appURLHider = new AppURLHider(this.appURLs);
+  this._perfMeter = new PerfMeter(this.appURLs);
```
oyiptong: only enable performance tracking if a performance profiling pref is turned on (or option passed)

oyiptong: Nevermind, I see that you only attach the tab listeners based on a pref below

lib/PerfMeter.js
```diff
@@ -0,0 +1,143 @@
...
+
+PerfMeter.prototype = {
+
+  _tabs: {},
+
+  get events() {
+    return this._tabs;
+  },
+
+  init: function() {
```
oyiptong: how about using the new way to declare functions?
```javascript
init() {
  ...
}
```

This is consistent with the rest of the code

lib/PerfMeter.js
```diff
@@ -0,0 +1,143 @@
...
+  },
+
+  onReady: function(tab) {
+    if (this.isActivityStreamsURL(tab.url)) {
+      this.log(tab, "TAB_READY");
+    } else {
+      // not an activity stream tab, get rid of it
+      delete this._tabs[tab.id];
+    }
+    tab.removeListener("ready", this.onReady);
```
oyiptong: Does this do what we want?

If two tabs trigger `onReady`, what will happen? Do we want the listener to be removed?
There might be multiple tabs triggering this even.

lib/PerfMeter.js
```diff
@@ -0,0 +1,143 @@
...
+
+  onOpen: function(tab) {
+    let item = {tag: "TAB_OPEN", start: 0};
+    this._tabs[tab.id] = {
+      tab: tab,
+      openAt: Date.now(),
+      events: [item],
+      requests: new Map(),
+    };
+    tab.on("ready", this.onReady);
```
oyiptong: this will trigger multiple new listeners to the `ready` event in case of multiple tabs open. Is that what we want?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+const {ActivityStreams} = require("lib/ActivityStreams");
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/Services.jsm");
+
+let ACTIVITY_STREAMS_URL;
+let app;
+
+function openTestTab(notifyEvent) {
+  let tabData = {};
+  tabData.showPromise = new Promise(resolve => {
```
oyiptong: why not return the promise itself?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+    function onShow(tab) {
+      tabData.tab = tab;
+      tabs.removeListener("pageshow", onShow);
+      if (!notifyEvent) {
+        setTimeout(function() {
```
oyiptong: why `setTimeout` here?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+          resolve();
```
oyiptong: why not resolve with the tab here?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+        });
+      }
+    }
+
+    let observer = {
+      observe: function(subject, topic, data) {
+        Services.obs.removeObserver(this, notifyEvent);
+        setTimeout(function() {
```
oyiptong: why timeout here?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+        });
+      }
+    };
+
+    if (notifyEvent) {
+      Services.obs.addObserver(observer, notifyEvent);
+    }
+    tabs.on("pageshow", onShow);
+  });
+  tabs.open(ACTIVITY_STREAMS_URL);
```
oyiptong: pass the URL to open as a function parameter

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+const simplePrefs = require("sdk/simple-prefs");
+
+const {before, after} = require("sdk/test/utils");
+const tabs = require("sdk/tabs");
+const {setTimeout} = require("sdk/timers");
+const {ActivityStreams} = require("lib/ActivityStreams");
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/Services.jsm");
+
+let ACTIVITY_STREAMS_URL;
```
oyiptong: global state: bad

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+      tabs.removeListener("pageshow", onShow);
+      if (!notifyEvent) {
+        setTimeout(function() {
+          resolve();
+        });
+      }
+    }
+
+    let observer = {
+      observe: function(subject, topic, data) {
```
oyiptong:
```javascript
observe (subject, topic, data) {
  ...
}
```

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+  ]);
+
+  let tabData = openTestTab("performance-log-complete");
+  let notifyData = yield tabData.showPromise;
+
+  let performanceData = app.performanceData;
+  assert.ok(performanceData[tabData.tab.id], "Entry with expected tab id exists");
+
+  let events = performanceData[tabData.tab.id].events;
+  for (let i = 0; i < exectedEvents.length; i++) {
```
oyiptong:
```javascript
for (let event of events) {
  ...
}
```

lib/PerfMeter.js
```diff
@@ -0,0 +1,143 @@
...
+    if (simplePrefs.prefs.performanceLog) {
+      tabs.removeListener("open", this.onOpen);
+    }
+    this.clearTabs();
+    simplePrefs.removeListener("performanceLog", this.onPrefChange);
+  },
+
+  clearTabs: function() {
+    // remove tab listeners
+    Object.keys(this._tabs).forEach(id => {
```
oyiptong:
```javascript
for (let id of this._tabs) {
  ...
}
```

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+    assert.ok(exectedEvents.has(events[i].tag), "Received expected event " + exectedEvents[i]);
```
oyiptong: what is this testing? isn't this a tautology?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+  // the first start must be 0
+  assert.equal(events[0].start, 0, "First event always start at 0");
+
+  // check that "performance-log-complete" notification returns same tabId
+  assert.equal(notifyData.tabId, tabData.tab.id, "Tab ids are identical");
+  // check that "performance-log-complete" notification events are included in application events
+  let eventsSet = new Set(events.map(item => JSON.stringify(item)));
+  notifyData.events.forEach(item => {
```
oyiptong:
```javascript
for (let item of notifyData.events) {
  ...
}
```

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+  let events = notifyData.events;
+  assert.equal(events[events.length - 1].tag, "NOTIFY_PERFORMANCE", "Expected last event");
+  tabData.tab.close();
+
+  function prefSetter(value) {
+    return new Promise(resolve => {
+      let observer = {
+        observe: function() {
+          Services.obs.removeObserver(this, "performance-pref-changed");
+          setTimeout(function() {
```
oyiptong: why timeout?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+        }
+      };
+      Services.obs.addObserver(observer, "performance-pref-changed");
+      simplePrefs.prefs.performanceLog = value;
+    });
+  }
+
+  assert.equal(simplePrefs.prefs.performanceLog, true, "performanceLog is on");
+
+  yield prefSetter(false);
```
oyiptong: the function is mislabeled. You're testing that setting a pref causes an observer notification to fire, correct?
It might be more explicit to declare have a promise that waits for an observer notification to occur, to set the pref, then to yield on the promise

test/test-PerfMeter.js
```diff
@@ -0,0 +1,162 @@
...
+const tabs = require("sdk/tabs");
+const {setTimeout} = require("sdk/timers");
+const {ActivityStreams} = require("lib/ActivityStreams");
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/Services.jsm");
+
+let ACTIVITY_STREAMS_URL;
+let app;
+
+function openTestTab(notifyEvent) {
```
oyiptong: need another test where one opens multiple tabs

oyiptong:
There are style issues, and code correctness issues. Please fix

lib/ActivityStreams.js
mzhilyaev: I would like to keep this indirection as we may need some other logic executed before we sent an event to the perfMeter.  For example, I am customary using this API to place debug output and see which logs and when are generated by the application rather then the PerfMeter.

lib/PerfMeter.js
mzhilyaev: This style of function definitions is only used inside ActivityStreams.js, other modules like TabTracker and PlacesProvider use the old style of function definitions.  I would suggest either accepting the existence of two styles of function definitions (in which case this should not be an issue).  Or doing a clean up across all modules to use the new style.  There are no guidelines on function styling as of now, so I wonder if we need to be adamant about it in this particular case?  Not a problem for me change, but perhaps we should be consistent.

lib/PerfMeter.js
mzhilyaev: Yes, it does.  Note that the tab.removeListener(...) removes the listener attached to the tab passed in.  The "ready" listeners are attached to individual tabs not the tabs object.

lib/PerfMeter.js
mzhilyaev: Again, note that the "ready" listener is attached to the "tab" not "tabs"

test/test-PerfMeter.js
mzhilyaev: Because I need the tab object to close it, and you never know which tab was closed when in the situation of multiple tabs being open simultaneously

test/test-PerfMeter.js
mzhilyaev: You want all other observers of "pageshow" event to handle that event before you resolve the promise.  Which is why you stick resolve inside the timeout, so the promise is resolved after everyone else finish with  "pageshow"  event.

test/test-PerfMeter.js
mzhilyaev: Not sure i understand the question, but if you wonder why the promise is resolved inside setTimeout, see the previous comment

test/test-PerfMeter.js
mzhilyaev: see comments above

test/test-PerfMeter.js
mzhilyaev: I frankly don't see it as a major issue, given the test context. 
We have used the exact same trick in TabTracker  https://github.com/mozilla/activity-streams/blob/master/test/test-TabTracker.js#L15, and I think it's acceptable as it saves a bit of time not to pass the exact same url around.   It does make the test code more readable to me, but I will be happy to change if you disagree.

test/test-PerfMeter.js
mzhilyaev: This ensures that the events recorded by perfMeter are what we expect them to be.

test/test-PerfMeter.js
mzhilyaev: Yes, in this case the timeout could be removed

test/test-PerfMeter.js
mzhilyaev: I am not really testing the pref being changed (although i do check it).  This function is called prefSetter because I need a promise, so I can wait until PerfMeter switches into inactive mode upon the pref change.  Does this make sense?

test/test-PerfMeter.js
mzhilyaev: Ok, will do.

content-src/components/NewTabPage/NewTabPage.js
oyiptong: Asking the question again. I'm curious. When would NEWTAB_RENDER fire multiple times?
You set a global at line 11 and disable it at line 21

lib/ActivityStreams.js
oyiptong: >  For example, I am customary using this API to place debug output and see which logs and when are generated by the application rather then the PerfMeter.

I would say that if you want to put dump statements, they should not affect how the code is written.
For your own development, you could put dump statements inside of PerfMeter's log function.

This adds unnecessary indirection which complicates the code for no obvious gain. We now have extra functions that serve no apparent function in the running of the program.

> Another reason is a potential change to the perfMeter.log() signature, for example we decide to send in yet another parameter to the perfMeter (_appURLs could be a candidate) - we only need to change one call site instead of all.

Are we doing this right now? If not, that sounds like over engineering

lib/PerfMeter.js
oyiptong: PlacesProvider is an exception, because we ported the code over.

For all new code, I'd like us to be consistent, as you say. We should perhaps change TabTracker.

lib/PerfMeter.js
oyiptong: Ah, excellent.

lib/PerfMeter.js
oyiptong: Got it, my mistake

test/test-PerfMeter.js
oyiptong: The code could be made simpler and easier to read in the tests by just passing a parameter instead of using a global state.

test/test-PerfMeter.js
oyiptong: you can return the tab in the `resolve` function. That would make the code simpler

test/test-PerfMeter.js
oyiptong: You comment above mentioned the need to return the tab. You can return the `tab` object in the resolve function.

test/test-PerfMeter.js
oyiptong: Got it

test/test-PerfMeter.js
oyiptong: which other observers? you're removing your specific onShow listener, you're not removing all other listeners

test/test-PerfMeter.js
oyiptong: see comment above

test/test-PerfMeter.js
oyiptong: you don't need the timeout

test/test-PerfMeter.js
oyiptong: argh, sorry wrong comment. I think resolve will not return immediately and I can find out. If you absolutely want to run in the next tick, you can always do `promise.then`, no need for timeouts.

test/test-PerfMeter.js
mzhilyaev: I want  perfMeter.onPrefChange to explicitly exit before the promise is resolved, which explains the need to for time out.  Here's the relevant lines of code;

PerfMeter.js:
```
  onPrefChange: function() {
    ....
    Services.obs.notifyObservers(null, "performance-pref-changed", null);
  },
``` 
Test code
```
      observe () {
        Services.obs.removeObserver(this, "performance-pref-changed");
        setTimeout(function() {
          resolve();
        });
      }
```

lib/PerfMeter.js
pdehaan: FWIW, I believe this is ESLint's [`object-shorthand`](http://eslint.org/docs/rules/object-shorthand) rule.

If I add `"object-shorthand": 1,` to the .eslintrc file, I currently get about 24 warnings [in **master**, I haven't tried against this branch specifically yet].

test/test-PerfMeter.js
```diff
@@ -0,0 +1,196 @@
...
+
+function prefSetter(value) {
+  return new Promise(resolve => {
+    let observer = {
+      observe() {
+        Services.obs.removeObserver(this, "performance-pref-changed");
+        resolve();
+      }
+    };
+    Services.obs.addObserver(observer, "performance-pref-changed");
```
oyiptong: FYI you could just use a function there:

```javascript
let someFunc = (subject, topic, data) => {
  Services.obs.removeObserver(someFunc, "some-notif");
  ...
};
Services.obs.addObserver(someFunc, "some-notif");
```

test/test-PerfMeter.js
```diff
@@ -0,0 +1,196 @@
...
+    "RECENT_BOOKMARKS_RESPONSE",
+    "SEARCH_STATE_RESPONSE",
+    "NOTIFY_PERFORMANCE",
+  ]);
+
+  let seenEvents = new Set();
+  for (let event of eventsArray) {
+    assert.ok(exectedEvents.has(event.tag), "Received expected event " + event.tag);
+    assert.ok(!seenEvents.has(event.tag + event.data), "Event tag and data pair is unique");
+    if (event.tag === "NOTIFY_PERFORMANCE") {
```
oyiptong: why not add `NOTIFY_PERFORMANCE`, `BASE_MOUNTED` and `NEWTAB_RENDER` in `expectedEvents?

test/test-PerfMeter.js
```diff
@@ -0,0 +1,196 @@
...
+
+exports.test_PerfMeter_tab_hygiene = function*(assert) {
+  // open two activity streams tabs and two example.com tabs
+  let as_tab1 = yield openTestTab(app.appURLs[1], "performance-log-complete");
+  let as_tab2 = yield openTestTab(app.appURLs[1], "performance-log-complete");
+  let example_tab1 = yield openTestTab("http://example.com");
+  let example_tab2 = yield openTestTab("http://example.com");
+
+  // check notifyData for both tabs for sanity
+  assert.equal(as_tab1.notifyData.tabId, as_tab1.tab.id, "Macthed tab ids");
```
oyiptong: matched

test/test-PerfMeter.js
```diff
@@ -0,0 +1,196 @@
...
+  assert.equal(as_tab2.notifyData.tabId, as_tab2.tab.id, "Macthed tab ids");
```
oyiptong: ditto

package.json
```diff
@@ -118,5 +118,13 @@
...
     "yamscripts": "yamscripts compile",
     "__": "# NOTE: THESE SCRIPTS ARE COMPILED!!! EDIT yamscripts.yml instead!!!"
-  }
+  },
+  "preferences": [
+    {
+      "name": "performanceLog",
+      "title": "Enable Performance Logging",
+      "type": "bool",
+      "value": true
```
oyiptong: we want the default to be `false`. Enable in test mode in the `before` statement.

test/test-PerfMeter.js
oyiptong: Just an observation, btw. (pun intended ;-))

test/test-PerfMeter.js
```diff
@@ -0,0 +1,196 @@
...
+
+  let events = performanceData[tabData.tab.id].events;
+  // did events come as expected
+  verifyPerfEvents(events, assert);
+  // check that "performance-log-complete" notification returns same tabId
+  assert.equal(notifyData.tabId, tabData.tab.id, "Tab ids are identical");
+  // check that "performance-log-complete" notification events are included in application events
+  let eventsSet = new Set(events.map(item => JSON.stringify(item)));
+  notifyData.events.forEach(item => {
+    assert.ok(eventsSet.has(JSON.stringify(item)), "Macthed item " + item.tag);
```
oyiptong: Matched item

oyiptong: Also, why are you comparing JSON stringified items? Unsure the order is guaranteed.

oyiptong:
an additional note so we don't forget: we need to find out why the performance logging doesn't work with big profiles

content-src/components/NewTabPage/NewTabPage.js
mzhilyaev: @k88hudson  could you comment on @oyiptong question.  Here's the relevant code snippet in NewTabPage.js:
```
  componentWillUpdate(nextProps, nextState) {
    if (nextProps.isReady && REPORT_RENDER_EVENT) {
      this.props.dispatch(actions.NotifyTelemetry("NEWTAB_RENDER"));
      REPORT_RENDER_EVENT = false;
    }
```
I vaguely recall that the reason was not to send another NEWTAB_RENDER events to perfMeter if the page is re-rendered due to a history event.  Could you verify that this theory is correct.

k88hudson: The `componentDidUpdate` function fires every time the component's props or state changes, so it will fire any time the component changes after it is initially ready, such as  when a user deletes an item.

oyiptong: When is the page ready?

Does the initial empty page creation cause the event to fire?

Does the page populated with the original query data possibly make this fire as well?

mzhilyaev: So, there's something funny going on with these notifications.  I tested at a bit by placing window.confirm() right after the notification is sent, and sure enough - only half of the page rendered.  Recent history appears after I click confirm button. And yes, we are going through this function multiple times on first render, hence the question why?  Will bug @k88hudson on Monday

oyiptong: Can we try just printing out all the events?

mzhilyaev:
@oyiptong  - fixed logging issue and pushed, so you can use the branch to meter performance.  Not yet ready for your review - there are few more problems you flagged that i need to correct.

mzhilyaev:
all done

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -13,6 +13,11 @@ const NewTabPage = React.createClass({
   onSearch(value) {
     this.props.dispatch(actions.NotifyPerformSearch(value));
   },
+  componentWillUpdate(nextProps, nextState) {
+    if (nextProps.isReady) {
+      this.props.dispatch(actions.NotifyTelemetry("NEWTAB_RENDER"));
```
oyiptong: We're now not checking the state. What was the result of that conversation?

oyiptong:
@mzhilyaev, there are still a couple of issues:
* on reload, counters are not being reset
* there are some typos

I see you're now setting the default log level to info. What is it set at for "production use"? We should leave the default and only set it to log at `info` level when performance profiling is turned on.

I see you're no longer setting the global to send `NEWTAB_RENDER`. What was the behavior and what caused you to remove it?

package.json
```diff
@@ -118,5 +118,19 @@
     "help": "yamscripts help",
     "yamscripts": "yamscripts compile",
     "__": "# NOTE: THESE SCRIPTS ARE COMPILED!!! EDIT yamscripts.yml instead!!!"
-  }
+  },
+  "preferences": [
+    {
+      "name": "performanceLog",
```
oyiptong: nit: performance.log

content-src/components/NewTabPage/NewTabPage.js
mzhilyaev: changing to componentDidUpdate() , this will send the event when newtab page is ready

package.json
mzhilyaev: I would like to keep the old name as it allows for 
```
simplePrefs.prefs.performanceLog
```
As oppose to
```
simplePrefs.prefs["performance.log"]
```
Not critical, can fix if reviewer insists

oyiptong: I'd say to change it to `performance.log` to remain in style with what we do in gecko. We will need to change it when we do land there anyway.

mzhilyaev:
Responding to @oyiptong  comment https://github.com/mozilla/activity-streams/pull/278#issuecomment-195660476

- changed componentWillUpdate() to componentDidUpdate() in commit 5b9522a8b05ca497b39705ab7661af1c93280b8f, to send notification when page is rendered
- the global state is not needed per manual tests: we do go through that function multiple times, but only once when this.props.isReady is set.

As for sdk.console.logLevel = info, this is left at the discretion of the addon developer.  There's no production use default, meaning no logging will be seen at all fron an extension, unless the extension sets its own logging level.  Since we need to log, I think we should set own own log level at info, and do console.info()  when needed.   Relevant documentation can be found here: https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/console#Setting_the_Logging_Level

oyiptong:
@mzhilyaev logging is of no use to a production user. How about we do not set it in `packages.json`, but rather, set it in `dev-prefs.json` and change npm scripts to use this configuration file?

mzhilyaev:
@oyiptong we are using extension-specific logLevel setting exactly as this pref is intended to be used, in the exact use-case this pref is intended to be used.  Why are we so hung on it?  We want to log info message, and we do so? What's the problem of having extension log level be set at info?  A production user has no use for performance.log either - should we remove it from prefs as well?  
If you absolutely do not want to have prefs inside package.json, then ok, let's move all the prefs into dev-prefs.json, but in my humble opinion this will be a complete overkill.  I am really reluctant to do so.

oyiptong:
Ah, perhaps I miscommunicated.

My comment attempted to produce this situation:

* A production user doesn't see any log by default
* A development user can see logs but not performance logs by default
* A development user sees the logs if they turn on `performance.log`

Usually, applications run on a log level `error` in production, which I assume is the default if `sdk.console.level` is not set.

A developer doesn't need to see the performance log all the time. (though we probably should keep track of it through automated means).

What do you think of the above three scenarios?

oyiptong:
r+

Issue #221, opened 2016-02-29
Delete UI MVP


k88hudson:
This is the MVP spec for April first launch (discussed with @bryanbell):

#### Functionality

Delete buttons should **delete the individual history entry** (not bookmarks, if they exist).

#### Spotlight UI
![image](https://cloud.githubusercontent.com/assets/1455535/13404507/dfe3a520-dee7-11e5-8204-4a5ba624be49.png)

#### Activity Feed UI
* comp (show on hover, delay showing menu by `250ms`) ![image](https://cloud.githubusercontent.com/assets/1455535/13404330/248e24da-dee7-11e5-853e-aaae0b07d153.png)
*  Share button could just use a [twitter intent](https://dev.twitter.com/web/tweet-button/web-intent) for now
* `...` could be an alert saying "we're working on a better/blocking delete solution"

Note that this will all be expanded via https://github.com/mozilla/activity-streams/issues/105 post-MVP.

rlr:
landed. :rocket:

PR #280, opened 2016-03-08
feat(content) #221 Delete UI

rlr:
r? @k88hudson @pdehaan @anybodyelse


rlr:
just added a fix for the addon side ^^

content-src/components/ActivityFeed/ActivityFeed.scss
```diff
@@ -9,8 +9,12 @@
 .feed-item {
   display: flex;
   padding-left: $feed-gutter-h + $feed-circle-size;
-  padding-bottom: $feed-gutter-v;
+  padding-right: $feed-gutter-h;
+  padding-bottom: $feed-gutter-v / 2;
+  padding-top: $feed-gutter-v / 2;
   position: relative;
+  border: 1px solid $white;
```
pdehaan: I think the `$white` may be causing some mild UI issues here. You can see a slight gap in the vertical bar. Not sure if we can do a transparent border instead. Will investigate further.

pdehaan:
This is pretty awesome!

So far it looks pretty good from a UI/UX standpoint, although a few minor CSS issues we may need to fiddle with (in a future PR).

new_tab_and_slack

1. You can see the bookmark star leaks outside of the hover border.
2. There is a slight gap in the vertical line in the activity stream timeline.
3. The "Delete", "Share", and "..." buttonbar slightly leaks outside the border.

content-src/components/ActivityFeed/ActivityFeed.scss
```diff
@@ -64,6 +68,58 @@
...
+    opacity: 0;
+  }
+
+  .action-item {
+    cursor: pointer;
+    width: 40px;
+    height: 32px;
+    background-position: center center;
+    background-repeat: no-repeat;
+    border-right: solid 1px $light-grey;
```
k88hudson: These should all be `#C6C6C6`, so we'll need to add a specific color for that

content-src/components/ActivityFeed/ActivityFeed.scss
```diff
@@ -64,6 +68,58 @@
     margin-top: -($feed-gutter-v / 2 + 18 / 2);
     cursor: pointer;
   }
+
+  .action-items-container {
+    background: $white;
```
k88hudson: The spec gives us `#FBFCFC` for this although i think `$bg-grey` is close enough

content-src/components/ActivityFeed/ActivityFeed.scss
k88hudson: Actually scratch that, use `$faintest-black` so that the line on the side can fade through

k88hudson:
Right now there is a big of a weird alignment issue on the side, you need to move everything in a little so the hover state doesn't look squished
image

content-src/components/ActivityFeed/ActivityFeed.scss
k88hudson: The correct thing to do here is use `transparent`, however we're still going to have an issue because there will be a gap appearing in between elements; 

The quick way to fix this is to set the position of the `::before` element to `top: -1px` `bottom: `-1px`, or you could create an inner element that has a border, have the `::before` stuff outside that element.

rlr:
Feedback/fixes applied ^^. Much better I think. r? @k88hudson

k88hudson:
R+!

PR #292, opened 2016-03-10
feat(addon): #291 Setup ping for tab tracking telemetry.

### LINKED ISSUE #291: Perform XHR ping with telemetry

emtwo:
Fixes #291

### DIFF
--- a/lib/TelemetrySender.js
+++ b/lib/TelemetrySender.js
@@ -2,16 +2,29 @@
 
 const {Cu} = require("chrome");
 Cu.import("resource://gre/modules/Services.jsm");
+Cu.importGlobalProperties(["fetch"]);
 
-function TelemetrySender() {
+function TelemetrySender(pingEndpoint) {
   Services.obs.addObserver(this, "tab-session-complete");
+  this._pingEndpoint = pingEndpoint;
 }
 
 TelemetrySender.prototype = {
-  observe: function(subject, topic) {
+  observe: function(subject, topic, data) {
     if (topic === "tab-session-complete") {
-      // TODO: send ping to server.
+      this._sendPing(data);
     }
+  },
+
+  _sendPing: function(data) {
+    fetch(this._pingEndpoint, {method: "POST", body: data}).then(function(response) {
+      if (!response.ok) {
+        Cu.reportError(`Ping failure with response code: ${response.status}`);
+      }
+    })
+    .catch(function(e) {
+      Cu.reportError(`Ping failure with error code: ${e.message}`);
+    });
   },
 
   uninit: function() {
--- a/lib/main.js
+++ b/lib/main.js
@@ -8,6 +8,8 @@ const {Cu} = require("chrome");
 Cu.import("resource://gre/modules/ClientID.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
 
+const PING_ENDPOINT = "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream";
+
 let app = null;
 
 Object.assign(exports, {
@@ -20,6 +22,7 @@ Object.assign(exports, {
 
     Task.spawn(function*() {
       options.clientID = yield ClientID.getClientID();
+      options.pingEndpoint = PING_ENDPOINT;
       app = new ActivityStreams(options);
     }.bind(this));
   },
--- a/lib/main.js
+++ b/lib/main.js
@@ -3,12 +3,11 @@
 const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {ActivityStreams} = require("lib/ActivityStreams");
+const simplePrefs = require("sdk/simple-prefs");
 const {Cu} = require("chrome");
 
 Cu.import("resource://gre/modules/ClientID.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
-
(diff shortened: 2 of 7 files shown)


lib/TelemetrySender.js
```diff
@@ -1,19 +1,48 @@
 /* globals Services */
 
-const {Cu} = require("chrome");
+const {Cu, Cc, Ci} = require("chrome");
+
 Cu.import("resource://gre/modules/Services.jsm");
+Cu.importGlobalProperties(["XMLHttpRequest"]);
```
oyiptong: how about we use `fetch` instead?

oyiptong: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

lib/TelemetrySender.js
```diff
@@ -1,19 +1,48 @@
...
+
+function TelemetrySender(pingPref, pingEndpoint) {
+  this._pingPref = pingPref;
 
-function TelemetrySender() {
   Services.obs.addObserver(this, "tab-session-complete");
+  this._prefs = Cc["@mozilla.org/preferences-service;1"]
```
oyiptong: `Preferences.jsm` will be easier to use: https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/Preferences.jsm

oyiptong: an example in action: https://dxr.mozilla.org/mozilla-central/source/browser/components/nsBrowserGlue.js#408

lib/TelemetrySender.js
```diff
@@ -1,19 +1,48 @@
...
+  observe: function(subject, topic, data) {
     if (topic === "tab-session-complete") {
-      // TODO: send ping to server.
+      this._sendPing(data);
     }
   },
 
+  _sendPing: function(data) {
+    // Create a new XMLHttpRequest that is anonymous, i.e., doesn't send cookies
+    let ping = new XMLHttpRequest({mozAnon: true});
```
oyiptong: fetch will make this code much simpler

lib/main.js
```diff
@@ -8,6 +8,9 @@ const {Cu} = require("chrome");
 Cu.import("resource://gre/modules/ClientID.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
 
+const PING_PREF = "browser.activity-streams.ping";
```
oyiptong: Why have these both as constants and as prefs? We should have one or the other, not both.
I'm in favor of hardcoding this and eschewing the use of prefs entirely.

oyiptong:
Good feature. I'd recommend the use of fetch as it's a nicer API.

We need to choose between prefs and consts, but not both. You can set a pref in `package.json`. It will populate the pref on addon startup.

We need to make sure to remove the pref on addon uninstall (but not on upgrade/downgrade/disable etc).

Should you choose to use a hardcoded const, lib/main.js is a good location and might be the simples for now.

lib/main.js
```diff
@@ -8,6 +8,8 @@ const {Cu} = require("chrome");
 Cu.import("resource://gre/modules/ClientID.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
 
+const PING_ENDPOINT = "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream";
```
pdehaan: Should this be in a env-specific config or something? It seems weird that we'd always send data to a staging server from a prod environment.

oyiptong: I agree, we should set it in `package.json` as a default pref.

e.g.

```javascript
  "preferences": [
    {
      "name": "telemetry.ping.endpoint",
      "title": "Telemetry server endpoint",
      "type": "string",
      "value": "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream"
    }
  ]
```

Here's one can use `simplePrefs`:

```javascript
const simplePrefs = require("sdk/simple-prefs");

simplePrefs.prefs["telemetry.ping.endpoint"];
```

For development, we probably want to give that pref another value. We should use the `dev-prefs.json` config file.
JPM commands for test and for command line use (`npm run once` etc.) should use `dev-prefs.json`.

e.g.
`$ jpm run --prefs ./dev-prefs.json`

We should amend existing npm scripts to make that work. The contents of `dev-prefs.json` could be the staging server.

@ncloudioj what would be the two endpoints we need here for staging and prod?

oyiptong: One more note about the endpoints, the test environment should not send any telemetry pings.

oyiptong:
Good patch. we're almost there. It just needs some more work with the prefs/runtime parameters and then it's a go!

lib/main.js
ncloudioj: - stage: https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream
- prod: https://tiles.services.mozilla.com/v3/links/activity-stream

package.json
```diff
@@ -101,11 +101,11 @@
     "start:css": "npm run bundle:css && npm run bundle:css -- -w",
     "start:html": "npm run bundle:html -- --baseUrl http://localhost:1963/",
     "start:server": "live-server data/content --port=1963 --no-browser",
-    "firefox": "jpm run -b beta",
+    "firefox": "jpm run -b beta --prefs ./dev-prefs.json",
```
oyiptong: you might need to change yamscripts.yml and run `npm run yamscripts` to regenerate the package.json for the tasks

oyiptong: you did that. nevermind

oyiptong:
r+

Issue #285, opened 2016-03-09
opening new tab clears url bar after i've type a couple characters


edmoz:
1. open new tab
2. type one or two keystrokes

actual: you see the first key, then when new tab is loading, the url bar is cleared out.  This seems to be a performance issue of typing just before the new tab loads which clears out the URL bar.

expected: any characters entered should not be deleted.

edmoz:
this hurt me deeply and I think we need this fixed by public release

rlr:
grabbing to work on today

rlr:
landed fix.

edmoz:
:heart:

PR #301, opened 2016-03-15
fix(addon) #285: dont clear newtab after typing

rlr:
r? @oyiptong @k88hudson

### DIFF
--- a/lib/AppURLHider.js
+++ b/lib/AppURLHider.js
@@ -31,9 +31,13 @@ AppURLHider.prototype = {
   /**
    * Hide the URL of the most recent window.
    */
-  hideURL: function() {
+  _hideURL: function(url) {
     let browserWindow = Services.wm.getMostRecentWindow("navigator:browser");
-    browserWindow.gURLBar.value = "";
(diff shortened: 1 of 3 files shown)


lib/AppURLHider.js
```diff
@@ -31,9 +31,13 @@ AppURLHider.prototype = {
   /**
    * Hide the URL of the most recent window.
    */
-  hideURL: function() {
+  _hideURL: function(url) {
     let browserWindow = Services.wm.getMostRecentWindow("navigator:browser");
-    browserWindow.gURLBar.value = "";
+    // Only hide the url if it matches the expected app url.
+    if (browserWindow.gURLBar.value === url) {
```
pdehaan: what's the difference between this line, and the `this.isAppURL() {}` method?

rlr: this is checking what the browser chrome is actually displaying. The isAppURL checks the tab actual URL to our list of know app URLs. Make sense? I know, kind of confusing.

k88hudson:
Looks good to me!

Issue #288, opened 2016-03-10
Replace Embedly


pdehaan:
I was investigating a bug yesterday (where we were mishandling sites with query params), but it brought up an interesting point of what exactly we need/use Embedly for.

It looks like without Embedly, we get about 9 keys back from our queries:
- `bookmarkDateCreated`
- `bookmarkGuid`
- `favicon` (base64 encoded data uri)
- `frecency`
- `lastVisitDate`
- `parsedUrl{}`
- `title`
- `type`
- `url`

Results injected with the Embedly middleware add the following ~22 keys (see [Embedly `extract` API docs](http://docs.embed.ly/docs/extract#response-attributes)):
- `app_links[]`
- `authors[]`
- `cache_age`
- `content`
- **`description`**
- `embeds[]`
- `entities[]`
- **`favicon_colors[]`**
- **`favicon_url[]`**
- **`images[]`**
- `keywords[]`
- `language`
- `lead`
- `media{}`
- `offset`
- `original_url` (ie: "https://news.ycombinator.com")
- **`provider_display`** (ie: "news.ycombinator.com")
- **`provider_name`** (ie: "Ycombinator")
- `provider_url` (ie: "https://news.ycombinator.com")
- `published`
- `related[]`
- `safe`

So I think the only keys we use above are:
- `description` &mdash; used for the spotlight results.
- `favicon_colors[0]` &mdash; for the background color on "Top Sites" and fallback icons for "Top Activity"
- `favicon_url` &mdash; kind of redundant since we have a base64 encoded version already which would save us a network request.
- `images[0]` &mdash; only used for the 3 "Spotlight" results.
- `provider_display` and `provider_name` &mdash; only used for the fallback icon letter and possibly "Top Sites" friendly name. Probably easily extracted from the `parsedUrl` object.

Not sure if I'm missing anything, but it seems like we're barely using any of the features from Embedly, and it may be pretty easy to build our own version. The big features being:

1. How can we get the primary favicon colors from our existing base64 encoded `favicon` data uri?
2. How can we get the primary article image for "Spotlight" results?
3. How can we get the site description for "Spotlight" results?

But if we can live without `favicon_colors`, I think we'd only to use the Embedly middleware for the 3 spotlight URLs so we can get the spotlight `description` and `images[0]`.

----

```json
"FrecentHistory": {
  "rows": [
    {
      "bookmarkDateCreated": null,
      "bookmarkGuid": null,
[...]

PR #304, opened 2016-03-15
feat(addon): #258 Implement client in the addon

### LINKED ISSUE #258: Implement Embedly client in the addon
This tracks the task to implement an embedly client on the add-on side.

## Embedly client

The embedly client, given a list of urls should:

1. clean up URLs to remove privacy-sensitive information (e.g. HTTP basic auth, many query parameters)
2. check a local disk-based cache for an existing URLs and obtain embedly summary
3. for urls which cause a cache miss, formulate one POST query to send to our embedly client
  * upon receiving a response, store embedly responses in disk-based cache, together with a timestamp
4. return results that have embedly summaries to content

One possible design is to have this embedly cache/request run as a filter that gets called after each DB query results, probably in `lib/ActivityStreams.js`.

For urls it can return immediately, it should send all to content. For urls for which it has no summary, it should send the data without summaries yet.

Content should expect the summaries to be present with a message from chrome and not call out to our proxy directly.

## Asynchronous tasks

We need to run two asynchronous tasks:

1. periodically clean up the disk-based cache. If items are > 24H, the items should be evicted. If we find some data storage that has automatic expiry, this is a moot point. Annotations could be a data store we could use, it's an abstraction on top of sqlite with the notion of expiry
2. on addon boot up, and perhaps every 6 hours, we should trigger all queries that a normal content page load would, and pro-actively prime our caches by running our client

sarracini:
Things this does:
1. checks a local disk-based cache for an existing URLs and obtain embedly data using simple-storage (on addon side)
2. for urls which cause a cache miss, formulates POST query to send to embedly. Upon receiving a response, stores embedly data in cache + a timestamp
3. return results that have embedly data to content
4. periodically cleans up the cache for items > 24H
5. on addon boot up, and every 6 hours, asynchronously primes the cache with queries that a normal content page load would 
6. removes content-side embedly middleware which made requests every single time (yay!)
7. filters incoming URLs for allowed protocols and disallowed hosts
8. sanitizes incoming URLs to remove sensitive information in the query, as well as normalizes the URLs
9. dedupes the incoming URLs

As we discussed, this only sends down things that are cached and so it doesn't block on embedly returning the data. Instead, it will add it to the cache and the next time a request for that link comes up, it will have the stored data. The fetching is done asynchronously so we don't block on the embedly server for getting the cached links down. Since the server handles 25 urls per request, a request queue is made with 25 links at a time, and a new async task to request embedly data is spawned for each bundle of 25.

Things to do:
1. Tests

A note about testing: 
Right now localhosts are on a blacklist for URL hosts. When testing we'll be serving things (like images) on localhost and so we'll need to pref off the filters to allow for those links to be tested.

Review can be started on the parts that are done while I work on the things that are left to do
@oyiptong @k88hudson 

Fixes #258

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/304)


content-src/lib/filters.js
```diff
@@ -16,7 +13,7 @@ function createFilter(definition, name) {
     let result = true;
     definition.forEach((test, i) => {
       if (!test(item)) {
-        log(`omitting ${item.url}\nbecause it failed test #${i} of ${name}`);
+        console.log(`omitting ${item.url}\nbecause it failed test #${i} of ${name}`);
```
pdehaan: We may want to add back the `const {log} = require("lib/logger");` bit so we aren't always sending the output to the user's DevTools console:

```
/home/travis/build/mozilla/activity-streams/content-src/lib/filters.js
  16:9  warning  Unexpected console statement  no-console
```

content-src/lib/filters.js
```diff
@@ -27,9 +24,6 @@ function createFilter(definition, name) {
 const URL_FILTERS = [
   (item) => !!item.url,
   (item) => !!item.parsedUrl,
-  // This is temporary, until we can use POST-style requests
-  // see https://github.com/mozilla/embedly-proxy/issues/1
-  (item) => item.url && item.url.length <= TEMP_MAX_LENGTH,
```
pdehaan: We may need to delete some tests around this as well:

```
FAILED TESTS:
  urlFilter
    ✖ should remove urls > max characters
      Firefox 45.0.0 (Linux 0.0.0)
    invalid array length
    @/home/travis/build/mozilla/activity-streams/content-test/index.js:52687:45 <- webpack:///content-test/lib/filters.test.js:52:43
```

lib/ActivityStreams.js
```diff
@@ -75,6 +80,27 @@ ActivityStreams.prototype = {
...
+        PreviewProvider.preview.primeCache(links);
+      });
+      PlacesProvider.links.getRecentLinks.then(links => {
+        PreviewProvider.preview.primeCache(links);
+      });
+      PlacesProvider.links.getFrecentLinks.then(links => {
+        PreviewProvider.preview.primeCache(links);
+      });
+      this._primeCache();
+    }.bind(this), PRIME_CACHE_TIMEOUT);
```
pdehaan: Not sure if we could switch to an arrow function here to avoid needing to use `.bind(this)`. :shrug:

lib/PreviewProvider.js
```diff
@@ -0,0 +1,138 @@
+/* globals Task */
+"use strict";
+
+const {Cu} = require("chrome");
+const ss = require("sdk/simple-storage");
+const {setTimeout} = require("sdk/timers");
+const {siteFilter} = require("content-src/lib/filters");
+
+const EMBEDLY_PROXY_MAX_LINKS = 25; // number of links embedly proxy accepts per request
+const EMBEDLY_ENDPOINT = "https://embedly-proxy.dev.mozaws.net/v2/extract";
```
pdehaan: We probably want to pull this from a config so we can tweak for stage/production.

lib/ActivityStreams.js
```diff
@@ -82,22 +108,30 @@ ActivityStreams.prototype = {
     switch (msgName) {
       case am.type("TOP_FRECENT_SITES_REQUEST"):
         PlacesProvider.links.getTopFrecentSites(msg.data).then(links => {
-          this.send(am.actions.Response("TOP_FRECENT_SITES_RESPONSE", links), worker);
+          PreviewProvider.preview.saveNewLinks(links);
+          let cachedLinks = PreviewProvider.preview.getCachedLinks(links);
```
pdehaan: nit: none of these values change, we could switch to a `const` instead. Eventually we should add an ESLint rule that enforces `let`+`const` over `var`, and using `const` for "unchanging" values.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,138 @@
...
+
+Cu.importGlobalProperties(["fetch"]);
+Cu.import("resource://gre/modules/Task.jsm");
+
+function PreviewProvider() {
+  this.init();
+}
+
+PreviewProvider.prototype = {
+  // TO:DO santize, filter, and dedupe urls
```
pdehaan: uber nit: change to "TODO:" (for easier grep, etc)

lib/PreviewProvider.js
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Makes the necessary requests to embedly to get data and store it in the cache
+   */
+  fetchAndCache(newLinks) {
+    let linkURLs = [];
+    // extract only the link urls to send to embedly
+    newLinks.forEach(link => {
+      linkURLs.push(link.url);
+    });
+    // const sites = this.dedupeURLs(newLinks.filter(urlFilter));
```
pdehaan: nit: delete this line if it isn't used.

content-test/lib/filters.test.js
```diff
@@ -48,10 +48,6 @@ describe("urlFilter", () => {
     const urls = [{url: null}, {}, {url: ""}];
     assert.deepEqual(urls.filter(urlFilter), []);
   });
-  it("should remove urls > max characters", () => {
-    const urls = [addParsedUrl("http://" + new Array(TEMP_MAX_LENGTH + 1).join("d") + ".com")];
```
pdehaan: Need to delete the `TEMP_MAX_LENGH` import on L2 as well:

```
/home/travis/build/mozilla/activity-streams/content-test/lib/filters.test.js
  2:45  error  'TEMP_MAX_LENGTH' is defined but never used  no-unused-vars
```

sarracini:
thanks @pdehaan! :)

lib/ActivityStreams.js
```diff
@@ -75,6 +80,27 @@ ActivityStreams.prototype = {
   },
 
   /**
+   * Every 6 hours pro-actively prime our emebedly cache with a normal content page load
+   */
+  _primeCache() {
```
oyiptong: a more suggestive name would be great, e.g. `_buildPreviewCache()`

lib/ActivityStreams.js
```diff
@@ -75,6 +80,27 @@ ActivityStreams.prototype = {
...
+    setTimeout(() => {
+      PlacesProvider.links.getTopFrecentSites.then(links => {
```
oyiptong: these should be function calls

lib/ActivityStreams.js
```diff
@@ -75,6 +80,27 @@ ActivityStreams.prototype = {
...
+        PreviewProvider.preview.primeCache(links);
+      });
+      PlacesProvider.links.getRecentLinks.then(links => {
+        PreviewProvider.preview.primeCache(links);
+      });
+      PlacesProvider.links.getFrecentLinks.then(links => {
+        PreviewProvider.preview.primeCache(links);
+      });
+      this._primeCache();
+    }, PRIME_CACHE_TIMEOUT);
```
oyiptong: The function should also run at boot time. Perhaps setTimeout is called in ActivityStream's init function

lib/ActivityStreams.js
```diff
@@ -5,14 +5,17 @@
...
 const {ActionButton} = require("sdk/ui/button/action");
 const tabs = require("sdk/tabs");
 const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {TabTracker} = require("lib/TabTracker");
+const {PreviewProvider} = require("lib/PreviewProvider");
 const {TelemetrySender} = require("lib/TelemetrySender");
 const {AppURLHider} = require("lib/AppURLHider");
 const am = require("content-src/actions/action-manager");
+const PRIME_CACHE_TIMEOUT = 216000; // every 6 hours, prime the cache
```
oyiptong: How about making that an option passed to ActivityStreams? That would make testing easier

lib/PreviewProvider.js
oyiptong: Indeed. We should add this as a pref in `package.json`

lib/ActivityStreams.js
sarracini: it's currently called in ActivityStream's constructor, right after the cache is cleaned here: https://github.com/mozilla/activity-streams/pull/304/files#diff-cd9b1d133214f36f1c20099d7f6817e6R52 
is that what you mean?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+function PreviewProvider() {
+  this.init();
+}
+
+PreviewProvider.prototype = {
+  // TODO: santize, filter, and dedupe urls
+  /**
+    * Clean up cache
+    */
+  cleanUpCache: Task.async(function*() {
```
oyiptong: cleanUpCache will run 24 hours in the future after this function is called.
You actually intend to call this right away, and have a timeout function seperately, correct?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+    setTimeout(() => {
+      let currentTime = Date.now();
+      Object.keys(ss.storage.embedlyData).forEach(item => {
```
oyiptong: style nit, feel free to take or drop:

```javascript
for (let item of ss.storage.embedlyData) {
  ...
}
```

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+  }),
+
+  /**
+    * Dedupe urls
+    */
+  dedupeURLs(items) {
+    let createKey =  (item => item);
+    let compare = (() => false);
+    const valueMap = new Map();
+    items.forEach(value => {
```
oyiptong: style nit, feel free to take or drop:

```javascript
for (let item of items) {
  ...
}
```

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+      this.cleanUpCache();
+    }, CLEAR_CACHE_TIMEOUT);
+  }),
+
+  /**
+    * Dedupe urls
+    */
+  dedupeURLs(items) {
+    let createKey =  (item => item);
+    let compare = (() => false);
```
oyiptong: what does this do? compare always returns `false`

oyiptong: I suppose you intend to compare canonical urls here?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+    return Array.from(valueMap.values());
+  },
+
+  /**
+    * Filter cached links out, and store their embedly data
+    */
+  getCachedLinks(links) {
+    let cachedLinks = [];
+    links.forEach(link => {
```
oyiptong: style nit, feel free to take or drop:

```javascript
for (let link of links) {
  ...
}
```

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+      if (link && ss.storage.embedlyData[link.url] && link.url === ss.storage.embedlyData[link.url].url) {
```
oyiptong: you might want to canonicalize the url before comparing to the cache

oyiptong: and of course, only storing the canonicalized url

lib/PreviewProvider.js
sarracini: don't mind this for now..i'm working on the deduping part

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+        cachedLinks.push(ss.storage.embedlyData[link.url]);
+      }
+    });
+    return cachedLinks;
+  },
+
+  /**
+    * Prime the cache with embedly data
+    */
+  primeCache: Task.async(function*(links) {
```
oyiptong: "priming" implies a certain action is only run once. it might be better to call `saveNewLinks` directly.
What was the intention with the `Task.async`?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+   */
+  saveNewLinks: Task.async(function*(links) {
+    let newLinks = links.filter(link => link && !ss.storage.embedlyData[link.url]);
+    let requestQueue = [];
+    while (newLinks.length !== 0) {
+      // we have some new links we need to fetch the embedly data for, put them on the queue
+      requestQueue.push(newLinks.splice(0, EMBEDLY_PROXY_MAX_LINKS));
+    }
+    // for each bundle of 25 links, spawn off a new request to embedly
+    requestQueue.forEach(requestBundle => {
```
oyiptong: style nit, feel free to take or drop:

```javascript
for (let requestBundle of requestQueue) {
  ...
}
```

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+      Task.spawn(function*() {
```
oyiptong: You could collect all the promises in an array and `yield Promise.all(promiseArray)`;

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }),
+
+  /**
+   * Makes the necessary requests to embedly to get data and store it in the cache
+   */
+  fetchAndCache(newLinks) {
+    let linkURLs = [];
+    // extract only the link urls to send to embedly
+    newLinks.forEach(link => {
```
oyiptong: style nit, feel free to take or drop:

```javascript
for (let link of newLinks) {
  ...
}
```

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+      Task.spawn(function*() {
+        this.fetchAndCache(requestBundle);
+      }.bind(this));
+    });
+  }),
+
+  /**
+   * Makes the necessary requests to embedly to get data and store it in the cache
+   */
+  fetchAndCache(newLinks) {
```
oyiptong: This could perhaps be a Task.async function?

It would avoid chaining the `then`'s below

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+
+const EMBEDLY_PROXY_MAX_LINKS = 25; // number of links embedly proxy accepts per request
+const EMBEDLY_ENDPOINT = "https://embedly-proxy.dev.mozaws.net/v2/extract";
+const CLEAR_CACHE_TIMEOUT = 86400000; // 24 hours to clear cache
+
+Cu.importGlobalProperties(["fetch"]);
+Cu.import("resource://gre/modules/Task.jsm");
+
+function PreviewProvider() {
+  this.init();
```
oyiptong: why not put the contents of init() in here?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,137 @@
...
+}
+
+PreviewProvider.prototype = {
+  // TODO: santize, filter, and dedupe urls
+  /**
+    * Clean up cache
+    */
+  cleanUpCache: Task.async(function*() {
+    setTimeout(() => {
```
oyiptong: You also need to keep track of the timeout so you can clear it on `unload`... oh yeah, you need an `unload` function

lib/PreviewProvider.js
sarracini: Yeah I was thinking of that too. I was concerned on blocking on all the promises to resolve before adding it to the cache. Say if we request 100 links, the first 25 (chunk 1) would resolve first then need to wait until the next 75 (chunks 2,3,4) resolve before being added to the cache, correct?

lib/PreviewProvider.js
oyiptong: I retract my comment. It might be useful in setting up your tests.

lib/ActivityStreams.js
oyiptong: If it does what I think it does, when you call `setTimeout`, it won't execute the callback you provide immediately.

i.e. if you run:
```javascript
let start = (new Date()).getTime();
setTimeout(() => {
  let step = (new Date()).getTime();
  console.log(`elapsed: ${step - start}`);
}, 3000);
```

The result will be close to 3000, meaning that the callback doesn't execute when you set the timeout. In this case, it will only execute after `PRIME_CACHE_TIMEOUT` milliseconds.

If I understand correctly, your intention is to populate the cache at bootup time, correct?

lib/PreviewProvider.js
oyiptong: There's a design we could do to make this better, but the current implementation can suffice: it is much better than the current situation.

With the code in this PR, `saveNewLinks` will potentially result in an async `fetchAndCache` task, depending on what the filter returns.

The idea behind a new design would be to have a pool of fetchers, so we can cap the number of outbound requests by a set number.

A way to implement this would be to have a queue of links that each call to `saveNewLinks` would insert into. Upon inserting, `saveNewLinks` would be careful not to duplicate canonicalized URLs in the queue.

Asynchronous "threads of executions" check this queue periodically and perform `fetchAndCache` queries. If the queue is empty, the threads of execution "sleep". Of course, they don't really sleep, they timeout for a short amount of time.

Basically, the idea is a common pattern called a "job queue".

lib/PreviewProvider.js
sarracini: hmm, the docs for fetch seem to always chain their ```then```'s. I'm not sure we could get away with getting rid of them if we use Task.async. https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

lib/PreviewProvider.js
sarracini: should i also clear the ```_buildCachePreview``` timeout in ActivityStream's unload also?

lib/PreviewProvider.js
oyiptong: `fetch` returns a promise. You can verify as follows:

```javascript
let fetchPromise = fetch('flowers.png');
if (fetchPromise instanceof Promise) {
  console.log("fetch returns a promise");
}
```

Using `Task.async`, you can `yield` on the promise when it completes successfully, and use `try`/`catch` blocks for promise rejections.

In the case of a succesful HTTP response, here's how you'd write it with `Task.async`:

```javascript
let someTask = Task.async(function*(url) {
  let jsonData = null;
  try {
    let response = yield fetch(url);
    if (response.ok) {
       jsonData = yield response.json();
    }
  } catch(e) {
    Cu.reportError(e);
  }
  return jsonData;
});

someTask("/api/flowers/");
```

lib/PreviewProvider.js
oyiptong: yes, we should keep track of each call to `setTimeout` that we make. The function returns an ID that we can use to clear later on if we need to.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,150 @@
...
+  }),
+
+  /**
+   * Makes the necessary requests to embedly to get data and store it in the cache
+   */
+  _fetchAndCache: Task.async(function*(newLinks) {
+    let linkURLs = [];
+    let json = null;
+    // extract only the link urls to send to embedly
+    newLinks.forEach(link => {
```
oyiptong: Maybe this?
```javascript
let linkUrls = newLinks.map(link => link.url);
```

lib/ActivityStreams.js
```diff
@@ -5,15 +5,18 @@
...
 const tabs = require("sdk/tabs");
 const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {TabTracker} = require("lib/TabTracker");
+const {PreviewProvider} = require("lib/PreviewProvider");
 const {TelemetrySender} = require("lib/TelemetrySender");
 const {PerfMeter} = require("lib/PerfMeter");
 const {AppURLHider} = require("lib/AppURLHider");
 const am = require("content-src/actions/action-manager");
+const PRIME_CACHE_TIMEOUT = 216000; // every 6 hours, prime the cache
```
oyiptong: maybe `PREVIEW_CACHE_TIMEOUT`?

lib/ActivityStreams.js
```diff
@@ -78,6 +84,34 @@ ActivityStreams.prototype = {
...
+      PreviewProvider.preview.saveNewLinks(links);
+    });
+    this._setBuildPreviewCacheTimeout(primeCacheTimeout);
+  },
+
+  /**
+   * Set up preview cache to be primed every 6 hours
+   */
+  _setBuildPreviewCacheTimeout(primeCacheTimeout) {
+    this._timeoutID = setTimeout(() => {
```
oyiptong: maybe `this._previewCacheTimeoutID`?

lib/ActivityStreams.js
pdehaan: Is this value in milliseconds?

According to [MDN `setTimeout()` docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/setTimeout), the `delay` should be specified in milliseconds.
According to Google's conversion, 216000ms is 0.06h (or 3.6 minutes).
According to https://www.unitjuggler.com/convert-time-from-h-to-ms.html?val=6, 6 hours is 21600000 ms.


Not sure if the comment is wrong, or the value, or maybe I'm missing something else.

lib/ActivityStreams.js
```diff
@@ -78,6 +84,34 @@ ActivityStreams.prototype = {
   },
 
   /**
+   * Prime our emebedly cache with a normal content page load
```
pdehaan: typo: "Embedly"

lib/ActivityStreams.js
sarracini: good catch! forgot a couple 0's...

lib/ActivityStreams.js
```diff
@@ -12,10 +13,12 @@ const {Memoizer} = require("lib/Memoizer");
 const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {TabTracker} = require("lib/TabTracker");
+const {PreviewProvider} = require("lib/PreviewProvider");
 const {TelemetrySender} = require("lib/TelemetrySender");
 const {PerfMeter} = require("lib/PerfMeter");
 const {AppURLHider} = require("lib/AppURLHider");
 const am = require("content-src/actions/action-manager");
+const PREVIEW_CACHE_TIMEOUT = 21600000; // every 6 hours, prime the cache
```
oyiptong: nit: every 6 hours, rebuild/repopulate the cache

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+        }
+      });
+      return result;
+    };
+  },
+
+  /**
+    * Santize the URL to remove any unwanted or sensitive information about the link
+    */
+  _sanitizeURL(site) {
```
oyiptong: nit: use `link` or `url`, not `site`, the 2 former are semantically different than the latter

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    }
+    // extract and parse the query parameters, if any
+    url.search.slice(1, url.search.length);
+    let paramArray = url.search.slice(1, url.search.length).split("&");
+    paramArray.forEach(item => {
+      let tmp = item.split("=");
+      queryParams.set(tmp[0], tmp[1]);
+    });
+    // remove the unwanted parameters and update the query string
+    REMOVE_KEYS.forEach(key => queryParams.delete(key));
```
oyiptong: please use a whitelist instead of a blacklist.

You can probably just lift it from: https://github.com/mozilla/activity-streams/blob/master/content-src/lib/utils.js#L3

Please remove code from content that will be turned redundant. After using `ALLOWED_QUERY_PARAMS`, `REMOVE_KEYS` in utils.js becomes redundant too.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    // remove extra slashes in pathname and construct back a safe pathname
+    if (url.pathname) {
+      let pathItems = url.pathname.split("/");
+      let safePathItems = pathItems.filter(item => item.length !== 0);
+      let safePath = "/";
+      url.pathname = safePath + safePathItems.join("/");
+    }
+
+    // if the url doesn't have a protocol, use https
+    if (!url.protocol) {
```
oyiptong: the url will always have a protocol, otherwise URL object creation will fail

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    REMOVE_KEYS.forEach(key => queryParams.delete(key));
+    queryParams.forEach((val, key) => {
+      if (val.length !== 0 && key.length !== 0) {
+        safeQueryParam = safeQueryParam + key + "=" + val + "&";
+      }
+    });
+    url.search = safeQueryParam.slice(0, safeQueryParam.length - 1);
+    url.hash = "";
+
+    // remove extra slashes in pathname and construct back a safe pathname
```
oyiptong: to normalize the url, you should also take into account `/../` and `/./`

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+      let safePathItems = pathItems.filter(item => item.length !== 0);
+      let safePath = "/";
+      url.pathname = safePath + safePathItems.join("/");
+    }
+
+    // if the url doesn't have a protocol, use https
+    if (!url.protocol) {
+      url.protocol = "https:";
+    }
+
```
oyiptong: please add:
* the removal of basic HTTP auth if present
* port normalizing, i.e. if it is 80, ensure http protocol, if it is :443, ensure https protocol

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    url = new URL(url);
+    let key = url.host.replace(/www\.?/, "");
+    key = key + url.pathname;
+    return key;
+  },
+
+  /**
+    * Canonicalize urls by deduping them, then sanitizing them
+    */
+  _dedupeURLs(sites) {
```
oyiptong: please use `urls` instead of `sites`

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    let dedupedSites = new Map();
```
oyiptong: `uniqueURLs`

lib/PreviewProvider.js
oyiptong: actually, following the code, they should instead be called `links`

lib/PreviewProvider.js
oyiptong: `uniqueLinks`

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+      url.protocol = "https:";
+    }
+
+    return url.toString();
+  },
+
+  /**
+    * Create a key based on a URL in order to dedupe sites
+    */
+  _createKey(url) {
```
oyiptong: the key should also drop the protocol

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+  _dedupeURLs(sites) {
+    let dedupedSites = new Map();
+    sites.forEach(site => {
+      let urlkey = this._createKey(site.url);
+      if (!dedupedSites.has(urlkey)) {
+        dedupedSites.set(urlkey, site);
+      } else {
+        // if two sites come in with 'http' and 'https', pick 'https'
+        if (!dedupedSites.get(urlkey).url.contains("https")) {
+          dedupedSites.delete(urlkey);
```
oyiptong: no need to delete before setting

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+      this._fetchAndCache(requestBundle);
+    });
+  }),
+
+  /**
+   * Makes the necessary requests to embedly to get data and store it in the cache
+   */
+  _fetchAndCache: Task.async(function*(newLinks) {
+    let json = null;
+    let sites = this._dedupeURLs(newLinks.filter(this._URLFilter(URL_FILTERS)));
```
oyiptong: we might still be sending multiple duplicates to our embedly proxy if the urls happen to not be in the expected batch

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+      }
+    });
+    return cachedLinks;
+  },
+
+  /**
+   * Filter out new links and request them from embedly
+   */
+  saveNewLinks: Task.async(function*(links) {
+    let newLinks = links.filter(link => link && !ss.storage.embedlyData[link.url]);
```
oyiptong: might be better to dedupe at this point

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+      site.url = this._sanitizeURL(site.url);
+      freshSites.push(site);
+    });
+    return freshSites;
+  },
+
+  /**
+    * Filter cached links out, and store their embedly data
+    */
+  getCachedLinks(links) {
```
oyiptong: We actually want to keep a mapping between sanitzed urls and real urls.

We will use sanitized urls to send to the embedly proxy, and to obtain a summary from cache.
At the end of the day though, we want to return to the user their original URL.

The site summary from embedly will be obtained from a sanitized url, but if they are going to https://foo@bar:example.com/thing?id=blah&author=quuz&date=today we should let them still go to that link

oyiptong:
Thanks for the work on the preview module!

There's some more work to be done. Some about the sanitization of urls, and some more on the overall functionality.

The fundamental takeaway is: we generate summaries of sites through a sanitized version of the urls sent to our proxy to protect the user, however, we show the original URL back to them.

lib/PreviewProvider.js
oyiptong: Let me actually rephrase:

We want to "augment" the links we obtain from queries with summaries we have in the cache.

Say the output of our query, and the input to this function is:
```javascript
[
  {url: "https://foo:bar@example.com/thing?id=blah&author=quuz&date=today", title: "A day at Foo Bar"},
  {url: "https://reddit.com/r/programming", title: "/r/programming"}
]
``` 

We want to return the exact same thing, except with data from our cache augmenting the original, in the following two examples, the embed, description keywords and more:

```javascript
[
  {url: "https://foo:bar@example.com/thing?id=blah&author=quuz&date=today", title: "A day at Foo Bar",
   embeds: [], description: "...", other_embedly_metadata: ...},
  {url: "https://reddit.com/r/programming", title: "/r/programming",
   embeds: [], description: "...", other_embedly_metadata: ...}
]
```
In effect, we'll use our sanitized URL to both fetch a preview through embedly and as a key to our preview cache.

oyiptong: The function will also filter out the URLs for which we have no preview

lib/PreviewProvider.js
oyiptong: Also, the `embedlyData` object will contain sanitized URLs as keys, the input `links` will not.
You need to generate a sanitized URL to be able to compare in the preview cache.

oyiptong:
Another couple of notes:
* There might be some redundant and/or dead code to remove from content
* I heard travis CI calling and it said: Can I haz tests plz? ;-)

oyiptong:
more on embedly metadata here: #288

lib/PreviewProvider.js
sarracini: making a new URL object automatically removes the default port if there is one. 
you mean the "username" and "password" fields of the URL object if present, right?

oyiptong: Oh! Awesome!

And yes, that's what I meant

lib/PreviewProvider.js
sarracini: the key only returns the host (without www) and the pathname. it drops protocol and the query.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+        if (!dedupedSites.get(urlkey).url.contains("https")) {
+          dedupedSites.delete(urlkey);
+          dedupedSites.set(urlkey, site);
+        }
+      }
+    });
+    let freshSites = [];
+    dedupedSites.forEach((site) => {
+      // sanitize and normalize each site
+      site.url = this._sanitizeURL(site.url);
```
k88hudson: There's a possibility this could cause side effects since this modifies the original `site`/`link` object, it would be better to use `Object.assign`:

````js
return Array.from(dedupedLinks.values())
  .map(link => Object.assign({}, link, {url: this._sanitizeURL(link.url)})));
```

k88hudson:
Nice work @sarracini, this is looking great. Big improvement on UX.

lib/PreviewProvider.js
k88hudson: `REMOVE_KEYS` refers to keys on the url object, i.e. used to remove basic auth.
Since however I don't think `URL` allows deleting the username/password properties to modify basic auth, so you'll need to use a different strategy

sarracini: the ```REMOVE_KEYS``` approach is nice for when the auth is in the query string, if the url has basic auth, the URL object picks it up and puts it in a "username" and "password" field, which i can just clear if they exist. 
maybe a more appropriate name for ```REMOVE_KEYS``` would be ```DISALLOWED_QUERY_PARAMS```?

lib/ActivityStreams.js
```diff
@@ -103,22 +137,30 @@ ActivityStreams.prototype = {
     switch (msgName) {
```
jaredkerim: It looks like this switch body is doing the same thing for every response, you could just build a mapping between the event name "TOP_FRECENT_SITES_REQUEST" and its memoized data source this._memoized.getTopFrecentSites and then extract the rest out instead of repeating it.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    let queryParams = new Map();
+    let safeQueryParam = "";
+    let url = new URL(site);
+
+    if (url.search.length === 0) {
+      return url.toString();
+    }
+    // extract and parse the query parameters, if any
```
jaredkerim: This can be written a little more succinctly as

new Map(query.slice(1).split('&').map((pair) => pair.split('=').map(decodeURIComponent)))

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+  this.init();
+}
+
+PreviewProvider.prototype = {
+  _timeoutID: null,
+
+  /**
+    * Clean up cache on addon bootup
+    */
+  cleanUpCache: Task.async(function*() {
```
jaredkerim: You could check if a key's timeout has expired at key lookup time rather than setting up a timeout, and then you don't need a background process at all.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+      this.cleanUpCache();
+    }, CLEAR_CACHE_TIMEOUT);
+  },
+
+  /**
+    * Filter out sites that do not have acceptable protocols and hosts
+    */
+  _URLFilter(definition) {
+    return function(item) {
+      let result = true;
```
jaredkerim: I think you can say

```js
return function (item) {
  return definition.every(test => test(item));
}
```

lib/PreviewProvider.js
jaredkerim: I experimented with this on the proxy side, but the point was raised that many sites store page information in the query string, ie bugzilla puts the ticket id in the query string, google searches, etc and @k88hudson raised the point that embedly can return different metadata for different pages based solely on the query string.  If that's the case then we should drop only the protocol here.

lib/PreviewProvider.js
```diff
@@ -0,0 +1,239 @@
...
+    * Canonicalize urls by deduping them, then sanitizing them
+    */
+  _dedupeURLs(sites) {
+    let dedupedSites = new Map();
+    sites.forEach(site => {
+      let urlkey = this._createKey(site.url);
+      if (!dedupedSites.has(urlkey)) {
+        dedupedSites.set(urlkey, site);
+      } else {
+        // if two sites come in with 'http' and 'https', pick 'https'
```
jaredkerim: I'm worried about two things here

1) separating and potentially duplicating logic between the client and server with respect to how we decide to group or cluster URLs around protocol or domain or what have you.  I'm in favour of the client filtering out potentially sensitive data from URLs such as query strings which might contain private user IDs, but removing protocol can actually introduce issues and 

2) I've already witnessed situations where embedly's responses will vary solely based on protocol, so I think it might be advisable to actually preserve both URLs rather than deduping them.

lib/PreviewProvider.js
jaredkerim: Furthermore @pdehaan has discovered that embedly may respond differently for the same URL with different protocols, I think we should preserve protocol to cut back on confusing or unexpected behaviours.

lib/PreviewProvider.js
jaredkerim: I think we need to consolidate this code base on a concept of the original URL and the 'sanitized' URL which will serve as both a cache key as well as the URL which is sent to embedly, rather than having an additional mapping concept of original/sanitized/cachekey.  Particularly if we drop the concept of scrubbing query string and protocol altogether from the cache key (which I believe introduces more harmful behaviours than removes them).

lib/PreviewProvider.js
oyiptong: We don't need to blacklist query params if we have a whitelist

lib/PreviewProvider.js
oyiptong: to @jaredkerim's point, the `key` for deduping should be the sanitized url. It should contain the whitelisted url query params. If the only differentiating factor is a `www` sub-domain name, (keeping in mind that path and query params are the same), I'd say it's not a differentiating factor we care about.

That said, I maintain that our sanitized pages should still be protocol agnostic. One way to think about it is as follows:

Do we care if http://www.reddit.com and https://www.reddit.com both appear on the timeline? We do, and we don't want it to happen.

lib/PreviewProvider.js
oyiptong: I agree. the cache key should be the sanitized URL, at least to a first approximation.

I think @sarracini was using the cachekey to dedupe URLs to send to the embedly proxy. The cache key should be the sanitized URL, or something very close to it.

The only gotcha is the case of url scheme/protocol. We *do* want to send a url scheme to embedly, but we don't want to consider it for caching, given that if the only difference for a url is the scheme, then the content is certainly the same.

jaredkerim: @oyiptong @pdehaan and myself have observed that embedly's content can be different depending solely on protocol and this can lead to confusing and unexpected behaviours.

oyiptong: Are there examples you could share?

oyiptong:
To clarify:

Upon querying, we want links from a database query to go through a filter, which will:

1. verify for presence of summaries for urls in an in-memory cache via a `cache key` (a transformation of the url)
2. filter out links that are not in the cache
3. for links that are not present:
  1. check for duplicates via a cache key
  2. fetch and cache asynchronously with a `sanitzed url`
  3. if there are two original URLs that have the same cache key, e.g. https://reddit.com vs http://reddit.com, choose the https version of the sanitized url
4. the filter augments the original link objects with summaries found in the cache

### Cache key vs Sanitized URL

The `cache key` and `sanitized URL`s are almost the same. The only difference are:

1. If the hostname starts with `www.` drop those 4 leading characters
2. Drop the hostname http scheme

e.g. https://www.reddit.com/r/programming?utm=1231234 will yield:

original URL | cache key | sanitized URL |
---- | ---- | ---- |
https://www.reddit.com/r/programming?utm=1231234 | reddit.com/r/programming | https://www.reddit.com/r/programming |
http://reddit.com/r/programming?utm=1231234 | reddit.com/r/programming | http://reddit.com/r/programming

If both of those are in the same fetch, they'd only result in 1 URL to be sent to embedly, the https URL.

The reasoning is that if there are 2 history entries, both of those URLs successfully loaded in the past.
Given two equivalent URLs, choose to make a preview of the most secure one.

### Definitions

* The `cache key` is used to: deduplicate URLs while constructing a set of URLs to send to our embedly proxy. It is also used as a key to look up summaries in our in-memory summary cache
* The `sanitized URL` is the actual URL we send to our embedly proxy
* The `original URL` is returned to the user unmodified via content. It is the link that is offered to the user to click

lib/PreviewProvider.js
oyiptong: We saw that:

http://groupon.com redirects to https://www.groupon.com

Which means that the equivalent API response of http://embed.ly/docs/explore/extract?url=http%3A%2F%2Fgroupon.com returns an error.

I verified that there are actually two history items created in the example above, because Firefox considers redirects successful loads.

We can actually discriminate between the different kinds of loads:

 `sqlite> select p.url, v.visit_type, v.session from moz_historyvisits v, moz_places p where p.id = v.place_id order by v.visit_date DESC limit 5;`

```
url                                       visit_type  session
------------------------------------      ----------  ----------
https://www.groupon.com/                  5           0
http://www.groupon.com/                   2           0
```

We'll leave that as an exercise for later, however.

In most cases, a successful redirect is what we want.

oyiptong: More about the possible visit types at https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/nsINavHistoryService.idl#1196
The constants start with `TRANSITION_`

oyiptong:
another possible good compromise for a tie breaker for choosing between two URLs returning the same cache keys in a fetch would be to pick the URL with the latest visit date.

lib/PreviewProvider.js
jaredkerim: After discussing this with @oyiptong I agree that we should use the following layers:

1) unmodified URL (this is what is actually stored in the history and what will be loaded if the user clicks on this in their history or activity stream)

2) sanitized URL (this has some query params and basic auth scrubbed, path is normalized, includes protocol, and is what will be sent to the embedly proxy)

3) cache key (this is the sanitized URL except with the protocol dropped) this is used just for looking the URL metadata up in our local simplestorage cache

The proxy will perform no URL transformations and will query and cache URLs exactly as they are sent to it.  This way we don't wind up with a confusing situation where a URL is being transformed or miscached and we're not sure where that transformation is taking place.  If we only transform URLs in the Firefox client, then we can be sure transformation errors are handled by that code base.

The case raised by @pdehaan (the case where HTTP is disallowed, HTTPS is allowed, and we have mistakenly cached the HTTP empty dataset under the same key as the HTTPS URL) should be prevented in the client because only URLs which the browser is able to establish a successful connection to will appear in the history, so we should not be mistakenly caching data for URLs which contain no metadata.

lib/PreviewProvider.js
jaredkerim: @oyiptong @sarracini I think we can simplify this using the URL layers described above

1) Receive a list of URLs from the places db

2) Sanitize them using the above method

3) Passing that sanitized list through a set will remove duplicates

4) Generate the cache keys of the sanitized URLs to look up in the simple storage local cache

5) Find the set of URLs which were not locally cached and pass those to the embedly proxy (these are sanitized, de duped, and not locally cached)

sarracini:
Here's what I gathered from everyone here, if this is agreed upon then I'll make the changes:

1. An array of links come in from the places db
2. Filter out the immediate links that have disallowed hosts, and that do not have ```http``` or ```https``` protocols
3. These filtered links are then **ALL** sanitized. Sanitized URLs **include** protocol, scrubs the query params to remove sensitive information, normalizes the path, and removes the hash
4. Pass the sanitized set of links through a function to dedupe them 

     4.1. In order to dedupe we use a cache key which **drops** the protocol, and keeps the pathname
     4.2. Use this key to find duplicates within the set of sanitized links that we passed in.
     4.3. If two sanitized URLs have the same cache key, use the ```https``` version
5. At this point we have a filtered, sanitized, and deduped set of links, each with a sanitized URL. Now we can check the cache.
6. Check the cache using the same cache key mentioned above (which drops the protocol) 

     6.1. If the link is not in the cache, send the sanitized URL to embedly asynchronously.
     6.2. Embedly will get the data we need, and we will cache the **non-sanitized** URL. Instead we cache the URL returned from the **cache key**, along with it's respective embedly data. 
     6.3. For entries that are in the cache (i.e. they match the cache key), retrieve their respective embedly data
     6.4. Finally, augment the original link that came in, with it's respective cached embedly data.
7. Return the set of cached links

We will use Olivier's criteria for cache key and sanitized URLs to drop the protocol if it is ```www``` and use ```https``` over ```http```

Am I getting this right?
@jaredkerim @oyiptong

lib/PreviewProvider.js
```diff
@@ -0,0 +1,226 @@
...
+    requestQueue.forEach(requestBundle => {
+      this._fetchAndCache(requestBundle);
+    });
+  }),
+
+  /**
+   * Makes the necessary requests to embedly to get data and store it in the cache
+   */
+  _fetchAndCache: Task.async(function*(newLinks) {
+    let json = null;
```
jaredkerim: - you should avoid using `json` as a varname because that can overwrite the name of the actual json processing library (i know in JS it's usually `JSON` not `json`, it's just a practice you should always keep in your mind), so try to call things like response_json or url_json or something to indicate what the data the json is holding is for

- why do we need to declare this here?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,226 @@
...
+    // extract only the sanitized link urls to send to embedly
+    let linkURLs = newLinks.map(link => link.sanitizedURL);
```
jaredkerim: :+1:

lib/PreviewProvider.js
sarracini: good point! i'll move it into the try

lib/PreviewProvider.js
```diff
@@ -0,0 +1,226 @@
...
+    try {
+      let response = yield fetch(EMBEDLY_ENDPOINT, {
+        method: "POST",
+        body: JSON.stringify({urls: linkURLs}),
+        headers: {"Content-Type": "application/json"}
+      });
+      if (response.ok) {
```
pdehaan: What happens (or should happen) if we get a response from the embedly-proxy and the response is NOT `ok`?
Do we need an `else { ... }` here, or am I missing something? I just wanted to make sure there isn't some weird case where the proxy returns an error but something isn't thrown/caught and we're just silently ignoring something.

sarracini: Hmmm, from what I understand the proxy sends out an error if it fails (https://github.com/mozilla/embedly-proxy/blob/master/embedly/embedly/api/views.py#L64) and that error gets caught in the reducers here https://github.com/mozilla/activity-streams/blob/master/content-src/reducers/reducers.js#L12. So we should get an indication if something went wrong, even though the preview provider doesn't deal with it directly (unless it was a problem with the fetch in which case the catch will report the error)

lib/PreviewProvider.js
```diff
@@ -0,0 +1,226 @@
...
+      return "";
+    }
+
+    let queryParams = {};
+    let safeQueryParam = "";
+    let newURL = new URL(url);
+
+    // extract and parse the query parameters, if any
+    if (newURL.search.length !== 0) {
+      let tmp = new Map(newURL.search.slice(1).split("&").map(pair => pair.split("=").map(decodeURIComponent)));
```
jaredkerim: Try to avoid calling something 'tmp', here it's the query params, so call it that.  queryParamsMap or something.

Also do we need to create an empty object on line 77, and then create a map, and then populate the object from the map?  why not just say

    let queryParams = new Map(newURL.search.slice(1).split("&").map(pair => pair.split("=").map(decodeURIComponent)));

sarracini: so this was kinda a tricky and frustrating thing...
queryParams needs to be an actual object, not a map, so that i can filter the allowed query params. so i make a temporary map, but then convert it to an object so i can easily filter across it. 

apparently filtering on a map is not a thing?

lib/PreviewProvider.js
```diff
@@ -0,0 +1,226 @@
...
+    });
+
+    return cachedLinks;
+  },
+
+  /**
+   * Filter out new links and request them from embedly
+   */
+  saveNewLinks: Task.async(function*(links) {
+    let sites = this._uniqueLinks(links.filter(this._URLFilter(URL_FILTERS)));
```
jaredkerim: We're doing this twice, here and on line 144.  Is it possible to do this once and then pass that dataset around?

lib/PreviewProvider.js
jaredkerim: let queryParams = new Map(newURL.search.slice(1).split("&").map(pair => pair.split("=").map(decodeURIComponent)));
    Array.from(queryParams.keys()).filter(.......)

The only difference is that Object.keys() returns an Array and Map.keys() returns an Iterator which you need to convert to an Array using Array.from, which is a little annoying but not impossible, and even then it's an extra funcall vs a whole extra variable/object and a loop to copy the map into the object.

lib/PreviewProvider.js
jaredkerim: It looks to me like if the embedly-proxy response is an error (not 200 status) the response is just completely ignored.

lib/PreviewProvider.js
sarracini: oh excellent :dancer:

jaredkerim:
@sarracini Looking great!  Much more succinct, and we love succinct. :clap: :smile_cat: Keep me posted when tests come around.

lib/PreviewProvider.js
jaredkerim: So I don't know how the client wants to handle it.  A failure in this case means that there will be no data returned at all.  The proxy does not return partial results (yet).  Soon it will attempt to do result level validation and then it may throw out a result.  So later we will have to amend the client to do things like compare which URLs it sent with which it got back and possibly retry or something, but I think that can come later, for now that behaviour is poorly defined so let's just leave it out of this PR and handle it in a separate ticket.  In fact I will file that ticket.

lib/PreviewProvider.js
sarracini: https://github.com/mozilla/activity-streams/pull/304/commits/150797a51e2ac0f32e61fb8533b30a2ab094def7#diff-7da8abc4c82d27cce198b3adf58c2df4R134 and https://github.com/mozilla/activity-streams/pull/304/commits/150797a51e2ac0f32e61fb8533b30a2ab094def7#diff-cd9b1d133214f36f1c20099d7f6817e6R137  is this what you mean? :)

lib/PreviewProvider.js
k88hudson: Maybe we should file this as a follow-up ticket

lib/PreviewProvider.js
k88hudson: Filed https://github.com/mozilla/activity-streams/issues/439 to handle errors from embedly, we can do as a follow-up.

k88hudson:
This is looking good to land once we have tests :+1:

oyiptong:
Reviewed 1 of 4 files at r7.
Review status: 1 of 10 files reviewed at latest revision, 14 unresolved discussions.

---

*[lib/ActivityStreams.js, line 137 \[r2\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEUmm5IA02NFlkT2bap-r2-137:-KEXq2qk12XhBFfzFLdt:-1370556393) ([raw file](https://github.com/mozilla/activity-streams/blob/740cfe65947dea9625265e781e0491e20cb63b53/lib/ActivityStreams.js#L137)):*
let's leave it as-is for now. I know it's ugly, but we only have 4 queries. Let's refactor when we have more

---

*[lib/ActivityStreams.js, line 106 \[r8\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEXoXto5N6qEZbmwZyi:-KEXoXtptBA-YCPCvPQO:1724820894) ([raw file](https://github.com/mozilla/activity-streams/blob/a194db6010e03260f80333f290b6bdde620f1bd2/lib/ActivityStreams.js#L106)):*
we can actually make this one query to the proxy instead of 4 by collecting all links first then making the call to `saveNewLinks`

---

*[lib/PreviewProvider.js, line 35 \[r2\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEUmm5IA02NFlkT2bar-r2-35:-KEXqHPqHZYk7iWpDYFv:563901209) ([raw file](https://github.com/mozilla/activity-streams/blob/740cfe65947dea9625265e781e0491e20cb63b53/lib/PreviewProvider.js#L35)):*
thanks kate

---

*[lib/PreviewProvider.js, line 84 \[r2\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEUmm5IA02NFlkT2bar-r2-84:-KEXqZy8ub0xsl99TJ_C:-542168073) ([raw file](https://github.com/mozilla/activity-streams/blob/740cfe65947dea9625265e781e0491e20cb63b53/lib/PreviewProvider.js#L84)):*
I actually disagree. I value explicitness rather than conciseness. Chaining commands are notoriously hard to debug. That said, since the change is in, let's go with it

---

*[lib/PreviewProvider.js, line 101 \[r2\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEUmm5IA02NFlkT2bar-r2-101:-KEXtOY6LwkL87pterhd:-1089407949) ([raw file](https://github.com/mozilla/activity-streams/blob/740cfe65947dea9625265e781e0491e20cb63b53/lib/PreviewProvider.js#L101)):*
actually, `/../` and `/./` are taken care of with the `URL` constructor. Which means we won't need the following code

---

*[lib/PreviewProvider.js, line 131 \[r2\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEUmm5IA02NFlkT2bar-r2-131:-KEXuaZUcY6I179Tmo5b:2039822990) ([raw file](https://github.com/mozilla/activity-streams/blob/740cfe65947dea9625265e781e0491e20cb63b53/lib/PreviewProvider.js#L131)):*
nit: please change all mention of `sites` to `links`

---

*[lib/PreviewProvider.js, line 84 \[r8\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEXrFHNnoWrU9C--40W:-KEXrFHNnoWrU9C--40X:1388203097) ([raw file](https://github.com/mozilla/activity-streams/blob/a194db6010e03260f80333f290b6bdde620f1bd2/lib/PreviewProvider.js#L84)):*
I would re-write this a bit differently: this is creating a new string each time safeQueryParam is assigned.

Since strings are immutable in JS, this is a O(n^2) operation, where n is the number of characters

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
Reviewed 2 of 7 files at r1, 4 of 6 files at r3, 2 of 4 files at r7, 1 of 1 files at r8.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
:lgtm: r=myself

---

Review status: all files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-:-KEXwjIrjL9Daf3JyGrV:1545415986)*

oyiptong:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[lib/PreviewProvider.js, line 84 \[r8\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEXrFHNnoWrU9C--40W:-KEXyjHvBJwCDRudBS8A:-778827438) ([raw file](https://github.com/mozilla/activity-streams/blob/a194db6010e03260f80333f290b6bdde620f1bd2/lib/PreviewProvider.js#L84)):*
You want to collect all parameters, and call Array.join at the end.

e.g.

let params = ["foo=bar", "baz=qux].join("&");
newURL.search = params;

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
It's almost there. With a few fixes and a commit squash, we should be all set. Good job, @sarracini!

---

Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-:-KEXzOCPzv5pSkCmeeWv:1195539150)*

sarracini:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[lib/PreviewProvider.js, line 101 \[r2\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEUmm5IA02NFlkT2bar-r2-101:-KEY2A7TU9ghjfjRVgEe:1059446521) ([raw file](https://github.com/mozilla/activity-streams/blob/740cfe65947dea9625265e781e0491e20cb63b53/lib/PreviewProvider.js#L101)):*
i actually noticed this too when i was testing... the URL object doesn't remove extra slashes though, i replaced it with a regex for removing extra slashes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
Review status: 8 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[lib/ActivityStreams.js, line 106 \[r8\]](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-KEXoXto5N6qEZbmwZyi:-KEYNHtcSs347zXYKWE3:-509473754) ([raw file](https://github.com/mozilla/activity-streams/blob/a194db6010e03260f80333f290b6bdde620f1bd2/lib/ActivityStreams.js#L106)):*
you actually want to do `linksToSend.push(...links);`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
:lgtm:

---

Review status: 8 of 10 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-:-KEYPnymxfzU1kMS_3Yz:1417088721)*

oyiptong:
Reviewed 2 of 2 files at r9.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
r+ with the change to `...links`

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304#-:-KEYPxi5VpbWoxvtw9uk:2138218801)*

oyiptong:
Reviewed 1 of 4 files at r7.
Review status: 8 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

oyiptong:
Reviewed 2 of 2 files at r10.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-streams/304)*

PR #305, opened 2016-03-15
feat(addon): #295 places query caching

### LINKED ISSUE #295: Add in-memory query caching
This tracks a new feature:

In-memory query caching. The feature will allow a request/response flow from content to occur, storing results in-memory.

The data cached will, at first not include the site summaries, but can later include them.
This cache is invalidated whenever there is a history update that might cause any change.

For instance, when a new history item is inserted, the query cache for the recent sites is invalidated and repopulated. This is done by adding listeners to PlacesProvider.

The query cache is primed on addon startup.

oyiptong:
closes #295 

r? anyone?

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -1,4 +1,4 @@
-/* globals NewTabURL, EventEmitter, XPCOMUtils, windowMediator */
+/* globals NewTabURL, EventEmitter, XPCOMUtils, windowMediator, Task, Services */
 
 "use strict";
 
@@ -7,6 +7,8 @@ const {data} = require("sdk/self");
 const {PageMod} = require("sdk/page-mod");
 const {ActionButton} = require("sdk/ui/button/action");
 const tabs = require("sdk/tabs");
+const simplePrefs = require("sdk/simple-prefs");
+const {Memoizer} = require("lib/Memoizer");
 const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {TabTracker} = require("lib/TabTracker");
@@ -17,10 +19,14 @@ const am = require("content-src/actions/action-manager");
 
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource:///modules/NewTabURL.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
 
 XPCOMUtils.defineLazyServiceGetter(this, "windowMediator",
                                    "@mozilla.org/appshell/window-mediator;1",
                                    "nsIWindowMediator");
+
+XPCOMUtils.defineLazyModuleGetter(this, "Task",
+                                  "resource://gre/modules/Task.jsm");
 
 XPCOMUtils.defineLazyGetter(this, "EventEmitter", function() {
   const {EventEmitter} = Cu.import("resource://devtools/shared/event-emitter.js", {});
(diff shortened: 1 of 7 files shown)


lib/Memoizer.js
```diff
@@ -0,0 +1,58 @@
...
+
+  memoize(key, func) {
+    return () => {
+      if (!this._memo_all[key]) {
+        this._memo_all[key] = {};
+      }
+      let memo = this._memo_all[key];
+
+      // turn arguments to array, skip cache key
+      let args = Array.prototype.slice.call(arguments).slice(1);
```
pdehaan: Do we need this in ES6-syntax, or can't we just do something like:
```js
memoize(key, func) {
  return (cacheKey, ...args) => {
```

oyiptong:
adding @sarracini to review request

lib/Memoizer.js
oyiptong: wasn't aware of this syntax. I like. Thanks @pdehaan

oyiptong: `...args` is neat!

lib/Memoizer.js
```diff
@@ -0,0 +1,72 @@
...
+function Memoizer() {
+  this._cacheEnabled = simplePrefs.prefs["query.cache"];
+  this._memo_all = {};
+  this._onPrefChange = this._onPrefChange.bind(this);
+  simplePrefs.on("", this._onPrefChange);
+}
+
+Memoizer.prototype = {
+
+  memoize(key, func) {
```
mzhilyaev: perhaps memorize() instead of memoize

lib/ActivityStreams.js
```diff
@@ -209,6 +220,48 @@ ActivityStreams.prototype = {
   },
 
   /**
+   * Returns an object of functions with results cached
+   */
+  _get_memoized(cache) {
```
mzhilyaev: is name memoized() intended?  shouldn't it be memorized()?

oyiptong: I actually intended to call it memoize after the term "memoization": https://en.wikipedia.org/wiki/Memoization, which, according to wikipedia comes from "memorandum"

lib/ActivityStreams.js
```diff
@@ -132,17 +142,19 @@ ActivityStreams.prototype = {
...
-  /**
-   * Broadcast bookmark changes to pages
-   */
-  _handleBookmarkChanges(eventName, data) {
-    this.broadcast(am.actions.Response("RECEIVE_BOOKMARKS_CHANGES", data));
+    /* note: this will execute for each of the 3 notifications that occur
+     * when adding a visit: frecency:-1, frecency: real frecency, title */
+    this._invalidatePlacesCache();
+    this._asyncBuildPlacesCache();
+
```
mzhilyaev: What would happen if a user opens a page and does "Command Click" on 10 links from this page?
That will open 10 tabs, and will generate 10 history events, which in turn may cause cascading invalidate and cache rebuilds calls.  Not sure how bad it is, but at the very list we should not clean the cache if we are rebuilding it.  Something like:
```
_invalidatePlacesCache() {
if (_building_cache_flag) { 
   // do not disturb
   return;
}
_building_cache_flag = true;
// build the cache
....
// done with cache
_building_cache_flag = false;
```
It would be good to unit test this use case to make sure we don't get in some crazy concurrency situation when cache is cleaned while being rebuilt.

oyiptong: I was going to handle this in a later issue. In your outlined scenario, this will actually cause up to 30 history events.
3 for each new link:

1. link creation (with frecency -1)
2. frecency adjustment
3. title change

Also, the solution you propose might not work as a user expects: it will rebuild cache the first time a user creates a new history entry, but potentially not the other times, meaning the results might be stale.

The concurrency scenario you outline will not happen: Javascript is single-threaded. Every synchronous operation will occur one after the other. A sequence of invalidate/build will occur one after the other.

oyiptong: We can probably put the number of updates down to 10 rebuilds if we detect a new link and build a state machine.

oyiptong: But that's going to be hard. There may be many concurrent builds.

Issue #222, opened 2016-02-29
Add pagination UI


k88hudson:
Blocked on https://github.com/mozilla/activity-streams/issues/198 for implementation

We just need some UI for loading more stuff/next on the Timeline

pdehaan:
Not sure how relevant [the invision spec](https://mozilla.invisionapp.com/share/GA5RW90CW#/screens/130804847) is anymore, but it shows a "Show 7 more" link at the bottom of the page.

bryanbell:
When the user gets to the bottom we should show a "loading more items" spinner. Hopefully, it will load so fast that they never get to see that. 

When in doubt it should work like the Twitter Feed.

bryanbell:
![pasted_image_3_2_16__9_44_am](https://cloud.githubusercontent.com/assets/228979/13469247/539f8226-e05b-11e5-9bd6-152dacddd9a6.png)

example

k88hudson:
Blocked don https://github.com/mozilla/activity-streams/issues/312

pdehaan:
@k88hudson Is this finished now that #318 has merged, or are we still blocked on something?

k88hudson:
This is mostly completed by #318, we just need to add pagination to the bookmarks view

PR #318, opened 2016-03-17
feat(content): Add pagination to TimelineHistory

k88hudson:
This fixes half of #222, the rest is blocked by https://github.com/mozilla/activity-streams/issues/312

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -96,25 +96,26 @@ ActivityStreams.prototype = {
    */
   _respondToPlacesRequests(msgName, params) {
     let {msg, worker} = params;
+    const append = msg.meta && msg.meta.append;
     switch (msgName) {
       case am.type("TOP_FRECENT_SITES_REQUEST"):
         this._memoized.getTopFrecentSites(msg.data).then(links => {
-          this.send(am.actions.Response("TOP_FRECENT_SITES_RESPONSE", links), worker);
+          this.send(am.actions.Response("TOP_FRECENT_SITES_RESPONSE", links, {append}), worker);
         });
         break;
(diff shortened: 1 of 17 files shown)


pdehaan:
@k88hudson Needs a rebasin'.

lib/ActivityStreams.js
```diff
@@ -96,25 +96,26 @@ ActivityStreams.prototype = {
    */
   _respondToPlacesRequests(msgName, params) {
     let {msg, worker} = params;
+    const append = msg.meta && msg.meta.append;
```
oyiptong: The memoization key generator will not handle this correctly.

I would suggest to flatten the `msg` object by adding `append` as a parameter of `msg` and removing `meta` altogether, or to improve the cache key generation.

oyiptong:
The code looks OK to me.

The addon bits get an r+ if we take the simple route (using `msg.append` instead of `msg.meta.append`). If we want to improve the cache key generator, we'll need to do more work, and 
I'll take another review when it's ready.

I took a look at the content code and while it looks good to me, I have to admit I'm not familiar with React's conventions and how it will fit in a bigger app. I'll defer to @rlr to review the content side of things.

lib/ActivityStreams.js
oyiptong: Actually, I remove my comment. `meta` is not sent as a parameter, so query caching will work correctly

oyiptong:
I took a look at the code, and it looks like `msg.meta` is fine. we only cache based on `params`. The addon code looks good.

k88hudson:
if @rlr isn't available, maybe @pdehaan you could take a look?

oyiptong:
r+

Issue #314, opened 2016-03-16
Time to remove the subtitles inside Top Activity 


bryanbell:
The subtitles in Top Activity are only somewhat helpful, but usually visually dilute the section title since they're exactly the same size and weight. Top Activity is often closely followed by a day of the week header. Let's just drop that. 

<img width="888" alt="pasted_image_3_16_16__3_09_pm" src="https://cloud.githubusercontent.com/assets/228979/13830469/1357c96e-eb8a-11e5-8492-6c81b92dab14.png">

Let's instead try the horizontal rules in the mockup. 

<img width="788" alt="mockup" src="https://cloud.githubusercontent.com/assets/228979/13830513/5b341922-eb8a-11e5-9018-f60a08295f09.png">

pdehaan:
Unrelated, but in @bryanbell's screenshot above, I see a broken favicon image in his Spotlight result for Slate magazine bookmark.
Not sure if we have any way of detecting that and failing a bit more gracefully than a broken image icon (such as a fallback image of a blank 1px image).

k88hudson:
To clarify, spec here is:

* use plain lines to divide sections
* Leave "Top activity" label at the top
* change the date display to be default settings for `moment.calendar()` (http://momentjs.com/docs/#/displaying/calendar-time/), so it will include the day as well

rlr:
landed :ship:

PR #323, opened 2016-03-17
fix(content) #314 remove date subtitles and show .calendar() on side

rlr:
R? @k88hudson is this right?

### DIFF
--- a/content-src/components/ActivityFeed/ActivityFeed.scss
+++ b/content-src/components/ActivityFeed/ActivityFeed.scss
@@ -1,3 +1,13 @@
+.grouped-activity-feed {
+  > div {
+    border-bottom: solid 1px $faintest-black;
+
+    &:last-child {
+      border-bottom: none;
+    }
+  }
+}
+
 .activity-feed {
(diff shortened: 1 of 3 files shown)


rlr:
@k88hudson updated as discussed ^^

content-src/components/ActivityFeed/ActivityFeed.scss
```diff
@@ -1,9 +1,14 @@
 .activity-feed {
+  border-bottom: solid 1px $faintest-black;
   font-size: 13px;
   line-height: 1.4;
   list-style: none;
   margin: 0;
   padding: 0;
+
+  &:last-child {
```
k88hudson: This isn't working because the `ul` is nested inside a div, you need to either apply it to the container or change the structure

rlr:
gah! fixed ^^

k88hudson:
@rlr R+!

pdehaan:
Here's what it currently looks like with some sparse browsing data:

activity_stream

PR #324, opened 2016-03-17
Closes #312 - Add before/after filters to getRecentBookmarks

### LINKED ISSUE #312: Add before/after filters to getRecentBookmarks
We need to add the `afterDate`, `beforeDate` options available for history items that @mzhilyaev implemented to `getRecentBookmarks` as well.

This is needed for #222, so we can load more bookmarks.

mzhilyaev:
FIxes #312


k88hudson:
R+, thanks for doing this so quick!

PR #326, opened 2016-03-17
Add placeholders for required payload attributes and send session_duration in milliseconds

emtwo:


test/test-TelemetrySender.js
```diff
@@ -20,9 +21,21 @@ const samplePing = JSON.stringify({
   unload_reason: "close",
   client_id: "b8abfda8-0c57-9b48-af8b-c5ef87b2b673",
   session_duration: 10
+
```
emtwo: nit: remove newline

lib/TelemetrySender.js
```diff
@@ -1,21 +1,45 @@
...
+// The preference that tells what locale the user selected
+const PREF_SELECTED_LOCALE = "general.useragent.locale";
+
 function TelemetrySender(pingEndpoint) {
   Services.obs.addObserver(this, "tab-session-complete");
   this._pingEndpoint = pingEndpoint;
 }
 
 TelemetrySender.prototype = {
+  get locale() {
```
oyiptong: use Locale.jsm instead: https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/Locale.jsm#20

lib/TelemetrySender.js
```diff
@@ -1,21 +1,45 @@
...
+      data = this._appendRequiredPingData(data);
       this._sendPing(data);
     }
   },
 
+  _appendRequiredPingData: function(data) {
+    let ping = JSON.parse(data);
+    ping.addon_version = self.version;
+    ping.locale = this.locale;
+    ping.load_reason = "other";
```
oyiptong: the idea is that the values will be correct once we collect them?

lib/TelemetrySender.js
```diff
@@ -1,21 +1,45 @@
...
   observe: function(subject, topic, data) {
     if (topic === "tab-session-complete") {
+      data = this._appendRequiredPingData(data);
       this._sendPing(data);
     }
   },
 
+  _appendRequiredPingData: function(data) {
+    let ping = JSON.parse(data);
+    ping.addon_version = self.version;
```
oyiptong: since this is unlikely to change, i suggest we move all of this data to TabTracker instead of requiring a deserialize/serialize

lib/TabTracker.js
```diff
@@ -51,7 +51,7 @@ TabTracker.prototype = {
       return;
     }
     if (this._tabData.start_time) {
-      this._tabData.session_duration = (Date.now() - this._tabData.start_time) / 1000;
+      this._tabData.session_duration = (Date.now() - this._tabData.start_time);
       delete this._tabData.start_time;
     }
```
oyiptong: IMO, it would make sense to add the tab metadata at this point

lib/TelemetrySender.js
emtwo: Yes. Just putting in these placeholders for now so the pings are not discarded by the server.

oyiptong:
r+

PR #338, opened 2016-03-21
fix(content): #337 Use white bg as default SiteIcon bg

### LINKED ISSUE #337: Incorrect background colors in Top Activity
Background colors in Top Activity don't seem to be using dominant colors and instead seem to be mostly random and aren't consistent internally or with Top Sites.

screen shot 2016-03-21 at 10 26 51 am

k88hudson:
Fixes #337

### DIFF
--- a/content-src/selectors/selectors.js
+++ b/content-src/selectors/selectors.js
@@ -7,6 +7,7 @@ const urlParse = require("url-parse");
 const SPOTLIGHT_LENGTH = module.exports.SPOTLIGHT_LENGTH = 3;
 const TOP_SITES_LENGTH = module.exports.TOP_SITES_LENGTH = 6;
 const BACKGROUND_FADE = 0.5;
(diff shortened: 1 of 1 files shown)


pdehaan:
@k88hudson But this won't use the same logic as "Top Sites" to choose a background-color though, correct?
For example, in @nchapman's screenshot in #337, the Top Sites shows a Twitter "chip" with a slight blue background color and Twitter logo. But in the Top Activity section it's a red background (similarly, GitHub uses two different favicon background colors). With this PR, it should choose an almost-white background color if a favicon is found (but not a blue background in the Twitter scenario).

pdehaan:
Unsolicited :+1:

Ignore my earlier comment. I tested this locally and it looks pretty awesome! Seeing one odd side effect on Facebook where half the time it is blue, and the rest of the time the background is white.

Here's some random browsing data:
activity_stream

PR #341, opened 2016-03-21
Closes #308 - Add load latency to telemetry ping

### LINKED ISSUE #308: Add load latency to telemetry ping
per bug #27

mzhilyaev:
Fixes #308

### DIFF
--- a/lib/TabTracker.js
+++ b/lib/TabTracker.js
@@ -10,6 +10,7 @@ Cu.import("resource://gre/modules/Locale.jsm");
 
 const TELEMETRY_PREF = "telemetry";
 const COMPLETE_NOTIF = "tab-session-complete";
+const PERF_LOG_COMPLETE_NOTIF = "performance-log-complete";
 
 function TabTracker(trackableURLs, clientID) {
   this._tabData = {};
@@ -17,13 +18,13 @@ function TabTracker(trackableURLs, clientID) {
   this._clientID = clientID;
   this._trackableURLs = trackableURLs;
   this.onOpen = this.onOpen.bind(this);
-  this.onLoadComplete = this.onLoadComplete.bind(this);
+  this.onPerfLoadComplete = this.onPerfLoadComplete.bind(this);
 
   this._onPrefChange = this._onPrefChange.bind(this);
   this.enabled = simplePrefs.prefs[TELEMETRY_PREF];
   if (this.enabled) {
     tabs.on("open", this.onOpen);
-    Services.obs.addObserver(this.onLoadComplete, "performance-log-complete");
+    Services.obs.addObserver(this.onPerfLoadComplete, PERF_LOG_COMPLETE_NOTIF);
   }
   simplePrefs.on(TELEMETRY_PREF, this._onPrefChange);
 }
@@ -43,7 +44,7 @@ TabTracker.prototype = {
       tab.removeListener("close", this.logDeactivate);
     }
     tabs.removeListener("open", this.onOpen);
-    Services.obs.removeObserver(this.onLoadComplete, "performance-log-complete");
+    Services.obs.removeObserver(this.onPerfLoadComplete, PERF_LOG_COMPLETE_NOTIF);
   },
 
   uninit() {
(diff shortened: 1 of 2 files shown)


lib/TabTracker.js
```diff
@@ -13,7 +13,9 @@ function TabTracker(trackableURLs, clientID) {
   this._clientID = clientID;
   this._trackableURLs = trackableURLs;
   this.onOpen = this.onOpen.bind(this);
+  this.onLoadComplete = this.onLoadComplete.bind(this);
   tabs.on("open", this.onOpen);
+  Services.obs.addObserver(this.onLoadComplete, "performance-log-complete");
```
oyiptong: nit: please use a `const` for consistency; an upstream version of the file uses `const`s instead of strings:
https://github.com/mozilla/activity-streams/blob/59806985ab1138fe9fbc83f352f3ab4d53cc7eb8/lib/TabTracker.js

lib/TabTracker.js
```diff
@@ -31,6 +33,7 @@ TabTracker.prototype = {
       tab.removeListener("close", this.logDeactivate);
     }
     tabs.removeListener("open", this.onOpen);
+    Services.obs.removeObserver(this.onLoadComplete, "performance-log-complete");
```
oyiptong: ditto

lib/TabTracker.js
```diff
@@ -121,6 +125,13 @@ TabTracker.prototype = {
     tab.on("deactivate", this.logDeactivate("unfocus"));
     tab.on("close", this.logClose("close"));
   },
+
+  onLoadComplete: function(subject, topic, data) {
```
oyiptong: nit: `onPerfLoadComplete`. `TabTracker` also measures engagement stats and the function name could be confusing

test/test-TabTracker.js
```diff
@@ -150,6 +154,47 @@ exports.test_TabTracker_reactivating = function*(assert) {
...
+            tabs.removeListener("pageshow", onOpen);
+          });
+        }, 10);
+      }
+      Services.obs.addObserver(onPageLoaded, "performance-log-complete");
+    };
+
+    tabs.on("pageshow", onOpen);
+
+    let Observer = {
```
oyiptong: nit: please use the function parameter for observers. it is more concise

oyiptong:
I'd like to see one more test: one where the load isn't completed and the tab is closed.
All data should be there except load latency.

You can do that by manufacturing a page that send almost all events... except for the one triggering `performance-log-complete`.

mzhilyaev:
@oyiptong 
thanks for the review :)
Regarding the last comment for the test that checks for latency absence if page load is no finished:
Take a look at https://github.com/mozilla/activity-streams/blob/master/test/test-TabTracker.js#L63
This test closes the tab session on tab ready, long before the NEWTAB_RENDER event arrives, and then it checks that ping consists only of the keys listed here: https://github.com/mozilla/activity-streams/blob/master/test/test-TabTracker.js#L14
Which doesn't mention latency, so U woudl think this should qualify for the test you want.
Let me know if this is sufficient.

test/test-TabTracker.js
```diff
@@ -143,6 +143,10 @@ exports.test_TabTracker_reactivating = function*(assert) {
   for (let i in pingData) {
     let ping = pingData[i];
     assert.equal(ping.unload_reason, unloadReasons[i], "Unloaded for the expected reason");
+    // the page may or may not had enough time to be completely loaded, hence load_latency may
```
oyiptong: The comment should read that the page will definitely not have the time to load. Otherwise we'll have an intermittent failure here.

Perhaps we need to load an artificial HTML file to guarantee this won't happen.

oyiptong:
I think the current test is fine, however, I'm anxious about intermittent issues later down the road.

One way to guarantee that NEWTAB_RENDER will never occur is to create a situation where it can't occur.

What are your thoughts?

mzhilyaev:
i think we are good with the existing test. It closes the tab right on the "pageshow" event, which should happen way-way-way before content done rendering.  If NEWTAB_RENDER arrives before or on  "pageshow" the test should break - something is definitely screwed.

oyiptong:
r+ with one nit

test/test-TabTracker.js
```diff
@@ -213,7 +227,7 @@ exports.test_TabTracker_latency = function*(assert) {
   let pingData;
```
mzhilyaev: I think the nit is fixed in this commit.  Unless i missed yet another nit

Issue #346, opened 2016-03-22
Add load_reason.


emtwo:

pdehaan:
@emtwo Was this fixed/resolved by #347?

PR #347, opened 2016-03-22
feat(addon): #346 Add load_reason.

emtwo:


lib/TabTracker.js
```diff
@@ -65,6 +64,7 @@ TabTracker.prototype = {
       // didn't even load yet. Let's say it's been active for 0 seconds.
       this._tabData.url = tab.url;
       this._tabData.tab_id = tab.id;
+      this._tabData.load_reason = this._tabData.load_reason || "other"; // Page didn't load at all.
```
oyiptong: if the page didn't load, we should be explicit. "no_load" or whatnot

oyiptong: maybe "none"?

test/test-TabTracker.js
```diff
@@ -16,6 +16,37 @@ const EXPECTED_KEYS = ["url", "tab_id", "session_duration", "client_id", "unload
 let ACTIVITY_STREAMS_URL;
 let app;
 
+function createPingSentPromise(pingData, expectedPingCount) {
+  return new Promise(resolve => {
+    let Observer = {
+      observe: function(subject, topic, data) {
```
oyiptong: nit: the code can be made simpler by just passing a named function to the observer service.

note that it has to be a named function:

```javascript
function someFunc(subject, topic, data) {
  ...
  Services.obs.removeObserver(someFunc, someNotif);
}
Services.obs.addObserver(someFunc, someNotif);

oyiptong:
r+ with nits

PR #351, opened 2016-03-23
fix(build): Add common dir to webpack

k88hudson:
This doesn't actually cause a bug but prevents one from happening in the future


pdehaan:
:+1: LGTM. Only seems to add ~1KB to the bundle.js file.

```
           Asset    Size  Chunks             Chunk Names
vendor.bundle.js  308 kB       0  [emitted]  vendor
       bundle.js  264 kB       1  [emitted]  app
   [0] multi vendor 52 bytes {0} [built]
    + 359 hidden modules
```

PR #361, opened 2016-03-23
fix(ActivityFeed): Make date/time formats consistent

### LINKED ISSUE #343: Date headers are slightly inconsistent
Ref: #323

It looks like a small edge case where we're formatting dates slightly inconsistently based on whether it's the first item in a group. Since that was a garbage explanation, here's a free screenshot:

activity_stream
**Figure 1:** "2:11PM" versus "2:01 pm"

pdehaan:
FIxes #343


pdehaan:
activity_stream

k88hudson:
R+

PR #364, opened 2016-03-24
Closes #309 - Add total_bookmarks and total_history_size to telemetry ping

### LINKED ISSUE #309: Add total_bookmarks and total_history_size to telemetry ping
will require two separate queries against places.

mzhilyaev:
* added PlacesProvider API to get history and bookmarks sizes
  * added corresponding queries to memoizer cache
  * added notification data containing history and bookmarks sizes to "activity-streams-places-cache-complete" event
  * added "activity-streams-places-cache-complete" observer to TabTracker which updates appropriate keys inside tabData
  * a bunch of test changes 

Fixes #309

### DIFF
--- a/lib/TabTracker.js
+++ b/lib/TabTracker.js
@@ -12,12 +12,12 @@ const TELEMETRY_PREF = "telemetry";
 const COMPLETE_NOTIF = "tab-session-complete";
 const PERF_LOG_COMPLETE_NOTIF = "performance-log-complete";
 
-function TabTracker(trackableURLs, clientID, placesGate) {
+function TabTracker(trackableURLs, clientID, placesQueries) {
   this._tabData = {};
 
   this._clientID = clientID;
   this._trackableURLs = trackableURLs;
-  this._placesGate = placesGate;
+  this._placesQueries = placesQueries;
   this.onOpen = this.onOpen.bind(this);
   this.onPerfLoadComplete = this.onPerfLoadComplete.bind(this);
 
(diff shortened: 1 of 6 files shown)


lib/TabTracker.js
```diff
@@ -45,6 +52,16 @@ TabTracker.prototype = {
     }
     tabs.removeListener("open", this.onOpen);
     Services.obs.removeObserver(this.onPerfLoadComplete, PERF_LOG_COMPLETE_NOTIF);
+    Services.obs.removeObserver(this.onCacheUpdate, CACHE_UPDATE_NOTIF);
```
oyiptong: If we listen to the cache update notification, we would only obtain data if there is a cache update. This is incorrect: we want the history and bookmark sizes with or without cache updates.

A better way to obtain this data is to get ahold of the memoized query object instance and retrieving the data from there

mzhilyaev: Well,  we UPDATE our values on cache updates.  So there's always something in the tabData, right?
And on the cache update, that something changes to the the current history and bookmark size, which seems reasonably OK.   Now, this indeed relies on the fact that we update cache every time something places change happens, but .....   is this a problem?

oyiptong: A good side effect of this is that it makes for clearer and more explicit code.

Should we start needing database queries in other modules than `ActivityStreams.js`, it makes sense to create a `PlacesQueryService` or similarly named object which will be exported as a singleton instance whose purpose is to keep track of caching Places queries. It could even be exported from `PlacesProvider`.

Thoughts?

oyiptong: Not having the cache turned on will result in the data not being populated. It makes for an non-obvious coupling with an unrelated pref state, even if in all likelihood, that pref will not be tampered with.

We want to minimize such couplings.

lib/TabTracker.js
```diff
@@ -12,19 +12,19 @@ const TELEMETRY_PREF = "telemetry";
 const COMPLETE_NOTIF = "tab-session-complete";
 const PERF_LOG_COMPLETE_NOTIF = "performance-log-complete";
 
-function TabTracker(trackableURLs, clientID) {
+function TabTracker(trackableURLs, clientID, placesGate) {
```
oyiptong: nit: placesQueries might be more suggestive

oyiptong:
r+ with one nit

PR #368, opened 2016-03-24
chore(tests): Better fake data for tests

### LINKED ISSUE #212: Create better fake data test utils for content-tests
Right now it's really hard to update them if the data model changes.

When we do this, we should expand tests for selectors that landed in #220

k88hudson:
This will allow us to do more granular tests on the content side, still working on filling out the tests, need to publish tippy-top to npm, etc.

Fixes #212


content-test/faker.js
```diff
@@ -0,0 +1,179 @@
...
+    Object.assign(site, options.override);
+  }
+
+  // Add images if options.images > 0
+  if (options.images) {
+    site.images = arrayOf(createImage, options.images);
+  }
+
+  // Only create favicon_colors if there is a favicon_url
+  if (site.favicon_url && tiptopSite.background_color_rgb) {
```
pdehaan: Bad news, we lost `background_color_rgb[]` for now in the tippy-top-sites data set (per https://github.com/nchapman/tippy-top-sites/pull/19). We'll have to convert the hex to RGB ourselves from this end for now, and add a `// TODO` to change the code once we've locked down tippy-top-sites with near final data and we can put the color array back there.

Probably easiest just to do something like this pseudo-silliness near the top of the file:

```js
const hexToRgb = require("hex-to-rgb");

tiptop = tiptop.map((site) => {
  site.background_color_rgb = hexToRgb(site.background_color);
  return site;
});
```

k88hudson: Ah ok, noted. Perhaps we can add a programmatic API to tippy-top

pdehaan: @k88hudson: Great idea! What would you need/want for an API?

If it's just to dynamically inject the `rgb[]` color array, it may be as easy as adding a "server.js" with the following (and then tweaking the main entry point in package.json):

```js
var hexToRgb = require("hex-to-rgb");

var data = require("./top_sites.json");

module.exports = data.map(function (site) {
  site.background_color = site.background_color.toUpperCase();
  site.background_color_rgb = hexToRgb(site.background_color);
  return site;
});
```

Or if you want a base64 version of the `image_url` that should be really simple too (although will add a considerable amount of weight to the JSON payload).

pdehaan:
LGTM, apart from my one typo nit. And you may want to rename bug title above to remove "work in progress", and then squash commits and do a "fix(data): ..." commit message for our changelog gods.

I'll let @rlr give the final :eyes: and :+1:/:-1:.

content-test/components/ActivityFeed.test.js
```diff
@@ -112,8 +113,14 @@ describe("ActivityFeedItem", function() {
 describe("GroupedActivityFeed", function() {
   let instance;
   let el;
+  let sites;
   beforeEach(() => {
-    instance = TestUtils.renderIntoDocument(<GroupedActivityFeed sites={fakeSites} />);
+    sites = [
+      faker.createSite({moment: faker.moment()}),
+      faker.createSite({moment: faker.moment().subtract(2, "days")}),
+      faker.createSite({moment: faker.moment().subtract(4, "days")}),
```
rlr: super cool! :+1:

rlr:
this is sweet. lgtm!

pdehaan:
@k88hudson: Does this fix #212?

content-test/faker.js
k88hudson: Oops! Sorry :stuck_out_tongue_winking_eye:

k88hudson:
Yes!

PR #372, opened 2016-03-28
fix(content): #352 Renamed Spotlight and Top Activity

### LINKED ISSUE #352: Rename Spotlight and Top Activity
We need to rename the two lower sections of the new tab page.

**Rename Spotlight -> Featured**

Avoids any issues with Apple's trademark.

**Rename Top Activity -> Recent Activity**

It is actually recent activity and not top activity. I think it will avoid confusion. We can revisit this again when we have the idea of "Top".

nchapman:
Renamed Spotlight to Featured and Top Activity to Recent Activity. Both appear to be showing up correctly on new tab and the timeline.

Fixes #352.

@pdehaan @k88hudson r?

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -39,7 +39,7 @@ const NewTabPage = React.createClass({
           </section>
 
           <section>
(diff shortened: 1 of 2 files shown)


pdehaan:
LGTM. I'll let @k88hudson give the final :eyes: and :+1:.

PR #424, opened 2016-03-30
fix(content): #422 disable menu on header

### LINKED ISSUE #422: Disable dropdown menu on Timeline
We don't need the dropdown menu for now, although we will need it in the future.

k88hudson:
We don't need the dropdown for now since it doesn't do much, but we will in the future.

Fixes #422


k88hudson:
@rlr R?

rlr:
works for me! tests pass. r+

PR #429, opened 2016-03-31
Experimental localization support with L20n

stasm:
I was able to get L20n to work with Activity Streams.

stream1
stream2

This is highly experimental and part of my exploration into what it would take to use L20n in Firefox Desktop with React.  In this attempt, I'm simply reusing L20n's Mutation Observer to get notified when React re-renders elements with `data-l10n-id` and `data-l10n-args` attributes.

I've outlined other possible approaches in https://github.com/stasm/l20n-react-experiments/ with parts of the conversation happening in https://groups.google.com/d/msg/mozilla.tools.l10n/XtxHgBEokCA/onHthNvtBgAJ.  This PR is an invitation to join the discussion :)

cc @zbraniecki @Pike

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/429)

### DIFF
--- a/bin/generate-html.js
+++ b/bin/generate-html.js
@@ -15,6 +15,10 @@ function template(rawOptions) {
     <title>${options.title}</title>
     <link rel="stylesheet" href="${options.baseUrl}main.css" />
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
+
+    <meta name="defaultLanguage" content="en-US">
+    <meta name="availableLanguages" content="en-US, pl">
+    <link rel="localization" href="./locales/{locale}/app.properties">
   </head>
   <body>
     <div id="root"></div>
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -3,7 +3,7 @@ const {connect} = require("react-redux");
 const {justDispatch} = require("selectors/selectors");
 const {actions} = require("common/action-manager");
 const SiteIcon = require("components/SiteIcon/SiteIcon");
-const {prettyUrl, getRandomFromTimestamp} = require("lib/utils");
+const {prettyUrl, getRandomFromTimestamp, l10n} = require("lib/utils");
 const moment = require("moment");
 const classNames = require("classnames");
 
@@ -40,8 +40,11 @@ const ActivityFeedItem = React.createClass({
 
     let dateLabel = "";
     if (date && this.props.showDate) {
+      // TODO how to implement this in L20n?
+      // See http://momentjs.com/docs/#/displaying/calendar-time/
       dateLabel = moment(date).calendar();
     } else if (date) {
+      // TODO use Intl's DateTimeFormat
       dateLabel = moment(date).format("h:mm A");
     }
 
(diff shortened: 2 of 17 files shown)


bin/generate-html.js
```diff
@@ -15,6 +15,10 @@ function template(rawOptions) {
     <title>${options.title}</title>
     <link rel="stylesheet" href="${options.baseUrl}main.css" />
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
+
+    <meta name="defaultLanguage" content="en-US">
+    <meta name="availableLanguages" content="en-US, pl">
+    <link rel="localization" href="./locales/{locale}/app.properties">
```
stasm: I tried using `${options.baseUrl}` but got CORS errors.

content-src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -40,8 +40,11 @@ const ActivityFeedItem = React.createClass({
 
     let dateLabel = "";
     if (date && this.props.showDate) {
+      // TODO how to implement this in L20n?
+      // See http://momentjs.com/docs/#/displaying/calendar-time/
       dateLabel = moment(date).calendar();
```
stasm: @zbraniecki Any thoughts on how to implement this in L20n/Intl?

content-src/components/Spotlight/Spotlight.js
```diff
@@ -76,7 +93,7 @@ const Spotlight = React.createClass({
       blankSites.push(<li className="spotlight-item spotlight-placeholder" key={`blank-${i}`} />);
     }
     return (<section className="spotlight">
-      <h3 className="section-title">Featured</h3>
+      <h3 data-l10n-id="spotlight-title" className="section-title">Featured</h3>
```
stasm: Here and in other places:  I didn't remove the original English strings from the JSX for now, but we should consider doing it.

For one, if kept around they tend to get out of sync pretty quickly.  Also, React renders a number of `<span>` elements around text content (especially when you interpolate expressions in `{…}`) to help with the reconciliation process.  These `<span>` elements are currently removed when L20n inserts the translation into the parent element, so they're don't serve their purpose at all and aren't needed.

There might be alternative approaches to integrating L20n with React which would operate on the virtual DOM instead of the actual DOM and thus preserve whatever React ends up rendering.

content-src/lib/utils.js
```diff
@@ -79,5 +79,18 @@ module.exports = {
   getRandomFromTimestamp(percent, site) {
     const n = (site.lastVisitDate || site.bookmarkDateCreated || Math.round(Math.random() * 100)) % 100;
     return n <= percent * 100;
+  },
+
+  // before: data-l10n-id={this.props['data-l10n-id']}
+  // after:  {...l10n(this.props)}
+  l10n(props) {
```
stasm: I wrote this helper because I found myself typing `data-l10n-id={this.props['data-l10n-id']}` a lot.  It works well because components in Activity Streams are well designed and generally one component only needs a single `data-l10n-id` to be passed.

webpack.config.js
```diff
@@ -62,7 +63,8 @@ module.exports = {
       "selectors": absolute("./content-src/selectors"),
       "lib": absolute("./content-src/lib"),
       "strings": absolute("./strings"),
-      "test": absolute("./content-test")
+      "test": absolute("./content-test"),
+      "l20n": absolute("./node_modules/l20n/dist/compat/web/l20n.js")
```
stasm: I had to use the `compat` variant which is writtern in es5 because uglifyjs doesn't like some of the es6 syntax from `dist/bundle/web/l20n.js`.

k88hudson:
This is awesome! :star: :star: :star: Sorry, we're a bit swamped getting this cycle's release out, will take a closer look at this soon.

webpack.config.js
k88hudson: I know, super annoying. :( we could also try specifically excluding it from the optimizer 

Can't wait til that harmony branch lands in uglify...

PR #432, opened 2016-03-31
Closes #430 - Add page identifier to tab-session ping

### LINKED ISSUE #430: Add page identifier to tab-session ping
see bug #358

mzhilyaev:
Fixes #430


k88hudson:
R+

I think we should consider using a constants file to ensure we don't mess up our strings, but we can do that in a future patch

Issue #83, opened 2016-02-10
Replace font-awesome with actual icons


k88hudson:
Right now we're using font-awesome via the cdn, let's replace with real icons;

Should we use 

a) A subset of font-awesome?
b) Custom icon svgs?

If so, we should get final specs/assets for icons

pdehaan:
Potentially useful: https://www.npmjs.com/package/react-svg-use

pdehaan:
Interesting SVG post from GitHub: https://github.com/blog/2112-delivering-octicons-with-svg

pdehaan:
I found the following FontAwesome [references in the code](https://github.com/mozilla/activity-streams/search?utf8=%E2%9C%93&q=fa):

| ICON | FILE |
|----|----|
| `fa-search` | [/content-src/components/Search/Search.js:17](https://github.com/mozilla/activity-streams/blob/a7f93917b357379b5c3c4f195439f978159e26e5/content-src/components/Search/Search.js#L17)
| `fa-arrow-right` | [/content-src/components/Search/Search.js:20](https://github.com/mozilla/activity-streams/blob/a7f93917b357379b5c3c4f195439f978159e26e5/content-src/components/Search/Search.js#L20)
| `fa-firefox` | [/content-src/components/TimelinePage/TimelinePage.js:11](https://github.com/mozilla/activity-streams/blob/728749aec929a441ac455672f0f2e8face945fcf/content-src/components/TimelinePage/TimelinePage.js#L11)
| `fa-star` | [/content-src/components/TimelinePage/TimelinePage.js:12](https://github.com/mozilla/activity-streams/blob/728749aec929a441ac455672f0f2e8face945fcf/content-src/components/TimelinePage/TimelinePage.js#L12)
| `fa-timeline` | [/content-src/components/TimelinePage/TimelinePage.js:17](https://github.com/mozilla/activity-streams/blob/728749aec929a441ac455672f0f2e8face945fcf/content-src/components/TimelinePage/TimelinePage.js#L17)
| `fa-chevron-down` | [/content-src/components/Header/Header.js:19](https://github.com/mozilla/activity-streams/blob/728749aec929a441ac455672f0f2e8face945fcf/content-src/components/Header/Header.js#L19)

pdehaan:
In other _"@pdehaan never shuts up about SVG news"_...
[...]

PR #435, opened 2016-04-01
fix (content): #83 replace font awesome with svg images

rlr:
r? @k88hudson 

I was unable to get Firefox to set the fill color from our CSS, so I set it in the stylesheet within each SVG.


k88hudson:
Thanks for doing this, R+

We can look into optimizing after

PR #441, opened 2016-04-01
chore(build): use npm for tippy-top-sites

k88hudson:
Now that we've published to the registry, might as well use it


pdehaan:
:+1:

Unrelated: we should possibly see about fixing the unparsed HTML in the npm page: https://www.npmjs.com/package/tippy-top-sites

PR #464, opened 2016-04-05
fix(caching): make preview cache repopulation deterministic

oyiptong:
#304 introduced a bug where cache rebuilds would be non-deterministic: the function triggered asynchronous tasks but didn't wait for them to complete before launching the next task


oyiptong:
partly fixes #466

rlr:
LGTM and WFM. Doesn't fix the bug I was looking at so I'll take it from here. r+ :+1:

Issue #393, opened 2016-03-30
Verify fill in behavior after deleting spotlight items (as a last resort include mozilla tiles)


pdehaan:
Ref: #389

Not sure how this differs from #389

nchapman:
> Not sure how this differs from #389

This has to do with filling in more items as you delete them rather than just making sure they're filled in to begin with.

nchapman:
We should continue to fill in Top Sites and Highlights as you delete them rather than showing empty boxes.

PR #476, opened 2016-04-06
Block instead of delete for spotlight, top sites

k88hudson:
This changes delete for spotlight, top sites to a "block" strategy and warns that it's not yet complete.

Requires a follow-up for filling in empty boxes after deleting, https://github.com/mozilla/activity-streams/issues/393

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/476)

### DIFF
--- a/common/action-manager.js
+++ b/common/action-manager.js
@@ -13,6 +13,7 @@ const am = new ActionManager([
   "RECENT_LINKS_RESPONSE",
   "FRECENT_LINKS_REQUEST",
   "FRECENT_LINKS_RESPONSE",
+  "BLOCK_URL",
   "NOTIFY_HISTORY_DELETE",
   "NOTIFY_PERFORM_SEARCH",
   "RECEIVE_CURRENT_ENGINE",
@@ -20,7 +21,6 @@ const am = new ActionManager([
   "SEARCH_STATE_RESPONSE",
   "NOTIFY_ROUTE_CHANGE",
   "NOTIFY_PERFORMANCE",
-  "NOTIFY_TELEMETRY",
   "NOTIFY_USER_EVENT"
 ]);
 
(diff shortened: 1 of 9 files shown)


k88hudson:
R? @rlr @pdehaan

content-src/selectors/selectors.js
```diff
@@ -36,14 +36,24 @@ module.exports.selectNewTabSites = createSelector(
...
+    selectSpotlight,
+    state => state.Blocked
   ],
-  (TopSites, FrecentHistory, History, Spotlight) => {
-    let [topSitesRows, spotlightRows] = dedupe.group([TopSites.rows.slice(0, TOP_SITES_LENGTH), Spotlight.rows]);
+  (TopSites, FrecentHistory, History, Spotlight, Blocked) => {
+
+    // Remove duplicates
+    let [topSitesRows, spotlightRows] = dedupe.group([TopSites.rows.slice(0, TOP_SITES_LENGTH), Spotlight.rows]).map(sites => {
+      return sites.filter(site => !Blocked.urls.has(site.url));
```
rlr: shouldn't you be removing the blocked sites before deduping so that you end up with the right number of sites in each group? or am I reading this wrong? To me it looks like if one of the TopSites is blocked, it will end up having less than TOP_SITES_LENGTH?

rlr:
I have that one question but otherwise things look good. I'm gonna play with it locally and see how it behaves.

rlr:
Tested it... When I remove something from top sites, I kind of expected something from below to be promoted up to top sites to fill in the blank square, but this didn't happen. I think it's related to the order in which you do the filtering (I think filtering out blocked sites should be happen before anything else). But given this is still not implemented for real, I'm not opposed to just shipping as is for now. :+1:

k88hudson:
@rlr yeah i think the problem is that we only ask for exactly 6 top sites (it's a different query). I'm going to address this in #393 since it's kind of complicated how this has to be done

content-src/selectors/selectors.js
k88hudson: :+1:

common/action-manager.js
```diff
@@ -104,6 +105,14 @@ function RequestSearchState() {
   return RequestExpect("SEARCH_STATE_REQUEST", "SEARCH_STATE_RESPONSE");
 }
 
+function BlockUrl(url) {
+  alert("We're still working on this feature. Thanks for your patience!");
```
pdehaan: :sob: My kingdom for our beautiful modal dialogs.

rlr:
:shipit: :rocket:

common/action-manager.js
k88hudson: :soon:

PR #481, opened 2016-04-07
Closes #475: Fix main thread CPU pegging

### LINKED ISSUE #475: Blocked main thread during async task
There is a bug since 1.0.7 where an async task is causing Firefox to peg CPU at 100% utilization on 1 thread or more (including the main thread).

It seems caused by an async task, potentially one introduced in #304

oyiptong:
This fixes the main thread CPU pegging issue.

Closes #475 
r? @rlr

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/481)


lib/ActivityStreams.js
```diff
@@ -118,31 +112,31 @@ ActivityStreams.prototype = {
...
+          this._previewProvider.asyncSaveNewLinks(processedLinks);
           const cachedLinks = this._previewProvider.getCachedLinks(processedLinks);
           this.send(am.actions.Response("TOP_FRECENT_SITES_RESPONSE", cachedLinks, {append}), worker);
         });
         break;
       case am.type("RECENT_BOOKMARKS_REQUEST"):
         this._memoized.getRecentBookmarks(msg.data).then(links => {
           let processedLinks = this._previewProvider.processLinks(links);
-          this._previewProvider.saveNewLinks(processedLinks);
+          this._previewProvider.asyncSaveNewLinks(processedLinks);
```
rlr: :+1: +1000 for the `async...()`

rlr: (there probably should be a yield in front of this right? which is the bug I was looking at)

oyiptong: Probably a `then` which encloses whatever needs to follow sequentially.

And yes, I didn't fix it because you want to take it on, right?

lib/ActivityStreams.js
```diff
@@ -290,47 +281,69 @@ ActivityStreams.prototype = {
...
+
+      yield Promise.all(promises);
+      yield this._previewProvider.asyncSaveNewLinks(linksToSend);
+      this._populatingCache.preview = false;
+      Services.obs.notifyObservers(null, "activity-streams-previews-cache-complete", null);
+    }
   }),
 
   /**
+   * Set up preview cache to be repopulated every 6 hours
```
rlr: This is only called from init. I would expect the timeout to be cleared and restarted every time the cache is populated.

oyiptong: do you mean `this._previewCacheTimeoutID` ?

oyiptong: The cache population function and the timeout have been decoupled in part so that they can be unit tested

rlr: yeah. I would expect that to be cleared out every time the cache gets built. Then this function called again to reset it. Also, as it is right now, it gets set up once and never gets set again after being triggered.

rlr: (so it really isn't periodic as it says it is)

rlr: then it could work off of the `Services.obs.notifyObservers(null, "activity-streams-previews-cache-complete", null);` event

lib/ActivityStreams.js
```diff
@@ -290,47 +281,69 @@ ActivityStreams.prototype = {
...
+   */
+  _startPeriodicBuildPreviewCache(previewCacheTimeout) {
+    if (previewCacheTimeout) {
+      // only set a timeout if a non-null value is provided otherwise this will
+      // effectively be an infinite loop
+      this._previewCacheTimeoutID = setTimeout(() => {
+        this._asyncBuildPreviewCache();
+        this._startPeriodicBuildPreviewCache(previewCacheTimeout);
```
rlr: oh duh, i see it here lol

lib/ActivityStreams.js
rlr: nvm... then.. .i guess it could be fancier by resetting off the cache-complete event. but whatevs. it seems like we will be constantly rebuilding cache anyway while the user is active.

oyiptong: It is periodic: there is a recursive call at line 341. We also test this: https://github.com/mozilla/activity-streams/pull/481/files#diff-afc7a0b5d5d8bab654f47ee7ac5fd5b7R68

oyiptong: Oh, that test is a misnomer, it's not an invalidation, it's a repopulation. I'll rename it

lib/ActivityStreams.js
```diff
@@ -443,22 +456,31 @@ ActivityStreams.prototype = {
...
+      this._appURLHider.uninit();
+      this._perfMeter.uninit();
+      this._memoizer.uninit();
+      this._populatingCache.places = false;
+      this._populatingCache.preview = false;
+    };
+
     switch (reason){
       // can be one of: uninstall/disable/shutdown/upgrade/downgrade
+      case "uninstall":
```
rlr: this will never get called https://bugzilla.mozilla.org/show_bug.cgi?id=627432

rlr: specifically comment 12 https://bugzilla.mozilla.org/show_bug.cgi?id=627432#c12

You might want to do `disable` instead?

oyiptong: Oh! I didn't know "uninstall" is never called. Thank you!

test/test-ActivityStreams-async-preview-cleanup.js
```diff
@@ -0,0 +1,88 @@
...
+  previewsCachePromise = makeCachePromise("previews");
+  let app = new ActivityStreams({previewCacheTimeout: 100});
+  yield placesCachePromise;
+  yield previewsCachePromise;
+
+  let expectedInvalidations = 3;
+  let previewsCountPromise = makeCountingCachePromise("previews", expectedInvalidations);
+  let numInvalidations = yield previewsCountPromise;
+
+  assert.equal(numInvalidations, expectedInvalidations, "preview cache successfully invalidated periodically");
```
rlr: cool!

rlr:
awesome! sorry about the confusion. Async is hard before my 5th :coffee: :laughing: ... I'm going to play with it locally for sanity checking...

rlr:
travis broke with latest push ^

Seems to be working better now. Is it faster or is it me? r+ w/ travis fix

oyiptong:
hmm that failure is actually in karma. and it is intermittent. let me take a look

oyiptong:
@k88hudson says the intermittent issue may have been fixed in 6fdef5d76f9d629cc62be329d333c826b9577447.

I'll merge and close

PR #499, opened 2016-04-08
chore(content): #310 Add meta tag for CSP.

### LINKED ISSUE #310: Add CSP rules to content page
We need to see how restrictive we can get with CSP without breaking functionality.

This would help us prevent a lot of bad behavior from happening, and we still need to be able to run test automation with the rules.

At the minimum, we can set the CSP rules to be report-only.

emtwo:
Fixes #310 

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/499)

### DIFF
--- a/bin/generate-html.js
+++ b/bin/generate-html.js
@@ -11,6 +11,7 @@ function template(rawOptions) {
 <!doctype html>
 <html lang=en-us>
   <head>
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src http: data:; style-src 'unsafe-inline'">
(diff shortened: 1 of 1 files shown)


emtwo:
Some comments on this:

1. `default-src 'none'` blocks anything set for a `src` attribute. The following override it
2. `img-src` is `data` and `http` since the favicons can come from anywhere at the moment
3. `style-src` is `unsafe-inline` since the `style` attribute is set inline in multiple locations.
 
This is the most restrictive I was able to get the CSP tag without breaking anything.

oyiptong:
r+

PR #524, opened 2016-04-14
Closes #501 - Remove "source" property for tab session duration ping

### LINKED ISSUE #501: Remove "source" property for tab session duration ping
The "source" property only makes sense for the user action and is now handled by sending a user-event-ping as in https://github.com/mozilla/activity-streams/blob/master/data_dictionary.md#example-activity-stream-event-log.   Implemented in https://github.com/mozilla/activity-streams/pull/465

This make keeping "source" property in the tab-session ping meaningless, and it's always being set to "other".  Suggest removing it.
Need @ncloudioj  or @oyiptong , or @tspurway  OK on it.

mzhilyaev:
Fixes #501

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/524)


oyiptong:
r+

PR #530, opened 2016-04-15
Closes #311 - A performance logger that collects statistics

### LINKED ISSUE #311: A performance logger that collects statistics
Instead of just collecting and printing, the raw stats, we should collect the statistics and return mean and median data.

This would be a stepping stone for finer grained performance analysis.

We could observe observer notifications currently sent from `PerfMeter` to collect and build our data distribution.

mzhilyaev:
Fixes #311 

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/530)


oyiptong:
r+

Issue #389, opened 2016-03-30
Prevent empty UI in Spotlight, Top Sites


pdehaan:
Currently it is too easy to get empty Spotlight/Featured/Highlight boxes if you delete/hide items or in other cases.

This will be partially mitigated somewhat on a new profile via #200, but we should come up with an alternate solution like re-querying Firefox/Embedly on item delete, or in the case of not enough data, merging it with the default data set from #200.

nchapman:
It could be really interesting for us to use this [tool from embedly](http://embed.ly/tools/generator) to create a query that would show content results rather than "recent top sites".

nchapman:
Here's a rather epic (and obviously unoptimized) query based on the embedly tool with some additional news sites added for good measure:

```sql
SELECT DISTINCT
  p.id, p.url
FROM
  moz_historyvisits hv
INNER JOIN
  moz_places p ON p.id = hv.place_id
WHERE
  -- http://embed.ly/tools/generator -- all minus google maps
[...]

PR #531, opened 2016-04-18
New highlight query

oyiptong:
As requested in #389, this implements a new highlight query.

Given that we're likely to iterate rapidly on the query, there isn't thorough testing for the results of the query, just a sanity check.

I started implementing the UI integration, however, I found that the deduping logic yields potentially undesirable effects, resulting in empty spotlight spots. We'll have to file a follow-up bug to play around with the query and its integration overall.

r? @k88hudson, @mzhilyaev

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/531)

### DIFF
--- a/content-src/components/Base/Base.js
+++ b/content-src/components/Base/Base.js
@@ -15,6 +15,8 @@ const Base = React.createClass({
 
     this.props.dispatch(actions.RequestSearchState());
 
+    this.props.dispatch(actions.RequestHighlightsLinks());
+
     this.props.dispatch(actions.NotifyPerf("BASE_MOUNTED"));
   },
   render() {
--- a/content-src/selectors/selectors.js
+++ b/content-src/selectors/selectors.js
@@ -68,10 +68,11 @@ module.exports.selectNewTabSites = createSelector(
     state => state.TopSites,
     state => state.FrecentHistory,
     state => state.History,
+    state => state.Highlights,
     selectSpotlight,
     state => state.Blocked
   ],
-  (TopSites, FrecentHistory, History, Spotlight, Blocked) => {
+  (TopSites, FrecentHistory, History, Highlights, Spotlight, Blocked) => {
 
     // Removed blocked
     [TopSites, Spotlight] = [TopSites, Spotlight].map(item => {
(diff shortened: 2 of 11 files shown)


content-src/components/Base/Base.js
```diff
@@ -15,6 +15,8 @@ const Base = React.createClass({
 
     this.props.dispatch(actions.RequestSearchState());
 
+    this.props.dispatch(actions.RequestHighlightsLinks());
```
k88hudson: You should remove line 12,  `this.props.dispatch(actions.RequestFrecentLinks())`, since it's not being used anymore

content-src/selectors/selectors.js
```diff
@@ -68,10 +68,11 @@ module.exports.selectNewTabSites = createSelector(
     state => state.TopSites,
     state => state.FrecentHistory,
     state => state.History,
+    state => state.Highlights,
     selectSpotlight,
     state => state.Blocked
   ],
-  (TopSites, FrecentHistory, History, Spotlight, Blocked) => {
+  (TopSites, FrecentHistory, History, Highlights, Spotlight, Blocked) => {
```
k88hudson: You need to remove `FrecentHistory` here and add `Highlights` to the `selectSpotlight` instead (line 36)

Alternatively, leave this selector as is and file a follow-up ticket

oyiptong: That's what I did originally, however, it produced an inferior highlights UX.

I did file a follow-up bug so we can iterate on this feature: #533

lib/PlacesProvider.js
```diff
@@ -430,6 +447,89 @@ Links.prototype = {
...
+                               b.guid as bookmarkGuid,
+                               b.title as bookmarkTitle,
+                               b.lastModified / 1000 as lastModified,
+                               b.dateAdded / 1000 as bookmarkDateCreated
+                        FROM moz_places p
+                        INNER JOIN moz_bookmarks b
+                          ON b.fk = p.id
+                        LEFT JOIN moz_favicons f
+                          ON p.favicon_id = f.id
+                        WHERE date(b.dateAdded / 1000 / 1000, 'unixepoch') > date('now', '${HIGHLIGHTS_THRESHOLDS.created}')
```
k88hudson: Just confirming, this is supposed to be bookmarks in the last last three days, not older than last three days?

k88hudson:
There is some additional stuff here to use the new query including removing the old event FrecentLinks dispatch event and updating some more tests; the selectors also need a bit of work/tests updated.

If you want to do a follow-up ticket to do that, that's also OK

lib/PlacesProvider.js
oyiptong: Yes, the goal is to obtain a bookmark that has been created within the last 3 days

oyiptong:
The follow-up bug has been filed. I don't think we're ready to have this take over FrecentLinks just yet.

lib/PlacesProvider.js
k88hudson: 👍

PR #541, opened 2016-04-19
Fix some preview caching issues

oyiptong:
This PR:
* increases the TTL for previews to 30 days
* makes cleanups run lazily (allowing for long browser inactivity)
* adds a feature where links are updated periodically

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/541)

### DIFF
--- a/lib/PreviewProvider.js
+++ b/lib/PreviewProvider.js
@@ -3,7 +3,6 @@
 
 const {Cu} = require("chrome");
 const ss = require("sdk/simple-storage");
-const {setTimeout, clearTimeout} = require("sdk/timers");
 const simplePrefs = require("sdk/simple-prefs");
 const EMBEDLY_PREF = "embedly.endpoint";
 const ENABLED_PREF = "previews.enabled";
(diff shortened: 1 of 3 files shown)


lib/PreviewProvider.js
```diff
@@ -237,13 +279,17 @@ PreviewProvider.prototype = {
...
           if (!data) {
             return;
           }
-          ss.storage.embedlyData[link.cacheKey] = data;
-          ss.storage.embedlyData[link.cacheKey].accessTime = Date.now();
+          ss.storage.embedlyData[link.cacheKey] = Object.assign({}, ss.storage.embedlyData[link.cacheKey], data);
+          if (updateAccessTime) {
+            ss.storage.embedlyData[link.cacheKey].accessTime = currentTime;
+          }
+          ss.storage.embedlyData[link.cacheKey].refreshTime = currentTime;
```
rlr: was gonna ask about `accessTime` vs `refreshTime` but the tests make it obvious what's they are for 👍

lib/PreviewProvider.js
```diff
@@ -173,40 +181,74 @@ PreviewProvider.prototype = {
...
       } else {
         return null;
       }
     })
     .filter(link => link);
+
+    // gives the opportunity to have at least one run with an old
+    // preview cache. This is for the case where one hasn't opened
+    // the browser since `cacheTTL`
+    this.cleanUpCacheMaybe();
```
rlr: I assume this should never take too much time? It is blocking the returning of cached results. Any reason it wasn't done as part of the periodic updates instead?

rlr:
LGTM... I like the tests, they helped me understand it all. I ran it all locally to make sure nothing was obviously busted. WFM R+

lib/PreviewProvider.js
oyiptong: Yeah, I *hope* it won't take too much time.

That said, the periodic run has a chance to never run: if a user never keeps their browser open for longer than 24 hours, or whatever the timeout was.

One way to fix it would've been to run a cleanup job at startup, but that is also not desirable because a user may have not opened their browser for a long time.

We *could* solve the problem with timeouts, which would be better, but I thought we'd give this a try first.

PR #542, opened 2016-04-20
Always return links

oyiptong:
This PR contains commits that:
* ensure that links are always returned for history type queries, even if there are no previews available
* makes the cache cleanup job more tolerant to errors (better for tests)

r? @rlr

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/542)


rlr:
LGTM. R+!

Issue #483, opened 2016-04-07
Block/delete UI for ActivityFeed


k88hudson:
For this iteration we have decided we should add a small menu to the activity feed that allows users to remove (`NotifyHistoryDelete`) or block (`BlockUrl`) urls. We should also ask people to confirm when they delete something from their history permanently.

![image](https://cloud.githubusercontent.com/assets/1455535/14360446/70cd5b7c-fcc4-11e5-8975-b7f201d32ee6.png)

k88hudson:
Top sites should have label: `Remove from Top Sites`

PR #543, opened 2016-04-20
Gh483 - Delete

k88hudson:
This patch:

* adds a `ContextMenu` component (a little drop down with options)
* centralizes delete logic in `DeleteMenu` component that includes two options: 
     * Remove from History (this is actually deleting the item for history)
     * Never show this site (this is adding to the block list)

The delete menu has clearer language around what delete actually does, plus a confirmation alert for deleting from history. It also extends the  block list to Activity feed items as well.

Note that the block list is still in memory only, with an alert that says "we're still working on this feature"

See #483 for reference mocks.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/543)

### DIFF
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -3,6 +3,7 @@ const {connect} = require("react-redux");
 const {justDispatch} = require("selectors/selectors");
 const {actions} = require("common/action-manager");
 const SiteIcon = require("components/SiteIcon/SiteIcon");
+const DeleteMenu = require("components/DeleteMenu/DeleteMenu");
 const {prettyUrl, getRandomFromTimestamp} = require("lib/utils");
 const moment = require("moment");
 const classNames = require("classnames");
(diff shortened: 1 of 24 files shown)


k88hudson:
@rlr mind taking a look?

rlr:
on it! 👍

content-test/index.js
```diff
@@ -5,16 +5,20 @@ const {overrideConsoleError} = require("test/test-utils");
...
   let originalAlert;
+  let originalConfirm;
   before(() => {
     originalAlert = window.alert;
     window.alert = () => {};
     restore = overrideConsoleError(message => {
       throw new Error(message);
     });
+    originalConfirm = window.confirm;
+    window.confirm = () => true;
```
rlr: sneaky :)

content-src/components/ActivityFeed/ActivityFeed.js
```diff
@@ -70,15 +79,26 @@ const ActivityFeedItem = React.createClass({
         </div>
       </a>
       <div className="action-items-container">
-        <div className="action-item icon-delete" ref="delete" onClick={() => this.props.onDelete(site.url)}></div>
```
rlr: I see that `getDefaultProps()` still has an `onDelete` placeholder. Can remove now?

k88hudson: ah, good catch 👍

rlr:
Awesome stuff! I like that you broke out `ContextMenu` into it's own component...

Everything seems to be working awesome, except.... If you remove an item from histoy, it doesn't show up on our pages anymore, but it still shows up in the history menu and in AwesomeBar results. So it doesn't really seem to be deleting properly from everywhere. Or something? Is that expected?

I guess it would be a bug on the addon side and this pull is about the UI.

so R+ 🚀 🎸

rlr:
wait, scratch that... Now I can't repro what I just said... I think I may have had the tab for that url open still in the background.

PR #580, opened 2016-04-26
Continuous deployment

oyiptong:
This change in automation deploys to the bucket at https://moz-activity-streams-dev.s3.amazonaws.com/dist/latest.html

We'll know if the "master"-only branch works when we merge this PR

r? @k88hudson

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/580)

### DIFF
--- a/.jpmignore
+++ b/.jpmignore
@@ -4,6 +4,7 @@ bin/
 content-src/
 content-test/
 dist/
+artifacts/
 firefox/
 logs/
 node_modules/
--- a/bin/continuous-integration.sh
+++ b/bin/continuous-integration.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+if [ -z "$CONTINUOUS_INTEGRATION" ]
+then
+    set -x # sets bash debugging mode
+    echo "setting up development virtualenv"
+    export CFLAGS='-std=c99'
+fi
+
+fab deploy:dev_deploy=True || exit 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: node_js
 node_js:
   - "5.5.0"
 
+python:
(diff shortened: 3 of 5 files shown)


dist/latest/latest.html
```diff
@@ -0,0 +1,81 @@
+<html>
```
k88hudson: Might not be a good idea to put this in `dist` since it would complicate any cleaning scripts if we used them to clear out build directories. Not a problem right now but yeah if it's not hard I think it would be better to move it

bin/continuous-integration.sh
```diff
@@ -0,0 +1,22 @@
...
+if [ -z "$CONTINUOUS_INTEGRATION" ]
+then
+    set -x # sets bash debugging mode
+    echo "setting up development virtualenv"
+    export CFLAGS='-std=c99'
+fi
+
+echo "\nTRAVIS_BRANCH=$TRAVIS_BRANCH TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST"
+
+if [ -n "$TRAVIS_BRANCH" ] && [ -n "$TRAVIS_PULL_REQUEST" ] && [ "$TRAVIS_BRANCH" = "master" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]
```
k88hudson: don't you not need this anymore?

.jpmignore
```diff
@@ -32,3 +33,4 @@ npm-shrinkwrap.json
 requirements.txt
 webpack.config.js
 yamscripts.yml
+continuous-integration.sh
```
k88hudson: This is in bin, no?

bin/continuous-integration.sh
oyiptong: I thought so too, but I saw a build where the else condition triggered. wait up

oyiptong: https://travis-ci.org/mozilla/activity-streams/builds/125946931

That was most def not a merge build.

oyiptong: Oh. that was pre-travis deploy. Ok, i must've been mistaken. I'll remove the condition

k88hudson:
LGTM 👍

PR #588, opened 2016-04-27
chore (content): #310 Add meta tag for CSP

### LINKED ISSUE #310: Add CSP rules to content page

rlr:
r? @oyiptong 

This is a little bit more permissive but a lot better than nothing. Some of the directives are only for local dev and/or testing, so we could possibly look into having separate rules or just disable it altogether for dev/test. Thoughts?

Fixes #310 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/588)

### DIFF
--- a/bin/generate-html.js
+++ b/bin/generate-html.js
@@ -2,15 +2,20 @@
 "use strict";
 const defaults = {
   baseUrl: "",
-  title: "Loading..."
+  title: "Loading...",
+  csp: "on"
 };
 
 function template(rawOptions) {
   const options = Object.assign({}, defaults, rawOptions || {});
+  const csp = options.csp === "on" ?
+    "<meta http-equiv=\"Content-Security-Policy\" content=\"default-src 'none'; script-src 'self'; img-src http: https: data:; style-src 'self' 'unsafe-inline'\">" :
+    "";
   return `<!doctype html>
 <html lang="en-us">
   <head>
     <meta charset="utf-8">
+    ${csp}
     <title>${options.title}</title>
     <link rel="stylesheet" href="${options.baseUrl}main.css" />
     <link rel="icon" type="image/svg+xml" href="${options.baseUrl}img/newtab-icon.svg">
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "start:static": "npm run bundle:static -- -w",
     "start:webpack": "webpack -w",
     "start:css": "npm run bundle:css && npm run bundle:css -- -w",
-    "start:html": "npm run bundle:html -- --baseUrl http://localhost:1963/",
+    "start:html": "npm run bundle:html -- --baseUrl http://localhost:1963/ --csp off",
     "start:server": "live-server data/content --port=1963 --no-browser",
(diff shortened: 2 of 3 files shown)


pdehaan:
I think if we're doing some of the configs just for dev/test, it may make sense to pass something via a flag or option so we can toggle it so we have a tighter set of rules for prod.

But I'll defer to @oyiptong and @k88hudson

oyiptong:
We want the rule to be permissive enough to run our test automation. Using a different rule for production code is a possiblity. For instance, we could disallow unsafe-eval and inline scripts... if react doesn't use them.

@fmarier, how do you think we should approach this problem?

rlr:
Here is additional info on some of the "dangerous" things we are allowing:

`connect-src ws:;`: this is needed by the dev server for live reload. Tests don't need it. Packaged addon doesn't need it. We can probably make it even more specific specifying the hostname (`ws://localhost:1963`)

`script-src 'unsafe-eval';` this only seems to be needed for local dev (the `vendor.bundle.js` has a bunch of `evals` at the bottom). Tests pass without it and the bundled xpi works fine.

`script-src 'unsafe-inline';` this is needed for local dev only (the live-server injects js into the page). Tests pass and the bundled xpi works without it.

rlr:
That said, it almost makes sense to have a more restrictive CSP directive (without those `unsafe-*`) and just somehow not use it when developing on `http://localhost:1963`.

fmarier:
> @fmarier, how do you think we should approach this problem?

If you have a way to configure dev/test/prod differently and inject a different policy in the HTML, that would be ideal since you could a much strong policy as @rlr points out.

Another thing you may want to consider is to use `default-src 'none'` and whitelist the directives you need for `self` too.

oyiptong:
@rlr we could add the CSP directive in config.[ENV].yml files.

You mentioned that we might need `unsafe-eval` for the uglifier to work. Can we please verify?

@fmarier would the rules as implemented in this PR work, in view of more strict CSP for different environments implemented in a later bug?

rlr:
> You mentioned that we might need unsafe-eval for the uglifier to work. Can we please verify?

Yeah. I only found eval in the bundle that gets generated for dev. The packaged xpi didn't have any issues without the `unsafe-eval` directive. So I guess it isn't the uglifier that was putting those evals in.

> we could add the CSP directive in config.[ENV].yml files.

This is script generates a static template so I don't think it has access to any config files. I think we can pass in parameters to the script though. Might be tricky.

rlr:
@oyiptong ^^ updated with stricter policy and disabled for localhost dev. Fingers crossed for travis.

oyiptong:
r+

fmarier:
> @fmarier would the rules as implemented in this PR work, in view of more strict CSP for different environments implemented in a later bug?

That looks good to me. It would of course be nice if we didn't need `unsafe-inline` within the `style-src` directive but what is in this PR is way better than the defaults :)

PR #598, opened 2016-04-28
484 persist block list

oyiptong:
This provides the data backend as well as content API to block links.

It provides the ability to:
1. block a url
2. unblock a url (e.g. for an undo functionality)
3. unblock all urls

r? @k88hudson since you'll be the consumer of this API

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/598)

### DIFF
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -78,34 +78,46 @@ class BlockedURLs extends Set {
 
   /**
    * Add url and persist to pref
+   *
+   * @param {String} url a url to block
+   * @returns {Boolean} true if the item has been added
    */
   save(url) {
-    let oldSize = this.size;
-    this.add(url);
-    if (this.size !== oldSize) {
+    if (!this.has(url)) {
+      this.add(url);
       simplePrefs.prefs[this._prefName] = JSON.stringify(this.items());
+      return true;
     }
+    return false;
   }
 
   /**
    * Remove a url and persist to pref
+   *
+   * @param {String} url a url to unblock
+   * @returns {Boolean} true if the item has been remove
    */
   remove(url) {
-    let oldSize = this.size;
-    this.delete(url);
-    if (this.size !== oldSize) {
+    if (this.has(url)) {
+      this.delete(url);
       simplePrefs.prefs[this._prefName] = JSON.stringify(this.items());
+      return true;
     }
+    return false;
   }
 
   /**
    * Clear blocked url set and persist
+   *
+   * @returns {Boolean} whether or not blocklist was cleared
    */
   clear() {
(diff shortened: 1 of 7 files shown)


lib/PlacesProvider.js
```diff
@@ -50,6 +51,85 @@ const HIGHLIGHTS_THRESHOLDS = {
...
+   * Add url and persist to pref
+   *
+   * @param {String} url a url to block
+   * @returns {Integer} number of items added
+   */
+  save(url) {
+    let oldSize = this.size;
+    this.add(url);
+    let sizeDiff = this.size - oldSize;
+    if (sizeDiff > 0) {
```
k88hudson: Is there a reason you want to do this instead of `.has`?

lib/PlacesProvider.js
```diff
@@ -50,6 +51,85 @@ const HIGHLIGHTS_THRESHOLDS = {
...
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Return url set as an array ordered by insertion time
+   */
+  items() {
+    return [...this];
```
k88hudson: heh, that's a need way of converting to an array

lib/PlacesProvider.js
oyiptong: Ah! Good point. `has` is much better

lib/PlacesProvider.js
```diff
@@ -50,6 +51,85 @@ const HIGHLIGHTS_THRESHOLDS = {
...
+   * Remove a url and persist to pref
+   *
+   * @param {String} url a url to unblock
+   * @returns {Integer} number of items removed
+   */
+  remove(url) {
+    let oldSize = this.size;
+    this.delete(url);
+    let sizeDiff = oldSize - this.size;
+    if (sizeDiff > 0) {
```
oyiptong: ditto here

lib/PlacesProvider.js
```diff
@@ -204,6 +285,33 @@ Links.prototype = {
   },
 
   /**
+   * Blocks a URL
+   */
+  blockURL(url) {
+    if (this.blockedURLs.save(url) > 0) {
```
k88hudson: Personally I think something like returning a boolean/something like this would be more readable:
```js
const didChange = this.blockedURLs.save(url);
if (didChange) {
   this.emit...
}
```
but yeah, idk

oyiptong: I totally agree

k88hudson:
R+ after those nits 👍

k88hudson:
There's gonna be some happy people when this lands heh ;)

PR #616, opened 2016-05-02
feat (addon) #444 enhance sites with tippy top data

### LINKED ISSUE #444: Use tippy-top-sites as an Embedly pre-proxy?
Currently it looks like we're using the [**tippy-top-sites**](https://github.com/mozilla/activity-streams/search?utf8=%E2%9C%93&q=tippy-top-sites) module in 1 place, [content-test/faker.js](https://github.com/mozilla/activity-streams/blob/fc67ec4954c02f7684bcba980353e2030dadf88c/content-test/faker.js#L3), for our tests.

Since we can sometimes get suboptimal results from the Embed.ly proxy (pinterest, lowes, hackernews, etc), I'm not sure if it makes sense to try to use tippy-top-sites as a pre-proxy so we know we'll always have high quality favicons and background colors, site titles, etc, and only query embedly when necessary (such as the 99.9999% of the sites _not_ in our curated tippy-top-sites JSON).
I don't imagine there is much point in repeatedly querying youtube.com if we know we'll get back a poor quality favicon and not great titles, or whatever.

I'm sure there are several flaws with my logic (tippy-top-sites only cover top-level domains and aren't page specific, for example), so feel free to knock me down a few pegs.

Ref: #304
Ref: #384

rlr:
r? @oyiptong and/or @k88hudson

Fixes #444 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/616)


lib/TippyTopProvider.js
```diff
@@ -0,0 +1,67 @@
...
+   * Initialize the sites object
+   */
+  init() {
+    // If the sites were passed into the constructor (for testability), use those.
+    // Else, read them from top_sites.json.
+    let sites;
+    if (this.options.sites) {
+      sites = this.options.sites;
+    } else {
+      sites = JSON.parse(data.load("content/favicons/top_sites.json"));
```
pdehaan: Not sure if you want to touch the JSON file directly vs the `require("tippy-top-sites")` exported version via [index.js](https://github.com/nchapman/tippy-top-sites/blob/master/index.js) (since we inject a couple values on the fly). /cc @k88hudson

k88hudson: yeah i agree, i think it would be better to just required the file

lib/TippyTopProvider.js
```diff
@@ -0,0 +1,67 @@
...
+      site.background_color = tippyTopSite.background_color;
+    }
+    return site;
+  },
+
+  /**
+    * Create a key using the domain (minus the www.).
+    */
+  _getKey(url) {
+    let domain = URL(url).host;
```
k88hudson: I think we'll probably want to expand this to include all subdomains of a host as well, although that will result in some conflicts

lib/TippyTopProvider.js
```diff
@@ -0,0 +1,67 @@
...
+    */
+  processSite(site) {
+    let key;
+    try {
+      key = this._getKey(site.url);
+    } catch (e) {
+      key = null;
+    }
+
+    if (key && key in this._sitesByDomain) {
```
k88hudson: I would recommend creating and returning a new object instead of modifying `site` to avoid unintended side effects

lib/PreviewProvider.js
```diff
@@ -36,6 +38,7 @@ const DEFAULT_OPTIONS = {
 function PreviewProvider(options = {}) {
   this.options = Object.assign({}, DEFAULT_OPTIONS, options);
   this._onPrefChange = this._onPrefChange.bind(this);
+  this._tippyTopProvider = new TippyTopProvider();
```
k88hudson: technically you only need one instance of this right?

test/test-TippyTopProvider.js
```diff
@@ -0,0 +1,93 @@
...
+    sites: [{
+      url: "https://mozilla.org",
+      image_url: "mozilla-org.png",
+      background_color: "#fff"
+    }]
+  });
+  assert.deepEqual(site, tippyTopProvider.processSite(site));
+};
+
+exports["test TippyTopProvider processSite"] = function*(assert) {
```
k88hudson: can you add a test for a subdomain other than `www`? (even if it means no match for now)

k88hudson:
YAY! This is awesome. A couple of small notes in the code. For the issue with matching subdomains, I think it's fine to do that as a follow-up.

lib/TippyTopProvider.js
rlr: Yeah, I had talked about this with @oyiptong and we decided it was safer to match the full domain. I don't think there are that many cases in the top sites where we want to include all it's subdomains? I could be wrong.

lib/TippyTopProvider.js
rlr: How do I do that? Do I need to copy over that index.js into the addon data directory? Can I use it from there with require? Since it's in node_modules, wasn't sure how.

lib/TippyTopProvider.js
rlr: 👍

lib/PreviewProvider.js
rlr: hmm. yeah. I think in the long run we want to refactor this code to something more flexible to create a sort of pipeline to process the sites data. But are you suggesting I change something here?

rlr:
OK, I think I fixed the <strike>important</strike> easy things. I can file followups for 1) figuring out a better way to use the tippy tops package 2) look into if we want to match all subdomains.

lib/TippyTopProvider.js
pdehaan: I think it's just a simple `require("tippy-top-sites")`, per [./content-test/faker.js:3](https://github.com/mozilla/activity-streams/blob/fc67ec4954c02f7684bcba980353e2030dadf88c/content-test/faker.js#L3).

oyiptong: The issue with using require is that this is addon js. The `tippy-top-sites` module won't be available in the addon at runtime.

One way to get around this is to duplicate the module in a directory that will be packaged with the addon.

lib/PreviewProvider.js
oyiptong: Since `PreviewProvider` is a singleton, there will also be only one instance of `TippyTopProvider`. Since `PreviewProvider` is the only consumer, it makes sense to only instantiate it here IMO

lib/TippyTopProvider.js
pdehaan: Ah yes. This is a bit more verbose, but we can wire it up during the `npm run bundle` step:

```js
const writeFile = require("fs").writeFileSync;
const join = require("path").join;
const tippy = require("tippy-top-sites");

const OUT_FILE = join("data", "content", "favicons", "top_sites.json");

writeFile(OUT_FILE, JSON.stringify(tippy));
```

k88hudson:
Ok, looks good to me!

PR #619, opened 2016-05-03
use rimraf and mkdirp instead of shell #125

### LINKED ISSUE #125: Use rimraf instead of rm -rf
Currently we're using [`rm -rf data/content/*`](https://github.com/mozilla/activity-streams/blob/d083dabacb069703e71d5aaf35a25ff21a3a12cd/package.json#L84) in the package.json file (via [./yamscripts.yml:11](https://github.com/mozilla/activity-streams/blob/d083dabacb069703e71d5aaf35a25ff21a3a12cd/yamscripts.yml#L11)).
We should probably switch to the [**rimraf**](https://github.com/isaacs/rimraf) module for better OS support (lookin' at you, Windows).

I still need to hack on my Windows 10 laptop to try and get this all working. I fell down a sharp well of Node 4+ support and needing specific versions of Python.

gasolin:
use rimraf and mkdirp to replace native command.

Need set `"react": "^0.14.7"` to pass dependency check (will install `0.14.8`)

Fixes #125 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/619)


pdehaan:
Looks good to me. r? @k88hudson

k88hudson:
Thanks!

PR #624, opened 2016-05-03
fix (addon): #579 hide simple prefs

### LINKED ISSUE #579: hide checkboxes and urls in about:addons prefs
This was discussed to leave or hide checkboxes and urls in addon prefs

screen shot 2016-04-26 at 2 00 01 pm

rlr:
r? @oyiptong 

Fixes #579

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/624)

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,8 +85,7 @@
       "name": "performance.log",
       "title": "Enable Performance Logging",
       "type": "bool",
-      "value": false,
-      "hidden": true
+      "value": false
     },
     {
       "name": "query.cache",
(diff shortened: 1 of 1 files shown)


rlr:
Here is the new prefs screen with this commit:

screen shot 2016-05-03 at 4 16 17 pm

oyiptong:
Do we want to hide all of them?

I'm good either way, because as a developer I know how to hide prefs. However, as an end-user, it's not Firefox-like to not provide options for users.

rlr:
Just following what the issue says. IMO they shouldnt be preferences if nobody can change them. not sure what the point is.

oyiptong:
following from #579:

We should:
* hide query caching checkbox
* leave preview checkbox as-is
* leave perf logging checkbox as-is

rlr:
OK, ^^ updated to only add query caching to the hidden list.

PR #633, opened 2016-05-04
Improve cache invalidations

### LINKED ISSUE #629: Persist query cache
Ensure query cache is persisted to diminish the cold-start problem

### LINKED ISSUE #630: Only invalidate query cache when new data is ready
Instead of invalidating data before repopulating, only invalidate when new data is ready.

Due to some profiles being big causing queries to take 30+ seconds, one way to ensure this takes less time is to only clear the existing cache for a query when new data is ready.

oyiptong:
r? @rlr 
closes #629 
closes #630

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/633)


oyiptong:
This PR includes 3 commits:

1. Persists the query cache to minimize the cold-boot problem (i.e. a user's cache is empty every time they start their browser)
2. Highlights link performance was not being logged, it now is
3. Instead of invalidating the query cache before repopulating, replace the query cache when new data is ready. This should improve performance for profiles with large histories where queries may take very long

oyiptong:
@ncloudioj we should see how performance improves when this lands/gets deployed

lib/Memoizer.js
```diff
@@ -1,61 +1,117 @@
...
-      let repr = concat.call(parts);
+      let repr = this._makeRepr(args);
 
-      if (this._cacheEnabled && repr in memo) {
+      if (this._cacheEnabled && this.hasRepr(key, repr)) {
         Services.obs.notifyObservers(null, `${key}-cache`, "hit");
-        return memo[repr];
+        return new Promise(resolve => resolve(this.getData(key, repr)));
       } else {
         Services.obs.notifyObservers(null, `${key}-cache`, "miss");
```
rlr: if cache is disabled, you are reporting it as a cache miss.

lib/Memoizer.js
```diff
@@ -1,61 +1,117 @@
...
+      ss.storage.queryCache[key] = {};
+    }
+    ss.storage.queryCache[key][repr] = data;
+  },
+
+  getData(key, repr) {
+    if (this.hasRepr(key, repr)) {
+      return ss.storage.queryCache[key][repr];
+    }
+    return undefined;
```
rlr: This line isn't needed since that's the default behavior of functions in js if nothing is returned.

lib/Memoizer.js
```diff
@@ -1,61 +1,117 @@
...
+   * Return a function that will wrap a given function and cache its results.
+   *
+   * @param {String} key
+   *   a key to store/retrieve cached values
+   *
+   * @param {Function} func
+   *   a function to wrap for caching
+   *
+   * @return {Promise}
+   *   A promise whose result contains data
```
rlr: This actually returns a function that returns a Promise if I am reading it right.

ncloudioj:
Yes, we're looking to create a "load_latency" monitoring chart in the telemetry dashboard.

Also @tspurway and @k88hudson mentioned about the performance regression test with a standard fixture, i think that's also worth a try.

rlr:
Otherwise, LGTM.

lib/Memoizer.js
oyiptong: You are correct!

lib/Memoizer.js
oyiptong: good point

lib/Memoizer.js
oyiptong: good catch!

lib/Memoizer.js
rlr: although I kinda like being explicit about it. heh

oyiptong: how about `return null;`? that is pretty explicit

rlr: sure 👍

lib/Memoizer.js
oyiptong: Actually, it's intentional. I call it a "miss" because it didn't hit the cache

rlr: true, it won't skew the data? but do you still neeed to spawn the Task below? i guess nobody will have it disabled anyway

oyiptong: Yeah, the Task is there to execute an async query if need be.
I.e. it will always execute async.

You are right, it might make data non-straightforward if we are counting how often there are cache misses. We should still obtain information on relative changes in this number though.

That said, we're not counting cache misses yet. I think we should tackle that when we do.

PR #645, opened 2016-05-05
style(content): #610 Fade in main sections on init

### LINKED ISSUE #610: Recent Activity has a half-second load time which makes the UI feel jerky
The Recent Activity section in Activity Stream takes a brief moment to load (maybe half a second). It's a really small lag, but I think it would benefit from having a loading gif to both 

a) make it seem like the UI isn't broken (for users loading up Activity Stream for the second or third time, expecting to see Recent Activity immediately) and 
b) make the UI feel less jerky.

I said a loading gif above; I don't think this would need to _literally_ be a loading gif as in a spinner. It could be a that we use divs with gray backgrounds that we replace with content à la Facebook.

k88hudson:
This should help a bit to make the UI feel less "janky" on first load

Fixes #610 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/645)

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -50,7 +50,7 @@ const NewTabPage = React.createClass({
           <Search onSearch={this.onSearch} />
         </section>
 
-        <div className="delayedFadeIn">
+        <div className={classNames("show-on-init", {on: this.props.isReady})}>
           <section>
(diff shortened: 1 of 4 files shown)


k88hudson:
@oyiptong what do you think?

oyiptong:
this looks great!

oyiptong:
r+

PR #654, opened 2016-05-06
fix (addon): #646 do not throw exception on invalid uri component

### LINKED ISSUE #646: Bookmark urls containing % cause Bookmarks view to not render
I have some bookmarks saved but one of them is causing an error in the add-on code: `URIError: malformed URI sequence PreviewProvider.js:124:79`

I looked through my bookmarks and the problem was my search Bugzilla bookmark: `https://bugzilla.mozilla.org/buglist.cgi?quicksearch=%s`

If I removed the `%` from the bookmark URL then the Bookmarks view in Activity Streams works great.

rlr:
r? @oyiptong @k88hudson

Fixes #646

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/654)

### DIFF
--- a/lib/PreviewProvider.js
+++ b/lib/PreviewProvider.js
@@ -121,7 +121,7 @@ PreviewProvider.prototype = {
 
     // extract and parse the query parameters, if any
     if (newURL.search.length !== 0) {
-      let queryParams = new Map(newURL.search.slice(1).split("&").map(pair => pair.split("=").map(decodeURIComponent)));
+      let queryParams = new Map(newURL.search.slice(1).split("&").map(pair => pair.split("=")));
 
       // filter out the allowed query params and update the query string
       newURL.search = Array.from(queryParams.keys())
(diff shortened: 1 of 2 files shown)


lib/PreviewProvider.js
```diff
@@ -110,6 +110,17 @@ PreviewProvider.prototype = {
   },
 
   /**
+    * A decodeURIComponent that doesn't throw on invalid URI components
+    */
+  _safeDecodeURIComponent(str) {
+    try {
+      return decodeURIComponent(str);
+    } catch (e) {
+      return str;
```
rlr: maybe should return empty string instead?

oyiptong: AFAICS `decodeURIComponent` fails when there's an invalid URL passed.

Is there a situation where it is failing on valid URL's?

For instance, this passes: `decodeURIComponent("foo://example.com/bar?baz=quux&foo=bar%20baz")`

But this doesn't: `decodeURIComponent("foo://example.com/bar?baz=quux&foo=bar%baz")`

This is rightly so, because the URL is invalid

k88hudson: @oyiptong and I discussed this, let's verify that the URLs which come back are *not* URI encoded, and if that is the case, we shouldn't be decoding at all

rlr: good point. It doesnt seem like they are encoded. I'll remove

rlr:
^^ updated to remove the decoding.

k88hudson:
Looks good to me R+ (this should pass if you rebase)

rlr:
squashed, rebased and waiting on travis to give green light.

PR #660, opened 2016-05-06
fix(concurrency): guard against handlers called when ActivityStream object is gone

oyiptong:
r? @k88hudson

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/660)


k88hudson:
R+

PR #662, opened 2016-05-06
Loading improvements

k88hudson:
This patch:

* removes the `previewsOnly` constraint for Highlights/TopSites
* increases the timeout content-side from `5s` to `20s`
* shows a loading spinner that says `Hang on tight! We are analyzing your history to personalize your experience` if loading is taking longer than 20 seconds

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-streams/662)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -111,7 +111,7 @@ ActivityStreams.prototype = {
     switch (msgName) {
       case am.type("TOP_FRECENT_SITES_REQUEST"):
         this._memoized.getTopFrecentSites(msg.data).then(links => {
-          this._processAndSendLinks(links, "TOP_FRECENT_SITES_RESPONSE", append, worker, true);
+          this._processAndSendLinks(links, "TOP_FRECENT_SITES_RESPONSE", append, worker);
         });
         break;
       case am.type("RECENT_BOOKMARKS_REQUEST"):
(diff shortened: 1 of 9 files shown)


k88hudson:
I still need to work on the tests this, but would you mind taking a look at this @oyiptong?

lib/ActivityStreams.js
```diff
@@ -67,6 +68,12 @@ function ActivityStreams(options = {}) {
     preview: false,
   };
 
+  // This lets us know whether or not this is the first time someone has
+  // installed the add-on, so we can wait for previews to come back
+  simpleStorage.storage.waitForFirstPreviews = {};
```
k88hudson: What do you think about this?

oyiptong: why don't we use a pref? That would also make it easier to test

oyiptong: `simplePrefs` is already included in `lib/ActivityStreams.js`, and we can have a default value in package.json.
We could have two prefs: one for top_frecent and another for highlights_links.

The code here would blow away the `storage` every time you boot up the addon it looks. You want a more persistent pref, right?

oyiptong: discussed with @k88hudson in person. It might be preferable to show the load indicator if the query cache is empy rather than to rely on prefs.

It is less error prone. The only edge case to watch out for is when a history is truly empty.

I.e. the loading indicators should show if history is not empty and the query cache is empty.

k88hudson:
Ok, I've updated the PR not to include the first run experience stuff

oyiptong:
r+ with test case fixes

PR #683, opened 2016-05-11
feat(addon): add the API to remove a bookmark

### LINKED ISSUE #670: Add a function to remove a bookmark
Content needs an API to be able to remove a bookmark.

This will consist of creating a function in `PlacesProvider.js` that will make a call to `Bookmarks.jsm` to remove the said bookmark.

It will be similar to https://github.com/mozilla/activity-stream/blob/4d5c47a05a1caca0864b1b7342e700443d5b8333/lib/PlacesProvider.js#L275

But will make a call to https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/Bookmarks.jsm#385

Additionally, the bookmark API bindings need to be implemented in `lib/ActivityStreams.js` and in `common/action-manager.js` so that they are accessible to content.

rittme:
Fixes #670

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/683)

### DIFF
--- a/lib/PlacesProvider.js
+++ b/lib/PlacesProvider.js
@@ -273,6 +273,16 @@ Links.prototype = {
   },
 
   /**
+   * Removes a bookmark
+   *
+   * @param {String} bookmarkGuid the bookmark guid
+   * @returns {Promise} Returns a promise set to an object representing the removed bookmark
+   */
+  asyncDeleteBookmark: function PlacesProvider_asyncDeleteBookmark(bookmarkGuid) {
+    return Bookmarks.remove(bookmarkGuid);
+  },
+
+  /**
    * Removes a history link
    *
    * @param {String} url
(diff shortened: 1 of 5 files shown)


lib/PlacesProvider.js
```diff
@@ -273,6 +273,16 @@ Links.prototype = {
   },
 
   /**
+   * Removes a bookmark
+   *
+   * @param {String} bookmarkGuid the bookmark guid
+   * @returns {Promise} Returns a promise set to an object representing the removed bookmark
+   */
+  deleteBookmark: function PlacesProvider_deleteBookmark(bookmarkGuid) {
```
oyiptong: nit: we're trying something from now on: every function that is async (i.e. returns a promise) should have `async` in the name.

this method would be `asyncDeleteBookmark`.

We haven't applied the same convention to other functions yet, and we should.

oyiptong:
you could also add `BOOKMARK_DELETE` as an event constant: https://github.com/mozilla/activity-stream/blob/master/common/event-constants.js

PR #688, opened 2016-05-12
feat(addon): #679 Send addon version to embedly proxy

### LINKED ISSUE #679: Send addon version to embedly proxy
When we query the embedly proxy, we should also send the addon version as a query parameter so that during debugging on the server side we can have a clearer sense of where traffic is coming from.

ex: POST /v2/extract?addon_version=1.0.15

sarracini:
easy peasy lemon squeeze @jaredkerim r?

Fixes #679 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/688)

### DIFF
--- a/lib/PreviewProvider.js
+++ b/lib/PreviewProvider.js
@@ -4,10 +4,12 @@
 const {Cu} = require("chrome");
 const ss = require("sdk/simple-storage");
 const simplePrefs = require("sdk/simple-prefs");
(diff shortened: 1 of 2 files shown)


pdehaan:
@sarracini Me thinks this needs a rebasin' due to a mid-air collision.

jaredkerim:
r+ lgtm!

PR #689, opened 2016-05-12
tests(addon): #428 Ensure embedly client throws out non-requested responses

### LINKED ISSUE #428: Check for missing results from Embedly Proxy
Soon, the embedly proxy may exclude a result from its response due to data validation or other errors.  So for instance if you send urls [url1, url2, url3] you may only get back [url1, url3].  The client may want to check that it got back what it asked for and potentially re-request failed attempts or log them in some way.

sarracini:
@jaredkerim r? 

Just a quick little test

Fixes #428 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/689)

### DIFF
--- a/test/test-PreviewProvider.js
+++ b/test/test-PreviewProvider.js
@@ -305,6 +305,49 @@ exports.test_dedupe_urls = function*(assert) {
   });
 };
 
+exports.test_throw_out_non_requested_responses = function*(assert) {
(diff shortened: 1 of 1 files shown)


jaredkerim:
This is good, this ensures that we only pay attention to URLs which we requested.  Can you also add a test for the inverse case:

ie Client requests sites a b c, and only receives a b, puts them in the cache properly and carries on happily

jaredkerim:
r+! 🍰

PR #695, opened 2016-05-12
fix (content): #675 use cacheKey instead of lastVisitDate as react child key

### LINKED ISSUE #675: Bug 1271974 - One or two "Highlights" tiles are not displayed after importing history from Chrome profile
via https://bugzilla.mozilla.org/show_bug.cgi?id=1271974

<hr>

[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.16

[Affected Platforms]:
- All Windows
- All Linux
- All Mac OS

[Notes]:
- Attached a Chrome profile data as history dataset to reproduce the issue. I have added in the attachment the steps for how to create the Chrome profile with the attached data (https://goo.gl/Fld5F5). 
- I have reproduced this issue with several different Chrome profiles, the one attached was a new Chrome profile on which I created some history in the last 2-3 days.
- Attached a screen recording of the issue https://goo.gl/9FK9xB.
- The issue is reproducible on the New Tab page as well. 

[Prerequisites]:
- Latest Activity Stream add-on version (1.0.16) was previously installed on a clean profile.  

[Steps to reproduce]:
2. Click the Activity Stream button from the browser toolbar.
3. Press the "Ctrl+Shift+B" key combination.
4. After the "Library" window is opened, click "Import and Backup" button form the top of the window.
5. From the displayed context menu, click on the "Import data from another browser..." button.
6. Choose to import history from Chrome and select the profile from the provided attachment. 
7. Wait until the history is imported and after close the "Library" window.
8. Refresh the Timeline page.

[Expected result]:
- All 3 "Highlights" tiles are correctly populated with history from Chrome profile.

[Actual result]:
- Only two of the 3 tiles are displayed and populated with history from Chrome profile.

[Additional Notes]:
- If the all three "Highlights" tiles are displayed, remove the first "Highlight" tile from history. This will leave the "Timeline" with just 2 of the 3 tiles displayed.

rlr:
The issue was that the profile used for testing had sites with identical `lastVisitDate` values and we were using those for child keys in react and react doesn't like the dupes.

What do you think of using the `cacheKey` instead? @k88hudson f? r?

Fixes #675 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/695)


k88hudson:
R+

k88hudson:
The best thing to do would be to pass an `id` from the add-on i think, but this is fine

Issue #518, opened 2016-04-13
embedly-proxy receiving null URLs


jaredkerim:
The activity stream addon is sending requests to the embedly-proxy client with empty URLs like so:
	
{"urls":[null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null]}

Each request has a variable number of URLs in the list, but are always all null.

I suspect this is coming directly from deployed addons (rather than network calls being inadvertently made during automated testing) because I see a distribution of user agents which looks like actual users.

pdehaan:
Too bad we aren't passing the add-on version to the proxy to see if its all older add-ons.

jaredkerim:
It's possible this is coming from automated tests but sentry reports that this is coming from multiple versions of Firefox and from multiple operating systems including OSX, Windows 7-10, and Linux.

oyiptong:
oh interesting. it sounds like it's a deeper issue than just an automated testing config problem

pdehaan:
@jaredkerim Are you still seeing this nullery in the logs?

I wasn't able to reproduce https://github.com/mozilla/embedly-proxy/issues/70 earlier today w/ testpilot.firefox.com, so wondering if it was just a transient error or solar flares or something.

rlr:
I found the issue here. Or one of the issues that causes this. On startup, `ActivityStreams.js` builds a preview cache `this._asyncBuildPreviewCache();` but this function is missing one step. It calls `this._previewProvider.asyncSaveLinks(links);` without calling `links = this._previewProvider.processLinks(links);`. `processLinks()` is what sets the `sanitizedUrl` to pass to embedly among other things. Since it isn't called, it is passing null instead.

Want me to grab this @jaredkerim ?

rlr:
This also means that the previewCache we create on startup isn't really working so fixing this might give a perf boost in some cases.

rlr:
This should fix it... The hard part is writing the tests 😆 

```
diff --git a/lib/ActivityStreams.js b/lib/ActivityStreams.js
index 0c79f6a..268097b 100644
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -342,6 +342,7 @@ ActivityStreams.prototype = {
       }));

       yield Promise.all(promises);
+      linksToSend = this._previewProvider.processLinks(linksToSend);
       yield this._previewProvider.asyncSaveLinks(linksToSend);
[...]

PR #697, opened 2016-05-13
fix(addon): #518 Fix Embedly proxy receiving null URLs

sarracini:
Working on a test for it now, but this fixes the null problem. yay!

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/697)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -342,6 +342,7 @@ ActivityStreams.prototype = {
       }));
 
       yield Promise.all(promises);
+      linksToSend = this._previewProvider.processLinks(linksToSend);
       yield this._previewProvider.asyncSaveLinks(linksToSend);
       this._populatingCache.preview = false;
       Services.obs.notifyObservers(null, "activity-streams-previews-cache-complete", null);
--- a/lib/PreviewProvider.js
+++ b/lib/PreviewProvider.js
@@ -243,7 +243,8 @@ PreviewProvider.prototype = {
         links.push(link);
       }
     }
-    yield this.asyncSaveLinks(links, false, false);
+    let linksToUpdate = this.processLinks(links);
+    yield this.asyncSaveLinks(linksToUpdate, false, false);
     Services.obs.notifyObservers(null, "activity-streams-preview-cache-update", null);
   }),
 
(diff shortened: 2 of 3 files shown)


sarracini:
what's going on with travis? https://travis-ci.org/mozilla/activity-stream/builds/130038880#L419 
this doesn't look like it was related to my patch?

oyiptong:
That's the error:

```

  fail:

  undefined - "http://localhost:8199/dummy-activitystreams.html#/timeline/bookmarks" == "about:privatebrowsing"

console.trace: activity-streams: 

_ecated/unit-test.js 108 fail

_/sdk/test/assert.js 88 fail

_/sdk/test/assert.js 123 equal

_private-browsing.js 32 exports["test activity stream doesn't load in private windows"]/</<.onReady

_s/sdk/event/core.js 112 emitOnObject

_s/sdk/event/core.js 89 emit

_tabs/tab-firefox.js 264 tabEmit

_tabs/tab-firefox.js 328 tabEventListener

_tabs/tab-firefox.js 347 

_s/sdk/event/core.js 112 emitOnObject

_s/sdk/event/core.js 89 emit

_s/sdk/event/core.js 112 emitOnObject

_s/sdk/event/core.js 89 emit

_s/sdk/event/core.js 123 emitOnObject

_s/sdk/event/core.js 89 emit

_dk/remote/parent.js 96 messageReceived
```

perhaps this an intermittent issue? otherwise this patch may have caused the test to fail somehow.
In any case @rlr had been working on the private browsing patch and could help if you need it.

rlr:
awesomesauce 👍  R+!

We might want to refactor this later so that `asyncSaveLinks()` calls `processLinks()` so that we don't have this issue of forgetting. But I didn't look to see if that would break something else. Something to consider.

rlr:
I'll take a look at that intermittent failure to see if I can find what is going on. Also, are we releasing this quickly?

sarracini:
I agree about the refactoring. 
We're going to deploy at the end of the day, with this change in it.

oyiptong:
There should not be any rush to release things quickly because we can deploy literally any time.

However, if this is merged today, it will go out with this afternoon's deploy

Issue #604, opened 2016-04-29
Bug 1268870 - If no new website is visited, the yesterday history is still displayed as just now


SoftVision-PaulOiegas:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1268870
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.13

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- Install latest Activity Stream add-on version (1.0.13) on a new profile.
- There are no bookmarks and there are no visited websites (all data was removed from history including defaults).

[Steps to reproduce]:
1. Start Firefox browser with the profile where Activity Stream add-on was previously installed.
2. Click on "Activity Stream" button from browser toolbar.
3. Visit website X.
4. Visit website Y and bookmark it.
5. Click on "Activity Stream" button from browser toolbar.
6. Observe Activity Stream Timeline and Bookmarks pages.
7. Wait until the next day, (change system date by one day in future).
8. Click on "Activity Stream" button from browser toolbar.
9. Observe Activity Stream Timeline and Bookmarks pages.

[Expected results]:
2. "Just now" header is not displayed because there is no History stored.
6. Visited webpages are displayed in "Just now" area from Timeline.
9. "Just now" header is updated to "Yesterday".

[Actual results]:
2. "Just now" header is not displayed because there is no History stored.
6. Visited webpages are displayed in "Just now" area from Timeline.
9. Visited webpages are still displayed in "Just now" area from Timeline, "Yesterday" section was not created. 

[Notes]:
After a website is visited in the "next day", all the other items from "Just now" area are sent to "Yesterday" section. If the visited website from "Just now" is deleted from History, the items that were sent to "Yesterday" are returned to "Just now" section.
![timestamp is not automatically updated](https://cloud.githubusercontent.com/assets/17258026/14917117/19c20d32-0e26-11e6-826c-a5dc91dcffd8.jpg)

oyiptong:
As currently implemented, the page does not respond to events occurring outside the page.

here's a related behaviour: if you open a timeline in tab 1, navigate to a new web page in tab 2 that creates a history entry, return to the timeline in tab 1, the history item will not be in tab 1.

It behaves very much like a traditional web page at the moment, and not like the newtab page. (web page vs browser chrome).

In this case, the page also won't adjust the time stamps as time goes by.

softvision-VladBacia:
[...]

PR #710, opened 2016-05-16
fix (content): #604 remove custom titles from GroupedActivityFeed

rlr:
r? @k88hudson 

What do you think of this? The New Tab page will keep the "Recent Activity" but the Timeline pages will show "Today"/"Yesterday"/etc instead of "Just now".

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/710)


k88hudson:
R+!

PR #715, opened 2016-05-17
feat(content): add Remove Bookmark UI

### LINKED ISSUE #659: Bug 1270842 - Add "remove bookmark" UI
via https://bugzilla.mozilla.org/show_bug.cgi?id=1270842
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.15-dev

[Affected Platforms]:
- All Windows
- All Mac OS X
- All Linux

[Prerequisites]:
- Latest Activity Stream add-on version (1.0.15-dev) was previously installed on a clean profile.  

[Steps to reproduce]:
2. Open Activity Stream from the browser toolbar and click the on the Bookmarks tab.
3. Hover one of the saved bookmarks and click the "X" button.

[Expected result]:
The option to remove a bookmark from Bookmarks should be available in the context menu.

[Actual result]:
Instead of having an option to remove a bookmark from Bookmarks, we only have the option to remove from History.

[Notes]:
- If you remove a bookmark from History, it is removed from the page, but after a refresh it reappears.
- From the User's perspective the option to remove from History doesn't make much sense in the Bookmarks tab since this can be achieved from Timeline->All History page. 
- Attached a screen recording of the issue https://goo.gl/m9MBDq.

rittme:
Fixes #659 
This still don't work on the Top Sites, since we don't get the bookmarkGuid for the top frecent sites. Should I file a follow up issue to fix this?
Note that I replaced a "NOTIFY_HISTORY_DELETE" event listener at the ActivityStream.js since it was duplicated.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/715)

### DIFF
--- a/content-src/components/ActivityFeed/ActivityFeed.js
+++ b/content-src/components/ActivityFeed/ActivityFeed.js
@@ -86,6 +86,7 @@ const ActivityFeedItem = React.createClass({
         visible={this.state.showContextMenu}
         onUpdate={val => this.setState({showContextMenu: val})}
         url={site.url}
+        bookmarkGuid={site.bookmarkGuid}
         page={this.props.page}
         index={this.props.index}
         source={this.props.source}
--- a/content-src/components/DeleteMenu/DeleteMenu.js
+++ b/content-src/components/DeleteMenu/DeleteMenu.js
@@ -15,22 +15,34 @@ const DeleteMenu = React.createClass({
       }));
     }
   },
-  onDelete() {
+  onDeleteHistory() {
     this.props.dispatch(actions.NotifyHistoryDelete(this.props.url));
     this.userEvent("DELETE");
+  },
+  onDeleteBookmark() {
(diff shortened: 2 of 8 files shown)


rlr:
LGTM and works pretty much as I expected. There is just one minor issue. When on the newtab or history page, removing the bookmark should only unstar the item instead of removing the url from the page. Because if you refresh it, the url is back since it wasn't deleted from history.

So... I think if instead of filtering out the url in the reducer, maybe just setting the `bookmarkGuid` to null would work?

rittme:
Only setting the `bookmarkGuid` to `null` works for the newtab and the history page, but for the bookmarks page it's not the expected behavior, because the item stays on the list.

But now I'm wondering if the behavior should be removing it from the history too. When the user clicks remove, he wants it to only not be a bookmark anymore or he wants it to be removed from the list?

PR #725, opened 2016-05-19
fix(addon): Add an unique id for each item sent to content

### LINKED ISSUE #712: Send unique id for each item for use in react keys
Ideally we should be using an unique identifier to for the `key` attribute of repeated items in React to improve performance (this is a good explanation http://jaero.space/blog/react-performance-1 for more information about why this is the case).

Let's add a new field in our queries in the `PlacesProvider` to return a unique id for each item (this includes the `getRecentBookmarks`, `_getHistoryLinks` methods, etc) and use it in our components (for example, in the `ActivityFeed` component.

rittme:
Fixes #712

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/725)


rlr:
awesome! It WFM, yay! R+ 🚀

PR #736, opened 2016-05-24
fix(addon): Bug 1273164 - The search bar from New Tab page is no longer working on latest Nightly version

sarracini:
@oyiptong r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/736)

### DIFF
--- a/lib/SearchProvider.js
+++ b/lib/SearchProvider.js
@@ -52,8 +52,19 @@ NewTabSearchProvider.prototype = {
 
   get state() {
     return Task.spawn(function*() {
-      let state = yield ContentSearch._currentStateObj();
-      return state;
+      let state;
+      try {
+        if ("currentStateObj" in ContentSearch) {
+          // ContentSearch.jsm from Firefox 49+
+          state = yield ContentSearch.currentStateObj(true);
+        } else {
+          // legacy ContentSearch
+          state = yield ContentSearch._currentStateObj();
+        }
+        return state;
+      } catch (e) {
+        Cu.reportError(e);
+      }
     }.bind(this));
   },
 


lib/SearchProvider.js
```diff
@@ -52,7 +52,15 @@ NewTabSearchProvider.prototype = {
...
     return Task.spawn(function*() {
-      let state = yield ContentSearch._currentStateObj();
+      // a new version of ContentSearch.jsm landed in nightly but not firefox
+      // therefore, for now, we need to accomodate for both versions.
+      let state;
+      try {
+        state = yield ContentSearch.currentStateObj(true);
+      } catch (e) {
+        state = yield ContentSearch._currentStateObj();
+        Cu.reportError(e);
```
oyiptong: This is expected behavior and we shouldn't report an error. Maybe it could be more explicit to do this as follows:

```javascript
let state;
try {
  if ("currentStateObj" in ContentSearch) {
    // ContentSearch.jsm from Firefox 49+
    state = yield ContentSearch.currentStateObj(true);
  } else {
    // legacy ContentSearch
    state = yield ContentSearch._currentStateObj();
  }
} catch(e) {
  Cu.reportError(e);
}
```

lib/SearchProvider.js
```diff
@@ -52,7 +52,15 @@ NewTabSearchProvider.prototype = {
 
   get state() {
     return Task.spawn(function*() {
-      let state = yield ContentSearch._currentStateObj();
+      // a new version of ContentSearch.jsm landed in nightly but not firefox
```
oyiptong: This comment is outdated the moment it lands because it does not contain the specific firefox version number.
Please include the version.

oyiptong:
We should be clear and explicit that this is to support API changes in a specific version of Firefox.
Please see my suggested changes.

Issue #454, opened 2016-04-04
Bug 1259396 - "See more activity" button is displayed when one website is available in history


pdehaan:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1259396

---

Created attachment 8734318 [details]
see more activity button

#### [Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.4

#### [Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

#### [Prerequisites]:
- Latest Activity Stream add-on version (1.0.4) was previously installed on a profile.
- There is no history for the profile used.

#### [Steps to reproduce]:

1. Start Firefox browser with the profile where Activity Stream add-on was previously installed.
2. Visit a website (e.g. 9gag.com)
3. After the page is loaded click on "Activity Stream" button from browser toolbar.

#### [Expected results]:
One website is displayed in the Timeline.
There is no "See more activity" button.

#### [Actual results]:
"See more activity" button is displayed even if there are no other websites that can be loaded in Timeline.

pdehaan:
I was able to reproduce this today. I only have 2 items in my history

![activity_stream](https://cloud.githubusercontent.com/assets/557895/14393067/b64a187e-fd79-11e5-9f33-6705a996bcdd.png)

softvision-VladBacia:
This issue was verified on Activity Stream v1.0.14-dev and it is reproducible.

k88hudson:
This is resovled with 005ba6f

PR #739, opened 2016-05-24
fix(addon): #454 Hide Load More if < 20 items on history

k88hudson:
This fixes an issue where we still see the load more box if < 20 items are returned from the query.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/739)

### DIFF
--- a/content-src/reducers/SetRowsOrError.js
+++ b/content-src/reducers/SetRowsOrError.js
@@ -8,7 +8,7 @@ const DEFAULTS = {
   canLoadMore: true
 };
 
-module.exports = function setRowsOrError(requestType, responseType) {
+module.exports = function setRowsOrError(requestType, responseType, querySize) {
   return (prevState = DEFAULTS, action) => {
     const state = {};
     const meta = action.meta || {};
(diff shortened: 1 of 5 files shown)


content-src/reducers/SetRowsOrError.js
```diff
@@ -25,9 +25,15 @@ module.exports = function setRowsOrError(requestType, responseType) {
           state.init = true;
           state.rows = meta.append ? prevState.rows.concat(action.data) : action.data;
           state.error = false;
+          // If there is no data, we definitely can't load more.
           if (!action.data || !action.data.length) {
             state.canLoadMore = false;
           }
+          // If the results returned are less than the query size,
+          // we should be on our last page of results.
```
rlr: is there going to be an edge case where we are on the last page of results but it is full?

rlr:
Not sure if that matters or not. Otherwise, this looks awesome. R+

content-src/reducers/SetRowsOrError.js
k88hudson: There is, but there's not really a way around that

PR #744, opened 2016-05-24
fix(prefs): Switch to production embedly-proxy URL

### LINKED ISSUE #737: Dev Addon should use Production Embedly Proxy
The dev prefs points to the staging embedly proxy for local development use, see here:

https://github.com/mozilla/activity-stream/blob/master/dev-prefs.json#L5

However, this should point to the production embedly proxy endpoint here:

https://embedly-proxy.services.mozilla.com/v2/extract

The reason for this is that the stage embedly proxy is meant to test embedly proxy code deploys which may break, and if the stage instance breaks (which is an expected result) then that will block use of any addons built using the dev prefs which can block addon development.

pdehaan:
Fixes #737

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/744)


jaredkerim:
r+

Issue #705, opened 2016-05-16
Collect proxy telemetries from addon


ncloudioj:
We wanted to collect some pings for the proxy related metrics. It needs following steps:
- sort out the target metrics @jaredkerim 
- client side ping collecting and notification @sarracini 
- server side telemetry processing and persistence @ncloudioj

sarracini:
Based on #680 we want to collect:
```js
{
    source: "TOP_FRECENT_SITES_RESPONSE",  // which SQL query produced this list of URLs
    count: 17, // the total number of URLs returned by the SQL query
    cacheHit: 13, // the number found in our local simple storage cache
    cacheMiss: 4, // the number we failed to find and will request from embedly-proxy
    success: true, // was the call to embedly-proxy successful
    urlsRequested: 4, // how many URLs did we send to the proxy
    urlsReceived: 2, // how many URLs did we get back
    requestLatency: 13.4, // how long did it take for the call to complete
}
```
from the client's side

ncloudioj:
Ah, great. A couple of suggestions:
* each ping from addon has a required field `action`, here we can use `action: "activity_stream_proxy"`
* let's stick with the snake case, e.g. "cacheHit" -> "cache_hit", just be consistent with other pings.
* use integer "millisecond" for the "requestLatency" rather than float point, e.g "13.4" -> "13400"

jaredkerim:
We would like to track the following messages in the following places

https://github.com/mozilla/activity-stream/blob/7c51cd6175849718bf562df0f085982564a80376/lib/PreviewProvider.js#L193

    track('previewCacheRequest', links.length)

    track('previewCacheHits', result.length)

    track('previewCacheMisses', links.length - result.length)

https://github.com/mozilla/activity-stream/blob/7c51cd6175849718bf562df0f085982564a80376/lib/PreviewProvider.js#L288

    track('embedlyProxyRequestSentCount', newLinks.length)

    track('embedlyProxyRequestTime', <duration of request in milliseconds>)

    track('embedlyProxyRequestSuccess', 1) if request is successful ie 200

    track('embedlyProxyRequestFailure', 1) if request is failure ie !200

    track('embedlyProxyRequestReceivedCount', responseJson.urls.length)

    track('embedlyProxyRequestError', 1) if response.error is not null


A single performance ping should contain:

(timestamp: unix timestamp, userID: str, eventName: str, value: int)

PR #749, opened 2016-05-26
Add a description of performance ping #705

jaredkerim:
@ncloudioj r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/749)


ncloudioj:
Could you add some comments for the `value` field? Like, using -1 for the missing value, latency will be stored in millisecond as an integer.

r+

PR #755, opened 2016-05-27
feat(content): show inline video players for youtube and vimeo

### LINKED ISSUE #415: Add embeds (images, videos, etc) to timeline

rittme:
Fixes #415

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/755)


content-src/components/MediaPreview/MediaPreview.js
```diff
@@ -0,0 +1,43 @@
...
+  getInitialState() {
+    return {
+      showPlayer: false
+    };
+  },
+
+  getDefaultProps() {
+    return {
+      site: {},
+      height: null,
```
k88hudson: It doesn't look like you're using these props

content-src/components/MediaPreview/MediaPreview.js
```diff
@@ -0,0 +1,43 @@
...
+const MediaPreview = React.createClass({
+  getInitialState() {
+    return {
+      showPlayer: false
+    };
+  },
+
+  getDefaultProps() {
+    return {
+      site: {},
```
k88hudson: You should define `PropTypes` for `MediaPreview` so that you incorrect prop definitions will throw errors in tests/create warnings. It would probably be a good idea to use [`React.PropTypes.shape`](https://facebook.github.io/react/docs/reusable-components.html) to define the shape of the `site` prop.

content-src/components/MediaPreview/MediaPreview.scss
```diff
@@ -0,0 +1,39 @@
...
+  background: $black no-repeat center center;
+  width: 367px;
+  height: 206px;
+  border-radius: 6px;
+  transition: 0.2s opacity;
+  overflow: hidden;
+  margin-top: 20px;
+}
+
+.video-preview::after {
```
k88hudson: you can add a bit of nesting here

```sass
.video-preview {
   &::after {
   }
   ..
   &.isPlaying::after {
   }
   ...
}
```

content-src/components/MediaPreview/MediaPreview.scss
```diff
@@ -0,0 +1,39 @@
...
+  width: 65px;
+  height: 40px;
+  background: url('img/playButton@2x.png') no-repeat center center;
+  background-size: 65px 40px;
+  display: block;
+  left: 50%;
+  top: 50%;
+  position: absolute;
+  margin-left: -32px;
+  margin-top: -20px;
```
k88hudson: This centering strategy definitely works, but I would recommend using flexbox centering on `.video-preview` instead since it doesn't involve hard-coding margin values:
```
display: flex;
align-items: center;
justify-content: center;
```

content-src/components/MediaPreview/MediaPreview.js
```diff
@@ -0,0 +1,43 @@
...
+  render() {
+    const site = this.props.site;
+
+    let player;
+    if (this.state.showPlayer) {
+      player = (<iframe className="video-preview-player" width="367" height="206" type="text/html" src={site.previewURL} frameborder="0" allowFullScreen></iframe>);
+    }
+
+    const style = {backgroundImage: `url(${site.thumbnail.url})`};
+    return (<div className={classNames("video-preview", {isPlaying: this.state.showPlayer})}
```
k88hudson: instead of having a class for this, you could just use the `hidden` field, i.e. `hidden={this.state.showPlayer}`

content-src/lib/getVideoPreview.js
```diff
@@ -0,0 +1,36 @@
+module.exports = function getVideoPreview(site) {
+  switch (site.provider_url) {
+    case "https://www.youtube.com/":
```
k88hudson: I think it might be a better idea to test `site.url` instead with a slightly more forgiving regex, to account for embedly errors/inconsistencies

k88hudson:
Great work 👍 this is looking awesome!

A few comments in the code, also can you add some tests? Ping me if you need more info on anything

k88hudson:
One more thing: the dot on the activity feed item should always be in the centre of the icon, like this:

screenshot

content-src/lib/getVideoPreview.js
pdehaan: We've also used [**get-youtube-id**](https://www.npmjs.com/package/get-youtube-id) in other projects (such as [**meandavejustice/min-vid**](https://github.com/meandavejustice/min-vid)).

Not sure if there is a similar library to try and parse URLs and extract vimeo ids. I did find [**vimeo-regex**](https://github.com/regexhq/vimeo-regex), but not sure how much it will help vs bloat.

pdehaan:
Not sure if we're going to need to make a collapsable UI or add the ability to delete a preview. After a bit of testing and YouTube-in' my timeline will get pretty cluttered up.

new_tab

package.json
```diff
@@ -9,6 +9,7 @@
   "dependencies": {
     "classnames": "2.2.3",
     "fancy-dedupe": "0.1.0",
+    "get-youtube-id": "^1.0.0",
```
k88hudson: can you lock this to `1.0.0`?

content-test/selectors/selectors.test.js
```diff
@@ -252,6 +254,57 @@ describe("selectors", () => {
...
+      assert.isObject(state.thumbnail);
+      assert.property(state.thumbnail, "url");
+      assert.equal(state.thumbnail.url, siteWithMedia.images[0].url);
+    });
+    it("should have a type", () => {
+      assert.equal(state.type, siteWithMedia.media.type);
+    });
+    it("should return null thumbnail if no images exists", () => {
+      state = selectSitePreview(Object.assign({}, siteWithMedia, {images: []}));
+      assert.equal(state.thumbnail, null);
```
k88hudson: this is fine, there's actually an `assert.isNull` method you can use if you want

content-test/lib/getVideoPreview.test.js
```diff
@@ -0,0 +1,27 @@
+const {assert} = require("chai");
+
+const getVideoPreview = require("lib/getVideoPreview");
+
+describe("getVideoPreview", () => {
+  it("should return null if url is falsey", () => {
+    assert.equal(getVideoPreview(), null);
```
k88hudson: you can use `assert.isNull` if you want

content-test/lib/getVideoPreview.test.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+  it("should return an embed url for a valid youtube url", () => {
+    const videoId = "lDv68xYHFXM";
+    assert.equal(getVideoPreview(`https://www.youtube.com/watch?v=${videoId}`), `https://www.youtube.com/embed/${videoId}?autoplay=1`);
+  });
+  it("should return an embed url for a valid vimeo url", () => {
+    const videoId = "1202674";
+    assert.equal(getVideoPreview(`https://vimeo.com/${videoId}`), `https://player.vimeo.com/video/${videoId}?autoplay=1`);
+  });
+});
```
k88hudson: 👍

content-test/components/MediaPreview.test.js
```diff
@@ -0,0 +1,57 @@
...
+    TestUtils.Simulate.click(el);
+    const previewPlayer = instance.refs.previewPlayer;
+    assert.equal(previewPlayer.src, fakeProps.previewInfo.previewURL);
+  });
+
+  it("should not show the iframe when clicked and previewURL is falsey", () => {
+    setup({previewURL: null});
+    TestUtils.Simulate.click(el);
+    const previewPlayer = instance.refs.previewPlayer;
+    assert.equal(typeof previewPlayer, "undefined");
```
k88hudson: `assert.isUndefined`

PR #758, opened 2016-05-27
feat(metrics): #705 Collect proxy telemetries from addon

sarracini:
@jaredkerim r? 🐱

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/758)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -62,7 +62,7 @@ function ActivityStreams(options = {}) {
   this._telemetrySender = new TelemetrySender();
   this._tabTracker = new TabTracker(this.appURLs, options.clientID, this._memoized);
 
-  this._previewProvider = new PreviewProvider();
+  this._previewProvider = new PreviewProvider(this._tabTracker);
   this._populatingCache = {
     places: false,
     preview: false,
@@ -153,10 +153,12 @@ ActivityStreams.prototype = {
   _processAndSendLinks(links, responseType, worker, options) {
     let {append, previewsOnly, skipPreviewRequest} = options || {};
     let processedLinks = this._previewProvider.processLinks(links);
+    const event = this._tabTracker.generateEvent({source: responseType});
     if (!skipPreviewRequest) {
-      this._previewProvider.asyncSaveLinks(processedLinks);
+      this._previewProvider.asyncSaveLinks(processedLinks, event);
     }
     const cachedLinks = this._previewProvider.getEnhancedLinks(processedLinks, previewsOnly);
+    this._handlePerformanceEvent(event, processedLinks, cachedLinks);
     this.send(am.actions.Response(responseType, cachedLinks, {append}), worker);
   },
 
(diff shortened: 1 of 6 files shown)


lib/ActivityStreams.js
```diff
@@ -152,9 +152,11 @@ ActivityStreams.prototype = {
    * Process the passed in links, save them, get from cache and response to content.
    */
   _processAndSendLinks(links, responseType, append, worker, previewsOnly = false) {
+    const event = this._tabTracker.generateEvent({source: responseType});
     let processedLinks = this._previewProvider.processLinks(links);
-    this._previewProvider.asyncSaveLinks(processedLinks);
+    this._previewProvider.asyncSaveLinks(processedLinks, event);
```
pdehaan: A bit of a nit, but a lot of the other methods specify `event` First and other args after. Not sure if worth reordering.

lib/ActivityStreams.js
```diff
@@ -354,8 +362,9 @@ ActivityStreams.prototype = {
       }));
 
       yield Promise.all(promises);
+      const event = this._tabTracker.generateEvent({source: "BUILD_PREVIEW_CACHE"});
       linksToSend = this._previewProvider.processLinks(linksToSend);
-      yield this._previewProvider.asyncSaveLinks(linksToSend);
+      yield this._previewProvider.asyncSaveLinks(linksToSend, event);
```
pdehaan: See above, re `asyncSaveLinks(event, linksToSend)` reordering.

lib/ActivityStreams.js
sarracini: unfortunately in the tests, since we only pass one parameter to ```asyncSaveLinks```, it expects it to be the links, so we would have to pass an empty first parameter in the tests like ```asyncSaveLinks({}, links)```, which is kind of ugly

jaredkerim: Regardless of the order, should be explicitly passing in an event object in the tests as well.  We don't want to wind up in a situation where the method definition expects two arguments and the tests only invoke one of them and are precariously dependent on that ordering to silently hide the second arg, and then someone comes along and changes something and can't figure out why things are failing.

lib/TelemetrySender.js
```diff
@@ -44,6 +46,7 @@ TelemetrySender.prototype = {
       if (this.enabled && !newValue) {
         Services.obs.removeObserver(this, COMPLETE_NOTIF);
         Services.obs.removeObserver(this, ACTION_NOTIF);
+        Services.obs.removeObserver(this, PERFORMANCE_NOTIF);
       } else if (!this.enabled && newValue) {
         Services.obs.addObserver(this, COMPLETE_NOTIF);
         Services.obs.addObserver(this, ACTION_NOTIF);
```
jaredkerim: Shouldn't we register PERFORMANCE_NOTIF here as well?

jaredkerim:
r+wc! Incoming data!  📈

jaredkerim:
We should mute the logging of the performance pings to the stdout so that it doesn't fill up the travis log.

PR #761, opened 2016-05-27
Consolidate action handlers

k88hudson:
I was thinking something like this might be a better pattern for handling and responding to messages from content, since this way:

- removes several bug vectors (forgetting to add/remove event listeners, bind each handler)
- consolidates the location for adding a new listener to a single place (`lib/handlers/handler.js`)
- slight change to the function signature

Suggestions welcome 👍 

What do you think? @oyiptong @rlr @sarracini @emtwo

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/761)


rlr:
YES! Me gusta! Much better. I am not a big fan of switch statements but that's another issue.

Travis didn't likey.

k88hudson:
@rlr I kinda redid this a bit/fixed the tests in #791

PR #776, opened 2016-05-31
feat(addon): #764 SearchProvider manages your engines

### LINKED ISSUE #764: SearchProvider.jsm needs to manage your search engines
The search provider as of right now only does the bare minimum of getting the initial state of search and performs search. It needs to be able to manage which search engines the user is using and needs to be able to cycle through them

sarracini:
Fixes #764 

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/776)


lib/ActivityStreams.js
```diff
@@ -182,6 +182,15 @@ ActivityStreams.prototype = {
           SearchProvider.search.performSearch(browser, searchData);
         });
         break;
+      case am.type("NOTIFY_MANAGE_ENGINES"): // eslint-disable-line
```
pdehaan: What are we `eslint disabling` here?

sarracini: i honestly don't know what it's complaining about - "disallows lexical declarations" (http://eslint.org/docs/rules/no-case-declarations) but i'm not sure why it doesn't like it

pdehaan: Ah, I think you're the first one to declare `let` or `const` within a `case` statement here.
So if I understand the rule/error correctly, it's saying that the `win`, `gBrowser`, and `browser` vars will have their scopes hoisted up [to the `switch` statement's scope].
The workaround is to just ignore the rule like you did (or just ignore the specific error), or else maybe wrap the case statement within `{...}`, like-a-so:
```js
      case am.type("NOTIFY_MANAGE_ENGINES"): {
        const win = windowMediator.getMostRecentWindow("navigator:browser");
        const gBrowser = win.getBrowser();
        const browser = gBrowser.selectedBrowser;
        SearchProvider.search.manageEngines(browser);
        break;
      }
```

PR #777, opened 2016-05-31
feat(addon): #732 Create experiment for DeleteMenu

### LINKED ISSUE #732: Create Experiment for Delete/Block Menu Item Order
This experiment will simply reverse the order of the 'X' menu items (block/delete) in 10% of the population.

1.  instrument addon so that on start-up, it will determine if it is part of the experiment (10%) or control (90%) and persist the result (based on hash of client_id?)
2.  assign an experiment_id to uniquely identify the test (hardcoded)
3.  if the experiment is enabled, reverse the 'X' menu items
4.  if the experiment is enabled, when sending event pings, make sure to include the experiment_id field
5.  modify data pipeline / infernyx job / database schema to process and persist experiment_id
6.  add a comparative re:dash graph to show any differences between the experiment and it's control

ncloudioj:
Fixes #732 

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/777)

### DIFF
--- a/content-src/components/DeleteMenu/DeleteMenu.js
+++ b/content-src/components/DeleteMenu/DeleteMenu.js
@@ -1,18 +1,22 @@
 const React = require("react");
 const {connect} = require("react-redux");
-const {justDispatch} = require("selectors/selectors");
 const {actions} = require("common/action-manager");
 const ContextMenu = require("components/ContextMenu/ContextMenu");
 
 const DeleteMenu = React.createClass({
   userEvent(event) {
     if (this.props.page && this.props.source) {
-      this.props.dispatch(actions.NotifyEvent({
+      let payload = {
         event,
         page: this.props.page,
         source: this.props.source,
         action_position: this.props.index
-      }));
+      };
+
+      if (this.props.experimentData.reverseMenuOptions) {
+        payload.experiment_id = this.props.experimentData.id;
+      }
+      this.props.dispatch(actions.NotifyEvent(payload));
     }
   },
   onDeleteHistory() {
@@ -39,6 +43,10 @@ const DeleteMenu = React.createClass({
 
     menuOptions.push({label: "Never show on this page", onClick: this.onBlock});
 
+    if (this.props.experimentData.reverseMenuOptions) {
+      menuOptions.reverse();
+    }
+
     return (<ContextMenu
       visible={this.props.visible}
       onUpdate={this.props.onUpdate}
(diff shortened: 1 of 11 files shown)


content-src/components/DeleteMenu/DeleteMenu.js
```diff
@@ -1,18 +1,22 @@
...
+      let payload = {
         event,
         page: this.props.page,
         source: this.props.source,
         action_position: this.props.index
-      }));
+      };
+
+      if (this.props.experimentData.reverseMenuOptions) {
+        payload.experiment_id = this.props.experimentData.id;
```
k88hudson: This should always be included I think, not just for a truthy value

lib/ActivityStreams.js
```diff
@@ -60,6 +61,17 @@ function ActivityStreams(options = {}) {
   this._memoized = this._get_memoized(this._memoizer);
 
   this._telemetrySender = new TelemetrySender();
+
+  this._experimentData = {};
+  let rng = seedrandom(options.clientID);
+  // Targeting 10% of users for this experiment
+  if (rng() <= 0.1) {
+    this._experimentData.id = "exp-001-delete-menu";
+    this._experimentData.reverseMenuOptions = true;
```
k88hudson: I would recommend using string or integer based variant IDs such as `"CONTROL"` and `"REVERSE"`.  Also, you might want to call the experiment something neutral like `deleteMenuOrder`

content-src/components/DeleteMenu/DeleteMenu.js
```diff
@@ -57,4 +65,6 @@ DeleteMenu.propTypes = {
   index: React.PropTypes.number
 };
 
-module.exports = connect(justDispatch)(DeleteMenu);
+module.exports = connect(state => {
+  return {experimentData: state.Experiments.data};
+})(DeleteMenu);
```
k88hudson: Could we have a test for this?

k88hudson:
Ok, we are going to do a follow-up patch to refactor the randomizer stuff into its own module

lib/ActivityStreams.js
```diff
@@ -60,6 +61,17 @@ function ActivityStreams(options = {}) {
   this._memoized = this._get_memoized(this._memoizer);
 
   this._telemetrySender = new TelemetrySender();
+
+  this._experimentData = {};
+  let rng = seedrandom(options.clientID);
+  // Targeting 10% of users for this experiment
```
yati-sagade: Hi! Just a curious Joe here. Did this code run on every new tab? If so, is this not putting 10% of new tab opens (as opposed to 10% of FF eligible FF installs) in the variant?

yati-sagade: Just saw the seedrandom(), so that should be ok :)

PR #780, opened 2016-05-31
fix(addon): #759 Add session_id for all pings.

### LINKED ISSUE #759: Event and session pings should have session_id
We should add this because it will make joins between the two tables a lot simpler for finding out what events happened during what sessions.

emtwo:
Fixes #759 

---

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/780)

### DIFF
--- a/lib/TabTracker.js
+++ b/lib/TabTracker.js
@@ -85,6 +85,7 @@ TabTracker.prototype = {
     payload.addon_version = self.version;
     payload.locale = Locale.getLocale();
     payload.page = this._resolvePageType(url);
+    payload.session_id = this._tabData.session_id;
   },
 
   uninit() {
@@ -130,6 +131,15 @@ TabTracker.prototype = {
     return Object.assign({}, eventData, {event_id: String(uuid())});
   },
 
+  _initTabSession(tab, loadReason) {
+    // For session IDs that were set on tab open, let's make sure we
+    // don't overwrite them.
+    this._tabData.session_id = this._tabData.session_id || String(uuid());
+    this._tabData.url = tab.url;
+    this._tabData.tab_id = tab.id;
+    this._tabData.load_reason = loadReason;
+  },
+
   navigateAwayFromPage(tab, reason) {
     // we can't use tab.url, because it's pointing to a new url of the page
     // we have to use the URL stored in this._openTabs object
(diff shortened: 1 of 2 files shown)


lib/TabTracker.js
```diff
@@ -164,8 +173,6 @@ TabTracker.prototype = {
     // already ended, likely with an 'unfocus' unload reason.
     if (this.isActivityStreamsURL(tab.url) && tabs.activeTab.id === tab.id) {
       if (!this._tabData.url) {
-        this._tabData.url = tab.url;
-        this._tabData.tab_id = tab.id;
```
emtwo: I removed these two since `logActivate()` will be called a few lines down and it already sets them

test/test-TabTracker.js
```diff
@@ -348,7 +348,7 @@ exports.test_TabTracker_action_pings = function*(assert) {
       data: {
         source: "topsites",
         action_position: 3,
-        event: "click"
+        event: "CLICK"
```
emtwo: `handleUserEvent()` is specifically looking for uppercase `CLICK`

test/test-TabTracker.js
```diff
@@ -359,13 +359,29 @@ exports.test_TabTracker_action_pings = function*(assert) {
   for (let key of additionalKeys) {
     assert.ok(pingData[key], `The ping has the additional key ${key}`);
   }
-  assert.deepEqual(eventData.msg.data, pingData, "We receive the expected ping data.");
+  assert.deepEqual(eventData.msg.data.source, pingData.source, "the ping has the correct source");
+  assert.deepEqual(eventData.msg.data.event, pingData.event, "the ping has the correct event");
+  assert.deepEqual(eventData.msg.data.action_position, pingData.action_position, "the ping has the correct action_position");
 };
 
 exports.test_TabTracker_unload_reason_with_user_action = function*(assert) {
```
emtwo: In this test, I added promises to wait for tab open and close. This is because `session_id` will only be set after a tab is open (and we must close tabs at the end of every test).

test/test-TabTracker.js
```diff
@@ -359,13 +359,29 @@ exports.test_TabTracker_action_pings = function*(assert) {
   for (let key of additionalKeys) {
     assert.ok(pingData[key], `The ping has the additional key ${key}`);
   }
-  assert.deepEqual(eventData.msg.data, pingData, "We receive the expected ping data.");
+  assert.deepEqual(eventData.msg.data.source, pingData.source, "the ping has the correct source");
+  assert.deepEqual(eventData.msg.data.event, pingData.event, "the ping has the correct event");
+  assert.deepEqual(eventData.msg.data.action_position, pingData.action_position, "the ping has the correct action_position");
```
emtwo: Since this test doesn't open a new tab, `session_id` will be set to `undefined`. Thus, we only check the other attributes instead of all

emtwo:
r? @ncloudioj or @k88hudson 

I just left some github notes in the code which should hopefully make it easier to follow.

k88hudson:
Looks good to me! R+

ncloudioj:
r+ @emtwo 🚢

PR #782, opened 2016-05-31
chore(addon): #781 use webpack in addon for external deps

### LINKED ISSUE #781: Use webpack for external addon dependencies
I think we can have an intermediary step here to introduce external dependencies to the addon -- writing a "bridge" file that essentially proxies a bunch of stuff required from node_modules.

k88hudson:
This adds a webpack task to build a file in `lib/` called `vendor.bundle.js`, which can expose dependencies from npm.

Obviously it would nice if you could require stuff directly, but this is better than copy-pasting

Fixes #781 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/782)


webpack.addon.config.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+const WebpackNotifierPlugin = require("webpack-notifier");
+const webpack = require("webpack");
+const path = require("path");
+
+module.exports = {
+  entry: {
+    "vendor": path.join(__dirname, "./addon-src/vendor.js")
```
pdehaan: Not sure if you want the `./` and `/` in the paths here, or just use the more platform agnostic:
```js
path.join(__dirname, "addon-src", "vendor.js")
```

package.json
```diff
@@ -141,17 +142,19 @@
     "prebundle": "npm run clean && npm run copyTestImages && npm run copyTopSitesJson",
     "bundle:static": "cpx \"content-src/static/**/*\" data/content",
     "bundle:webpack": "NODE_ENV=production webpack",
+    "bundle:webpackAddon": "webpack --config webpack.addon.config.js",
```
pdehaan: Does this bit need a `NODE_ENV=production` as well?

package.json
```diff
@@ -21,6 +21,7 @@
     "redux-logger": "2.4.0",
     "redux-thunk": "1.0.3",
     "reselect": "2.0.3",
+    "seed-random": "^2.2.0",
```
pdehaan: Me thinks drop the semver range operator (`^`) here for consistency with other deps above.

k88hudson: darn, i thought i saw that for sure!

package.json
k88hudson: in this case we don't since there are no production settings... might not be a bad idea for future proofing though

webpack.addon.config.js
k88hudson: eh, `path` knows what to do to make it right for the platform. the initial "./" is kinda unnecessary though

k88hudson:
@ncloudioj what do you think?

package.json
```diff
@@ -21,6 +21,7 @@
     "redux-logger": "2.4.0",
     "redux-thunk": "1.0.3",
     "reselect": "2.0.3",
+    "seed-random": "2.2.0",
```
ncloudioj: I think the package we're using is called "seedrandom".

ncloudioj:
R+ after fixing the package name.

This is great! Like it! Why didn't you get this working earlier?!!

PR #785, opened 2016-06-01
feat (addon): #724 set home page to Activity Stream

### LINKED ISSUE #724: replace about:home with activity stream
As a windows user who views each site in a new window, I want to see activity stream so that I can get value out of my recent items.  

A number of users who open a single window for every site often see about:home or the custom home page that they've set.  New Tab is for multi-tab users.  To maximize the impact of the activity stream experiment, let's take over about:home.  If users want to fall back to the old about:home experience, we should verify they can set that as their default home page in preferences.

rlr:
r? @oyiptong

Fixes #724 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/785)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -499,10 +499,12 @@ ActivityStreams.prototype = {
    * Replace the home page with the ActivityStream new tab page.
    */
   _setHomePage() {
-    // Only hijack the home page if it is set to about:home or about:blank and
-    // the user didn't previously override the preference.
-    if (["about:home", "about:blank"].includes(prefService.get(HOME_PAGE_PREF)) &&
-        !ss.storage.homepageOverriden) {
+    // Only hijack the home page if it isn't set by user or if it is set to
+    // about:home/about:blank
+    // AND the user didn't previously override the preference.
+    if (!ss.storage.homepageOverriden &&
+        (!prefService.isSet(HOME_PAGE_PREF) ||
+         ["about:home", "about:blank"].includes(prefService.get(HOME_PAGE_PREF)))) {
       prefService.set(HOME_PAGE_PREF, this._newTabURL);
     }
   },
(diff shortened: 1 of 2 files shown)


oyiptong:
@rlr could you instead import NewTabPrefsProvider from https://dxr.mozilla.org/mozilla-central/source/browser/components/newtab/NewTabPrefsProvider.jsm ?

We also want to listen to any changes from the pref. If the user manually changes the pref, we need to remember not to reset the pref on uninstall

oyiptong:
for instance, we don't want to `prefService.reset` if the user set it to something else than about:home

rlr:
@oyiptong I am already checking on unload if the user changed the setting to something else I won't reset. What I am not sure how to do is where to keep track if the user has done that in a way that persists across sessions (loads/unloads). Where is a good place to keep that state?

lib/ActivityStreams.js
```diff
@@ -483,6 +489,23 @@ ActivityStreams.prototype = {
...
+  _setHomePage() {
+    // Only hijack the home page if it is set to about:home or about:blank
+    if (["about:home", "about:blank"].includes(prefService.get(HOME_PAGE_PREF))) {
+      prefService.set(HOME_PAGE_PREF, this._newTabURL);
+    }
+  },
+
+  _unsetHomePage() {
+    // Reset home page back if user didn't change it.
+    if (prefService.get(HOME_PAGE_PREF) === this._newTabURL) {
```
rlr: This is where I conditionally reset it back. I think it is way simpler than adding listeners and all that jazz.

rlr:
@oyiptong This is updated so we respect the user if they override the home page or reset it back to the default. I looked at NewTabPrefsProvider but that doesn't expose the pref we need. I think we don't need to listen for the change because we can detect it on unload.

Anyway, I think this is ready for r? now.

lib/ActivityStreams.js
```diff
@@ -57,14 +59,19 @@ const PLACES_CHANGES_EVENTS = [
   "bookmarkChanged"
 ];
 
+const HOME_PAGE_PREF = "browser.startup.homepage";
+
 function ActivityStreams(options = {}) {
   this.options = Object.assign({}, DEFAULT_OPTIONS, options);
   EventEmitter.decorate(this);
 
+  this._newTabURL = `${this.options.pageURL}#/`;
```
oyiptong: I know this is only slightly different from what there was before, however, I think a comment explaning why there's a `#/` at the end of the url for readers unfamiliar with how our content code works

PR #787, opened 2016-06-01
Refactor cache updating strategy

k88hudson:
This patch removes the `_runPeriodicUpdate` function and replaces it with a strategy where when links from a query are sent to `asyncSaveLinks`, they are re-requested if they have expired.

This will eliminate unnecessary update requests for items in the cache that won't ever be needed again/requested.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/787)


k88hudson:
R @jaredkerim?

jaredkerim:
r+ from me, looks fantastic, just asking @sarracini to review as well because she worked on this more than I did.

sarracini:
looks good to me 👍

PR #792, opened 2016-06-03
feat(addon): set browser pref to store bigger favicons

### LINKED ISSUE #731: Show better favicons by setting optimizeToDimension preference
Even though the browser places need more work to store bigger favicons in a better way, most of the fetching and rendering work is already implemented. Setting the pref for the favicon size hard cap can be a temporary solution to be able to store better favicons.

`places.favicons.optimizeToDimension: 64`

If we set this pref to 64 (32px favicons with 2x for retina displays) we allow the browser to store bigger images that we can use with no change to our code.

rittme:
Fixes #731 

I'm not sure we need tests for this. What do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/792)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -54,6 +54,8 @@ function ActivityStreams(options = {}) {
   this._setupButton();
   NewTabURL.override(`${this.options.pageURL}#/`);
 
+  Services.prefs.setIntPref("places.favicons.optimizeToDimension", 64);
+
   this._appURLHider = new AppURLHider(this.appURLs);
   this._perfMeter = new PerfMeter(this.appURLs);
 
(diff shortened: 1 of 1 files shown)


oyiptong:
no need for tests. r+

k88hudson:
\o/

pdehaan:
@rittme How will this work? Will Firefox fetch higher resolution favicons for existing sites, or only fetch higher resolution favicons for future visited sites, or ?

rittme:
@pdehaan It will only be able to fetch higher resolution favicons for future visited sites. 
But the ones that are already in cache will expire in at most one week 😃  (as set here: https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/FaviconHelpers.h#39).

Issue #762, opened 2016-05-30
SearchProvider.jsm collects suggestions for search box


sarracini:
The search provider as of right now only does the bare minimum of getting the initial state of search and performs search. It needs to be able to get suggestions for search too

sarracini:
Fixed with #799

PR #798, opened 2016-06-07
feat(content): #765 Search drop down

sarracini:
@rlr this is the content work for the search drop down.
This closes #765 and #767 

It needs to land at the same time as #797

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/798)

PR #797, opened 2016-06-07
feat(addon): #775 Remove ContentSearch dependency

sarracini:
@rlr this is the addon-side work.
This closes #762 #763 #764 #766 #775 

@oyiptong if you want to review also feel free

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/797)

Issue #775, opened 2016-05-31
Remove SearchProvider's dependency on ContentSearch


sarracini:
Advantages to writing our own slimmed down version of ContentSearch for ActivityStream:
 * No more backwards compatibility issues
 * No more having to load in ContentSearch and all the useless stuff that comes along with it
 * No more calling into private functions like ```_onMessageGetSuggestions``` which is not what we want anyways
 * We can make it do what we want, how we want

Disadvantages:
 * Code overlap/duplication with existing ContentSearch
 * Having to maintain it, instead of just using the built ContentSearch functionality 
 * It can go away when the refactored version of ContentSearch lands in FFx

what do you think @oyiptong?

oyiptong:
We'll want to customize search results anyway. I think it's a good idea

sarracini:
Fixed with #799

PR #799, opened 2016-06-07
feat(addon): Get Search working in Activity Stream

sarracini:
This is the giant PR which got split into #797 for addon-side work and #798 for content-side work

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/799)

### DIFF
--- a/content-src/components/Search/Search.js
+++ b/content-src/components/Search/Search.js
@@ -1,27 +1,316 @@
 const React = require("react");
+const {connect} = require("react-redux");
+const {actions} = require("common/action-manager");
+const PAGE_NAME = "NEW_TAB";
 
 const Search = React.createClass({
-  getDefaultProps() {
+  getInitialState: function() {
     return {
-      onSearch: function() {}
+      focus: false,
+      activeIndex: -1,
+      activeSuggestionIndex: -1,
+      activeEngineIndex: -1,
+      searchString: this.props.searchString
     };
   },
-  doSearch(value) {
-    this.props.onSearch(value);
+  resetState: function() {
+    this.setState(this.getInitialState());
+  },
+  manageEngines: function() {
+    this.props.dispatch(actions.NotifyManageEngines());
+  },
+  setValueAndSuggestions: function(value) {
+    this.setState({activeIndex: -1, activeSuggestionIndex: -1, searchString: value});
+    this.props.dispatch(actions.NotifyUpdateSearchString(value));
+    this.props.dispatch(actions.RequestSearchSuggestions({engineName: this.props.currentEngine.name, searchString: value}));
+  },
+  getActiveSuggestion: function() {
+    const suggestions = this.props.formHistory.concat(this.props.suggestions);
+    const index = this.state.activeSuggestionIndex;
+    return (suggestions && suggestions.length && index >= 0) ? suggestions[index] : null;
+  },
+  getActiveEngine: function() {
+    const index = this.state.activeEngineIndex;
+    return (index >= 0) ? this.props.engines[index].name : this.props.currentEngine.name;
+  },
+  getSettingsButtonIsActive: function() {
+    const index = this.state.activeIndex;
+    const numSuggestions = this.props.formHistory.concat(this.props.suggestions).length;
+    const numEngines = this.props.engines.length;
+    return index === numSuggestions + numEngines;
+  },
+  getActiveDescendantId: function() {
+    // Returns the ID of the element being currently in focus, if any.
+    const index = this.state.activeIndex;
+    const numSuggestions = this.props.formHistory.concat(this.props.suggestions).length;
+    const numEngines = this.props.engines.length;
+    if (index < numSuggestions) {
+      return "search-suggestions-" + "history-search-suggestions-" + index;
+    } else if (index < numSuggestions + numEngines) {
+      return "search-partners-" + (index - numSuggestions);
+    } else if (index === numSuggestions + numEngines) {
+      return "search-settings-button";
+    }
+    return null;
+  },
+  getDropdownVisible: function() {
+    return !!(this.props.searchString && this.state.focus);
+  },
+  performSearch: function(options) {
+    let searchData = {
+      engineName: options.engineName,
+      searchString: options.searchString,
+      searchPurpose: "d"
+    };
+    this.props.dispatch(actions.NotifyPerformSearch(searchData));
+    this.props.dispatch(actions.NotifyEvent({
+      event: "SEARCH",
+      page: PAGE_NAME
+    }));
+  },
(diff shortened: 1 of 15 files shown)


content-src/components/Search/Search.js
```diff
@@ -1,27 +1,318 @@
...
+      </section>
+      <section className="search-partners" role="group">
+            <ul>
+            {engines.map(option => {
+              const icon = option.icon;
+              const active = (this.state.activeEngineIndex === enginesIdIndex);
+              return (<li key={option.name} className={active ? "active" : ""}>
+                <a id={"search-partners-" + enginesIdIndex++ } aria-selected={active}
+                      onClick={() => this.performSearch({engineName: option.name, searchString: this.getActiveSuggestion()})}>
+                <img src={icon} alt={option.name} width="16px" height="16px"/></a>
```
pdehaan: I don't think you need to specify `"px"` units here when using the `width` and `height` attributes (only in CSS).

content-src/reducers/setSearchContent.js
```diff
@@ -0,0 +1,59 @@
+/*globals require, module */
```
pdehaan: This is singular in other files (see [ESLint "Specifying Globals"](http://eslint.org/docs/user-guide/configuring#specifying-globals)), plus not 100% certain you need them, or if they're implied if you're using an ESLint `env` of [`node: true`](https://github.com/mozilla/activity-stream/blob/c4813d2de280a5a6094d3f689da1d6c3eb5eb0f6/.eslintrc#L12).

content-src/styles/variables.scss
```diff
@@ -81,9 +81,13 @@ $search-input-left-label-width: 35px;
 $search-input-border-color: #BFBFBF;
 $search-blue: #0996F8;
 $search-shadow: rgba(96, 181, 255, 0.75);
+$search-box-shadow: rgba(0, 0, 0, 0.1);
```
pdehaan: Since this is Sass, you can do `rgba(#000, 0.1)` here, or just use the `$faintest-black` variable from above:

```sass
$faintest-black: rgba($black, 0.1);
```

content-src/styles/variables.scss
```diff
@@ -81,9 +81,13 @@ $search-input-left-label-width: 35px;
...
 $search-button-grey: #FBFBFB;
 $search-button-border-color: #D4D4D4;
 $search-button-width: 64px;
+$search-text: #666;
+$search-section-background: #F7F7F7;
+$search-border-color: #EAEAEA;
```
pdehaan: Not sure if we want to use `$very-light-grey: #EBEBEB;` here since the colors are remarkably similar.
Maybe we should run something like [**colorguard**](https://www.npmjs.com/package/colorguard) to make sure we aren't using a bunch of excessive colors that are super close to other colors in the style guide.

sarracini: I agree there's a lot of colours going on here, but I took the colors from about:newtab, I think it's ok to leave it for now, to keep it consistent with the current newtab

pdehaan: Sounds great! I'll file a new bug re:colorguard and we can discuss duplicate colors there [so it isn't blocking this PR].

sarracini: thanks @pdehaan :)

sarracini:
@rlr @oyiptong if you guys want to review from this PR it might be easier to just land all in one piece so travis can be happy. Up to you guys. This'll close #762, #763, #764, #765, #766, #767, #775

oyiptong:
bug: using a non-default search engine does not fill in the query string

sarracini:
@oyiptong can you expand?

oyiptong:
Steps to Reproduce:

1. open activity stream newtab page
2. enter search term
3. wait for suggestions menu and other search engine options to pop up
4. click on a different search engine than the default

Expected:

Search occurs on selected search engine with given search terms

Observed:

Search occurs with empty search parameters

oyiptong:
bug:

hitting the back button doesn't restore search query input box value when input box is not focused

Steps to reproduce:

1. enter search string
2. unfocus input box by clicking outside
3. press the -> button
4. hit the back button

-- alternatively

1. enter search string
2. click on an alternate search engine
3. hit the back button

Expected:

Input box shows search query.

Observed:

Input box shows the default text. When the input box is clicked, the search values are seen again.

lib/SearchProvider.js
```diff
@@ -46,35 +65,210 @@ NewTabSearchProvider.prototype = {
...
-          state = yield ContentSearch._currentStateObj();
-        }
-        return state;
-      } catch (e) {
-        Cu.reportError(e);
+  /**
+   *  Sets the UI strings.
+   */
+  get searchSuggestionUIStrings() {
+    return Object.assign({}, {
```
oyiptong: Why do you need Object.assign here? And why not just:
```
let output =  {
    ...
}
return output;
```

lib/SearchProvider.js
```diff
@@ -46,35 +65,210 @@ NewTabSearchProvider.prototype = {
...
+   *  Gets the current engine - a combination of the engine name and the icon URI.
+   */
+  get currentEngine() {
+    const engine = Services.search.currentEngine;
+    const favicon = engine.getIconURLBySize(ENGINE_ICON_SIZE, ENGINE_ICON_SIZE);
+    let obj = {
+      name: engine.name,
+      iconBuffer: favicon,
+    };
+    return Object.assign({}, obj);
```
oyiptong: Why not return `obj` directly?

lib/SearchProvider.js
```diff
@@ -46,35 +65,210 @@ NewTabSearchProvider.prototype = {
...
+    const hiddenEngines = pref ? pref.split(",") : [];
+    let result =  Services.search.getVisibleEngines().filter(engine => !hiddenEngines.includes(engine.name));
+    for (let engine of result) {
+      let favicon = engine.getIconURLBySize(ENGINE_ICON_SIZE, ENGINE_ICON_SIZE);
+      state.engines.push({
+        name: engine.name,
+        iconBuffer: favicon,
+      });
+    }
+    return Object.assign({}, state);
```
oyiptong: Why make a copy here?

rlr:
If I enter text and hit enter to do the search (works great). Then hit back button, the input text is still there and the input is focus (great). But the dropdown doesn't show even if I start typing more into the input (not great).

rlr:
The arrow button and tab behavior are working as I expect them to 👍

sarracini:
I'm going to file the follow up ticket for the back button bug, so this can land without blocking on that

rlr:
Sounds reasonable to me! I don't think I ever go back to the newtab page.

PR #801, opened 2016-06-07
Add new context menu

### LINKED ISSUE #701: Clicking on "X" should just dismiss site from activity stream, not give me a bunch of options
I wish clicking "X" would just dismiss the site from highlights, rather than force me to make a choice on "removing from history" or "never to show on this page"

screen shot 2016-05-14 at 3 28 23 pm

If I visited the site again, I might want to see it again (though the fact that I just dismissed it is a strong hint that I might not)! It just might not be relevant now. I imagine for a non-technical audience these options are going to be even more confusing. In any case, I think forcing me to take some kind of destructive action to remove those sites/links is not a great user experience, at least for me.

k88hudson:
This patch adds a new context menu which includes

- adding/removing bookmarks
- opening a link in a new window
- deleting/blocking
- tests, tests, tests

The add bookmark and open window APIs are in separate commits for easier reviewing.

TODO (in a follow-up patch):
- more sophisticated blocking 
- add icons

Fixes #701

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/801)

### DIFF
--- a/content-src/components/LinkMenu/LinkMenu.js
+++ b/content-src/components/LinkMenu/LinkMenu.js
@@ -1,0 +1,106 @@
+const React = require("react");
+const {connect} = require("react-redux");
+const ContextMenu = require("components/ContextMenu/ContextMenu");
+const {actions} = require("common/action-manager");
+const {FIRST_RUN_TYPE} = require("lib/first-run-data");
+
+const LinkMenu = React.createClass({
+  getDefaultProps() {
+    return {
+      visible: false,
+      allowBlock: true
+    };
+  },
+  userEvent(event) {
+    const {page, source, index, Experiments, dispatch} = this.props;
+    if (page && source) {
+      let payload = {
+        event,
+        page: page,
+        source: source,
+        action_position: index
+      };
+      if (["BLOCK", "DELETE"].includes(event) && Experiments.data.reverseMenuOptions) {
+        payload.experiment_id = Experiments.data.id;
+      }
+      dispatch(actions.NotifyEvent(payload));
+    }
+  },
+  getOptions() {
+    const {site, allowBlock, Experiments, dispatch} = this.props;
+    const isNotDefault = site.type !== FIRST_RUN_TYPE;
+
+    // Don't add delete options for default links
+    // that show up if your history is empty
+    const deleteOptions = isNotDefault ? [
+      {type: "separator"},
+      allowBlock && {
+        ref: "dismiss",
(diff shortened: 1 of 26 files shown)


content-src/components/LinkMenu/LinkMenu.js
```diff
@@ -0,0 +1,106 @@
...
+  userEvent(event) {
+    const {page, source, index, Experiments, dispatch} = this.props;
+    if (page && source) {
+      let payload = {
+        event,
+        page: page,
+        source: source,
+        action_position: index
+      };
+      if (["BLOCK", "DELETE"].indexOf(event) > -1 && Experiments.data.reverseMenuOptions) {
```
rlr: nit: `[].includes()` is a little more readable

content-src/components/LinkMenu/LinkMenu.js
```diff
@@ -0,0 +1,106 @@
...
+  },
+  getOptions() {
+    const {site, allowBlock, Experiments, dispatch} = this.props;
+    const isNotDefault = site.type !== FIRST_RUN_TYPE;
+
+    // Don't add delete options for default links
+    // that show up if your history is empty
+    const deleteOptions = isNotDefault ? [
+      {type: "separator"},
+      allowBlock && {
```
rlr: whoa, you can do that? I guess it skips `false` when rendering the menu.

content-src/components/LinkMenu/LinkMenu.js
```diff
@@ -0,0 +1,106 @@
...
+        userEvent: "OPEN_NEW_WINDOW",
+        onClick: () => dispatch(actions.NotifyOpenWindow({url: site.url}))
+      },
+      {
+        ref: "openPrivate",
+        label: "Open in a Private Window",
+        userEvent: "OPEN_PRIVATE_WINDOW",
+        onClick: () => dispatch(actions.NotifyOpenWindow({url: site.url, isPrivate: true}))
+      }]
+      .concat(deleteOptions).filter(o => o);
```
rlr: oh I see, you filter out the `false`s here 👍

rlr:
An issue I found is that Top Sites always seem to show "Bookmark" regardless of if it is a bookmark or not. I don't see the bug in your code though. Guessing it is an issue with the query on the addon side?

rlr:
Otherwise, looks awesome! If it's an issue with the places provider query, feel free to spinoff a new issue.

Issue #774, opened 2016-05-31
Bug 1276932 - Infinite scrolling is not working after zooming out the Timeline page, if the scrollbar is not available


SoftVision-PaulOiegas:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1276932
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.18

[Affected Platforms]:
- All Windows
- All Mac OS X
- All Linux

[Prerequisites]:
- Latest Activity Stream add-on version (v1.0.18) was previously installed on an old profile.
- Have enough History and Bookmarks in the profile so when scrolling is performed, a new batch of History items is listed.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites. 
2. Open Activity Stream Timeline from the browser toolbar.
3. Zoom out the Timeline page at lest 50%.
4. Refresh the Timeline page.

[Expected result]:
- Websites that were not displayed in viewport are listed while scrolling.

[Actual result]:
- Websites that were not displayed in viewport are not listed while scrolling because the scrollbar is not available.

[Notes]:
- This issue is reproducible when after zooming out the scroll bar is no longer displayed.
- On 4k monitors, this issue is reproducible instantly. When accessing Activity Stream Timeline there is no scroll bar and the user cannot load more visited websites while scrolling. 
- Attached a screen recording of the issue (https://goo.gl/01eYdy).

SoftVision-VladBacia:
This issue was verified on Windows 8.1 x64 with latest Nightly (49.0a1, Build ID: 20160602030220) using Activity Stream v1.0.18-today build, and it is still reproducible.

[Prerequisites]:
- The browser window is in Fullscreen mode (1920x1080).
- Open a New Tab page and set the zoom level to 30%.

[STR]:
1. Click on Activity Stream button from the browser toolbar.
2. Scroll through the Timeline History items.
3. Click on Bookmarks button from the Timeline sidebar.
4.  Scroll through the Activity Stream Bookmark History items.

[Expected]:
- When scroll action is performed, more items are loaded.

[Actual]:
- When opening the Activity Stream Timeline, two batches of items are displayed but they are not enough to fill page.
- When switching between pages (Bookmarks and Timeline) a refresh should be automatically made, so the new items will be auto displayed. Every switch between pages requires a manual refresh in order to bring two batches of items but they are not enough.

[Notes]:
- [Here](https://goo.gl/1wXpj4) is a video of this behavior.

[...]

PR #804, opened 2016-06-08
fix (content): #774 more fixes related to more data on zoom/scroll/resizing

rlr:
This fixes a couple more of cases:
* switching between routes
* when more than two pages fit on the screen - it was only loading up to two pages

Also I renamed the scroll handlers since they aren't only called during scrolling now. Open to better suggestions.

r? @k88hudson

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/804)

### DIFF
--- a/content-src/components/TimelinePage/TimelineFeed.js
+++ b/content-src/components/TimelinePage/TimelineFeed.js
@@ -24,7 +24,7 @@ const TimelineFeed = React.createClass({
     }));
   },
   windowHeight: null,
-  handleScroll(values) {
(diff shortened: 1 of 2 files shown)


k88hudson:
The only issue with this is that we will also do the check every time someone deletes something etc.... although I guess that's ok, since we're debouncing and if the scroll positions are right nothing will happen anyway, right?

rlr:
yeah. There are some cases where we will waste a few extra cycles doing a little math but that's it.

Issue #808, opened 2016-06-08
Having intermittent building errors on TravisCI


ncloudioj:
@sarracini has noticed that there are intermittent test failures on TravisCI. In those tests, the "load_reason" in the session payload could be either "newtab" or "back_button" at random, it's not reproducible in the dev environment though.

I suspect the assumption that the "pageshow" event is emitted after the "ready" event is not always the case on certain platforms. So, if the "pageshow" event gets fired prior to the "ready" one, or its event handler gets executed earlier, the "load_reason" will be set as "back_button" by mistake when the user opens a new tab.

To work around this issue, when the user clicks on the "back" button to get back to the default Activity Stream page, we can simply start a new session with "load_reason" set as "newtab" rather than "back_button".

ncloudioj:
Fixed in #809

PR #809, opened 2016-06-08
fix(addon): #808 fix uncertain load_reason for the back button

ncloudioj:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/809)


sarracini:
looks good to me, is it possible to just put a comment in the code as to why we're doing this now?

ncloudioj:
Thanks @sarracini 

Good idea, will drop a comment in the code.

Issue #843, opened 2016-06-13
Bug 1279914 - Keywords written in the New Tab Search Bar cannot be deleted using the "Delete" key


SoftVision-VladBacia:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1279914
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.1.0-dev

[Affected Platforms]:
- All Windows
- All Mac OS X
- All Linux

[Prerequisites]:
- Latest Activity Stream add-on version (1.1.0-dev) was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites. 
2. Open a New Tab and write a keyword in the Search bar.
3. Select the keyword and press the keyboard "Delete" button.

[Expected result]:
- The selection is deleted.

[Actual result]:
- The "Delete" keyboard button does not delete the selected string. 

[Notes]:
- [Here](https://goo.gl/2T8J9E), is a screen recording of the issue.

sarracini:
Fixed with #866

SoftVision-VladBacia:
This issue has been verified on Win 10 x64, Mac OS X 10.11 and Ubuntu 14.04 x64 with latest Nightly 50.0a1 (Build ID: 20160620030215) using latest Activity Stream v1.1.0-custom build and could not reproduce it. 
Thus, marking this issue as Verified-Fixed.

PR #866, opened 2016-06-20
fix(content): #843 Fix delete key bindings for search

sarracini:
@rlr r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/866)


rlr:
I wasn't really able to reproduce the video. But this makes sense to me. R+

Issue #849, opened 2016-06-14
Bug 1279983 - Default browser home page is displayed after opening the browser


SoftVision-PaulOiegas:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1279983
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.1.0-dev

[Affected Platforms]:
- All Windows
- All Mac OS X
- All Linux

[Prerequisites]:
- Latest Activity Stream add-on version (1.1.0-dev) was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites. 
2. Observe the Home Page tab.

[Expected result]:
- The Activity Stream New Tab page is displayed because it has replaced the default browser home page.

[Actual result]:
- The default browser Home Page page is displayed instead of Activity Stream's New Tab page.

[Notes]:
- This issue is related to #724.
- Attached a screen recording of the issue https://goo.gl/l79tnp.

rlr:
The issue here is that when the first window opens on Firefox startup, the addon hasn't (fully) loaded yet to replace the home page. I'll see if I can tweak the unload code so that we only restore the home page back to the original when the addon is getting disabled.

rlr:
fixed via #878 

Note that when starting up the browser, you'll be able to see `resource://..blablabla...` in the awesome bar until the addon loads and hides it. Can't really fix that until we ship for real in firefox.

SoftVision-VladBacia:
This issue has been verified on Win 10 x64, Mac OS X 10.11 and Ubuntu 15.10 x64 with latest Nightly 50.0a1 (Build ID: 20160622030210) using latest Activity Stream v1.1.0-custom build and could not reproduce it.
Thus, marking this issue as Verified-Fixed.

PR #878, opened 2016-06-23
fix (addon): #849 only reset back home page on disable

rlr:
r? @sarracini or @oyiptong 

Basically, we don't reset the home page back on unload unless the user is disabling (uninstalling).

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/878)


oyiptong:
r+

Issue #848, opened 2016-06-14
Keystrokes Lost in Search Bar


kdoske:
As I type random keystrokes appear to be instantly backspacing on there own and my next key stroke overwrites the previous letter. Its not every key stroke but probably ever 3rd one or so. I had to turn it off. It was driving me mad.

sarracini:
I don't actually know if this can be fixed...it can be a result of a lot of things but most likely because message passing from content to addon happens when the a new key stroke happens and it needs to fetch the new suggestions - if it's slow this might be a result of that.
Not sure how we can fix that

wlusink:
Right now I think I can see that the suggestion code is re-writing the contents of the text box? I don't see why it should do that. If I am just typing a character in that instance, my character is overwritten with the older value from the suggestion code.

I think the suggestion-code should never write anything in the textbox. It should only read.

rlr:
fixed via pull #879

SoftVision-VladBacia:
This issue has been verified on Win 8.1 x64, Mac OS X 10.11 and Ubuntu 14.04 x64 with latest Nightly 50.0a1 (Build ID: 20160626030213) using latest Activity Stream v1.1.0-custom build and could not reproduce it. The fix will be available in the next release.
Thus, marking this issue as Verified-Fixed.

PR #879, opened 2016-06-23
fix (content): #848 fix fast typing issue with search box

rlr:
This bit me a couple times so looked into it... @sarracini what do you think of this fix? We already update `searchString` when handling the `NOTIFY_UPDATE_SEARCH_STRING` action so we shouldn't need to update it again when suggestions come back from the addon side. It seems to fix the issue for me.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/879)


sarracini:
works for me! thanks for this ricky, it was a good catch :)
r+

Issue #859, opened 2016-06-17
Bug 1280457 - "Delete from History" should not be available in Bookmarks screen


SoftVision-PaulOiegas:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1280457
<hr>
Affected versions]:
- Firefox 45.0 and up
- Activity Stream v1.1.0

[Affected Platforms]:
- All Windows
- All Linux
- All Mac OS

[Prerequisites]:
- Latest Activity Stream add-on (v1.1.0) was previously installed on a profile.
- Have more bookmarked websites.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Click the Activity Stream button from the browser toolbar.
3. Click on "Bookmarks" button from the Timeline sidebar.
4. Hover a bookmarked website and click on "..." context menu button.
5. Click on "Delete from History" button.

[Expected result]:
- "Delete from History" button is not available in Bookmarks History page.

[Actual result]:
- The bookmarked website is no longer displayed in Bookmarks History page. After a page refresh the website will be redisplayed.

[Notes]:
- Attached a screen recording of the issue https://goo.gl/AnO8yB.

sarracini:
Fixed with #883

SoftVision-VladBacia:
This issue has been verified on Win 8.1 x64, Mac OS X 10.11 and Ubuntu 14.04 x64 with latest Nightly 50.0a1 (Build ID: 20160627030215) using latest Activity Stream v1.1.0-custom build and could not reproduce it.
Thus, marking this issue as Verified-Fixed.

PR #883, opened 2016-06-24
fix(content): #859 'Delete from History' should not be available for bookmarks

sarracini:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/883)


k88hudson:
R+

PR #886, opened 2016-06-27
Experiment definition framework

k88hudson:
This patch adds a experiment definition framework in the form of a json file.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/886)

### DIFF
--- a/lib/ExperimentProvider.js
+++ b/lib/ExperimentProvider.js
@@ -1,0 +1,46 @@
+const seedrandom = require("lib/vendor.bundle").SeedRandom;
+
+exports.ExperimentProvider = class ExperimentProvider {
+  constructor(clientID, experiments, rng) {
+    experiments = experiments || require("experiments.json");
+    this._clientID = clientID;
+    this._rng = rng || seedrandom(clientID);
+    this._data = {};
+    this._experimentId = null;
+
+    const randomNumber = this._rng();
+    let floor = 0;
+
+    Object.keys(experiments).forEach(key => {
+      const experiment = experiments[key];
+      if (experiment.active === false) {
+        return;
+      }
+      const {variant, control} = experiment;
+      const ceiling = variant.threshold + floor;
+      if (ceiling > 1) {
+        throw new Error("Your variant cohort sizes should add up to less than 1.");
+      }
+      const inExperiment = randomNumber >= floor && randomNumber < ceiling;
+      if (inExperiment) {
+        this._experimentId = variant.id;
+      }
+      this.data[key] = {
+        value: inExperiment ? variant.value : control.value,
+        inExperiment
+      };
+      floor = ceiling;
+    });
+  }
+
+  // This is an object representing all experiments
+  get data() {
+    return this._data;
+  }
+
+  // This returns null if the user is part of a control group,
+  // or an id indicating the experiment/variant if they are part of it.
+  get experimentId() {
+    return this._experimentId;
+  }
+};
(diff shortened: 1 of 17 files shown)


test/test-ExperimentProvider.js
```diff
@@ -0,0 +1,105 @@
...
+    foo: {id: "foo_01", value: 42, inExperiment: false}
+  }, "clientID 'baz' should result in control being picked");
+
+  setup("012j");
+  assert.deepEqual(experimentProvider.data, {
+    foo: {id: "foo_01", value: 84, inExperiment: true}
+  }, "clientID '012j' should result in variant being picked");
+};
+
+// exports["test ExperimentProvider.validateExperiment"] = assert => {
```
k88hudson: `assert.throws` seems to intermittently fuck up the tests, I get this output sometimes:

```
  warnings and/or errors were logged.

26 of 27 tests passed.
There were test failures...

The following tests failed:


```

i.e., it causes a failure, but no tests actually failed?

k88hudson:
@ncloudioj want to take a look? I wrote some documentation in https://github.com/k88hudson/activity-stream/blob/fa451804114b2dca71ea823462aa3ea006320b82/experiments_how_to.md

k88hudson:
@jaredkerim suggested that maybe we just automatically add the cohort/experiment ID to every ping instead of requiring a specific code path for it/a whitelist. what do you think @ncloudioj?

lib/ExperimentProvider.js
```diff
@@ -0,0 +1,59 @@
...
+    this._rng = seedrandom(clientID);
+    this._data = {};
+    Object.keys(experiments).forEach(key => {
+      const experiment = experiments[key];
+      this.validateExperiment(experiment, key);
+      if (experiment.active === false) {
+        return;
+      }
+      const {id, variant, control} = experiment;
+      const inExperiment = this._rng() <= variant.threshold;
```
ncloudioj: A single threshold is fine for the single experiment case, but we need a better experiment user selection mechanism to support those multi-experiment scenarios.

For instance, say we have three independent experiments - e1, e2, e3 running at the same time, each of them is targeting 10% of the addon users, respectively. Ideally, the whole user base should be divided into four groups as follows,
* control group, if 0 <= `rng` < 0.7
* e1 group, if 0.7 <= `rng` < 0.8
* e2 group, if 0.8 <= `rng` < 0.9
* e3 group otherwise

lib/ExperimentProvider.js
```diff
@@ -0,0 +1,59 @@
...
+    if (!experiment.variant || typeof experiment.variant !== "object") {
+      errors.push("missing variant, which should be an object");
+    }
+    if (experiment.variant && typeof experiment.variant.value === "undefined") {
+      errors.push("missing variant.value");
+    }
+    if (experiment.variant && typeof experiment.variant.threshold !== "number") {
+      errors.push("missing variant.threshold, which should be a number");
+    }
+    if (experiment.variant && experiment.variant.threshold > 1) {
```
ncloudioj: also, threshold > 0

ncloudioj:
LGTM except a few nits. This framework largely simplifies the work of experimenting. Awesome work!!!

k88hudson:
OK I added:
* automatic `experiment_id` added to ping
* tests/validator that will throw if the total control cohort is less than 20%
* moved `id` to variant, so if we want to add multiple variants in the future, this is easy to support

ncloudioj:
r+ :shipit:

Issue #838, opened 2016-06-10
Bug 1279511 - Search keyword is removed after navigating with the arrow keys through the search suggestion dropdown


SoftVision-PaulOiegas:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1279511
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.19-dev

[Affected Platforms]:
- All Windows
- All Linux
- All Mac OS

[Prerequisites]:
- Latest Activity Stream add-on version (1.0.19) was previously installed on a profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Open a New Tab so the Activity Stream New Tab page is displayed.
3. Enter a keyword in the Search text box.
4. Navigate through the search suggestion dropdown with "down arrow" key from keyboard.
5. Return to the Search text box using the "up arrow" key from keyboard.

[Expected result]:
- The keyword written in the Search text box is displayed

[Actual result]:
- When returning to the Search text box using the "up arrow" key from keyboard, the keyword written is removed.

[Notes]:
- A video describing the issue was attached https://goo.gl/07ZYY9.

SoftVision-CosminMuntean:
While testing the latest Activity Stream add-on (v1.1.0-custom build) we observed that when navigating through the search suggestion dropdown with "down arrow" or "up arrow" from keyboard, and after that hovering with the mouse cursor on the suggestion keywords, two elements will be highlighted (one will be highlighted from keyboard and one from mouse cursor).
<hr>
![two elements highlighted](https://cloud.githubusercontent.com/assets/19226534/16231700/120f1674-37d0-11e6-88cd-b1959bba4108.png)

sarracini:
Fixed with #890

SoftVision-VladBacia:
The issue described above by @SoftVision-CosminMuntean is still reproducible. If an item from the search suggestion dropdown list is hovered with the mouse cursor, it will be highlighted. While the item is highlighted press the "down arrow" or "up arrow" from keyboard. In this case there will be 2 items highlighted.
In `about:newtab` page when "down arrow" or "up arrow" button are pressed, only the last selected item is highlighted.

sarracini:
I'm going to file a separate issue for that one

SoftVision-VladBacia:
[...]

PR #890, opened 2016-06-28
fix(content): #838 Search keyword is removed after navigating with the arrow keys through the search suggestion dropdown

sarracini:
the onMouseMove stuff is to prevent the double selection stuff that was going on

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/890)

### DIFF
--- a/content-src/components/Search/Search.js
+++ b/content-src/components/Search/Search.js
@@ -122,6 +122,17 @@ const Search = React.createClass({
       case "ArrowUp":
         if (index > -1) {
           newIndex--;
+
+          // if we've reached the top of the suggestions list and press 'up' once more,
+          // we keep the table visible but display the origial search string. This
+          // behaviour is meant to match about:newtab
+          if (newIndex === -1) {
+            this.setState({focus: true});
+            this.props.dispatch(actions.NotifyUpdateSearchString(originalSearchString));
+            newSuggestionIndex = -1;
+            newIndex = -1;
+            break;
+          }
           if (index < numSuggestions) {
             // We are in suggestions, move on up.
             newSuggestionIndex--;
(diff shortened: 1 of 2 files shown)


content-src/components/Search/Search.js
```diff
@@ -122,6 +122,17 @@ const Search = React.createClass({
...
+          // if we've reached the top of the suggestions list and press 'up' once more,
+          // we keep the table visible but display the origial search string. This
+          // behaviour is meant to match about:newtab
+          if (newIndex === -1) {
+            this.setState({focus: true});
+            this.props.dispatch(actions.NotifyUpdateSearchString(originalSearchString));
+            newSuggestionIndex = -1;
+            newIndex = -1;
+            break;
+          }
```
rlr: 👍 WFM

content-src/components/Search/Search.js
```diff
@@ -229,6 +240,13 @@ const Search = React.createClass({
       activeEngineIndex: newEngineIndex
     });
   },
+
+  onMouseMove(newIndex) {
+    this.setState({
+      activeIndex: newIndex,
+      activeSuggestionIndex: newIndex,
+    });
+  },
```
rlr: this too 👍

content-src/components/Search/Search.js
```diff
@@ -259,8 +277,10 @@ const Search = React.createClass({
           {formHistory.map(suggestion => {
             const active = (this.state.activeSuggestionIndex === suggestionsIdIndex);
             const activeEngine = this.getActiveEngine();
-            return (<li key={suggestion} role="presentation">
+            const suggestionIndex = suggestionsIdIndex++;
+            return (<li key={suggestion} role="option">
                   <a id={"history-search-suggestions-" + suggestionsIdIndex++ }
```
rlr: do you need to remove the `++` here since you are doing it two lines above? I think you are incremementing it twice per row but it's like impossible to inspect in dev tools so I am not sure.

rlr: Actually, I was able to see it in the inspector and it looks fine.

rlr: Gah, I was looking at the wrong thing. I see they go `*-1`, `*-3`, and so on. So it should probably be changed to look like line 298 below?

rlr:
r+ with that history index fix

content-src/components/Search/Search.js
sarracini: good catch! will do :)

Issue #674, opened 2016-05-11
Bug 1271941 - Content is not automatically displayed for "Timeline" and "New Tab" pages when `remember windows and tabs from last time` option is set in `about:preferences`


SoftVision-VladBacia:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1271941
<hr>
[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.16

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- Create a new profile.
- Install Activity Stream through Test Pilot.
- Browse a few links as a new user (50-100 visited websites, 1 bookmarked website).

[Steps to reproduce]:
1. Go to about:preferences->General and select "When Firefox starts: Show my windows and tabs from last time".
2. Click on "Activity Stream" button from browser toolbar.
3. Close all other tabs.
4. Close Firefox.
5. Start Firefox.

[Expected results]:
- Activity Stream Timeline content is automatically displayed with all items in viewport.

[Actual results]:
- For a few seconds the Mozilla start page is displayed and then it is replaced by Activity Stream Timeline. Activity Stream Timeline is empty although there is content in history. Even the Highlight tiles are hidden. Resource link is displayed in the browser address bar.
- If we visit the Bookmarks page from Activity Stream Timeline, we can see that this page is also empty (1 website is bookmarked).
- Content is displayed if the user performs a manual refresh of the page.

[Notes]:
- If we repeat the same scenario for New Tab page, we can observe that the loading spinner is displayed indefinitely with the message "Hang on tight! We are analyzing your history to personalize your experience", until the user refreshes the page.
- Videos are available [here](https://drive.google.com/open?id=0B6QogV2AJFKiSXNWUkpZdDdsMlk)

rlr:
I've been unable to reproduce this on dev edition. I'll follow the exact STR to see if I can get it to happen. Otherwise this might be fixed already.

SoftVision-CosminMuntean:
This issue was retested on the latest Nightly (49.0a1) build with Activity Stream v1.0.18 but it is reproducible only with e10s disabled.

The issue is reproducible also with this scenarios:
[Prerequisites]:
- Latest Activity Stream add-on version (1.0.18) was previously installed on a old profile.
- Have more history so that the "Highlights" area is populated with visited websites.
- Make sure that e10s is disabled.

[STR]:
1. Click on Activity Stream button from browser toolbar.
[...]

Issue #816, opened 2016-06-09
Bug 1279257 - Timeline /New Tab page content is lost at refresh if the page tab is dragged and opened in a new window when e10s is enabled


SoftVision-PaulOiegas:
via https://bugzilla.mozilla.org/show_bug.cgi?id=1279257

<hr>

[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.0.19-dev

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- Firefox has e10s enabled. 
- Latest Activity Stream add-on version (1.0.19-dev) was previously installed on a clean profile.

[Steps to reproduce]:
1. Start Firefox browser with the profile where Activity Stream add-on was previously installed.
2. Open a "Timeline" page from browser toolbar or "New Tab" page.
3. Click and hold the opened page tab and drag it on the screen to open it in a new window.
4. Refresh the page on the new window and observe its behavior.

[Expected results]:
The page content is correctly displayed after refresh.

[Actual results]:
The page content disappears after the refresh and it's never displayed back.

[Notes]:
- This issue only reproduces with e10s enabled.
- The content is displayed back only when a new page of its kind is opened.
- The "New Tab" page displays the "Hang on tight! We are analyzing your history to personalize your experience" message for indefinite time.
- Attached a screen recording of the issue https://goo.gl/dL5Fu4 .

sarracini:
This looks related to #674, similar behaviour is happening

SoftVision-PaulOiegas:
No luck with this one. Still reproducible after #891 was merged. So there must be something else here since #674 now correctly works.

sarracini:
Closing this as this is no longer reproducible

PR #896, opened 2016-06-30
fix (content): #891 wait for content bridge to be setup before render

### LINKED ISSUE #891: first run after addon update results in 'hang on' message forever
1. use a profile with an average amount of history and data.
2. install dev channel activity stream
3. upgrade via about:addons UI or automatically after 24hr ping
4. launch new tab

actual: the view displays 'hang on tight' for a long long time.  work around is refresh the browser.

expected: it should only sit on 'hang on tight for maybe 5-10sec.

rlr:
There are some cases where we end up dispatching actions before the content bridge is set up, so the addon side never receives the initial messages to get data. Then the actions time out and the user is stuck hanging tight...

It only seems to happen in some cases when firing up Firefox with the Activity Stream page open. Happens more with e10s disabled. See issue for more details. I was able to reliably reproduce by having Firefox open my home page on startup. I would install the addon, make sure it was working, close Firefox, reopen Firefox, it still worked, close Firefox, reopen Firefox and then it was busted. Obviously it was a painful troubleshooting/debugging session.

r? @k88hudson 

Fixes #891

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/896)

### DIFF
--- a/content-src/main.js
+++ b/content-src/main.js
@@ -17,4 +17,13 @@ const Root = React.createClass({
   }
 });
 
-ReactDOM.render(<Root />, document.getElementById("root"));
+function renderRootWhenAddonIsReady() {
+  if (window.navigator.activity_streams_addon) {
+    ReactDOM.render(<Root />, document.getElementById("root"));
+  } else {
+    // If the content bridge to the addon isn't set up yet, try again soon.
+    setTimeout(renderRootWhenAddonIsReady, 50);
+  }
+}
+
+renderRootWhenAddonIsReady();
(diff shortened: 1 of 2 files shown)


sarracini:
This will actually also fix #674 and possibly also #816 🎉

oyiptong:
hopefully those issues are timing issues and not just system restore not sending the events

k88hudson:
Kinda gross 🔨, but it works! LGTM, R+

Issue #863, opened 2016-06-20
Persist the page's meta data


ncloudioj:
This is the meta bug for tracking the persistence of page's meta data, which is currently generated by embedly. The plan is to use a separate SQLite db as the data store, and to provide various "join" queries against the "Places" database.

jaredkerim:
This is a duplicate of #770, however this can now be considered the canonical tracking ticket for this effort and we will close #770.

ncloudioj:
Thanks @jaredkerim! Let's use this one to track the design and implementation of #770

k88hudson:
These are all the fields we currently use in our UI:

- url (provided by places)
- title (provided by places)
- description (provided by embedly)
- favicon url (provided by embedly, falls back to placesDB)
- favicon color as RGB (we compute this locally using the favicon in placesDB) 
- image url (from embedly)
- video/media url (from embedly)

k88hudson:
cc @jaredkerim @sarracini

pdehaan:
Is this related to https://github.com/mozilla/loci in any way?

Should we be capturing the page's full text from Fathom?

Also, should we include some TTL in each record so we can rescrape pages every X days (similar to Embedly's `cache_age`)?

> `cache_age` &mdash; How long Embedly is going to `cache` the response for? Generally, this is for a day, unless some external factor tells us to reevaluate the resource.
> 
> &mdash; via http://docs.embed.ly/docs/extract ([example](http://embed.ly/docs/explore/extract?url=http%3A%2F%2Ftechcrunch.com%2F2013%2F03%2F26%2Fembedly-now-goes-beyond-embedding-with-new-products-extract-display-for-making-sense-of-links-resizing-images%2F))

ncloudioj:
These are all the fields we currently use in our UI, plus the feedback from @jaredkerim @sarracini :

- url (provided by places)
- title (provided by places)
- type (provided by embedly)
- description (provided by embedly)
- favicon url (provided by embedly, falls back to placesDB)
- favicon color as RGB (we compute this locally using the favicon in placesDB) 
- image url (from embedly)
- video/media url (from embedly)
- canonical url (provider_url from embedly)
- cache_key url (generated by the addon) 
- expire_at (a timestamp)

pdehaan:
[...]

PR #899, opened 2016-07-07
Metadata store

ncloudioj:
This is the first cut of metadata persistence implementation #863. It now supports:
* insertion of metadata, images(favicon and preview), and relations of those two
* table dropping
* data expiry (for metadata only, not for images yet)
* executeQuery (raw SQL statements)

To be added shortly:
* more specific query APIs (once we know how we want to query against those tables)
* data expiry for images

Also, table schema is open to change anytime.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/899)

### DIFF
--- a/lib/MetadataStore.js
+++ b/lib/MetadataStore.js
@@ -1,0 +1,380 @@
+ /* globals XPCOMUtils, Task, OS, Sqlite, PlacesUtils, NetUtil, Services */
+"use strict";
+
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Sqlite.jsm");
+Cu.import("resource://gre/modules/NetUtil.jsm");
+Cu.import("resource://gre/modules/Timer.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "Task",
+                                  "resource://gre/modules/Task.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PlacesUtils",
+                                  "resource://gre/modules/PlacesUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "OS",
+                                  "resource://gre/modules/osfile.jsm");
+
+const METASTORE_NAME = "metadata.sqlite";
+
+const SQL_DDLS = [
+  `CREATE TABLE IF NOT EXISTS page_metadata (
+      id INTEGER PRIMARY KEY,
+      cache_key LONGVARCHAR UNIQUE,
+      places_url LONGVARCHAR,
+      title TEXT,
+      type VARCHAR(32),
+      description TEXT,
+      media_url LONGVARCHAR,
+      created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
+      expired_at LONG
+  )`,
+  `CREATE TABLE IF NOT EXISTS page_images (
+    id INTEGER PRIMARY KEY,
+    url LONGVARCHAR UNIQUE,
+    type INTEGER,
+    height INTEGER,
+    width INTEGER,
+    color VARCHAR(32)
+  )`,
+  `CREATE TABLE IF NOT EXISTS page_metadata_images (
+    metadata_id INTEGER,
+    image_id INTEGER,
+    FOREIGN KEY(metadata_id) REFERENCES page_metadata(id) ON DELETE CASCADE,
+    FOREIGN KEY(image_id) REFERENCES page_images(id) ON DELETE CASCADE
+  )`,
+  "CREATE UNIQUE INDEX IF NOT EXISTS page_metadata_cache_key_uniqueindex ON page_metadata (cache_key)",
+  "CREATE UNIQUE INDEX IF NOT EXISTS page_images_url_uniqueindex ON page_images (url)"
+];
+
+const SQL_LAST_INSERT_ROWID = "SELECT last_insert_rowid() AS lastInsertRowID";
+
+/**
+ * A place holder for the database migrations in the future
+ */
+const SQL_MIGRATIONS = [];
+
+const SQL_INSERT_METADATA = `INSERT INTO page_metadata
+  (cache_key, places_url, title, type, description, media_url, expired_at)
+  VALUES
+  (:cache_key, :places_url, :title, :type, :description, :media_url, :expired_at)`;
+
+const SQL_INSERT_IMAGES = `INSERT INTO page_images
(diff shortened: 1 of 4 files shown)


ncloudioj:
@oyiptong Could you r? it please?

@jaredkerim @sarracini Could you also double check the table schema and APIs? Thanks a lot.

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+  this._conn = null;
+  this._cron = null;
+}
+
+MetadataStore.prototype = {
+  /**
+   * Creates the table schema for the metadata database. This function must be called
+   * after a database connection has been established
+   */
+  _createTableSchema() {
```
oyiptong: nit: use `Task.async` instead of `return Task.spawn`

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+  },
+
+  /**
+  * Convert a RGB array to the Hex form, e.g. [10, 1, 2] => #0A0102
+  */
+  _rgbToHex(rgb) {
+    return "#" + ((1 << 24) + (rgb[0] << 16) + (rgb[1] << 8) + rgb[2]).toString(16).slice(1).toUpperCase();
+  },
+
+  _getLastInsertRowID() {
```
oyiptong: nit: use `Task.async`

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+    return Task.spawn(function*() {
+      let result = yield this._conn.execute(SQL_LAST_INSERT_ROWID);
+      return result[0].getResultByName("lastInsertRowID");
+    }.bind(this));
+  },
+
+  _getImageIDByURL(url) {
+    return Task.spawn(function*() {
```
oyiptong: nit: use Task.async

lib/MetadataStore.js
oyiptong: And also rename the function to `_asyncCreateTableSchema`

lib/MetadataStore.js
oyiptong: and please rename function

lib/MetadataStore.js
oyiptong: and please rename function

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+   * @param {Object} [optional] aOptions
+   *        aOptions.columns - an array of column names. if supplied the return
+   *        items will consists of objects keyed on column names. Otherwise
+   *        array of raw values is returned in the select order
+   *        aOptions.param - an object of SQL binding parameters
+   *        aOptions.callback - a callback to handle query raws
+   *
+   * Returns a promise with the array of retrieved items
+   */
+  executeQuery(aSql, aOptions = {}) {
```
oyiptong: nit: use `Task.async`. The function with a similar name in `PlacesProvider` could use some refactoring too

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+      });
+      if (queryError) {
+        throw new Error(queryError);
+      }
+      return items;
+    }.bind(this));
+  },
+
+  /**
+  * Enables the data expiry cron. The database connection needs to
```
oyiptong: s/cron/job/g

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    this._cron = setInterval(() => {
+      this._conn.execute(SQL_DELETE_EXPIRED).catch(error => {
+        Cu.reportError("Failed to delete expired metadata: " + error.message);
+      });
+    }, interval);
+  },
+
+  disableDataExpiryCron() {
```
oyiptong: 1,$s/[Cc]ron/job/g

lib/MetadataStore.js
```diff
@@ -0,0 +1,385 @@
...
+  asyncInsert: Task.async(function*(aRows) {
+    const principal = Services.scriptSecurityManager.getSystemPrincipal();
+    for (let row of aRows) {
+      yield this._conn.executeTransaction(function*() {
+        let metadata_id;
+        let image_ids = [];
+
+        /* attach favicon to canonical_url for Places */
+        PlacesUtils.favicons.setAndFetchFaviconForPage(
+          NetUtil.newURI(row.canonical_url),
```
oyiptong: how is `canonical_url` obtained?

ncloudioj: Good catch, it's a bug. The `canonical_url` should be obtained from the embed.ly response payload for now.

@jaredkerim can you confirm that which field in the current embed.ly payload could be used as the `canonical_url`? `provider_url`, `original_url`, or `url`?

I recall you mentioned aobut `provider_url` last week, but it does not look like that way to me.

ncloudioj: Having confirmed with @jaredkerim, the `canonical_url` field will be obtained from the Places, and the PreviewProvider will take care of populating it.

oyiptong: IIUC `canonical_url` == `places.url`?

lib/MetadataStore.js
```diff
@@ -0,0 +1,380 @@
...
+
+  /**
+   * Inserts the metadata into the database. It consists of two tables,
+   * i.e. page_metadata for the regular meta information of the page, and
+   * page_images for the favicon and preview images.
+   *
+   * aRows, an array of metadata objects that currently generated by embed.ly
+   * Returns a promise that is resolved upon success, or rejected if an exception occurs
+   */
+  asyncInsert: Task.async(function*(aRows) {
```
oyiptong: it's confusing to call this parameter `aRows`. perhaps `metaObjs` makes more sense. `aRows` gives the impression these are database rows, which they aren't.

lib/MetadataStore.js
```diff
@@ -0,0 +1,380 @@
...
+  *
+  * @param {Number} interval
+  *        an time interval in millisecond for this cron job
+  */
+  enableDataExpiryJob(interval) {
+    if (!this._conn) {
+      throw new Error("The database connection is not open yet");
+    }
+
+    if (this._cron) {
```
oyiptong: %s/cron/job/g

oyiptong:
r+ after nits are fixed

k88hudson:
\o/

lib/MetadataStore.js
ncloudioj: I think so. Perhaps we have various url fields in this table, the prefixes help remove the ambiguity.

oyiptong: how about `places_url`?

PR #904, opened 2016-07-13
MetadataParser and PageScraper

k88hudson:
This patch adds metadata parsing and page scraping. There are two parts:
- a `MetadataParser`, which takes a url or a raw html as text
- a `PageScraper` which uses `MetadataParser` to parse pages as they are loaded in the browser. By default it is pref'ed off.

It's not yet hooked up to the metadata db since it kind of depends on implementing update/insert, and I figured it would be better as a subsequent patch

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/904)

### DIFF
--- a/data/page-scraper-content-script.js
+++ b/data/page-scraper-content-script.js
@@ -1,0 +1,14 @@
+function sendData() {
+  const text = document.documentElement.innerHTML;
+  const url = document.documentURI;
+  self.port.emit("message", {type: "PAGE_HTML", data: {text, url}});
+}
+
+sendData();
+if (document.readyState === "complete") {
+  sendData();
+} else {
+  window.addEventListener("load", sendData);
+}
+
+window.addEventListener("pagehide", sendData);
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -24,6 +24,7 @@ const {AppURLHider} = require("lib/AppURLHider");
 const am = require("common/action-manager");
 const {CONTENT_TO_ADDON, ADDON_TO_CONTENT} = require("common/event-constants");
 const {ExperimentProvider} = require("lib/ExperimentProvider");
+const {PageScraper} = require("lib/PageScraper");
 
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource:///modules/NewTabURL.jsm");
@@ -97,6 +98,7 @@ function ActivityStreams(metadataStore, options = {}) {
   );
 
   this._previewProvider = new PreviewProvider(this._tabTracker, metadataStore);
+  this._pageScraper = new PageScraper();
 
   this._populatingCache = {
     places: false,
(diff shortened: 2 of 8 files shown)


data/page-scraper-content-script.js
```diff
@@ -0,0 +1,14 @@
+function sendData() {
+  const text = document.documentElement.innerHTML;
+  const url = document.documentURI;
+  self.port.emit("message", {type: "PAGE_HTML", data: {text, url}});
+}
+
+sendData();
+if (document.readyState === "complete") {
+  sendData();
```
oyiptong: Food for thought: you might want to wait a bit before sending data. That's to account for dynamically generated pages. Something like ~3 secs should be reasonable.

data/page-scraper-content-script.js
```diff
@@ -0,0 +1,14 @@
+function sendData() {
+  const text = document.documentElement.innerHTML;
```
oyiptong: you want outerHTML, right?

lib/ActivityStreams.js
```diff
@@ -97,6 +98,7 @@ function ActivityStreams(metadataStore, options = {}) {
   );
 
   this._previewProvider = new PreviewProvider(this._tabTracker, metadataStore);
+  this._pageScraper = new PageScraper();
```
oyiptong: `PageScraper` should only be instantiated if the pref is on, otherwise it's consuming memory even if it's not being used. If we default to off, we should not instantiate

lib/PageScraper.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  observe(subject, topic, data) {
+    this.emit(topic, data);
+  }
+  load() {
+    if (this.pageMod) {
+      return;
+    }
+    Services.obs.addObserver(this, PageScraper.PAGE_PARSED_NOTIF);
+    this.pageMod = PageMod({
```
oyiptong: instead of using PageMod, you could take a look at the message manager's `loadFrameScript` command.
http://searchfox.org/mozilla-central/source/dom/base/nsIMessageManager.idl#416

As an experiment, it should be OK to use `PageMod` for this, however, as we get more performance-focused, it's worthwhile to look at what an implementation for Firefox would look like.

The PageMod infrastructure uses a set of abstractions (content loaders, workers, rules, etc) that are loaded for each tab created and aren't needed if we're using `include: "*"` anyway... the framescript needs to load regardless.

A framescript + message-manager based implementation should be fairly quick and easy, without the overhead.

oyiptong: Here's an example of how to attach a framescript and to attach a listener: https://github.com/mozilla/loci/blob/032db76fc58d708ecbc503ec2c1926790d017055/lib/DOMFetcher.js

oyiptong: Here's more info about Message Manager's and framescripts: https://developer.mozilla.org/en-US/Firefox/Multiprocess_Firefox/Message_Manager/Message_manager_overview

jaredkerim:
- Change PageMod to a frame script
- Attach metadatastore.js

pdehaan:
I think the new PR for this is #1157? maybe?

Issue #870, opened 2016-06-21
Implement recommendation highlight


edmoz:
This ticket should replace one of the tiles in the Highlights section with a "recommended" tile, if possible.

We should be able to use the new pocket recommendation engine to do this.

*[Edit by @k88hudson for clarity]*

spec:
https://trello.com/c/TQcZZhMe

k88hudson:
We still need to work out:
- [ ] What should be the source of the recommendation? Should it be related to an existing tile that is also shown? Should it *replace* the tile it is related to? Should it be a general "trending" recommendation?
- [ ] What if a recommendation is not available from the designated source? Should we replace it with a general recommendation? Should we just not show it?
- [ ] How frequently should recommendations be shown? All the time when we have one? Only sometimes?

Also note: we should try running this as an experiment.

cc @bryanbell @aaronrbenson

k88hudson:
If you want, I can try something like replacing one of the tiles with its recommendation, and we can see how that goes.

k88hudson:
Note: this is also blocked on implementing pocket endpoints in the proxy

aaronrbenson:
![new tab - recommended content spec](https://cloud.githubusercontent.com/assets/139451/16470763/b7f6fcfc-3e1d-11e6-839e-57353089eddb.png)

k88hudson:
@tspurway I wasnt able to make the meeting on Thursday, did you end up asking about the short URLs?

k88hudson:
Also thanks for the clarification @aaronrbenson! Looks great

sarracini:
Fixed with #906

PR #906, opened 2016-07-14
feat(content): #870 Turn recommendations on as experiment

sarracini:
This sets up recommendations as an experiment and flips the pref on
Last patch for #870 

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/906)


ncloudioj:
The experiment setup looks good to me. Loop @k88hudson in for another check? 

Do you have any tests for this experiment? Like, if the user is enrolled in this experiment, there should be some recommendations showing up. Like we did [here](https://github.com/mozilla/activity-stream/blob/master/content-test/components/LinkMenu.test.js#L178)

sarracini:
@ncloudioj I can definitely add one

ncloudioj:
Yes! R+ :shipit:

PR #907, opened 2016-07-14
feat(addon): Add Recommendation Provider

sarracini:
@oyiptong @jaredkerim r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/907)

### DIFF
--- a/lib/RecommendationProvider.js
+++ b/lib/RecommendationProvider.js
@@ -1,0 +1,125 @@
+/* globals exports, require, Task */
+"use strict";
+const {Cu} = require("chrome");
+const {setTimeout, clearTimeout} = require("sdk/timers");
+const simplePrefs = require("sdk/simple-prefs");
+
+const POCKET_API_URL = "pocket.endpoint";
+
+const DEFAULT_TIMEOUTS = {
+  pocketTimeout: 60000, // every 10 minutes, refresh the Pocket recommendations
+};
+
+Cu.import("resource://gre/modules/Task.jsm");
+Cu.importGlobalProperties(["fetch"]);
+
+function RecommendationProvider(previewProvider, options = {}) {
+  this.options = Object.assign({}, DEFAULT_TIMEOUTS, options);
+  this._recommendedContent = [];
+  this._blockedRecommendedContent = new Set();
+  this._currentRecommendation = null;
+  this._pocketEndpoint = simplePrefs.prefs[POCKET_API_URL];
+  this._previewProvider = previewProvider;
+  this._asyncUpdateRecommendations(this.options.pocketTimeout);
+}
+
+RecommendationProvider.prototype = {
+  _pocketTimeoutID: null,
+
+  /**
+    * If we need a recommendation, go get one
+    */
+  getRecommendation(refresh = false) {
+    // if we need a new recommendation purge the current recommendation and get a new one
+    if (refresh) {
+      this._currentRecommendation = null;
+    }
+    // if we are already displaying a recommendation, just show that one and don't get a new one
+    if (this._currentRecommendation) {
+      return this._currentRecommendation;
+    }
+    this._currentRecommendation = this._getRandomRecommendation();
+    return this._currentRecommendation;
+  },
+
+  /**
+    * Randomly pick a recommendation from the allowed set of recommendations (i.e ones that are not blocked)
+    */
+  _getRandomRecommendation() {
+    const allowedRecommendations = this._recommendedContent.filter(item => !this._blockedRecommendedContent.has(item.url));
+    if (allowedRecommendations.length > 0) {
+      let index = Math.floor(Math.random() * (allowedRecommendations.length));
+      return Object.assign({}, {
+        url: allowedRecommendations[index].url,
+        timestamp: allowedRecommendations[index].timestamp,
+        recommended: true});
+    }
+    return null;
+  },
+
+  /**
+    * Fetches Pocket recommendations and saves them to a global list of recommended content
+    */
+  asyncSetRecommendedContent: Task.async(function*() {
+    try {
+      let response = yield this._asyncGetRecommendationsFromPocket();
+      if (response.ok) {
+        let responseJson = yield response.json();
+        if (responseJson.urls.length > 0) {
+          // wait for the recommendation's metadata to come back (we don't want to show ugly recommendations)
+          this._recommendedContent  = yield this._previewProvider.getLinkMetadata(responseJson.urls);
+        }
+      } else {
+        Cu.reportError(`Response failed with status ${response.status} ${response.statusText}`);
+      }
+    } catch (err) {
+      Cu.reportError(err);
+      throw err;
+    }
+  }),
+
+  /**
+    * Update the list of recommendations the user has blocked (they do not want to see these again)
+    */
+  setBlockedRecommendation(link) {
+    this._blockedRecommendedContent.add(link);
+    this._currentRecommendation = null;
+  },
+
+  _asyncGetRecommendationsFromPocket: Task.async(function*() {
+    try {
+      let response = yield fetch(this._pocketEndpoint, {
+        method: "GET",
+        headers: {"Content-Type": "application/json"}
+      });
+      return response;
+    } catch (err) {
+      Cu.reportError(err);
+      throw err;
+    }
+  }),
+
+  /**
+    * Update the list of recommendations from Pocket
+    */
+  _asyncUpdateRecommendations: Task.async(function*(pocketTimeout = 5000) {
+    if (pocketTimeout) {
+      this._pocketTimeoutID = setTimeout(() => {
+        this.asyncSetRecommendedContent();
+        this._asyncUpdateRecommendations(pocketTimeout);
+      }, pocketTimeout);
+    }
+  }),
+
+  /**
+    * Uninitialize the recommendation provider
+    */
+  uninit() {
+    clearTimeout(this._pocketTimeoutID);
+    this._recommendedContent = new Set();
+    this._blockedRecommendedContent = [];
+    this._currentRecommendation = null;
+  }
+};
+
+exports.RecommendationProvider = RecommendationProvider;
--- a/package.json
+++ b/package.json
@@ -130,6 +130,19 @@
       "type": "string",
       "value": "https://embedly-proxy.services.mozilla.com/v2/extract",
       "hidden": true
+    },
+    {
+      "name": "pocket.endpoint",
+      "title": "Pocket Endpoint",
+      "type": "string",
+      "value": "",
+      "hidden": true
+    },
+    {
+      "name": "recommendations",
+      "title": "Show me recommendations",
+      "type": "bool",
+      "value": true
     }
   ],
   "repository": "mozilla/activity-stream",
--- a/test/test-RecommendationProvider.js
+++ b/test/test-RecommendationProvider.js
@@ -1,0 +1,168 @@
+"use strict";
+
+const {before, after, waitUntil} = require("sdk/test/utils");
+const {RecommendationProvider} = require("lib/RecommendationProvider");
+const simplePrefs = require("sdk/simple-prefs");
+const {Loader} = require("sdk/test/loader");
+const loader = Loader(module);
+const httpd = loader.require("./lib/httpd");
+const gPort = 8079;
+
+let gRecommendationProvider;
+let gPrefPocket = simplePrefs.prefs["pocket.endpoint"];
+let fakeResponse = {"urls": [
+  {url: "http://example.com/1", timestamp: 1467781200000},
+  {url: "http://example.com/2", timestamp: 1467694800000},
+  {url: "http://example.com/3", timestamp: 0},
+  {url: "http://example.com/4", timestamp: 1467383640000},
+  {url: "http://example.com/5", timestamp: 1467262800000},
+]};
+
+exports.test_get_recommended_content = function*(assert) {
+  assert.ok(gRecommendationProvider._pocketEndpoint, "The pocket endpoint is set");
+  let srv = httpd.startServerAsync(gPort);
+
+  srv.registerPathHandler("/pocketRecommendations", function handle(request, response) {
+    response.setHeader("Content-Type", "application/json", false);
+    response.write(JSON.stringify(fakeResponse));
+  });
+
+  // set the recommendations and check that it got some metadata
+  yield gRecommendationProvider.asyncSetRecommendedContent();
+
+  assert.equal(gRecommendationProvider._recommendedContent.length, 5, "set the recommended content");
+  gRecommendationProvider._recommendedContent.forEach(recommendation => {
+    assert.ok(recommendation.metadata, "recommendation got some metadata from PreviewProvider");
+  });
+
+  yield new Promise(resolve => {
+    srv.stop(resolve);
+  });
+};
+
+exports.test_update_recommendations = function*(assert) {
+  // set the recommendations
+  gRecommendationProvider._recommendedContent = fakeResponse.urls;
+  assert.equal(gRecommendationProvider._recommendedContent.length, 5, "currently we have 5 recommendations");
+  assert.ok(gRecommendationProvider._pocketEndpoint, "The pocket endpoint is set");
+  let srv = httpd.startServerAsync(gPort);
+
+  // update the recommendations response
+  let newFakeResponse = {"urls": fakeResponse.urls.slice(0, 4)};
+  srv.registerPathHandler("/pocketRecommendations", function handle(request, response) {
+    response.setHeader("Content-Type", "application/json", false);
+    response.write(JSON.stringify(newFakeResponse));
+  });
+
+  // start the timer to update the data and wait for it to trigger
+  gRecommendationProvider._asyncUpdateRecommendations(100);
+  yield waitUntil(() => {return true;}, 1000);
+
+  // check that the recommendations got updated
+  assert.equal(gRecommendationProvider._recommendedContent.length, 4, "updated the recommended content to 4 recommendations");
+  gRecommendationProvider._recommendedContent.forEach(recommendation => {
+    assert.ok(recommendation.metadata, "recommendation got some metadata from PreviewProvider");
+  });
+  yield new Promise(resolve => {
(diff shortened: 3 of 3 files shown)


lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
+/* globals exports, require, Task */
+"use strict";
+const {Cu} = require("chrome");
+const {setTimeout, clearTimeout} = require("sdk/timers");
+const simplePrefs = require("sdk/simple-prefs");
+
+const MILLISECONDS_IN_TEN_MINUTES = 600000;
+const MILLISECONDS_IN_AN_HOUR = 3600000;
+const MILLISECONDS_IN_A_DAY = 86400000;
+const MILLISECONDS_IN_A_WEEK = 604800000;
```
pdehaan: Not that it helps, but you could possibly just chain these:

```js
const MILLISECONDS_IN_TEN_MINUTES = 600000;
const MILLISECONDS_IN_AN_HOUR = MILLISECONDS_IN_TEN_MINUTES * 6;
const MILLISECONDS_IN_A_DAY = MILLISECONDS_IN_AN_HOUR * 24;
const MILLISECONDS_IN_A_WEEK = MILLISECONDS_IN_A_DAY * 7;
```

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
...
+const POCKET_PREF = "pocket.endpoint";
+
+const DEFAULT_TIMEOUTS = {
+  pocketTimeout: 600000, // every 10 minutes, refresh the Pocket recommendations
```
pdehaan: Nit: Use the `MILLISECONDS_IN_TEN_MINUTES` constant.

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
...
+      let response = yield this._asyncGetRecommendationsFromPocket();
+      if (response.ok) {
+        let responseJson = yield response.json();
+        if (responseJson.urls.length > 0) {
+          // wait for the recommendation's metadata to come back (we don't want to show ugly recommendations)
+          let richRecommendations = yield this._previewProvider.getLinkMetadata(responseJson.urls);
+          this._recommendedContent = richRecommendations;
+          this._recommendedContent.forEach(item => item.timestamp = this._convertTime(item.timestamp));
+        }
+      }
```
pdehaan: Not sure if this matters, but there may be a potential silent failure here if `response.ok` is falsey and no error gets thrown. The `if` would skip, the `catch` wouldn't get caught, and there is no `else` statement.
It may not matter, but hey.

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
...
+    * Convert milliseconds into human readable form and style it accordingly
+    */
+  _convertTime(timestamp) {
+    let publishedTime = Date.now() - timestamp;
+    if (timestamp === 0 || publishedTime < MILLISECONDS_IN_TEN_MINUTES) {
+      return "Now";
+    } else if (publishedTime >= MILLISECONDS_IN_TEN_MINUTES && publishedTime < MILLISECONDS_IN_AN_HOUR) {
+      return Math.floor(((publishedTime / (1000 * 60)) % 60)) + "m";
+    } else if (publishedTime >= MILLISECONDS_IN_AN_HOUR && publishedTime < MILLISECONDS_IN_A_DAY) {
+      return Math.floor(((publishedTime / (1000 * 60 * 60)) % 24)) + "h";
```
pdehaan: Use `MILLISECONDS_IN_AN_HOUR` const here?
1000 * 60 * 60 = 3,600,000 &mdash; see `const MILLISECONDS_IN_AN_HOUR = 3600000;`

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
...
+    } else if (publishedTime >= MILLISECONDS_IN_A_DAY && publishedTime < MILLISECONDS_IN_A_WEEK) {
+      return Math.floor(((publishedTime / (1000 * 60 * 60 * 24)) % 7)) + "d";
```
pdehaan: Use `MILLISECONDS_IN_A_DAY` const here?
1000 * 60 * 60 * 24 = 86,400,000 &mdash; `const MILLISECONDS_IN_A_DAY = 86400000;`

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
...
+    } else {
+      return Math.floor((publishedTime / (1000 * 60 * 60 * 24 * 7))) + "w";
```
pdehaan: Same, `MILLISECONDS_IN_A_WEEK = 604800000;`

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+    * Convert milliseconds into human readable form and style it accordingly
+    */
+  _convertTime(timestamp) {
+    let publishedTime = Date.now() - timestamp;
+    if (timestamp === 0 || publishedTime < MILLISECONDS_IN_TEN_MINUTES) {
+      return "Now";
+    } else if (publishedTime >= MILLISECONDS_IN_TEN_MINUTES && publishedTime < MILLISECONDS_IN_AN_HOUR) {
+      return Math.floor(((publishedTime / (1000 * 60)) % 60)) + "m";
```
pdehaan: I don't see any constants here that would make sense for `1000 * 60`. We'd either need to add a new constant for `MILLISECONDS_IN_A_MINUTE` (because `MILLISECONDS_IN_TEN_MINUTES / 10` would be hella weird).

package.json
```diff
@@ -130,6 +130,19 @@
       "type": "string",
       "value": "https://embedly-proxy.services.mozilla.com/v2/extract",
       "hidden": true
+    },
+    {
+      "name": "pocket.endpoint",
+      "title": "Pocket Endpoint",
+      "type": "string",
+      "value": "http://cloudvm.jaredkerim.com/v2/recommendations",
```
pdehaan: Do we have a more legit looking dev/stage URL for this? Or at least something w/ HTTPS?
Not that I don't trust @jaredkerim, but...

test/test-RecommendationProvider.js
```diff
@@ -0,0 +1,198 @@
+/* globals require, exports */
```
pdehaan: Not sure if we need these `globals`. They _should_ be covered by our ESLint `env` of `node`, I think. Or maybe I'm wrong.

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+    } else if (publishedTime >= MILLISECONDS_IN_A_DAY && publishedTime < MILLISECONDS_IN_A_WEEK) {
+      return Math.floor(((publishedTime / (MILLISECONDS_IN_A_DAY)) % 7)) + "d";
+    } else {
+      return Math.floor((publishedTime / (MILLISECONDS_IN_A_WEEK))) + "w";
+    }
+  },
+
+  _asyncGetRecommendationsFromPocket: Task.async(function*() {
+    try {
+      let response = yield fetch(this._pocketEndpoint, {
```
pdehaan: Just so I understand (I'm a bit slow), but this will always fail now, correct?
Now it seems like the [`pocket.endpoint` config in /package.json](https://github.com/mozilla/activity-stream/pull/907/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R138) is an empty string, so I'm guessing `fetch()` may potentially implode and throw an error which gets caught by the `catch() {}` below which just logs + re-throws the error.
Not sure if this will clear the timers, or if it will just re-attempt and fail every 10 minutes.

sarracini: There's no one using the RecommendationProvider just yet, so it will actually not ever even fire. I just removed the end point under Jared gives us a stage endpoint (the one that was there previously was just for testing purposes and shouldn't have made it to the PR)

pdehaan: Perfect, thanks.

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+  pocketTimeout: MILLISECONDS_IN_TEN_MINUTES, // every 10 minutes, refresh the Pocket recommendations
+};
+
+Cu.import("resource://gre/modules/Task.jsm");
+Cu.importGlobalProperties(["fetch"]);
+
+function RecommendationProvider(previewProvider, options = {}) {
+  this.options = Object.assign({}, DEFAULT_TIMEOUTS, options);
+  this._recommendedContent = [];
+  this._blockedRecommendedContent = [];
```
oyiptong: it will be better to make this a `Set`. Every time you do a `includes`, it does a linear scan.
While this list is probably not going to be big all in all, it is still more correct to use a `Set`

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+  getANewRecommendation(getRecommendation) {
+    this._currentRecommendation = null;
+    this.getARecommendation(getRecommendation);
+  },
+
+  /**
+    * Randomly pick a recommendation from the allowed set of recommendations (i.e ones that are not blocked)
+    */
+  _getRandomRecommendation() {
+    const allowedRecommendations = this._recommendedContent.filter(item => !this._blockedRecommendedContent.includes(item.url));
```
oyiptong: using a `Set` would make this an amortized `O(1 * n)` operation, where n is the number of recommendations available. Using an `Array` makes it a `O(n * m)` operation where `m` is the number of blocked items and `n` is the number of recommendations.

Again, as stated above, the numbers `m` and `n` will not be big, but it's still more correct to use a `Set` here.

lib/RecommendationProvider.js
oyiptong: I agree, I think we should catch the case when the response status is not `200 OK` and log.

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+    */
+  setBlockedRecommendation(link) {
+    this._blockedRecommendedContent.push(link);
+    this._currentRecommendation = null;
+  },
+
+  /**
+    * Convert milliseconds into human readable form and style it accordingly
+    */
+  _convertTime(timestamp) {
```
oyiptong: instead of humanizing the timestamps here, can we do them in-content using `moment` (which we're already using for the highlights cards)?

http://momentjs.com/docs/#/durations/humanize/

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+  this._asyncExpireRecommendations(this.options.pocketTimeout);
+}
+
+RecommendationProvider.prototype = {
+  _pocketTimeoutID: null,
+
+  /**
+    * If we need a recommendation, go get one
+    */
+  getARecommendation(getRecommendation) {
```
oyiptong: when would one pass a `false` value here?

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+      return this._getRandomRecommendation();
+    }
+    // if we don't need a recommendation, don't return one
+    return null;
+  },
+
+  /**
+    * Get a new recommendation from Pocket
+    */
+  getANewRecommendation(getRecommendation) {
```
oyiptong: why not combine `getARecommendation` and `getANewRecommendation` together?

Also, perhaps a better name is just `getRecommendation` which takes a boolean paramenter, `refresh`.
When `refresh` is `true`, the `this._currentRecommendation` cache is purged.

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+  asyncSetRecommendedContent: Task.async(function*() {
+    try {
+      let response = yield this._asyncGetRecommendationsFromPocket();
+      if (response.ok) {
+        let responseJson = yield response.json();
+        if (responseJson.urls.length > 0) {
+          // wait for the recommendation's metadata to come back (we don't want to show ugly recommendations)
+          let richRecommendations = yield this._previewProvider.getLinkMetadata(responseJson.urls);
+          this._recommendedContent = richRecommendations;
+          this._recommendedContent.forEach(item => item.timestamp = this._convertTime(item.timestamp));
```
oyiptong: perhaps doing this in-content with `moment` will work better

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }),
+
+  /**
+    * Update the list of recommendations from Pocket
+    */
+  _asyncExpireRecommendations: Task.async(function*(pocketTimeout) {
+    if (pocketTimeout) {
+      this._pocketTimeoutID = setTimeout(() => {
+        this.asyncSetRecommendedContent();
```
oyiptong: a better name for this function is `_asyncUpdateRecommendations`. Expiring doesn't seem to be the main intent of the function.

Why not have a default value for `pocketTimeout`?

Be aware that the first recommendation fetch will only occur `pocketTimeout` after startup time. You might want to fetch at module startup as well.

lib/RecommendationProvider.js
sarracini: when ActivityStream.js calls this it'll pass in three conditions:
1. are we in the experiment? - If we are not, don't get a recommendation
2. is the recommendations pref on? - If not, don't get a recommendation
3. did the incoming message come from HIGHLIGHTS_REQUEST? - If not we don't need to get a recommendation (ex Top Sites doesn't need a recommendation)

oyiptong: If `ActivityStream.js` has the state available, can it just not call this function?

lib/RecommendationProvider.js
sarracini: ActivityStream.js will fetch the recomendations on start-up.
We do have a default value for pocket timeout: https://github.com/mozilla/activity-stream/pull/907/files#diff-bf979e8f6ec950aa1049414bee6435efR15

oyiptong: I mean, can you have a default value passed in this function, e.g.

```javascript
_asyncExpireRecommendations: Task.async(function*(pocketTimeout = 5000) {
}
```

The reason is that this function seems to require a parameter to function at all and it might be developer friendly for this to work especially since there is no status returned.

lib/RecommendationProvider.js
sarracini: Sure - but I would keep this here because 1) it makes testing a little easier, 2) it abstracts away activity stream needing to know if it needs to get a recommendation or not. That way activitystream doesn't need to worry about that logic and 3) it also gives us flexibility to return different things within the same function in the future if we decide to change the logic on how to get a recommendation

oyiptong: Do you have the code handy for `ActivityStream.js`? I'm having a little trouble visualizing how that will work.

sarracini: https://github.com/mozilla/activity-stream/pull/906/files#diff-cd9b1d133214f36f1c20099d7f6817e6R22 

The PR for the entire thing hooked up (including content) is here https://github.com/mozilla/activity-stream/pull/906/files

oyiptong: The line in question is this, right?

```javascript
let recommendation = this._recommendationProvider.getARecommendation(getRecommendation && simplePrefs.prefs.recommendations && inExperiment);
```

If your concern is added lines of code in AS, why not make this:

```javascript
let shouldGetRecommendation = getRecommendation && simplePrefs.prefs.recommendations && inExperiment;
let recommendation = shouldGetRecommendation ? this._recommendationProvider.getARecommendation() : null;
```

It might be marginally preferable to wrap it in an `if` block for readability.

It seems wasteful to do a function call when there is enough context to know that the function call was not needed. AS contains all the context needed and the boolean evaluation happens there anyway.

Another way to think about this is to think of `RecommendationProvider` as to being as decoupled from the rest of the application as possible. There is no point calling a function if no result was expected.

1. Unless I'm missing something, how does it make testing easier?
2. Point 2 is voided because AS already knows if there needs to be a recommendation or not
3. When the changes happen, we could refactor. The current code doesn't show what could be in the future yet

oyiptong: Also, I assume ActivityStream.js will be in another PR?

sarracini: Ah, I see. Sure I'll make the switch. 

And yep, this PR is just to stick the recommendation provider in there. There'll be an integration PR with AS once this one lands

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,157 @@
...
+const {Cu} = require("chrome");
+const {setTimeout, clearTimeout} = require("sdk/timers");
+const simplePrefs = require("sdk/simple-prefs");
+
+const MILLISECONDS_IN_A_MINUTE = 1000 * 60;
+const MILLISECONDS_IN_TEN_MINUTES = MILLISECONDS_IN_A_MINUTE * 10;
+const MILLISECONDS_IN_AN_HOUR = MILLISECONDS_IN_A_MINUTE * 60;
+const MILLISECONDS_IN_A_DAY = MILLISECONDS_IN_AN_HOUR * 24;
+const MILLISECONDS_IN_A_WEEK = MILLISECONDS_IN_A_DAY * 7;
+const POCKET_PREF = "pocket.endpoint";
```
jaredkerim: Name this something like POCKET_API_URL, pref doesn't tell us anything.

lib/RecommendationProvider.js
sarracini: Sure, I'll just remove this from this PR and make sure to do it in the content side PR

lib/RecommendationProvider.js
```diff
@@ -0,0 +1,126 @@
...
+  getRecommendation(refresh = false) {
+    // if we need a new recommendation purge the current recommendation and get a new one
+    if (refresh) {
+      this._currentRecommendation = null;
+    }
+    // if we are already displaying a recommendation, just show that one and don't get a new one
+    if (this._currentRecommendation) {
+      return this._currentRecommendation;
+    }
+    return this._getRandomRecommendation();
```
jaredkerim: Can we do

this._currentRecommendation = this._getRandomRecommendation();
return this._currentRecommendation;

here, and then _getRandomRecommendation has no side effects and just always return a random selection from available recommendations or null, and then the case of setting it to null is just automatic?

PR #914, opened 2016-07-18
feat(addon): Start/close the Metadata DB

sarracini:
@ncloudioj @jaredkerim @k88hudson r?
this is step 1 of replacing simple storage

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/914)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -16,7 +16,6 @@ const ss = require("sdk/simple-storage");
 const {Memoizer} = require("lib/Memoizer");
 const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
-const {MetadataStore} = require("lib/MetadataStore");
 const {TabTracker} = require("lib/TabTracker");
 const {PreviewProvider} = require("lib/PreviewProvider");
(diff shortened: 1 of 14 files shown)


lib/ActivityStreams.js
```diff
@@ -98,7 +98,8 @@ function ActivityStreams(options = {}) {
   );
 
   this._metadataStore = new MetadataStore();
-  this._previewProvider = new PreviewProvider(this._tabTracker);
+  this._metadataStore.asyncConnect();
```
jaredkerim: @sarracini @ncloudioj should we be doing this here, or should this happen inside the MetadataStore init?

ncloudioj: We can't do the auto-connection in the constructor is mainly due to the fact that the underlying `connect()` in Sqlite.jsm is an async function, the user might have issued some queries upon this connection before it actually gets connected. So we have to explicitly connect to the metadata store, wait until the connection is established, and then proceed queries upon it.

That being said, it's preferred to just create the MetadataStore object here, and call the "asyncConnect" in the PreviewProvider before calling any query APIs.

ncloudioj:
Looks good to me. A solid first step 👍 !

PR #915, opened 2016-07-19
feat(content): Add react components to recommendations

sarracini:
Some content side stuff for the recommendations:
1. Styling for spotlight + tooltip
2. Styling settings 
3. Styling for link context menu
4. Tests

@k88hudson r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/915)

### DIFF
--- a/content-src/components/LinkMenu/LinkMenu.js
+++ b/content-src/components/LinkMenu/LinkMenu.js
@@ -38,9 +38,16 @@ const LinkMenu = React.createClass({
           label: "Dismiss",
           icon: "dismiss",
           userEvent: "BLOCK",
-          onClick: () => dispatch(actions.NotifyBlockURL(site.url))
+          onClick: () => {
+            if (site.recommended) {
+              dispatch(actions.NotifyBlockRecommendation(site.url));
+              dispatch(actions.RequestHighlightsLinks());
+            } else {
+              dispatch(actions.NotifyBlockURL(site.url));
+            }
+          }
         },
-        {
+        !site.recommended && {
           ref: "delete",
           label: "Delete from History",
           icon: "delete",
@@ -102,7 +109,8 @@ LinkMenu.propTypes = {
   allowBlock: React.PropTypes.bool,
   site: React.PropTypes.shape({
     url: React.PropTypes.string.isRequired,
-    bookmarkGuid: React.PropTypes.string
+    bookmarkGuid: React.PropTypes.string,
+    recommended: React.PropTypes.bool
   }).isRequired,
 
   // This is for events
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -17,27 +17,33 @@ const PAGE_NAME = "NEW_TAB";
 const NewTabPage = React.createClass({
   getInitialState() {
     return {
-      showSettingsMenu: false
+      showSettingsMenu: false,
+      renderedOnce: false
     };
   },
-  resetBlockList() {
+  toggleRecommendation() {
     this.props.dispatch(actions.NotifyEvent({
-      event: "UNBLOCK_ALL",
+      event: "TOGGLE_RECOMMENDATION",
       page: PAGE_NAME
     }));
-    this.props.dispatch(actions.NotifyUnblockAll());
+    this.props.dispatch(actions.NotifyToggleRecommendations());
+    this.props.dispatch(actions.RequestHighlightsLinks());
   },
   componentDidMount() {
     document.title = "New Tab";
     setFavicon("newtab-icon.svg");
   },
   componentDidUpdate() {
-    if (this.props.isReady) {
+    if (this.props.isReady && !this.state.renderedOnce) {
       this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
+      this.setState({renderedOnce: true});
     }
   },
   render() {
     const props = this.props;
+    const recommendationLabel = "Show Trending Highlights";
+    const recommendationIcon = props.Spotlight.recommendationShown ? "check" : "   ";
+    const showRecommendationOption = props.showRecommendationOption;
     return (<main className="new-tab">
       <div className="new-tab-wrapper">
         <section>
@@ -70,6 +76,7 @@ const NewTabPage = React.createClass({
             <span className="link-wrapper-right">
               <a
                 ref="settingsLink"
+                hidden={!showRecommendationOption}
                 className={classNames("bottom-link expand", {active: this.state.showSettingsMenu})}
                 onClick={() => this.setState({showSettingsMenu: !this.state.showSettingsMenu})} >
                   <span className="icon icon-spacer icon-settings" /> <span className="text">Settings</span>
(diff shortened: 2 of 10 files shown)


content-src/components/LinkMenu/LinkMenu.js
```diff
@@ -38,9 +38,16 @@ const LinkMenu = React.createClass({
           label: "Dismiss",
           icon: "dismiss",
           userEvent: "BLOCK",
-          onClick: () => dispatch(actions.NotifyBlockURL(site.url))
+          onClick: () => {
+            if (site.recommended) {
+              dispatch(actions.NotifyBlockRecommendation(site.url));
+              dispatch(actions.RequestHighlightsLinks());
```
k88hudson: If this needs to be re-requested, you might run into a race condition situation here. Maybe trigger re-firing highlights from the add-on side, when the pref changes?

k88hudson: In practice this will probably fine since blocking is pretty quick, but these two things are async a race condition is possible

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -17,27 +17,33 @@ const PAGE_NAME = "NEW_TAB";
...
   },
   componentDidMount() {
     document.title = "New Tab";
     setFavicon("newtab-icon.svg");
   },
   componentDidUpdate() {
-    if (this.props.isReady) {
+    if (this.props.isReady && !this.state.renderedOnce) {
       this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
+      this.setState({renderedOnce: true});
```
k88hudson: 👍

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -17,27 +17,33 @@ const PAGE_NAME = "NEW_TAB";
...
     }
   },
   render() {
     const props = this.props;
+    let recommendationLabel = "Show Trending Highlights";
+    let recommendationIcon = this.props.Spotlight.recommendationShown ? "check" : "   ";
```
k88hudson: you can just use `props`

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -17,27 +17,33 @@ const PAGE_NAME = "NEW_TAB";
...
   componentDidUpdate() {
-    if (this.props.isReady) {
+    if (this.props.isReady && !this.state.renderedOnce) {
       this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
+      this.setState({renderedOnce: true});
     }
   },
   render() {
     const props = this.props;
+    let recommendationLabel = "Show Trending Highlights";
```
k88hudson: these should be `const`

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -79,7 +86,7 @@ const NewTabPage = React.createClass({
                 visible={this.state.showSettingsMenu}
                 onUpdate={showSettingsMenu => this.setState({showSettingsMenu})}
                 options={[
-                  {label: "Reset Block List", onClick: this.resetBlockList}
+                  {icon: `${recommendationIcon}`, label: `${recommendationLabel}`, onClick: this.toggleRecommendation}
```
k88hudson: `{icon: recommendationIcon, label: recommendationLabel ...`

content-src/components/LinkMenu/LinkMenu.js
sarracini: is it possible to waitUntil ```NotifyBlockRecommendation``` finishes and then that function can make a call to ```RequestHighlightsLinks``` when it's done?

k88hudson: If you turned it into a `RequestExpect` instead of a `Notify` you could do it, but it will require some refactoring. This is kinda a problem with all the delete/block stuff right now, let's leave this the way it is for now and file a follow-up to address the side effects stuff separately

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -17,27 +17,33 @@ const PAGE_NAME = "NEW_TAB";
...
+    if (this.props.isReady && !this.state.renderedOnce) {
       this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
+      this.setState({renderedOnce: true});
     }
   },
   render() {
     const props = this.props;
+    let recommendationLabel = "Show Trending Highlights";
+    let recommendationIcon = this.props.Spotlight.recommendationShown ? "check" : "   ";
+    let showRecommendationOption = props.showRecommendationOption;
```
k88hudson: Is this actually set anywhere? I don't see it

content-src/components/Spotlight/Spotlight.js
```diff
@@ -22,12 +23,23 @@ const SpotlightItem = React.createClass({
...
+      this.setState({hover: false});
+    }
+  },
   render() {
     const site = this.props;
     const image = site.bestImage;
     const imageUrl = image.url;
     const description = site.description || site.url;
     const isPortrait = image.height > image.width;
+    let timestamp = moment(site.timestamp).fromNow();
```
k88hudson: this should be `const`, also can you name it something more descriptive like `relativeTimestamp`

content-src/components/Spotlight/Spotlight.js
```diff
@@ -98,14 +123,17 @@ const Spotlight = React.createClass({
...
+  onClickFactory(index, site) {
     return () => {
       this.props.dispatch(actions.NotifyEvent({
         event: "CLICK",
         page: this.props.page,
         source: "FEATURED",
         action_position: index
       }));
+      if (site.recommended) {
+        this.props.dispatch(actions.NotifyBlockRecommendation(site.url));
```
k88hudson: Why does clicking on the recommendation block it?

content-src/components/NewTabPage/NewTabPage.js
sarracini: So this will only be set once it gets set up as an experiment. for now, we don't want to show a "settings" button - since there are no settings

sarracini: https://github.com/mozilla/activity-stream/pull/906/files#diff-922cf0fba908e58d1f2dd6a8bb8a2200R124 (for reference on how it'll be set)

k88hudson:
Ok, this is looking good, although I can't really manually test it without the new data structures in redux (I guess that's in a separate patch?)

content-src/components/Spotlight/Spotlight.js
sarracini: the idea is that if you click on the recommendation, you shouldn't be shown that recommendation again, so we just put it on the "block list" of recommendations. the addon picks a new one that isn't blocked.
i guess blocked is kind of a strong word

sarracini:
@k88hudson yeah, i'll have another patch up with the redux stuff. Basically this patch does nothing except lay some of the ground work so the highlight card can show a recommendation. But since everything is pref'ed off, it doesn't actually do anything

k88hudson:
Ok, LGTM R+ We might have to revisit a couple of things after everything has combined

content-src/components/NewTabPage/NewTabPage.js
Mardak: Do we actually want this in `componentDidUpdate` instead of `componentDidMount` or `componentWillMount`? The update method seems like it only gets called /after/ the initial render: https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate

Additionally, I was trying to turn on an eslint rule `react/no-did-update-set-state` which errors on calling `setState` from `componentDidUpdate` as it would trigger a rerendering.

Mardak: Oh actually, I would guess isReady is not true initially, so that's why we need to wait on an update.

Issue #916, opened 2016-07-19
Ensure the drop transaction gets committed in metadata store 


ncloudioj:
We've found an issue of transaction handling in `Sqlite.jsm`, i.e. the transaction promise resolves before it actually gets committed in SQLite. This might end up with a transaction failure if the user closes the database connection inside of the `then()` of the transaction promise.

Before this issue gets fixed in `Sqlite.jsm`, we can work around it by polling the transaction status indicator as well as checking the table status.

ncloudioj:
This has landed in #917

PR #917, opened 2016-07-19
fix(metadata-store): #916 Ensure drop transaction gets committed

ncloudioj:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/917)

### DIFF
--- a/lib/MetadataStore.js
+++ b/lib/MetadataStore.js
@@ -7,6 +7,7 @@ Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Sqlite.jsm");
 Cu.import("resource://gre/modules/NetUtil.jsm");
 Cu.import("resource://gre/modules/Timer.jsm");
+const fileIO = require("sdk/io/file");
 
 XPCOMUtils.defineLazyModuleGetter(this, "Task",
                                   "resource://gre/modules/Task.jsm");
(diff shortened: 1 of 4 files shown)


ncloudioj:
@oyiptong r?

@sarracini and @k88hudson have found the `asycnDrop` was a bit annoying if it couldn't guarantee the drop transaction gets committed. So I made a workaround to make it easier to use.

lib/MetadataStore.js
```diff
@@ -280,6 +281,38 @@ MetadataStore.prototype = {
   }),
 
   /**
+   * This function could be used to ensure the "asyncDrop" actually commit
+   * the transaction. It appears that the transaction might be still in
+   * the uncommitted state despite its promise is resolved in Sqlite.jsm.
+   * Hence, it has to poll for the table info periodically to comfirm that
+   */
+  _waitForDrop() {
```
oyiptong: please rename, it's not waiting for a drop, it's waiting for a reset

oyiptong:
from our conversation IRL: upon uninstall/disable, it is preferable to delete the database rather than reset the database, otherwise we'd leave an empty (but fully created) sqlite database  behind

ncloudioj:
Very good points. Will make those changes accordingly.

ncloudioj:
@sarracini FYI, I've refactored the MetadataStore a bit, specifically,
* rename `asyncDrop` to `asyncReset`, it resets all the tables (essentially, it does a combo - "drop" and then "create tables"
* add `asyncDelete` to nuke the SQLite file, I've also updated the code that handles the "disable/uninstall" events in `main.js`

ncloudioj:
@oyiptong One more round, please? :)

lib/MetadataStore.js
```diff
@@ -280,21 +282,70 @@ MetadataStore.prototype = {
   }),
 
   /**
+   * This function could be used to ensure the "asyncReset" actually commit
+   * the transaction. It appears that the transaction might be still in
+   * the uncommitted state despite its promise is resolved in Sqlite.jsm.
+   * Hence, it has to poll for the table info periodically to comfirm that
+   */
+  _waitForAsyncReset() {
```
oyiptong: makes sense to move this back to the `test/lib/utils.js` now right?

oyiptong:
r+ with a nit: please move the `waitForReset` function back to `test/lib/utils.js`

lib/MetadataStore.js
ncloudioj: Sure, will do.

Issue #918, opened 2016-07-19
Mock a MetadataStore in addon tests


sarracini:
Having to do this: https://github.com/mozilla/activity-stream/pull/914/files#diff-2df7ac17116a705b1d479c76202bd4e3R26 in every test because it makes a real instance of Activity Stream is annoying. We need to mock MetadataStore. As @jaredkerim suggested, we should have some kind of 'test-utils' file which allows you to make a custom instance of Activity Stream which mocks all the things you need it to mock, and the tests can import that file instead of just doing ```let app = new ActivityStreams()```

PR #919, opened 2016-07-19
fix(test): #918 Mock the ActivityStream instance

sarracini:
🎉  @jaredkerim r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/919)

### DIFF
--- a/test/lib/utils.js
+++ b/test/lib/utils.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const {Cc, Ci, Cu, components} = require("chrome");
+const {ActivityStreams} = require("lib/ActivityStreams");
 const {stack: Cs} = components;
 
 function doGetFile(path, allowNonexistent) {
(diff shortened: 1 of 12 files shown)


test/lib/utils.js
```diff
@@ -52,7 +53,17 @@ function doDump(object, trailer) {
   dump(JSON.stringify(object, null, 1) + trailer); // eslint-disable-line no-undef
 }
 
+function mockActivityStream(options = {}) {
```
jaredkerim: We shouldn't call this 'mockActivityStream', because that signals to the caller that we're returning a fake ActivityStream that doesn't implement any real behaviour, which isn't true, it's a real instance of ActivityStream just with a mocked out MetadataStore.  So we should call this getActivityStream or getTestActivityStream to indicate that it's a real ActivityStream, but should only be used for test purposes, not real code.  Other than that this looks great!

jaredkerim:
r+wc

PR #925, opened 2016-07-20
feat(addon): Invoke the Recommendation Provider on addon-side

sarracini:
@oyiptong r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/925)

### DIFF
--- a/lib/ActivityStreams.js
+++ b/lib/ActivityStreams.js
@@ -18,6 +18,7 @@ const {PlacesProvider} = require("lib/PlacesProvider");
 const {SearchProvider} = require("lib/SearchProvider");
 const {TabTracker} = require("lib/TabTracker");
 const {PreviewProvider} = require("lib/PreviewProvider");
+const {RecommendationProvider} = require("lib/RecommendationProvider");
 const {TelemetrySender} = require("lib/TelemetrySender");
 const {PerfMeter} = require("lib/PerfMeter");
 const {AppURLHider} = require("lib/AppURLHider");
@@ -47,6 +48,7 @@ const DEFAULT_OPTIONS = {
   onRemoveWorker: null,
   previewCacheTimeout: 21600000, // every 6 hours, rebuild/repopulate the cache
   placesCacheTimeout: 1800000, // every 30 minutes, rebuild/repopulate the cache
+  recommendationTTL: 3600000, // every hour, get a new recommendation
 };
 
 const PLACES_CHANGES_EVENTS = [
@@ -107,6 +109,15 @@ function ActivityStreams(metadataStore, options = {}) {
 
   this._asyncBuildPreviewCache();
   this._startPeriodicBuildPreviewCache(this.options.previewCacheTimeout);
+
+  // Only create RecommendationProvider if they are in the experiment
+  if (this._experimentProvider.data.recommendedHighlight) {
+    this._recommendationProvider = new RecommendationProvider(this._previewProvider);
+    if (simplePrefs.prefs.recommendations) {
+      this._recommendationProvider.asyncSetRecommendedContent();
+    }
+    this._refreshRecommendations(this.options.recommendationTTL);
+  }
 }
 
 ActivityStreams.prototype = {
@@ -114,6 +125,7 @@ ActivityStreams.prototype = {
   _pagemod: null,
   _button: null,
   _previewCacheTimeoutID: null,
+  _newRecommendationTimeoutID: null,
 
   /**
    * Send a message to a worker
@@ -195,6 +207,9 @@ ActivityStreams.prototype = {
       case am.type("NOTIFY_UNBLOCK_ALL"):
         PlacesProvider.links.unblockAll();
         break;
+      case am.type("NOTIFY_BLOCK_RECOMMENDATION"):
+        this._recommendationProvider.setBlockedRecommendation(msg.data);
+        break;
     }
   },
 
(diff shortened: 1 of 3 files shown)


lib/ActivityStreams.js
```diff
@@ -107,13 +109,20 @@ function ActivityStreams(metadataStore, options = {}) {
 
   this._asyncBuildPreviewCache();
   this._startPeriodicBuildPreviewCache(this.options.previewCacheTimeout);
+
+  this._recommendationProvider = new RecommendationProvider(this._previewProvider);
```
oyiptong: So, we're instantiating the recommendation provider regardless of whether we will use it or not as implemented.
Can we watch a pref and set it instead?

Should the pref be on, `_asyncGetRecommendations` can be called and `_startRecommendationRefresh` can be invoked if it isn't. since it's a timeout, it's easy to see if it's been set or not

lib/ActivityStreams.js
```diff
@@ -443,6 +466,34 @@ ActivityStreams.prototype = {
   },
 
   /**
+    * Start a timer to fetch a new recommendation every hour
+    */
+  _startRecommendationRefresh(recommendationTTL) {
```
oyiptong: if you only call it once, you can say you "start" something. If you call it repeatedly, why not just call it `_refreshRecommendation`?

lib/ActivityStreams.js
```diff
@@ -443,6 +466,34 @@ ActivityStreams.prototype = {
...
+    if (recommendationTTL) {
+      this._newRecommendationTimeoutID = setTimeout(() => {
+        if (simplePrefs.prefs.recommendations) {
```
oyiptong: it should not even run if the pref is not on, see a comment above

lib/ActivityStreams.js
```diff
@@ -443,6 +466,34 @@ ActivityStreams.prototype = {
...
+        }
+        this._startRecommendationRefresh(recommendationTTL);
+      }, recommendationTTL);
+    }
+  },
+
+  /**
+    * Every hour we want to get a new recommendation
+    */
+  _asyncGetNewRecommendation: Task.async(function*() {
```
oyiptong: why not call the function directly?

lib/ActivityStreams.js
```diff
@@ -443,6 +466,34 @@ ActivityStreams.prototype = {
...
+    this._recommendationProvider.getRecommendation(true);
+  }),
+
+  /**
+    * Fetch the recommendations
+    */
+  _asyncGetRecommendations: Task.async(function*() {
```
oyiptong: as above, why the indirection?

lib/ActivityStreams.js
```diff
@@ -574,7 +625,9 @@ ActivityStreams.prototype = {
     let defaultUnload = () => {
       clearTimeout(this._previewCacheTimeoutID);
       clearTimeout(this._placesCacheTimeoutID);
+      clearTimeout(this._newRecommendationTimeoutID);
```
oyiptong: it should only be cleared if set, correct?

lib/ActivityStreams.js
```diff
@@ -574,7 +625,9 @@ ActivityStreams.prototype = {
...
       this._previewProvider.uninit();
+      this._recommendationProvider.uninit();
```
oyiptong: it should only be uninit if _recommendationProvider has been initted

lib/ActivityStreams.js
```diff
@@ -195,16 +204,23 @@ ActivityStreams.prototype = {
...
+        break;
     }
   },
 
   /*
    * Process the passed in links, save them, get from cache and response to content.
    */
   _processAndSendLinks(placesLinks, responseType, worker, options) {
-    let {append, previewsOnly, skipPreviewRequest} = options || {};
+    let {append, previewsOnly, skipPreviewRequest, getRecommendation} = options || {};
```
oyiptong: when is this parameter passed?

oyiptong:
We're very close!

A few points we discussed IRL and one we haven't:

1. not instantiating the RecommendationProvider if not in an experiment, and not doing network requests every hour either
2. removing the coupling with content about `getRecommendation` because all information is available in the addon: it could be renamed `isHighlights` and it would be the same. Also, the parameter should be removed from content.
3. remove unnecessary indirection. one can call the provider methods directly

sarracini:
round two? @oyiptong ?

lib/ActivityStreams.js
```diff
@@ -443,6 +470,19 @@ ActivityStreams.prototype = {
   },
 
   /**
+    * Start a timer to fetch a new recommendation every hour. This will only
+    * run for those in the experiment
+    */
+  _startRecommendationRefresh(recommendationTTL) {
+    if (recommendationTTL && simplePrefs.prefs.recommendations) {
```
sarracini: actually this needs to be:
```js 
 _startRecommendationRefresh(recommendationTTL) {
    if (recommendationTTL) {
      this._newRecommendationTimeoutID = setTimeout(() => {
         if (simplePrefs.prefs.recommendations) {
           this._recommendationProvider.getRecommendation(true);
         }
        this._startRecommendationRefresh(recommendationTTL);
      }, recommendationTTL);
    }
  },
```

lib/ActivityStreams.js
```diff
@@ -443,6 +470,19 @@ ActivityStreams.prototype = {
   },
 
   /**
+    * Start a timer to fetch a new recommendation every hour. This will only
+    * run for those in the experiment
+    */
+  _startRecommendationRefresh(recommendationTTL) {
```
oyiptong: nit: rename to `_refreshRecommendation`?

lib/ActivityStreams.js
oyiptong: 👍

oyiptong: Wait, you need to check if the pref is on, right?

oyiptong: If the pref is off, you don't want to set the timeout again

sarracini: in that case, then if the preff is off at the start, the timeout would never start, and that function wouldn't be called again and you would never refresh recommendations
right?

lib/ActivityStreams.js
```diff
@@ -443,6 +470,21 @@ ActivityStreams.prototype = {
   },
 
   /**
+    * Start a timer to fetch a new recommendation every hour. This will only
+    * run for those in the experiment
+    */
+  _refreshRecommendations(recommendationTTL) {
+    if (recommendationTTL) {
+      this._newRecommendationTimeoutID = setTimeout(() => {
```
oyiptong: That means the timeout will run forever for everyone, regardless of whether the pref is on

sarracini: only if they're in the experiment

oyiptong: hmm i was looking at the RecommendationProvider.

You're also setting a timeout there:

https://github.com/mozilla/activity-stream/blob/790d04b0a4db4ba2c9dce16069f0e697c7dc6f9d/lib/RecommendationProvider.js#L23

This makes a network request

oyiptong:
👍

PR #933, opened 2016-07-21
feat(metrics): Collect metrics on Pocket Recommendations

sarracini:
@emtwo r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/933)

### DIFF
--- a/content-src/components/LinkMenu/LinkMenu.js
+++ b/content-src/components/LinkMenu/LinkMenu.js
@@ -8,17 +8,20 @@ const LinkMenu = React.createClass({
   getDefaultProps() {
     return {
       visible: false,
-      allowBlock: true
+      allowBlock: true,
+      site: {}
     };
   },
   userEvent(event) {
-    const {page, source, index, dispatch} = this.props;
+    const {page, source, index, dispatch, site} = this.props;
     if (page && source) {
       let payload = {
         event,
         page: page,
         source: source,
-        action_position: index
+        action_position: index,
+        url: site.recommended ? site.url : null,
+        recommender_type: site.recommended ? site.recommender_type : null
       };
       dispatch(actions.NotifyEvent(payload));
     }
@@ -116,7 +119,9 @@ LinkMenu.propTypes = {
   // This is for events
   page: React.PropTypes.string,
   source: React.PropTypes.string,
-  index: React.PropTypes.number
+  index: React.PropTypes.number,
+  url: React.PropTypes.string,
+  recommender_type: React.PropTypes.string
 };
 
 module.exports = connect(({Experiments}) => {
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -18,16 +18,19 @@ const NewTabPage = React.createClass({
   getInitialState() {
     return {
       showSettingsMenu: false,
-      renderedOnce: false
+      renderedOnce: false,
+      showRecommendations: true,
     };
   },
   toggleRecommendation() {
     this.props.dispatch(actions.NotifyEvent({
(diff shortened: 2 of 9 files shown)


lib/RecommendationProvider.js
```diff
@@ -13,13 +13,14 @@ const DEFAULT_TIMEOUTS = {
 Cu.import("resource://gre/modules/Task.jsm");
 Cu.importGlobalProperties(["fetch"]);
 
-function RecommendationProvider(previewProvider, options = {}) {
+function RecommendationProvider(previewProvider, tabTracker, options = {}) {
```
sarracini: I was going to land that bit (found here https://github.com/mozilla/activity-stream/pull/906/files#diff-cd9b1d133214f36f1c20099d7f6817e6R115) in a separate PR because no one was instantiating RecommendationProvider yet, but that patch landed so I can add it to this PR

lib/RecommendationProvider.js
```diff
@@ -39,10 +40,28 @@ RecommendationProvider.prototype = {
...
+    let url = null;
+    let recommenderType = null;
+    if (currentRecommendation) {
+      url = currentRecommendation.url;
+      recommenderType = currentRecommendation.recommender_type;
+    }
+    this._tabTracker.handleUserEvent({
+      event: "NEW_RECOMMENDATION",
+      url: url,
+      recommender_type: recommenderType
```
emtwo: nit: you could simplify some of the code like this:
```
url = currentRecommendation ? currentRecommendation.url : null
recommenderType = currentRecommendation ? currentRecommendation.recommender_type : null
```

emtwo:
beauty! r+ with nits addressed

content-src/components/LinkMenu/LinkMenu.js
```diff
@@ -12,13 +12,15 @@ const LinkMenu = React.createClass({
...
+    const {page, source, index, dispatch, site} = this.props;
     if (page && source) {
       let payload = {
         event,
         page: page,
         source: source,
-        action_position: index
+        action_position: index,
+        url: site.recommended ? site.url : null,
+        recommender_type: site.recommended ? site.recommender_type : null
```
pdehaan: Do we need to do any checks that `site` is defined? Or are we guaranteed that we'll have a `site` object if `page` and `source` are truthy?

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -18,16 +18,19 @@ const NewTabPage = React.createClass({
   getInitialState() {
     return {
       showSettingsMenu: false,
-      renderedOnce: false
+      renderedOnce: false,
+      showRecommendations: true,
```
pdehaan: FWIW, trailing comma. ¯\\\_(ツ)\_/¯

content-src/components/LinkMenu/LinkMenu.js
sarracini: hmmm. good question. doesn't hurt to add it in! thanks @pdehaan

sarracini: actually let me double check with kate, i don't want to to add a condition that will never be met and then break all the metrics

pdehaan: Yeah, definitely double check before listening to me (good advice in general).
I did see we specify `propTypes` below which defines a `site` shape object with a required `url` string (lines 112-116 below), so maybe that's enough to guarantee we'll always have that object.

sarracini: kate said we're pretty much guaranteed a site, but to define ```site: {}``` in default props that way we don't have to keep checking if we have it. so i will do that

PR #938, opened 2016-07-21
feat(content): Add redux flow to recommendations

sarracini:
This change:
1. adds the redux part to actually show recommendations. 
2. fixes a typo I had in RecommendationProvider which went unnoticed because it was never being invoked until now

This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/938)

### DIFF
--- a/common/action-manager.js
+++ b/common/action-manager.js
@@ -42,7 +42,9 @@ const am = new ActionManager([
   "NOTIFY_UPDATE_SEARCH_STRING",
   "NOTIFY_BLOCK_RECOMMENDATION",
   "NOTIFY_TOGGLE_RECOMMENDATIONS",
-  "RECEIVE_RECOMMENDATION_TOGGLE"
+  "RECEIVE_RECOMMENDATION_TOGGLE",
+  "PREFS_REQUEST",
+  "PREFS_RESPONSE"
 ]);
 
 // This is a a set of actions that have sites in them,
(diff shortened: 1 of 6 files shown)


content-src/reducers/SetRowsOrError.js
```diff
@@ -5,7 +5,8 @@ const DEFAULTS = {
   error: false,
   init: false,
   isLoading: false,
-  canLoadMore: true
+  canLoadMore: true,
+  recommendationShown: true
```
k88hudson: Is this supposed to be `true` by default?

common/action-manager.js
```diff
@@ -19,7 +19,6 @@ const am = new ActionManager([
   "EXPERIMENTS_RESPONSE",
   "NOTIFY_BLOCK_URL",
   "NOTIFY_UNBLOCK_URL",
-  "NOTIFY_UNBLOCK_ALL",
```
k88hudson: I think the idea was to add this to the dev page, could you maybe file a ticket for that?

k88hudson:
Looks good, but I think we want to keep the unblock stuff so we can add it to the debug page

common/action-manager.js
sarracini: yup, i'll add these back and file a ticket

k88hudson:
LGTM

Issue #929, opened 2016-07-20
Add query APIs to metadata store


ncloudioj:
We need following query APIs:
* given an array of cache_keys, fetching all the pages metadata including the images
...

@sarracini Feel free to add more API requests to the list if you find out more.

ncloudioj:
This has landed in 0955245

PR #943, opened 2016-07-21
feat(metadata-store): #929 Add metadata query to search cache keys

ncloudioj:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/943)

### DIFF
--- a/lib/MetadataStore.js
+++ b/lib/MetadataStore.js
@@ -371,6 +371,61 @@ MetadataStore.prototype = {
   }),
 
   /**
+   * Get page metadata (including images) for the given cache_keys.
+   * For the missing cache_keys, it simply ignores them and will not
+   * raise any exception
+   *
+   * @param {Array} cacheKeys an cache key array
+   *
+   * Returns a promise with the array of retrieved metadata records
+   */
+  asyncGetMetadataByCacheKey: Task.async(function*(cacheKeys) {
+    const quoted = cacheKeys.map(key => {return `'${key}'`;}).join(",");
+
+    let metaObjects = yield this.asyncExecuteQuery(
+      `SELECT * FROM page_metadata WHERE cache_key IN (${quoted})`,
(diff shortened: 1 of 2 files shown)


lib/MetadataStore.js
```diff
@@ -371,6 +371,61 @@ MetadataStore.prototype = {
...
+   *
+   * @param {Array} cacheKeys an cache key array
+   *
+   * Returns a promise with the array of retrieved metadata records
+   */
+  asyncGetMetadataByCacheKey: Task.async(function*(cacheKeys) {
+    const quoted = cacheKeys.map(key => {return `'${key}'`;}).join(",");
+
+    let metaObjects = yield this.asyncExecuteQuery(
+      `SELECT * FROM page_metadata WHERE cache_key IN (${quoted})`,
```
oyiptong: Just curious, did you try comparing the performance of one big query vs many small queries?

test/test-MetadataStore.js
```diff
@@ -158,6 +158,22 @@ exports.test_async_insert_all = function*(assert) {
   assert.deepEqual(items[7], [3, 5]);
 };
 
+exports.test_async_get_by_cache_key = function*(assert) {
```
oyiptong: nit: I'd add a couple more tests, if only for regression testing and ensuring the code path has been run:
1. testing for a key that doesn't exist
2. testing for an image type that doesn't exist

lib/MetadataStore.js
ncloudioj: I haven't done any query benchmark yet. But I did try the one gigantic query, the problem was that you had to sort out the flattened many-to-many result afterward. For instance,

```
metadata1, image1
metadata1, image2
metadata1, favicon1
metadata1, favicon2
...
metadataN, image1
...
metadataN, imageN
```

I think I just chose the simplicity other than performance for now.

test/test-MetadataStore.js
ncloudioj: Good idea, will do.

Issue #928, opened 2016-07-20
Death to simplestorage


sarracini:
Now that we have metadata database all good and going we can replace simple storage to read/write the embedly data to that database.

Along with this goes changing many tests

k88hudson:
🌷 💀💀 💀 💀 💀 💀 💀 🌷

sarracini:
Fixed with #970

PR #970, opened 2016-07-26
feat(addon): #928 Replace simplestorage to read/write from metadata db

sarracini:
GOODBYE SIMPLE STORAGE 💀 

@oyiptong   @k88hudson  @jaredkerim  @ncloudioj who can review?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/970)


oyiptong:
@sarracini can we do a perf analysis on this? we may still need simplestorage for caching

The querying may be done, but we may need to make use of the `memoizer` before we sent the data to content

oyiptong:
you may be able to easily use the `memoizer` by adding a call to `previewProvider` after the `placesProvider` query

https://github.com/mozilla/activity-stream/blob/master/lib/ActivityStreams.js#L390

content-src/selectors/selectors.js
```diff
@@ -129,7 +135,8 @@ module.exports.selectNewTabSites = createSelector(
 const selectSiteIcon = createSelector(
   site => site,
   site => {
-    const favicon = site.favicon_url || site.favicon;
+    const metadataFavicon = site.favicons && site.favicons[0] && site.favicons[0].url;
+    const favicon = site.favicon_url || metadataFavicon || site.favicon;
```
jaredkerim: If this

https://github.com/sarracini/activity-stream/blob/21a364cfb68abb389c3cd78e576d32c9f3cde818/content-src/selectors/selectors.js#L31-L35

and this

https://github.com/sarracini/activity-stream/blob/21a364cfb68abb389c3cd78e576d32c9f3cde818/content-src/selectors/selectors.js#L138-L141

share a bunch of logic, can we pull that into a separate function?

lib/PreviewProvider.js
```diff
@@ -26,12 +25,9 @@ const URL_FILTERS = [
 Cu.importGlobalProperties(["fetch"]);
 Cu.importGlobalProperties(["URL"]);
 Cu.import("resource://gre/modules/Task.jsm");
-Cu.import("resource://gre/modules/Services.jsm");
 
 const DEFAULT_OPTIONS = {
-  cacheCleanupPeriod: 86400000, // a cache clearing job runs at most once every 24 hours
-  cacheRefreshAge: 259200000, // refresh a link every 3 days
-  cacheTTL: 2592000000, // cached items expire if they haven't been accessed in 30 days
+  metadataTTL: 259200000,
```
jaredkerim: Just a personal nit, if you can represent times in ms like this as:

24 * 60 * 60 * 1000

it makes it easier to understand how long it is

lib/PreviewProvider.js
```diff
@@ -211,9 +169,10 @@ PreviewProvider.prototype = {
       this._asyncSaveLinks(processedLinks, event);
     }
 
-    let cachedLinks = this._getEnhancedLinks(processedLinks, previewsOnly, event);
-    return this._getFaviconColors(cachedLinks).then(linksToSend => {
-      return linksToSend;
+    return this._asyncGetEnhancedLinks(processedLinks, previewsOnly, event).then(cachedLinks => {
+      return this._getFaviconColors(cachedLinks).then(linksToSend => {
+        return linksToSend;
```
jaredkerim: Why do you need to do this step?

lib/PreviewProvider.js
```diff
@@ -277,15 +221,27 @@ PreviewProvider.prototype = {
   }),
 
   /**
+   * Find the metadata for each link in the database
+   */
+  _asyncFindItemsInDB: Task.async(function*(links) {
+    let cacheKeyArray = [];
+    links.forEach(link => cacheKeyArray.push(link.cache_key));
```
jaredkerim: I think you can just say

const cacheKeyArray = links.map(link => link.cache_key);

lib/PreviewProvider.js
```diff
@@ -277,15 +221,27 @@ PreviewProvider.prototype = {
...
+    let items = yield this._metadataStore.asyncGetMetadataByCacheKey(cacheKeyArray);
+    return items;
+  }),
+
+  /**
    * Request links from embedly, optionally filtering out known links
    */
-  _asyncSaveLinks: Task.async(function*(processedLinks, event, updateAccessTime = true) {
+  _asyncSaveLinks: Task.async(function*(processedLinks, event) {
+    let DBItems = yield this._asyncFindItemsInDB(processedLinks);
```
jaredkerim: I think 'items' in this case means 'links', so we should just use the word links:

dbLinks

lib/PreviewProvider.js
```diff
@@ -277,15 +221,27 @@ PreviewProvider.prototype = {
...
+    let setItems = new Set();
```
jaredkerim: I think this is storing the links for which we already have data in the db, so we should call it something like

existingLinks

lib/PreviewProvider.js
```diff
@@ -342,18 +298,16 @@ PreviewProvider.prototype = {
...
           }
-          ss.storage.embedlyData[link.cache_key] = Object.assign({}, ss.storage.embedlyData[link.cache_key], data);
-          if (updateAccessTime) {
-            ss.storage.embedlyData[link.cache_key].accessTime = currentTime;
-          }
-          ss.storage.embedlyData[link.cache_key].refreshTime = currentTime;
-        });
+          // Save the entry in the metadata database
+          let entry = Object.assign({}, link, {expired_at: (this.options.metadataTTL) + Date.now()});
+          yield this._metadataStore.asyncInsert([Object.assign({}, data, entry)]);
```
jaredkerim: Do we need to yield here?

sarracini: perhaps @ncloudioj could clarify if we need to yield - i think it has to do with making sure we insert uniques (race conditions if we don't make sure we've inserted that link before requesting to insert that same one)

lib/PreviewProvider.js
sarracini: we compute some colors based on the tippy top site that we we're using to display it nicely before we have embedly's colours. that way we don't have to use the grey fallback color

sarracini: .... i think

lib/PreviewProvider.js
ncloudioj: We don't need to yield here, instead, as @sarracini said, we should catch the exceptions here. Unique key violation is one of the possible exceptions. Other exceptions might also be raised if the SQLite runs into other issues, e.g. out of disk space, oom, transaction timeout etc.

ncloudioj:
@oyiptong It's worth noting that the underlying SQLite library of MetadataStore also has cache feature equipped out of the box (executeCached). So I agree with you that we should do some perf analysis here. I can take this job in a dedicated issue.

jaredkerim:
@oyiptong @ncloudioj I agree.  My preliminary testing shows that it is a few 100ms slower with a fully populated history and metadata cache to do it this way than with ss, so we should definitely add some caching, but as @ncloudioj mentions, it can be done by metadatastore without needing to alter Preview Provider, as a separate issue.

lib/PreviewProvider.js
jaredkerim: No I mean that .then(stuff => stuff) is a no-op, you can just remove that.  I just tried and the tests still pass.

lib/PreviewProvider.js
```diff
@@ -277,15 +219,26 @@ PreviewProvider.prototype = {
   }),
 
   /**
+   * Find the metadata for each link in the database
+   */
+  _asyncFindItemsInDB: Task.async(function*(links) {
+    const cacheKeyArray = links.map(link => link.cache_key);
+    let items = yield this._metadataStore.asyncGetMetadataByCacheKey(cacheKeyArray);
```
jaredkerim: Try to avoid using variable names like 'items' or 'things' or 'puppies'.  Describe what the thing actually is, in this case I think it's a bunch of links with their metadata?  linksMetadata?  Also I think this whole function is:

return yield this._metadataStore.asyncGetMetadataByCacheKey(links.map(link => link.cache_key));

jaredkerim:
r+ from me.  @oyiptong @k88hudson can you do a quick sanity check?

jaredkerim:
@oyiptong @ncloudioj We will handle caching at the metadatstore.js level with this ticket: https://github.com/mozilla/activity-stream/issues/981

lib/PreviewProvider.js
```diff
@@ -342,18 +295,16 @@ PreviewProvider.prototype = {
...
           }
-          ss.storage.embedlyData[link.cache_key] = Object.assign({}, ss.storage.embedlyData[link.cache_key], data);
-          if (updateAccessTime) {
-            ss.storage.embedlyData[link.cache_key].accessTime = currentTime;
-          }
-          ss.storage.embedlyData[link.cache_key].refreshTime = currentTime;
-        });
+          // Save the entry in the metadata database
+          let entry = Object.assign({}, link, {expired_at: (this.options.metadataTTL) + Date.now()});
+          this._metadataStore.asyncInsert([Object.assign({}, data, entry)]);
```
ncloudioj: It's preferred to do a batch insert here to improve the write performance as it has done the unique key deduping in advance. For instance,
```js
linksToInsert = newLinks.filter(link => {// filter out bad links here});
this._metadataStore.asyncInsert(linksToInsert);
```
Under the hood, the `asyncInsert()` takes an array of link objects, and inserts them into the metadata store in *one* transaction.

ncloudioj:
Also noticed two typos in PreviewProvider.js, @sarracini could you correct them as well please?
L69: "Santize", should be "Sanitize"
L321: "Initialize the simple storage"

k88hudson:
Looks good to me!

ncloudioj:
Just realized that there is a missing feature in metadata store PreviewProvider. Unlike simple storage, we didn't update the expiry time for table `page_metadata`.  The current expiring strategy is kinda dumb as it doesn't take the link's frecency and like into account.

I think we can simply use the access time to update the expiry timestamp for now. MetadataStore can expose an API for this.

lib/PreviewProvider.js
```diff
@@ -238,35 +197,18 @@ PreviewProvider.prototype = {
...
-    if (!cachedLink) {
-      return false;
-    }
-    let currentTime = Date.now();
-    return (currentTime - cachedLink.refreshTime) > this.options.cacheRefreshAge;
-  },
+  }),
 
   /**
    * Build the preview cache
```
ncloudioj: This function never gets called by anyone. Perhaps it's for the PreviewProvider cache :)

sarracini: good catch!

sarracini:
adding an integration test for preview provider and metadatastore

test/test-PreviewProvider-MetadataStore.js
```diff
@@ -0,0 +1,143 @@
...
+  srv.registerPathHandler("/previewProviderMetadataStore", function handle(request, response) {
+    response.setHeader("Content-Type", "application/json", false);
+    response.write(JSON.stringify(fakeResponse));
+  });
+
+  // only request the items that are not in the database and wait for them to
+  // successfully be inserted
+  yield gPreviewProvider._asyncSaveLinks(links);
+  yield waitUntil(function*() {
+    if (gMetadataStore.transactionInProgress) {
```
ncloudioj: It should return `true` if the transaction is done, i guess `return !gMetadataStore.transactionInProgress` does that :)

sarracini:
```js
  yield waitUntil(() => {
    if (gMetadataStore.transactionInProgress) {
      return false;
    }
    return true;
  });
```

like that @ncloudioj ?

ncloudioj:
```
yield waitUntil(() => {
  return !gMetadataStore.transactionInProgress;
});
```
I wonder if it's necessary to do the waiting here, did you have any database error here? We did this in other tests, cause the Sqlite library had some buggy transaction support.

Issue #497, opened 2016-04-08
A few outdated npm modules


pdehaan:
Not sure if we want to bump any of these.
It looks like **react@15** was released yesterday https://facebook.github.io/react/blog/2016/04/07/react-v15.html
And **history** had a major version bump.

```sh
$ npm outdated
```

| Package |                      Current |  Wanted |  Latest |  Location | Type |
|----|----:|----:|-----:|----|----|
| [**history**](//npm.im/history)                       | 1.17.0 | 1.17.0  | 2.0.1 | activity-streams | [dep **(major)**](https://github.com/mjackson/history/blob/master/CHANGES.md)
| [**moment**](//npm.im/moment)                       | 2.11.2 | 2.11.2 | 2.12.0 | activity-streams | [dep (minor)](https://github.com/moment/moment/blob/develop/CHANGELOG.md)
| [**react**](//npm.im/react)                         | 0.14.7 | 0.14.7 | 15.0.1 | activity-streams | [dep **(major)**](https://github.com/facebook/react/blob/master/CHANGELOG.md)
| [**react-dom**](//npm.im/react-dom)                     | 0.14.7 | 0.14.7 | 15.0.1 | activity-streams | [dep **(major)**](https://github.com/facebook/react/tree/master/packages/react-dom)
| [**react-redux**](//npm.im/react-redux)                   |  4.1.2 |  4.1.2  | 4.4.1 | activity-streams | [dep (minor)](https://github.com/reactjs/react-redux/releases)
| [**react-router**](//npm.im/react-router)                  |  1.0.3  | 1.0.3  | 2.0.1 | activity-streams | [dep **(major)**](https://github.com/reactjs/react-router/blob/master/CHANGES.md)
| [**redux**](//npm.im/redux)                         |  3.2.0 |  3.2.0  | 3.3.1 | activity-streams | [dep (minor)](https://github.com/reactjs/redux/releases)
| [**redux-logger**](//npm.im/redux-logger)                  |  2.4.0 |  2.4.0 |  2.6.1 | activity-streams | [dep (minor)](https://github.com/fcomb/redux-logger/releases)
| [**redux-thunk**](//npm.im/redux-thunk)                   |  1.0.3 |  1.0.3  | 2.0.1 | activity-streams | [dep **(major)**](https://github.com/gaearon/redux-thunk/releases)
[...]

PR #999, opened 2016-07-29
chore(package): Update npm dev dependenices to latest versions.

Standard8:
This is part of #497. It updates most of the dev dependencies to their latest versions.

There's a couple I'm still investigating (likely need additional code changes), and I will look at those next week. I'll also follow-up with the production dependencies.

The *.scss changes are due to the sass-lint update, which now picks up on the missing double colon:

```
content-src/components/DebugPage/DebugPage.scss
  61:22  warning  Pseudo-elements must start with double colons  pseudo-element
```

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/999)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - "5.5.0"
+  - "5.11.1"
```
Standard8: The version bump for node here, is to get travis to use a newer version of npm as it seems to handle installation better.

rlr:
LGTM and travis is happy 🎉 R+

Let me know if you need me to merge, but I'm assuming you have the powers.

PR #1023, opened 2016-08-02
chore(package): Update npm prod dependencies to latest versions.

Standard8:
This is part of #497. It updates most of the prod dependencies to their latest versions.

The only one that isn't updated to the latest is the `history` dependency, as `react-router` currently only works with 2.x.

There's a couple of minor code changes for compatibility updates for the `history` and `redux-thunk` modules.

I couldn't see any obvious issues resulting from the react and redux module updates.

This will mean we've got most of the modules up to date, and there's just the slightly more complicated updates to work out.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1023)


rlr:
LGTM. I built the xpi and tested it in my everyday browsing profile and it all seems to work fine. R+ 🚀

PR #1036, opened 2016-08-04
fix (addon): #1034 Fix event ping to include experiment data only if necessary

### LINKED ISSUE #1034: Clicks counts on highlights look incorrect
When running the query below we see a drastic drop in clicks on highlights on July 28th, which is when we released the new recommendation.

This is likely a bug in the code for collecting this data. I would expect a pretty small drop if people are clicking on recommendations more but this drop looks too big.

```
SELECT date, COUNT(*)
FROM activity_stream_events_daily
WHERE event IN ('CLICK', 'click')
AND source = 'FEATURED'
AND date >= '07/15/16'
AND (recommendation_url IS NULL
     OR recommendation_url = 'n/a')
GROUP BY date
ORDER BY date
```

ncloudioj:
This fixes #1034

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1036)

### DIFF
--- a/lib/RecommendationProvider.js
+++ b/lib/RecommendationProvider.js
@@ -48,13 +48,14 @@ RecommendationProvider.prototype = {
     * Handle the NEW_RECOMMENDATION event
     */
   _handleNewRecommendationEvent(currentRecommendation) {
-    let url = currentRecommendation ? currentRecommendation.url : null;
-    let recommenderType = currentRecommendation ? currentRecommendation.recommender_type : null;
-    this._tabTracker.handleUserEvent({
-      event: "NEW_RECOMMENDATION",
-      url: url,
-      recommender_type: recommenderType
-    });
+    if (currentRecommendation && currentRecommendation.url &&
+      currentRecommendation.recommender_type) {
+      this._tabTracker.handleUserEvent({
+        event: "NEW_RECOMMENDATION",
+        url: currentRecommendation.url,
+        recommender_type: currentRecommendation.recommender_type
+      });
+    }
   },
 
   /**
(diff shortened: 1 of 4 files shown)


lib/RecommendationProvider.js
```diff
@@ -48,13 +48,13 @@ RecommendationProvider.prototype = {
...
-    this._tabTracker.handleUserEvent({
-      event: "NEW_RECOMMENDATION",
-      url: url,
-      recommender_type: recommenderType
-    });
+    if (currentRecommendation && currentRecommendation.url &&
+      currentRecommendation.recommender_type) {
+      let payload = {event: "NEW_RECOMMENDATION"};
+      payload.url = currentRecommendation.url;
+      payload.recommender_type = currentRecommendation.recommender_type;
```
emtwo: nit: is there a reason `url` and `recommender_type` are not part of the `payload` object when you first generate it? I think it should be safe to do that since you added that if statement right?

emtwo:
r+ aside from minor nit. Thanks Nan!

lib/RecommendationProvider.js
ncloudioj: In the unit test [here](https://github.com/mozilla/activity-stream/blob/master/test/test-RecommendationProvider.js#L88), there is a special case where the `currentRecommendation` is `null`. So I guess if the `currentRecommendation` is null, we shuoldn't send out a ping with both "url" and "recommender_type" as null.

Does it make sense?

ncloudioj: I meant, just wanted to make sure both two fields are valid before send the ping out.

ncloudioj:
Thanks for pointing that out! @emtwo 👍

PR #1039, opened 2016-08-05
React storybook

dmose:
I've removed the storybooks.io integration for the moment, in the interest of getting something landed, and because the current version of React storybook seems to have an odd install issue with it.

Presumably I'll need to squash this before landing, I've left the commits for reviewing, though, as they _might_ be useful in understanding the motivation for given changes.

r? @k88hudson

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1039)

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,9 @@
+const {configure} = require("@kadira/storybook");
+
+function loadStories() {
+  require("../content-test/components/Spotlight.story");
+  require("../content-test/components/ContextMenu.story");
+  // require as many stories as you need.
+}
+
+configure(loadStories, module);
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,0 +1,10 @@
+"use strict";
+const webpack_common = require("../webpack.common");
+
+module.exports = storybookBaseConfig =>
+  Object.assign(storybookBaseConfig, {
+    module: webpack_common.module,
+    resolve: webpack_common.resolve,
+    devtool: "eval-sourcemap",
+    plugins: storybookBaseConfig.plugins.concat(webpack_common.plugins)
+  });
--- a/content-src/components/LinkMenuButton/LinkMenuButton.scss
+++ b/content-src/components/LinkMenuButton/LinkMenuButton.scss
@@ -6,7 +6,7 @@
   width: $link-menu-button-size;
   height: $link-menu-button-size;
   background-color: $white;
-  background-image: url('img/glyph-more-16.svg');
+  background-image: url('#{$image-path}glyph-more-16.svg');
   background-position: 65%;
   background-repeat: no-repeat;
   background-clip: padding-box;
--- a/content-src/styles/icons.scss
+++ b/content-src/styles/icons.scss
@@ -12,86 +12,86 @@
   }
 
   &.icon-activity-stream {
-    background-image: url('img/glyph-activityStream-16.svg');
+    background-image: url('#{$image-path}glyph-activityStream-16.svg');
   }
 
   &.icon-arrow {
-    background-image: url('img/glyph-showmore-16.svg');
+    background-image: url('#{$image-path}glyph-showmore-16.svg');
   }
 
   &.icon-bookmark {
-    background-image: url('img/glyph-bookmark-16.svg');
+    background-image: url('#{$image-path}glyph-bookmark-16.svg');
   }
 
   &.icon-bookmark-white {
-    background-image: url('img/glyph-bookmark-16-white.svg');
+    background-image: url('#{$image-path}glyph-bookmark-16-white.svg');
   }
 
   &.icon-bookmark-remove {
-    background-image: url('img/glyph-bookmark-remove-16.svg');
+    background-image: url('#{$image-path}glyph-bookmark-remove-16.svg');
   }
 
   &.icon-delete {
-    background-image: url('img/glyph-delete-16.svg');
+    background-image: url('#{$image-path}glyph-delete-16.svg');
   }
 
   &.icon-dismiss {
-    background-image: url('img/glyph-dismiss-16.svg');
+    background-image: url('#{$image-path}glyph-dismiss-16.svg');
   }
 
   &.icon-firefox-white {
-    background-image: url('img/glyph-firefox-white-16.svg');
+    background-image: url('#{$image-path}glyph-firefox-white-16.svg');
   }
 
   &.icon-new-window {
-    background-image: url('img/glyph-newWindow-16.svg');
+    background-image: url('#{$image-path}glyph-newWindow-16.svg');
   }
 
   &.icon-new-window-private {
-    background-image: url('img/glyph-newWindow-private-16.svg');
+    background-image: url('#{$image-path}glyph-newWindow-private-16.svg');
   }
 
   &.icon-forward {
-    background-image: url('img/glyph-forward-16.svg');
+    background-image: url('#{$image-path}glyph-forward-16.svg');
   }
 
   &.icon-search {
-    background-image: url('img/glyph-search-16.svg');
+    background-image: url('#{$image-path}glyph-search-16.svg');
   }
 
   &.icon-settings {
-    background-image: url('img/glyph-settings-16.svg');
+    background-image: url('#{$image-path}glyph-settings-16.svg');
   }
 
   &.icon-showMore {
-    background-image: url('img/glyph-showmore-16.svg');
+    background-image: url('#{$image-path}glyph-showmore-16.svg');
   }
 
   &.icon-timeline {
-    background-image: url('img/list-icon.svg');
+    background-image: url('#{$image-path}list-icon.svg');
   }
 
   &.icon-check {
-    background-image: url('img/glyph-check-16.svg');
+    background-image: url('#{$image-path}glyph-check-16.svg');
   }
 
   &.icon-pocket {
-    background-image: url('img/glyph-pocket-16.svg');
+    background-image: url('#{$image-path}glyph-pocket-16.svg');
   }
 
   &.icon-historyItem {
-    background-image: url('img/glyph-historyItem-16.svg');
+    background-image: url('#{$image-path}glyph-historyItem-16.svg');
   }
 
   &.icon-sync {
-    background-image: url('img/glyph-sync-16.svg');
+    background-image: url('#{$image-path}glyph-sync-16.svg');
   }
 
   &.icon-tab {
-    background-image: url('img/glyph-tab-16.svg');
+    background-image: url('#{$image-path}glyph-tab-16.svg');
   }
 
   &.icon-topic {
-    background-image: url('img/glyph-topic-16.svg');
+    background-image: url('#{$image-path}glyph-topic-16.svg');
   }
 }
--- a/content-test/components/ContextMenu.story.js
+++ b/content-test/components/ContextMenu.story.js
@@ -1,0 +1,46 @@
+const React = require("react");
+const {storiesOf} = require("@kadira/storybook");
+const ContextMenu = require("components/ContextMenu/ContextMenu");
+
+const DEFAULT_OPTIONS = [
+  {label: "Apples", onClick: () => {}},
+  {label: "Oranges", onClick: () => {}},
+  {label: "Grapes", onClick: () => {}}
+];
+
+// This is so we can show the context menu.  Note that this is only
+// necessary because ContextMenu has certain implicit dependencies on
+// its container.  Fully encapsulated components (hopefully any new
+// new ones!) should be renderable without needing a container!
+const ContextMenuContainer = React.createClass({
+  getInitialState() {
+    return {visible: true};
+  },
+  render() {
+    const style = {
+      position: "relative",
+      display: "flex",
+      alignItems: "center",
+      justifyContent: "center",
+      width: 150,
+      height: 60,
+      border: "1px dashed #DDD"
+    };
+    const buttonStyle = {
+      fontSize: 11,
+      display: "inline-block"
+    };
+    return (<div style={style}>
+      <ContextMenu
+        options={this.props.options}
+        visible={this.state.visible}
+        onUpdate={visible => this.setState({visible})} />
+      <button style={buttonStyle} onClick={() => this.setState({visible: !this.state.visible})}>Toggle menu</button>
+    </div>);
+  }
+});
+
+storiesOf("ContextMenu", module)
+  .add("default props", () => (
+    <ContextMenuContainer options={DEFAULT_OPTIONS} />
+  ));
--- a/content-test/components/Spotlight.story.js
+++ b/content-test/components/Spotlight.story.js
@@ -1,0 +1,47 @@
+const React = require("react");
+const {SpotlightItem} = require("components/Spotlight/Spotlight");
+const {storiesOf, action} = require("@kadira/storybook");
+const {createMockProvider} = require("../test-utils");
+const {createSite} = require("../faker");
+const Provider = createMockProvider({dispatch: action("dispatched a redux action")});
+
+// XXX should get rid of container here.  See comment in ContextMenu.story
+// for details.
+const Container = props => (
+  <Provider>
+    <div style={{padding: "20px"}}>
+      {props.children}
+    </div>
+  </Provider>
+);
+
+// Note that if a site image is not in the cache, it can take a while
+// (eg 10 seconds) to load, because the image load starts very late, for
+// unclear reasons.  Presumably something to do with faker, tippy-top-sites,
+// lorempixel (used by faker for the images), or Spotlight (aka Highlight)
+// itself.
+storiesOf("Highlight", module)
+  .add("All valid properties", () => {
+    const site = createSite({images: 1});
+    site.bestImage = site.images[0];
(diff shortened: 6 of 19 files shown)


dmose:
"npm run travis" gets much further locally than on travis itself.  It kinda feels like somehow webpack is being run in a context on travis where the JS dialect doesn't support destructuring assignment.  Hmmm...

Will investigate further tomorrow.

.storybook/config.js
```diff
@@ -0,0 +1,11 @@
+const {configure} = require('@kadira/storybook');
+
+require("../data/content/main.css");
+
+function loadStories() {
+  require('../content-test/components/Spotlight.story');
+  require('../content-test/components/ContextMenu.story');
```
pdehaan: Nit: odd mix of single and double quotes in this file, which will surely (hopefully) make ESLint cry.

.storybook/webpack.config.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = {
+  resolve,
+  module: {
+    loaders: [
+      {test: /\.json$/, loader: "json"},
+      {
+        test: /\.css?$/,
+        loaders: ['style', 'raw'],
+        include: path.resolve(__dirname, '../')
```
pdehaan: The fact that ESLint isn't imploding on the unholy union of single and double quotes is proof that I have no idea what I'm doing. My guess is that ESLint is ignoring this folder because it starts with a ".". We may need to explicitly lint this folder as well in our lint task.

content-src/components/LinkMenuButton/LinkMenuButton.scss
```diff
@@ -6,7 +6,7 @@
   width: $link-menu-button-size;
   height: $link-menu-button-size;
   background-color: $white;
-  background-image: url('img/glyph-more-16.svg');
+  background-image: url($image-path + 'glyph-more-16.svg');
```
pdehaan: Nit: Looks like some inconsistent interpolation stuffs between this file and the previous ActivityFeed.scss file:

| FILE | CODE |
|:----|:----|
| ActivityFeed/ActivityFeed.scss | `url('#{$image-path}glyph-bookmark-16-blue.svg')`
| LinkMenuButton/LinkMenuButton.scss | `url($image-path + 'glyph-more-16.svg')`

May be a bit cleaner to stick to one style (either `'#{$foo}bar'` or `$foo + 'bar'`).
Not sure if sass-lint has rules for this. I tried looking at https://github.com/sasstools/sass-lint/tree/master/docs/rules but nothing immediately jumps out.

pdehaan: Unrelated annoying comment, but I was playing with [**sass-inline-svg**](http://npm.im/sass-inline-svg) a couple weeks ago and we could easily inline all the SVGs into the CSS which may reduce loads/lag.

May be as simple as specifying the SVG inliner in an exported function and then specifying the function file using the `--functions` flag when calling node-sass:

```js
// funcs.js
const path = require("path");

const inliner = require('sass-inline-svg');

const SVG_PATH = path.join(__dirname, 'static', 'images');

module.exports = {
  svg: inliner(SVG_PATH)
};
```

```sh
$ node-sass content-src/main.scss -o data/content --functions ./funcs.js
```

... Where we change these `url()` calls to our custom `svg()` function:

```sass
.logo-icon {
  background: svg("glyph-bookmark-16-blue.svg"); // `$image-path` is defined in ./funcs.js
  height: 20px;
  width: 20px;
}
```

webpack.common.js
```diff
@@ -0,0 +1,48 @@
...
+      "components": absolute("./content-src/components"),
+      "reducers": absolute("./content-src/reducers"),
+      "actions": absolute("./content-src/actions"),
+      "selectors": absolute("./content-src/selectors"),
+      "lib": absolute("./content-src/lib"),
+      "strings": absolute("./strings"),
+      "test": absolute("./content-test")
+    }
+  },
+  plugins: plugins
```
pdehaan: If this is ES6, you can probably shorten this to [object shorthand](http://eslint.org/docs/rules/object-shorthand) form `plugins` instead of `plugins: plugins`.

webpack.common.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+const WebpackNotifierPlugin = require("webpack-notifier");
+const webpack = require("webpack");
+const path = require("path");
+const absolute = (relPath) => path.join(__dirname, relPath);
+const EnvLoaderPlugin = require("webpack-env-loader-plugin");
+
+let env = process.env.NODE_ENV || "development";
```
pdehaan: This is now duplicated below in `webpack.config.js:11`. Not sure if we should possibly just `module.exports` this value as well so everything stays in sync.

Can't think of any other clever ideas, but it looks like we're also using `env === "development"`, `env !== 'test'`, and `env === 'production'` in these files. Probably overkill to export `isDevelopmentEnv()`, `isTestEnv()`, and `isProductionEnv()` methods.

.storybook/webpack.config.js
pdehaan: OK, did a bit more poking around after Pizza Friday(tm).

I think we'll need to fix this using a few minor tweaks:

1. Update ESLint to **eslint@latest** (which I believe is expertly handled in #1049)

1. Add `!.*` to the .eslintignore, so we don't automatically ignore the files starting with a dot.

1. Explicitly lint the .storybook/ directory by tweaking yamscripts.yml (and in turn package.json) to something like:

  ```yaml
lint: eslint --ext=.js,.jsx,.json . .storybook && sass-lint -v -q
```

---

After those hoops, I get the following output:

```sh
/Users/pdehaan/dev/tmp/del/activity-storybook/.storybook/config.js
  1:29  error  Strings must use doublequote  quotes
  6:11  error  Strings must use doublequote  quotes
  7:11  error  Strings must use doublequote  quotes

/Users/pdehaan/dev/tmp/del/activity-storybook/.storybook/webpack.config.js
   2:7   error  'webpack' is defined but never used  no-unused-vars
  13:19  error  Strings must use doublequote         quotes
  13:28  error  Strings must use doublequote         quotes
  14:42  error  Strings must use doublequote         quotes
```

dmose: Thanks!  Will update.

dmose: (and fix the quoting issues)

content-src/styles/icons.scss
```diff
@@ -12,86 +12,86 @@
...
   }
 
   &.icon-settings {
-    background-image: url('img/glyph-settings-16.svg');
+    background-image: url($image-path + 'glyph-settings-16.svg');
   }
 
   &.icon-showMore {
-    background-image: url('img/glyph-show-more-16.svg');
+    background-image: url($image-path + 'glyph-show-more-16.svg');
```
pdehaan: Nit: I think this should be "glyph-showmore-16.svg" (not hyphenated "show-more").

content-src/components/LinkMenuButton/LinkMenuButton.scss
dmose: The inconsistent interpolation is only to workaround sasstools/sass-lint#63 .   I only made the switch to `#{$foo}` in cases where sass-lint was actually falling over on the existing syntax. Agreed that it's probably cleaner to change everything, but I didn't want to do that without more input.  @k88hudson what are your thoughts?

dmose: I propose we leave the inlining for another PR, as I'd like to avoid this one growing any more.

webpack.common.js
dmose: Sure; fair enough.

webpack.common.js
dmose: I think we want the requires to stay in-sync with what's in the file.  It's not clear to me that having it in sync with the other file makes any difference.

dmose:
*[content-src/styles/icons.scss, line 67 \[r3\]](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-KON51IU5c0WjJk6Bi8p-r3-67:-KOhHi_WxYUqUJqcrNxQ:b-8zyycv) ([raw file](https://github.com/mozilla/activity-stream/blob/e7fc8e4fc9336b4dc4c84d950ad182a3fb25bd56/content-src/styles/icons.scss#L67)):*
<details><summary><i>Previously, pdehaan (Peter deHaan) wrote…</i></summary>
> Nit: I think this should be "glyph-showmore-16.svg" (not hyphenated "show-more").
</details>
This patch is not changing any filenames.  Let's not entrain more than is necessary in the interest of landing this sooner.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039)*

.storybook/config.js
```diff
@@ -0,0 +1,11 @@
+const {configure} = require('@kadira/storybook');
+
+require("../data/content/main.css");
```
k88hudson: This might actually be better as a static directory (it would reduce the additional configuration needed in webpack) `https://github.com/kadirahq/react-storybook/blob/master/docs/configure_storybook.md#static-directory`

content-src/components/LinkMenuButton/LinkMenuButton.scss
k88hudson: `'#{$foo}bar'` does look a little nicer, but it's fine if we leave this as a follow-up

content-test/components/ContextMenu.story.js
```diff
@@ -0,0 +1,46 @@
...
+const DEFAULT_OPTIONS = [
+  {label: "Apples", onClick: () => {}},
+  {label: "Oranges", onClick: () => {}},
+  {label: "Grapes", onClick: () => {}}
+];
+
+// This is so we can show the context menu.  Note that this is only
+// necessary because ContextMenu has certain implicit dependencies on
+// its container.  Fully encapsulated components (hopefully any new
+// new ones!) should be renderable without needing a container!
```
k88hudson: 👍

package.json
```diff
@@ -179,6 +182,7 @@
     "package": "npm run bundle && jpm xpi && mv activity-streams.xpi dist/activity-streams-$npm_package_version.xpi",
     "travis": "npm run test",
     "prepush": "npm run test:lint && npm run yamscripts",
+    "storybook": "start-storybook -p 9001 -s ./data/content",
```
k88hudson: Ah, so this is already in here! Should work without the require css line then, no?

content-test/components/Spotlight.story.js
```diff
@@ -0,0 +1,48 @@
...
+    <div style={{padding: "20px"}}>
+      {props.children}
+    </div>
+  </Provider>
+);
+
+// Note that if a site image is not in the cache, it can take a while
+// (eg 10 seconds) to load, because the image load starts very late, for
+// unclear reasons.  Presumably something to do with faker, tippy-top-sites,
+// or Spotlight (aka Highlight) itself.
```
k88hudson: Might be lorempixel which is providing the faker images

k88hudson:
Looks great!

Do you think when you're ready to land this you could squash some of your commits? We use the [angular commit style](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines) with the github issue added in if necessary (for example `feat(content): #1 Added a feature`)

I noticed the webpack build is throwing out weird messages and being really slow, I'm thinking it might be because react storybook is trying to process `node_modules` with babel loader instead of just source files:
```
[BABEL] Note: The code generator has deoptimised the styling of "/Users/khudson/github/activity-stream/node_modules/es6-shim/es6-shim.js" as it exceeds the max of "100KB".
[BABEL] Note: The code generator has deoptimised the styling of "/Users/khudson/github/activity-stream/node_modules/moment/moment.js" as it exceeds the max of "100KB".
[BABEL] Note: The code generator has deoptimised the styling of "/Users/khudson/github/activity-stream/node_modules/faker/lib/locales/en/system/mimeTypes.js" as it exceeds the max of "100KB".
```

k88hudson:
Weirdly, their code appears to exclude node_modules: https://github.com/kadirahq/react-storybook/blob/0be9294e485e51fa7c5f7d1cc5570269227d4ffd/src/server/config/paths.js

content-src/styles/icons.scss
pdehaan: Filed as #1057

dmose:
After merging your latest commit, these messages are gone, so I'm betting your guess was correct.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-:-KOlCX-_eKOZ4DS-C3CO:b-raiv3w)*

dmose:
*[package.json, line 185 \[r3\]](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-KON51IU5c0WjJk6Bi8r-r3-185:-KOlGwn7DQpbioXq6aBq:bivbf4u) ([raw file](https://github.com/mozilla/activity-stream/blob/e7fc8e4fc9336b4dc4c84d950ad182a3fb25bd56/package.json#L185)):*
<details><summary><i>Previously, k88hudson (Kate Hudson) wrote…</i></summary>
> Ah, so this is already in here! Should work without the require css line then, no?
</details>
Removing the require("main.css") breaks things (i.e. no CSS is loaded), which makes sense to me.  It has to get the CSS from somewhere, and activity stream itself gets it from a link rel="stylesheet" in the HTML.  We could do that here too, I guess, but it seems like 6 of 1, half-dozen of the other...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039)*

dmose:
*[.storybook/config.js, line 3 \[r3\]](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-KON51IU5c0WjJk6Bi8u-r3-3:-KOlIVnubwt1D23p7jfp:biz656m) ([raw file](https://github.com/mozilla/activity-stream/blob/e7fc8e4fc9336b4dc4c84d950ad182a3fb25bd56/.storybook/config.js#L3)):*
<details><summary><i>Previously, k88hudson (Kate Hudson) wrote…</i></summary>
> This might actually be better as a static directory (it would reduce the additional configuration needed in webpack) `https://github.com/kadirahq/react-storybook/blob/master/docs/configure_storybook.md#static-directory`
</details>
So would you prefer this to be loaded via an HTML tag?  (which itself would require a bit of special config)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039)*

dmose:
*[.storybook/webpack.config.js, line 14 \[r3\]](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-KON51IU5c0WjJk6Bi8v-r3-14:-KOlMlbXEr7eqicSY6FW:b-896fix) ([raw file](https://github.com/mozilla/activity-stream/blob/e7fc8e4fc9336b4dc4c84d950ad182a3fb25bd56/.storybook/webpack.config.js#L14)):*
<details><summary><i>Previously, dmose (Dan Mosedale) wrote…</i></summary>
> (and fix the quoting issues)
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039)*

dmose:
*[content-test/components/Spotlight.story.js, line 23 \[r3\]](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-KON51IU5c0WjJk6Bi8x-r3-23:-KOlNTU7d3CzuvbfhpX0:b-vj294z) ([raw file](https://github.com/mozilla/activity-stream/blob/e7fc8e4fc9336b4dc4c84d950ad182a3fb25bd56/content-test/components/Spotlight.story.js#L23)):*
<details><summary><i>Previously, k88hudson (Kate Hudson) wrote…</i></summary>
> Might be lorempixel which is providing the faker images
</details>
I've updated the comment to include that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039)*

dmose:
*[content-src/styles/icons.scss, line 67 \[r3\]](https://reviewable.io:443/reviews/mozilla/activity-stream/1039#-KON51IU5c0WjJk6Bi8p-r3-67:-KOlOpwQpUf68lO5BuqV:b32w1t) ([raw file](https://github.com/mozilla/activity-stream/blob/e7fc8e4fc9336b4dc4c84d950ad182a3fb25bd56/content-src/styles/icons.scss#L67)):*
<details><summary><i>Previously, pdehaan (Peter deHaan) wrote…</i></summary>
> Filed as #1057
</details>
Ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/activity-stream/1039)*

content-src/components/LinkMenuButton/LinkMenuButton.scss
dmose: I've just gone ahead and fixed this here.

dmose:
OK, here's an updated patch.  Highlights:

* stop using webpack to load CSS, which simplifies webpack config and removes 2 deps
* switch all the url()s in the scss files to the new syntax
* move more stuff into webpack.common
* updated to latest react-storybook available on npm (by the time you read this, there will probably be an even slightly newer on)
* fixed eslint issues.

package.json
dmose: Good catch; fixed!

k88hudson:
@dmose  🔥 🔥  🔥 🔥 🔥 Land it!

PR #1040, opened 2016-08-05
chore(lint): Enable various eslint rules that only need a few fixes

Mardak:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1040)

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -36,8 +36,8 @@ const NewTabPage = React.createClass({
     document.title = "New Tab";
     setFavicon("newtab-icon.svg");
   },
-  componentDidUpdate() {
-    if (this.props.isReady && !this.state.renderedOnce) {
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.isReady && !this.state.renderedOnce) {
       this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
       this.setState({renderedOnce: true});
     }
--- a/content-src/components/Search/Search.js
+++ b/content-src/components/Search/Search.js
@@ -45,7 +45,7 @@ const Search = React.createClass({
     const numSuggestions = this.props.formHistory.concat(this.props.suggestions).length;
     const numEngines = this.props.engines.length;
     if (index < numSuggestions) {
-      return "search-suggestions-" + "history-search-suggestions-" + index;
+      return `${index < this.props.formHistory.length ? "history-" : ""}search-suggestions-${index}`;
     } else if (index < numSuggestions + numEngines) {
       return "search-partners-" + (index - numSuggestions);
     } else if (index === numSuggestions + numEngines) {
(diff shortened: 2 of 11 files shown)


content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -36,8 +36,8 @@ const NewTabPage = React.createClass({
     document.title = "New Tab";
     setFavicon("newtab-icon.svg");
   },
-  componentDidUpdate() {
-    if (this.props.isReady && !this.state.renderedOnce) {
+  componentWillReceiveProps(nextProps) {
```
sarracini: just out of curiosity, what difference does this make here? @Mardak

sarracini:
r+ for ```NewTabPage.js``` & ```Search.js```

content-src/components/NewTabPage/NewTabPage.js
Mardak: The react eslint rule has some information https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-update-set-state.md and the descriptions of nearby lifecycle methods provides some context https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate

Main thing is that calling `setState` from `componentDidUpdate` will cause `render` to run again as something in `props` or `state` changed. Calling `setState` from `componentWillReceiveProps` will effectively coalesce the two `render` calls into one as it will appear as if the `props` and `state` changed at the same time.

content-src/components/Search/Search.js
```diff
@@ -45,7 +45,7 @@ const Search = React.createClass({
     const numSuggestions = this.props.formHistory.concat(this.props.suggestions).length;
     const numEngines = this.props.engines.length;
     if (index < numSuggestions) {
-      return "search-suggestions-" + "history-search-suggestions-" + index;
+      return `${index < this.props.formHistory.length ? "history-" : ""}search-suggestions-${index}`;
```
pdehaan: This seems odd to me, and a big change in logic.

The old version looks like we were returning:
- `search-suggestions-history-search-suggestions-${index}`

... but the new version seems like we're returning one of the following:

- `history-search-suggestions-${index}`
- `search-suggestions-${index}`

lib/PreviewProvider.js
```diff
@@ -82,7 +82,7 @@ PreviewProvider.prototype = {
       // filter out the allowed query params and update the query string
       newURL.search = Array.from(queryParams.keys())
         .filter(param => ALLOWED_QUERY_PARAMS.has(param))
-        .map(param => param = param + "=" + queryParams.get(param)).join("&");
+        .map(param => param + "=" + queryParams.get(param)).join("&");
```
pdehaan: Seems like another logic change, but the old behavior looks decidedly odd with the extra `param +`.

Nit: Not sure if you want to move that `.join("&")` onto the following line so it aligns with `.filter()` and `.map()`.

Also not sure if we need to URL-escape the `queryParams.get(param)` value, or if that is done elsewhere.

content-src/components/Search/Search.js
Mardak: Yes. `eslint` caught a coding error that is fixed here. :)

lib/PreviewProvider.js
Mardak: Yup. Another coding error although not incorrect but dangerous. Reassigning the parameter does nothing.

Mardak: Moved `join` to newline.

Issue #1010, opened 2016-07-29
Add a sorting mechanism for highlights based on weights/score


piatra:
As a user of AS I want more than just 3 recommendations, the highlights section should expand to replace recent activity.

Deliver an initial option to score highlights and allow for more than 3 results.
Recommendations that do not fit in the top 3 positions will fall down into recent activity.
This will also help drive https://github.com/mozilla/activity-stream/issues/1006 because it will make the requirements for that module more obvious.

oyiptong:
might need to be broken up in separate issues to fit the milestones

PR #1041, opened 2016-08-05
Add scoring function based on weights #1010

piatra:
There are some failing JPM tests which I want to investigate before I request review. Same tests are failing for me on activity-stream/master as well.

*EDIT* but they are passing on Travis. Interesting.

I chose not to split the patch in two because there would be no way to actually see how the feature works. It adds a minimal setup required to score, sort highlight results and display them. I would like to also surface a UI component that allows you to change the [scoring parameters](https://github.com/piatra/activity-stream/blob/e44403d34d19801626b2c0ff655ef1a71cd3e5b8/common/recommender/Baseline.js#L33) and the currently selecting scoring mechanism with future pull requests. (and the pull request got pretty big as is). Right now it should be thought of as being equivalent of the SQL query we have but with scores (meaning nothing really fancy is happening in the background, but it is more extensible so new types can easily be added in with appropriate `priorities` -- scoring some higher than others).

This pull req adds a `/recommender/` folder in `/common/` which contains both an entry point for initializing a recommender system and (in the future) all the available classifiers or scoring functions.

It adds a new pref `weightedHighlights` set as `false` in package.json and `true` in `dev-prefs.json` so the feature should be available when launching dev-AS.

It uses the existing `getHighlightsLinks` from the places provider as set of sites to score and recommend. I would like this to be different, it's something I will change with future pull requests.

It uses the existing `ActivityFeed` component which doesn't really make sense (scored items are not chronologically ordered) but it was less code than implementing a different component. I will try with a future pull req to either refactor `ActivityFeed` and have it render differently based on `prop.weightedHighlights` or add a new component for this. At the moment the `ActivityFeed` rendering logic is pretty tightly coupled.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1041)


piatra:
Please review @k88hudson, r? @emtwo for `common/recommender/*` or anyone else who can get to it 😄

piatra:
@emtwo reported some issues with the patch. I am looking into this right now, I would delay any review until it is fixed.

piatra:
@emtwo @k88hudson please review.
I initially delayed this review due to some issues, I wanted to give some more insight into that and the issues I encountered.

1. sharing an xpi file resulted in a broken install of activity stream. This was because I was doing `jpm xpi` I was not aware of the packaging task so the result was obviously broken

2. the development server has to always be on when testing (`npm run start` command) otherwise the addon tries to communicate with an endpoint that does not exist. The issue here is that the fail is silent no obvious error messages.

3. (not 100% certain about this) doing `npm install` is required before `npm run start` and trying to run the addon.

It was a combination of the first two steps that made it difficult to debug why sharing the xpi of activity-stream with my patch applied did not work. Seems obvious now but now when I was trying to debug this in completely the wrong places.
I will make a pull request for the README to mention the npm packaging task, I think that would be useful to have in there

oyiptong:
cc @ncloudioj

common/recommender/Baseline.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Scoring function for an array of URLs.
+   *
+   * @param {Array.<URLs>} entries
+   * @returns {Array.<URLs>} sorted and with the associated score value.
+   */
+  score(entries) {
+    return entries.map(e => this.scoreEntry(e))
```
k88hudson: `e` is kind of a weird choice here IMO, since it usually means event

common/recommender/Baseline.js
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @param {Array.<URLs>} history - User history used to assign higher score to popular domains.
+ */
+class Baseline {
+  constructor(history) {
+    this.domainCounts = history.reduce(this.countDomainOccurrences, new Map());
+  }
+
+  scoreEntry(entry) {
+    let urlObj = URL(entry.url);
```
k88hudson: this can all (mostly?) be `const`

common/recommender/Baseline.js
```diff
@@ -0,0 +1,106 @@
...
+    }, 0);
+
+    return value * Math.pow(Math.E, -exp);
+  }
+
+  sortDescByScore(a, b) {
+    return b.score - a.score;
+  }
+
+  dedupHosts(e, idx, arr) {
```
k88hudson: typo here (`dedupe`

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -42,11 +43,31 @@ const NewTabPage = React.createClass({
       this.setState({renderedOnce: true});
     }
   },
+  _hasWeightedHighligts() {
+    return this.props.WeightedHighlights.weightedHighlights && this.props.WeightedHighlights.rows.length;
```
k88hudson: if `WeightedHighlights` is a reducer, you are enforcing that it always has `rows`, so this check shouldn't be necessary

content-src/selectors/selectors.js
```diff
@@ -31,6 +31,21 @@ function getBackgroundRGB(site) {
   return favicon ? DEFAULT_FAVICON_BG_COLOR : getRandomColor(label);
 }
 
+function assignImageAndBackgroundColor(rows) {
```
k88hudson: Would you mind actually putting this in its own file, and maybe adding a comment? I know most of the selectors are in the same file right now but it's getting kind of big, and it would be nice if we started refactoring things out

content-src/components/NewTabPage/NewTabPage.js
piatra: But that check says something else `if pref is on && we have results`.

piatra:
@k88hudson updated.

package.json
```diff
@@ -142,6 +142,12 @@
       "title": "Show me recommendations",
       "type": "bool",
       "value": true
+    },
+    {
+      "name": "weightedHighlights",
+      "title": "New recommendations method based on weights",
+      "type": "bool",
+      "value": false
```
k88hudson: Let's make this a hidden pref for now

lib/PlacesProvider.js
```diff
@@ -719,6 +721,22 @@ Links.prototype = {
   }),
 
   /**
+   * Get all history items
```
k88hudson: Can you add a comment about what this is for? i.e. to determine frequency of visits

content-test/recommender/Baseline.test.js
```diff
@@ -0,0 +1,129 @@
+/* globals describe, it, beforeEach */
+
+const {Baseline} = require("common/recommender/Baseline");
```
k88hudson: awesome!!

dev-prefs.json
```diff
@@ -3,5 +3,6 @@
   "xpinstall.signatures.required": false,
   "extensions.@activity-streams.sdk.console.logLevel": "info",
   "extensions.@activity-streams.telemetry.ping.endpoint": "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream",
-  "extensions.@activity-streams.embedly.endpoint": "https://embedly-proxy.services.mozilla.com/v2/extract"
+  "extensions.@activity-streams.embedly.endpoint": "https://embedly-proxy.services.mozilla.com/v2/extract",
+  "extensions.@activity-streams.weightedHighlights": true
```
k88hudson: turn this off for now

content-src/components/NewTabPage/NewTabPage.js
k88hudson: oops, you're right!

Let's actually move this to the NewTabPage selector, as we discussed

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -42,11 +43,31 @@ const NewTabPage = React.createClass({
...
+  },
+  _renderGroupedSectionTitle() {
+    let title = "Recent Activity";
+
+    if (this._hasWeightedHighligts()) {
```
k88hudson: if you still need to do this, return a `hasWeightedHighlights` prop from the selector

content-src/selectors/selectors.js
```diff
@@ -123,6 +80,7 @@ module.exports.selectNewTabSites = createSelector(
       History.rows])[2];
 
     return {
+      WeightedHighlights: Object.assign({}, WeightedHighlights, {rows: weightedHighlightsRows}),
```
k88hudson: can you add `WeightedHighlights.init` to the `isReady` line (87)?

k88hudson:
This is looking great! 🔥🔥🔥

k88hudson:
I noticed you're not applying sorting logic for stuff like images, titles etc. That's fine, but we should make a note in a future PR to take those into account

piatra:
@k88hudson https://github.com/piatra/activity-stream/blob/f2bf8c67ddb7c545cfe89ee038dcee425e29644c/common/recommender/Baseline.js#L30 ?

piatra:
@k88hudson updated the pull req. Also when moving stuff to selectors I realized it would be easier to revert back to the old "Recent Activity" view because it would mean I don't need extra props to be passed down to components, so I did that. So now only top highlights get changed.

common/recommender/Baseline.js
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Scoring function for an array of URLs.
+   *
+   * @param {Array.<URLs>} entries
+   * @returns {Array.<URLs>} sorted and with the associated score value.
+   */
+  score(entries) {
+    return entries.map(entry => this.scoreEntry(entry))
+      .filter(e => e.score > 0)
```
k88hudson: The only problem with filtering our 0 scores is that sometimes, we end up with only 2 items (I noticed this happening on a kind of old profile) – maybe let's not remove stuff, just sort it to the bottom (as you're already doing), until we have a better strategy to deal with not enough items

k88hudson:
Awesome! Just one small nit, but pretty much ready to go!

Before you land this, please squash your commits to match the [angular commit style](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit) we use, e.g.:

```
 feat(highlights): #1010 Add scoring function based on weights
```

common/recommender/Baseline.js
piatra: 👍

Issue #1033, opened 2016-08-04
Turn on multiprocess compatibility for Activity Stream


Standard8:
With Firefox turning on multiprocess compatibility for users, and that will soon be turned on for compatible add-ons, we should move towards making activity stream be declared as compatible.

Although it works at the moment, Firefox is implementing some shims for us to keep it working. The [documentation page](https://developer.mozilla.org/Add-ons/SDK/Guides/Multiprocess_Firefox_and_the_SDK#Testing_with_shims_enabled) also states that using the shims may lead to performance issues.

dmose:
Apart from the shims, presumably things not marked as multiprocess compatible are going to take different code paths and therefore have other different performance characteristics as well.  So any profiling that's folks are currently doing while in Firefox is e10s mode may not ultimately be as useful as one would hope, since AS isn't really running as an e10s add-on yet.

Standard8:
I have a patch in progress for this. It looks like most of the incompatibilities are that `Services.obs.addObserver` has been called with two arguments instead of three.

sarracini:
Fixed with #1042

PR #1042, opened 2016-08-05
fix(addon): #1033 Turn on the Multiprocess compatibility flag, correct the uses of addObserver, and avoid the use of shims.

Standard8:
This seems to get us enough to turn on the multiprocess compatibility flag and to avoid the use of the shims (which may impact performance in e10s mode).

Most of the work here is to correct the uses of `Services.obs.addObserver` this is meant to take [three arguments](https://dxr.mozilla.org/mozilla-central/rev/0ba72e8027cfcbcbf3426770ac264a7ade2af090/xpcom/ds/nsIObserverService.idl#25), but for some reason with the shims / non-e10s we were getting away with two.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1042)

### DIFF
--- a/lib/TabTracker.js
+++ b/lib/TabTracker.js
@@ -1,7 +1,7 @@
-/* globals Services, Locale */
+/* globals Services, Locale, XPCOMUtils */
 
 const tabs = require("sdk/tabs");
-const {Cu} = require("chrome");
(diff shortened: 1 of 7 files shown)


sarracini:
R+ this looks good to me

PR #1049, opened 2016-08-05
chore(package): Update eslint to latest version; fix reported issues with defining generators when yield isn't used and also fix indentation in a couple of places.

Standard8:
This upgrades eslint to the latest version.

Post upgrade eslint was complaining about a) indentation (easy fix), b) generators without yield.

I've removed the generators and changed a few names of functions to update the fact they are no longer async.

In `ActivityStreams.js` I ended up moving the logging for the performance data for requests. This was made to log the data before the functions that process the incoming requests. This means the appropriate data is always already logged in a synchronous response case.

Without that, some of the tests were failing due to the `item.delta` being undefined, as it couldn't find the request to generate it from.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1049)

Note: part of the work for #497


pdehaan:
The code superficially looks fine to me (and passes linting and Travis tests). But I'm not an expert on removing generator functions or `Task.async()`, so it will definitely need some eyes from @k88hudson @rlr @oyiptong, etc.

lib/SearchProvider.js
```diff
@@ -134,7 +132,7 @@ NewTabSearchProvider.prototype = {
   /**
    *  Gets the state - a combination of the current engine and all the visible engines.
    */
-  asyncGetCurrentState: Task.async(function*() {
+  get currentState() {
```
Mardak: This is the change in question. If we do want to keep this async, we can have eslint ignore this and undo the other changes related to `asyncGetCurrentState` -> `currentState`.

Mardak: Or actually, we could just keep the Task.async and have it be `function()` instead of `function*()`

Mardak:
I've pushed a new PR #1055 to land this for Standard8 as well as removed the API change to `asyncGetCurrentState` if that's what's desired.

k88hudson:
Closing as this is included in #1055

PR #1055, opened 2016-08-09
chore(package): Update eslint to latest version; fix reported issues with defining generators when yield isn't used and also fix indentation in a couple of places.

Mardak:
Clone of #1049 but removed `asyncGetCurrentState` API change. Closes #1049.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1055)


k88hudson:
Awesome! Could I possibly get you to rewrite your commit to fit the angular commit style (https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit)? (i.e. `chore(lint)....`

lib/main.js
```diff
@@ -69,11 +69,11 @@ Object.assign(exports, {
...
-        // increment the connect counter to avoid the endless retry
-        connectRetried++;
-        this.reconnectMetadataStore();
-      });
+        .catch(error => {
+          // increment the connect counter to avoid the endless retry
+          connectRetried++;
+          this.reconnectMetadataStore();
+        });
     }.bind(this), 500);
```
pdehaan: If we switch to arrow functions, do we need still need`.bind(this)`?

test/test-PreviewProvider.js
```diff
@@ -68,7 +68,7 @@ exports.test_only_request_links_once = function*(assert) {
   });
 };
 
-exports.test_filter_urls = function*(assert) {
+exports.test_filter_urls = function(assert) {
```
pdehaan: Unrelated, but now that my eyes are seeing it...
In the file above this, we use something like:
```js
exports["test blocklist basic format"] = function (assert) {
```

But in this file we use:
```js
exports.test_filter_urls = function (assert) {
```

lib/main.js
Mardak: Nope. Probably should do this as part of turning on http://eslint.org/docs/rules/prefer-arrow-callback than individually.

Issue #892, opened 2016-06-28
Implement Share Toolbar button


edmoz:
feature card: https://trello.com/c/uRoUOISr

pdehaan:
This feels weird to me.
Why is this an Activity-Stream thing and not a separate Test Pilot experiment add-on that we validate separately? If I opt out of Activity Stream, I can no longer share anything?

And as a user, installing the Activity Stream add-on and now having TWO new toolbar icons appearing would be pretty confusing.

rlr:
Quick update: the toolbar is implemented (minus "send to device") in https://github.com/mixedpuppy/sharemenu. Now working on porting the code over to Activity Stream. Then there is some polish and some messing with preferences so our users have a default set of social sites enabled to share to out of the box.

rlr:
@edmoz @nchapman,

I'm going to enable this default set of providers for users that haven't ever enabled any before. Are these good? Should any others be added? The full list of providers is here: https://activations.cdn.mozilla.net/en-US/

<img width="249" alt="screen shot 2016-08-10 at 12 30 51 pm" src="https://cloud.githubusercontent.com/assets/36629/17561962/7da2fe4e-5ef6-11e6-8454-21296dd0f9fc.png">

edmoz:
Not sure why yahoo mail would be in the list.  The others are social feeds, but mail seems the same as 'email link'. Otherwise it looks like a good list to me, it matches what i see on my Android Fx Share button.

edmoz:
The mocks have yahoo mail - so we should include it.  Rlr also mentioned the order of display is based on frencency (though i'm on a new profile).  

'Email Link...' launched the default mail and if it's gmail in chrome, it focuses the browser but doesn't launch a new window with gmail in it.

nchapman:
> Are these good? Should any others be added?

I'd like to see the list be the following:

- Facebook
- Twitter
- Tumblr
- LinkedIn
- Yahoo Mail
- Gmail

This is based somewhat on the numbers that Shane shared previously:

> For activations of share providers [in order], facebook 50%, yahoo email 20%, G+ 9%, GMail 5%, odnoklassniki + vkontakte come up to 8%, twitter 3%.  all others <1%

But I've also grouped them by type. Can we make sure that we instrument which ones are being clicked on?

bryanbell:
Icons needed 

[_final Activity Stream icons.zip](https://github.com/mozilla/activity-stream/files/412329/_final.Activity.Stream.icons.zip)

PR #1070, opened 2016-08-11
feat (addon): #892 implement share toolbar

rlr:
The ShareProvider is based heavily on https://github.com/mixedpuppy/sharemenu. Refactored for our needs and linted for our rules.

@k88hudson can you do a general r?

@ncloudioj @emtwo I need feedback for the payload of the user events I am sending. I'll tag you below.

@mixedpuppy if you have time maybe you can do a sanity check?

It should look like this:
screen shot 2016-08-11 at 11 41 57 am

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1070)


lib/ShareProvider.js
```diff
@@ -0,0 +1,448 @@
...
+    writable: true,
+    value: {
+      copyLink: url => {
+        let clipboard = Cc["@mozilla.org/widget/clipboardhelper;1"].getService(Ci.nsIClipboardHelper);
+        clipboard.copyString(url);
+        eventTracker.handleUserEvent({
+          event: "SHARE_FROM_TOOLBAR",
+          url,
+          provider: "copy-link"
+        });
```
rlr: @ncloudioj @emtwo ^^ here is the event payload. Plus what gets added automatically by `TabTracker.handleUserEvent()`

lib/ShareProvider.js
```diff
@@ -0,0 +1,448 @@
...
+      },
+      emailLink: (browser, url) => {
+        window.MailIntegration.sendLinkForBrowser(browser);
+        eventTracker.handleUserEvent({
+          event: "SHARE_FROM_TOOLBAR",
+          url,
+          provider: "email-link"
```
rlr: @ncloudioj @emtwo ^^ here is another

lib/ShareProvider.js
```diff
@@ -0,0 +1,448 @@
...
+        }
+
+        let shareEndpoint = OpenGraphBuilder.generateEndpointURL(provider.shareURL, pageData);
+        window.open(shareEndpoint, "share-dialog", "chrome");
+
+        eventTracker.handleUserEvent({
+          event: "SHARE_FROM_TOOLBAR",
+          url: pageData.url,
+          provider: provider.origin
+        });
```
rlr: @ncloudioj @emtwo ^^ and another one

rlr: Should we send the URL the user is sharing back? I wasn't sure if that was cool or not privacy-wise.

rlr: OK, talked to @ncloudioj... He said NO URLS PLZ...

Also, we discussed possibly dropping provider and including it in the event name instead. Like for example:
"SHARE_FROM_TOOLBAR:https://facebook.com", "SHARE_FROM_TOOLBAR:email", SHARE_FROM_TOOLBAR:https://twitter.com", etc. @emtwo what do you think?

ncloudioj: Given that we want to track the popularity for the different share providers, and the providers might change over time. I'd recommend to have a separate "provider" in the RS events table. It won't need too much extra work in the data pipeline :)

Thoughts? @emtwo

lib/ShareProvider.js
ncloudioj: Yes, that's the way to go. The url is perhaps not the thing that we want to track.

ncloudioj:
@rlr Could you also update the document for this new event type? [here](https://github.com/mozilla/activity-stream/blob/master/data_events.md)

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+Cu.import("resource://gre/modules/Preferences.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "Services",
+                                  "resource://gre/modules/Services.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "CustomizableUI",
+                                  "resource:///modules/CustomizableUI.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "Social",
+                                  "resource:///modules/Social.jsm");
+
+const DEFAULT_MANIFEST_PREFS = {
```
k88hudson: could this go in its own file maybe?

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+    "pageSize": {
+      "share": {
+        "height": 430,
+        "width": 550
+      }
+    },
+    "shareURL": "https://www.tumblr.com/share?v=3&u=%{url}&t=%{title}&s=%{description}",
+    "version": 1,
+    "postActivationURL": "https://activations.cdn.mozilla.net/en-US/activated/tumblr.html",
+    "updateDate": Date.now(),
```
k88hudson: just confirming, this is supposed to be the date/time AS was supposed to be installed?

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+// based on SocialShare.sharePage in browser-social.js
+// target would be item clicked on for a context menu, but not handled in this case
+function windowProperty(window, eventTracker) {
+  return {
+    configurable: true,
+    enumerable: true,
+    writable: true,
+    value: {
+      copyLink: url => {
+        let clipboard = Cc["@mozilla.org/widget/clipboardhelper;1"].getService(Ci.nsIClipboardHelper);
```
k88hudson: does this need to be re-imported every time this method gets called?

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+        clipboard.copyString(url);
+        eventTracker.handleUserEvent({
+          event: "SHARE_FROM_TOOLBAR",
+          provider: "copy-link"
+        });
+      },
+      emailLink: (browser, url) => {
+        window.MailIntegration.sendLinkForBrowser(browser);
+        eventTracker.handleUserEvent({
+          event: "SHARE_FROM_TOOLBAR",
```
k88hudson: should this co in go in `common/event-constants`?

k88hudson:
Not sure if this is your patch, but the gmail option is not working for me with the following console error:

```
MozSocialAPI injectController: unable to attachToWindow for
https://accounts.google.com/ServiceLogin?service=mail&passive=true&rm=false&continue=https://mail.google.com/mail/?.......
TypeError: containingBrowser is nullMozSocialAPI.jsm:84

Error: Permission denied to access property "type"[Learn More]  ServiceLogin:2798:561
```

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+          provider: "email-link"
+        });
+      },
+      sharePage: (providerOrigin, graphData, target) => {
+        let SocialUI = window.SocialUI;
+        let Social = window.Social;
+        let messageManager = window.messageManager;
+        let gBrowser = window.gBrowser;
+        let OpenGraphBuilder = window.OpenGraphBuilder;
+        let ShareUtils = window.ShareUtils;
```
k88hudson: total nit, but you could  just do
```js
const {
  SocialUI,
  Social,
  ...
} = window;
```

k88hudson:
After disabling, and enabling, I get the following errors:

no ui showing up for share menu
```
console.error: CustomizableUI:
  TypeError: can't access dead object -- resource://app/modules/CustomizableUI.jsm:2115
console.error: CustomizableUI:
  TypeError: can't access dead object -- resource://app/modules/CustomizableUI.jsm:2115
console.error: CustomizableUI:
  TypeError: can't access dead object -- resource://app/modules/CustomizableUI.jsm:2115
console.error: CustomizableUI:
  TypeError: can't access dead object -- resource://app/modules/CustomizableUI.jsm:2115
...
```

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+        view.appendChild(item);
+        item = createElementWithAttrs(doc, "toolbarbutton", {
+          "class": "subviewbutton",
+          "label": "Email Link...",
+          "image": data.url("content/img/glyph-email-16.svg"),
+          "oncommand": "ShareUtils.emailLink(gBrowser.selectedBrowser, gBrowser.currentURI.spec);"
+        });
+        view.appendChild(item);
+        item = createElementWithAttrs(doc, "menuseparator", {"id": "menu_shareMenuSeparator"});
+        view.appendChild(item);
```
k88hudson: oldschool 😎

k88hudson:
This is working really well except for that uninstall/re-enable failure! Nice work

k88hudson:
Also one more thing -- for the event, I noticed it was being sent on a share *attempt*, not a successful share – is that intended?

mixedpuppy:
There isn't an easy way to know if the share was successful, share dialogs don't signal that.  

I did another addon at https://github.com/mixedpuppy/annotator that does detect share success and adds an annotation to history.  That addon has the ability to catch any share, whether via share menu or share dialogs such as those on youtube.  It currently only catches a couple providers though.

mixedpuppy:
regarding the MozSocialAPI message commented on earlier, that is part of the socialapi functionality that is being removed in bug 1289549.  That shouldn't cause any problem.

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+    "pageSize": {
+      "share": {
+        "height": 300,
+        "width": 550
+      }
+    },
+    "shareURL": "https://twitter.com/intent/tweet?url=%{url}&text=%{text}",
+    "postActivationURL": "https://activations.cdn.mozilla.net/en-US/activated/twitter.html",
+    "updateDate": Date.now(),
+    "installDate": Date.now()
```
mixedpuppy: Dates do not need to be a part of this, they are added by the manifest parser when installing the provider.  It may be better to look at https://github.com/mixedpuppy/socialapi-directory/blob/master/templates/data-en.json for reference, though there are some fields that would no longer be needed after bug 1289549 lands.

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+  },
+  "social.manifest.www-facebook.com": {
+    "activities": {
```
mixedpuppy: "activities" is cruft from an experiment, remove it.

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+          return;
+        }
+
+        let provider = Social._getProviderFromOrigin(providerOrigin);
+        if (!provider || !provider.shareURL) {
+          return;
+        }
+
+        let shareEndpoint = OpenGraphBuilder.generateEndpointURL(provider.shareURL, pageData);
+        window.open(shareEndpoint, "share-dialog", "chrome");
```
mixedpuppy: Looks like we're going to need to handle a default size and maybe some controls as well in the window.open call.  Opening gmail results in a zero sized window you can never close.

lib/ShareProvider.js
rlr: I keep forgetting how cool js is these days 👍

lib/ShareProvider.js
rlr: ah, I guess we can use the dimensions from the manifest

mixedpuppy: Note that those are for the content frame, not the outer dimension of a window.  But those are good indicators for pages that dont properly size,  I'd consider using that size +100px for width/height.  That will ensure space for the window plus some.  I'd also try to use min-width and min-height rather than width/height on the window (not sure if window args can do that)

lib/ShareProvider.js
```diff
@@ -0,0 +1,466 @@
...
+    this._createButton();
+    Overlay.init(this.eventTracker);
+  }),
+
+  /**
+   * Set up the prefs and enable a default set of social providers if the user
+   * hasn't enabled any.
+   */
+  _setupProviders: Task.async(function*() {
+    if (!Services.prefs.prefHasUserValue("social.activeProviders")) {
```
mixedpuppy: This also needs a check that there are not existing providers already (to prevent changing providers for existing share users).  && !SocialService.hasEnabledProviders

mixedpuppy: Oh, I just realized that's kind of what you're doing :)  Sorry.  hasEnabledProviders may be better than direct pref access though.

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
+const DEFAULT_MANIFEST_PREFS = {
+  "social.manifest.twitter-com": {
+    "author": "Twitter, Inc.",
+    "description": "Use your Firefox Twitter Share to quickly author tweets about the pages you discover across the web.  We shorten URLs for quick sharing using the t.co link shortener.",
+    "homepageURL": "https://twitter.com",
+    "icon32ULR": "images/twitter/icon32.png",
```
pdehaan: typo: `icon32ULR`?

pdehaan: Also, should we base64 inline the icon32.png like we do for `icon64URL` and `iconURL` below?

mixedpuppy: heh.  fixed on the activations site (not yet updated).  If it's easier, you can grab the base64 from here:
https://github.com/mixedpuppy/socialapi-directory/blob/production/en-US/twitter.html

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
+    "name": "Twitter",
+    "origin": "https://twitter.com",
+    "pageSize": {
+      "share": {
+        "height": 300,
+        "width": 550
+      }
+    },
+    "shareURL": "https://twitter.com/intent/tweet?url=%{url}&text=%{text}",
+    "postActivationURL": "https://activations.cdn.mozilla.net/en-US/activated/twitter.html"
```
pdehaan: Sucks that we seemingly have to embed a specific /en-US/ locale in this URL, but it 404ed hard on me when I omit the `/en-US/` bit and hoped that the server would detect my locale and serve me a translated page [like we do for most other mozilla.org pages]. :(

mixedpuppy: You don't have to include postActivationURL.  I could also modify the activation site to generate a per-local json file with all the providers.  Translations exist on the site for 16 or 17 languages already.  Then rather than embedding the json here, you could just read a localized json file.

mixedpuppy: btw, there really isn't a server, it's a set of static pages on a cdn.

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
...owNTgwMTE3NDA3MjA2ODExODA4M0NDMTM4MEMyQTVFQiIgeG1wTU06RG9jdW1lbnRJRD0ieG1wLmRpZDpCOEE0QzYzMUE2MTYxMUUyOEJFQUJDRTMzOERDQjM5MCIgeG1wTU06SW5zdGFuY2VJRD0ieG1wLmlpZDpCOEE0QzYzMEE2MTYxMUUyOEJFQUJDRTMzOERDQjM5MCIgeG1wOkNyZWF0b3JUb29sPSJBZG9iZSBQaG90b3Nob3AgQ1M2IChNYWNpbnRvc2gpIj4gPHhtcE1NOkRlcml2ZWRGcm9tIHN0UmVmOmluc3RhbmNlSUQ9InhtcC5paWQ6QUM3QUJGQTkzODIwNjgxMThDMTQ5OEFGOTgxQUJBQ0UiIHN0UmVmOmRvY3VtZW50SUQ9InhtcC5kaWQ6MDU4MDExNzQwNzIwNjgxMTgwODNDQzEzODBDMkE1RUIiLz4gPC9yZGY6RGVzY3JpcHRpb24+IDwvcmRmOlJERj4gPC94OnhtcG1ldGE+IDw/eHBhY2tldCBlbmQ9InIiPz4XCE4/AAABzElEQVR42mK0Dp/LyMDAUAHEOUAsxTC8wDMgngLEHSxQT7YxDE8gBfXbfyYgkc0w/EEOyKPSI8Cj0kwMIwSMenTUo6MeHdyAZTA6ipebncHBXIHBTF+aQVFGgIGfl4NBAIgZgU2bL99+Mfz4+Yfh89dfDK/ffQPj9hmHhpZHOTlYGZJDjBgCXNUZONixO42Hiw2MRQS5wIEAAkPKowoyggwdJc4MMhJ8wzfpyojzM0yp82IQ4GMfvoUROxsLQ0eZM009OSg8GuGtw6AgLTC8qxdQgRPpqzP861ErIzlwCTrs61EzPcIdp20H7zCs23Wd4e6jdwy/f/8dmh5VUxTCK7/1wB2i6shBn3TFhXnwym/ce2N45FFODvwJ6tHTD8PDo2yszHjlQe3a0d7LqEdHPTrq0VGPDkVA05bRkRVJNNe/78R9hroJ+4d/jN68925kJN2b91+PDI/efjACYvTlm68MHz//GP4evfXg7cioXm7eezMyPEpKjNK0HrWJmEdRPUlI/2jLaNSjox4d9eioR0c9OurRUY+OenTUo6MeHfXoqEdHPTrq0VGP0t+jT0eAP5+CPDp1BHh0KmhwrBMas1kMw3ODDygiOwECDADJwGV3tLQaBAAAAABJRU5ErkJggg==",
```
pdehaan: Interestingly, this is 58x58, according to my Nightly developer tools.
Odd that the 16x16 `iconURL` data URI below looks significantly longer than the 58x58 URI.

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
+    "version": 2,
+    "postActivationURL": "https://activations.cdn.mozilla.net/en-US/activated/facebook.html"
+  },
+  "social.manifest.mg-mail.yahoo.com": {
+    "author": "Yahoo! Inc",
+    "description": "Quickly send web pages to yourself and others with enhanced link previews and more.",
+    "homepageURL": "http://mail.yahoo.com",
+    "icon32URL": "https://s.yimg.com/nq/nr/icons/ymail32.png",
+    "icon64URL": "https://s.yimg.com/nq/nr/icons/ymail64.png",
+    "iconURL": "https://s.yimg.com/nq/nr/icons/ymail16.png",
```
pdehaan: All the Facebook icons were data URIs, one of the Twitter images was a local/relative image path.
All these Yahoo! Mail icons are absolute. Not sure if we need to tweak these, or if there is any risk of cross domain loading images may leak any browsing history.

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
+    "name": "Yahoo Mail",
+    "origin": "https://mg.mail.yahoo.com",
+    "shareURL": "https://mg.mail.yahoo.com/d/app/compose?url=%{url}&title=%{title}",
```
pdehaan: Missing `pageSize` properties here?

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
+    "iconURL": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAIAAACQkWg2AAABEElEQVQoz2Ng2PCWNERVDevBCKcGNOl1b7k3vQMihjVvUMThqg33f5Dd8Z5h7RuI6rCTn298+gNEDde+RZ76zIyuYc2bumvfrnz8Y3/wI8Pat+wb3z769vc/DLTc+MayHk3DujeuRz7+/Pv/19//u17+mnzn+7c/UNWffv9T3PmeAV3D2jeRpz///fcfE9z/+pdvyzsMDevfCmx+t+PlL0wN8x78gHoMPZTWvsk8/+X7HxRbPvz6p7PnA7ZQ2gDyqNCmt723vsFVf/79Dxg+KMYjO8n64McVj3+8/vnvx99/L378Xfvkp8X+j+iq0SKOd/M71d3vdfZ+EN/2HhhuwNggImmsB8UalhRBm8SHDQEA5UPgBrwMoL4AAAAASUVORK5CYII=",
+    "name": "Twitter",
+    "origin": "https://twitter.com",
+    "pageSize": {
+      "share": {
+        "height": 300,
+        "width": 550
+      }
+    },
+    "shareURL": "https://twitter.com/intent/tweet?url=%{url}&text=%{text}",
```
pdehaan: I don't know anything about anything, but do we need to URI encode the `text` here, or is that handled somewhere else, or not needed?

lib/ShareManifests.js
mixedpuppy: no, those are optional.  I only used them for situations where a share page does weird stuff and we cannot size the iframe based on then content.

lib/ShareManifests.js
mixedpuppy: We should not have any icon pulling from the web since they are directly used in fx ui.  I'll also look at why the build scripts for the activation site didn't catch those.

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
...dO/XAL39xYQdLQdhVnQb8tCmEMLoyq7aJsOaSZg4mR/U1Qfo4UuvHFTyUknd0BM9yWdJVoeqcFmDXaqDmokuXX/t4IMW+sbIJDkUnouwLmMxKclBhgixjrBOqRfl9dX8beDhNP3teDm7erOHlxAbbnwZVqZ+FK+S4VIo5HmQ0/1O3ANccja4OjRF7B2+dv4sR3c5C0aw+5PWsaqJJXtmy3uSlNdEn7GC1L764XyNDpO5XQmjlG4pDUhIoFdshim4TYNBe6Mnjn6CHaF1hx7rcVfH1tGWP9nXhxrL+56dT3f6A/l8Irj9yLQUpIScC/Lm/gg+kiZpbKyFJoWAZH19ghSgZbaTsKm7aHrpSJFx7ON5lLZQufziyiryMZoqcoy58d3g9D2zZlqLcDTz/Yi4nTM5gulpEmy9kTptDjDIbmEbBDFrtsteOGmDU6YuwvW4bpz430hUAbD7v9+OOD4CrIMjkcnhefH5qi5FDkZsUBiSxiHsgbKmbvmdklvPT5j/h4ejFEHz2YQ2/GDHKAi1JNQkrZ7mquWJJy2vF4hBGYB85qL9w8zlFROXZm1n//iOJ6/94sHit0+d+daQOdKQPLVPkMTYNUDlnvtFvMYDaNGg0nYhofMdY6qtC3c6v+Ic0kDd9Ts5RQTUsoBIJGzQ3kun4PiTnHLruaDfItiwA36BFXc8Fh//sK0WxFcoPlcGwdQbNQKi65DDbSt4xevKjFKp5eT4AmP66oNPisWywwgwoVCFKRYq7qCrXTw8AqItqr81Uw4noEtNbGFlVe1aXE0kP7oqchZL3SY6qI0SYl0tb8vhqJkmisr4/W3AnzFDKZDH5fmHfe//LsA47rjk5OTp7P5/Orhgoh7wAgYvpdIwRKtbGFqCNzrU6msLS4qE9NTR0mxqOjo6Ol8fHxC9sW0/3ol78qOPzuxSborTtbAJXLS3+uhegLa1WSaAQK0HyKuthZ6tsNUQvU2aBp9Th7wjB0t6en+0rfgbwcHh6eC1y9rSY26TJweX615aqo8cFEuSrb6XogGHRmiwRUXGnp5RzT+k2IslqjKUFfV4eGhq4ODAzUz7vGNUQZvhBeHG36d6M3Zl34Hou7guTSZiWtS0v5lwLXr2KmaUJ7KlV6XbOtO7Gd/D89AglprR0b1N5MG8LigqJp21eofwUYABnolxra9d85AAAAAElFTkSuQmCC",
```
pdehaan: Also interesting, this image is actually 30x25

lib/ShareManifests.js
mixedpuppy: It's handled in the call to OpenGraphBuilder.generateEndpointURL

lib/ShareManifests.js
mixedpuppy: 16x16 may not have been optimized, 58x58 is indeed strange, I probably messed up at some point there.

lib/ShareProvider.js
```diff
@@ -0,0 +1,348 @@
...
+    }
+  },
+  observe: window => {
+    Overlay.setWindowScripts(window);
+  },
+  setWindowScripts: (window, eventTracker) => {
+    Object.defineProperty(window, "ShareUtils", windowProperty(window, eventTracker));
+
+    if (window.SocialShare.shareButton) {
+      window.SocialShare.shareButton.setAttribute("hidden", "true");
```
pdehaan: Should `"true"` have quotes here?

lib/ShareManifests.js
pdehaan: Is `version` a required field in this schema? I noticed it was missing for Twitter.
Or more specifically/annoyingly, I wrote a quick Joi schema validation for these manifests and it complained about the following:

```
social.manifest.twitter-com:
  child "icon32URL" fails because ["icon32URL" is required].
  child "version" fails because ["version" is required].
  "icon32ULR" is not allowed.
```

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
+    "origin": "https://mail.google.com",
+    "pageSize": {
+      "share": {
+        "height": 500,
+        "width": 500
+      }
+    },
+    "shareURL": "https://mail.google.com/mail/?view=cm&ui=2&tf=0&fs=1&body=%{body}&su=%{title}",
+    "postActivationURL": "https://activations.cdn.mozilla.net/en-US/activated/gmail.html"
+  },
```
pdehaan: Same comment as above re: required `version` key:

```
social.manifest.mail-google.com:
  child "version" fails because ["version" is required]
```

lib/ShareManifests.js
pdehaan: Is `homepageURL` a required field?

```
social.manifest.www-linkedin.com:
  child "homepageURL" fails because ["homepageURL" is required].
  child "version" fails because ["version" is required]
```

lib/ShareManifests.js
```diff
@@ -0,0 +1,108 @@
...
+        "width": 550
+      }
+    },
+    "shareURL": "https://www.tumblr.com/share?v=3&u=%{url}&t=%{title}&s=%{description}",
+    "version": 1,
+    "postActivationURL": "https://activations.cdn.mozilla.net/en-US/activated/tumblr.html"
+  }
+};
+
+module.exports = DEFAULT_MANIFEST_PREFS;
```
pdehaan: For anybody interested, here was my Joi schema validator script thing:

```js
const Joi = require('joi');

const manifests = require('./ShareManifests');

function lintManifest(manifest, cb) {
  const dataUriSchema = Joi.string().uri({
    scheme: ['data']
  });
  const itemSchema = Joi.object().keys({
    author: Joi.string().required(),
    description: Joi.string().required(),
    homepageURL: Joi.string().uri().optional(), // OPTIONAL
    iconURL: dataUriSchema.required(),   // NOTE: `data:` URI
    icon32URL: dataUriSchema.required(), // NOTE: `data:` URI
    icon64URL: dataUriSchema.required(), // NOTE: `data:` URI
    name: Joi.string().required(),
    origin: Joi.string().uri().required(),
    pageSize: Joi.object().optional().keys({ // OPTIONAL
      share: Joi.object().required().keys({
        height: Joi.number().integer().required(),
        width: Joi.number().integer().required()
      })
    }),
    postActivationURL: Joi.string().uri().required(),
    shareURL: Joi.string().required(), // This isn't a `uri()` because of odd tokens.
    version: Joi.number().optional() // OPTIONAL
  });

  // HACK: I'm a terrible QA and person.
  const obj = {};
  Object.keys(manifest).forEach((site) => obj[site] = itemSchema);

  Joi.validate(manifest, Joi.object().keys(obj), {abortEarly: false}, cb);
}

lintManifest(manifests, (err, value) => {
  if (err) {
    const messages = err.details.map(({message, path}) => `${path}: ${message}`);
    console.error(messages.join('\n'));
    process.exit(1);
  }
  console.log(`manifest is valid!`);
});
```

### Output:
```
social.manifest.twitter-com.icon32URL: "icon32URL" is required
social.manifest.twitter-com.icon32ULR: "icon32ULR" is not allowed
social.manifest.mg-mail.yahoo.com.iconURL: "iconURL" must be a valid uri with a scheme matching the data pattern
social.manifest.mg-mail.yahoo.com.icon32URL: "icon32URL" must be a valid uri with a scheme matching the data pattern
social.manifest.mg-mail.yahoo.com.icon64URL: "icon64URL" must be a valid uri with a scheme matching the data pattern
```

lib/ShareProvider.js
rlr: `SocailService.hasEnabledProviders()` is perfect 👍

lib/ShareManifests.js
rlr: that's cool @pdehaan thanks!

lib/ShareProvider.js
rlr: I *think* the answer is it doesn't matter. Just did a quick search and it looks like both are all over mozilla-central: http://searchfox.org/mozilla-central/search?q=setAttribute%28%22hidden%22&path=

You should land our eslint rules with the fixes to mozilla central 😜

data_events.md
```diff
@@ -119,6 +119,21 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as
 }
 ```
 
+#### Sharing a site from toolbar
```
k88hudson: Can you add a note here that this is a share *attempt* (i.e. the user clicks a share button), not necessarily a successful or failed share flow?

k88hudson:
Ok, all the issues I noted were fixed.I added a small note about documentation, but this looks good to go!

pdehaan:
For giggles, I wrote a dirty scraper which extracted all the data URIs and wrote them to disk so I could see the file sizes (since a few data:uri strings looked somewhat "unoptimized" to my special QA eyes):

```sh
$ ls -lsh output | awk '{print $6,$10}'

296B social.manifest.mail-google.com-iconURL.png
1.1K social.manifest.mail-google.com-icon32URL.png
2.2K social.manifest.mail-google.com-icon64URL.png

2.9K social.manifest.mg-mail.yahoo.com-iconURL.png
3.0K social.manifest.mg-mail.yahoo.com-icon32URL.png
3.2K social.manifest.mg-mail.yahoo.com-icon64URL.png

329B social.manifest.twitter-com-iconURL.png
4.4K social.manifest.twitter-com-icon32URL.png
1.4K social.manifest.twitter-com-icon64URL.png

3.8K social.manifest.www-facebook.com-iconURL.png
956B social.manifest.www-facebook.com-icon32URL.png
1.4K social.manifest.www-facebook.com-icon64URL.png

716B social.manifest.www-linkedin.com-iconURL.png
1.2K social.manifest.www-linkedin.com-icon32URL.png
2.5K social.manifest.www-linkedin.com-icon64URL.png

241B social.manifest.www-tumblr.com-iconURL.png
329B social.manifest.www-tumblr.com-icon32URL.png
512B social.manifest.www-tumblr.com-icon64URL.png

$ du -sh output # 76K
```

And for my own future reference, here is my embarassing script:

```js
const fs = require('fs');
const path = require('path');

// npm install data-uri-to-buffer mime-types
const dataUriToBuffer = require('data-uri-to-buffer');
const mime = require('mime-types');

const ShareManifests = require('./ShareManifests');

// Convert mime-type to file extension.
const getFileExt = (dataUri) => mime.extension(/data:(.*?);/.exec(dataUri)[1]);

for (let key in ShareManifests) {
  const manifest = ShareManifests[key];
  const icons = ['iconURL', 'icon32URL', 'icon64URL'];
  icons.forEach((size) => {
    const dataUri = manifest[size];
    const filename = path.join('output', `${key}-${size}.${getFileExt(dataUri)}`);
    const contents = dataUriToBuffer(dataUri);
    fs.writeFileSync(filename, contents);
  });
}
```

mixedpuppy:
Since these are being "built in", it might be better to move those icons out to files, and use some css to load them.

pdehaan:
In fact, here's my crude optimizer using imagemin+zopfli. If my garbage-fire-esque code can be trusted (disclaimer: usually not), we could be saving about 58% filesize if we pre-optimize the images using imagemin+zopfli:

| NAME                                            | ORIGINAL | OPTIMIZED | %    |
|----|----:|----:|----:|
| social.manifest.mail-google.com-icon32URL.png   | 1087     | 1029      | 5.3  |
| social.manifest.mail-google.com-icon64URL.png   | 2268     | 2261      | 0.3  |
| social.manifest.mail-google.com-iconURL.png     | 296      | 296       | 0.0  |
| social.manifest.mg-mail.yahoo.com-icon32URL.png | 3116     | 244       | 92.2 |
| social.manifest.mg-mail.yahoo.com-icon64URL.png | 3305     | 331       | 90.0 |
| social.manifest.mg-mail.yahoo.com-iconURL.png   | 3004     | 198       | 93.4 |
| social.manifest.twitter-com-icon32URL.png       | 4542     | 618       | 86.4 |
| social.manifest.twitter-com-icon64URL.png       | 1465     | 1461      | 0.3  |
| social.manifest.twitter-com-iconURL.png         | 329      | 326       | 0.9  |
| social.manifest.www-facebook.com-icon32URL.png  | 956      | 788       | 17.6 |
| social.manifest.www-facebook.com-icon64URL.png  | 1438     | 411       | 71.4 |
| social.manifest.www-facebook.com-iconURL.png    | 3858     | 555       | 85.6 |
| social.manifest.www-linkedin.com-icon32URL.png  | 1263     | 1169      | 7.4  |
| social.manifest.www-linkedin.com-icon64URL.png  | 2544     | 2048      | 19.5 |
| social.manifest.www-linkedin.com-iconURL.png    | 716      | 655       | 8.5  |
| social.manifest.www-tumblr.com-icon32URL.png    | 329      | 238       | 27.7 |
| social.manifest.www-tumblr.com-icon64URL.png    | 512      | 364       | 28.9 |
| social.manifest.www-tumblr.com-iconURL.png      | 241      | 165       | 31.5 |
|                                                 | 31269    | 13157     | 57.9 |

```js
const fs = require('fs');
const path = require('path');

// npm i cli-table imagemin imagemin-zopfli -S
const Table = require('cli-table');
const imagemin = require('imagemin');
const imageminZopfli = require('imagemin-zopfli');

const optimizedPct = (original, optimized) => ((original - optimized) / original * 100).toFixed(1);

imagemin(['output/*.png'], 'optimized', {
  use: [
    imageminZopfli({
      more: true
    })
  ]
}).then(() => {
  console.log('Images optimized');

  const table = new Table({
    head: ['NAME', 'ORIGINAL', 'OPTIMIZED', '%'],
    chars: { 'top': '' , 'top-mid': '' , 'top-left': '' , 'top-right': ''
         , 'bottom': '' , 'bottom-mid': '' , 'bottom-left': '' , 'bottom-right': ''
         , 'left': '|' , 'left-mid': '' , 'mid': '' , 'mid-mid': ''
         , 'right': '|' , 'right-mid': '' , 'middle': '|' }
  });

  const files = fs.readdirSync('output');
  let optimizedTotal = 0;
  let originalTotal = 0;

  files.forEach((file) => {
    const originalSize = fs.statSync(path.join('output', file)).size;
    const optimizedSize = fs.statSync(path.join('optimized', file)).size;
    table.push([file, originalSize, optimizedSize, optimizedPct(originalSize, optimizedSize)]);
    originalTotal += originalSize;
    optimizedTotal += optimizedSize;
  });

  table.push(['', originalTotal, optimizedTotal, optimizedPct(originalTotal, optimizedTotal)]);

  console.log(table.toString());
});
```

rlr:
@pdehaan Can you file a followup bug for the image optimization and size inconsistencies? I don't want to delay the cutting the release further. Thanks!

pdehaan:
Filed as https://github.com/mozilla/activity-stream/issues/1097: "Optimize (and resize) images in lib/ShareManifests.js?"

PR #1091, opened 2016-08-15
1072 - Add PrefsProvider

k88hudson:
This patch:

- adds a `PrefsProvider` which dispatches changes and handles requests from content
- adds a new `Prefs` reducer, which stores all prefs from the add-on
- replaces code in `setRowsOrError` that sends recommended, highlights with selectors instead
- adds unit tests via mocha (uses a depedency injection strategy here)
- adds integration tests with activity stream via jpm tests

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1091)


k88hudson:
@sarracini r?

k88hudson:
cc @piatra, if you are using prefs for stuff, this might affect you

k88hudson:
Switching the review to @piatra since he needs it for a patch he's working on

lib/ActivityStreams.js
```diff
@@ -310,13 +317,6 @@ ActivityStreams.prototype = {
     this.send(am.actions.Response("EXPERIMENTS_RESPONSE", this._experimentProvider.data), worker);
   },
 
-  _respondToPrefsRequest() {
```
piatra: I can't add the comment where I *really* want to 😞 , `ActivityStreams.js:355` calls this function which was removed.

piatra: Also the old behavior was to respond with a PREFS_RESPONSE to recommendation toggles. Is that case still covered?

content-test/reducers/Prefs.test.js
```diff
@@ -0,0 +1,74 @@
...
+    const state = Prefs(undefined, {});
+    assert.deepEqual(state, Prefs.INITIAL_STATE);
+  });
+  describe("PREFS_REQUEST", () => {
+    it("should set isLoading to true", () => {
+      const state = Prefs(undefined, {type: "PREFS_REQUEST"});
+      assert.isTrue(state.isLoading);
+    });
+    it("should return a new object", () => {
+      checkIndependence("PREFS_REQUEST");
```
piatra: How about moving this in a `beforeEach`?

lib/ActivityStreams.js
```diff
@@ -113,6 +114,12 @@ function ActivityStreams(metadataStore, options = {}) {
   this._setupButton();
   NewTabURL.override(this._newTabURL);
   this._setHomePage();
+  this.prefsProvider = new PrefsProvider({
```
piatra: question/nit Any reason why this isn't following the same underscore naming convention like the other instances initiated in the constructor?

content-src/selectors/selectors.js
```diff
@@ -80,13 +82,13 @@ module.exports.selectNewTabSites = createSelector(
...
     let topHighlights = spotlightRows;
-    if (WeightedHighlights.weightedHighlights && weightedHighlightsRows.length) {
+    if (weightedHighlightsEnabled && weightedHighlightsRows.length) {
       topHighlights = weightedHighlightsRows;
     }
 
     return {
       TopSites: Object.assign({}, TopSites, {rows: topSitesRows}),
-      Spotlight: Object.assign({}, Spotlight, {rows: topHighlights}),
+      Spotlight: Object.assign({}, Spotlight, {rows: topHighlights, weightedHighlights: weightedHighlightsEnabled}),
```
piatra: 👍

content-test/selectors/selectors.test.js
```diff
@@ -51,18 +51,29 @@ describe("selectors", () => {
...
     let state = selectSpotlight(fakeState);
 
+    function setup(custom) {
+      state = selectSpotlight(Object.assign({}, fakeState, custom));
+      return state;
+    }
+
+    beforeEach(() => setup());
+
     // Tests that provided sites don't get selected, because they don't
```
piatra: Is this comment valid? `selectSpotlight` Does not filter things out it just sorts them in the order they should be selected, and the assert in the function below tests that length is increased by 2 so everything is selected.

content-test/selectors/selectors.test.js
```diff
@@ -175,10 +186,8 @@ describe("selectors", () => {
     });
     it("should render the correct Spotlight items", () => {
       let weightedHighlights = {
-        WeightedHighlights: {
-          rows: [{url: "http://foo.com"}, {url: "http://www.foo.com"}],
-          weightedHighlights: true
-        }
+        WeightedHighlights: {rows: [{url: "http://foo.com"}, {url: "http://www.foo.com"}]},
+        Prefs: {prefs: {weightedHighlights: true}}
```
piatra: Can you add an assertion that this prefs is present in the state?

common/action-manager.js
```diff
@@ -46,7 +46,9 @@ const am = new ActionManager([
   "NOTIFY_TOGGLE_RECOMMENDATIONS",
   "RECEIVE_RECOMMENDATION_TOGGLE",
```
piatra: The code for this action & tests have been removed. Are we still using this somewhere?

lib/ActivityStreams.js
piatra: Similar thing on `ActivityStreams.js:408`

common/action-manager.js
k88hudson: good catch 👍

content-test/selectors/selectors.test.js
k88hudson: you're right, this function was changed to sort rather than filter

lib/ActivityStreams.js
k88hudson: you're right, it should be `_` (we use that for "private" properties, albeit somewhat inconsistently)

lib/ActivityStreams.js
k88hudson: oops, must have added those back in, in the rebase somehow – I will remove

k88hudson:
@piatra ok, updated!

piatra:
All green 👍 r+ 
🚢 🚢 🚢

PR #1104, opened 2016-08-16
chore(package): Use svgo to optimise and minimise svg files during packaging.

Standard8:
The talk of optimisation elsewhere reminded me that we used to run svgo on Hello's svg files prior to landing them (we never got around to making it automatic). Most of our images were svg files, and a lot came with extra header data that wasn't needed, hence wanting to run svgo.

With the bundling system we have for activity stream it seems we can easily hook svgo up to run during bundle.

When I bundle locally this branch versus current master, it is saving about 16K on the size of files in the data/content/img directory, and about 8.1K on the xpi size.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1104)


k88hudson:
Very nice!! Thank you

Issue #1101, opened 2016-08-16
Debug the weights in the recommender system


piatra:
Need: As a developer of activity stream I want to change the weights associated with the features used in scoring a URL for highlights. I want the update to be as easy as possible (no browser restarts).

Deliverable: A pref in `about:config` called `weightedHighlightsCoefficients` which allows you to change the weights in the system without having to do a browser restart to see the changes. *You still need to refresh the page* 😅 
Figuring out what the numbers mean requires [reading the comments](https://github.com/mozilla/activity-stream/blob/2a72af00acd6b13627d1f9b3c0c3f38bf3fb5cf9/common/recommender/Baseline.js#L27)

pdehaan:
@piatra Was this resolved by #1106?

sarracini:
^ Yes it was

PR #1106, opened 2016-08-16
Weights prefs

piatra:
As described in #1101 

@sarracini || @k88hudson please review.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1106)

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -2,16 +2,19 @@
 
 const URL = require("../../lib/vendor.bundle.js").urlParse;
 const {INFINITE_SCROLL_THRESHOLD} = require("../constants");
+const COEFFICIENTS = [0.4, 0.7, 0.1, -0.4, -0.2, -0.1];
 
 /**
  * Score function for URLs.
  * See tests and `scoreEntry` comments for more insight into how the score is computed.
  *
  * @param {Array.<URLs>} history - User history used to assign higher score to popular domains.
+ * @param {Object} options - settings for the scoring function.
  */
 class Baseline {
-  constructor(history) {
+  constructor(history, options = {}) {
     this.domainCounts = history.reduce(this.countDomainOccurrences, new Map());
+    this.options = options;
   }
 
   scoreEntry(entry) {
@@ -30,9 +33,13 @@ class Baseline {
        imageCount, isBookmarked, hasDescription],
       // Features weights: Positive values decrease the score proportional to a factor of feature * weight.
       //                   Negative values increase score proportional to a factor of feature * weight.
-      [0.4, 0.7, 0.1, -0.4, -0.2, -0.1]);
+      this.options.highlightsCoefficients || COEFFICIENTS);
 
     return Object.assign({}, entry, {score}, {host});
+  }
+
+  updateOptions(options) {
+    this.options = options;
   }
 
   /**
--- a/content-test/recommender/Baseline.test.js
+++ b/content-test/recommender/Baseline.test.js
@@ -126,4 +126,20 @@ describe("Baseline", () => {
     let items = baseline.score(fakeUrls);
     assert.equal(items[items.length - 1].title, "Very visited");
   });
+
+  describe("options", () => {
+    const options = {highlightsCoefficients: [1, 2, 3]};
+    it("should instantiate the recommender with empty object by default", () => {
+      baseline = new Baseline(fakeHistory);
+      assert.deepEqual(baseline.options, {});
+    });
+    it("should instantiate the recommender with the correct settings", () => {
+      baseline = new Baseline(fakeHistory, options);
+      assert.deepEqual(baseline.options, options);
+    });
+    it("should update the correct settings", () => {
+      baseline.updateOptions(options);
+      assert.deepEqual(baseline.options, options);
+    });
+  });
 });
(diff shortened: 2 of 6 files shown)


common/recommender/Baseline.js
```diff
@@ -10,8 +11,9 @@ const {INFINITE_SCROLL_THRESHOLD} = require("../constants");
  * @param {Array.<URLs>} history - User history used to assign higher score to popular domains.
  */
 class Baseline {
-  constructor(history) {
+  constructor(history, options = {}) {
```
sarracini: are you able to update the @param comment with the addition of options?

common/recommender/Baseline.js
```diff
@@ -30,11 +32,15 @@ class Baseline {
...
       // Features weights: Positive values decrease the score proportional to a factor of feature * weight.
       //                   Negative values increase score proportional to a factor of feature * weight.
-      [0.4, 0.7, 0.1, -0.4, -0.2, -0.1]);
+      this.options.highlightsCoefficients || COEFFICIENTS);
 
     return Object.assign({}, entry, {score}, {host});
   }
 
+  updateOptions(options = {}) {
+    this.options = options;
```
sarracini: i'm assuming the only time you're ever going to call ```updateOptions``` is when you need to pass in new options - so i think it's safe to assume you'll have a param here, so you can remove ```= {}```

content-test/recommender/Baseline.test.js
```diff
@@ -126,4 +126,20 @@ describe("Baseline", () => {
...
+    it("should instantiate the recommender with empty object by default", () => {
+      baseline = new Baseline(fakeHistory);
+      assert.deepEqual(baseline.options, {});
+    });
+    it("should instantiate the recommender with the correct settings", () => {
+      baseline = new Baseline(fakeHistory, options);
+      assert.deepEqual(baseline.options, options);
+    });
+    it("should update the correct settings", () => {
+      baseline.updateOptions(options);
```
sarracini: you didn't actually change your options ;) 
maybe 
```js
const newOptions = {highlightsCoefficients: [4, 5, 6]};
baseline.updateOptions(newOptions);
assert.deepEqual(baseline.options, newOptions);
```

lib/ActivityStreams.js
```diff
@@ -170,6 +176,23 @@ ActivityStreams.prototype = {
...
+   * @private
+   */
+  _loadRecommender() {
+    // Only need to load history items once per session.
+    if (this._baselineRecommender !== null) {
+      return;
+    }
+
+    this._memoized.getAllHistoryItems().then(historyItems => {
+      let highlightsCoefficients = JSON.parse(simplePrefs.prefs.weightedHighlightsCoefficients);
```
sarracini: 👍 🎉

lib/ActivityStreams.js
```diff
@@ -411,6 +432,17 @@ ActivityStreams.prototype = {
       this._prefsProvider.actionHandler(args);
     };
     this.on(CONTENT_TO_ADDON, this._contentToAddonHandlers);
+
+    simplePrefs.on("", prefName => {
```
sarracini: hmm is it possible use the PrefsProvider to do this? that way we have a communal place where the prefs are being changed? @k88hudson thoughts?

k88hudson: Right now there isn't really a convenient way to use `PrefsProvider` for this use case – it's more for handling communication with content, although we could add something to do it (i.e. an `addListener`/`on` method) – the advantage of that would be we could handle removing listeners on unload in once place.

For now, this seems reasonable (we could file the enhancement as separate ticket) -- although this method needs to be cached on the instance and removed at unload time.

common/recommender/Baseline.js
```diff
@@ -2,6 +2,7 @@
 
 const URL = require("../../lib/vendor.bundle.js").urlParse;
 const {INFINITE_SCROLL_THRESHOLD} = require("../constants");
+const COEFFICIENTS = [0.4, 0.7, 0.1, -0.4, -0.2, -0.1];
```
pdehaan: This seems to be the same values as in [package.json](https://github.com/mozilla/activity-stream/pull/1106/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R185) edits below:

```js
 +    {
 +      "name": "weightedHighlightsCoefficients",
 +      "title": "recommender weights",
 +      "description": "weightedHighlightsCoefficients",
 +      "type": "string",
 +      "value": "[0.4, 0.7, 0.1, -0.4, -0.2, -0.1]",
 +      "hidden": true
      }
```

Not sure if there is a way to consolidate the values so they don't get out of sync if somebody tweaks these values in the future.

content-test/recommender/Baseline.test.js
piatra: The beforeEach created a new baseline for every `it` block which by default has no options. If the assert passes it means it got the new options.

common/recommender/Baseline.js
piatra: @pdehaan you can load the package.json and read the values. Not sure if that's a good practice™®

lib/ActivityStreams.js
piatra: /note to self: unload method.

lib/ActivityStreams.js
sarracini: let's wrap this in a try/catch to handle faulty pref values and then make the constructor fallback to defaults

sarracini:
🎉  R+ from me!

PR #1117, opened 2016-08-17
fix (addon): #1109, #1107 use separate window per provider and focus it on share

### LINKED ISSUE #1107: Sharing pages doesn't bring the popup window to the foreground
add-ons_manager

### Steps to reproduce:
1. Go to some page on The Internet, for example http://www.ign.com/articles/2016/08/16/no-mans-sky-review?read=&page=2
2. Share the page to, say, Twitter using the Share button in the toolbar:
    screenshot_8_16_16__3_21_pm
3. Make sure the Twitter popup is slightly obscured behind some other window.
4. Share the page again.

### Actual results:
If the Twitter popup isn't fully obscured, you can see that it refreshes, but doesn't come into focus. So it's _super_ easy to use the share panel, but then completely lose a window since it never is forced into the foreground.

### LINKED ISSUE #1109: Share window doesn't resize itself after switching between different providers
add-ons_manager

### Steps to reproduce:
1. Go to a page and try sharing it to Twitter (but don't submit).
2. Try sharing the page to Gmail (but don't submit).
3. Close the sharing window.
4. Share the page to Gmail (but don't submit).

### Actual results:
share_a_link_on_twitter
**Figure 1:** Initial Twitter share page.

---

gmail_-_free_storage_and_email_from_google
**Figure 2:** Initial Gmail share page (after launching [unclosed] Twitter share page).

---

gmail_-_free_storage_and_email_from_google
**Figure 3:** Initial Gmail share page (if launched first).

rlr:
This fixes makes each provider have it's own window so you can have multiple share windows open at once. And it makes sure to focus the window on share if it was already open in the background.

r? @ncloudioj 

Fixes #1107 
Fixes #1109 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1117)

### DIFF
--- a/lib/ShareProvider.js
+++ b/lib/ShareProvider.js
@@ -108,7 +108,7 @@ function windowProperty(window, eventTracker) {
         if (size && size.height) {
           features = `${features},innerHeight=${size.height}`;
         }
-        window.open(shareEndpoint, "share-dialog", features);
+        window.open(shareEndpoint, `share-dialog-${provider.name}`, features).focus();
 
         eventTracker.handleUserEvent({
           event: "SHARE_FROM_TOOLBAR",


ncloudioj:
r+

PR #1118, opened 2016-08-17
feat(content): Closes #1074 Add rating system for metadata

### LINKED ISSUE #1074: Add Rating System for Metadata
@k88hudson and I discussed adding a rating system for the metadata based on the source (either embedly or fathom) that it came from. You'll be able to switch between metadata sources and give each highlight a rating based on their image, title, description, etc... 

It'll be hidden behind a pref. Switching the pref on will let you pick your metadata source and then show the UI for the rating. Each rating will be sent as a ping with a payload {url, metadata_source, rating} so we can compare which urls from which source are better than others. 

ref: #1054 

⭐ ⭐ ⭐ ⭐ ⭐ 

cc @edmoz @nchapman

sarracini:
@k88hudson r?

Fixes #1074

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1118)


content-src/components/Rating/Rating.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+  },
+  createStars() {
+    let stars = [];
+    for (let i = 1; i <= this.state.numStars; i++) {
+      stars.push(<span
+        id={`star-${i}`}
+        className={classNames("icon icon-rating-default", {active: this.state.activeRatings[i]})}
+        onClick={() => this.onClick(i)}
+        style={{transform: `translate(${((i - 1) * 25)}px)`}}
```
pdehaan: The `25` here feels a bit magic. Is that just the width of the SVG star icon?

sarracini: heh it is magic - its just a nice distance to put the stars apart from each other. i have no oppositions on changing this to be a little more calculated. any suggestions?

pdehaan: :shrug:

We could just use some canned component like https://github.com/ndyag/react-rater/
But I don't think anything we find won't be as artisanal or lean as your module.

content-src/components/Rating/Rating.scss
```diff
@@ -0,0 +1,17 @@
+.rating-container {
+  padding-top: 10%;
```
k88hudson: make this `$base-gutter/2`

content-src/components/Rating/Rating.scss
```diff
@@ -0,0 +1,17 @@
...
+  transform: scaleX(-1);
```
k88hudson: Why all the transforms here? why not just use `display:flex`

package.json
```diff
@@ -184,6 +184,20 @@
...
+    },
+    {
+      "name": "metadataRatingSystem",
+      "title": "Rate this metadata",
+      "type": "bool",
+      "value": false,
+      "hidden": true
+    },
+    {
+      "name": "metadataSource",
```
k88hudson: couldn't we just pass along the URL associated? `i.e. embedly.endpoint`? why the extra pref?

content-src/components/Rating/Rating.js
```diff
@@ -0,0 +1,45 @@
+const React = require("react");
+const {connect} = require("react-redux");
+const {justDispatch} = require("selectors/selectors");
+const {actions} = require("common/action-manager");
+const classNames = require("classnames");
+
+const Rating = React.createClass({
+  getInitialState() {
+    return {numStars: 5, activeRatings: {}, currentRating: null};
```
k88hudson: why do you need both activeRating and currentRating?

content-src/components/Rating/Rating.scss
sarracini: so this is a trick for the hovering css to work. it needs to be flipped so that when you hover over the svg the previous siblings highlight instead of the succeeding siblings. this was the only way i could get it to work, is there another css trick that's better?

content-src/components/Rating/Rating.scss
```diff
@@ -0,0 +1,17 @@
...
+
+  .icon-rating-default {
+    position: absolute;
+    margin-left: 122px;
+  }
+
+  .icon-rating-default:hover,
+  .icon-rating-default:hover ~ .icon-rating-default,
```
k88hudson: nice!

k88hudson: ohhh i see - the transforms are to allow for this to work

content-src/components/Rating/Rating.js
k88hudson: make `numStars` a prop and instead of using `activeRatings` just check if your index is less than `currentRating`

content-src/components/Rating/Rating.scss
k88hudson: Can you add a comment about the purpose of the transforms?

package.json
sarracini: this pref will come in handy for future patches to dynamically set the metadata endpoint based on the source, and also to store it nicely in the metadata db as an easily identifiable source

k88hudson:
R+. Test failure looks unrelated

piatra:
@sarracini After rating an item and refreshing the page the rating system is still there, my rating does not show up. Is that intended? It should either not show up for that item again or show me the rating that I gave it.

sarracini:
@piatra meh since this is just mostly for debugging i don't think it's a super big deal that this happens. adding either a persistent rating for that item is more overhead than its worth

Issue #1046, opened 2016-08-05
Intermittent or Perma failure: ./test/test-PerfMeter.test_PerfMeter_tab_restore: timed out, failure


Standard8:
I'm not sure if this is intermittent or perma failure, seems to be happening frequently in 49+:

```
console.error: activity-streams: 
  fail:
  Timed out (after: START)
.console.error: activity-streams: 
  fail:
  Should not be any unexpected tabs open
console.trace: activity-streams: 
_ecated/unit-test.js 108 fail
_ecated/unit-test.js 346 done/<
_/Promise-backend.js 937 Handler.prototype.process
_/Promise-backend.js 816 this.PromiseWalker.walkerLoop
```

ncloudioj:
Fixed in #1120

PR #1120, opened 2016-08-17
fix(test): #1046 Intermittent failures in PerfMeter

ncloudioj:
There are two failures found in this unit test for Firefox 49+:
* `tabData.tab` is undefined
* the timeout failure when executing the line `yield promiseLoadURI(tab, "about:about")` in `test_PerfMeter_tab_restore`

The first one is due to the race condition where the `onShow` didn't get executed as the promise got resolved in the "onNotify".

The cause of the second failure is a little bit obscure, looks like `gBrowser.addTab(appUrl)` sometimes might take some time to do its job, calling `tab.linkBrowser.loadURI(uri)` before it's done will hit the timeout in this case.

I think the similar issues exist in `test-TabTracker.js` as well, will fix it shortly.

@oyiptong @emtwo @k88hudson can anyone r? it, please?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1120)

### DIFF
--- a/test/test-PerfMeter.js
+++ b/test/test-PerfMeter.js
@@ -1,15 +1,19 @@
-/* globals Services, XPCOMUtils, windowMediator, SessionStore */
+/* globals Services, XPCOMUtils, windowMediator, SessionStore, Task */
 
 "use strict";
 const simplePrefs = require("sdk/simple-prefs");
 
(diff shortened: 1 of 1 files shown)


rlr:
changes look good to me and they fix the tests for me locally on nightly. R+

Issue #1111, opened 2016-08-16
Adding new Services adds new toolbar icon and doesn't seem to add new service to doorhanger


pdehaan:
### Steps to reproduce:
1. In the Share toolbar doorhanger, select "Add Services". It should take you to https://activations.cdn.mozilla.net/en-US/
2. Scroll down to Google+ and click Activate button.
3. Click "Enable Services" in doorhanger.
    <img width="522" alt="menubar_and_google__for_firefox" src="https://cloud.githubusercontent.com/assets/557895/17718618/b0439ab6-63c9-11e6-8c84-95b8f4a48e75.png">

### Actual results:
I see a page saying:

> Congratulations, Google+ is now activated in Firefox!

Despite that, I don't see "Google+" added to my share panel:

<img width="207" alt="screenshot_8_16_16__3_55_pm" src="https://cloud.githubusercontent.com/assets/557895/17718652/db0a14d2-63c9-11e6-80b9-286e99c5a498.png">

... But I do see yet another toolbar icon added with the same share services:

<img width="395" alt="google__for_firefox" src="https://cloud.githubusercontent.com/assets/557895/17718682/0e72295e-63ca-11e6-8452-cb101066aa3c.png">

---

I can also see Google+ added to my Services settings page:

<img width="1280" alt="add-ons_manager" src="https://cloud.githubusercontent.com/assets/557895/17718713/3a865c68-63ca-11e6-9625-bf2a3f6f41f3.png">

rlr:
yeah, same underlying issue as #1099. PR coming in 3...2.....

rlr:
Half of this is fixed now. Now need to figure out how to hide that reappearing shareplane.

rlr:
Actually, scratch that... Adding/enabling still seems to be busted. As in they don't appear in the doorhanger until a disable/remove happens. weird.

ncloudioj:
I've tried to add a new service in my AS. Looks like the new service got merged into the share toolbar after a few moments. And the shareplane was gone eventually.

ncloudioj:
@rlr FYI, it works, amigo. I just successfully shared some links to my Chinese social networks 🎱

rlr:
fixed via #1122

PR #1122, opened 2016-08-17
fix (addon): #1111 don't add shareplane to toolbar on activation

rlr:
@mixedpuppy r? and if you have any tips on how to trigger what the activations page does in a test, I'll add one. I tried but couldn't trigger the scenario where the shareplane gets added to the toolbar.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1122)

### DIFF
--- a/lib/ShareProvider.js
+++ b/lib/ShareProvider.js
@@ -135,6 +135,10 @@ const Overlay = {
       if (win.SocialShare.shareButton) {
         win.SocialShare.shareButton.removeAttribute("hidden");
       }
+
+      // unmonkeypatch SocialActivationListener.receiveMessage
+      win.SocialActivationListener.receiveMessage = win.SocialActivationListener.originalReceiveMessage;
+      delete win.SocialActivationListener.originalReceiveMessage;
     }
   },
   observe: window => {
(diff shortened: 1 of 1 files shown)


lib/ShareProvider.js
```diff
@@ -146,6 +150,44 @@ const Overlay = {
...
+        SocialSidebar
+      } = window;
+      // if the source if the message is the share panel, we do a one-click
+      // installation. The source of activations is controlled by the
+      // social.directories preference
+      let options;
+      if (browser === SocialShare.iframe && Services.prefs.getBoolPref("social.share.activationPanelEnabled")) {
+        options = {bypassContentCheck: true, bypassInstallPanel: true};
+      }
+
```
mixedpuppy: The above code [170-176] is not necessary here since you're disabling the share panel.

lib/ShareProvider.js
```diff
@@ -146,6 +150,44 @@ const Overlay = {
...
+      Social.installProvider(data, manifest => {
+        Social.activateFromOrigin(manifest.origin, provider => {
+          if (provider.sidebarURL) {
+            SocialSidebar.show(provider.origin);
+          }
```
mixedpuppy: remove the sidebar code, that is removed once my other patches land.

lib/ShareProvider.js
```diff
@@ -135,6 +135,10 @@ const Overlay = {
       if (win.SocialShare.shareButton) {
         win.SocialShare.shareButton.removeAttribute("hidden");
       }
+
+      // unmonkeypatch SocialActivationListener.receiveMessage
+      win.SocialActivationListener.receiveMessage = win.SocialActivationListener.originalReceiveMessage;
+      delete win.SocialActivationListener.originalReceiveMessage;
     }
```
mixedpuppy: not fond of monkey patching, but ok.

lib/ShareProvider.js
```diff
@@ -146,6 +150,32 @@ const Overlay = {
...
+        Social,
+        SocialShare
+      } = window;
+
+      Social.installProvider(data, manifest => {
+        Social.activateFromOrigin(manifest.origin, provider => {
+          if (provider.postActivationURL) {
+            // if activated from an open share panel, we load the landing page in
+            // a background tab
+            gBrowser.loadOneTab(provider.postActivationURL, {inBackground: SocialShare.panel.state === "open"});
```
mixedpuppy: drop the SocialShare.panel check (just use true or false as you prefer) then r+ from me.

rlr:
landed 589c77a76b1d49e30aae57f64ee05b34a8dd1484

PR #1124, opened 2016-08-18
New highlights query

piatra:
The purpose of this pull requests is to move away from the getHighlightsLinks, to be able to make more significant changes to the query without affecting the highlights.
It is supposed to get all links in the past 4 days or the last 20 items you visited.
It also fixes a bug where weighted highlights were not shown in the Timeline view / See all activity.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1124)


piatra:
@ncloudioj || @k88hudson please review. This pull request has higher priority than all the others currently open by me.
[This line](https://github.com/piatra/activity-stream/blob/8b4a1fdd92686549abd551d7da62a0d33956ad6f/addon/PlacesProvider.js#L741) is currently up for debate. My thinking is that getting all results over the past 4 days and sorting them should be fine as long as I only send only 20 or so to content side. But I'm not the one who worked on performance so your judgement is probably best here. I should note that getting more URLs is an important step.
Right now all links are sent over to content because it helps judge the quality of the sorting/scoring but we can change it just before we land this. [Similar thing happening here](https://github.com/piatra/activity-stream/blob/8b4a1fdd92686549abd551d7da62a0d33956ad6f/common/constants.js#L17).

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
   constructor(history, options = {}) {
     this.domainCounts = history.reduce(this.countDomainOccurrences, new Map());
     this.options = options;
+    // Features that are extracted from URLs and need normalization.
+    // Key 0 holds the min, key 1 holds max, using arrays for brevity.
+    this.normalizeFeatures = {
+      idf: [1, 0],
```
k88hudson: I think `idf: {min: 1, max: 0}` and normalizeFeatures[key].min`is a lot more readable, I would just go with an object

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
...
+      queryLength: [1, 0]
+    };
   }
 
-  scoreEntry(entry) {
+  extractFeatures(entry) {
     const urlObj = URL(entry.url);
-    const host = URL(entry.url).host;
+    const host = urlObj.host;
+    const occurences = this.domainCounts.get(host) || 1;
```
k88hudson: two rs in occurrence

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
...
+    this.updateFeatureMinMax(features);
+
+    return Object.assign({}, entry, {features, host});
+  }
+
+  normalize(features) {
+    let delta;
+    const normalizeKeys = Object.keys(this.normalizeFeatures);
+
+    normalizeKeys.forEach(key => {
```
k88hudson: this would also be a lot more readable with max/min instead of array indices, i.e.
```js
  normalize(features) {
    Object.keys(this.normalizeFeatures).forEach(key => {
      const {min, max} = this.normalizeFeatures[key];
      if (max !== min) { // No division by 0.
        let delta = max - min;
        features[key] = (features[key] - max) / delta;
      }
    });
  }
```

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
...
+    const largestImage = this.extractLargestImage(entry.images);
+    const queryLength = urlObj.query.length;
+
+    const features = {age, tf, idf, imageCount, isBookmarked, description, pathLength, queryLength, largestImage};
+    this.updateFeatureMinMax(features);
+
+    return Object.assign({}, entry, {features, host});
+  }
+
+  normalize(features) {
```
k88hudson: can you put a quick comment here

common/recommender/Baseline.js
k88hudson: also if you can, it's better to return a new object rather than cause side effects on the one you input

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
...
+  }
+
+  scoreEntry(entry) {
+    this.normalize(entry.features);
+
+    const {age, tf, idf, queryLength, imageCount, pathLength, description, largestImage} = entry.features;
+    let score = this.decay(tf * idf, // Score
+      // Features: Age in hours, number of visits to url, how often you go to the domain, number of images,
+      //           length of path, length of description, size of biggest image.
+      [age, tf, idf, queryLength, imageCount, pathLength, description, largestImage],
```
k88hudson: Does the order of these matter? If not, perhaps you could store `highlightsCoefficients` as an object and then just use  `keys.map(key => entry.features[key])`, `keys.map(key => highlightsCoefficients[key])`

k88hudson: If order does matter, you could also use a `Map`

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
...
+      }
+    });
+  }
+
+  scoreEntry(entry) {
+    this.normalize(entry.features);
+
+    const {age, tf, idf, queryLength, imageCount, pathLength, description, largestImage} = entry.features;
+    let score = this.decay(tf * idf, // Score
+      // Features: Age in hours, number of visits to url, how often you go to the domain, number of images,
```
k88hudson: This comment might be more useful up top with the `highlightsCoefficients`

common/recommender/Baseline.js
```diff
@@ -15,27 +15,134 @@ class Baseline {
...
+
+  /**
+   * Update min and max values of the features that require normalization.
+   *
+   * @param {Object} newFeatures - features and associated values.
+   */
+  updateFeatureMinMax(newFeatures) {
+    const keys = Object.keys(this.normalizeFeatures);
+    keys.forEach(key => {
+      this.normalizeFeatures[key][0] = this.selectMinValue(this.normalizeFeatures[key][0], newFeatures[key]);
```
k88hudson: again, `max` and `min` keys would make this a lot more readable

k88hudson:
This is looking great, my main concern here is making the "features" a little more flexible/readable; right now, changing the default coefficients/adding/removing any features requires code changes in a lot of different places

piatra:
@k88hudson I've made some changes let me know what you think.
* [dropped the coefficient values](https://github.com/piatra/activity-stream/commit/2c388630c1848423240aa00ebfb550a2ff936b06#diff-4fd4e9cc2b26862198efdd2354d207d9L5) only read them from __one place__ the package.json, if not provided throw an error.
* [instead of specifying parameters in specific order](https://github.com/piatra/activity-stream/commit/2c388630c1848423240aa00ebfb550a2ff936b06#diff-4fd4e9cc2b26862198efdd2354d207d9R73) send an object with all of them.
* the price to pay: [weights have to be specified in alphabetical order](https://github.com/piatra/activity-stream/commit/2c388630c1848423240aa00ebfb550a2ff936b06#diff-4fd4e9cc2b26862198efdd2354d207d9R223) to ensure they are matched with the correct coefficient value
* I agree with the refactoring of `normalize` using min/max instead of array keys *WIP*

You mentioned using a `Map` I don't know how/if that would prevent having a specific order for feature names. The problem occurs when you have to match weights to features. One idea: we could store serialized objects as preferences. Then we would have an object with keys as feature names and values as the current coefficients.
```
pref.weightedHighlights: {
  // pref : weight
  featureA: 0.5,
  featureB: 0.2
}

// when computing score
featureNames = getPrefKeys();
// score = feature value for a given url * preference weight for same feature
score = entry.features[featureName] * prefs.weightedHighlights[featureName]

```

piatra:
@k88hudson I am going ahead with this solution.

piatra:
I think https://github.com/mozilla/activity-stream/pull/1124/commits/d42d798baa9ce8a5f83164b595214a16db7e6ebb is starting to look p good.

common/recommender/Baseline.js
```diff
@@ -15,27 +14,141 @@ class Baseline {
   constructor(history, options = {}) {
     this.domainCounts = history.reduce(this.countDomainOccurrences, new Map());
     this.options = options;
+    // Features that are extracted from URLs and need normalization.
+    // Key 0 holds the min, key 1 holds max, using arrays for brevity.
+    this.normalizeFeatures = {
+      idf: {min: 1, max: 0},
```
k88hudson: much better 👍

common/recommender/Baseline.js
```diff
@@ -51,20 +54,20 @@ class Baseline {
...
+  // Adjust all values in [0, 1].
   normalize(features) {
-    let delta;
-    const normalizeKeys = Object.keys(this.normalizeFeatures);
-
-    normalizeKeys.forEach(key => {
-      if (this.normalizeFeatures[key][1] !== this.normalizeFeatures[key][0]) { // No division by 0.
-        delta = this.normalizeFeatures[key][1] - this.normalizeFeatures[key][0];
-        features[key] = (features[key] - this.normalizeFeatures[key][0]) / delta;
+    return Object.keys(this.normalizeFeatures).reduce((acc, key) => {
```
k88hudson: 👍

k88hudson:
I agree having a serialized pref is not the best experience ever, but the alternative is to not have a record of what each thing actually is; we could add some UI to make this more painful to the debug menu later. A similar approach needs to be taken for the experiment overrides

k88hudson:
You've got a couple of lint errors (and a typo) but this is looking great! Refactor is +++

piatra:
@k88hudson  😱 where is the typo I can't find it!

common/recommender/Baseline.js
k88hudson: typo here ^

pdehaan: (and same typo three other places in this PR, if I ctrl+F for "occurence" on this page.)

piatra:
@k88hudson This is ready for a new review. 

Right now the query fetches all the links in the past 4 days, processes all (with some logic to filter out those that are < 30 minutes if we have enough links to show) and only return 20 to content space. We can discuss any performance issues if you have concerns, personally I did not feel any change with just 20 links on content side and a reasonable amount of history ~250 links to process. 

`dev-prefs` has weightedHighlights set to `true` on purpose. We all need to have them on at least when we are testing so that we may catch any problems.

Only potential change that may occur is with weights.

k88hudson:
@piatra Ok, I think this is all good to go except I ran into one more problem: I tried this with a profile in which everything was >4 days old, and I got 3 blank highlights show up

k88hudson:
Squash and land it! R+ 👍 👍 👍 👍 👍

tspurway:
🎉‼️😀

tspurway:
🎉‼️😀

PR #1163, opened 2016-08-23
feat(metadata): #1159 add migration support for metadata store

### LINKED ISSUE #1159: Improve the migration support for metadata store
The migration for metadata store should be implemented as follows:

* add a migration table ("metastore_migration") to track the current schema version
* add a js module to store the whole history of migrations
* check whether the migration is needed when the user connects to the metadata store

ncloudioj:
This fixes #1159

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1163)

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -8,6 +8,8 @@ Cu.import("resource://gre/modules/Sqlite.jsm");
 Cu.import("resource://gre/modules/NetUtil.jsm");
 Cu.import("resource://gre/modules/Timer.jsm");
 
+const {MIGRATIONS} = require("addon/MetadataStoreMigration.js");
+
 XPCOMUtils.defineLazyModuleGetter(this, "Task",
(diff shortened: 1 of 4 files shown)


ncloudioj:
r? @oyiptong 

cc @sarracini PR is out, you can take a quick look at the `MetadataStoreMigration.js`, that's the file in which all the migrations will be added.

addon/MetadataStore.js
```diff
@@ -17,6 +19,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "OS",
 
 const METASTORE_NAME = "metadata.sqlite";
 
+const SQL_MIGRATIONS = migrations;
```
oyiptong: why not use the `migrations` const declared above directly?

ncloudioj: Just wanted to keep the convention as the uppercase. I think I can rename the `migration` directly.

Issue #1165, opened 2016-08-25
Metadata store reset is broken with migration


ncloudioj:
Found this issue while @sarracini made the first migration for #gh1156

We should drop the "migrations" table while resetting these tables, otherwise, the migration won't be committed correctly.

sarracini:
Fixed with #1166

PR #1166, opened 2016-08-25
fix(metadata-store): #1165 - Drop the migrations table in reset

ncloudioj:
@sarracini r?

Turns out it wasn't `sqlite.jsm` fault, it's mine. I should take the "credit" ;)

All the tests in your branch should pass now with this patch.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1166)


sarracini:
Oooooh man. R+ thanks for figuring this out @ncloudioj

ncloudioj:
Thank you for revealing this nasty bug! @sarracini 
Turns out the GO train is a very good place to tackle these sneaky issues :)

PR #1183, opened 2016-08-26
Fix debug page results for weighted highlights

piatra:
Debug page is showing incorrect results when selecting `WeightedHighlights` due to selectors being given incorrect state.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1183)


pdehaan:
@piatra I think this needs a rebasin'.

PR #1199, opened 2016-08-31
feat(content): Add basic token matching for Timeline all/bookmark views with search in header.

### LINKED ISSUE #416: Add filter to Timeline for filtering/searching history and bookmarks
This ticket should add a filter/search box that allows searching the Timeline.

This will require:
- [ ] add-on APIs for searching history/bookmarks, as well as new data structures in the store
- [ ] new search input UI
- [ ] new search results display UI

Might be worth splitting this up into different tickets.

Trello spec here: https://trello.com/c/UMGCuxJy

*[edit by @k88hudson for clarity]*

Mardak:
Fix #416. This approach does the filtering in content after passing the query string to a new notify action that gets selected by the timeline page views. Tests are updated to handle the now-connected Header in addition to new tests covering dispatching notify action, passing filter query from `TimelinePage`, and filtering in `GroupedActivityFeed` component. Styles are mostly reusing the input from the `Search` component, so refactored accordingly.

Todo: file followup performance bugs, ui tweaks

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1199)

### DIFF
--- a/content-src/components/Header/Header.js
+++ b/content-src/components/Header/Header.js
@@ -1,4 +1,7 @@
 const React = require("react");
+const {connect} = require("react-redux");
+const {actions} = require("common/action-manager");
+const {justDispatch} = require("selectors/selectors");
 const {Link} = require("react-router");
 const classNames = require("classnames");
 
@@ -32,7 +35,13 @@ const Header = React.createClass({
           {props.links.map(link => <li key={link.to}><Link to={link.to}>{link.title}</Link></li>)}
         </ul>
       </section>
-      <section className="spacer" />
+      <section className="filter">
+        <input
+          onChange={e => props.dispatch(actions.NotifyFilterQuery(e.target.value))}
+          placeholder="Search your Activity Stream"
+          ref="filter"
+          type="search" />
+      </section>
       <section className="user-info">
         {props.userName && <span>
           {props.userName}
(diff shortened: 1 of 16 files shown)


content-src/components/Header/Header.js
```diff
@@ -33,6 +36,13 @@ const Header = React.createClass({
         </ul>
       </section>
       <section className="spacer" />
+      <section className="filter">
+        <input
+          onChange={e => props.dispatch(actions.NotifyFilterQuery(e.target.value))}
```
piatra: Don't we want some `debounce` logic here?

piatra: And I think we want `dispatch` as a required prop since we always use it.

sarracini: @piatra what do you mean by debounce?
and also nah i don't think you need to declare dispatch as a required prop

content-src/components/Header/Header.js
```diff
@@ -33,6 +36,13 @@ const Header = React.createClass({
...
+          placeholder="Filter your activity stream"
```
sarracini: can we change this to say "Search your activity stream"?

Mardak: Changed to Search and capitalized Activity Stream. Also fixed the sizing/positioning as the longer placeholder text got cut off.

sarracini:
looks good r+!

Mardak:
Filed followup issues #1260 and #1261.

PR #1209, opened 2016-09-02
feat(addon): add benchmarking support

ncloudioj:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1209)

### DIFF
--- a/test/lib/Benchmark.js
+++ b/test/lib/Benchmark.js
@@ -1,0 +1,291 @@
+/* globals XPCOMUtils, Services, Task */
+"use strict";
+
+const {Cu} = require("chrome");
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "Task",
+                                  "resource://gre/modules/Task.jsm");
+
+// The default benchmark time is 1 second, i.e. 1e6 microseconds
+const DEFAULT_BENCH_TIME = 1e6;
+
+// For the performance.now(), so the timer resolution is 1 microsecond
+const performance = Services.appShell.hiddenDOMWindow.performance;
+
+// The max iteration number is 1e6 as the timer resolution is 1 microsecond
+const MAX_ITERATION = 1e6;
+
+/*
+ * Get the DOMHighResTimeStamp in microseconds
+ *
+ * Note that the actual resolution is subject to the implementation and platforms
+ *
+ * See more detail at
+ *   https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
+ *   https://www.w3.org/TR/2013/WD-hr-time-2-20131203/#sec-DOMHighResTimeStamp
+ */
+function _performanceNowInUs() {
+  return performance.now() * 1e3;
+}
+
+/*
+ * A simple benchmark framework for Activity Stream that runs on top of jpm test
+ *
+ * If you're new to Javascript benchmarking, here is a list of references:
+ *
+ *   https://developer.mozilla.org/en-US/docs/Mozilla/Benchmarking
+ *   https://mathiasbynens.be/notes/javascript-benchmarking
+ *   http://ejohn.org/blog/javascript-benchmark-quality/
+ *
+ * Suitable use cases:
+ *   * functions without or with little side effect, i.e. IO/file system/database dependency
+ *
+ * Unsuitable use cases:
+ *   * benchmark functions running at the nanoseconds level, it doesn't support nanosecond resolution
+ *   * measure the functions or code paths with side effects. Although, you may mitigate that by
+ *     leveraging the timer functions. But the results could be incorrect and hard to interpret
+ *
+ * Benchmarking Activity Stream requires extra effort in following cases,
+ *   * measure the load latency of Activity Stream with different profiles and browsering history
+ *   * measure the time-consuming functions or code paths found by the profiler. In particular,
+ *     the performance difference after the improvement or optimization
+ * Since Activity Stream is IO intensive, and makes extensive use of cache, the benchmark could be
+ * biased by those confounding variables. It's highly recommended to take proper actions inside benchmark
+ * function. For instance, cleaning up cache to benchmark the cold start. The benchmark timer functions
+ * can be used to prevent benchmark from timing those unrelated operations
+ *
+ * Example:
+
+  function fib(n) {
+    if (n < 2) {
+      return n;
+    }
+
+    return fib(n - 1) + fib(n - 2);
+  }
+
+  exports.test_benchmark = function*() {
+    let bench = new Benchmark(b => {
+      for (let i = 0; i < b.N; i++) {
+        fib(20);
+      }
+    });
+    let result = yield bench.launch();
+    console.log(result);
+  };
+*/
+
+/*
+ * Constructor of benchmark
+ *
+ * {Params} func is the benchmark function/generator that is of the form
+ *
+   function benchFunc(b) {
+     for (let i = 0; i < b.N; i++) {
+       // do your stuff here
+     }
+   }
+
+ * Where the argument b is the current benchmark instance. A couple of timer functions
+ * are availale to get the finer-grained control of benchmark timing. For instance,
+ * avoid timing the expensive irrelevant functions, like setup of benchmark fixture.
+ * See more detail in startTimer(), resetTimer(), and stopTimer()
+ *
+ * It's intentional to have a for loop inside the benchmark function, otherwise the framework
+ * has to call the benchmark function for b.N times, which might introduce some overheads
+ * if b.N is a big integer
+ *
+ * Also note that the framework will choose a good iteration number b.N based on how fast
+ * the target benchmark function runs. Therefore, the user should never mutate it
+ *
+ * {Params} benchTime is the benchmark time in microseconds, by default it's one second
+ */
+function Benchmark(func, benchTime = DEFAULT_BENCH_TIME) {
+  this._failed = false;
+  this._benchFunc = func;
+  this._benchTime = benchTime;
+  this._timerOn = false;
+  this._start = 0;
+  this._duration = 0;
+  // N is public to the benchmark functions
+  this.N = 0;
+}
+
+Benchmark.prototype = {
+  /*
+   * Starts the timer for the benchmark. This function is called automatically
+   * before each benchmark. It also can be used to resume timing after the
+   * stopTimer function call
+   */
+  startTimer() {
+    if (!this._timerOn) {
+      this._start = _performanceNowInUs();
+      this._timerOn = true;
+    }
+  },
+
+  /*
+   * Resets the timer if it is in use. It also zeros the benchmark's duration
+   */
+  resetTimer() {
+    if (this._timerOn) {
+      this._start = _performanceNowInUs();
+    }
+    this._duration = 0;
+  },
+
+  /*
+   * Stops the timer if it is in use. It also calculates the benchmark's duration
+   */
+  stopTimer() {
+    if (this._timerOn) {
+      this._duration += (_performanceNowInUs() - this._start);
+      this._timerOn = false;
+    }
+  },
+
+  /*
+   * Calculates the duration per operation in microseconds
+   */
+  _usPerOp() {
+    if (!this.N) {
+      return 0;
+    }
+    return Math.round(this._duration / this.N);
+  },
+
+  /*
+   * Rounds the given integer up to be one of [1eX, 2eX, 3eX, 5eX, 10eX]
+   *
+   * We wanted to apply this adjustment to the iteration numbers to:
+   * 1. be more human readable
+   * 2. mimic the traditional benchmark approach, where the benchmark designer
+   *    specifies some frequently used iteration numbers beforehand
+   */
+  _roundsUp(n) {
+    let nn = n;
+    let tens = 0;
+    while (nn >= 10) {
+      nn /= 10;
+      tens++;
+    }
+
+    let base = 1;
+    for (let i = 0; i < tens; i++) {
+      base *= 10;
+    }
+
+    if (n <= base) {
+      return base;
+    } else if (n <= 2 * base) {
+      return 2 * base;
+    } else if (n <= 3 * base) {
+      return 3 * base;
+    } else if (n <= 5 * base) {
+      return 5 * base;
+    }
+    return 10 * base;
+  },
+
+  /*
+   * Runs the benchmark once, this function is used to get the first estimate of
+   * the speed of the benchmark
+   *
+   * Note that this funciton will catch the exceptions in the benchmark function,
+   * and will set the flag upon those exceptions accrodingly
+   */
+  _asyncRun1: Task.async(function*() {
+    try {
+      yield this._asyncRunN(1);
+    } catch (e) {
+      this._failed = true;
+    }
+  }),
+
+  /*
+   * Runs the benchmark for a given times with timer
+   *
+   * It will not handle any exceptions thrown from the benchmark function
+   */
+  _asyncRunN: Task.async(function*(n) {
+    this.N = n;
+    this.resetTimer();
+    this.startTimer();
+    yield this._benchFunc(this);
+    this.stopTimer();
+  }),
+
+  /*
+   * Choose a good iteration number for the benchmark
+   *
+   * {Params} hz is the current benchmark frequency
+   * {Params} last is the last iteration number
+   *
+   * It returns a integer that is considered as a good iteration number
+   */
+  _chooseN(hz, last) {
+    let n = hz;
+
+    // It runs more (1.2x) than the estimated frequency
+    const amplified = Math.round(n * 1.2);
+
+    // Do not grow too fast in case of the incorrect previous timing,
+    // so it'll be capped by 100x of last iteration
+    const capped = 100 * last;
+    n = Math.min(amplified, capped);
+
+    // Make sure it at least runs one more than last iteration
+    n = Math.max(n, last + 1);
+
+    // Rounds up the estimate to make it easy to read
+    return this._roundsUp(n);
+  },
+
+  /*
+   * Launches and times the benchmark
+   *
+   * To begin with, it executes the benchmark function once to figure out its
+   * speed. Then it attempts to predict a reasonable iteration number given
+   * the pre-defined benchmark time and estimated speed. In case of the incorrect
+   * estimations, it will retry the benchmark with a biggr loop until the actual
+   * benchmark duration reaches the expected duration
+   *
+   * This function returns an object with the benchmark status, the iteration number,
+   * the actual duration, and the duration per operation in microseconds. The status
+   * will be set as failed if any exception is thrown during the benchmark, and the
+   * benchmark will abort immediately
+   */
+  launch: Task.async(function*() {
+    let n = 1;
+
+    yield this._asyncRun1();
+    while (!this._failed && this._duration < this._benchTime && n < MAX_ITERATION) {
+      let last = n;
+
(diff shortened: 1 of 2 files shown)


piatra:
nit: The microseconds project added as a dependency has 0 tests and we are using this for our benchmark tests. A bit meta & also seems wrong to me.

https://github.com/kamicane/microseconds

test/lib/Benchmark.js
```diff
@@ -0,0 +1,181 @@
...
+  this._benchFunc = func;
+  this._benchTime = benchTime;
+  this._timerOn = false;
+  this._start = 0;
+  this._duration = 0;
+  // N is public to the benchmark functions
+  this.N = 0;
+}
+
+Benchmark.prototype = {
```
piatra: I'm seeing these warnings in the browser regularly 
```
[JavaScript Warning: "mutating the [[Prototype]] of an object will cause your code to run very slowly;
```
Why not use `classes`?

dmose: The style of assigning to .prototype in this patch is used all over the mozilla codebase; is there evidence to suggest that this code is triggering the warning?

http://stackoverflow.com/questions/23807805/why-is-mutating-the-prototype-of-an-object-bad-for-performance seems to suggest that the warning is talking about something else.

dmose:
As far as the microseconds dependency, it looks to me like the only real value we're getting from it is https://github.com/kamicane/microseconds/blob/master/now.js#L16-L19, and we don't even want the Date.now fallback, so I'd be a fan of letting that go and just testing/writing a function that does that.

test/lib/Benchmark.js
ncloudioj: @piatra I have asked the similar question before. Not sure if there is any difference between those two.

As @dmose mentioned above, I stuck to this one because it's ubiquitous in our codebase.

dmose:
One of the exciting things about performance.now is that "SHOULD" have 5us resolution, but it apparently is allowed to fall back to 1ms resolution, and there doesn't appear to be an API to tell the caller what the actual resolution is.  

https://www.w3.org/TR/2013/WD-hr-time-2-20131203/#sec-DOMHighResTimeStamp

has the fun details.

It's probably worth finding out what resolution Firefox provides on what systems, and if we're lucky, it'll be 5 microseconds everywhere, and commenting on that somewhere in the code.

Perhaps it's worth echoing something to the console about this (in Benchmark.js?), since it effects interpretation of the benchmarks, possibly very substantially, if we don't have microseconds everywhere.

ncloudioj:
@piatra Thanks for pointing out the problematic 3rd party dependent library.

Looks like we can leverage the `Window.performance` to get the job done. Although, as the benchmark works inside the addon, we have to use the hidden window to access it.

dmose:
Doing this review forced me to learn a bunch about JS benchmarking gotchas, which was very interesting.  For the benefit of future maintainers of this code, I'd suggest putting these links in a comment at the top of the file to make it easier for first-time readers to get their bearings:

https://mathiasbynens.be/notes/javascript-benchmarking
http://ejohn.org/blog/javascript-benchmark-quality/

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "Task",
+                                  "resource://gre/modules/Task.jsm");
+// For the performance.now()
+const performance = Services.appShell.hiddenDOMWindow.performance;
+
+/*
+ * A simple benchmark framework that runs on top of jpm test
```
dmose: I'd suggest writing here a list of what this framework is and isn't suitable for.  In particular, at least as it's currently written, it can't be used by itself to test "first-time" or "startup" costs (eg performance of the first time new tab is shown in a browser instance).

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+*/
+
+/*
+ * {Params} func is the benchmark function/generator that is of the following form
+ *
+   function(b) {
+     for (let i = 0; i < b.N; i++) {
+       // do your stuff here
+     }
+   }
```
dmose: Out of curiousity, why does the API make the caller do the looping?  It's probably worth at least a comment...

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+ * Note that the framework will decide the best iteration number b.N based on
+ * how fast the target benchmark function runs
```
dmose: I think "will choose a good number of iterations, b.N,"... would be clearer.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+
+Benchmark.prototype = {
+  /*
+   * Starts the timer for the benchmark. This function is called automatically
+   * before each benchmark. It also can be used to resuming timing after the
+   * stopTimer function call
+   */
+  startTimer() {
+    if (!this._timerOn) {
+      this._start = performance.now() * 1e3;
```
dmose: Maybe use a constant for readability?

Eg

const MS_TO_US_MULTIPLIER = 1e3;

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+  stopTimer() {
+    if (this._timerOn) {
+      this._duration += (performance.now() * 1e3 - this._start);
+      this._timerOn = false;
+    }
+  },
+
+  _usPerOp() {
+    if (this.N <= 0) {
+      return 0;
```
dmose: Is it valid for this.N to go negative?  If so, please comment under what circumstances that would happen.  If not, perhaps we want to throw in that case?

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+    return Math.round(this._duration / this.N);
+  },
+
+  /*
+   * Rounds the given integer up to a number of the form [1eX, 2eX, 3eX, 5eX, 1e(X+1)]
+   */
```
dmose: It would be good to have the comments here include a description of why this is useful.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+    this.startTimer();
+    yield this._benchFunc(this);
+    this.stopTimer();
+  }),
+
+  launch: Task.async(function*() {
+    let n = 1;
+
+    yield this._asyncRun1();
+    while (!this._failed && this._duration < this._benchTime && n < 1e6) {
```
dmose: 1e6 wants to be a constant for readability.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+      let last = n;
+      if (!this._usPerOp()) {
+        n = 1e6;
+      } else {
+        n = Math.round(this._benchTime / this._usPerOp());
+      }
```
dmose: The above segment might want to be it's own function, but at the very least, it should have a comment about what it's doing at a high-level.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,182 @@
...
+      // Run more (1.2x) than we predict, don't grow too fast in case of
+      // the incorrect previous timing. At least run 1 more than last time
+      n = Math.max(Math.min(Math.round(n + n / 5), 100 * last), last + 1);
```
dmose: I think each one of the function calls in the above line wants to be its own statement with a comment about why it's doing what it's doing so that future maintainers can easily decode this.

test/lib/Benchmark.js
dmose: Out of curiousity what are the specific use cases here, besides measuring (eg) tab-open times?

test/test-Benchmark.js
```diff
@@ -0,0 +1,128 @@
...
+XPCOMUtils.defineLazyModuleGetter(this, "Task",
+                                  "resource://gre/modules/Task.jsm");
+
+const {Benchmark} = require("./lib/Benchmark.js");
+const {getTestActivityStream} = require("./lib/utils");
+
+let app;
+let ACTIVITY_STREAMS_URL;
+
+const roundsUpTest = [
```
dmose: Nice concise unit test here!

test/test-Benchmark.js
```diff
@@ -0,0 +1,128 @@
...
+  });
+  const result = yield bench.launch();
+  assert.ok(result.ok, "it should run the benchmark and report the status");
+  assert.equal(bench._usPerOp(), result.usPerOp);
+};
+
+before(exports, function*() {
+  // initialize the app now
+  let clientID = yield ClientID.getClientID();
+  simplePrefs.prefs.telemetry = true;
```
dmose: I'm assuming this is because we want to exercise the same code-paths as we would in production?  Probably worth a comment...

dmose: Also, I presume that although the telemetry code is activated, we're not actually sending this telemetry data to a real server?

test/test-Benchmark.js
```diff
@@ -0,0 +1,128 @@
...
+
+  app = getTestActivityStream({
```
dmose: I imagine using a testActivityStream is the most appropriate thing to do, but it may be worth opening another ticket to investigate how different the mocks this uses are from typical performance in the wild.

dmose:
To get real numbers when testing with nightly, we're going to need to explicitly tweak the async stacks and JS GC poisoning stuff somewhere.

dmose:
https://developer.mozilla.org/en-US/docs/Mozilla/Benchmarking documented here

dmose:
It could be that an "npm run perf" script to run jpm perf tests with the appropriate prefs and environment could be the way to go.

dmose:
Nan, thanks for putting this together!  I'd like to have another look at it once it's been tweaked & the questions have been answered...

test/lib/Benchmark.js
ncloudioj: The main reason is that the overhead of executing the benchmark function for b.N times might affect the benchmark's accuracy if b.N is a large integer. 
```js
b.startTimer();
for (let i = 0; i < b.N; i++) {
  b.benchFunc();
}
b.stopTimer();
```
whereas if the looping gets wrapped into the benchmark function
```js
b.startTimer();
b.benchFunc(); // only call the benchmark function once 
b.stopTimer();
```

test/lib/Benchmark.js
ncloudioj: Good point. The purpose of doing this is to make the b.N more human readable, other than reporting some random integer, e.g. `b.N = 12,312,145`. Will add a comment for this.

test/test-Benchmark.js
ncloudioj: We have a dev-pref for the test env, so it won't send any telemetry to the prod server.

ncloudioj:
@dmose Thanks for the detailed inputs and feedback. Will fix and tweak a bit accordingly.

test/lib/Benchmark.js
ncloudioj: The main use cases we want to benchmark:
* the load latency of A-S with various profile sizes (e.g. history/bookmarks)
* the expensive utility functions or code blocks found by the profiler (e.g. color-analyzer, tippy-top-site, sqlite-queries...)
* new features and its influence on the addon performance

test/lib/Benchmark.js
dmose: Seems like it could simplify the code to use b.N.toExponential(0).  Is there some reason not to do that?

ncloudioj: We wanted to roundup the iteration number to [1eX, 2eX, 3eX, 5eX, 10eX], instead of letting it completely go wild. This somewhat mimics the traditional benchmark approach, where the benchmark designer specifies some frequently used iteration numbers beforehand.

test/lib/Benchmark.js
dmose: Ah, makes sense.  Definitely put something like that in as a comment so that future maintainers don't naively change that invariant.

ncloudioj:
@dmose I've added some comments as well as some code refactoring. One more run plz?

Also, I haven't yet added the "npm run perf", and the "Poisoning" stuff in this PR. I think they both deserve a separate one to make it right. Will file issues to track them right away. Make sense?

test/lib/Benchmark.js
```diff
@@ -0,0 +1,259 @@
...
+// For the performance.now(), so the timer resolution is 1 microsecond
+const performance = Services.appShell.hiddenDOMWindow.performance;
+
+// The max iteration number is 1e6 as the timer resolution is 1 microsecond
+const MAX_ITERATION = 1e6;
+
+/*
+ * Get the DOMHighResTimeStamp in microsecond
+ * See more detail at https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
+ */
```
dmose: Maybe in microseconds.  I think this comment needs to at least mention some of the bits/links from https://github.com/mozilla/activity-stream/pull/1209#issuecomment-244446467, and we probably still want to address that, even if it's by opening another ticket to do that work.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,259 @@
...
+ * A simple benchmark framework for Activity Stream that runs on top of jpm test
+ *
+ * If you're new to Javascript benchmarking, here is a list of references:
+ *
+ *   https://developer.mozilla.org/en-US/docs/Mozilla/Benchmarking
+ *   https://mathiasbynens.be/notes/javascript-benchmarking
+ *   http://ejohn.org/blog/javascript-benchmark-quality/
+ *
+ * Suitable use cases:
+ *   * measure the load latency of Activity Stream with different profiles and browsering history
```
dmose: I'm not convinced this is really suitable for that, since it involves lots of I/O, and the methodology used here means that all of the OS/filesystem/browser/JIT caches will be pre-primed with the content in question, which seems likely to be fairly unrepresentative of the typical use in the wild.

dmose: That said, I'm not entirely sure what to do for this use case, and am open to suggestions.  It might well be that this is the best we can do without substantially more work (eg finding and flushing all the caches that we can think of that could interfere).  

At the very least, we should document the likely confounding variables with this sort of test.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,259 @@
...
+ *   * measure the time-consuming functions or code paths found by the profiler. In particular,
+ *     the performance difference after the improvement or optimization
+ *   * other functions or features (i.e. CPU intensive or IO bound) that are worth monitoring early
+ *     in order to get the peace of mind
```
dmose: This has the same problem, as I/O is typically cached, which is a side-effect.

test/lib/Benchmark.js
dmose: One thing the comment could suggest would be adding performance telemetry for everything that we benchmark so that we can also see distributions and averages in the wild.

test/lib/Benchmark.js
```diff
@@ -0,0 +1,259 @@
...
+ *
+   function benchFunc(b) {
+     for (let i = 0; i < b.N; i++) {
+       // do your stuff here
+     }
+   }
+
+ * Where the argument b is the current benchmark instance. A couple of timer functions
+ * are availale to get the finer-grained control of benchmark timing. For instance,
+ * avoid timing the expensive irrlevant functions, like setup of benchmark fixture.
```
dmose: s/irrlevant/irrelevant/

test/lib/Benchmark.js
dmose: If you could address https://github.com/mozilla/activity-stream/pull/1209/files#r77722101 , I think that would enhance readability.

ncloudioj: Oops, I missed this one. Sure, will clarify it.

test/lib/Benchmark.js
ncloudioj: I agreed with you that the current methodology won't work in certain cases. It hasn't yet done too much to create a reasonably fair benchmark environment. Like, force a GC before each benchmark, turn off the JIT optimization, or other system level tweaks. I am open to any suggestions on that though. 

My understanding of this benchmark thing is,
* try the best first, having a working one is better than flying in blind, although it could start from being inaccurate in some cases. We can improve it as we understand more in an incremental fashion
* the benchmark designer also could take some actions inside the benchmark function to mitigate the side-effect. i.e. clean up the cache, reset the table in the database etc

Let's have a chat about this to sort things out, Dan. Really appreciate your insights!

test/test-Benchmark.js
```diff
@@ -0,0 +1,129 @@
...
+    for (let i = 0; i < b.N; i++) {
+      fib(20);
+    }
+  });
+  const result = yield bench.launch();
+  assert.ok(result.ok, "it should run the benchmark and report the status");
+  assert.equal(bench._usPerOp(), result.usPerOp);
+};
+
+const openActivityStream = Task.async(function*(openUrl) {
```
dmose: I have a suspicion that this measures all of openActivityStream, as well as some of closeActivityStream.  The reason is that even though the code doesn't wait for the tab to close before starting the next open, the closing will eat some of the cycles on the UI thread (and, in the e10s case, in the content process). 
So the wall clock time measure is likely to in part reflect that the event queue on both of those threads spent time away from the opening task.

 In the (future) case of multiple content processes, this may or may not get diluted more, depending on how the tabs are sorted across content processes.

It might make sense to explicitly do and wait for teardown completion with the timer turned off before starting the next iteration of the test.

Not closing the tabs until the whole test finishes is another option, though various caching strategies (eg XUL "brutal sharing") will kick in that case.

test/lib/Benchmark.js
dmose: That sounds exactly right to me, which is why I think just being sure to add comments about the gotchas is totally sufficient.  Crawl -> walk -> run FTW

test/test-Benchmark.js
dmose: A comment here (even just pasting the above, if you want) as well as opening another issue is probably sufficient for now too.

dmose:
Almost there, thanks for tolerating my over-reviewing.  :-)  I'd like to have one more quick look.

dmose:
(Once you've addressed the current round of review comments, even if mostly by commenting)

ncloudioj:
@dmose heh, no worries. Will tweak it a bit, particularly, clarify some confusing/misleading comments. Then let's do it again.

ncloudioj:
Hey @dmose, just pushed one commit, let's do it again!

test/lib/Benchmark.js
```diff
@@ -208,6 +218,29 @@ Benchmark.prototype = {
...
 
   /*
+   * Choose a good iteration number for the benchmark
+   *
+   * {Params} hz is the current benchmark frequency
+   * {Params} last is the last iteration number
+   *
+   * It returns a integer that is considered as a good iteration number
+   * by the following rules:
+   *   1. it runs more (1.2x) than the estimated frequency
```
dmose: Please put each of these comments next to the line of code that implements it.

test/lib/Benchmark.js
```diff
@@ -208,6 +218,29 @@ Benchmark.prototype = {
...
+   *   2. it shouldn't grow too fast in case of the incorrect previous timing,
+   *      so it'll be capped by 100x of last iteration
+   *   3. it should at least run one more than the last iteration
+   *   4. it rounds up the estimate to make it easy to read
+   */
+  _chooseN(hz, last) {
+    let n = hz;
+
+    n = Math.min(Math.round(n * 1.2), 100 * last);
```
dmose: Which will mean splitting this into two lines, which will improve readability.

dmose:
Looks good, r=dmose with those last readability tweaks.  Thanks for the hard work!

ncloudioj:
@dmose those're now fixed.

Thanks again for doing all this. The code looks much more reasonable now.

PR #1239, opened 2016-09-07
chore(addon): Use sharable strategy for vendor imports in common

k88hudson:
This patch makes a slight change to our vendor dependency importing strategy:

- moved `vendor` to common, instead of addon folder
- changed filename to `vendor.js` so you can just do `require("common/vendor")`
- aliases `common/vendor` to `common/vendor-src` for content, so that files which use `common/vendor` can be used in both content and addon code

I need this for the page worker refactor stuff

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1239)


k88hudson:
@sarracini hey want to review this?

sarracini:
r+ looks good to me

PR #1153, opened 2016-08-23
New design for grouped highlights

piatra:
Replace the `Recent activity` section with more highlights. Design mockups are here: https://mozilla.invisionapp.com/share/TN84I7MB2#/screens/177705068
Size chart is here: http://i.imgur.com/RmRnfXK.png

screen shot 2016-08-22 at 5 00 55 pm

Work in progress. Amusing how my top recommendations are now CSS related. Centering stuff never works as expected!

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1153)

PR #1242, opened 2016-09-07
Fixes #1241 - First time use with weighted highlights

### LINKED ISSUE #1241: First time use issues with weighted highlights
On a fresh profile there are some issues to do the lack of user history.
One of them is that the score is not computed correctly, the other is we have to few highlights to show so we need to keep the first run suggestions a bit longer.

Added to this iteration because a pull request is ready.

Blocks https://github.com/mozilla/activity-stream/pull/1153

piatra:
Default history size to 1 if it doesn't exist.
Default idf value to 1.
Keep first time use cards around until we have at least 6 highlights. This helps https://github.com/mozilla/activity-stream/pull/1153.
Add tests for the changes described above.

Fixes #1241 

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1242)

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -39,7 +39,7 @@ class Baseline {
     const occurrences = this.domainCounts.get(host) || 1;
     const domainCountsSize = this.domainCounts.size || 1;
     const tf = entry.visitCount || 1;
-    const idf = Math.log(domainCountsSize / occurrences) || 1;
+    const idf = Math.log(1 + domainCountsSize / occurrences);
 
     const age = this.normalizeTimestamp(entry.lastVisitDate);
     const imageCount = entry.images ? entry.images.length : 0;
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -35,11 +35,13 @@ class Baseline {
   extractFeatures(entry) {
     const urlObj = URL(entry.url);
     const host = urlObj.host;
+    // For empty profiles.
     const occurrences = this.domainCounts.get(host) || 1;
+    const domainCountsSize = this.domainCounts.size || 1;
+    const tf = entry.visitCount || 1;
+    const idf = Math.log(domainCountsSize / occurrences) || 1;
 
     const age = this.normalizeTimestamp(entry.lastVisitDate);
-    const tf = Math.max(entry.visitCount, 1);
-    const idf = Math.log(this.domainCounts.size / occurrences);
     const imageCount = entry.images ? entry.images.length : 0;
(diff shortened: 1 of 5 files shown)


content-src/selectors/selectors.js
```diff
@@ -91,7 +91,7 @@ module.exports.selectNewTabSites = createSelector(
 
     let topHighlights = spotlightRows;
     if (prefWeightedHighlights) {
-      const weightedRows = WeightedHighlights.rows.concat(WeightedHighlights.init ? firstRunData.Highlights : []);
+      let weightedRows = WeightedHighlights.rows.length > 5 ? WeightedHighlights.rows : WeightedHighlights.rows.concat(firstRunData.Highlights);
```
ncloudioj: Let's have a constant for `5` here for better readability.

content-test/selectors/selectors.test.js
```diff
@@ -244,6 +247,19 @@ describe("selectors", () => {
 
       state = selectNewTabSites(Object.assign({}, fakeState, weightedHighlights));
       assert.equal(state.Spotlight.rows.length, firstRunData.Highlights.length);
+      state.Spotlight.rows.forEach((row, i) => {
+        assert.equal(row.url, firstRunData.Highlights[i].url);
+      });
+    });
+    it("should append First Run data if less < 6 highlights", () => {
```
ncloudioj: s/less </less than/

common/recommender/Baseline.js
```diff
@@ -35,11 +35,13 @@ class Baseline {
   extractFeatures(entry) {
     const urlObj = URL(entry.url);
     const host = urlObj.host;
+    // For empty profiles.
     const occurrences = this.domainCounts.get(host) || 1;
+    const domainCountsSize = this.domainCounts.size || 1;
+    const tf = entry.visitCount || 1;
+    const idf = Math.log(domainCountsSize / occurrences) || 1;
```
ncloudioj: I'm not sure whether or not 1 is a good default idf here. Consider using the "idf smooth" instead?

[Wikipedia TF-IDF](https://en.wikipedia.org/wiki/Tf%E2%80%93idf)

ncloudioj:
Looks good. Just a few nits, and a possible optimization. I'd like to take another look after the question gets answered.

common/recommender/Baseline.js
piatra: @ncloudioj the value `1` can only occur if you only visit a single domain which seems difficult to achieve but smoothing makes sense.

ncloudioj:
r+ 🚢

Issue #1090, opened 2016-08-15
[Meta] Cache the rendered addon to improve perceived performance


jaredkerim:
[edit @k88hudson]

- [x] Change vendor strategy to allow dependencies to be used in common #1256
- [x] Refactor redux store to be used on addon and content side #1257
- [x] Add page worker to cache and allow rehydration of redux state #1258 
- [ ] Add html caching via rendering to a string and storing in locastorage (needs bug)

We should find a way to pre render the display content of the new tab page and then quickly access that from a cache so that when someone starts a new tab we don't go through the entire render/query flow each time.

rlr:
see also issue #320

jaredkerim:
Close attention needs to be paid to how this affects UX and data staleness.

jaredkerim:
This will be handled in #1092

jaredkerim:
Reopening and assigning to @k88hudson with the design using page workers:

https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/page-worker

to asynchronously render the page in page worker and store the output in a way that can quickly be sent to new tabs.

PR #1243, opened 2016-09-08
feat(redux): #1090 move store to common

k88hudson:
This PR moves our store logic from `content-src` to `common`. It includes:

- a change from a store singleton to a store creator, which can be used both in addon and content
- tests
- replaces the redux logger, which depends on some console-specific stuff (that mostly looked bad in Firefox anyway) with a simple custom logger
- an empty store in addon (will be filled in in future PRs)

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1243)


common/ReduxChannel.js
```diff
@@ -0,0 +1,87 @@
+const DEFAULT_OPTIONS = {
+  timeout: 20000,
```
pdehaan: Nit: i assume this is in ms, but it'd be nice to have a friendly comment w/ times with a user friendly value. 
For example: `timeout: 20000, // 20 seconds`

common/ReduxChannel.js
```diff
@@ -0,0 +1,87 @@
...
+        }, time);
+        timeouts.set(meta.expect, timeout);
+      }
+
+      // If the action has the right "broadcast" property on the meta property,
+      // send it to the other side of the channel.
+      if (meta.broadcast === this.options.outgoing) {
+        this.broadcast(action);
+      }
+    }.bind(this);
```
pdehaan: Do we need a `.bind(this)` if we convert this to an arrow function above on L49?

common/redux-logger.js
```diff
@@ -0,0 +1,4 @@
+module.exports = store => next => action => {
+  console.log("ACTION", action); // eslint-disable-line no-console
```
pdehaan: Do we need this `console.log()`, or was that for debugging?

common/vendor-src.js
```diff
@@ -4,11 +4,16 @@
...
+  "redux": require("redux"),
+  "redux-thunk": require("redux-thunk"),
   "seedrandom": require("seedrandom"),
   "url-parse": require("url-parse"),
   "DoublyLinkedList": require("DoublyLinkedList/doubly-linked-list").DoublyLinkedList
 };
 
 module.exports = function vendor(moduleName) {
+  if (!vendorModules[moduleName]) {
+    throw new Error(`Tried to import '${moduleName}' but it was not defined in common/vendor-src.js. Maye you need to add it?`);
```
pdehaan: Maye => Maybe.

common/redux-logger.js
k88hudson: this is specifically a logger (i.e. logging redux state to console), so we do need it

common/ReduxChannel.js
```diff
@@ -0,0 +1,87 @@
+const DEFAULT_OPTIONS = {
```
k88hudson: This is just moved over from `content-src`, with one change (had to check existence of `window` so it doesn't throw)

common/create-store.js
```diff
@@ -0,0 +1,71 @@
+const vendor = require("common/vendor");
```
k88hudson: this is very similar to `content-src/store`

common/parse-url-middleware.js
```diff
@@ -0,0 +1,25 @@
+const urlParse = require("common/vendor")("url-parse");
```
k88hudson: this is just moved over from `content-src`

common/ReduxChannel.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+    if (!options.outgoing) {
+      throw new Error("You must specify an outgoing event name with the 'outgoing' option.");
+    }
+    this.options = Object.assign({}, DEFAULT_OPTIONS, options);
+    this.callbacks = new Set();
+    this.timeouts = new Map();
+
+    this._onReceiveEvent = this.runCallbacks.bind(this);
+    this.options.target.addEventListener(this.options.incoming, this._onReceiveEvent);
```
sarracini: is there any reason we can't just do ```this.options.target.addEventListener(this.options.incoming, this.runCallbacks.bind(this))``` directly? so just get rid of ```this._onReceiveEvent```

common/create-store.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * rehydrateFromLocalStorage - Fetches initial state from local storage
+ *
+ * @return {obj}  Redux initial state as a plain object
+ */
+function rehydrateFromLocalStorage() {
+  let state = {};
+  try {
+    state = JSON.parse(localStorage.getItem(LOCAL_STORAGE_KEY)) || {};
```
sarracini: is there any way for JSON.parse to be successful but we still choose to have an empty object instead?

common/ReduxChannel.js
k88hudson: do you mind if I just do that in a follow-up ticket since this is just a straight up copy

common/create-store.js
sarracini: also just out of curiosity, where the heck is ```localstorage``` declared/coming from in this case?

common/ReduxChannel.js
sarracini: yup no problemo

content-src/main.js
```diff
@@ -3,7 +3,14 @@ const ReactDOM = require("react-dom");
...
 const Routes = require("components/Routes/Routes");
-const store = require("./store");
+const createStore = require("common/create-store");
+const {ADDON_TO_CONTENT, CONTENT_TO_ADDON} = require("common/event-constants");
+
+const store = createStore({
+  incoming: ADDON_TO_CONTENT,
+  outgoing: CONTENT_TO_ADDON,
+  logger: __CONFIG__.LOGGING
+});
```
sarracini: 🎉

common/create-store.js
k88hudson: localStorage is a global on window

PR #1251, opened 2016-09-09
Add page worker

### LINKED ISSUE #1090: [Meta] Cache the rendered addon to improve perceived performance

k88hudson:
This PR adds a page worker for caching state on the add-on side, and replaces messages on the content side with a call to `localStorage` to restore state from a copy of the global one. It includes:

- a global "master" add-on redux store (see `._store` on `ActivityStream.js`)
- removing messages like `HIGHLIGHTS_REQUEST` and replaces with a with a function called `._refreshAppState`, which dispatches changes directly to the master store
- a new `PageWorker` constructor, which takes advantage of [sdk/page-worker](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/page-worker) to create a hidden browser window to store a copy of the browser store in `localStorage`
- the content side "rehydrates" from the copy of the master store in `localStorage`, i.e. it loads a copy of the store when creating its child store

Fixes #1090


k88hudson:
Tests failing due to intermittents

k88hudson:
Ok so, perf meter tests are now re-written as of #1255, but now we have tab tracker failures 😭

k88hudson:
@sarracini i'm giving this to @piatra since he doesn't have any reviews right now

content-src/main.scss
```diff
@@ -53,12 +53,23 @@ a {
   z-index: 100;
 }
 
+@keyframes fadeIn {
```
k88hudson: This adds a little fade in on first load

data/page-worker/page-worker.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Page Worker</title>
+  </head>
+  <body>
+    <script>
+      window.addEventListener("addon-to-content", e => {
```
k88hudson: Will use proper constants once this is converted to use a webpack build, but that won't happen until we do the html cacheing

addon/PageWorker.js
```diff
@@ -0,0 +1,39 @@
...
+const {data} = require("sdk/self");
+const {LOCAL_STORAGE_KEY} = require("common/constants");
+const {ADDON_TO_CONTENT} = require("common/event-constants");
+
+class PageWorker {
+  constructor({store}) {
+    this._page = null;
+    this._onDispatch = null;
+    this._unsubscribe = null;
+    this._store = store;
```
piatra: Since it's expected to be called with a `store` property should we throw an error if `store` is `undefined`?

addon/PageWorker.js
```diff
@@ -0,0 +1,39 @@
...
+      contentScriptFile: data.url("content-bridge.js"),
+      contentScriptWhen: "start"
+    });
+    this._onDispatch = () => {
+      this._page.port.emit(ADDON_TO_CONTENT, {
+        type: LOCAL_STORAGE_KEY,
+        data: this._store.getState()
+      });
+    };
+    this._unsubscribe = this._store.subscribe(this._onDispatch);
```
piatra: #nit

```
To unsubscribe the change listener, invoke the function returned by subscribe.
```
That is __not obvious__ unless you go and read the docs. Can you please add a comment about this otherwise reading this line might seem very confusing.

addon/PageWorker.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+  destroy() {
```
piatra: Shouldn't you also reset `this._onDispatch` to `null` in this method?

content-src/main.js
```diff
@@ -9,7 +9,8 @@ const {ADDON_TO_CONTENT, CONTENT_TO_ADDON} = require("common/event-constants");
 const store = createStore({
   incoming: ADDON_TO_CONTENT,
   outgoing: CONTENT_TO_ADDON,
-  logger: __CONFIG__.LOGGING
+  logger: __CONFIG__.LOGGING,
+  rehydrate: true
```
piatra: Where is this state being used?

content-src/main.scss
piatra: Has to be pretty!

addon/PageWorker.js
```diff
@@ -0,0 +1,39 @@
...
+  constructor({store}) {
+    this._page = null;
+    this._onDispatch = null;
+    this._unsubscribe = null;
+    this._store = store;
+  }
+  connect() {
+    this._page = Page({
+      contentURL: data.url("page-worker/page-worker.html"),
+      contentScriptFile: data.url("content-bridge.js"),
```
piatra: Not sure if it matters too much but, from the docs:
```
From Firefox 34, you can use "./myFile.html" as an alias for self.data.url("myFile.html")
```

piatra: It should be safe to use.

content-src/main.scss
piatra: Not 💯 % sure about this but if loading takes too long I don't see the fade in, just a sudden appearance of everything. Maybe fade happens when `render()` rendered ... nothing?

piatra:
This is a gif of AS loading on my machine (this branch). Not exactly accurate but it still takes ~2s to see the page. Is it faster for you?

gif of AS loading on my machine

addon/ActivityStreams.js
```diff
@@ -170,6 +176,67 @@ ActivityStreams.prototype = {
     }
   },
 
+  _processAndDispatchLinks(links, type) {
```
piatra: There's a part I don't understand. Before this method used to call `this.send` how are the dispatched changes propagated to the content side now by `dispatch`?

addon/PageWorker.js
k88hudson: Despite the docs, this doesn't seem to work 😕

addon/PageWorker.js
k88hudson: I'm actually going to move the onDispatch stuff to the constructor so this isnt necessary

addon/PageWorker.js
piatra: :(

content-test/addon/PageWorker.test.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+  it("should throw if you try to create a PageWorker without a store", () => {
+    assert.throws(() => {
+      assert.ok(new PageWorker());
+    }, "options.store is required");
+  });
+  describe("#connect", () => {
+    it("should create a Page instance and ._unsubscribe", () => {
+      pageWorker.connect();
+      assert.isObject(pageWorker._page);
```
piatra: nit: http://chaijs.com/api/assert/#method_instanceof

k88hudson: weirdly, `Page` is not actually a constructor

content-test/addon/PageWorker.test.js
```diff
@@ -0,0 +1,85 @@
...
+      assert.isFunction(pageWorker._unsubscribe);
+    });
+    it("should subscribe to the store with ._onDispatch", () => {
+      sinon.spy(pageWorker, "_onDispatch");
+      pageWorker.connect();
+      store.dispatch({type: "add", number: 42});
+      assert.equal(store.getState(), 42);
+      assert.called(pageWorker._onDispatch);
```
piatra: nit: this test is using side effects to make assertions. judging by the description you want to:
1. `stub` (so no side effects) the subscribe method of the store 
2. call `connect`
3. assert that `store.subscribe` was `calledOnce` and `calledWithExactly` `_onDispatch`

content-test/addon/PageWorker.test.js
piatra: sorry it's not using `new` my bad.

content-test/addon/PageWorker.test.js
k88hudson: While I originally stubbed out `subscribe`, I actually refactored this to actually test if `_onDispatch` is called to catch any breaking changes in `redux` as well (for example, if they change the way unsubscribes work, which is not very obvious, I want to know about it)

piatra: Faith in JS projects === 0. Fair enough!

k88hudson:
@piatra re: loading improvements, we won't *actually* see too much improvement until we do the HTML part (coming up next!)

addon/ActivityStreams.js
piatra: cleared over IRC.

addon/ActivityStreams.js
```diff
@@ -593,6 +603,8 @@ ActivityStreams.prototype = {
       contentScriptWhen: "start",
       attachTo: ["existing", "top"],
       onAttach: worker => {
+        this._refreshAppState();
```
piatra: Cleared over IRC.

addon/ActivityStreams.js
```diff
@@ -190,9 +190,9 @@ ActivityStreams.prototype = {
    *                    TODO: Refactor this in to a different functions that handle refreshing data separately
    */
   _refreshAppState() {
-    console.log("Refresh app state");
```
piatra: Just a thought. It would be nice to have some logging in place when running debug mode.

addon/ActivityStreams.js
```diff
@@ -239,7 +239,7 @@ ActivityStreams.prototype = {
     this._store.dispatch(am.actions.Response("PREFS_RESPONSE", simplePrefs.prefs));
 
     // Share
-    this._store.dispatch(am.actions.Response("SHARE_PROVIDERS_RESPONSE", this._shareProvider.socialProviders));
+    this._store.dispatch(am.actions.Response("SHARE_PROVIDERS_RESPONSE", this._shareProvider.socialProviders || []));
```
piatra: Doesn't social providers return an empty array by default? I see the [return statement](https://github.com/k88hudson/activity-stream/blob/bdcd57107194a4925a59c6848fd4c7aabbed6890/addon/ShareProvider.js#L147) is a map, even if empty you still get `[]` and if it's undefined it will crash either way.
I personally dislike this kind of defensive programming because it can propagate bugs away from the real source and make you chase them in different places in your codebase.

k88hudson: There is a *really* weird bug in the tests in which it somehow returns `undefined`, which should be impossible given the code path... the only thing i can think of is that it's got a reference to an object that isn't in the right scope or something. It appears to be a race condition related to the `Social` service. I filed https://github.com/mozilla/activity-stream/issues/1314 as a follow-up, since @rlr knows that code better than me

piatra: Can you please add a comment with or without the bug number at that line?

PR #1272, opened 2016-09-12
fix(metrics): #1271 Send ping on didUpdate, not receiveProps

### LINKED ISSUE #1271: Measure the right event for first render data
We should be looking at componentDidUpdate/componentDidMount instead of componentWillReceiveProps (which is before render)

Note that this will affect our latency numbers (they will go up) cc @emtwo

k88hudson:
This PR sends our key performance ping *after* a render occurs, not before. It will result in the load times recorded appearing to be slower in our metrics, since they will include render time.

Test this by adding `"extensions.@activity-streams.performance.log": true` to your dev prefs.

@emtwo do you think this is worth doing? if so, is there anywhere we can add a note for this?

Fixes #1271

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -13,13 +13,14 @@ const setFavicon = require("lib/set-favicon");
 const classNames = require("classnames");
 const PAGE_NAME = "NEW_TAB";
 
+let renderedOnce = false;
+
 const {MAX_TOP_ACTIVITY_ITEMS} = require("common/constants");
 
 const NewTabPage = React.createClass({
   getInitialState() {
     return {
       showSettingsMenu: false,
-      renderedOnce: false,
       showRecommendations: true
     };
   },
@@ -37,10 +38,12 @@ const NewTabPage = React.createClass({
     document.title = "New Tab";
     setFavicon("newtab-icon.svg");
   },
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.isReady && !this.state.renderedOnce) {
+  componentDidUpdate() {
+    // Note that isReady is set in the newTabPage selector (check the selectors/ folder)
+    // and it is possible it never fires, if responses are not received.
+    if (!renderedOnce && this.props.isReady) {
       this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
-      this.setState({renderedOnce: true});
+      renderedOnce = true;
     }
   },
   render() {


pdehaan:
Travis failed, but not sure if this is root cause or something else entirely:

```
JPM [error]   Message: Error: The ID "action-button--activity-streams-activity-streams-link" seems already used.
  Stack:
    create@resource://gre/modules/commonjs/sdk/ui/button/view.js:120:11
setup@resource://gre/modules/commonjs/sdk/ui/button/action.js:60:5
@resource://gre/modules/commonjs/sdk/core/disposable.js:48:56
dispatch@resource://gre/modules/commonjs/method/core.js:119:12
Disposable<.initialize@resource://gre/modules/commonjs/sdk/core/disposable.js:71:17
constructor@resource://gre/modules/commonjs/sdk/core/heritage.js:146:23
_setupButton@resource://activity-streams/addon/ActivityStreams.js:629:20
ActivityStreams@resource://activity-streams/addon/ActivityStreams.js:127:3
getTestActivityStream@resource://activity-streams/test/lib/utils.js:69:17
@resource://activity-streams/test/test-PrefsProvider.js:4:12
findTests/</getNextTest@resource://gre/modules/commonjs/sdk/deprecated/unit-test-finder.js:162:25
findTests/</<.getNext@resource://gre/modules/commonjs/sdk/deprecated/unit-test-finder.js:195:32
runNextTest@resource://gre/modules/commonjs/sdk/deprecated/unit-test.js:532:14
start/this.onDone@resource://gre/modules/commonjs/sdk/deprecated/unit-test.js:563:7
done/</<@resource://gre/modules/commonjs/sdk/deprecated/unit-test.js:389:31
notify@resource://gre/modules/commonjs/sdk/timers.js:40:9
```

k88hudson:
:'(

k88hudson:
@emtwo ok, thanks! 
@dmose R?

dmose:
I just kicked travis, maybe things will go better this time.

Interestingly, NEWTAB_RENDER is never actually logged with this patch when I test it, but I think I see part of why...

dmose:
So the if clause is never getting called because nextProps.isReady is apparently never being set.  It's not actually obvious to me where in the code isReady would ever be passed to NewTabPage.  I'll poke at that more later...

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -37,10 +38,10 @@ const NewTabPage = React.createClass({
     document.title = "New Tab";
     setFavicon("newtab-icon.svg");
   },
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.isReady && !this.state.renderedOnce) {
+  componentDidUpdate(nextProps) {
+    if (!renderedOnce && nextProps.isReady) {
```
dmose: The problem was that componentDidUpdate doesn't take any params at here.  Changing 

```
nextProps.isReady
```

to 

```
this.props.isReady
```

seems to fix things.

That said, it's still not clear what is causing this.props.isReady, which makes this code a bit hard to reason about.  Can you add a comment describing why this is the right test?

dmose:
Note that since we're not checking in componentDidMount there's an implicit assumption that isReady could never be true when the component first mounts.  I can believe that this is the case with the current code structure; whether that continues to be true over time, I can't guess.  Might be worth a comment to that effect as well.

content-src/components/NewTabPage/NewTabPage.js
k88hudson: Lol 😜  good catch.. totally derp on my part

`isReady` is set by the new tab selector, which is right here:

https://github.com/mozilla/activity-stream/blob/master/content-src/selectors/selectors.js#L107

It will get triggered when "init" is true for all of these things

k88hudson:
We used to set `init` on timeout, although we don't seem to anymore. I'm going to file a ticket for that and add a comment.

PR #1273, opened 2016-09-12
fix(content): Closes #1149 Remove best image logic

### LINKED ISSUE #1149: Remove the 'best image' logic from activity stream
Right now activity stream receives multiple images from embedly and chooses a best image to use as a preview, however the new metadata service will only return maximum one image which it decides is the best image.  Right now we use the image sizes to determine which one to use, however if we let the service (or local parser) decide based on semantic markup then we remove the need for that logic and also the need for any kind of image size calculations altogether.

sarracini:
@jaredkerim r?

cc @piatra for sanity check 

Fixes #1149

### DIFF
--- a/common/getBestImage.js
+++ b/common/getBestImage.js
@@ -1,29 +1,7 @@
-const IMG_HEIGHT = 226;
-const IMG_WIDTH = 124;
-
 module.exports = function getBestImage(images) {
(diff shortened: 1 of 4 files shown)


piatra:
👍 from me. 🚢 it!

jaredkerim:
R+!

PR #1280, opened 2016-09-13
1274 score consecutive images

### LINKED ISSUE #1274: Lower the score of consecutive items with the same image URL
If two or more consecutive items show up with the same image we should try to separate them.

Example
screen shot 2016-09-12 at 5 28 45 pm 1

Solution
Use the same strategy as for results from the same domain: lower the score by a certain amount.

piatra:
Fixes #1274

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -290,25 +290,27 @@ class Baseline {
   }
 
   /**
-   *  Decrease the score for consecutive items from the same domain.
-   *  Combined with sorting by score the result is we don't see consecutive
-   *  items from the same domain.
+   *  Decrease the score for consecutive items which are similar (see `_similarItems`).
+   *  Combined with sorting by score the result is we don't see similar consecutive
+   *  items.
    *
    *  @param {Array} entries - scored and sorted highlight items.
    */
   dedupe(entries) {
     let penalty = 0.6;
 
-    return entries.map((entry, i, arr) => {
-      if (i > 0 && this._similarItems(arr[i - 1], entry)) {
-        let score = entry.score * penalty;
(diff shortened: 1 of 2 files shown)


common/recommender/Baseline.js
```diff
@@ -284,7 +306,7 @@ class Baseline {
     let penalty = 0.8;
 
     return entries.map((entry, i, arr) => {
-      if (i > 0 && arr[i - 1].host === entry.host) {
+      if (i > 0 && this._similarItems(arr[i - 1], entry)) {
```
sarracini: I think this is cleaner and makes more sense:

```js
_similarItems(prev, curr) {
    const imgPrev = getBestImage(prev.images);
    const imgCurr = getBestImage(curr.images);
    const hasImage = imgPrev && imgCurr;
    return prev.host === curr.host || (hasImage && imgPrev.url === imgCurr.url);
  }
  
dedupe(entries) {
    let penalty = 0.8;
    if (entries.length > 0) {
      entries.reduce((prev, curr) => {
        if (this._similarItems(prev, curr)) {
          let score = curr.score * penalty;
          penalty -= 0.2;
          curr.score = score;
        }
        penalty = 0.8;
        return curr;
      });
    }
    return entries;
  }

```

So essentially you can use reduce here, since you're comparing to previous values in the array and modifying it. I tested it exactly like this and it still passes the ```Baseline.test.js``` tests

piatra:
I am not sure about the refactoring of the `dedupe` method.
Calling `reduce` makes it seem like it's accumulating something when actually it's using side effects for the benefit of having nicer argument names. If I added an accumulator as the prev value it would defeat the purpose of the refactoring.
I could add more tests for this method to make sure more scenarios are covered.

PR #1282, opened 2016-09-13
fix(metadata): Closes #1278 Fix YouTube Top Sites icon

### LINKED ISSUE #1278: YouTube Top Sites Icon is 😔
image 2016-09-13 at 9 19 25 am

The YouTube icon doesn't look very good. That top site links to `http://www.youtube.com/`. It seems like it should be the nicer icon from Tippy Top sites.

sarracini:
- Ensures we _always_ add tippytop data if we have it 
- As a result the youtube icon is nice
- Update the tests

@jaredkerim r?

### DIFF
--- a/addon/PreviewProvider.js
+++ b/addon/PreviewProvider.js
@@ -208,17 +208,18 @@ PreviewProvider.prototype = {
     let existingLinks = new Map();
     dbLinks.forEach(item => existingLinks.set(item.cache_key, item));
     let results = processedLinks.map(site => {
-      let enhancedLink = null;
-
-      if (site) {
-        // Find the item in the map and return it if it exists
-        if (existingLinks.has(site.cache_key)) {
-          enhancedLink = Object.assign({}, existingLinks.get(site.cache_key), site);
-        } else {
-          // Add tippy top data, if available
-          enhancedLink = this._tippyTopProvider.processSite(site);
-        }
+      if (!site) {
+        return null;
       }
+      let enhancedLink = Object.assign({}, site);
+      // Find the item in the map and return it if it exists
+      if (existingLinks.has(site.cache_key)) {
+        Object.assign(enhancedLink, existingLinks.get(site.cache_key));
+      }
+
+      // Add tippy top data, if available
+      Object.assign(enhancedLink, this._tippyTopProvider.processSite(enhancedLink));
+
       return enhancedLink;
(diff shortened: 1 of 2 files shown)


test/test-PreviewProvider.js
```diff
@@ -303,7 +303,58 @@ exports.test_no_metadata_source = function*(assert) {
...
+    response.write(JSON.stringify(fakeResponse));
+  });
+  // insert a link with some less nice icons in it and get them back
+  yield gPreviewProvider._asyncSaveLinks([fakeSite]);
+  cachedLink = yield gPreviewProvider._asyncGetEnhancedLinks([fakeSite]);
+  tippyTopFavicon = cachedLink[0].favicon_url;
+  tippyTopBackgroundColor = cachedLink[0].background_color;
+
+  assert.equal(tippyTopLink.favicon_url, tippyTopFavicon, "we still took the better tippyTop favicon_url");
+  assert.equal(tippyTopLink.background_color, tippyTopBackgroundColor, "we still took the better tippyTop background_color");
```
jaredkerim: Can we add an assertion to test that the link object we get back has both the fakeResponse metadata *and* the tippy top data?  If we only check the tippy top fields, we could remove line 351 and this would still pass.

test/test-PreviewProvider.js
```diff
@@ -387,11 +438,15 @@ before(exports, () => {
     },
     asyncGetMetadataByCacheKey(cacheKeys) {
```
jaredkerim: If this method isn't called anybody we should remove it.

addon/PreviewProvider.js
```diff
@@ -209,16 +209,18 @@ PreviewProvider.prototype = {
...
-
-      if (site) {
-        // Find the item in the map and return it if it exists
-        if (existingLinks.has(site.cache_key)) {
-          enhancedLink = Object.assign({}, existingLinks.get(site.cache_key), site);
-        } else {
-          // Add tippy top data, if available
-          enhancedLink = this._tippyTopProvider.processSite(site);
-        }
+      if (!site) {
```
jaredkerim: If you move this one line up, you can get rid of 

    let enhancedLink = null;

altogether and just add a let to line 215.

jaredkerim:
R+ with test changes and nits.

Issue #1264, opened 2016-09-09
Inappropriate 'Hang on tight'


tspurway:
here's the traceback:

Content Security Policy: Directive ‘frame-src’ has been deprecated. Please use directive ‘child-src’ instead.(unknown)
TypeError: can't access dead object(unknown)
Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src resource://activity-streams”).activity-streams.html:1
A promise chain failed to handle a rejection. Did you forget to '.catch', or did you forget to 'return'?
See https://developer.mozilla.org/Mozilla/JavaScript_code_modules/Promise.jsm/Promise

Date: Fri Sep 09 2016 16:50:39 GMT-0400 (EDT)
Full Message: Error: Could not compute feature score
Full Stack: decay@resource://gre/modules/commonjs/toolkit/loader.js -> resource://activity-streams/common/recommender/Baseline.js:263:13
scoreEntry@resource://gre/modules/commonjs/toolkit/loader.js -> resource://activity-streams/common/recommender/Baseline.js:73:17
score/results<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://activity-streams/common/recommender/Baseline.js:168:35
score@resource://gre/modules/commonjs/toolkit/loader.js -> resource://activity-streams/common/recommender/Baseline.js:166:19
scoreEntries@resource://gre/modules/commonjs/toolkit/loader.js -> resource://activity-streams/common/recommender/Recommender.js:13:12
_respondToPlacesRequests/</<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://activity-streams/addon/ActivityStreams.js:233:43
Handler.prototype.process@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:937:23
this.PromiseWalker.walkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:816:7
this.PromiseWalker.scheduleWalkerLoop/<@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:750:11

piatra:
Likely related to https://github.com/mozilla/activity-stream/pull/1242 which that pull request will fix.
Keeping this open until that pull request lands.

piatra:
Pull request for #1241 landed and closed the bug, this should also be fixed. Closing this unless it's reproduced.

tspurway:
[...]

PR #1294, opened 2016-09-14
Bug 1264 missing image size

piatra:
Fix for #1264, new metadata service does not return image size. Feature value ends up being `undefined` and therefore the final score is NaN. Using `width x height` to replace image size.

Small updates to the test as well to ensure all feature extraction methods provide defaults and that all function calls are made, because one test was actually making assertions on a stubbed method.

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -221,11 +221,11 @@ class Baseline {
   extractImage(images) {
     const image = getBestImage(images);
 
-    if (!image) {
+    if (!image || !image.width || !image.height) {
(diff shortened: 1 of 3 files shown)


common/recommender/Baseline.js
```diff
@@ -221,11 +221,11 @@ class Baseline {
   extractImage(images) {
     const image = getBestImage(images);
 
-    if (!image) {
+    if (!image || !image.width || !image.height) {
```
k88hudson: can you just add a comment for this

PR #1308, opened 2016-09-15
fix (addon): #1228 simplify url hiding

### LINKED ISSUE #1228: URL overriding causes uneccessary performance hits
I noticed while digging around the the URL overriding code that we add a listener/do a few things to override URLS for new tab that I *think* are unnecessary and could cause locking of the URL bar. Need to discuss with @rlr.

rlr:
Fixes #1228


rlr:
@k88hudson I changed router to use memory instead of the hash. What you think? All seems to work fine.

PR #1313, opened 2016-09-16
switch timing uses of Date.now to performance.now wrapper for better accuracy 

dmose:
This PR switches the various timing uses of Date.now (except for httpd.js, which seems to be an external module, and I'd be surprised if we care about the timing there).

One can test some data collection by adding this to dev-prefs.json

```json
  "extensions.@activity-streams.performance.log": true
```

On my relative new Macbook Pro, opening and closing 8 new tabs on master without this patch yielded this for NEWTAB_RENDER:

```
SIZE=8 MEAN=519.63 STD=31.69 MEDIAN=514.5
```

with the commits in this PR, it yielded

```
SIZE=8 MEAN=526.43 STD=30.83 MEDIAN=511.6549072265625
```

Which I would guess is not statistically significant.  That said, I'd also expect that it'll come in handing for we're timing smaller things, plus it avoids the occasional "NTP clock shift" issues.

Note that this change will effect all performance logging that uses PerfMeter, making our new, slightly more accurate numbers theoretically hard to compare against the existing data.  CC @emtwo

### DIFF
--- a/addon/TabTracker.js
+++ b/addon/TabTracker.js
@@ -6,6 +6,7 @@ const self = require("sdk/self");
 const {uuid} = require("sdk/util/uuid");
 const simplePrefs = require("sdk/simple-prefs");
 const eventConstants = require("../common/event-constants");
+const {absPerf} = require("common/AbsPerf");
 
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/Locale.jsm");
@@ -177,7 +178,7 @@ TabTracker.prototype = {
       return;
     }
     if (this._tabData.start_time) {
-      this._tabData.session_duration = (Date.now() - this._tabData.start_time);
+      this._tabData.session_duration = (absPerf.now() - this._tabData.start_time);
       delete this._tabData.start_time;
     }
     delete this._tabData.active;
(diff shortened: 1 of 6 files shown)


addon/TabTracker.js
```diff
@@ -177,7 +178,7 @@ TabTracker.prototype = {
       return;
     }
     if (this._tabData.start_time) {
-      this._tabData.session_duration = (Date.now() - this._tabData.start_time);
+      this._tabData.session_duration = (absPerf.now() - this._tabData.start_time);
```
ncloudioj: nit: according to [here](https://github.com/mozilla/activity-stream/blob/master/data_dictionary.md), `session_duration` should be an integer in milliseconds, let's round up the result of performance.now here

ncloudioj: I'd recommend to expose two APIs for absPerf, i.e. one could be used as a drop-in replacement of `Date.now()` which returns an integer, and the other one returns a floatpoint value that has the microsecond resolution.

dmose: So I just did the drop-in replacement of Date.now() for the moment.  We can do further refinement once we have a caller for higher resolution timing.

ncloudioj:
Looks great! r+ :shipit:

PR #1328, opened 2016-09-19
fix(timeline): #1323 Fix infinite scroll events for timeline

### LINKED ISSUE #1323: Timeline loader spins for 20 seconds then shows nothing
An initial bisect seems to point to page worker from #1251 @k88hudson ?

Happens for both "All" and "Bookmarks"

k88hudson:
This PR fixes a bug introduced by the new page worker strategy, in which infinite scroll no longer works on the Timeline pages.

Fixes #1323

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -155,10 +155,12 @@ ActivityStreams.prototype = {
   /**
    * Send a message to a worker
    */
-  send(action, worker) {
+  send(action, worker, skipMasterStore) {
     // if the function is async, the worker might not be there yet, or might have already disappeared
     try {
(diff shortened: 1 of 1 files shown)


Mardak:
This looks to fix things for me. Does it make sense to have a test that makes sure expected requests are processed? Or at least the `Places` ones and stub out any calls to `PlacesProvider.links.*()`?

k88hudson:
@Mardak yeah, I definitely feel like we need to do some tests for this, the problem is that they are all sort of tacked on to `ActivityStream.js` so it makes it really hard test them independently. Do you might if I file a follow-up for that?

PR #1336, opened 2016-09-19
fix(highlights): Query was calling the wrong thing

k88hudson:


piatra:
R+ 🚢 before anyone notices 😆

PR #1344, opened 2016-09-20
Initialize experiment provider in constructor fixes #1343

### LINKED ISSUE #1343: Move experimentProvider init into constructor
Presently we initialize the experiment provider outside of its constructor which could theoretically lead to the case where we attempt to access its member variables before it is initialized we could lead to destructive behaviour in our experiments.  We should initialize it in its constructor.

jaredkerim:


jaredkerim:
@sarracini r?

PR #1346, opened 2016-09-20
fix(timeline): Filter links with SQL and only show filter-compatible results

### LINKED ISSUE #1261: Improve responsiveness when clearing timeline search
As noted in #1199, when clearing a search, all items that were filtered out try to immediately appear. There could be many items that are showing up at the same time causing a pause in responsiveness when rendering all that content.

screen shot 2016-09-15 at 10 21 05 am

### LINKED ISSUE #1302: Search endlessly loading...
When searching from the timeline, the loading spinner continues to load endlessly. It may finally complete, and I just didn't wait long enough.

screen recording 2016-09-15 at 10 10 am

Mardak:
Fix #1302 by passing the filter to `PlacesProvider` allowing it to do the multi-token matching on title/url. These SQL-filtered results are tagged with the filter they were matched with, so that the selector only shows the results that are relevant to the current filter. This is faster because the 20 results the timeline gets are already most likely going to match the filter instead of going through many "pages" of non-matches.

This also fixes #1261 at the same time as it doesn't try showing many pages of results when clearing search. By including the tagged filter, more specific searches can be quickly ignored when deleting in addition to there being fewer fetched rows in the first place.

r?@sarracini

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -335,7 +335,7 @@ Links.prototype = {
    * @returns {Promise} Returns a promise with the array of links as payload.
    */
   _getHistoryLinks: Task.async(function*(options = {}) {
-    let {limit, afterDate, beforeDate, order, ignoreBlocked} = options;
+    let {limit, afterDate, beforeDate, filter, order, ignoreBlocked} = options;
     if (!limit || limit.options > LINKS_QUERY_LIMIT) {
       limit = LINKS_QUERY_LIMIT;
     }
@@ -357,6 +357,22 @@ Links.prototype = {
     if (beforeDate) {
       beforeDateClause = "AND moz_places.last_visit_date < :beforeDate * 1000";
       params.beforeDate = beforeDate;
+    }
+
+    // setup filter binding and sql clause
+    let filterClause = "";
+    if (filter) {
+      // Match each token in either the title or url
+      let tokens = filter.trim().toLowerCase().split(/\s+/);
+      tokens.forEach((token, i) => {
+        filterClause = `${filterClause} AND (
+                          moz_places.title LIKE :filter${i} OR
+                          moz_places.url LIKE :filter${i})`;
+        params[`filter${i}`] = `%${token}%`;
+      });
+    }
+    else {
+      filter = "";
     }
 
     // setup order by clause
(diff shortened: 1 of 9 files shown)


common/action-manager.js
```diff
@@ -140,9 +140,9 @@ function RequestRecentLinks(options) {
   return RequestExpect("RECENT_LINKS_REQUEST", "RECENT_LINKS_RESPONSE", options);
 }
 
-function RequestMoreRecentLinks(beforeDate) {
+function RequestMoreRecentLinks(beforeDate, filter = "") {
   return RequestRecentLinks({
-    data: {beforeDate},
+    data: {beforeDate, filter},
```
Mardak: Yup. It's to include the query for fetching more when infinitely scrolling.

addon/PlacesProvider.js
```diff
@@ -359,6 +359,22 @@ Links.prototype = {
       params.beforeDate = beforeDate;
     }
 
+    // setup filter binding and sql clause
+    let filterClause = "";
+    if (filter) {
+      // Match each token in either the title or url
+      let tokens = filter.trim().toLowerCase().split(/\s+/);
+      tokens.forEach((token, i) => {
+        filterClause = `${filterClause} AND (
```
Mardak: The other Clauses start with "AND ..." because they're chaining to the `WHERE`. This is the same except there's a dynamic number of chains based on how many tokens are being used, e.g., " AND token1 ... AND token2 ... AND token3 ..." The empty string is the initial value so that the subsequent token clauses can be concatenated.

content-src/selectors/selectors.js
```diff
@@ -164,7 +164,10 @@ module.exports.selectHistory = createSelector(
     return {
       Spotlight: Object.assign({}, Spotlight, {rows}),
       Filter,
-      History
+      History: Object.assign({}, History, {
+        // Only include rows that are less filtered than the current filter
```
Mardak: A query of "query" is more filtered/specific than a query of "qu" which is more specific than "". If the current query is "que" it would include rows for "" and "qu" but not "query" because "query" is more specific. This is desired because the user might be searching for "queso".

sarracini:
looks good, R+!

PR #1355, opened 2016-09-21
chore(benchmark): force GC for each run in the benchmark

ncloudioj:
Came across this function `Cu.forceGC()` while working on writing some benchmark baselines. I think it's helpful to get a more comparable benchmark results with it.

### DIFF
--- a/test/lib/Benchmark.js
+++ b/test/lib/Benchmark.js
@@ -1,7 +1,7 @@
 /* globals XPCOMUtils, Services, Task */
 "use strict";
 
-const {Cu} = require("chrome");
+const {Ci, Cu} = require("chrome");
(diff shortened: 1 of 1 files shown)


dmose:
Good idea!  However, I'd suggest using nsIDOMWindowUtils.garbageCollect (see the note in https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.forceGC), as this will run both the garbage collector and the cycle collector.  https://adblockplus.org/blog/different-ways-to-force-garbage-collection has a code sample of how to 
get to that starting from the Window object.

r=dmose with that change

ncloudioj:
@dmose Nice! Now I already know two ways to force GC now. Thanks a lot!

Issue #1356, opened 2016-09-21
Color Analyzer still throwing 


k88hudson:
Just received a report from @jbuck that `ColorAnalyzer` is still throwing, on line `53`. This is on version `1.1.3`, which does include the latest fix.

k88hudson:
This is the site and the image that causes the error:

https://compat.spec.whatwg.org/
https://resources.whatwg.org/logo-compat.svg

k88hudson:
Also, it looks like the error happens in an asynchronous context (i.e. image `onLoad` event) so the try-catch block wont catch the error 😢

k88hudson:
I looked into whether it was possibly failing on svgs, and I found this was not the case. For example, https://css-tricks.com/wp-content/uploads/2015/05/kiwi.svg loads fine

ncloudioj:
😡 ColorAnalyzer, I hate you.

K, will take a look what we can do.

PR #1358, opened 2016-09-21
fix(addon): #1356 Import ColorAnalyzer to protect against errors thrown in callback

k88hudson:
In this patch, I moved ColorAnalyzer into tree and patched the areas where there could be uncaught errors, as well as adding tests for this.


ncloudioj:
r+ 💯

pdehaan:
Can we push these changes upstream too, so they can benefit from our/your genius and more superior error handling?

PR #1359, opened 2016-09-21
feat(benchmark): #1236 add baseline benchmarks

### LINKED ISSUE #1236: Add some benchmark baselines for the addon
As the benchmark framework goes live, we can write some benchmark baselines for the addon. To begin with, we can cover following typical scenarios:
* benchmarking the load latency of A-S as a cold start
* benchmarking the load latency of A-S with pre-warmming cache/memoizer
* measure other time-consuming functions or code blocks if improved/optimized versions have been developed

@dmose has already come up with a good idea that "npm run perf" will trigger all those benchmarks in one shot.

### LINKED ISSUE #1237: Investigate the Firefox benchmark caveats
[Benchmarking in Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Benchmarking)

This page mentions a great deal of gotchas of Firefox benchmarking. Let's figure out how to take advantage of those wisdom in order to get more accurate results.

Ref to #1236

ncloudioj:
This will fix #1236 and fix #1237 

Highlights:
* Add the first one to benchmark the "log-performance-complete", i.e. "NEWTAB_RENDER"
* Add "npm run benchmark"
* Add some documents for benchmarking


ncloudioj:
@k88hudson @dmose So this is the first cut of the benchmark initiative. Could both of you take a look at it at your convenience please?

dmose:
I hope to get to this today, but it may end up being Monday.

benchmark_how_to.md
```diff
@@ -0,0 +1,108 @@
+# Benchmarking in Activity Stream
+
+When you add a new feature or want to test the performance of a change, you may want to benchmark the feature or the add-on as a wholoe. You can do that using our simple benchmarking framework.
+
+## Some useful notices
+
+* All the benchmarks are running on top of "jpm test", so you should put your benchmark files in the "test" directory
+* Name the benchmark file as "test-****-benchmark.js", e.g. "test-Feature-abc-benchmark.js". This is *very* important as we use the "benchmark" suffix to differentiate the benchmark files from the regular test ones. Note the leading "test" is also mandatory as jpm only recognizes test/benchmark files with it
+* A new prefs file "benchmark-prefs.json" is being used for the benchmarking, which sets the log level to "error" to avoid the verbose jpm test outputs. See examples below to see how to log in the benchmark
```
dmose: You might also mention that it more clearly mimics a production run as well.

benchmark_how_to.md
```diff
@@ -0,0 +1,108 @@
+# Benchmarking in Activity Stream
```
dmose: Somewhere near the top of this file, please mention that this framework is suitable for some things, and less so for others, and then link to the commentary in Benchmark.js on github that discusses this.

test/lib/Benchmark.js
```diff
@@ -2,12 +2,22 @@
...
                                   "resource://gre/modules/Task.jsm");
 
+/*
+ * Disable "Async Stacks" to make the benchmark yield more stable results
+ *
+ * See more detail at
+ *  https://developer.mozilla.org/en-US/docs/Mozilla/Benchmarking
+ */
+Services.prefs.setBoolPref("javascript.options.asyncstack", false);
+
```
dmose: Presumably, this only works if javascript.options.asyncstack is checked at runtime rather than startup time.  Is there some reason not to just put this in benchmark-prefs.json ?

test/lib/Benchmark.js
```diff
@@ -299,4 +309,15 @@ Benchmark.prototype = {
   })
 };
 
+const PREF_LOG_LEVEL = `extensions.${self.id}.sdk.console.logLevel`;
+
+// bLog allows the user to use the console.log() regardless the current log level
```
dmose: It's not obvious to me what the b in bLog stands for.  Maybe call this forcedLog?

test/test-Baselines-benchmark.js
```diff
@@ -0,0 +1,103 @@
...
+});
+
+/*
+ * Creates a benchmark object for message performance_log_complete
+ *
+ * Params:
+ *   benchTime is an integer in microseconds
+ * Returns:
+ *   a benchmark object
+ */
```
dmose: I'd suggest using jsdoc style for this, as there are lots of good (IDE and standalone) tools for auto-generating and doing interesting stuff with such comments.  Github even (mostly) knows how to highlight them:

```js
/**
 * Creates a benchmark object for message performance_log_complete
 *
 * @param {benchTime}     an integer in microseconds
 * @returns {Benchmark} 
 */
```

https://github.com/mozilla/loop/blob/master/shared/js/utils.js#L389 has a bunch of examples, and http://usejsdoc.org/ is the canonical reference.

benchmark-prefs.json
```diff
@@ -0,0 +1,9 @@
+{
+  "xpinstall.signatures.required": false,
+  "javascript.options.strict": false,
+  "extensions.@activity-streams.sdk.console.logLevel": "error",
+  "extensions.@activity-streams.performance.log": false,
+  "extensions.@activity-streams.previews.enabled": false,
+  "extensions.@activity-streams.telemetry": true,
+  "extensions.@activity-streams.telemetry.ping.endpoint": "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream"
+}
```
dmose: Please also add

```
  "browser.crashReports.unsubmittedCheck.autoSubmit": true
```

so that we don't get those "Submit unsent crash reports?" bottom bars rendered, since they'll change the rendering timing.

benchmark_how_to.md
dmose: Nicely written doc, by the way!

ncloudioj:
Pushed various changes based on the review.

Thanks a lot for the feedback! @dmose Those points are just right on! 👍

test/lib/Benchmark.js
```diff
@@ -299,4 +301,15 @@ Benchmark.prototype = {
   })
 };
 
+const PREF_LOG_LEVEL = `extensions.${self.id}.sdk.console.logLevel`;
+
+// forceLog allows the user to use the console.log() regardless the current log level
+function forceLog(...theArgs) {
+  let logLevel = prefs.get(PREF_LOG_LEVEL, "error");
+  prefs.set(PREF_LOG_LEVEL, "info");
+  console.log.apply(console, theArgs); // eslint-disable-line
```
k88hudson: you could alternatively just do `console.log.call(console, ...arguments)` instead of using `theArgs`

test/lib/Benchmark.js
```diff
@@ -299,4 +301,15 @@ Benchmark.prototype = {
   })
 };
 
+const PREF_LOG_LEVEL = `extensions.${self.id}.sdk.console.logLevel`;
```
k88hudson: This would be great to use in other parts of the application

test/lib/Benchmark.js
ncloudioj: Ah, that's neat.

PR #1362, opened 2016-09-21
feat(highlights) - Min number of results

### LINKED ISSUE #1275: Return a minimum number of results
Increase the number of results we fetch in the query so we get better results after (possible) long periods of inactivity. This is for highlights.

piatra:
Fixes #1275 

The query should fetch past 4 days of results OR past 20 results so that we have something to show to a user if he's been away from Firefox recently.

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -667,11 +667,10 @@ Links.prototype = {
 
   /**
    * Obtain a set of links to rank.
-   * Get MAX(all links visited in the past 4 days, last 20 links).
+   * Get all links visited in the past 4 days or default to last 20 links.
    * Last 20 links is used as fallback in case there is no recent history.
    *
    * @param {Object} options
-   *        {Integer} limit: Maximum number of results to return. Max 20
    *
    * @returns {Promise} Returns a promise with the array of links as payload.
    */
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -679,34 +679,67 @@ Links.prototype = {
     let {ignoreBlocked} = options;
 
     let blockedURLs = ignoreBlocked ? [] : this.blockedURLs.items().map(item => `"${item}"`);
+    let params = {limit: LINKS_QUERY_LIMIT};
 
-    let sqlQuery = `SELECT p.url as url,
-                           p.guid as guid,
-                           f.data as favicon,
-                           f.mime_type as mimeType,
-                           p.title as title,
-                           p.frecency as frecency,
-                           p.visit_count as visitCount,
-                           p.last_visit_date / 1000 as lastVisitDate,
-                           "history" as type,
-                           b.id as bookmarkId,
-                           b.guid as bookmarkGuid,
-                           b.title as bookmarkTitle,
-                           b.lastModified / 1000 as lastModified,
-                           b.dateAdded / 1000 as bookmarkDateCreated
-                    FROM moz_places p
-                    LEFT JOIN moz_bookmarks b
-                      ON b.fk = p.id
-                    LEFT JOIN moz_favicons f
-                      ON p.favicon_id = f.id
-                    WHERE datetime(p.last_visit_date / 1000 / 1000, 'unixepoch') > datetime('now', '${HIGHLIGHTS_THRESHOLDS.ageLimit}')
-                    AND p.url NOT IN (${blockedURLs})
-                    AND p.title NOT NULL
-                    AND p.rev_host NOT IN (${REV_HOST_BLACKLIST})`;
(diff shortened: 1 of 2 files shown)


addon/PlacesProvider.js
```diff
@@ -667,46 +667,78 @@ Links.prototype = {
...
+                          b.guid as bookmarkGuid,
+                          b.title as bookmarkTitle,
+                          b.lastModified / 1000 as lastModified,
+                          b.dateAdded / 1000 as bookmarkDateCreated
+                        FROM moz_places p
+                        LEFT JOIN moz_bookmarks b
+                          ON b.fk = p.id
+                        LEFT JOIN moz_favicons f
+                          ON p.favicon_id = f.id
+                        AND p.url NOT IN (${blockedURLs})
```
Mardak: This `AND` used to be part of the `WHERE` but happens to work in the `ON`. Not sure if it would make sense to have a shared query template accepting an optional extra `WHERE` and `LIMIT` to avoid these slight differences.

addon/PlacesProvider.js
```diff
@@ -667,46 +667,78 @@ Links.prototype = {
...
-                    FROM moz_places p
-                    LEFT JOIN moz_bookmarks b
-                      ON b.fk = p.id
-                    LEFT JOIN moz_favicons f
-                      ON p.favicon_id = f.id
-                    WHERE datetime(p.last_visit_date / 1000 / 1000, 'unixepoch') > datetime('now', '${HIGHLIGHTS_THRESHOLDS.ageLimit}')
-                    AND p.url NOT IN (${blockedURLs})
-                    AND p.title NOT NULL
-                    AND p.rev_host NOT IN (${REV_HOST_BLACKLIST})`;
+    let sqlQuery = `SELECT DISTINCT * FROM (
```
ncloudioj: I think this query is doing more than what you've mentioned above in the document. If I understand correctly, you wanted to apply 20 as the limit for this function, correct? If that's the case, I'd recommend to use two queries separately. e.g. using the second sub-query as the fallback if it couldn't fetch enough data from the first one.

One reason is to avoid the unnecessary database overhead. More importantly, the following function `this.faviconBytesToDataURI` is quite expensive, we should avoid feeding it with extra data if possible.

What do you think?

piatra:
```
you wanted to apply 20 as the limit for this function
```
No. I want the last 4 days of activity OR last 20 items from whenever. If you have never used FFx for the past 4 days the second part of the `UNION` will return no rows and only the first part will succeed with 20 results. If you have used FFx you will get both `last 20 links` and `last 4 days` and the `SELECT DISTINCT` will filter out the duplicates.

piatra: If there are concerns about performance we should discuss a limit of results for this query.

addon/PlacesProvider.js
piatra: @Mardak Thanks for mentioning that. I only validated it though tests. Personally I am not excited about the current state of PlacesProvider and would not want to add more abstraction on top of it (tldr: hard to debug failing queries), but if anyone has suggestions on how to improve that it's definitely a win.

addon/PlacesProvider.js
ncloudioj: I see. So you are trying to handle a special case here. But the query is doing "and" not "or", correct?

I wonder how often we're gonna run into that special case? If it's rare, there is no point to run the second query everytime. Whereas if it does happen, we can query it once and store the result for the following uses. It won't change anyways.

Yes, performance is always a concern given that we are currently suffering from the performance trouble. I think we should try our best to avoid any extra work if possible.

Make sense?

piatra: @ncloudioj it's actually a likely scenario: happened to me, I was presented with what was basically a first time use experience because no links matched the `last 4 days` query. 
This is however a polish bug, if you think performance is an issue we can drop it and close the pull request.
Personally I find it hard to believe sql queries are *a real bottleneck* without seeing some numbers.

ncloudioj: Well, I think your polish to handle this special case is fine. My question is that is it really necessary to include the query for the special case in the main query? What if you main query succeeds to fetch the data as expected? I think it's pretty wasteful to do that in this case. At least, you might want to apply the `limit` to the whole query to avoid the following computing overheads.

Agreed, we can't tell whether it's a bottleneck or not without obvious evidences. But my main concern here is *not* introducing the unnecessary work. Therefore, I'd recommend to leave the original query as it is. Just adding a separate code path to handle the special case. It does that if and only if it's needed.

jaredkerim:
Transferring ownership of this patch to @Mardak

Mardak:
Superseded by #1474.

PR #1377, opened 2016-09-22
feat(topsites): Show shorter top sites label without top level domain

### LINKED ISSUE #1376: Show shorter top sites label without top level domain
Split from #1311: instead of showing "news.ycombinator.com" or "youtube.com" show "news.ycombinator" or "youtube" instead.

Mardak:
Fix #1376 by using Services.eTLD to get the effective top level domain of a top frecent site url. Use that eTLD to remove it from the displayed hostname label.

r?@piatra

### DIFF
--- a/content-src/selectors/siteMetadataSelectors.js
+++ b/content-src/selectors/siteMetadataSelectors.js
@@ -8,7 +8,12 @@ function selectSiteProperties(site) {
   const metadataFavicon = site.favicons && site.favicons[0] && site.favicons[0].url;
   const favicon = site.favicon_url || metadataFavicon || site.favicon;
   const parsedUrl = site.parsedUrl || urlParse(site.url || "");
-  const label = prettyUrl(parsedUrl.hostname);
+
+  // Remove the eTLD (e.g., com, net) and the preceding period from the hostname
+  const eTLDLength = (site.eTLD || "").length;
+  const eTLDExtra = eTLDLength > 0 ? -(eTLDLength + 1) : Infinity;
+  const label = prettyUrl(parsedUrl.hostname).slice(0, eTLDExtra);
+
   return {favicon, parsedUrl, label};
 }
 
--- a/content-test/selectors/siteMetadataSelectors.test.js
+++ b/content-test/selectors/siteMetadataSelectors.test.js
@@ -82,10 +82,15 @@ describe("siteMetadataSelectors", () => {
 
       assert.ok(result.parsedUrl);
     });
-    it("should set label prop", () => {
+    it("should set label prop without eTLD set", () => {
       let result = selectSiteProperties(validSpotlightSite);
 
-      assert.ok(result.label);
+      assert.equal(result.label, "cnn.com");
+    });
+    it("should set label prop with eTLD set", () => {
+      let result = selectSiteProperties(Object.assign({}, validSpotlightSite, {eTLD: "com"}));
+
+      assert.equal(result.label, "cnn");
     });
   });
 });
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -470,6 +470,14 @@ Links.prototype = {
     });
 
     links = this._faviconBytesToDataURI(links);
+    links = links.map(link => {
+      try {
+        link.eTLD = Services.eTLD.getPublicSuffix(Services.io.newURI(link.url, null, null));
+      } catch (e) {
+        link.eTLD = "";
+      }
+      return link;
+    });
     return links.filter(link => LinkChecker.checkLoadURI(link.url));
   }),
 
--- a/content-test/selectors/colorSelectors.test.js
+++ b/content-test/selectors/colorSelectors.test.js
@@ -49,6 +49,7 @@ describe("colorSelectors", () => {
 
   describe("selectSiteIcon", () => {
     const siteWithFavicon = {
+      eTLD: "com",
       url: "http://foo.com",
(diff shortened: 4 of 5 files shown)


content-src/selectors/siteMetadataSelectors.js
```diff
@@ -8,7 +8,12 @@ function selectSiteProperties(site) {
   const metadataFavicon = site.favicons && site.favicons[0] && site.favicons[0].url;
   const favicon = site.favicon_url || metadataFavicon || site.favicon;
   const parsedUrl = site.parsedUrl || urlParse(site.url || "");
-  const label = prettyUrl(parsedUrl.hostname);
+
+  // Remove the eTLD (e.g., com, net) and the preceding period from the hostname
+  const eTLDLength = (site.eTLD || "").length;
```
piatra: Do you need the `||` if [here](https://github.com/mozilla/activity-stream/pull/1377/files#diff-49c53f46436babc1b5d680484d645cdcR477) you are assigning in both cases?

content-src/selectors/siteMetadataSelectors.js
```diff
@@ -8,7 +8,12 @@ function selectSiteProperties(site) {
...
+  const eTLDExtra = eTLDLength > 0 ? -eTLDLength - 1 : Infinity;
```
piatra: Can you put parenthesis around the subtraction? I think it would help readability since there's already the shorthand if statement there.

content-test/selectors/siteMetadataSelectors.test.js
```diff
@@ -82,10 +82,15 @@ describe("siteMetadataSelectors", () => {
 
       assert.ok(result.parsedUrl);
     });
-    it("should set label prop", () => {
+    it("should set label prop without eTLD set", () => {
       let result = selectSiteProperties(validSpotlightSite);
 
-      assert.ok(result.label);
+      assert.equal(result.label, "cnn.com");
```
piatra: Shouldn't this be `validSpotlightSite.hostname` instead of the actual string?

content-src/selectors/siteMetadataSelectors.js
Mardak: The `|| ""` is to handle sites that don't have `eTLD` set

content-src/selectors/siteMetadataSelectors.js
Mardak: Sure. I'll make it `-(eTLDLength + 1)` to be more clear that it's 1 more.

content-test/selectors/siteMetadataSelectors.test.js
Mardak: Sounds reasonable.

Mardak: Well, turns out .hostname isn't set on the site.

content-src/selectors/siteMetadataSelectors.js
Mardak: Oh clarified that the `||` is to not require some other site-like thing using this selector to need `eTLD` as well.

content-test/selectors/siteMetadataSelectors.test.js
Mardak: So there's `result.parsedUrl.hostname` but that actually returns "www.cnn.com". Any reason this shouldn't be a plain string comparison for the test?

piatra: I know it's cosmetically the same thing but if I want to change `validSpotlightSite` to use mockData (like it should because APIs change and object names don't stay the same) I would also have to change your test.

piatra: Also looking at the test file no assert are being done against a string constant.

Mardak: An adjacent test `colorSelectors.test.js` checks things like `assert.equal(state.label, "foo");` and `assert.equal(state.backgroundColor, "rgba(11, 11, 11, 1)");`.

Changing test data requiring changes to tests doesn't seem that unexpected.

content-src/selectors/siteMetadataSelectors.js
```diff
@@ -8,7 +8,12 @@ function selectSiteProperties(site) {
   const metadataFavicon = site.favicons && site.favicons[0] && site.favicons[0].url;
   const favicon = site.favicon_url || metadataFavicon || site.favicon;
   const parsedUrl = site.parsedUrl || urlParse(site.url || "");
-  const label = prettyUrl(parsedUrl.hostname);
+
+  // Remove the eTLD (e.g., com, net) and the preceding period from the hostname
+  const eTLDLength = (site.eTLD || "").length;
+  const eTLDExtra = eTLDLength > 0 ? -(eTLDLength + 1) : Infinity;
+  const label = prettyUrl(parsedUrl.hostname).slice(0, eTLDExtra);
```
pdehaan: Will `parsedUrl.hostname` work here? It seems like we _may_ possibly fall back to "" above on L10 above.

piatra: That might be right but I don't see how an entry could be missing a URL.
Oddly enough [we have a rule](https://github.com/mozilla/activity-stream/blob/master/addon/PlacesProvider.js#L47) protecting us from this when we make the query.

piatra: Even so `urlParse("").hostname` will still work.

PR #1390, opened 2016-09-23
fix(addon): Remove the frame-src definition as it is now obsolete.

Standard8:
According to [devmo](https://developer.mozilla.org/docs/Web/Security/CSP/CSP_policy_directives) and the [implementation bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1045891), `child-src` was introduced in Firefox 45.

Currently the CSP lists both `child-src` and `frame-src`, however, as activity stream's minimum version is FF 45, all that needs to be listed is `child-src`.

This will also avoid the obsolete warning that is seen on the console when a new tab is opened.

### DIFF
--- a/bin/generate-html.js
+++ b/bin/generate-html.js
@@ -9,7 +9,7 @@ const defaults = {
 function template(rawOptions) {
   const options = Object.assign({}, defaults, rawOptions || {});
   const csp = options.csp === "on" ?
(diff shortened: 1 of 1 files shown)


rlr:
weird, this had been fixed in a56f196d863f5aabcf5d15a7fe06524f92e94022. it somehow came back. R+!

PR #1411, opened 2016-09-28
feat(metadata) Closes #1410 Create a metadata parser

### LINKED ISSUE #1410: Create a metadata parser
Step 1 in #1017

sarracini:
* Creates a MetadataParser module (not actually used anywhere yet)
    * Receieves a DOM, parses it
    * Gets metdata for that DOM (from remote service)
    * Formats it correctly for the metadataStore to insert it correctly 
* Module calls into remote page-metadata-parser service
* Import the page-metadata-parser package
* Tests for MetadataParser
* Adds relevant shims for MetadataParser

### DIFF
--- a/addon/MetadataParser.js
+++ b/addon/MetadataParser.js
@@ -1,0 +1,47 @@
+/* globals require, exports */
+"use strict";
+
+const {getMetadata} = require("common/vendor")("PageMetadataParser");
+const {Cc, Ci} = require("chrome");
+
+function MetadataParser() {
+}
+
+MetadataParser.prototype = {
+  /**
+   * Parse the outerHTML
+   */
+  _getDocumentObject(text) {
+    const parser = Cc["@mozilla.org/xmlextras/domparser;1"]
+                  .createInstance(Ci.nsIDOMParser);
+    return parser.parseFromString(text, "text/html");
+  },
+
+  /**
+   * Format the data so the metadata DB receives it
+   */
+  _formatData(data, url) {
+    const images = data.image_url ? [{url: data.image_url, height: 500, width: 500}] : [];
+    const formattedData = {
+      url,
+      images,
+      original_url: data.original_url,
+      title: data.title,
+      description: data.description,
+      favicon_url: data.icon_url
+    };
+    return formattedData;
+  },
+
+  /**
+   * Parse HTML, get the metadata from it, and format it
+   */
+  parseHTMLText(raw, url) {
+    return new Promise(resolve => {
+      const doc = this._getDocumentObject(raw);
+      resolve(this._formatData(getMetadata(doc, url), url));
+    });
+  }
+};
+
+exports.MetadataParser = MetadataParser;
--- a/content-test/addon/MetadataParser.test.js
+++ b/content-test/addon/MetadataParser.test.js
@@ -1,0 +1,77 @@
+/* globals assert, require */
+"use strict";
+const createMetadataParser = require("inject!addon/MetadataParser");
+const TEST_TITLE = "Test Page";
+const TEST_ICON = "https://www.foo.com/icon.png";
+const TEST_IMAGE = "https://www.foo.com/image.png";
+const TEST_DESCRIPTION = "This is a description";
+const TEST_HTML = `<html>
+  <head>
+    <title>${TEST_TITLE}</title>
+    <link rel="shortcut icon" type="image/x-icon" href="${TEST_ICON}" />
+    <meta property="og:description" content="${TEST_DESCRIPTION}" />
+    <meta property="og:image" content="${TEST_IMAGE}" />
+  </head>
+ </html>`;
+
+describe("MetadataParser", () => {
+  const {MetadataParser} = createMetadataParser({
+    "common/vendor.bundle": {url: {resolve(url) {return url;}}},
+    "chrome": {
+      Cc: {"@mozilla.org/xmlextras/domparser;1": {createInstance: () => new DOMParser()}},
+      Ci: {nsIDOMParser: {}}
+    }
+  });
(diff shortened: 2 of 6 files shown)


content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,72 @@
...
+  let metadataParser;
+  function setup() {
+    const mockPreviewProvider = {
+      processLinks(link) {
+        return link.map(link => Object.assign({}, link, {cache_key: link.url, places_url: link.url}));
+      }
+    };
+    const {MetadataParser} = createMetadataParser({
+      "common/vendor.bundle": {
+        PageMetadataParser: {getMetadata(doc) {return doc;}},
```
jaredkerim: The return signature of getMetadata is not a document, it's an object with the following format:

https://github.com/mozilla/page-metadata-parser#usage

But you should only mock this out if you have to force its behaviour in some way.  If you're just testing the success case, just actually invoke it.  It's synchronous, stateless, and self contained, so you lose nothing by actually calling it.

addon/MetadataParser.js
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Format the data so the metadata DB receives it
+   */
+  _formatData(data, url) {
+    data.images = data.image_url ? [{url: data.image_url, height: 500, width: 500}] : [];
+    data.favicon_url = data.icon_url ? data.icon_url : "";
+    data.url = url;
+    delete data.image_url;
+    delete data.icon_url;
+    let link = this._previewProvider.processLinks([data]);
```
jaredkerim: Let's not do this here at all.  If we simply pre process the link into an object with the fields that preview provider will accept then we can remove our dependency on it here.

content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,72 @@
...
+      "icon_url": "//www.foo.com/icon.png",
+      "image_url": "//www.foo.com/image.png",
+      "title": "Some title",
+      "url": "https:///"
+    };
+    let formattedData = metadataParser._formatData(incomingData, originalUrl)[0];
+    // format the data so the MetadataStore can store it properly
+    assert.equal(formattedData.url, originalUrl);
+
+    // removed the unusable properties
```
jaredkerim: Define two objects, the one that goes into format, and the one we expect to come out, and then just do a deep equality check, if you can.

content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+    // added the fields from PreviewProvider's processLinks
+    assert.ok(formattedData.cache_key);
+    assert.ok(formattedData.places_url);
+  });
+
+  it("should parse HTML text", done => {
+    metadataParser.parseHTMLText(TEST_HTML, "https://foo.com").then(result => {
+      assert.ok(result);
+      assert.equal(result[0].title, "Test Page");
```
jaredkerim: If you can, define the expected output object and use a deep equality check.

package.json
```diff
@@ -15,6 +15,7 @@
     "moment": "2.15.1",
     "react": "15.3.2",
     "react-dom": "15.3.2",
+    "page-metadata-parser": "^0.4.1",
```
jaredkerim: Update this to 0.4.2

content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,72 @@
...
+  it("should format the data properly", () => {
+    let originalUrl = "https://www.foo.com";
+    let incomingData = {
+      "description": "Some description",
+      "icon_url": "//www.foo.com/icon.png",
+      "image_url": "//www.foo.com/image.png",
+      "title": "Some title",
+      "url": "https:///"
+    };
+    let formattedData = metadataParser._formatData(incomingData, originalUrl)[0];
```
jaredkerim: If we remove the dependency on processLinks then we don't get an array back (since we only expect a single entry anyways) so you can remove all the [0] indexing.

content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,72 @@
...
+    <link rel="shortcut icon" type="image/x-icon" href="//www.foo.com/icon.png" />
+    <meta property="og:image" content="//www.foo.com/image.png" />
+  </head>
+ </html>`;
+
+describe("MetadataParser", () => {
+  let metadataParser;
+  function setup() {
+    const mockPreviewProvider = {
+      processLinks(link) {
```
jaredkerim: Removing this becuase we can remove the dependency on preview provider altogether.

content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,72 @@
...
+        return link.map(link => Object.assign({}, link, {cache_key: link.url, places_url: link.url}));
```
jaredkerim: Scratch this.

content-test/addon/MetadataParser.test.js
```diff
@@ -0,0 +1,73 @@
...
+    assert.equal(formattedData.url, expectedFormattedData.url);
+    assert.deepEqual(formattedData.images, expectedFormattedData.images);
+    assert.equal(formattedData.title, expectedFormattedData.title);
+    assert.equal(formattedData.description, expectedFormattedData.description);
+    assert.equal(formattedData.favicon_url, expectedFormattedData.favicon_url);
+  });
+
+  it("should parse HTML text", () => {
+    const url = "https://foo.com";
+    const expectedResult = {
```
jaredkerim: Add a description to this just so we're testing all the fields we expect in the success case.

jaredkerim:
r+!

PR #1421, opened 2016-09-30
New highlights & style changes (closes #1190)

### LINKED ISSUE #1190: Update style of existing Highlights
This is for Highlights:
highlights

Invision is here: https://mozilla.invisionapp.com/share/TN84I7MB2#/screens/177705068

NOTE: needs update, waiting on invision

dmose:
This addresses style changes suggested in #1190 as well as implemented the new grid Highlights layout.  It'll be much easier to review commit by commit, otherwise the reversion of the SpotlightFeed commit will be interspersed with everything else.

My expectation is that we'll want to squash everything except the reversion before landing.

I've disabled Recent Activity on the New Tab page when weighted highlights is turned on, as that's the desired UX.

I've also disabled the empty box "blank sites" stuff in the spotlight because it looks quite strange.  Perhaps @bryanbell or @k88hudson has opinions on this?

Most of the changes can be seen using "npm run storybook" and looking at the various stories.  The one thing that this doesn't show is the NewTabPage overall changes.  For that, one needs to do "npm run firefox" after a build (for the currently default-on-dev-branch weightedHighlights setup), and after a different build with that experiment turned off in package.json for the normal unweighted highlights version.


dmose:
r? @sarracini 

I'll be happy to do as much of as this as you like synchronously over vidyo since the hope is to get it in before Tuesday's meeting.  If you're overloaded with stuff, please let me know and I can find another reviewer...

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -79,10 +83,12 @@ const NewTabPage = React.createClass({
...
 
           <section>
-            <Spotlight page={PAGE_NAME} showRating={props.Spotlight.metadataRating} sites={props.Spotlight.rows} />
+            <Spotlight page={PAGE_NAME} length={spotlightLength}
+              showRating={props.Spotlight.metadataRating}
+              sites={props.Spotlight.rows} />
           </section>
 
-          {this._renderRecentActivity()}
+          { this.renderRecentActivity() }
```
dmose: Turns out the JSX grammar wants an expression, not a statement, so I ended up with this:

```javascript
 { props.Spotlight.weightedHighlights ? null : this.renderRecentActivity() }
```

content-src/components/Spotlight/Spotlight.scss
```diff
@@ -55,10 +71,16 @@
...
+    color: $mid-light-grey;
+    line-height: 1.8;
+    font-size: 10px;
+  }
+
   .spotlight-title {
     margin: 0 0 $spotlight-title-margin-bottom;
     font-size: inherit;
-    line-height: inherit;
+    line-height: 1.3;
```
dmose: I'm going to spin this off to another bug, probably on Monday, for two reasons:

* this problem already exists on master
* this particular solution looks nicer, but has the disadvantage of cropping the more important info (the title) rather than the less important info (the snippet), so I suspect we're going to want some design review...

dmose:
OK, just chatted with @sarracini over IRC.  She agreed that it was ok to land with comments addressed, so I'm changing the label to R+ with nits, with her consent.

dmose:
I've just pulled latest master, and it's suddenly very flaky, which is making it very hard to test the enabled/disable stuff we talked about on Vidyo, @sarracini .  

My suspicion is that there is some flakiness around experiment overrides, possibly a race condition, but I can't get master to show any weightedHighlights at all with a new profile.  

I can't in good conscience merge something else into master complicating things even more.  I'll dig into the master lossage on Monday, and then come back to this.

dmose:
Late Friday, I pulled latest master, and it was suddenly very flaky, which made it very hard to test the enabled/disable stuff @sarracini and I talked about on Vidyo.  

My suspicion is that there is some flakiness around experiment overrides, possibly a race condition, but I can't get master to show any weightedHighlights at all with a new profile.  

I can't in good conscience merge something else into master complicating things even more.  I'll dig into the master lossage on Monday, and then come back to this.

dmose:
After a bunch of poking around and some help from @rlr, it seems exceedingly likely that the problem @sarracini saw on Vidyo was #1493 introduced on master some time ago.  I suspect that this patch changes the timing a bit to make the race easier to hit on fast machines, but I don't think that needs to block this landing.

PR #1422, opened 2016-09-30
chore(storybook): support building of storybooks for PRs by storybook.io

dmose:
Once this lands, we'll be able to get automatically built storybooks.  Click on the last "View Deployment" button in https://github.com/dmose/activity-stream/pull/6 to see an example.

r? @Mardak

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules
 npm-debug.log
 .DS_Store
+.static-storybook
 *.sw[po]
 *.xpi
 *.pyc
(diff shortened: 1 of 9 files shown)


Mardak:
What am I looking for in dmose#6? Clicking on either deployment results in `Internal server error [500]`

dmose:
@Mardak They both seem to be working for me now.  I suspect they had some sort of infrastructure issue that they've since fixed.

You should see something like this:

screen shot 2016-09-30 at 06 41 49

dmose:
So it turns out that at this moment, you need to log in to storybooks.io using your github account to avoid the error.  I've pinged their support about this.

Mardak:
Got things working by logging in with Google (so it only requests email address instead of read/write access to github repositories)

PR #1429, opened 2016-09-30
feat(metadata): Closes #1409 Create a page scraper

### LINKED ISSUE #1409: Create a page scraper
Step 2 of #1017 
+ tests

sarracini:
* Create a PageSraper module (not instantiated anywhere yet)
    * Makes use of the MetadataParser to get metadata for links locally within page 
    * Injects a frame script (not yet created) to get metadata for the page you're browsing
    * Has a blacklist for links NOT to get metadata for (can be added to)
    * Inserts metadata into the MetadataStore DB (and respectively to it's cache)
* Tests for PageScraper
* A function in PreviewProvider to check if the single page you're visiting already has metadata for it - in that case don't re-parse and re-compute metadata
* Tests for the addition of PreviewProvider's function

Fixes #1409

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -492,6 +492,22 @@ MetadataStore.prototype = {
   }),
 
   /**
+   * Check if the link exists in the database by checking if it's cache key exists
+   *
+   * @param {String} key a cache key
+   *
+   * Returns a promise with the array of the retrieved metadata record
+   */
+  asyncCacheKeyExists: Task.async(function*(key) {
+    try {
+      let metadataLink = yield this.asyncExecuteQuery(`SELECT 1 FROM page_metadata WHERE cache_key = '${key}'`);
+      return metadataLink.length > 0;
+    } catch (e) {
+      return false;
+    }
+  }),
+
+  /**
   * Enables the data expiry job. The database connection needs to
   * be established prior to calling this function. Once it's triggered,
   * any following calls will be ignored unless the user disables
--- a/addon/PageScraper.js
+++ b/addon/PageScraper.js
@@ -1,0 +1,96 @@
+/* globals Services, Task */
+"use strict";
+
+const {MetadataParser} = require("addon/MetadataParser");
+const {Cu} = require("chrome");
+const options = require("@loader/options");
+
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
+Cu.importGlobalProperties(["URL"]);
+
+const DEFAULT_OPTIONS = {
+  framescriptPath: new URL("data/page-scraper-content-script.js", options.prefixURI),
+  blacklist: ["about:", "localhost:", "resource://"]
+};
+
+function PageScraper(previewProvider, options = {}) {
+  this.options = Object.assign({}, DEFAULT_OPTIONS, options);
+  this._previewProvider = previewProvider;
+  this._metadataParser = new MetadataParser();
+}
+
+/**
+ * Receives raw HTML from a page and sends it to a service to be parsed and the
+ * metadata extracted, and then stores that in the metadata DB. It takes an instance
+ * of PreviewProvider, since all the link processing takes places in PreviewProvider
+ */
+
+PageScraper.prototype = {
+  /**
+   * Check if the link is already in the metadata database. If not, parse the
+   * HTML and insert it in the metadata database
+   */
+  _asyncParseAndSave: Task.async(function*(rawHTML, url) {
+    let link = yield this._previewProvider.asyncLinkExist(url);
+    if (!link) {
+      let metadata;
+      try {
+        metadata = yield this._metadataParser.parseHTMLText(rawHTML, url);
+      } catch (e) {
+        Cu.reportError(`MetadataParser failed to parse ${url}`);
+      }
+      if (this._shouldSaveMetadata(metadata)) {
+        this._insertMetadata(metadata);
+      }
+    }
+  }),
+
+  /**
+   * Insert the metadata in the metadata database, along with it's source.
+   */
+  _insertMetadata(metadata) {
+    this._previewProvider.processAndInsertMetadata(metadata, "Local");
+  },
+
+  /**
+   * If metadata has neither a title, nor a favicon_url we do not want to insert
+   * it into the metadata DB
+   */
+  _shouldSaveMetadata(metadata) {
+    return (metadata && !!metadata.title && !!metadata.favicon_url);
+  },
+
+  /**
+   * Ensure that the page doesn't belong to a blacklist by checking that the
+   * url is not a substring of a restricted set of knows urls that should not
+   * collect metadata
+   */
+  _blacklistFilter(url) {
+    return (this.options.blacklist.every(item => url.indexOf(item) === -1));
+  },
+
+  /**
+   * Initialize the Page Scraper
+   */
+  init() {
+    Services.mm.loadFrameScript(this.options.framescriptPath, true);
+    Services.mm.addMessageListener("page-scraper-message", message => {
+      let {text, url} = message.data.data;
+      if (message.data.type === "PAGE_HTML" && this._blacklistFilter(url)) {
+        this._asyncParseAndSave(text, url);
+      }
+    });
+  },
+
+  /**
+   * Uninitialize the Page Scraper
+   */
+  uninit() {
+    Services.mm.removeMessageListener("page-scraper-message", this);
+    Services.mm.removeDelayedFrameScript(this.options.framescriptPath);
+  }
+};
+
+exports.PageScraper = PageScraper;
--- a/addon/PreviewProvider.js
+++ b/addon/PreviewProvider.js
@@ -359,13 +359,32 @@ PreviewProvider.prototype = {
   },
 
   /**
-   * Do some pre-processing on the links before inserting them into the metadata
+   * Do some pre-processing on the link before inserting it into the metadata
    * DB and adding them to the metadata cache
    */
-  processAndInsertMetadata(links) {
-    const processedLinks = this._processLinks(links);
-    this.insertMetadata(processedLinks);
+  processAndInsertMetadata(link, metadataSource) {
+    const processedLink = this._processLinks([link]);
+    this.insertMetadata(processedLink, metadataSource);
   },
+
+  /**
+   * Check if a single link exists in the metadata DB
+   */
+  asyncLinkExist: Task.async(function*(url) {
+    let key = this._createCacheKey(url);
+    if (!key) {
+      return false;
+    }
+
+    // Hit the cache first and return true immediately if we have a hit
+    const metadataFromCache = MetadataCache.cache.get(key);
+    if (metadataFromCache) {
+      return true;
+    }
+
+    const linkExists = yield this._metadataStore.asyncCacheKeyExists(key);
+    return linkExists;
+  }),
 
   /**
    * Initialize Preview Provider
--- a/test/test-PageScraper.js
+++ b/test/test-PageScraper.js
@@ -1,0 +1,129 @@
+/* globals require, exports */
+
+"use strict";
+
+const {before, after} = require("sdk/test/utils");
+const {PageScraper} = require("addon/PageScraper");
+const DUMMY_PARSED_METADATA = {
+  metadata: "Some dummy metadata",
+  title: "Some dummy title",
+  favicon_url: "Some dummy favicon_url"
+};
+
+let gMetadataStore = [];
+let gPageScraper;
+let parseCallCount;
+
+const mockPreviewProvider = {
+  processLinks(link) {
+    return link.map(link => Object.assign({}, link, {cache_key: link.url, places_url: link.url}));
+  },
+  asyncLinkExist(url) {
+    let doesExists = false;
+    if (gMetadataStore[0]) {
+      gMetadataStore.forEach(item => {
+        if (item.url === url) {
+          doesExists = true;
+        }
+      });
+    }
+    return doesExists;
+  },
+  processAndInsertMetadata(metadata, source) {
+    const processedLink = this.processLinks([metadata]);
+    this.insertMetadata(processedLink, source);
+  },
+  insertMetadata(metadata, source) {
+    const linkToInsert = Object.assign({}, metadata[0], {metadata_source: source});
+    gMetadataStore.push(linkToInsert);
+  }
+};
+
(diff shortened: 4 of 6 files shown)


addon/PageScraper.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }),
+
+  /**
+   * Insert the metadata in the metadata database, along with it's source.
+   * Also add it to the metadata cache
+   */
+  _insertMetadata(metadata) {
+    if (Object.keys(metadata).length) {
+      this._previewProvider.processAndInsertMetadata([metadata], "Local");
```
Mardak: This looks a bit odd as usually caching something has a separate key to reference the thing added, but we're getting those bits from various parts of the metadata object, yeah? Are there any required values or do we expect the length check to fail here? I suppose are there any relevant things we should be checking here in `PageScraper` vs relying on checks within `PreviewProvider`?

addon/PageScraper.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  },
+
+  /**
+   * Ensure that the page doesn't belong to a black list
```
Mardak: Similar to earlier comment of `blackList` -> `blacklist`, this is a single word term of "blacklist"

addon/PageScraper.js
```diff
@@ -0,0 +1,77 @@
...
+   */
+  _blackListFiter(url) {
+    return (this.options.blackList.every(item => url.indexOf(item) === -1));
```
Mardak: This seems like it warrants a comment of what it's doing now as it's a pretty simple match that most likely wants to be improved upon eventually. In particular, any url that contains the word "localhost" will be filtered out, e.g., https://example.com/article-about-10-things-you-never-knew-about-localhost ;)

addon/PreviewProvider.js
```diff
@@ -359,15 +359,34 @@ PreviewProvider.prototype = {
...
 
   /**
-   * Do some pre-processing on the links before inserting them into the metadata
+   * Do some pre-processing on the link before inserting it into the metadata
    * DB and adding them to the metadata cache
    */
-  processAndInsertMetadata(links) {
-    const processedLinks = this._processLinks(links);
-    this.insertMetadata(processedLinks);
+  processAndInsertMetadata(link, metadataSource) {
```
Mardak: If this really is expecting only to take a single link, the caller shouldn't pass in an array of one item.

addon/PreviewProvider.js
```diff
@@ -359,15 +359,34 @@ PreviewProvider.prototype = {
...
+   */
+  asyncDoesSingleLinkExist: Task.async(function*(url) {
+    let link = this._processLinks([{url}])[0];
+    if (!link) {
+      return false;
+    }
+    // Hit the cache first
+    const key = link.cache_key;
+    const metadataFromCache = MetadataCache.cache.get(key);
+
```
Mardak: Short circuit with `return true` on success should flow better here while unindenting the cache miss case.

addon/PreviewProvider.js
```diff
@@ -359,15 +359,34 @@ PreviewProvider.prototype = {
...
+    if (!metadataFromCache) {
+      const linksMetadata = yield this._metadataStore.asyncExecuteQuery(`SELECT * FROM page_metadata WHERE cache_key = '${key}'`);
```
Mardak: Probably best to call `asyncGetMetadataByCacheKey` instead of reaching in to the internals of `MetadataStore` from `PreviewProvider`

addon/PreviewProvider.js
```diff
@@ -359,15 +359,34 @@ PreviewProvider.prototype = {
...
-    this.insertMetadata(processedLinks);
+  processAndInsertMetadata(link, metadataSource) {
+    const processedLink = this._processLinks(link);
+    this.insertMetadata(processedLink, metadataSource);
   },
 
   /**
+   * Check if a single link exists in the metadata DB
+   */
+  asyncDoesSingleLinkExist: Task.async(function*(url) {
```
Mardak: A slightly shorter name of `asyncSingleLinkExists` or probably just `asyncLinkExists` would do.

test/test-PageScraper.js
```diff
@@ -0,0 +1,95 @@
...
+
+  // try to insert some empty metadata and make sure that it didn't get inserted
+  gPageScraper._insertMetadata(noMetadata);
+  assert.equal(gMetadataStore[0], undefined, "we didn't insert the page with no metadata");
+
+  // insert some proper metadata and check that it correctly inserted
+  gPageScraper._insertMetadata(DUMMY_PARSED_METADATA);
+  assert.equal(gMetadataStore[0].metadata, DUMMY_PARSED_METADATA.metadata, "we did insert a page with metadata");
+};
+
```
Mardak: A basic test for the blacklist filtering out urls would be good.

test/test-PreviewProvider.js
```diff
@@ -50,6 +50,19 @@ const gMockMetadataStore = {
       });
     }
     return items;
+  },
+  asyncExecuteQuery(query) {
+    // extract key from the query
+    let key = query.split("=")[1].trim().replace(/'/g, "");
```
Mardak: I'm curious how you ended up on this path of directly querying as this test shim is quite fragile if someone was expecting to do some other query. Is the performance of `asyncGetMetadataByCacheKey` undesirable or some other issue?

addon/PageScraper.js
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Check if the link is already in the metadata database. If not, parse the
+   *  HTML and insert it in the metadata database
+   */
+  _asyncParseAndSave: Task.async(function*(raw, url) {
+    let link = yield this._previewProvider.asyncDoesSingleLinkExist(url);
+    if (!link) {
+      let metadata = yield this._metadataParser.parseHTMLText(raw, url);
+      this._insertMetadata(metadata);
+    }
```
sarracini: We should definitely think of an update strategy, as of now it gets inserted in the metadata DB the same as any entry would (like from embedly), and is assigned a TTL of 3 days. After that there's auto expiry that happens in the DB and the next time they visit this page it won't be found in the DB and will get the metadata again for that page (acts as an update)

addon/PageScraper.js
sarracini: good catch! how about a more specific mach for ```localhost```? like ```localhost:```

addon/PreviewProvider.js
sarracini: The reason it needs an array is because the function ```insertMetadata``` expects an array when inserting into metadataStore, and when adding to the metadataCache

addon/PreviewProvider.js
sarracini: ```asyncExecuteQuery``` was exposed so that we could make custom queries - I didn't want to use ```asyncGetMetadataByCacheKey``` because it expects an array of cache keys and does a WHERE IN query + a bunch of other work (like calling ```asynExecuteQuery``` twice to grab images and whatnot), which is unnecessary if you have just a single cache key look up and just need to check if it exists in the DB or not

test/test-PreviewProvider.js
sarracini: As i kind of mentioned above, ```asyncGetMetadataByCacheKey``` does a lot of extra work that we don't actually need here, whereas doing the query itself is quicker in this case. This line in this test (i know i'm not a fan either) is a work around to extract the key to shim it. Alternatively, to make this a little more robust, I can extract the query string and  construct it once as a default option, and then override it during test time.
What do you think?

addon/PageScraper.js
sarracini: The check here is because if there is no metadata returned for that page, the service returns an empty object. Since we can't be computing cache keys and stuff if there's no ```url``` property, we don't want to even be doing the step of processing and inserting
I'll double check with @jaredkerim, maybe it throws and a try-catch is a better option

addon/PreviewProvider.js
Mardak: I can see how it's not as efficient as it's doing a bunch of unnecessary work, but calling the existing method is more maintainable. This very well could be a performance critical section of code, and there are things we can do to optimize, but if it's not, then we'll have some quirky code exposure going on.

At the very least, the `SELECT` should be done from `MetadataStore.js` to encapsulate that logic instead of having it in `PreviewProvier` as well as the related tests in `test-PreviewProvider`. A simple fix could be to allow `asyncGetMetadataByCacheKey` to have an option of fetching images or not, so in our use here, it would only do `SELECT .. IN ..` Alternatively, a separate method for just `asyncCacheKeyExists` that takes a cache key is much easier to maintain and test than an arbitrary SQL querying API.

test/test-PreviewProvider.js
Mardak: Part of the reason why these test changes are odd is that the API exposes the expectation of handling any SQL query, so to be complete, the test here should actually test for `INSERT` and `SELECT .. JOIN .. GROUP BY .. ORDER BY ..` etc. If instead the API is that it takes a single string key and returns true/false; that's quite a bit easier to test.

addon/PreviewProvider.js
sarracini: I'll add a ```asyncCacheKeyExists``` function in metadataStore so ```PreviewProvider``` doesn't have to do SQL, sound good?

Mardak: Yup. Not sure if this actually helps optimize, but you could probably do `SELECT 1 FROM .. LIMIT 1` so it's not actually packaging up columns and can short circuit once it gets the one result. Although I would also think the LIMIT 1 wouldn't be necessary if there's already an index / primary key set for the column.

ncloudioj: +1. I actually saw this one too earlier, just forgot to drop a comment here.

addon/PreviewProvider.js
jaredkerim: I agree with @Mardak.  Because we are always taking in exactly one, take it in as a single entity and then wrap it in an array on line 366.

addon/MetadataStore.js
```diff
@@ -492,6 +492,22 @@ MetadataStore.prototype = {
   }),
 
   /**
+   * Check if the link exists in the database by checking if it's cache key exists
+   *
+   * @param {String} key a cache key
+   *
+   * Returns a promise with the array of the retrieved metadata record
+   */
+  asyncCacheKeyExists: Task.async(function*(key) {
```
jaredkerim: If this should function is meant to check that a record exists, it should only return a boolean.  Right now it looks like it returns an integer or an empty array, we should clear that up.

addon/PageScraper.js
```diff
@@ -0,0 +1,96 @@
...
+ * of PreviewProvider, since all the link processing takes places in PreviewProvider
+ */
+
+PageScraper.prototype = {
+  /**
+   * Check if the link is already in the metadata database. If not, parse the
+   * HTML and insert it in the metadata database
+   */
+  _asyncParseAndSave: Task.async(function*(rawHTML, url) {
+    let link = yield this._previewProvider.asyncLinkExist(url);
```
jaredkerim: Check with @ncloudioj about whether MetadataStore has checks against inserting duplicate rows (I'm sure it does) in which case we can skip this check altogether.

addon/PageScraper.js
```diff
@@ -0,0 +1,96 @@
...
+    if (!link) {
+      let metadata;
+      try {
+        metadata = yield this._metadataParser.parseHTMLText(rawHTML, url);
+      } catch (e) {
+        Cu.reportError(`MetadataParser failed to parse ${url}`);
```
jaredkerim: Make sure that logging actual URLs is only done locally (ie the browser console).

addon/PageScraper.js
```diff
@@ -0,0 +1,96 @@
...
+      }
+      if (this._shouldSaveMetadata(metadata)) {
```
jaredkerim: 👍

addon/PageScraper.js
```diff
@@ -0,0 +1,96 @@
...
+        this._insertMetadata(metadata);
+      }
+    }
+  }),
+
+  /**
+   * Insert the metadata in the metadata database, along with it's source.
+   */
+  _insertMetadata(metadata) {
+    this._previewProvider.processAndInsertMetadata([metadata], "Local");
```
jaredkerim: Don't send this in as an array.

addon/PageScraper.js
```diff
@@ -0,0 +1,96 @@
...
+  },
+
+  /**
+   * If metadata has neither a title, nor a favicon_url we do not want to insert
+   * it into the metadata DB
+   */
+  _shouldSaveMetadata(metadata) {
+    return (metadata && !!metadata.title && !!metadata.favicon_url);
```
jaredkerim: 👍

addon/PreviewProvider.js
```diff
@@ -359,15 +359,35 @@ PreviewProvider.prototype = {
...
+  processAndInsertMetadata(link, metadataSource) {
+    const processedLink = this._processLinks(link);
+    this.insertMetadata(processedLink, metadataSource);
   },
 
   /**
+   * Check if a single link exists in the metadata DB
+   */
+  asyncLinkExist: Task.async(function*(url) {
+    let link = this._processLinks([{url}])[0];
```
jaredkerim: Can't you call this._createCacheKey?

addon/PreviewProvider.js
```diff
@@ -359,15 +359,35 @@ PreviewProvider.prototype = {
...
-    this.insertMetadata(processedLinks);
+  processAndInsertMetadata(link, metadataSource) {
+    const processedLink = this._processLinks(link);
+    this.insertMetadata(processedLink, metadataSource);
   },
 
   /**
+   * Check if a single link exists in the metadata DB
+   */
+  asyncLinkExist: Task.async(function*(url) {
```
jaredkerim: I think we should drop this step and this logic here altogether and let metadata store deal with whether it should be written.

addon/PageScraper.js
sarracini: it checks, but it throws an error and i don't want to go through the trouble of actually parsing and calling the service and attempting to insert and all that stuff, if we don't need to. it's a simple check so i think it's fine to leave it in

addon/PreviewProvider.js
sarracini: indeed i can

addon/PreviewProvider.js
sarracini: this needs to stay because it checks for the extra caching layer that lives inside preview provider

addon/PageScraper.js
jaredkerim: Okay that sounds good! 👍

PR #1435, opened 2016-09-30
chore(performance): Only set state if differences were found

k88hudson:
This PR changes the page worker behaviour so that it only saves a new copy of the redux state if anything has changed.

### DIFF
--- a/addon/PageWorker.js
+++ b/addon/PageWorker.js
@@ -2,6 +2,7 @@ const {Page} = require("sdk/page-worker");
 const {data} = require("sdk/self");
 const {LOCAL_STORAGE_KEY} = require("common/constants");
 const {ADDON_TO_CONTENT} = require("common/event-constants");
+const watch = require("common/vendor")("redux-watch");
 
 class PageWorker {
   constructor({store} = {}) {
@@ -25,9 +26,13 @@ class PageWorker {
       contentScriptFile: data.url("content-bridge.js"),
       contentScriptWhen: "start"
     });
+
+    // Note: watch only calls the callback (this._onDispatch) if something on
+    // the state object actually changed
+    const w = watch(this._store.getState);
     // Note: According to the redux docs, calling .subscribe on a store
     // returns a function which will unsubscribe
-    this._unsubscribe = this._store.subscribe(this._onDispatch);
+    this._unsubscribe = this._store.subscribe(w(this._onDispatch));
   }
   destroy() {
     if (this._page) {
--- a/common/reducers/Filter.js
+++ b/common/reducers/Filter.js
@@ -11,7 +11,8 @@ module.exports = function Filter(prevState = INITIAL_STATE, action) {
   switch (action.type) {
     case am.type("NOTIFY_FILTER_QUERY"):
       state.query = action.data || "";
-      break;
+      return state;
+    default:
+      return prevState;
   }
-  return state;
 };
--- a/content-test/addon/PageWorker.test.js
+++ b/content-test/addon/PageWorker.test.js
@@ -4,11 +4,11 @@ const {ADDON_TO_CONTENT} = require("common/event-constants");
 const redux = require("redux");
 
 // Test reducer that adds action.data to a total
-function reducer(state = 0, action) {
-  return action.type === "add" ? (state + action.number) : state;
+function reducer(state = {number: 0}, action) {
+  return action.type === "add" ? ({number: state.number + action.number}) : state;
 }
 function createStore() {
-  return redux.createStore(reducer, 0);
+  return redux.createStore(reducer, {number: 0});
 }
 
 describe("PageWorker", () => {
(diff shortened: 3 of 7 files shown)


k88hudson:
@dmose tests are experiencing intermittent failures at the moment

dmose:
Travis seems to have failed because an intermittent that @ncloudioj has already picked up.  I restarted the travis build, so assuming it succeeds, you're all set...

dmose:
The only nit referred to by the label is CI tests passing.

k88hudson:
@dmose  I changed the strategy a little here, moving to strict equality to save on performance (`redux-watch` basically does `===` on each reducer). This also required a slight change in one of our reducers, which was accidentally creating new objects. I added a test to prevent that from happening in the future.

common/reducers/Filter.js
```diff
@@ -11,7 +11,8 @@ module.exports = function Filter(prevState = INITIAL_STATE, action) {
   switch (action.type) {
```
dmose: It seems like we could do even better here by only creating the state object inside the NOTIFY_FILTER_QUERY, so that it never gets created at all in the default cause, since it will be unused...

addon/PageWorker.js
```diff
@@ -27,7 +28,8 @@ class PageWorker {
     });
     // Note: According to the redux docs, calling .subscribe on a store
     // returns a function which will unsubscribe
-    this._unsubscribe = this._store.subscribe(this._onDispatch);
+    const w = watch(this._store.getState);
+    this._unsubscribe = this._store.subscribe(w(this._onDispatch));
```
dmose: I'm confused.  Since _onDispatch doesn't use any of the arguments that redux-watch has w pass, what is the point of adding redux-watch?

content-test/addon/PageWorker.test.js
```diff
@@ -39,19 +39,23 @@ describe("PageWorker", () => {
...
       assert.called(pageWorker._onDispatch);
     });
   });
   describe("_onDispatch", () => {
     it("should emit a message to Page instance when store dispatches an action", () => {
-      store.dispatch({type: "add", number: 3});
-      assert(store.getState(), 3);
       pageWorker.connect();
-      pageWorker._onDispatch();
+      store.dispatch({type: "add", number: 3});
```
dmose: Nice job removing the impl detail from the test!

k88hudson:
@dmose is this good to merge?

content-test/common/reducers/equality.test.js
```diff
@@ -0,0 +1,35 @@
...
+describe("Reducer equality", () => {
+  let store;
+  function dispatchRows(type, length = 10) {
+    store.dispatch(actions.Response(type, createRows({length})));
+  }
+  beforeEach(() => {
+    store = createStore({logger: false});
+  });
+  it("should return a reference", () => {
+    assert.equal(store.getState(), store.getState());
```
dmose: Should this be assert.strictEqual?

content-test/common/reducers/equality.test.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+  it("should dispatch rows", () => {
```
dmose: The title here suggests that the test wants to test the dispatchRows function above, which it obviously doesn't.  

It seems like this is really testing that TOP_FRECENT_SITES_RESPONSE maps to newState.TopSites.rows, but perhaps I'm misinterpreting.

I think something wants to change here; it _might_ just be the description of the test.

content-test/common/reducers/equality.test.js
```diff
@@ -0,0 +1,35 @@
+const createStore = require("common/create-store");
+const {actions} = require("common/action-manager");
+const {createRows} = require("test/faker");
+
+describe("Reducer equality", () => {
```
dmose: There's something incongruent about this title.  All of the "it" clauses are not specifically referring to "Reducer equality", I don't think.  Perhaps this doesn't matter; perhaps there's another phrasing that would improve readability.

addon/PageWorker.js
k88hudson: watch does an equality check that doesn't call the callback if `prevState === newState`

content-test/common/reducers/equality.test.js
k88hudson: in chai `equal` is strict equality as opposed to `deepEqual`

addon/PageWorker.js
k88hudson: I will add a comment

dmose: Great; thanks

content-test/common/reducers/equality.test.js
dmose: If I'm reading the reference at http://chaijs.com/api/assert/ correctly, .equal is non-strict equality, and .strictEqual is what we want here...

k88hudson:
@dmose ok, added some review fixes and filed https://github.com/mozilla/activity-stream/issues/1463 and https://github.com/mozilla/activity-stream/issues/1462

common/reducers/Filter.js
dmose: I belatedly remembered that you can't put a declaration inside a case block.  That said, you could declare it using let above the switch, and Object.assign inside the case block.  Totally a micro-optimization though, so it's entirely up to you whether you do it.  If you do so, I'd suggest adding a comment explaining why so that some future maintainer doesn't undo it.

dmose:
Looks good; thanks for the tweaks!  r=@dmose

PR #1444, opened 2016-10-03
chore(debug): #1443 Add state download and state size estimates

k88hudson:
This patch adds a small amount of text and a download button to the debug page (press the debug button at the bottom right corner)


ncloudioj:
Looks great. I think the test failure is a known permanent one, so r+

PR #1447, opened 2016-10-03
fix (content): #1372 stopPropagation of LinkMenuButton click

### LINKED ISSUE #1372: The context menu doesn't appear for the same item in Highlights, Top Sites or Recent Activity after sharing it
via https://bugzilla.mozilla.org/show_bug.cgi?id=1304742
<hr>

[Affected versions]:
- Firefox 45.0 and up
- Activity Stream 1.1.4-dev

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Latest Activity Stream add-on version (1.1.4-dev) was previously installed on a clean profile.

[Steps to reproduce]:
2. Hover one of the Highlights tiles and click the context menu button.
3. Share the page with any service.
4. Click the context menu button for the same tile.

[Expected result]:
- The context menu is opened.

[Actual result]:
- Nothing happens.

[Additional Notes]:
- The issue is reproducible for Top Sites and Recent Activity tiles and on the Timeline page.
- In order for the context menu to be displayed you must click somewhere else in the page.
- Attached a screen recording of the issue https://goo.gl/N3auxL.

rlr:
The issue was that `hideContext` was getting called and hiding the just shown context menu. This happens after sharing and also if you try to open a context menu while already having one open for another site. It's because `window.removeEventListener("click", this.hideContext);` wasn't called yet. There might be other ways to fix this?

@k88hudson r?

Fixes #1372

### DIFF
--- a/content-src/components/ContextMenu/ContextMenu.js
+++ b/content-src/components/ContextMenu/ContextMenu.js
@@ -4,17 +4,16 @@ const ContextMenu = React.createClass({
   componentWillMount() {
     this.hideContext = () => {
       this.props.onUpdate(false);
-      window.removeEventListener("click", this.hideContext);
     };
-  },
-  componentWillUnmount() {
-    window.removeEventListener("click", this.hideContext);
   },
   componentDidUpdate(prevProps) {
     if (this.props.visible && !prevProps.visible) {
       setTimeout(() => {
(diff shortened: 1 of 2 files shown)


rlr:
After playing a little more with it, I like this fix better than the `stopPropagation` one ^^

PR #1450, opened 2016-10-04
feat(metadata): Closes #1430 Init PageScraper and pref it off

### LINKED ISSUE #1430: Initialize the Page scraper and pref it off
Step #4 for #1017

sarracini:
* Creates an instance of PageScraper (pref'ed off)
* Creates a pref listener so that you can manually create an instance of PageScraper if you switch on ```extensions.@activity-streams.pageScraper``` to true.
* Adds the pref in ```package.json```
* Mocks out an instance of PageScraper for the tests

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -30,6 +30,7 @@ const {Recommender} = require("common/recommender/Recommender");
 const {PrefsProvider} = require("addon/PrefsProvider");
 const createStore = require("common/create-store");
 const PageWorker = require("addon/PageWorker");
+const {PageScraper} = require("addon/PageScraper");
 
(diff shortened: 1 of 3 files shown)


addon/ActivityStreams.js
```diff
@@ -103,6 +105,20 @@ function ActivityStreams(metadataStore, options = {}) {
 
   this._previewProvider = new PreviewProvider(this._tabTracker, metadataStore, this._experimentProvider);
 
+  this._pageScraper = null;
+
+  // Only create a PageScraper if we are in the experiment
+  if (this._experimentProvider.data.localMetadata) {
```
jaredkerim: Can we change this so that it inits based solely on whether the pref is enabled and then have a separate check on activity stream startup that checks the experiment provider and sets the pref?

PR #1455, opened 2016-10-05
feat(metrics): Add metrics to metadata insertion

sarracini:
* Ability to track insertions into the metadata DB
* Update tests to mock tab tracker

cc @ncloudioj to make sure this is sane


jaredkerim:
We can forgo this for now by comparing the metadata processing events with cache hits and misses.  If we need to more finely monitor the actual operation of the metadata store we can reconsider this.

ncloudioj:
Agreed. Or just adding some dedicated benchmarks for it. Anyway, let's revisit this once it becomes necessary.

PR #1468, opened 2016-10-06
tests(metadata): Closes #1467 Integration test for framescript

### LINKED ISSUE #1467: Integration tests for local metadata framescript
needs an integration test

sarracini:
* Integration test for the local metadata parser's framescript

### DIFF
--- a/test/resources/dummy-frameScript.html
+++ b/test/resources/dummy-frameScript.html
@@ -1,0 +1,8 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf8"/>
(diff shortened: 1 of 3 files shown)


test/resources/dummy-frameScript.html
```diff
@@ -0,0 +1,6 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
```
k88hudson: any reason this can't just be html 5?

sarracini: probably not - I'll change it

Issue #1393, opened 2016-09-23
Weighted Highlights Intermittent Failure


sarracini:
Noticed this:
![failure](https://cloud.githubusercontent.com/assets/7219526/18797015/ce80fe5e-8199-11e6-95f1-1e2c0338a39f.png)

Stack Trace:

```
FAILED TESTS:
  ActivtyStreams
    selectors
      selectHistory weightedHighlights pref is true
        ✖ should select WeightedHighlights when weightedHighlights pref is true
          Firefox 46.0.0 (Linux 0.0.0)
        expected 'http://buzzlie.com/' to equal 'http://www.msn.com/'
        AssertionError@node_modules/chai/chai.js:5553:18
        [3]</module.exports/Assertion.prototype.assert@node_modules/chai/chai.js:206:13
        [6]</module.exports/assert.equal@node_modules/chai/chai.js:2277:5
        @webpack:///content-test/selectors/selectors.test.js:188:8 <- content-test/index.js:136803:10
        @webpack:///content-test/selectors/selectors.test.js:187:6 <- content-test/index.js:136802:8
```

sarracini:
cc @piatra

piatra:
Interesting thing to notice here is the version used `Firefox 46.0.0` maybe JS execution is different for that version somehow and that's why I'm not seeing this locally.

jaredkerim:
Closing as unable to reproduce.  Please reopen if its reappears.

dmose:
Excitingly, I've started seeing this locally, twice so far this AM.

sarracini:
Weighted highlights graduated so the test in question is no longer relevant - closing this

PR #1474, opened 2016-10-06
feat(highlights): Fetch a minimum number of results if not enough recent links

### LINKED ISSUE #1275: Return a minimum number of results

Mardak:
Fixes #1275. Supersedes #1362.

Does the existing query (recent links) first then if there aren't enough results, do a subsequent query fetching enough links to reach the limit.

r?@ncloudioj

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -707,11 +707,10 @@ Links.prototype = {
 
   /**
    * Obtain a set of links to rank.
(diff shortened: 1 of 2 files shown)


ncloudioj:
Looks great. r+ after fixing the test failure in `selectors.test.js`.

Mardak:
The test failure seems to be independent of this change. See #1393.

PR #1483, opened 2016-10-07
fix(addon): Closes #1480 Create an init function for ActivityStreams

### LINKED ISSUE #1480: Make the ActivityStream constructor not have side effects
Seriously we should have an init function and not a giant constructor for activity stream. It'll help split out a bunch of logic that doesn't belong in a constructor and will help with testing most likely

sarracini:
Fix #1480.
* Creates an initialization function for Activity Stream
* Removes all side effects of the constructor

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -72,86 +72,17 @@ const HOME_PAGE_PREF = "browser.startup.homepage";
 function ActivityStreams(metadataStore, tabTracker, telemetrySender, options = {}) {
   this.options = Object.assign({}, DEFAULT_OPTIONS, options);
   EventEmitter.decorate(this);
-
-  if (!this.options.searchProvider) {
-    this._searchProvider = new SearchProvider();
-  } else {
-    this._searchProvider = this.options.searchProvider;
-  }
-  this._searchProvider.init();
-
-  this._store = createStore();
-
-  this._newTabURL = `${this.options.pageURL}#/`;
-
-  this._perfMeter = new PerfMeter(this.appURLs);
-
-  Services.prefs.setIntPref("places.favicons.optimizeToDimension", 64);
-
-  this._appURLHider = new AppURLHider(this.appURLs);
-
-  this._memoizer = new Memoizer();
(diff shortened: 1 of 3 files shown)


k88hudson:
I will take a look this afternoon

addon/ActivityStreams.js
```diff
@@ -227,6 +172,117 @@ ActivityStreams.prototype = {
...
+    this._tabTracker = new TabTracker(
+      this.appURLs,
+      this.options.clientID,
+      memoized,
+      experimentProviderID
+    );
+  },
+
+  _initializeExperimentProvider() {
+    this._experimentProvider = new ExperimentProvider(
```
k88hudson: You could actually initialize this in the constructor since has no side effects, and just call `_experimentProvider.init` directly.

addon/ActivityStreams.js
```diff
@@ -227,6 +172,117 @@ ActivityStreams.prototype = {
     });
   },
 
+  _setUpPageWorker(store) {
+    this._pageWorker = new PageWorker({store});
+    this._pageWorker.connect();
+    this._refreshAppState();
```
k88hudson: What do you think abt putting `buildAsyncCache` and `refreshAppState` in their own separate method? since they are both kind of related to initializing the data for the app

k88hudson: It feels kind of weird here

sarracini: sure - ```initializeAppData``` for a name?

k88hudson:
Hmm, tests are failing. Intermittents?

addon/ActivityStreams.js
```diff
@@ -218,6 +172,99 @@ ActivityStreams.prototype = {
     });
   },
 
+  _initializeAppData() {
```
k88hudson: 👍

sarracini:
oops that's my bad, deleted something i shouldn't have when i was rebasing. updated PR should be fixed now

k88hudson:
Ok, looks good!

Issue #1314, opened 2016-09-16
Race condition in ShareProvider


k88hudson:
There is a race condition in `ShareProvider.js` I noticed while working on performance improvements that is revealed by the page loading much quicker.

If you access the property `.socialProviders`, there is a chance it will return `undefined`. This will in turn cause the `providers` property on the reducer to be `undefined`, which will cause a fatal error.

jaredkerim:
Requires feedback from @rlr

rlr:
I looked into this a bit but I can't really find any code path that would cause `_shareProvider.socialProviders` to be undefined. @k88hudson do you remember how you triggered this? Unless I am missing something, it will always be an array as long as `_shareProviders` has been created (`new ShareProvider()`). And `_refreshAppState()` is always called after the share provider is created. What am I missing?

rlr:
fixed via pull #1492

PR #1492, opened 2016-10-10
fix (addon): #1314 remove ShareProvider race condition workaround

rlr:
~~I haven't been able to trigger the race condition manually or via our tests after removing the workaround. So I just removed the workaround. We'll see if travis is happy.~~

UPDATE: I figured it out. The issue was with the mock provider used for testing.

r? @k88hudson


k88hudson:
Aha! Thank you!!

PR #1527, opened 2016-10-11
fix(rating): Closes #1475 Remove rating system

### LINKED ISSUE #1475: get rid of Rating system
In the new highlight layout, we've stopped displaying the star ratings for each spotlight item, as they were adding unnecessary complexity.

After discussion with @sarracini, it seems like the Ratings component and friends exceeds the cost/benefit ratio, and we should get rid of it.  If folks disagree, now would be a great time to chime in.

sarracini:
Remove rating system and all the things that it does. Fix #1475.


dmose:
Looks good!  r=@dmose, with two questions:

* presumably we want to remove the ./content-src/static/img/glyph-rating-*, yes?
* is there anything in other repos in the data pipeline specific to Rating stuff that can be removed too?

sarracini:
@dmose removed them in latest PR, and there shouldn't be anything in other repos that need to be removed as far as I know. Thanks!

PR #1530, opened 2016-10-11
shield(addon): Closes #1482 Pull tabTracker out of ActivitySteam scope

### LINKED ISSUE #1482: Pull tabTracker out of scope of ActivityStream and into scope of main.js
In prep for shield and also because reasons

sarracini:
* Pull TabTracker out of ActivityStream scope
* Put it in the scope of ```main.js```
* Update tests to reflect this change

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -17,7 +17,6 @@ const {Memoizer} = require("addon/Memoizer");
 const {PlacesProvider} = require("addon/PlacesProvider");
 const {SearchProvider} = require("addon/SearchProvider");
 const {ShareProvider} = require("addon/ShareProvider");
-const {TabTracker} = require("addon/TabTracker");
 const {PreviewProvider} = require("addon/PreviewProvider");
 const {RecommendationProvider} = require("addon/RecommendationProvider");
 const {TelemetrySender} = require("addon/TelemetrySender");
@@ -72,7 +71,7 @@ const PLACES_CHANGES_EVENTS = [
 
 const HOME_PAGE_PREF = "browser.startup.homepage";
 
-function ActivityStreams(metadataStore, options = {}) {
+function ActivityStreams(metadataStore, tabTracker, options = {}) {
   this.options = Object.assign({}, DEFAULT_OPTIONS, options);
   EventEmitter.decorate(this);
 
@@ -105,12 +104,8 @@ function ActivityStreams(metadataStore, options = {}) {
   );
   this._experimentProvider.init();
 
-  this._tabTracker = new TabTracker(
-    this.appURLs,
-    options.clientID,
-    this._memoized,
-    this._experimentProvider.experimentId
-  );
+  this._tabTracker = tabTracker;
+  this._tabTracker.init(this.appURLs, this._memoized, this._experimentProvider.experimentId);
 
   this._previewProvider = new PreviewProvider(this._tabTracker, metadataStore, this._experimentProvider);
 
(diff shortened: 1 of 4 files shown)


addon/ActivityStreams.js
```diff
@@ -105,12 +104,8 @@ function ActivityStreams(metadataStore, options = {}) {
...
   this._experimentProvider.init();
 
-  this._tabTracker = new TabTracker(
-    this.appURLs,
-    options.clientID,
-    this._memoized,
-    this._experimentProvider.experimentId
-  );
+  this._tabTracker = tabTracker;
+  this._tabTracker.init(this.appURLs, this._memoized, this._experimentProvider.experimentId);
```
jaredkerim: We will need TabTracker to operate independently of the ActivityStream object for it to track metrics for the tiles cohort of the SHIELD study, which means it will need to be able to operate without ActivityStream being enabled.  If any of these fields are necessary for TabTracker to send metrics payloads we should move them out as well.

sarracini: when we create the shield study we can change tab tracker to not depend on the memoizer or the experiment provider but for now this is a fine change, it just needs to be available out of the scope of activity stream

addon/TabTracker.js
```diff
@@ -43,6 +40,12 @@ TabTracker.prototype = {
     return this._tabData;
   },
 
+  init(trackableURLs, placesQueries, experimentId) {
```
Mardak: Do we expect to always have `trackableURLs` and sometimes the others (in the context of tracking existing tiles)? If so, it might be better to place the optional stuff in an options object to avoid worrying about ordering of optional parameters. We can do this in a followup once we're more certain about which pings we want for sure guiding us to what parameters are required. Please file and link back to this issue.

addon/ActivityStreams.js
```diff
@@ -17,7 +17,6 @@ const {Memoizer} = require("addon/Memoizer");
 const {PlacesProvider} = require("addon/PlacesProvider");
 const {SearchProvider} = require("addon/SearchProvider");
 const {ShareProvider} = require("addon/ShareProvider");
-const {TabTracker} = require("addon/TabTracker");
 const {PreviewProvider} = require("addon/PreviewProvider");
 const {RecommendationProvider} = require("addon/RecommendationProvider");
 const {TelemetrySender} = require("addon/TelemetrySender");
```
Mardak: I was taking a look at how `TabTracker` was actually sending data, and it looks like it sends notifications to `TelemetrySender`? Can you file a followup issue linking this or additionally move `TelemetrySender` init/`uninit` to `main.js`?

addon/ActivityStreams.js
Mardak: This is fine for here. I would assume we'll call `tabTracker.init` with appropriate arguments in the path that we don't initialize `ActivityStream`, so we'll want to do that in a followup issue. There is potential for oddness of passing around a partially-`init`-ed object as there's potential for accidentally multiply-`init`-ing. We could add some sanity checks, but probably not too important.

addon/TabTracker.js
sarracini: Yup I was thinking in the case for tiles we won't have both ```trackableURLs``` or ```placesQueries``` so we'll have to have a default for those. I'll file the ticket

addon/ActivityStreams.js
sarracini: Will do as a file up ticket! Good call

PR #1534, opened 2016-10-11
fix (addon) Remove timeline toolbar button

### LINKED ISSUE #1466: Remove Timeline Button
In order to remove the button from the menu bar in Firefox, we should remove the following from `ActivityStream.js`:

- the `_setupButton` function
- any references to `._button`

rlr:
Fixes #1466

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -772,9 +772,7 @@ ActivityStreams.prototype = {
       let baseUrl = this.options.pageURL;
       this._appURLs = [
         baseUrl,
-        `${baseUrl}#/`,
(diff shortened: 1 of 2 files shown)


k88hudson:
Can you add a second commit to this to remove `/timeline` and `/timeline/bookmarks` from `getAppURLs` actually? Since you are here anyway

rlr:
oh good point!

rlr:
@k88hudson should we remove the `RECENT_BOOKMARKS_REQUEST` and `RECENT_BOOKMARKS_RESPONSE` actions and related code too?

k88hudson:
@rlr there is actually a separate ticket for that

PR #1538, opened 2016-10-11
fix(highlights): Wait for async initializers before loading app data

### LINKED ISSUE #1382: highlights revert to default cards intermittently
Every once in a while, maybe 1x every 24hrs i see the default firefox cards (first run) even though I have been seeing history for multiple days.  

Opening another tab results in it being proper activity stream new tab view.

### LINKED ISSUE #1493: Weighted highlights don't display newly visited pages until second new tab
steps to reproduce:  on latest master
1. `npm run once`
2. open a new tab
3. click on the yahoo icon
4. command-click on 5 or so stories so that they open in separate tabs
5. open a new tab
6. notice that there are (usually!) still only the three initial highlights, instead of highlights including the newly created history. (bug)
7. open another new tab
8. notice that this one renders something that includes the new history (correct)

Mardak:
Fix #1382 and maybe fix #1493. On init when app states are initially cached, the recommender might not be loaded yet, so refresh things when it does get loaded. r?@piatra or r?@sarracini

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -92,6 +92,8 @@ ActivityStreams.prototype = {
   _isUnloaded: false,
 
   init() {
+    let initializePromises = [];
+
     this._store = createStore();
     this._initializePerfMeter();
     this._initializeAppURLHider();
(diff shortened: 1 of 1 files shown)


Mardak:
@sarracini just changed how init works in #1483. Any inputs for fixing this? To avoid the extra call to refresh, we would probably need to have `initializeBaselineRecommender` return a promise to conditionally call `initializeAppData` ?

piatra:
Yes you would need the loadRecommender to return a promise signaling it finished loading.

Mardak:
@piatra Added promises so no longer 1 line fix ;)

Mardak:
@sarracini if you want to review as you did the `init` refactoring

sarracini:
@Mardak yep i'll review this today!

PR #1545, opened 2016-10-12
Gh1503/rm react router

rlr:
Also removed the unused `Header` component.

What do you think of this approach for switching to debug view @k88hudson


package.json
```diff
@@ -19,7 +19,6 @@
     "react-dom": "15.3.2",
     "react-json-inspector": "7.0.0",
     "react-redux": "4.4.5",
-    "react-router": "2.8.1",
```
k88hudson: can you remove `history` as well?

content-src/components/Routes/Routes.js
```diff
@@ -1,31 +1,31 @@
...
-    </Router>);
+    let debugLinkText;
+    if (this.state.showDebugPage) {
+      debugLinkText = "newtab";
+    } else {
+      debugLinkText = "debug";
+    }
+
+    return (<div>
+      {this.state.showDebugPage &&
```
k88hudson: how bout `{this.state.showDebugPage? <DebugPage /> : <NewTabPage />}`

content-src/components/Routes/Routes.js
```diff
@@ -1,31 +1,31 @@
...
+  getInitialState() {return {showDebugPage: false};},
   render() {
-    return (<Router history={history}>
-      <Route path="/" component={require("components/Base/Base")}>
-        <IndexRoute title="Home" component={require("components/NewTabPage/NewTabPage")} />
-        <Route title="DebugPage" path="debug" component={require("components/DebugPage/DebugPage")} />
-      </Route>
-    </Router>);
+    let debugLinkText;
+    if (this.state.showDebugPage) {
```
k88hudson: maybe `const debugLinkText = this.state.showDebugPage ? "newtab" : "debug";`

content-src/components/Routes/Routes.js
```diff
@@ -1,31 +1,31 @@
 const React = require("react");
```
k88hudson: Can you add this stuff to the `Base` file instead and use that instead of `Routes`? That seems like it makes more sense

content-src/components/Routes/Routes.js
rlr: gah! didn't know that was possible. awesome

rlr:
OK, got those comments fixed

package.json
k88hudson: oh, right! sorry 😜

k88hudson:
Awesome! R+

PR #1549, opened 2016-10-13
remove(metadata): Closes #1507 Remove MetadataCache

### LINKED ISSUE #1507: Remove MetadataCache
Once the asynchronous regeneration of the Redux State object is merged, we will no longer need to cache MetadataStore calls using the MetadataCache and so we can remove this module.

sarracini:
* Removes MetadataCache
* Removes the tests for MetadataCache


ncloudioj:
R+ after cleaning up the deps.

PR #1550, opened 2016-10-13
Tooltip hints for Highlights, TopSites

k88hudson:
This patch adds:
- a `Hint` component (see comps here: https://mozilla.invisionapp.com/share/8F8V2GX3S#/screens)
- `Hint`s for Top Sites and Highlights
- a `Hints` reducer, to store (and persist) disabled hints 
- a button on the Debug page that can enable all of the hints again

To test it, please try:
- clicking on the "What's this?" text next to the Top Sites and Highlights titles on New Tab
- click on the blue button inside the hint (it should make the "What's this" text disappear)
- open another new tab, make sure "What's this" is still gone
- go to the debug page, and click the `Show all Tooltip Hints` button, and see if all the "What's this" reappear

### DIFF
--- a/content-src/components/Hint/Hint.js
+++ b/content-src/components/Hint/Hint.js
@@ -1,0 +1,68 @@
+const React = require("react");
+const {connect} = require("react-redux");
+const {DisableHint} = require("common/action-manager").actions;
+
+/**
+ * Hint - A component for helping onboard users to different features
+ *        It displays a small bit of text that, if clicked, shows a tooltip with
+ *        a description of the feature
+ */
+const Hint = React.createClass({
+  getDefaultProps() {
+    return {disabled: false};
+  },
+  getInitialState() {
+    return {active: false};
+  },
+  hide() {
+    this.setState({active: false});
+  },
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.active && !prevState.active) {
+      // Timeout needed or else this.hide will be triggered from onClickPrompt
+      setTimeout(() => {
+        window.addEventListener("click", this.hide, false);
+      }, 0);
+    }
+    if (!this.state.active && prevState.active) {
+      window.removeEventListener("click", this.hide);
+    }
+  },
+  componentWillUnmount() {
+    window.removeEventListener("click", this.hide);
+  },
+  onClickPrompt(e) {
+    e.preventDefault();
+    this.setState({active: !this.state.active});
+  },
+  onDisable(e) {
+    e.preventDefault();
+    e.stopPropagation();
+    this.setState({active: false});
+    this.props.dispatch(DisableHint(this.props.id));
+  },
+  render() {
+    // Note: refs are for testing hooks
+    return (<div className="hint-container" hidden={this.props.disabled} ref="container">
+      <span className="tooltip-tip" />
+      &ndash; <a className="prompt" href="#" onClick={this.onClickPrompt} ref="prompt"> What&rsquo;s this?</a>
+      <div className="explanation" hidden={!this.state.active} ref="explanation">
+        <h3 ref="title">{this.props.title}</h3>
+        <p ref="body">{this.props.body}</p>
+        <button className="close-button" onClick={this.onDisable} ref="closeButton">
+          <h4>Okay, got it!</h4>
+          <p>Don&rsquo;t show this tip again</p>
+        </button>
+      </div>
+    </div>);
+  }
+});
+
+Hint.propTypes = {
+  id: React.PropTypes.string.isRequired,
+  title: React.PropTypes.string.isRequired,
+  body: React.PropTypes.string.isRequired
+};
+
+module.exports = connect((state, props) => ({disabled: state.Hints[props.id] === false}))(Hint);
+module.exports.Hint = Hint;
--- a/content-src/components/Hint/Hint.scss
+++ b/content-src/components/Hint/Hint.scss
@@ -1,0 +1,92 @@
+.hint-container {
+  display: inline-block;
+  position: relative;
+  // This is needed to push the child component(i.e. the explanation above other stuff in the container)
+  z-index: $hint-z-index;
+
+  // This is the link that if clicked, opens the explanation
+  .prompt {
+    color: inherit;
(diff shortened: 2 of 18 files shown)


k88hudson:
@dmose could you review?

dmose:
It seems like a surprising interaction to me that the "What's this" thing simply disappears after it's been clicked once, with no obvious way to decide you made a mistake and dismissed too quickly.  That said, I don't think this necessarily needs to block this bug.

k88hudson:
@dmose yeah, that's a good point. @aaronrbenson what do you think?

dmose:
OK, I misremembered a bit, and see that this only happens after the "do not show this tooltip again".  I guess what really surprised me was that I feel like I mostly see "do not show again" on UI hints that are shown involuntarily, not ones that are shown after an explicit click.

dmose:
If I overlay the mock on top of the Storybook page using the Perfect Pixel add-on, the whitespace that is marked on the mock as "10px" doesn't appear to be match the mock.  That said, I don't really understand what the 10px measurement on the mockup is supposed to be measuring between, since there's whitespace above and below the length indicator.

aaronrbenson:
@dmose I see your point. The intention here is to allow those hints to be dismissed without creating extra 'settings' UI to turn hints on/off and the button text is an effort to explain what's going on. I think we roll with this for now and we can get some feedback during the shield study.

Re: measurements, I spoke with Kate and she did a little bit of adjusting. The measurement on the spec is to the elements container and not the fonts themselves.

dmose:
It looks to me like the button background doesn't quite match either the fill mentioned in text on the button background (#138CF6) nor the color that the mock actually uses.  @aaronrbenson , which of those two colors is desired?

dmose:
The button background appears to have a border-radius of 3px, but the spec suggests 4px.

content-src/components/Hint/Hint.scss
```diff
@@ -0,0 +1,90 @@
...
+    width: $hint-box-width;
+    border-radius: $hint-border-radius;
+    padding: $hint-padding;
+    background: $white;
+    position: absolute;
+    left: 100%;
+    margin-left: $hint-offset-left;
+    top: -$hint-offset-top;
+
+    &::before {
```
dmose: This wants a comment explaining what it's being used for.

content-src/components/Hint/Hint.scss
```diff
@@ -0,0 +1,90 @@
...
+      margin: 0 0 10px;
+    }
+
+    p {
+      color: $hint-body-color;
+      margin: 0 0 15px;
+    }
+
+    .close-button {
+      display: block;
```
dmose: Does the above line have any effect?  Disabling it in dev-tools doesn't seem to make an obvious visual difference....

content-src/components/Hint/Hint.scss
```diff
@@ -0,0 +1,90 @@
+.hint-container {
+  display: inline-block;
+  position: relative;
+  z-index: $hint-z-index;
```
dmose: Why put the z-index and positioning on the hint-container?  The prompt doesn't need that stuff, and I'm guessing it's really only the explanation that wants it.  If there is a good reason for this, I propose a comment...

content-test/common/reducers/Hints.test.js
```diff
@@ -0,0 +1,17 @@
+const Hints = require("common/reducers/Hints");
```
dmose: It might improve calling code readability if the object were named HintsReducer.

content-test/components/Hint.test.js
```diff
@@ -0,0 +1,133 @@
...
+    });
+    it("should render props.title", () => {
+      setup({id: "foo", title: "Everglade"});
+      assert.equal(instance.refs.title.innerHTML, "Everglade");
+    });
+    it("should render props.title", () => {
+      setup({id: "foo", title: "Everglade"});
+      assert.equal(instance.refs.title.innerHTML, "Everglade");
+    });
+    it("should render props.text", () => {
```
dmose: If I'm not misreading, this is the same test twice.

content-src/components/Hint/Hint.js
```diff
@@ -0,0 +1,62 @@
...
+    e.preventDefault();
+    this.setState({active: !this.state.active});
+  },
+  onDisable(e) {
+    e.preventDefault();
+    e.stopPropagation();
+    this.setState({active: false});
+    this.props.dispatch(DisableHint(this.props.id));
+  },
+  render() {
```
dmose: Might be worth adding a comment that the refs are only for testing.

content-test/components/Hint.test.js
```diff
@@ -0,0 +1,133 @@
...
+      setup({id: "foo", text: "Bloop"});
+      assert.equal(instance.refs.text.innerHTML, "Bloop");
+    });
+  });
+
+  describe("state", () => {
+    it("should show the explanation if state.active is true", () => {
```
dmose: Add a similar test for hiding the explanation?

content-test/components/Hint.test.js
```diff
@@ -0,0 +1,133 @@
...
+      assert.equal(instance.refs.explanation.hidden, false);
+    });
+    it("should set state.active = true if the prompt is clicked", () => {
+      assert.equal(instance.state.active, false);
+      TestUtils.Simulate.click(instance.refs.prompt);
+      assert.equal(instance.state.active, true);
+    });
+  });
+
+  describe("disable", () => {
```
dmose: Isn't the thing being tested here actually the close button?

content-test/components/Hint.test.js
```diff
@@ -0,0 +1,133 @@
...
+      TestUtils.Simulate.click(instance.refs.closeButton);
+      assert.equal(instance.state.active, false);
+    });
+    it("should dispatch a disable action if the close button is clicked", () => {
+      TestUtils.Simulate.click(instance.refs.closeButton);
+      assert.calledWith(dispatch, DisableHint("foo"));
+    });
+  });
+
+  describe("connected", () => {
```
dmose: Maybe "connected state"?

PR #1556, opened 2016-10-13
chore(lint): Remove no-vendor-prefix rule for sasslint

k88hudson:
I don't think this is a common issue we need to check for, and it's annoying seeing the warning every time.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -21,6 +21,7 @@ rules:
   no-ids: 0
   no-important: 0
   no-qualifying-elements: 0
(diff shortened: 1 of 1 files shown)


.sass-lint.yml
```diff
@@ -24,4 +24,5 @@ rules:
   no-warn: 1
   placeholder-in-extend: 2
   property-sort-order: 0
+  no-vendor-prefixes: 0
```
Mardak: nit: place below `no-warn` above `placeholder-in-extend`

Issue #1511, opened 2016-10-11
Persist active experiments in prefs


emtwo:
This will include the following steps:

1. Experiment provider starts up and loads experiments.json
2. It looks at each experiment and sees if a pref exists for that experiment. If not, create a pref, and this means we have new experiments
3. If there are experiments that a user is assigned to that are no longer active, unassign them.
4. If there are new experiments and the user is not assigned to an experiment, randomly decide if they should be assigned to an experiment

jaredkerim:
We need to verify that machine generated prefs are not wiped at update time if they're not defined in package.json

sarracini:
Closed with #1560

PR #1560, opened 2016-10-13
feature(addon): #1511 persist active experiments in prefs.

emtwo:


test/test-ExperimentProvider.js
```diff
@@ -1,14 +1,267 @@
 "use strict";
```
k88hudson: Why did you rewrite this as a `jpm` test?

emtwo: Mainly it's because the `data` object is now directly accessing prefs to look up values for experiments. So I needed to write a mock of prefs in order to keep the content-side tests. This just seemed to defeat the purpose of the tests to me, but I'm happy to put them back in content if there's a strong reason to.

k88hudson: Personally I think it's a good idea to move away from doing unit tests in `jpm` since they are fragile, hard to debug, etc., We do have a shim for `sdk/simple-prefs` already, is there a reason you need to use `sdk/preferences/service` instead?

emtwo: Let me know if you'd like me to move these back to content tests.

addon/ExperimentProvider.js
```diff
@@ -1,60 +1,101 @@
...
   }
 
   init() {
     this.setValues();
-    this._onPrefChange = () => this.setValues();
-    simplePrefs.on(OVERRIDE_PREF, this._onPrefChange);
+    this._onPrefChange = prefName => this.overrideExperimentPrefs(prefName);
+
+    Object.keys(this._experiments).forEach(experimentName => {
+      this._target.on(experimentName, this._onPrefChange);
```
k88hudson: Could we add addon namespacing to this, like simple prefs does? Setting the experiment name as-is could potentially overwrite stuff, no?

addon/ExperimentProvider.js
```diff
@@ -1,60 +1,101 @@
...
+    let inExperiment;
+
+    Object.keys(this._experiments).forEach(key => {
+      const experiment = this._experiments[key];
+      const {variant} = experiment;
+
+      // If the user is part of an inactive experiment,
+      // reset that experiment's pref.
+      if (prefService.get(key) && !experiment.active) {
+        prefService.set(key, false);
```
k88hudson: Is there another way to remove a pref other than set it to `false`? This seems unclear (no way to tell the difference between an inactive v.s. control group)

addon/ExperimentProvider.js
```diff
@@ -1,60 +1,101 @@
...
+      }
+
+      // If the user is already part of an active experiment, set the experiment id.
+      if (prefService.get(key) && experiment.active) {
```
k88hudson: In both if statements here, if the `variant` value of the experiment is falsey (e.g. `0`), this will incorrectly resolve to `false`; don't you want to check `undefined`?

addon/ExperimentProvider.js
```diff
@@ -1,60 +1,101 @@
...
+        throw new Error("Your variant cohort sizes should add up to less than 1.");
+      }
+
+      // If a user is in no experiments and there are new, active experiments,
+      // randomly assign them to a variant (or control)
+      if (experiment.active) {
         inExperiment = randomNumber >= floor && randomNumber < ceiling;
         if (inExperiment) {
           this._experimentId = variant.id;
+          prefService.set(key, true);
```
k88hudson: this should be the `variant` value, not `true`

addon/ExperimentProvider.js
```diff
@@ -69,8 +110,15 @@ exports.ExperimentProvider = class ExperimentProvider {
...
   destroy() {
-    simplePrefs.removeListener(OVERRIDE_PREF, this._onPrefChange);
-    this._data = {};
     this._experimentId = null;
+    Object.keys(this._experiments).forEach(experimentName => {
+      this._target.removeListener(experimentName, this._onPrefChange);
+    });
+  }
+
+  clearPrefs() {
```
k88hudson: just curious, why is this separate from `destroy`?

k88hudson: if for testability, maybe just make it an internal method and call it in `destroy`

k88hudson:
If the reason you couldn't use `simplePrefs` was because you need to reset, maybe you could just use `services` for that in separate file called `resetPrefs` or something? You can calculate the prefix the addon uses like this ( I found this in the source code for `simple-prefs`)

```
const {preferencesBranch} = require("./self");
const PREF_PREFIX = "extensions." + preferencesBranch + ".";
```

addon/ExperimentProvider.js
```diff
@@ -1,60 +1,101 @@
...
+const ss = require("sdk/simple-storage");
 
-exports.OVERRIDE_PREF = OVERRIDE_PREF;
 exports.ExperimentProvider = class ExperimentProvider {
-  constructor(clientID, experiments = require("../experiments.json"), rng) {
-    this._clientID = clientID;
+  constructor(experiments = require("../experiments.json"), rng) {
     this._experiments = experiments;
-    this._rng = rng || seedrandom(clientID);
+    this._rng = rng || Math.random;
```
emtwo: Thanks @ncloudioj, I'll remove those dependencies. @k88hudson yes, the prefs are persisted so we no longer a client to always have the same generated number

addon/ExperimentProvider.js
emtwo: If I understand correctly, `destroy()` is called on any `unload` reason, 2 of which may be `disable` or `uninstall`. Perhaps another may be browser shut down? https://github.com/mozilla/activity-stream/blob/master/addon/ActivityStreams.js#L834

In any case. These prefs need to be persisted in all cases besides disable/uninstall which is why I made this a separate function that is only triggered in those cases.

emtwo:
@k88hudson:
> we need to account for the experiment_id for people who have overridden one or more experiment prefs, which could potentially put them in more than one experiment cohort. We could either give them a null experiment id, or a special one indicating they are in an override state. The current behaviour is to give them a null experiment id.

Yes. The behaviour in this PR is also to set the experiment id to `null` when they override any of the prefs. A boolean in simple storage is also set to remember that someone has previously done a pref override and they should never be assigned to an experiment again: https://github.com/mozilla/activity-stream/pull/1560/commits/1becf9d33b0d58cdb437e3bf0a77f7bf45ee82f3#diff-339c2b8346504accdf9d1a33ad6f9d71R29

> there needs to be a way to listen for changes on experiments. if we can't use simple-prefs, is there another way we can add listeners?

Yes. Listening to experiment prefs is done here: https://github.com/mozilla/activity-stream/pull/1560/commits/1becf9d33b0d58cdb437e3bf0a77f7bf45ee82f3#diff-339c2b8346504accdf9d1a33ad6f9d71R20

addon/ExperimentProvider.js
k88hudson: Ah, ok, makes sense! Can you set it to the `control` value in that case instead of `false`?

emtwo: Yes! Will do that here and in the other spots you mentioned. Thanks!

addon/ExperimentProvider.js
emtwo: Ah, yes. I initially didn't realize that control/variant values could be anything, I assumed they could only be true or false. So yes it makes sense to check `undefined` here.

emtwo:
Made all the updates we talked about, @k88hudson

emtwo:
@k88hudson ready for round 3 now! :)

k88hudson:
Looks good! I filed https://github.com/mozilla/activity-stream/issues/1620 for that existing bug with tab tracker not updating

PR #1569, opened 2016-10-17
migration(metadata): Add a provider_name field in metadataStore

sarracini:
* Add provider_name to metadataStore


test/test-MetadataStoreMigration.js
```diff
@@ -1,6 +1,10 @@
 const {MetadataStore} = require("addon/MetadataStore.js");
 const {migrationV1Fixture, migrationV2Fixture} = require("./lib/MetastoreFixture.js");
 
+function retrieveNewColumn(columns) {
```
ncloudioj: nit: `retrieveColumnNames` might be a better name

Issue #1502, opened 2016-10-11
Remove Timeline-related Metrics and Actions


jaredkerim:
Remove all metrics handlers from TabTracker and associated data description files for all timeline related activity.

PR #1561, opened 2016-10-14
WIP feat(shield): Instrument tiles page for session and search/click-related events

Mardak:
Fix #1570. Don't merge as it has dummy logic to turn on instrumenting. This shield-specific stuff should probably should stay off of `master` anyway.

Implements the ping structure described in #1523.

Currently blocked on #1516 to determine `shield_variant` that'll be used to turn on instrumenter or activity stream.

PR #1575, opened 2016-10-17
chore(addon) #1502: Remove Timeline-related metrics and actions.

emtwo:


Mardak:
We're currently using the page type resolution for shield/about:newtab detection. https://github.com/mozilla/activity-stream/pull/1561/files#diff-aa995db3120476da0aa38895fb8a8a5bR3

I suppose this can still be removed and we'll just have an inline check in `setCommonProperties` to something like `url === "about:newtab" ? "TILES" : defaultPage`

emtwo:
> I suppose this can still be removed and we'll just have an inline check on the shield branch in setCommonProperties to something like url === "about:newtab" ? "TILES" : defaultPage

Oh! Thanks for pointing this out, @Mardak. I suppose if we're keeping shield in another branch and it only cares about `url === "about:newtab"` in one spot (`setCommonProperties`) then we can keep this PR.

PR #1581, opened 2016-10-18
shield(metrics): Closes #1516 Add 'shield_variant' to pings

### LINKED ISSUE #1516: Add a 'shield_variant' field to pings
Split from #1438. We'll use ClientID to track groups within a shield study, e.g., shield1A-<clientId> vs shield1B-<clientId>

sarracini:
* Add a ```shield_variant``` field to pings in order to track which variant we are in
* Give it a default value for non-shield users


sarracini:
cc @Mardak

PR #1585, opened 2016-10-19
Gh1504/remove handlers

### LINKED ISSUE #1504: Remove Unused placesRequest handlers
We will no longer need to request recent activity or bookmarks so we can remove the message handler for these calls.

rlr:
Fix #1504. I think this is all I can remove for now (until the Recent Activity component gets the 🔪 ). r? @k88hudson

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -369,11 +369,6 @@ ActivityStreams.prototype = {
    */
   _respondToPlacesRequests({msg, worker}) {
     switch (msg.type) {
-      case am.type("RECENT_LINKS_REQUEST"):
-        PlacesProvider.links.getRecentLinks(msg.data).then(links => {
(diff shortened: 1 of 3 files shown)


addon/ActivityStreams.js
```diff
@@ -369,11 +369,6 @@ ActivityStreams.prototype = {
    */
   _respondToPlacesRequests({msg, worker}) {
     switch (msg.type) {
-      case am.type("RECENT_LINKS_REQUEST"):
-        PlacesProvider.links.getRecentLinks(msg.data).then(links => {
-          this._processAndSendLinks(links, "RECENT_LINKS_RESPONSE", worker, msg.meta);
```
sarracini: i think recent links is the only one who uses ```_processAndSendLinks``` so you can probably remove that function altogether

rlr: yep, nice! 🔪

k88hudson:
Nice!

PR #1636, opened 2016-10-26
fix(content): Detect stabilization of state to reload from "Hang on tight"

### LINKED ISSUE #1593: "Hang on tight" should eventually resolve
Right now, if a user hits a "hang on tight" message, it never resolves.

We should update the page immediately when we move from an incomplete state to a complete state (this will only really happen when data is first initializing).

More info here: https://github.com/mozilla/activity-stream/issues/1493

Mardak:
Fix #1593. If the state for new tab page isn't ready yet, poll `localStorage` for stabilized changes to reload the page.

r?@sarracini

### DIFF
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -41,6 +41,27 @@ const NewTabPage = React.createClass({
     // Note that data may or may not be complete, depending on
     // the state of the master store
     this.props.dispatch(actions.NotifyPerf("NEWTAB_RENDER"));
+
+    // Check for stabilization of state changes if things aren't ready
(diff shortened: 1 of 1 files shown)


Mardak:
Will fix up test 
```
Firefox 46.0.0 (Linux 0.0.0) ERROR
  Some of your tests did a full page reload!
```

Mardak:
@sarracini can you take this review from @k88hudson ?

sarracini:
R+, with just one question regarding the interval

Mardak:
> is it necessary to clear the interval once ```this.props.isReady``` gets set to true?
It would be good to if the props would update. The current architecture doesn't notice `localStorage` changes, so that's why polling is needed.

PR #1644, opened 2016-10-26
feat(metadata): Closes #1571 Compute image size locally

### LINKED ISSUE #1571: Compute image sizes locally
During local metadata parsing we should use Firefox libraries to compute the sizes of images as we insert them into metadata store here:

https://github.com/mozilla/activity-stream/blob/master/addon/MetadataParser.js#L24

sarracini:
Fix #1571. Possible way to compute image sizes locally
Not sure if it's worth it? 
@Mardak can you give some feedback?

### DIFF
--- a/addon/MetadataParser.js
+++ b/addon/MetadataParser.js
@@ -21,7 +21,7 @@ MetadataParser.prototype = {
    * Format the data so the metadata DB receives it
    */
   _formatData(data, url) {
-    const images = data.image_url ? [{url: data.image_url, height: 500, width: 500}] : [];
+    const images = data.image_url ? [{url: data.image_url}] : [];
     const formattedData = {
       url,
       images,
--- a/addon/PageScraper.js
+++ b/addon/PageScraper.js
@@ -50,6 +50,7 @@ PageScraper.prototype = {
       }
 
       if (this._shouldSaveMetadata(metadata)) {
+        metadata.images = yield this._computeImageSize(metadata);
         this._insertMetadata(metadata);
         this._tabTracker.handlePerformanceEvent(event, "metadataParseSuccess", Date.now() - startTime);
       } else {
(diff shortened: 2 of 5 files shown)


addon/PageScraper.js
```diff
@@ -60,6 +64,26 @@ PageScraper.prototype = {
     }
   }),
 
+  _computeImageSize: Task.async(function*(metadata) {
+    this._setUpImageListener();
+    if (metadata.images[0]) {
+      Services.mm.broadcastAsyncMessage("get-image-size", {url: metadata.images[0].url});
```
Mardak: The `idl` seems to make `broadcastAsyncMessage` quite scary. https://dxr.mozilla.org/mozilla-central/source/dom/base/nsIMessageManager.idl#313-314 I would assume this broadcasts to all frame scripts, so that seems a bit wasteful.

You should be able to just use `new Services.appShell.hiddenDOMWindow.Image();` and use a regular `Promise` instead of the obsolete `defer` form. E.g.,

```js
compute(metadata) {
    if (images) {
      let image = new Image(); ...
      image.addEventListener("load", () => resolve(image.width));
    }
    else {
      resolve([]);
    }
  });
}
```

sarracini:
* Computes image sizes for a given metadata image url
* Update tests

r? @Mardak

test/test-PageScraper.js
```diff
@@ -109,14 +109,33 @@ exports.test_blacklist = function(assert) {
   assert.equal(isAccepted, false, `${url} exists in the blacklist`);
 };
 
+exports.test_compute_image_sizes = function*(assert) {
+  const imageWidth = 350;
+  const imageHeight = 150;
+  let metadataObj = {
+    url: "https://www.hasAnImage.com",
+    images: [{url: `http://placehold.it/${imageWidth}x${imageHeight}`}]
```
Mardak: Is this expecting to do a network request? Probably best to use an data uri, e.g., 1x1 pixel:

`data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEAAAAALAAAAAABAAEAAAI=`

addon/MetadataParser.js
```diff
@@ -21,7 +21,7 @@ MetadataParser.prototype = {
    * Format the data so the metadata DB receives it
    */
   _formatData(data, url) {
-    const images = data.image_url ? [{url: data.image_url, height: 500, width: 500}] : [];
+    const images = data.image_url ? [{url: data.image_url}] : [];
```
Mardak: Are we expecting some default sizes? Some images don't have explicit sizes, e.g., `svg`.

addon/PageScraper.js
```diff
@@ -61,6 +62,28 @@ PageScraper.prototype = {
   }),
 
   /**
+   * Locally compute the size of the preview image
+   */
+  _computeImageSize(metadata) {
+    return new Promise(resolve => {
+      if (metadata.images[0]) {
```
jaredkerim: Check here for the array length and resolve and return fast, otherwise carry on and alias the image like

const metadataImage = metadata.images[0];

and refer to that

addon/MetadataParser.js
sarracini: would it be sufficient to do  this:
```js

let imageWithSize = {
            url: image.src,
            width: image.width || 500,
            height: image.height || 500
          };
```

PR #1653, opened 2016-10-27
fix(highlights): Fixes #1553 Limit highlights to one item per domain

### LINKED ISSUE #1553: Limit highlights to one item per domain (weighted highlights)
I get duplicate highlights often on the new highlights view. 

Using version Activity Stream Dev 1.1.4-dev-1476371709

screen shot 2016-10-13 at 2 03 10 pm

sarracini:
- Dedupes the sites based on a key (in this case, domain)
- Update tests

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -172,11 +172,12 @@ class Baseline {
                     .sort(this.sortDescByScore)
                     .filter(entry => entry.score > 0);
 
-    // Decreases score for consecutive items from the same host.
-    results = this.dedupe(results);
+    // Decreases score for similar consecutive items and remove duplicates
+    results = this._adjustConsecutiveEntries(results);
+    let dedupedEntries = this._dedupeSites(results);
 
     // Sort again after adjusting score.
-    return results.sort(this.sortDescByScore).slice(0, INFINITE_SCROLL_THRESHOLD);
+    return dedupedEntries.sort(this.sortDescByScore).slice(0, INFINITE_SCROLL_THRESHOLD);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


common/recommender/Baseline.js
```diff
@@ -317,6 +318,24 @@ class Baseline {
...
     return entries;
   }
+
+  _dedupeSites(entries) {
+    let dedupedEntries = new Map();
+    entries.forEach(item => {
+      let key = this._createDedupeKey(item);
+      if (!dedupedEntries.get(key)) {
+        dedupedEntries.set(key, [item]);
+      }
```
jaredkerim: You forgot the case where the key exists and you want to append.

common/recommender/Baseline.js
```diff
@@ -317,6 +318,24 @@ class Baseline {
...
+    });
+
+    let results = [];
+    dedupedEntries.forEach(item => results.push(item[0]));
```
jaredkerim: You need to sort these by weight and take the first.

sarracini: they come in sorted order, so the first one we encounter will already have the greatest weight

k88hudson:
LGTM

common/recommender/Baseline.js
```diff
@@ -317,6 +318,24 @@ class Baseline {
 
     return entries;
   }
+
+  _dedupeSites(entries) {
```
k88hudson: you might want to rename to `dedupeSitesByHost` to be clearer

sarracini: I was thinking leaving it blank so that if we ever change the dedupe key we can still use the generic method

PR #1657, opened 2016-10-28
chore(metadata): Closes #1609 Unpack metadata fields onto site object

### LINKED ISSUE #1609: Explicitly unpack metadata fields on to site object
Rather than flattening the metadata object on to the site object in its entirety, let's explicitly transfer each field by name.  For example:

https://github.com/mozilla/activity-stream/blob/master/addon/PreviewProvider.js#L220

const cachedMetadata = existingLinks.get(site.cache_key);
enhancedSite.title = cachedMetadata.title;
enhancedSite.description = cachedMetadata.description;

etc

This accomplishes two things

1) It forces us to be explicit about which metadata fields we're using, and where they're being populated
2) If we want to add new fields or remove them it becomes more explicit
3) We can do additional post processing or validation here as well

sarracini:
Fix #1609 
* Explicitly unpacks the metadata fields that we care about onto a site object
* Update tests

@jaredkerim r?

### DIFF
--- a/addon/PreviewProvider.js
+++ b/addon/PreviewProvider.js
@@ -215,9 +215,20 @@ PreviewProvider.prototype = {
         return null;
       }
       let enhancedLink = Object.assign({}, site);
-      // Find the item in the map and return it if it exists
+
+      // Find the item in the map and return it if it exists, then unpack that
+      // object onto our new link
       if (existingLinks.has(site.cache_key)) {
-        Object.assign(enhancedLink, existingLinks.get(site.cache_key));
+        const cachedMetadataLink = existingLinks.get(site.cache_key);
+        enhancedLink.title = cachedMetadataLink.title;
+        enhancedLink.description = cachedMetadataLink.description;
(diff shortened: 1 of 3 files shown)


addon/PreviewProvider.js
```diff
@@ -215,9 +215,16 @@ PreviewProvider.prototype = {
...
+
+      // Find the item in the map and return it if it exists, then unpack that
+      // object onto our new link
       if (existingLinks.has(site.cache_key)) {
-        Object.assign(enhancedLink, existingLinks.get(site.cache_key));
+        const cachedMetadataLink = existingLinks.get(site.cache_key);
+        enhancedLink.description = cachedMetadataLink.description;
+        enhancedLink.provider_name = cachedMetadataLink.provider_name;
+        enhancedLink.images = cachedMetadataLink.images;
+        enhancedLink.favicons = cachedMetadataLink.favicons;
```
jaredkerim: Also add favicon_url

jaredkerim: And title

Issue #1611, opened 2016-10-21
Graduate Weighted Highlights?


k88hudson:
Once we have enough data, lets change weighted highlights from an experiment to a stable feature. We would need to do this in order to ship it for shield.

jaredkerim:
This depends on #1617

dmose:
If we make this a completely unflagged feature, and we discover a severe problem during post-Shield rollout, it'll be notably more work to temporarily roll this feature back.

It seems like it might be wiser to simply turn this into a flagged feature defaulted to "on" for a while.

Thoughts?

dmose:
#1666 has fixed the basic thing, so it no longer needs to be in La Mauricie, and I've moved it to Mount Revelstoke.

dmose:
What remains to be done is to go tear out all the old code paths for unweighted highlights, including the various XXX comments I left pointing to #1611.

Mardak:
Taking the remaining cleanup of this issue at the same time as #1518.

PR #1666, opened 2016-10-31
feature(Highlights): force on weightedHighlights for everyone

dmose:
This PR forces on the weightedHighlights for all activity-stream users.  This is the lowest-risk way to get it turned on for the Shield study.  After this lands, #1611 can be moved out of La Mauricie and into Mount Revelstoke, where another PR will tear out the non-weighted code entirely so that that issue can be closed.

The main things to test here is that 

npm run test
the XPI generated by 'npm run package'
npm run firefox

all default to the weighted highlights in a new profile.

r? @sarracini

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -5,7 +5,6 @@ const {configure} = require("@kadira/storybook");
 require("../data/content/main.css");
 
 function loadStories() {
-  require("../content-test/components/Spotlight-unweighted.story");
   require("../content-test/components/Spotlight.story");
   require("../content-test/components/ContextMenu.story");
   require("../content-test/components/Hint.story");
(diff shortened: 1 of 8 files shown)


emtwo:
Just a note, there is this thing: https://github.com/mozilla/activity-stream/pull/1602

Which means this will be ignored in shield.

sarracini:
Ah, good point @emtwo. So can we do this instead:
rename the pref in package.json from ```experiments.weightedHighlights``` to just ```weightedHighlights``` and then in ActivityStream.js here: https://github.com/mozilla/activity-stream/blob/master/addon/ActivityStreams.js#L188, check against the pref instead of the experiment provider.

What do you think @dmose?

PR #1669, opened 2016-10-31
Update cycle (still need to add comments)

### LINKED ISSUE #1493: Weighted highlights don't display newly visited pages until second new tab

k88hudson:
Fix #1493


dmose:
@k88hudson I'm confused; why was this closed?

PR #1671, opened 2016-10-31
fix(exceptions): Fixes #1667 add .catch to promises

### LINKED ISSUE #1667: Add .catch to promises for better exception handling
Go in and add .catch to our promises, where we can then do as simple as a Cu.reportError, so we can at least see the exceptions that are thrown

sarracini:
Adding some catch enforcement
So this patch only enforces the catching of promises - we should file another PR for logging it to a central place, once we decide what that is, so that we can essentially mimic what sentry does.

@k88hudson r?

### DIFF
--- a/addon/PreviewProvider.js
+++ b/addon/PreviewProvider.js
@@ -176,10 +176,10 @@ PreviewProvider.prototype = {
         }
         getColor(link.favicon, link.url).then(color => {
           resolve(Object.assign({}, link, {favicon_color: color}));
-        }, () => resolve(link));
+        }, () => resolve(link)).catch(err => Cu.reportError(err));
       })
       )
-    );
+    ).catch(err => Cu.reportError(err));
   },
 
   /**
@@ -290,7 +290,7 @@ PreviewProvider.prototype = {
     requestQueue.forEach(requestBundle => {
       promises.push(this._asyncFetchAndStore(requestBundle, event));
     });
-    yield Promise.all(promises);
+    yield Promise.all(promises).catch(err => Cu.reportError(err));
   }),
 
   /**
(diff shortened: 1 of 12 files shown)


addon/MetadataParser.js
```diff
@@ -41,7 +41,7 @@ MetadataParser.prototype = {
     return new Promise(resolve => {
       const doc = this._getDocumentObject(raw);
       resolve(this._formatData(getMetadata(doc, url), url));
-    });
+    }).catch(err => Cu.reportError(err));
```
sarracini: I think it's ok, both ```getMetadata``` and ```_formatData``` always return something, and if either of them throw an exception we'll be able to catch it here. So I *think* it's fine....

k88hudson:
Overall looks good, I will take a closer look as soon as I finish up some tests (I just want to make sure we aren't prematurely catching anything that needs a specific catch handler)

addon/MetadataParser.js
k88hudson: If you handle the error here, the error handling in `_asyncParseAndSave` in `PageScraper.js` will never get hit

addon/PreviewProvider.js
```diff
@@ -192,7 +192,7 @@ PreviewProvider.prototype = {
     }
 
     return this._asyncGetEnhancedLinks(processedLinks, event).then(
-      cachedLinks => this._getFaviconColors(cachedLinks));
+      cachedLinks => this._getFaviconColors(cachedLinks)).catch(err => Cu.reportError(err));
```
k88hudson: same thing here with the `catch` never getting hit

pdehaan: I think if you `.catch()` a promise error, it should then resume onto the next chained `.then()`, IIRC.

So in places where we return a promise, not sure if we'd want a catch + log + throw action, so the usage side can handle any errors as well.

Issue #1452, opened 2016-10-04
Only update top sites/highlights state when necessary


k88hudson:
Right now, we re-check on every new tab -- this isn't necessary. We should only update the value when something has changed in prefs etc.

sarracini:
Closed with #1677

PR #1677, opened 2016-10-31
feat(addon): #1452 Add 'Feeds' to allow for custom update cycles

k88hudson:
This patch:
- Adds `Feed` and `FeedController, which allow for different data sources to define (a) how they get data and (b) when they update
- Adds definitions for a `Highlights`, `TopSites`, and `History` feed
- Removes caching for all queries related to History, Highlights, and Top sites and related tests
- Removes the `processLinks`, `processAndDispatch` functions from `Activitystreams.js`
- Moves all weighted highlights related stuff, such as the recommendation engine initialization and logic, to the Highlights feed and out of `Activitystreams.js`
- Adds new actions for key app updates, such as initialization, new tab opening

I'm sorry it's really big... it was not a clear case of refactoring (there was a lot of investigation needed as I was doing it). Happy to walk through on Vidyo if necessary.


addon/ActivityStreams.js
```diff
@@ -142,24 +151,12 @@ ActivityStreams.prototype = {
    * Broadcast a message to all workers
    */
   broadcast(action) {
+    this._store.dispatch(action);
```
k88hudson: This change ensures that broadcast will dispatch even if no tabs are currently open.

PR #1686, opened 2016-11-01
fix (metadatastore): parameterize the images SELECT query

### LINKED ISSUE #1679: Audit all the SQL queries in metadata store
Let's take another look at those SQL queries in the `MetadataStore.jsm`. Like what we've found in #1577 , we'd like to avoid following risks:

* SQL query injections
* inefficient queries
* unnecessary data fetches

rlr:
r? @ncloudioj 

Fixes #1679

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -461,8 +461,11 @@ MetadataStore.prototype = {
            FROM page_metadata AS pm
               JOIN page_metadata_images AS pmi ON pm.id = pmi.metadata_id
               JOIN page_images AS pi ON pi.id = pmi.image_id
-           WHERE pm.id = ${metaObject.id}`,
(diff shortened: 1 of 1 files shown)


addon/MetadataStore.js
```diff
@@ -461,8 +461,11 @@ MetadataStore.prototype = {
            FROM page_metadata AS pm
               JOIN page_metadata_images AS pmi ON pm.id = pmi.metadata_id
               JOIN page_images AS pi ON pi.id = pmi.image_id
-           WHERE pm.id = ${metaObject.id}`,
-           {columns: ["url", "type", "height", "width", "color"]}
+           WHERE pm.id = :metaobject_id`,
```
ncloudioj: Ah, good kill. 👍

PR #1689, opened 2016-11-01
fix(package): Revert dependencies update causing build failures

Mardak:
Trying out a backout to see if things will go green. r?@ncloudioj


ncloudioj:
R+

PR #1691, opened 2016-11-01
chore(package): update various dependencies to latest version

Mardak:
Same as #1690 but pre-squashed. Trying to track down red on master.


Mardak:
Checks have passed in this pre-squashed version... I'll track things in #1690.

PR #1746, opened 2016-11-07
Rename WeightedHighlights to Highlights`

k88hudson:
This patch:

* Removes the `Highlights` reducer
* Renames `WeightedHighlights` reducer and related actions to just `Highlights`


k88hudson:
@dmose what do you think about this?

dmose:
@k88hudson Looks like a fine simplification; r=@dmose.

PR #1748, opened 2016-11-08
fix(contextmenu): Fix #1722 TopSites context menu not clickable

### LINKED ISSUE #1722: Dynamically added top sites context menu is not clickable
Using version: 1.1.6-dev

1) Open a new tab
2) Dismiss a top site
3) A new top site is added on the right to replace it
4) Can not click on the new top sites context menu

sarracini:
* Fixes the busted top sites context menu after a dismiss


k88hudson:
This has the interesting side effect of not closing existing open menus, but seems to be better than the alternative.

sarracini:
oh, i can fix that!

PR #1751, opened 2016-11-08
chore(package): update various dependencies to latest version

Mardak:
r?@sarracini


Mardak:
> why the move of ```.tooltip-container:hover``` though?

There was a `sasslint` warning of incorrectly putting nested stuff before other styles.

Issue #1768, opened 2016-11-09
Limit places db queries to prevent critical errors


jaredkerim:
I updated to the latest activity stream on dev, and found that my highlights were not updating.  When I started digging into browser consoles and toolboxes, I found that it was caught in an infinite loop of re requesting metadata, failing, and trying again.  

When I dug a little further, I found that the highlights feed refresh logic was waking up, querying places db for urls, finding 1014 candidate urls, passing those to preview provider, which breaks them into lots of 20 to send to the metadata service, makes 51 concurrent requests to metadata service, attempts to insert 1014 metadata entries into metadatastore, catastrophically fails, and then tries to do the whole thing over again.  

We should limit the number of URLs we receive from placesdb and attempt to send to preview provider.

I've logged some things here to help with debugging, but this caused my activity stream to become unresponsive and I was forced to uninstall.  For reference, I have both local metadata parsing and metadata service enabled.

![screenshot 2016-11-09 15 17 48](https://cloud.githubusercontent.com/assets/119884/20154031/88a0d8ae-a693-11e6-9ac1-5cb2acb0e5c1.png)

pdehaan:
One of the more interesting parts of that screenshot is that we aren't doing templating correctly:

> Failed to fetch metadata by cacheKey: ${e.message}

Looks like we have at least 3 potential violations in /addon/MetadataStore.js:

1. [/addon/MetadataStore.js:448](https://github.com/mozilla/activity-stream/blob/7b7436191542fec546ee3a37914812816ecf0285/addon/MetadataStore.js#L448) should be using backticks for templated strings:
    ```js
Cu.reportError("Failed to fetch metadata by cacheKey: ${e.message}");
```

1. [/addon/MetadataStore.js:471](https://github.com/mozilla/activity-stream/blob/7b7436191542fec546ee3a37914812816ecf0285/addon/MetadataStore.js#L471) should be using backticks for templated strings:
    ```js
Cu.reportError("Failed to fetch metadata by cacheKey: ${e.message}");
```

1. [/addon/MetadataStore.js:492](https://github.com/mozilla/activity-stream/blob/7b7436191542fec546ee3a37914812816ecf0285/addon/MetadataStore.js#L492) should be using backticks for templated strings, and is missing a `$` in `${image.type}`:
    ```js
throw new Error("Fetched unknown image types: {image.type}");
```

pdehaan:
[...]

PR #1770, opened 2016-11-09
chore(addon): Fix invalid template strings in addon/MetadataStore.js

pdehaan:
Ref: https://github.com/mozilla/activity-stream/issues/1768#issuecomment-259546442

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -445,7 +445,7 @@ MetadataStore.prototype = {
       );
     }
     catch (e) {
-      Cu.reportError("Failed to fetch metadata by cacheKey: ${e.message}");
+      Cu.reportError(`Failed to fetch metadata by cacheKey: ${e.message}`);
       throw e;
     }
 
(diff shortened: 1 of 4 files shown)


ncloudioj:
Nice catch! 👀

addon/MetadataStore.js
```diff
@@ -489,7 +489,7 @@ MetadataStore.prototype = {
             });
             break;
           default:
-            throw new Error("Fetched unknown image types: {image.type}");
+            throw new Error(`Fetched unknown image types: ${image.type}`);
```
pdehaan: Should this have a `Cu.reportError()` as well, or is that getting logged elsewhere?

Also does anybody know where these `Cu.reportError()` get logged so we can audit those errors for other curious looking strings?

ncloudioj: No. This is more like an "assert" here, i.e. this should never happen. If it does happen, it'll let the caller handle this exception.

PR #1776, opened 2016-11-10
feat(feeds): #1775 Add Search feed

### LINKED ISSUE #1775: Create Search feed
This is to replace the initialization of search UI strings and engines in `refreshAppState`

k88hudson:
Fix #1775. This patch replaces the initialization and updating of search engine information and strings with a new `Feed`.

It also looks for changes in the search engines on every new tab – ideally, it would be good to listen on an event for that, but I haven't found one yet 😢 

There's definitely some refactoring work that could be done here, but I wanted to try to stick to just moving stuff mostly for this patch


Mardak:
> ideally, it would be good to listen on an event for that, but I haven't found one yet

There's a global observable "engine-current" notification that is sent when the search engine changes:
https://dxr.mozilla.org/mozilla-central/source/browser/components/newtab/NewTabSearchProvider.jsm#29-37

k88hudson:
@Mardak isn't that just for the current engine though, not the list of all search engines in the list?

Mardak:
There's a list of related "browser-search-engine-modified" topics including "engine-added":
https://dxr.mozilla.org/mozilla-central/source/netwerk/base/nsIBrowserSearchService.idl#493-507

k88hudson:
@Mardak Aha! Thank you, that's what I was looking for!

k88hudson:
@Mardak ok, I added the search engine events to `SearchProvider` and removed the `ON_NEW_TAB` event. Weirdly, the change event still doesn't fire for hiding/showing search engines for me, but everything else seems to be fine

addon/SearchProvider.js
```diff
@@ -3,7 +3,7 @@ SearchSuggestionController, PrivateBrowsingUtils, exports, require */
 
 "use strict";
 const {Ci, Cu} = require("chrome");
-const CURRENT_ENGINE = "browser-search-engine-modified";
+const SEARCH_ENGINE_TOPIC = "browser-search-engine-modified";
 const HIDDEN_ENGINES = "browser.search.hiddenOneOffs";
```
Mardak: As discussed on IRC, we'll want to watch this pref for changes. Probably just emit `SEARCH_ENGINE_TOPIC`.

https://dxr.mozilla.org/mozilla-central/source/browser/components/preferences/in-content/search.js#292-300

PR #1778, opened 2016-11-10
fix(sqlite): Fixes #1768 Limit MetadataStore SQLite parameters

### LINKED ISSUE #1768: Limit places db queries to prevent critical errors

sarracini:
* Truncate the SQL parameters to the defined limit
@ncloudioj r?

Needs to be uplifted to 1.1.7

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -86,6 +86,9 @@ const IMAGE_TYPES = {
   "preview": 3
 };
 
+/* Maxium number of allowed parameters in a single SQL statement (defined here: https://www.sqlite.org/limits.html) */
+const SQL_PARAMETER_LIMIT = 999;
+
 function MetadataStore(path, migrations = null) {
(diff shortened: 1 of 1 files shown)


ncloudioj:
R+! @sarracini

PR #1781, opened 2016-11-10
logging(exceptions): Closes #1779 Do not report db insertion collisions

### LINKED ISSUE #1779: Do not report cache key insertion collisions as errors
When we attempt to insert a row into the metadata store, if we already have a row for that URL we will get a unique key constraint violation, which is expected.  We are currently logging this to the browser console as an error, which is causing people to think that something is not behaving correctly when this is an expected result.  We should stop logging this to the console.

sarracini:
* Doesn't report unique constraint failure as error
* Update test

@jaredkerim r?


test/test-MetadataStore.js
```diff
@@ -125,13 +125,8 @@ exports.test_insert_twice = function*(assert) {
...
   let items = yield gMetadataStore.asyncExecuteQuery("SELECT * FROM page_metadata");
   assert.equal(items.length, 1, "Page is inserted");
-  let error = false;
-  try {
-    yield gMetadataStore.asyncInsert([metadata]);
-  } catch (e) {
-    error = true;
-  }
-  assert.ok(error, "Re-insert the same page should be rejected!");
+  yield gMetadataStore.asyncInsert([metadata]);
```
jaredkerim: So we expect this to not throw errors and just fail silently as an expected silent failure case.  I like it. 

Can we also add a test that checks that a different kind of exception does still throw an error?

Issue #1783, opened 2016-11-10
Move selectors to common/


k88hudson:
This is so addon aside can use them if needed

PR #1784, opened 2016-11-10
chore(selectors): #1783 Move selectors to common/

k88hudson:
This patch moves selectors to common, so that they are available to the addon side if needed.


k88hudson:
@dmose R?

k88hudson:
@dmose I cant actually figure out what is causing the coverage change... arg... maybe it's just a change in the ratio of lines? or something? O_o

k88hudson:
@dmose Ok, so the problem is in fact that `common/` is not part of the coverage reporting. I will file a follow-up to address this

PR #1793, opened 2016-11-15
feat(Highlights): Closes #1767; show highlights after welcome popup w…

### LINKED ISSUE #1767: when welcome popup no longer necesarry, page should not reload
This is a pre-requisite for the originally specced, less jarring UI being tracked in #1675: making stuff fill-in behind the dialog as it loads.

dmose:
…/o reload.  

To test:

1. After building, do ```npm run firefox``` on a slow machine.  
2. Once the main window opens, quickly hit command-T to open a new tab.
3. Note the "Welcome..." Firefox dialog
4. After a few seconds, the dialog should go away and the rest of the page shoulder render without an actual reload happening

r=k88hudson?

### DIFF
--- a/common/create-store.js
+++ b/common/create-store.js
@@ -6,6 +6,20 @@ const {Channel} = require("common/ReduxChannel");
 const parseUrlMiddleware = require("common/parse-url-middleware");
 const loggerMiddleware = require("common/redux-logger");
 const {LOCAL_STORAGE_KEY} = require("common/constants");
+let store;
+
+/**
+ * Returns whether or not the selectors have finished initializing.
+ *
+ * XXX copy-pasted from selectors.js; it should really be exposed and unit
+ * tested there (#1798)
+ *
+ * @param  {Object} state  The state object.
+ * @return {Boolean}
+ */
+function areSelectorsReady(state) {
+  return state.TopSites.init && state.Highlights.init && state.Experiments.init;
+}
 
 /**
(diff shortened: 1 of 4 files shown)


common/create-store.js
```diff
@@ -6,6 +6,20 @@ const {Channel} = require("common/ReduxChannel");
...
+ *
+ * XXX copy-pasted from selectors.js; it should really be exposed and unit
+ * tested there.
+ *
+ * @param  {Object} state  The state object.
+ * @return {Boolean}
+ */
+function areSelectorsReady(state) {
+  return state.TopSites.init && state.Highlights.init && state.Experiments.init;
+}
```
k88hudson: you should actually be able to put this in `selectors.js` now, if you like!

common/create-store.js
```diff
@@ -6,6 +6,20 @@ const {Channel} = require("common/ReduxChannel");
 const parseUrlMiddleware = require("common/parse-url-middleware");
 const loggerMiddleware = require("common/redux-logger");
 const {LOCAL_STORAGE_KEY} = require("common/constants");
+let store;
```
k88hudson: This is kind of a departure from how `createStore` worked originally (i.e., you could create multiple  independent stores with no overlapping state),  although in practice right now we only have one store per context so it won't actually be a problem. I think it might be worth commenting at least though

common/create-store.js
dmose: Done.

PR #1797, opened 2016-11-15
fix(search): #1795 Listen to browser.search.hiddenOneOffs

### LINKED ISSUE #1795: Listen to changes on `browser.search.hiddenOneOffs`
In order to update the selected search choices, let's listen to the following pref:
 https://dxr.mozilla.org/mozilla-central/source/browser/components/preferences/in-content/search.js#292-300

See https://github.com/mozilla/activity-stream/pull/1776 for further notes

k88hudson:
Fix #1795. This patch triggers an event when the `browser.search.hiddenOneOffs` changes, which `SearchFeed` already listens to.

### DIFF
--- a/addon/SearchProvider.js
+++ b/addon/SearchProvider.js
@@ -3,8 +3,9 @@ SearchSuggestionController, PrivateBrowsingUtils, exports, require */
 
 "use strict";
 const {Ci, Cu} = require("chrome");
+const {PrefsTarget} = require("sdk/preferences/event-target");
 const SEARCH_ENGINE_TOPIC = "browser-search-engine-modified";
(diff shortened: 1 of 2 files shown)


addon/SearchProvider.js
```diff
@@ -41,6 +44,13 @@ SearchProvider.prototype = {
   ]),
 
   /**
+   * Event listener for the HIDDEN_ENGINES_PREF pref change
+   */
+  _onPrefChange(e) {
+    this.emit(SEARCH_ENGINE_TOPIC, "hiddenOneOffs");
```
Mardak: `SearchFeed` happens to just update its list of engines when it eventually gets the action broadcasted by `ActivityStreams`. But this is somewhat dangerous as the same action might have `data` of the string or the engine.

PR #1800, opened 2016-11-16
fix(recommender): Remove recent filtering logic relying on faulty SCROLL_THRESHOLD usage

### LINKED ISSUE #1670: Show bookmarks sooner than the 30 minute recent filter
Split from #1638. The fix for #1493 has allows new pages/bookmarks to show up immediately but potentially up to 15 minutes later. We can special case bookmarks to have them show up sooner -- potentially on the next new tab.

Mardak:
Fix #1670 by not filtering out recently bookmarked/visited pages that relied on a pixel threshold. There isn't as much of a need to filter recent items now that we show 9 highlights.

r?@sarracini

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -16,14 +16,6 @@ module.exports = {
     ageLimit: "-4 day"
   },
 
-  // This is how many pixels before the bottom that
-  // infinite scroll is triggered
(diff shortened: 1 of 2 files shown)


sarracini:
Nice! R+

PR #1801, opened 2016-11-16
fix(addon): #1756 - fix a race condition in TabTracker

### LINKED ISSUE #1756: TabTracker race
When @k88hudson and I were pairing, we saw a console error where isActivityStreamsURL blew up because it was called before this._trackableURLs had been set (i.e. presumably a new tab had been opened after the TabTracker constructor had been called, but before TabTracker.init had finished).

We speculated that a reasonable fix would be to moving more of the side-effecty stuff out of the constructor and into init (in particular, we shouldn't be setting up any listeners before init is called).

ncloudioj:
This closes #1756

### DIFF
--- a/addon/TabTracker.js
+++ b/addon/TabTracker.js
@@ -23,13 +23,7 @@ function TabTracker(options) {
   this._tabData = {};
   this._clientID = options.clientID;
   this.onOpen = this.onOpen.bind(this);
-
   this._onPrefChange = this._onPrefChange.bind(this);
-  this.enabled = simplePrefs.prefs[TELEMETRY_PREF];
-  if (this.enabled) {
-    this._addListeners();
-  }
-  simplePrefs.on(TELEMETRY_PREF, this._onPrefChange);
 }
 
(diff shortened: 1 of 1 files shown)


addon/TabTracker.js
```diff
@@ -43,6 +35,14 @@ TabTracker.prototype = {
     this._trackableURLs = trackableURLs;
     this._experimentID = experimentId;
     this._store = store;
+
+    this.onOpen = this.onOpen.bind(this);
```
k88hudson: Technically this could stay in the constructor, although I guess it's fine to have it here

addon/TabTracker.js
```diff
@@ -43,6 +35,14 @@ TabTracker.prototype = {
...
+    this.enabled = simplePrefs.prefs[TELEMETRY_PREF];
+    if (this.enabled) {
+      this._addListeners();
+    }
+    this._onPrefChange = this._onPrefChange.bind(this);
```
k88hudson: Same with this

PR #1803, opened 2016-11-16
fix (content): #1763 fix context menu button clicking

### LINKED ISSUE #1763: Clicking on a context menu button does not close context menu
If you click on the context menu button above a top site or a highlight, it will bring up the context menu, but then clicking the button again does not dismiss it.  Should we consider changing that so clicking the context menu button closes the menu as well?

rlr:
Breaking out `TopSitesItem` into a separate component like we do for Spotlights fixed the issue.

r? @sarracini 

Fixes #1763

### DIFF
--- a/content-src/components/LinkMenuButton/LinkMenuButton.js
+++ b/content-src/components/LinkMenuButton/LinkMenuButton.js
@@ -4,7 +4,6 @@ const LinkMenuButton = React.createClass({
   render() {
     return (<button className="link-menu-button" onClick={e => {
       e.preventDefault();
-      e.stopPropagation();
       this.props.onClick(e);
     }}>
       <span className="sr-only">Open context menu</span>
--- a/content-src/components/TopSites/TopSites.js
+++ b/content-src/components/TopSites/TopSites.js
@@ -67,14 +67,7 @@ const TopSites = React.createClass({
         action_position: index,
         metadata_source: site.metadata_source
       };
-      if (site.recommended) {
-        payload.url = site.url;
-        payload.recommender_type = site.recommender_type;
-      }
       this.props.dispatch(actions.NotifyEvent(payload));
-      if (site.recommended) {
-        this.props.dispatch(actions.NotifyBlockRecommendation(site.url));
-      }
     };
   },
   render() {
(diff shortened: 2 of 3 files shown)


content-src/components/LinkMenuButton/LinkMenuButton.js
```diff
@@ -4,7 +4,6 @@ const LinkMenuButton = React.createClass({
   render() {
     return (<button className="link-menu-button" onClick={e => {
       e.preventDefault();
-      e.stopPropagation();
```
sarracini: we should revert the changes here https://github.com/mozilla/activity-stream/pull/1748/files#diff-bea2b5a6e0bdde5d8bb69333c83d78e4 if we remove the stopPropogation here. Or else we get the bug #1372

content-src/components/TopSites/TopSites.js
```diff
@@ -11,6 +11,41 @@ const Hint = require("components/Hint/Hint");
...
 
+const TopSitesItem = React.createClass({
+  getInitialState() {
+    return {showContextMenu: false};
+  },
+  getDefaultProps() {
+    return {onClick() {}};
+  },
+  render() {
+    const site = this.props;
```
sarracini: nit: unpack index from props here, since you're using it multiple times

content-src/components/TopSites/TopSites.js
```diff
@@ -25,40 +60,37 @@ const TopSites = React.createClass({
...
+  onClickFactory(index, site) {
+    return () => {
+      let payload = {
+        event: "CLICK",
+        page: this.props.page,
+        source: "TOP_SITES",
+        action_position: index,
+        metadata_source: site.metadata_source
+      };
+      if (site.recommended) {
```
sarracini: you can remove this recommendation stuff - top sites will never have a recommendation and also we removed the recommendation logic so this can go. (lines 72-79)

content-src/components/LinkMenuButton/LinkMenuButton.js
rlr: ah. great catch!

rlr:
@sarracini i pushed all the fixes from r? ^^

PR #1804, opened 2016-11-17
fix(addon): Differentiate homepage-loaded activity stream for pings

### LINKED ISSUE #1606: Differentiate about:home vs about:newtab in pings
**Here's the deal:** 
Activity Stream sets about:home to be Activity Stream, but we don't have any way of knowing which interactions came from about:home and which ones came from about:newtab. That being said, in the SHIELD study we need to deal with that. We want to get the SHIELD study out as soon as possible so let's do the minimum amount of work while still pleasing everyone. Here are the options:

**Option a)** Do engineering work in Activity Stream to differentiate between clicks from about:home and about:newtab. Throw away about:home clicks for SHIELD study users in the Activity Stream variant. No data on about:home clicks 

**Option b)** Do engineering work in Activity Stream to differentiate between clicks from about:home and about:newtab. Monkey patch about:home for Tiles variant to collect the search clicks on about:home and compare ONLY search data from about:home in Tiles vs search data from about:home in Activity Stream

**Option c)** Turn off setting about:home to Activity Stream for SHIELD users in that variant. No data collected on about:home in either Tiles or Activity Stream

**Option d)** Ship it as is and get the first SHIELD study out as is and we'll get a proper fix in for the next SHIELD study

**Option e)** Set about:home to about:newtab for SHIELD study users in the Tiles variant. We don't need to differentiate between pages and we just group the data together as just 'Tiles' vs 'Activity Stream'

Let's decide which one of these we want to do.
cc @Mardak @emtwo @tspurway @nchapman @ncloudioj @jaredkerim @k88hudson @dmose

Mardak:
Fix #1606 by setting the home page to a special #/HOME hash that gets reported as the page in payloads.

r?@rlr

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -546,12 +546,12 @@ ActivityStreams.prototype = {
     if (!ss.storage.homepageOverriden &&
         (!prefService.isSet(HOME_PAGE_PREF) ||
          ["about:home", "about:blank"].includes(prefService.get(HOME_PAGE_PREF)))) {
-      prefService.set(HOME_PAGE_PREF, this._newTabURL);
+      prefService.set(HOME_PAGE_PREF, `${this._newTabURL}HOME`);
     }
   },
 
   _unsetHomePage() {
-    if (prefService.get(HOME_PAGE_PREF) === this._newTabURL) {
+    if (prefService.get(HOME_PAGE_PREF) === `${this._newTabURL}HOME`) {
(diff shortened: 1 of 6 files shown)


rlr:
changes look good. I think you need to update a test:

```
The following tests failed:

  ./test/test-ActivityStreams-home-page.test activity stream loads on home page when appropriate: failure
```

Mardak:
Ah yup! I also added a new test checking for page = "HOME"

PR #1818, opened 2016-11-22
Add commit hash in the version string

### LINKED ISSUE #1203: Tweak -dev- add-on to use short Git SHA instead of current_time?
Currently, when I install a `dev` add-on, it displays as "1.1.2-dev-1472745680" in my **about:addons** page.

I'm not sure what that timestamp is, or what commit it relates to. It may be easier if we showed the short Git SHA (instead of `current_time`):

``` sh
$ git rev-parse --short HEAD # d620b93
```

Not sure if this is as easy as modifying our [./fabfile.py:32-33](https://github.com/mozilla/activity-stream/blob/9dac5bda7f9a0afd11ed04a989c425e8eb77d596/fabfile.py#L32-L33), or if we generate builds from other scripts as well.

/cc @oyiptong @ncloudioj

### LINKED ISSUE #1799: add commit-hash to build somewhere
It's very handy to be able to tell exactly what build of the add-on someone is running.  We should expose the git commit-hash of the build somewhere in the UI (the debug page, perhaps?) so that we can get it when we need it...

Dan

ncloudioj:
This closes #1799 and closes #1203 

For the dev and pre-release channel, it'll attach the commit hash to the version string. Also, it'll show the version in the DebugPage.

### DIFF
--- a/content-src/components/DebugPage/DebugPage.js
+++ b/content-src/components/DebugPage/DebugPage.js
@@ -49,6 +49,11 @@ const DebugPage = React.createClass({
     return (<main className="debug-page">
       <div className="new-tab-wrapper">
 
+        <section>
+          <h2>Version</h2>
+          <p>{this.props.raw.Prefs.prefs["sdk.version"]}</p>
+        </section>
+
         <section className="experiments">
           <h2>Experiments</h2>
           <p>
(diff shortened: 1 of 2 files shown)


k88hudson:
Ok, looks good to me!

PR #1819, opened 2016-11-22
Emit action when experiment prefs are changed

k88hudson:
To test this patch, open the debug page (look at the Experiments section) and edit an experiment pref in `about:config`.

On master, no change should show up immediately.
In this branch, you should immediately see the correct value.

### DIFF
--- a/common/action-manager.js
+++ b/common/action-manager.js
@@ -53,7 +53,8 @@ const am = new ActionManager([
   "DISABLE_HINT",
   "APP_INIT",
   "PLACES_STATS_UPDATED",
-  "MERGE_STORE"
+  "MERGE_STORE",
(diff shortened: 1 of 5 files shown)


common/action-manager.js
```diff
@@ -146,10 +145,6 @@ function NotifyUpdateSearchString(searchString) {
   return Notify("NOTIFY_UPDATE_SEARCH_STRING", {searchString}, {skipMasterStore: true});
 }
 
-function RequestExperiments() {
```
k88hudson: These is just some unused stuff, since we don't actually request experiments any more (they just get dispatched directly to the store)

k88hudson:
Sorry, I have the wrong issue number on the name of this branch. Reopening

PR #1821, opened 2016-11-22
fix(metadata): Fixes 1658 Consolidate favicon fields

sarracini:
Consolidates favicon, favicons, and favicon_url into favicon_url
Consolidates background_color, favicons.color, favicon_colors, favicon_color into background_color
Does that on addon side so we only store the ones we need - and removes it from selectors side
Updates Tests

Don't review - needs addon tests for consolidation

### DIFF
--- a/addon/PreviewProvider.js
+++ b/addon/PreviewProvider.js
@@ -7,7 +7,9 @@ const self = require("sdk/self");
 const {TippyTopProvider} = require("addon/TippyTopProvider");
 const {getColor} = require("addon/ColorAnalyzerProvider");
 const {absPerf} = require("common/AbsPerf");
+const {consolidateBackgroundColors, consolidateFavicons} = require("addon/lib/utils");
 
+const {BACKGROUND_FADE} = require("../common/constants");
 const ENABLED_PREF = "previews.enabled";
 const METADATA_SOURCE_PREF = "metadataSource";
 const EMBEDLY_SOURCE_NAME = "Embedly";
@@ -167,19 +169,18 @@ PreviewProvider.prototype = {
   /**
     * Get the main colors from the favicon
     */
-  _getFaviconColors(links) {
-    return Promise.all(
-      links.map(link => new Promise(resolve => {
-        if (!link.favicon) {
-          resolve(link);
-          return;
-        }
-        getColor(link.favicon, link.url).then(color => {
-          resolve(Object.assign({}, link, {favicon_color: color}));
-        }, () => resolve(link)).catch(err => Cu.reportError(err));
-      })
-      )
-    ).catch(err => Cu.reportError(err));
+  _getFaviconColors(link) {
+    return new Promise(resolve => {
+      if (!link.favicon) {
+        resolve(null);
+      }
+      getColor(link.favicon, link.url)
+        .then(color => resolve([...color, BACKGROUND_FADE]))
+        .catch(err => {
+          Cu.reportError(err);
+          resolve(null);
+        });
+    });
   },
 
   /**
@@ -189,8 +190,7 @@ PreviewProvider.prototype = {
     let processedLinks = this._processLinks(links);
     this._asyncSaveLinks(processedLinks, event);
 
-    return this._asyncGetEnhancedLinks(processedLinks, event).then(
-      cachedLinks => this._getFaviconColors(cachedLinks));
+    return this._asyncGetEnhancedLinks(processedLinks, event);
   },
 
   /**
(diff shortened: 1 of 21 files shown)


addon/PreviewProvider.js
```diff
@@ -208,38 +207,86 @@ PreviewProvider.prototype = {
...
+  },
+
+  /*
+   * Consolidate the background colors from tippytop, firefox, and our own metadata.
+   * TippyTop background color and metadata background color both need to be converted
+   * from hex to an rgb array, whereas the firefox background color is already rgb
+   */
+  _consolidateBackgroundColors(tippyTopBackgroundColor, firefoxBackgroundColor, metadataFavicons) {
+    const metadataBackgroundColor = metadataFavicons && metadataFavicons[0] && metadataFavicons[0].color;
+    return this._hexToRgb(tippyTopBackgroundColor) || firefoxBackgroundColor || this._hexToRgb(metadataBackgroundColor);
```
sarracini: i think the firefoxbackgroundcolor first, i don't think we're guaranteed a favicon color from metadata, so relying on firefox is safer

addon/PreviewProvider.js
```diff
@@ -208,38 +207,86 @@ PreviewProvider.prototype = {
...
+      enhancedLink.bookmarkDateCreated = link.bookmarkDateCreated;
+
+      // get favicon and background color from firefox
+      const firefoxBackgroundColor = yield this._getFaviconColors(link);
+      const firefoxFavicon = link.favicon;
+
+      // get favicon and background color from tippytop
+      const {favicon_url, background_color, metadata_source} = this._tippyTopProvider.processSite(link);
+      const tippyTopFavicon = favicon_url;
+      const tippyTopBackgroundColor = background_color;
```
pdehaan: Not sure of the exact Firefox support for this, but I think it should be possible to assign destructured values to a new variable name directly, instead of having to reassign here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names

```js
const {favicon_url: tippyTopFavicon, background_color: tippyTopBackgroundColor, metadata_source} = this._tippyTopProvider.processSite(link);
```

addon/PreviewProvider.js
```diff
@@ -208,38 +207,86 @@ PreviewProvider.prototype = {
...
+  _consolidateBackgroundColors(tippyTopBackgroundColor, firefoxBackgroundColor, metadataFavicons) {
+    const metadataBackgroundColor = metadataFavicons && metadataFavicons[0] && metadataFavicons[0].color;
+    return this._hexToRgb(tippyTopBackgroundColor) || firefoxBackgroundColor || this._hexToRgb(metadataBackgroundColor);
+  },
+
+  /**
+   * Convert a hex color string to the RGB form, e.g. #0A0102 => [10, 1, 2]
+   */
+  _hexToRgb(hex) {
+    let result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
```
pdehaan: I know this is just copy/pasta, but we're always certain that the `#RRGGBB` will always be the full 7 characters and not the shortened `#RGB` syntax, right?

sarracini: darn... I just checked tippytop and some come in as #FFF and some as #FFFFFF so maybe it's a good idea to generalize this. Good catch peter

addon/PreviewProvider.js
sarracini: Ahh I see the concern there. You've convinced me 😉

sarracini:
@k88hudson can you check the content side stuff? (selectors, faker, that kinda thing)

addon/PreviewProvider.js
```diff
@@ -208,38 +207,93 @@ PreviewProvider.prototype = {
...
+    const metadataFavicon = metadataFavicons && metadataFavicons[0] && metadataFavicons[0].url;
+    return tippyTopFavicon || metadataFavicon || firefoxFavicon;
+  },
+
+  /*
+   * Consolidate the background colors from tippytop, firefox, and our own metadata.
+   * TippyTop background color and metadata background color both need to be converted
+   * from hex to an rgb array, whereas the firefox background color is already rgb
+   */
+  _consolidateBackgroundColors(tippyTopBackgroundColor, metadataFavicons, firefoxBackgroundColor) {
```
k88hudson: personally I would probably put this and `hexToRgb` in a separate file in `addons/lib` perhaps – it would make them easier to test/isolate all the different cases (with the background fade, etc.) If it's a lot of extra work don't worry about it too much though.

common/selectors/colorSelectors.js
```diff
@@ -15,19 +15,6 @@ const BACKGROUND_FADE = 0.5;
  * @return {array}     An RGB colour as an array of numbers. E.g. [150, 150, 150]
  */
 function getBackgroundRGB(site) {
```
k88hudson: maybe rename this to `getFallbackColor` or something like that?

common/selectors/colorSelectors.js
```diff
@@ -69,8 +56,8 @@ const selectSiteIcon = createSelector(
   site => site,
   site => {
     const {favicon, parsedUrl, label} = selectSiteProperties(site);
-    const backgroundRGB = getBackgroundRGB(site);
-    const backgroundColor = site.background_color || toRGBString(...backgroundRGB, favicon ? BACKGROUND_FADE : 1);
+    const backgroundRGB = site.background_color || [...getBackgroundRGB(site), favicon ? BACKGROUND_FADE : 1];
+    const backgroundColor = toRGBString(backgroundRGB);
     const fontColor = getBlackOrWhite(...backgroundRGB);
     const firstLetter = prettyUrl(parsedUrl.hostname)[0];
     return {
```
k88hudson: I think this file could use some further consolidation, but doesn't have to be in this patch. Like maybe the fallback color stuff could just be in the same place as the other background color stuff.

Anyways, I was kinda confused, but it checks out 👍

Issue #1829, opened 2016-11-25
Refactor _respondToSearchRequests into SearchFeed


k88hudson:
Once #1750 lands, we can refactor the rest of the search handlers into `SearchFeed`. It requires having access to sending and broadcasting.

Issue #1750, opened 2016-11-08
Allow sending/broadcasting from Feeds


k88hudson:
This would be helpful for dispatching actions that need to be broadcast to content immediately.

PR #1830, opened 2016-11-25
feat(feeds): #1750 PageModProvider, add broadcasting/send to feeds

k88hudson:
This patch refactors the worker management to use a `Map` with unique identifiers, so actions from each worker are tagged with a `workerId`. 

Actually using send/broadcast in feeds will be implemented in https://github.com/mozilla/activity-stream/issues/1829

### DIFF
--- a/addon/PageModProvider.js
+++ b/addon/PageModProvider.js
@@ -1,0 +1,120 @@
+const generateUUID = require("sdk/util/uuid");
+const privateBrowsing = require("sdk/private-browsing");
+const {Cu} = require("chrome");
+const {PageMod} = require("sdk/page-mod");
+const {data} = require("sdk/self");
+const {CONTENT_TO_ADDON} = require("common/event-constants");
+const {WORKER_ATTACHED_EVENT} = require("common/constants");
+
+module.exports = class PageModProvider {
+  constructor({pageURL, uuid, onAddWorker, onRemoveWorker} = {}) {
+    this.workers = new Map();
+    this.pageURL = pageURL;
+    this.uuid = uuid || (() => String(generateUUID.uuid()));
+    this.onAddWorker = onAddWorker;
+    this.onRemoveWorker = onRemoveWorker;
+    this._perfMeter = null;
+    this._pagemod = null;
+    this._logEvent = null;
+  }
+
+  /**
+   * Sets up communications with the pages and manages the lifecycle of workers
+   */
+  init({onAttach, onMessage, logEvent} = {}) {
+    this._logEvent = logEvent;
+    this._pagemod = new PageMod({
+      include: [`${this.pageURL}*`],
+      contentScriptFile: data.url("content-bridge.js"),
+      contentScriptWhen: "start",
+      attachTo: ["existing", "top"],
+      onAttach: worker => {
+        if (onAttach) {
+          onAttach();
+        }
+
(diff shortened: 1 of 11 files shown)


content-test/addon/PageModProvider.test.js
```diff
@@ -0,0 +1,182 @@
...
+          // This is the callback for a CONTENT_TO_ADDON message
+          return worker.port.on.firstCall.args[1];
+        }
+        it("should remove the worker if the event type is 'pagehide'", () => {
+          const callback = setupWithOnMessage();
+          assert.ok(instance.workers.get(worker));
+          callback({type: "pagehide"});
+          assert.isUndefined(instance.workers.get(worker), "pagehide event removes worker");
+        });
+        it("should remove the worker if the event type is 'pagehide'", () => {
```
sarracini: dupe of the test above?

content-test/addon/PageModProvider.test.js
```diff
@@ -0,0 +1,182 @@
...
+    });
+  });
+  describe("#getWorkerById", () => {
+    it("should get a worker by id", () => {
+      const id = "asdkl123";
+      const worker = {};
+      instance.workers.set(worker, id);
+      instance.workers.set({}, "wjklwlk");
+      instance.workers.set({}, "a3elaks");
+      assert.equal(instance.getWorkerById(id), worker);
```
sarracini: It's not super clear to me how this test is testing that it got the correct worker based on the id... if all the workers are just an empty object, they'll all pass this test no matter what id you pass in? Unless you're just testing that the ```getWorkerById``` function returns any worker?

addon/PageModProvider.js
```diff
@@ -0,0 +1,119 @@
...
+   * Adds a worker and returns the new id
+   */
+  addWorker(worker) {
+    if (this.workers.has(worker)) {
+      return this.workers.get(worker);
+    }
+    const id = this.uuid();
+    this.workers.set(worker, id);
+    if (this._logEvent) {
+      this._logEvent(worker.tab, "WORKER_ATTACHED");
```
sarracini: Maybe add ```WORKER_ATTACHED``` to constants? That way we don't have to update the tests if it changes

content-test/addon/PageModProvider.test.js
```diff
@@ -0,0 +1,182 @@
...
+      instance = new PageModProvider({onAddWorker});
+      instance.addWorker({});
+      assert.calledOnce(onAddWorker);
+    });
+    it("should call logEvent if it was passed to the init function", () => {
+      const logEvent = sinon.spy();
+      const worker = {tab: {}};
+      instance.init({logEvent});
+      instance.addWorker(worker);
+      assert.calledWith(logEvent, worker.tab, "WORKER_ATTACHED");
```
sarracini: grab this from constants, once it's added

content-test/addon/PageModProvider.test.js
k88hudson: `{} !== {}`, so only the reference to the actual object cached with `worker` will actually be correct

sarracini: oh javascript, you got me again

k88hudson:
@sarracini ok, changes in https://github.com/mozilla/activity-stream/pull/1830/commits/388c28272eae1ee05aaa1a556a1acb3c390ff3d4

PR #1833, opened 2016-11-28
chore(feeds): #1829 Refactor respondToSearchRequests into SearchFeed

k88hudson:
This patch removes the rest of the search-related items from `ActivityStreams.js` and adds them to `SearchFeed.js`, and adds unit tests for each thing.

### DIFF
--- a/addon/Feeds/SearchFeed.js
+++ b/addon/Feeds/SearchFeed.js
@@ -1,5 +1,7 @@
 const Feed = require("addon/lib/Feed");
 const am = require("common/action-manager");
+const getCurrentBrowser = require("addon/lib/getCurrentBrowser");
+const {Cu} = require("chrome");
 
 module.exports = class SearchFeed extends Feed {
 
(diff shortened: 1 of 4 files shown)


addon/Feeds/SearchFeed.js
```diff
@@ -23,6 +25,38 @@ module.exports = class SearchFeed extends Feed {
...
 
+  /**
+   * getSuggestions - Retrieves search suggestions for a search string (action.data)
+   *                  Dispatches an array of suggestions.
+   */
+  getSuggestions(action) {
+    const browser = getCurrentBrowser();
+    return this.options.searchProvider.asyncGetSuggestions(browser, action.data)
+      .then(suggestions => this.options.send(am.actions.Response("SEARCH_SUGGESTIONS_RESPONSE", suggestions), action.workerId, true))
+      .catch(e => Cu.reportError(e));
```
pdehaan: FWIW, I think this may "break" the promise chain since the `.catch()` will swallow any thrown errors and then return a successful promise. Not sure if we want to re-`throw` the error so that any methods calling `getSuggestions()` can handle error cases as appropriate. Or explicitly return an empty array/object in the `catch()` case... ¯\\\_(ツ)\_/¯

rlr: Interesting, I just did a quick grep and saw a lot of `}).catch(err => Cu.reportError(err));` in our repo.

addon/lib/getCurrentBrowser.js
```diff
@@ -0,0 +1,14 @@
+/* globals XPCOMUtils, windowMediator */
+const {Cu} = require("chrome");
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+
+XPCOMUtils.defineLazyServiceGetter(this, "windowMediator",
+                                   "@mozilla.org/appshell/window-mediator;1",
+                                   "nsIWindowMediator");
+
+module.exports = function getCurrentBrowser() {
```
rlr: brilliant! Thank you!

addon/Feeds/SearchFeed.js
k88hudson: The promise is actually just returned for testing, it doesn't actually get used by anything else

PR #1834, opened 2016-11-28
medata(feeds): Closes #1826 Create a MetadataFeed

### LINKED ISSUE #1826: Create a Metadata Feed
We should create a metadata feed which periodically wakes up and fetches metadata for new links - and the regular feeds become read-only operations

sarracini:
- Change existing TopSitesFeed and HighlightsFeed to only request cached data
- Create a MetadataFeed which accumulates links on places changes and after it's reached a certain number (5), send them to PreviewProvider to fetch metadata for them
- On app init, get the recently visited links and fetch their metadata first
- ```METADATA_FEED_REQUEST``` is the event source for when the feed goes and gets metadata 
- Rename functions to ```getCachedMetadata``` and ```fetchNewMetadata```
- send a ```METADATA_FEED_UPDATE``` action for the HighlightsFeed to refresh with
- Updates existing tests to take correct form of links
- New tests for MetadataFeed

@k88hudson r?


addon/ActivityStreams.js
```diff
@@ -90,9 +90,13 @@ function ActivityStreams(metadataStore, tabTracker, telemetrySender, options = {
     send: this.sendById.bind(this),
     searchProvider: this._searchProvider,
     // TODO: move this into Feeds. Requires previewProvider/tabTracker to be independent
-    getMetadata: (links, type) => {
+    getCachedMetadata: (links, type) => {
```
k88hudson: yes much better

PR #1835, opened 2016-11-28
fix(reporting): fix #1831 Missing TOP_SITES pings

### LINKED ISSUE #1831: Missing TOP_SITES pings
It looks like our top sites pings have dropped significantly since our last release on Nov. 24th. We can see it in this graph: https://sql.telemetry.mozilla.org/queries/1800#3196 with this query:
```
SELECT date, COUNT(*)
FROM activity_stream_events_daily
WHERE source = 'TOP_SITES'
GROUP BY date
ORDER BY date
```
This is likely a reporting bug since all other numbers are looking consistent (mau/dau/highlights clicks/searchs etc)

sarracini:
Fix those sneaky TOP_SITES pings


k88hudson:
Could we perhaps add a test for this?

sarracini:
@k88hudson yeah i'll add a test in a follow up PR

PR #1840, opened 2016-11-30
fix (addon): #1711 use OS specific share icon in toolbar

### LINKED ISSUE #1711: Share icon needs to be tailored to the host system
Particularly Windows user have a hard time recognizing the current share icon, which is only commonly used on Mac/iOS. 

We should use the Windows specific share icon for Windows users. 

image
Mac, Windows,  Linux

rlr:
r? @ncloudioj can you test on linux?

Fixes #1711

### DIFF
--- a/addon/ShareProvider.js
+++ b/addon/ShareProvider.js
@@ -3,6 +3,7 @@
 const {Cc, Ci, Cu} = require("chrome");
 const {data} = require("sdk/self");
 const DEFAULT_MANIFEST_PREFS = require("addon/ShareManifests");
+const runtime = require("sdk/system/runtime");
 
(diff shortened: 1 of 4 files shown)


addon/ShareProvider.js
```diff
@@ -403,7 +404,14 @@ ShareProvider.prototype = {
         if (!node || node.id !== id) {
           return;
         }
-        node.setAttribute("image", data.url("content/img/glyph-share-16.svg"));
```
pdehaan: Does anything still use the generic "glyph-share-16.svg", or should we delete that now that we're moving towards OS-specific iconography?

rlr: good point! I'll remove that before merging!

rlr: actually, I just realized that icon is still used in the context menu. I'll leave that for another time.

PR #1844, opened 2016-12-01
feat(highlights): Always show recent bookmarks even without images

### LINKED ISSUE #1794: All recently bookmarked items should make an apearance on NewTab.
There is a class of bookmarked content that never seems to make it to the New Tab. This is probably content that might not have "good enough" imagery. 

For example, this doesn't show on New Tab (for me). https://www.dropbox.com/s/nh706j38jyoz26u/SnoozeTabs%20UI%20v2.1.pdf?dl=0

In short, anything a user bookmarks would make the cut.

Mardak:
Fix #1794. Turns out enhanced links don't get bookmarkId passed but bookmarkDateCreated does. Also, the boost for bookmarks doesn't work for image-less ones because that score drops to 0, so multiplying doesn't do anything.

r?@ncloudioj

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -42,7 +42,7 @@ class Baseline {
 
     const age = this.normalizeTimestamp(entry.lastVisitDate);
     const imageCount = entry.images ? entry.images.length : 0;
-    const isBookmarked = entry.bookmarkId || 0;
+    const isBookmarked = !!entry.bookmarkDateCreated;
     const description = this.extractDescriptionLength(entry);
     const pathLength = urlObj.pathname.split("/").filter(e => e.length).length;
     const image = this.extractImage(entry.images);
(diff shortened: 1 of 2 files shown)


Mardak:
@ncloudioj does this fix things for you?

common/recommender/Baseline.js
```diff
@@ -42,7 +42,7 @@ class Baseline {
 
     const age = this.normalizeTimestamp(entry.lastVisitDate);
     const imageCount = entry.images ? entry.images.length : 0;
-    const isBookmarked = entry.bookmarkId || 0;
+    const isBookmarked = !!entry.bookmarkDateCreated;
```
ncloudioj: Duh, that's sneaky. Nice catch, Ed!

pdehaan:
If we _don't_ have a preview image, do we show a placeholder, or something else?

ncloudioj:
It shows an image-less tile instead. e.g.

screen shot 2016-12-01 at 10 04 04 am

pdehaan:
/cc @nchapman @bryanbell @aaronrbenson 

Is the image-less tile kosher, or should we have some placeholder monochrome Firefox logo or something?

PR #1860, opened 2016-12-12
feat(NewTab): (closes #1675) load placeholder content when "waiting"…

### LINKED ISSUE #1675: add placeholder topsites and highlights behind "analyzing history" dialog
This is part of https://trello.com/c/qsRbWroC/86-as-a-new-and-existing-user-i-should-have-a-clear-understanding-of-what-is-going-on-with-my-new-tab-experience. In particular, we're trying to implement https://mozilla.invisionapp.com/share/D995GXJVC#/201431382_First_Run_Messaging Scenario 2A.

After the "analyzing history" dialog is implemented in #1674, we should show skeleton top-sites and highlights behind it, and they should each fill in as soon as the data is ready.

dmose:
… dialog up

See #1675 for details of the intent here.

To test this out:

* view the storybook pages for the new and added files
* do "npm run start" in one shell; and then "npm run firefox -- -p some-profile-with-some-history" in another.  
  * Quickly click "Command-T" to open a new tab as soon as the browser comes up.  If you time it right, you'll see the skeleton page with the dialog, which will morph into a loaded page.

### DIFF
--- a/content-src/components/HighlightContext/HighlightContext.js
+++ b/content-src/components/HighlightContext/HighlightContext.js
@@ -28,5 +28,31 @@ HighlightContext.propTypes = {
   date: React.PropTypes.number
 };
 
-module.exports = HighlightContext;
-module.exports.types = highlightTypes;
+/**
+ * Only display a placeholder version (ie just outlines/shapes), for use
+ * before sufficient data is available to display.
+ *
+ * This should be a stateless, functional component.  Unfortunately, testing
+ * these kinda sucks; see
+ * http://stackoverflow.com/questions/36682241/testing-functional-components-with-renderintodocument)
+ *
+ * The bottom line: we should start using Enzyme and switch this over.
+ */
+const PlaceholderHighlightContext = React.createClass({
+  render() {
+    return (
+      <div className="highlight-context placeholder">
+        <div className="hc-icon">
+          <div className="icon" />
+        </div>
+        <div className="hc-label" />
+      </div>
+    );
+  }
+});
+
+module.exports = {
+  HighlightContext,
+  PlaceholderHighlightContext,
+  types: highlightTypes
+};
--- a/content-src/components/NewTabPage/NewTabPage.js
+++ b/content-src/components/NewTabPage/NewTabPage.js
@@ -7,7 +7,6 @@ const Search = require("components/Search/Search");
 const Loader = require("components/Loader/Loader");
 const {actions} = require("common/action-manager");
 const setFavicon = require("lib/set-favicon");
-const classNames = require("classnames");
 const PAGE_NAME = "NEW_TAB";
 const {HIGHLIGHTS_LENGTH} = require("common/constants");
 
@@ -58,14 +57,15 @@ const NewTabPage = React.createClass({
           title="Welcome to new tab"
           body="Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily."
           label="Identifying your Highlights" />
-        <div className={classNames("show-on-init", {on: this.props.isReady})}>
+        <div className="show-on-init on">
           <section>
-            <TopSites page={PAGE_NAME} sites={props.TopSites.rows} showHint={props.TopSites.showHint} />
+            <TopSites placeholder={!this.props.isReady} page={PAGE_NAME}
+              sites={props.TopSites.rows} showHint={props.TopSites.showHint} />
           </section>
 
           <section>
-            <Spotlight page={PAGE_NAME} length={HIGHLIGHTS_LENGTH}
-              sites={props.Highlights.rows} />
+            <Spotlight placeholder={!this.props.isReady} page={PAGE_NAME}
+              length={HIGHLIGHTS_LENGTH} sites={props.Highlights.rows} />
           </section>
         </div>
       </div>
--- a/content-src/components/SiteIcon/SiteIcon.js
+++ b/content-src/components/SiteIcon/SiteIcon.js
@@ -67,4 +67,25 @@ SiteIcon.propTypes = {
   }).isRequired
 };
 
-module.exports = SiteIcon;
+/**
+ * A placeholder version (ie just outlines/shapes), for use before sufficient
+ * data is available to display.
+ *
+ * Unfortunately, this can't be a function component as long as we have the
+ * ref and it is used in testing.
+ */
+const PlaceholderSiteIcon = React.createClass({
+  render() {
+    return (
+      <div ref="icon" className="spotlight-icon">
+        <div className="site-icon-wrapper" />
+        <div className="site-icon-title" />
+      </div>
+    );
+  }
+});
+
+module.exports = {
+  PlaceholderSiteIcon,
+  SiteIcon
+};
--- a/content-src/components/Spotlight/Spotlight.js
+++ b/content-src/components/Spotlight/Spotlight.js
@@ -4,10 +4,10 @@ const {justDispatch} = require("common/selectors/selectors");
 const getHighlightContextFromSite = require("common/selectors/getHighlightContextFromSite");
 const {selectSiteProperties} = require("common/selectors/siteMetadataSelectors");
 const {actions} = require("common/action-manager");
-const SiteIcon = require("components/SiteIcon/SiteIcon");
+const {SiteIcon, PlaceholderSiteIcon} = require("components/SiteIcon/SiteIcon");
 const LinkMenu = require("components/LinkMenu/LinkMenu");
 const LinkMenuButton = require("components/LinkMenuButton/LinkMenuButton");
-const HighlightContext = require("components/HighlightContext/HighlightContext");
+const {HighlightContext, PlaceholderHighlightContext} = require("components/HighlightContext/HighlightContext");
 const Hint = require("components/Hint/Hint");
 const classNames = require("classnames");
 
@@ -88,6 +88,22 @@ const SpotlightItem = React.createClass({
   }
 });
 
+const PlaceholderSpotlightItem = React.createClass({
+  render() {
+    return (
+      <li className="spotlight-item placeholder">
+        <a>
+          <div className="spotlight-image portrait" ref="image">
+            <PlaceholderSiteIcon />
+          </div>
+          <div className="inner-border" />
+        </a>
+        <PlaceholderHighlightContext />
+      </li>
+    );
+  }
+});
+
 SpotlightItem.propTypes = {
   page: React.PropTypes.string,
   source: React.PropTypes.string,
@@ -105,7 +121,8 @@ const Spotlight = React.createClass({
   getDefaultProps() {
     return {
       length: 3,
-      page: "NEW_TAB"
+      page: "NEW_TAB",
+      placeholder: false
     };
   },
   onClickFactory(index, site) {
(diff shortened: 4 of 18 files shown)


Mardak:
Where did the term `skeleton` come from? Isn't `placeholder` the usual one for this type of functionality?

I only quickly glanced so far, but initial comment: this approach seems to pass a flag all the way down to each individual item to render the outline. Will we want to show a mix of some outline and regular top sites / highlights?

If not, then couldn't the parent `TopSites`/`Spotlight` just handle creating the 6 outline top sites and 9 highlights? Although I suppose that might not be very React-component-y. I.e., when `!isReady`, have a completely different `render` path for either/both sections to avoid touching the existing individual components.

dmose:
You're right: placeholder is clearer; I'll change the PR.

We don't intend to mix them, and after earlier discussion with @k88hudson, that was my original intent.  I strayed from that in part just because I wanted to get the thing finally in and landed, and I wanted to avoid markup/CSS fragility (eg if the main box changes size, the placeholder box also needs to).  That I said, I think I can see how to mostly avoid that without too much trouble.  Furthermore, if we create separate Placeholder{TopSite, Spotlight}Items, I think it fits into the React idiom nicely and more cleanly too.

I'll update the patch.

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -56,14 +56,15 @@ const NewTabPage = React.createClass({
           title="Welcome to new tab"
           body="Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily."
           label="Identifying your Highlights" />
-        <div className={classNames("show-on-init", {on: this.props.isReady})}>
+        <div className={classNames("show-on-init", "on")}>
```
rlr: here you don't really need the `classNames()` function anymore?

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -56,14 +56,15 @@ const NewTabPage = React.createClass({
...
-            <TopSites page={PAGE_NAME} sites={props.TopSites.rows} showHint={props.TopSites.showHint} />
+            <TopSites placeholder={!this.props.isReady} page={PAGE_NAME}
+              sites={props.TopSites.rows} showHint={props.TopSites.showHint} />
           </section>
 
           <section>
-            <Spotlight page={PAGE_NAME} length={HIGHLIGHTS_LENGTH}
-              sites={props.Highlights.rows} />
+            <Spotlight placeholder={!this.props.isReady} page={PAGE_NAME}
+              length={HIGHLIGHTS_LENGTH} sites={props.Highlights.rows} />
```
rlr: so we'll show the empty placeholders with the `<Loader/>` modal on top?

content-src/components/TopSites/TopSites.js
```diff
@@ -19,35 +19,55 @@ const TopSitesItem = React.createClass({
...
+    } else {
+      siteIcon = <SiteIcon ref="icon" className="tile-img-container" site={site} faviconSize={32} showTitle={true} />;
+    }
+
+    let className = classNames("tile-outer",
+      {active: isActive, placeholder: this.props.placeholder});
+    return (
+      <div className={className} key={site.guid || site.cache_key || index}>
+        <a onClick={() => this.props.onClick(index)} className="tile" href={site.url} ref="topSiteLink">
+        {siteIcon}
```
rlr: nit: should be indented 2 more spaces

content-src/components/NewTabPage/NewTabPage.js
dmose: Yes; that's the design.  See scenario 2A in https://mozilla.invisionapp.com/share/D995GXJVC#/201431382_First_Run_Messaging Scenario

content-src/components/NewTabPage/NewTabPage.js
dmose: Indeed; good catch!

content-src/components/HighlightContext/HighlightContext.js
```diff
@@ -2,30 +2,42 @@ const React = require("react");
...
-        <span ref="icon" className={`icon icon-${icon}`} /><span className="sr-only">{this.props.type}</span>
-      </span>
-      <span ref="label" className="hc-label">{label}</span>
-      <span hidden={!timestamp} ref="timestamp" className="hc-timestamp">{timestamp}</span>
-      {tooltip && <Tooltip ref="tooltip" label={tooltip} />}
-    </div>);
+    return (
+      <div className={classNames("highlight-context", "tooltip-container", {placeholder: this.props.placeholder})}>
+        <span className="hc-icon">
+          <span ref="icon" className={`icon icon-${icon}`} /><span className="sr-only">{this.props.type}</span>
```
Mardak: nit: `{placeholder: this.props.placeholder}` and `this.props.type` can just be `{placeholder}` and `type` if you change the `const type` above to `const {placeholder, type} = this.props`

dmose:
Almost there; all that remains to clean up is the TopSites, and then it should be ready for more review.

dmose:
OK, all comments from @Mardak and @rlr so far are addressed; this baby is ready for another pass.

content-src/components/Spotlight/Spotlight.js
```diff
@@ -129,13 +146,24 @@ const Spotlight = React.createClass({
...
       <ul className="spotlight-list">
-        {sites.map((site, i) => <SpotlightItem
-          index={i}
-          key={site.guid || site.cache_key || i}
-          page={this.props.page}
-          source="FEATURED"
-          onClick={this.onClickFactory(i, site)}
-          {...site} />)}
+        {sites.map((site, i) => {
+          if (this.props.placeholder) {
```
rlr: I'm curious how this works. If the data isn't ready, I expected `sites` could be empty. If so, we won't show any placeholders. Or is `sites` somehow hydrated with placeholders?

dmose: Good question!

TopSites has a DEFAULT_LENGTH property of 6, but that's really just the maximum length, as its used to slice out of the sites array.

Spotlights has a default "length" property of 3, but NewTabPage overrides it with HIGHLIGHTS_LENGTH from common/constants.js.  Again, this is really a maximum length, as it's used to slice out of "sites".

console logging shows the things being passed from the selectors to NewTabs as 6 and 3, which I suspect is just our default set of TopSites and the three (soon-to-disappear) Highlights.

Which means that I suspect you're onto something here, and that at least for the Highlights, we should do something to ensure that we have some skeletons all the time.

dmose: Er, placeholders.  :-)

dmose: Talked to @aaronrbenson; he's requested that we use three for the highlights.

dmose:
OK, I've updated the placeholder highlights so that it always renders exactly 3, which also cleaned up the code even more, and puts us in a good place to start breaking this apart into stateless components, both for testability and performance.   Placeholder TopSites doesn't currently have this problem (we're not planning on removing the defaults), but it'll want a similar cleanup in the future anyway.

content-src/components/TopSites/TopSites.scss
```diff
@@ -16,6 +16,23 @@
...
+    // Only display a placeholder version (ie just outlines/shapes), for use
+    // before sufficient data is available to display.
+    &.placeholder {
+      .site-icon-wrapper {
+        width: 32px;
+        height: 74px;
+      }
+
+      .site-icon-title {
+        border-top: 2px solid $placeholder-grey;
```
Mardak: This line is thicker and lighter than that of the tile. It should probably just be the same:

`1px solid $faintest-black`
https://github.com/mozilla/activity-stream/blob/c9286d83db030fa7a99e8b0c9de6643d34589a52/content-src/main.scss#L45

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -56,14 +56,15 @@ const NewTabPage = React.createClass({
           title="Welcome to new tab"
           body="Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily."
           label="Identifying your Highlights" />
```
Mardak: When testing this, I noticed there was a delay before the loader appeared. Is this 2 second delay desired? Probably want to have a followup bug to figure that out. It was introduced here:

https://github.com/mozilla/activity-stream/commit/54ea1f816c5a84ca13466df53077aee9a96e01ad

content-src/components/TopSites/TopSites.js
```diff
@@ -25,20 +25,23 @@ const TopSitesItem = React.createClass({
...
+        </a>
+        <LinkMenuButton onClick={() => this.setState({showContextMenu: true, activeTile: index})} />
+        <LinkMenu
+          visible={isActive}
+          onUpdate={val => this.setState({showContextMenu: val})}
+          site={site}
+          page={this.props.page}
+          source="TOP_SITES"
+          index={index} />
+      </div>);
```
Mardak: nit: these changes can be undone? It looks to be just a pulling out of `className` and reindenting?

content-src/components/SiteIcon/SiteIcon.js
```diff
@@ -67,4 +67,25 @@ SiteIcon.propTypes = {
...
+ * A placeholder version (ie just outlines/shapes), for use before sufficient
+ * data is available to display.
+ *
+ * Unfortunately, this can't be a function component as long as we have the
+ * ref and it is used in testing.
+ */
+const PlaceholderSiteIcon = React.createClass({
+  render() {
+    return (
+      <div ref="icon" className="spotlight-icon">
```
Mardak: nit: to be consistent, does it make sense to also include `placeholder` in the class list? Although the current usage always has a `PlaceholderSiteIcon` within another parent `Placeholder*`.

content-test/components/NewTabPage.story.js
```diff
@@ -34,8 +34,19 @@ const Container = props => (
 // itself.
 
 storiesOf("NewTabPage", module)
-  .add("Be on the lookout popup", () =>
-    <Container>
-      <NewTabPage {...mockData} page="SPOTLIGHT_STORYBOOK" dispatch={dispatch} />,
-    </Container>
+  .add("data available (this.props.isReady==true)", () => {
```
Mardak: nit: Most other items (what's the term? story? view?) start with an uppercase. Also, not sure if the verbose `(this.props.isReady==true)` adds much over the "human" readable label. ;)

content-test/components/Spotlight.story.js
```diff
@@ -57,6 +57,11 @@ storiesOf("Highlight List", module)
     <Container>
       <Spotlight length={HIGHLIGHTS_LENGTH} page="SPOTLIGHT_STORYBOOK" sites={fakeSpotlightItems.slice(1)} />
     </Container>
+  )
+  .add("placeholder view (data still loading)", () =>
```
Mardak: These really do make things easier for me to test/review! Thanks

content-test/components/NewTabPage.story.js
dmose: Fixed the casing on the new stories.  The code snippet I expect to occasionally be useful to folks using storybook for development (like I do).

content-src/components/NewTabPage/NewTabPage.js
dmose: Talked this over with @aaronrbenson in IRC.  He suggests leaving the 2s delay before displaying the dialog, and making the code poll every second rather than every three, so I've done that.

content-test/components/NewTabPage.story.js
dmose: Casing fixed.  The code bit is more useful to folks who develop directly using storybook (like me :-).

content-test/components/Spotlight.story.js
dmose: I'm glad to hear it; feedback on stuff like this is always helpful.  I actually find it an easier way to develop for some things.

content-src/components/TopSites/TopSites.js
dmose: I've undone everything except the fact that I had to add a ref="icon" for testing.

content-src/components/TopSites/TopSites.scss
dmose: This line is actually gone on master now.  So I've just removed the rule entirely in the interest of consistency.

PR #1870, opened 2016-12-14
fix(highlights): Use a single recent query with a limit of 500

### LINKED ISSUE #1855: Increase the "limit" for the special case in the "GetRecentVisitedLinks"
There is a special case scenario in the `GetRecentVistedLinks` that only fetches 20 links if there are not enough recently visited links. It might end up with <9 highlights as follows:

screen shot 2016-12-07 at 12 03 51 pm

Mardak:
Fix #1855 and update tests to use the constant instead of hardcoded number.

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -29,10 +29,7 @@ XPCOMUtils.defineLazyGetter(this, "gPrincipal", () => {
   return Services.scriptSecurityManager.getNoAppCodebasePrincipal(uri);
 });
 
-const {
-  LINKS_QUERY_LIMIT,
-  HIGHLIGHTS_THRESHOLDS
-} = require("../common/constants");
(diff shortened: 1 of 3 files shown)


Mardak:
Hrm. Locally the tests pass but it does take a while now that it's adding and fetching 50x the links. So looks like we should either reduce the query limit or increase test timeout?

Mardak:
Reworking the patch to simplify the query to always use the fallback and not bother with "all pages in last 4 days" query. Also to rework the tests using a custom limit instead of the large 999 default.

jaredkerim:
@Mardak This is really good, but I want to run it past everybody at the product meeting on Monday to make sure we don't make a significant change to user facing behaviours without discussing it.  We wouldn't want to drastically alter our engagement metrics without telling people before we do it.  We might also consider making this change as an experiment to see *if* and *how* it affects engagement before rolling it out to everybody, but we'll see what people think on Monday.

PR #1871, opened 2016-12-15
fix(highlights): Remove first-run default highlights and related packaged images

### LINKED ISSUE #1864: Everywhere/Extend Firefox/Complete Your Install cards displayed under highlights despite having browser history
Save Everywhere/Extend Firefox/Complete Your Install cards are displayed under highlights even when the user has lots of browser history.

screen shot 2016-12-12 at 6 51 56 pm

Mardak:
Fix #1864 by removing first-run data and updating tests.


sarracini:
@Mardak looks good, there's some additional stuff you can remove in ```getHighlightContextFromSite.js``` (along with it's test) if you want. And also you'd have to change this test https://github.com/mozilla/activity-stream/blob/master/content-test/components/HighlightContext.test.js#L53.

Mardak:
> https://github.com/mozilla/activity-stream/blob/master/content-test/components/HighlightContext.test.js#L53

@sarracini Are you thinking of just removing that one test, the context type defaults, and the selector logic?

https://github.com/mozilla/activity-stream/blob/master/content-src/components/HighlightContext/types.js#L2
https://github.com/mozilla/activity-stream/blob/master/content-src/selectors/getHighlightContextFromSite.js#L15

I was starting to remove the first-run type as well, but `TopSites` uses that to determine `LinkMenu` items.

sarracini:
@Mardak I was thinking of just removing this https://github.com/mozilla/activity-stream/blob/master/common/selectors/getHighlightContextFromSite.js#L25-L26 and the test that goes along with it here https://github.com/mozilla/activity-stream/blob/master/content-test/common/selectors/getHighlightContextFromSite.test.js#L29-L34

Unless the name is misleading and it isn't only just for Highlights?

PR #1888, opened 2016-12-20
fix(highlights): Heavily boost recent bookmarks and not much from 3 days or older

### LINKED ISSUE #1887: Recently bookmarked pages should show on newtab no mater what, older ones shouldn't
@bryanbell suggests bookmarks made within 3 days get a boost so that just bookmarked pages show up for sure:
```
give a bookmark that’s 0-120minutes old a Plus 100 points
then a bookmark that’s 120-500mins old gets a +40
a bookmark that’s older than 3 days gets no point boost
```

Mardak:
Fix #1887. Instead of a fixed number, base the boost on the age of the bookmark. r?@ncloudioj

### DIFF
--- a/common/recommender/Baseline.js
+++ b/common/recommender/Baseline.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const {BOOKMARK_AGE_DIVIDEND} = require("common/constants");
 const URL = require("common/vendor")("url-parse");
 const getBestImage = require("../getBestImage");
 
@@ -24,7 +25,7 @@ class Baseline {
 
     // These are features used for adjusting the final score.
     // Used by decay function to filter out features.
-    this.adjustmentFeatures = ["isBookmarked", "imageCount", "age", "idf"];
+    this.adjustmentFeatures = ["bookmarkAge", "imageCount", "age", "idf"];
 
     if (!this.options.highlightsCoefficients) {
       throw new Error("Coefficients not specified");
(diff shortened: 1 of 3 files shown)


common/recommender/Baseline.js
```diff
@@ -42,13 +42,13 @@ class Baseline {
 
     const age = this.normalizeTimestamp(entry.lastVisitDate);
     const imageCount = entry.images ? entry.images.length : 0;
-    const isBookmarked = !!entry.bookmarkDateCreated;
+    const bookmarkAge = entry.bookmarkDateCreated ? Math.max(1, Date.now() - entry.bookmarkDateCreated) : 0;
```
ncloudioj: For the better readablity, please add a comment for `bookmarkAge`. For example, calculate the age in milliseconds (minimal as 1) if it's a bookmark, otherwise using 0 for non-bookmark links"

common/recommender/Baseline.js
```diff
@@ -104,9 +104,10 @@ class Baseline {
       newScore *= 0.2;
     }
 
-    // Boost boomarks even if they have low score or no images
-    if (features.isBookmarked) {
-      newScore += 1.8;
+    // Boost boomarks even if they have low score or no images giving a
+    // just-bookmarked page a near-infinite boost, and 3-day page 1.0 boost
+    if (features.bookmarkAge) {
+      newScore += 3 * 24 * 60 * 60 * 1000 / features.bookmarkAge;
```
ncloudioj: Consider using a constant here, and update the comment accordingly?

ncloudioj:
This looks good for now.

Since we're adding more and more magic sauce to here, documenting the process of the algorithm evolution is highly recommended :)

PR #1893, opened 2016-12-21
fix(newtab): Show a border between top site icon/title and highlight picture/details

### LINKED ISSUE #1892: Top Site and Highlights cards need a line separator between the icon/image and text sections
There should be a separator on Top Site and Highlight tiles, as described in the attached picture. 
tile-seperator

Mardak:
Fix #1892. r?@dmose There was a reference to "spotlight-icon" from `PlaceholderSiteIcon` that was being used for `PlaceholderSpotlightItem`, so I switched it to "site-icon" and updated both places.

After/before pictures here: https://github.com/mozilla/activity-stream/issues/1892#issuecomment-268619007

### DIFF
--- a/content-src/components/SiteIcon/SiteIcon.js
+++ b/content-src/components/SiteIcon/SiteIcon.js
@@ -77,7 +77,7 @@ SiteIcon.propTypes = {
 const PlaceholderSiteIcon = React.createClass({
   render() {
     return (
-      <div ref="icon" className="spotlight-icon">
+      <div ref="icon" className="site-icon">
         <div className="site-icon-wrapper" />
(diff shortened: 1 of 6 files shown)


Mardak:
After/before pictures here: https://github.com/mozilla/activity-stream/issues/1892#issuecomment-268619007

dmose:
Good catch on the bug.  

r=@dmose, as long as getting rid of the SiteIcon change and PlaceholderSiteIcon doesn't cause unintended consequences elsewhere.  You might also consider just making PlaceholderSiteIcon take an "extraClasses" property or something and keep using it.

PR #1894, opened 2016-12-22
feat(metadata): Closes #1685 Fetch page content locally

### LINKED ISSUE #1685: Fetch page content for async metadata parsing
If the local metadata parser fails to successfully parse the metadata for a page, then we will still fall back to making a network call to the remote metadata service.  Eventually we need to remove all dependency on remote metadata services.  

To facilitate this, we need to add the capacity for activity stream to make a network request for the content of a URL and parse and store the metadata asynchronously if we attempt to show a tile for a URL for which we do not have metadata.  

We should launch this as an experiment and monitor its effectiveness and impact on usage and performance.

We need to be careful about how we do this so that we do not inadvertently impact the user by impersonating them through web requests, ie if requesting a URL would cause some stateful side effect on the users behalf.  We should consider disabling all cookies and making such requests anonymously.

sarracini:
* Creates an experiment to fetch page content locally
* If the experiment is on - call a new function in PageScraper
* Split some logic in PageScraper for less code duplication with new function
* Create a separate function for framescript listener (having it in the init is yucky)
* Update existing tests
* Write new tests

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1894)

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -95,7 +95,8 @@ function ActivityStreams(metadataStore, tabTracker, telemetrySender, options = {
     fetchNewMetadata: (links, type) => {
       const event = this._tabTracker.generateEvent({source: type});
       return this._previewProvider.asyncSaveLinks(links, event);
-    }
+    },
+    fetchNewMetadataLocally: (links, type) => this._pageScraper.asyncFetchLinks(links, type)
   });
   this._store = createStore({middleware: this._feeds.reduxMiddleware});
   this._feeds.connectStore(this._store);
--- a/addon/PageScraper.js
+++ b/addon/PageScraper.js
@@ -9,10 +9,21 @@ Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
 Cu.importGlobalProperties(["URL"]);
+Cu.importGlobalProperties(["fetch"]);
 
 const DEFAULT_OPTIONS = {
   framescriptPath: new URL("data/page-scraper-content-script.js", options.prefixURI),
   blacklist: ["about:", "localhost:", "resource://"]
+};
+const PERFORMANCE_EVENT_NAMES = {
+  framescript_event: "framescriptMessageReceived",
+  local_fetch_event: "localFetchStarted",
+  metadata_raw_html: "metadataReceivedRawHTML",
+  metadata_exists: "metadataExists",
+  metadata_invalid: "metadataInvalidReceived",
+  metadata_sucess: "metadataParseSuccess",
+  metadata_fail: "metadataParseFail",
+  network_fail: "networkRequestFailed"
 };
 
 function PageScraper(previewProvider, tabTracker, options = {}) {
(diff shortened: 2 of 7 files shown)


addon/PageScraper.js
```diff
@@ -30,33 +31,64 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
+   * Parse the HTML and attempt to insert it in the metadata database
    */
   _asyncParseAndSave: Task.async(function*(rawHTML, url, event) {
-    let startTime = Date.now();
-    this._tabTracker.handlePerformanceEvent(event, "metadataReceivedRawHTML", startTime);
-
-    let link = yield this._previewProvider.asyncLinkExist(url);
-    if (!link) {
+    this._tabTracker.handlePerformanceEvent(event, "metadataReceivedRawHTML", Date.now());
+    let linkExists = yield this._previewProvider.asyncLinkExist(url);
```
jaredkerim: Move this check to asyncSaveMetadata

addon/PageScraper.js
```diff
@@ -30,33 +31,64 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
+      this._tabTracker.handlePerformanceEvent(event, "metadataExists", Date.now());
+    }
+  }),
+
+  /**
+   * Save the metadata in the MetadataStore DB
+   */
+  _asyncSaveMetadata: Task.async(function*(metadata, event) {
+    const startTime = Date.now();
+    if (this._shouldSaveMetadata(metadata)) {
```
jaredkerim: Fail fast out of this if statement instead of embedding the entire function body inside of it

jaredkerim: Add an additional check here for linkExists

addon/PageScraper.js
```diff
@@ -30,33 +31,64 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
-   * HTML and insert it in the metadata database
+   * Parse the HTML and attempt to insert it in the metadata database
    */
   _asyncParseAndSave: Task.async(function*(rawHTML, url, event) {
-    let startTime = Date.now();
-    this._tabTracker.handlePerformanceEvent(event, "metadataReceivedRawHTML", startTime);
-
-    let link = yield this._previewProvider.asyncLinkExist(url);
-    if (!link) {
+    this._tabTracker.handlePerformanceEvent(event, "metadataReceivedRawHTML", Date.now());
```
jaredkerim: Move all the performance ping key names into a constant object

    const TELEMETRY_EVENTS = {
      RECEIVED_HTML: "metadataReceivedRawHTML",
      ...
    }

addon/PageScraper.js
```diff
@@ -30,33 +31,64 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
   /**
+   * Make a network request for links that the MetadataFeed has requested metadata for.
+   * Attempt to parse the html from that page and insert into the DB
+   */
+  asyncFetchLinks: Task.async(function*(links, event) {
+    for (let link of links) {
+      let linkExists = yield this._previewProvider.asyncLinkExist(link.url);
+      if (!linkExists) {
+        const rawHTML = yield this._fetchContent(link.url);
+        const metadata = this._metadataParser.parseHTMLText(rawHTML, link.url);
```
jaredkerim: Just call asyncParseAndSave here

sarracini:
Leaving this PR here for now, but it's blocked on #1897

sarracini:
Actually, this can land with the experiment turned off - so that the code doesn't rot. @jaredkerim give it another look over and we can file a ticket to turn this experiment on in a couple weeks

addon/PageScraper.js
```diff
@@ -30,33 +38,63 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
+    const metadata = this._metadataParser.parseHTMLText(rawHTML, url);
+    this._asyncSaveMetadata(metadata, event);
+  },
+
+  /**
+   * Save the metadata in the MetadataStore DB
+   */
+  _asyncSaveMetadata: Task.async(function*(metadata, event) {
+    const startTime = Date.now();
+    let linkExists = yield this._previewProvider.asyncLinkExist(metadata.url);
```
jaredkerim: Can you move this into the 'should save metadata' method

addon/PageScraper.js
```diff
@@ -30,33 +38,63 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
+    if (linkExists) {
+      this._tabTracker.handlePerformanceEvent(event, PERFORMANCE_EVENT_NAMES.metadata_exists, Date.now());
+      return;
+    }
+    if (!this._shouldSaveMetadata(metadata)) {
+      this._tabTracker.handlePerformanceEvent(event, PERFORMANCE_EVENT_NAMES.metadata_invalid, Date.now() - startTime);
```
jaredkerim: If we handle both the case of metadata exists and metadata is invalid inside teh should save method, then we'll need to fire this event inside the should save method too.

addon/PageScraper.js
```diff
@@ -30,33 +38,63 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
+      return;
+    }
+    try {
+      metadata.images = yield this._computeImageSize(metadata);
+    } catch (e) {
+      Cu.reportError(`PageScraper failed to compute image size for ${metadata.url}`);
+    }
+    this._insertMetadata(metadata);
```
jaredkerim: Can we just unpack the contents of this method (calling previewprovider) right into here?

addon/ActivityStreams.js
```diff
@@ -95,6 +95,10 @@ function ActivityStreams(metadataStore, tabTracker, telemetrySender, options = {
     fetchNewMetadata: (links, type) => {
       const event = this._tabTracker.generateEvent({source: type});
       return this._previewProvider.asyncSaveLinks(links, event);
+    },
+    fetchNewMetadataLocally: (links, type) => {
+      const event = this._tabTracker.generateEvent({source: type});
+      return this._pageScraper.asyncFetchLinks(links, event);
```
jaredkerim: Okay I know this is a little annoying, but can you change this so that here we loop over links like so:


    for (let link of links) {
        const event = this._tabTracker.generateEvent({source: type});
        this._pageScraper.asyncFetchLink(link, event);
    }

And all the methods just deal with a single link, that way we have a single telemetry event for each link we're fetching and all the methods only deal with a single link?

addon/PageScraper.js
```diff
@@ -30,33 +38,63 @@ function PageScraper(previewProvider, tabTracker, options = {}) {
...
+      const rawHTML = yield this._fetchContent(link.url);
+      this._asyncParseAndSave(rawHTML, link.url, event);
     }
   }),
 
   /**
+   * Wrapper for requesting the URL and returning it's DOM
+   */
+  _fetchContent: Task.async(function*(url) {
+    let response = yield fetch(url);
```
jaredkerim: These can both be consts can't they?

addon/PageScraper.js
sarracini: you mean just call previewprovider's method directly here instead of adding an extra function call?

addon/ActivityStreams.js
sarracini: I can't actually do this - the feeds expect a function to be returned for getData to work properly. If I do this I get an error in the console saying: ```this.options.fetchNewMetadataLocally(...) is undefined```

Issue #1906, opened 2016-12-29
Add snippet component


k88hudson:
In this ticket, I will add snippet UI as per the trello card / comp, as well as test and an example in the debug page. Blocks #1904

Issue #1904, opened 2016-12-28
Run snippet experiment


k88hudson:
Let's run an experiment to test the effectiveness of snippets -- this will not be the final implementation of how we serve/manage snippets, just an experiment to determine how users engage with them. We will run a few variations (just hard-coded for now) and collect the following metrics:

- clicks on the snippets
- dismissals of the snippet

k88hudson:
@aaronrbenson I need some assets for a couple of snippets to try. What do you want to do for that?

PR #1907, opened 2016-12-29
feat(content): #1906 Add Snippet component

k88hudson:
This patch adds a `Snippet` component, as well as an example in the `Debug` page, a storybook, and tests. Refer to the trello card for more information: https://trello.com/c/LmsamLox/39-as-mozilla-we-want-to-inform-our-users-about-interesting-things-marketing-initiatives-friendly-tips-other

To try it:
- Open the debug page, you should see it in the bottom
- Run `npm run storybooks`

Notes:
- I had to use `dangerouslySetInnerHTML` for the snippet description, since it sometimes contains links/other HTML. However, because this text is controlled internally, this should be OK. We should keep this in mind whenever we decide how to store/ship snippet content.
- There is no implementation of the snippet yet on the New Tab page, I will do that once we confirm the variants for the experiment in #1904

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1907)

### DIFF
--- a/content-src/components/Snippet/Snippet.js
+++ b/content-src/components/Snippet/Snippet.js
@@ -1,0 +1,55 @@
+const React = require("react");
+const classNames = require("classnames");
+
+const Snippet = React.createClass({
+  onClose(e) {
+    e.preventDefault();
+    this.props.setVisibility(false);
+  },
+  render() {
+    const {visible, description, title, image} = this.props;
+    const imageAttributes = {
+      className: classNames("snippet-image", {"placeholder": !image}),
+      style: {backgroundImage: image ? `url(${image})` : "none"}
+    };
+    return (<div className={classNames("snippet", {"hide-with-fade-out": !visible})}>
+      <div className="snippet-container">
+        <div ref="image" {...imageAttributes} />
+        <div className="snippet-text">
+          {title && <h3 ref="title" className="snippet-title">{title}</h3>}
+          {/*
+            NOTE: This could potentially be dangerous if we end up changing
+            the data source to something other than commited strings to the repo.
+            We should consider sanitization/reevaluating this strategy when we
+            change the data source for snippets.
+          */}
+          <p
+            dangerouslySetInnerHTML={{__html: description}} // eslint-disable-line react/no-danger
+            className="snippet-description"
+            ref="description" />
+        </div>
+      </div>
+      <button ref="closeButton" className="snippet-close-button" onClick={this.onClose}>
+        <span className="icon icon-dismiss" />
+        <span className="sr-only">Close snippet</span>
+      </button>
+    </div>);
+  }
+});
+
+/**
+ * Snippet - A component for the New Tab page snippet
+ *
+ * @prop  {bool} visible        Determines whether the snippet is currently visible or not
+ * @prop  {string} description  ~140 character text in the snippet
+ * @prop  {string} title        Bold text at the top of the snippet
+ * @prop  {func} setVisibility  Called with false when the user clicks on the 'close' button
+ */
+Snippet.propTypes = {
+  visible: React.PropTypes.bool,
+  description: React.PropTypes.string.isRequired,
+  title: React.PropTypes.string,
+  setVisibility: React.PropTypes.func.isRequired
+};
+
+module.exports = Snippet;
--- a/content-src/components/Snippet/Snippet.js
+++ b/content-src/components/Snippet/Snippet.js
@@ -1,7 +1,15 @@
 const React = require("react");
 const classNames = require("classnames");
 
+/**
+ * This should be a stateless, functional component.  Unfortunately, testing these kinda sucks; see
+ * http://stackoverflow.com/questions/36682241/testing-functional-components-with-renderintodocument)
+ * So for now, we are using PureRenderMixin.
+ */
+const PureRenderMixin = require("react-addons-pure-render-mixin");
+
 const Snippet = React.createClass({
+  mixins: [PureRenderMixin],
   onClose(e) {
     e.preventDefault();
     this.props.setVisibility(false);
--- a/content-src/components/Snippet/Snippet.scss
+++ b/content-src/components/Snippet/Snippet.scss
@@ -1,0 +1,78 @@
+.snippet {
+  $snippet-padding: 26px;
+  $snippet-bg-color: rgba($white, 0.95);
+  $snippet-divider-color: rgba($black, 0.1);
+  $snippet-width: 500px;
+  $snippet-base-font-size: 12px;
+  $snippet-title-size: 14px;
+  $snippet-image-size: 64px;
+  $snippet-image-margin: 15px;
+  $snippet-link-color: $link-blue;
+
+  display: flex;
+  position: fixed;
+  bottom: 0;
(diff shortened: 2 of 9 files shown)


content-src/components/Snippet/Snippet.js
```diff
@@ -0,0 +1,37 @@
...
+const classNames = require("classnames");
+
+const Snippet = React.createClass({
+  onClose(e) {
+    e.preventDefault();
+    this.props.setVisibility(false);
+  },
+  render() {
+    const {visible, description, title, image} = this.props;
+    return (<div className={classNames("snippet", {"hidden": !visible})}>
```
dmose: We'll save a bit of render/layout time, and follow the React idiom better if, instead of passing in ```visible```, simple have the parent not render <Snippet> at all if visible is set to false.

content-src/components/Snippet/Snippet.js
```diff
@@ -0,0 +1,37 @@
...
+      <div className="snippet-container">
+        <div ref="image" className={classNames("snippet-image", {"placeholder": !image})} style={{backgroundImage: image ? `url(${image})` : "none"}} />
+        <div className="snippet-text">
+          <h3 ref="title" hidden={!title} className="snippet-title">{title}</h3>
```
dmose: Another place where we can simply not render if title isn't set, instead of rendering and hiding.

content-test/components/Snippet.test.js
```diff
@@ -0,0 +1,65 @@
...
+      assert.notInclude(ReactDOM.findDOMNode(createInstance()).className, "hidden");
+    });
+    it("should hide the snippet if props.visible is false", () => {
+      const instance = createInstance({visible: false});
+      assert.include(ReactDOM.findDOMNode(instance).className, "hidden");
+    });
+  });
+  describe(".setVisibility", () => {
+    it("should call props.setVisibility when the close button is clicked with false", () => {
+      const setVisibility = sinon.spy();
```
dmose: I suspect "with false" wants to move before the word "when"  :-)

dmose:
Since this is currently just an experiment, I'd be ok with leaving the hidden stuff as-is, as long as there are XXX comments to fix if we decide to productionalize.

content-src/components/Snippet/Snippet.scss
```diff
@@ -0,0 +1,77 @@
...
+  background-color: $snippet-bg-color;
+  padding: $snippet-padding;
+  border-top: 1px solid $snippet-divider-color;
+  z-index: 1000;
+  align-items: center;
+
+  &.hidden {
+    transition: opacity 250ms ease-out, transform 250ms ease-out;
+    opacity: 0;
+    transform: translate3d(0, 100%, 0);
```
k88hudson: Oh yeah sorry lol, derp

content-src/components/Snippet/Snippet.scss
```diff
@@ -0,0 +1,77 @@
...
+      text-decoration: none;
+
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+
+    .snippet-title {
+      font-size: $snippet-title-size;
+      margin: 0 0 0.2em;
```
k88hudson: In this case I want it to be relative to the actual title size

content-src/components/Snippet/Snippet.js
k88hudson: Hm, now that I look at it, we kind of need this to be the case to support the CSS transition. This gets a bit messy it we're handling those two cases (hidden v.s. never shown ever), let me think about it

k88hudson:
I added a new property called `enabled` separate from `visible`, what do you think? I feel like it might be better just to assume that the consumer of the component should do:

```js
{isThereASnippet && <Snippet .... />}

```

instead of relying on a specific prop, since it's top level anyway. What do you think?

dmose:
I think your ```{isThereASnippet && <Snippet ... />``` is just right, and is actually what I was doing a poor job communicating.  In other words, it's not clear to me why either visible or enabled are necessary props.

content-src/components/Snippet/Snippet.js
```diff
@@ -0,0 +1,37 @@
+const React = require("react");
+const classNames = require("classnames");
+
+const Snippet = React.createClass({
```
dmose: Sure, but along with the comment, please do the PureRenderMixin thing now -- it's a one-liner:

https://github.com/mozilla/loop/blob/616b1e9c2dd844a6068cd194ac6d17d771241981/add-on/panels/js/desktopViews.jsx#L19

content-src/components/Snippet/Snippet.scss
dmose: Sounds reasonable; add a comment explaining this so noone inadvertently undoes it in the future?

content-src/components/Snippet/Snippet.scss
```diff
@@ -0,0 +1,77 @@
...
+  bottom: 0;
+  left: 0;
+  right: 0;
+  background-color: $snippet-bg-color;
+  padding: $snippet-padding;
+  border-top: 1px solid $snippet-divider-color;
+  z-index: 1000;
+  align-items: center;
+
+  &.hidden {
```
dmose: Looking at this, I see that I missed part of what's going on because I'm so used to a ".hidden" always meaning "visibility: hidden;" or "display: none;".   Maybe a different className here?

content-src/components/Snippet/Snippet.js
```diff
@@ -0,0 +1,57 @@
...
+  onClose(e) {
+    e.preventDefault();
+    this.props.setVisibility(false);
+  },
+  render() {
+    const {enabled, visible, description, title, image} = this.props;
```
dmose: As mentioned before, I think your suggestion of {isThereASnippet && <Snippet ... /.>} is just right, which should get rid of the need for one of these props.

You might (I'm not sure) also be able to get rid of the need for the other using [ReactCSSTransitionGroup](https://facebook.github.io/react/docs/animation.html#animating-one-or-zero-items).  That may or may not be worth the effort this time around.

dmose:
Whoops.  Gotta love keeping multiple tabs open :-)

k88hudson:
@dmose, ok I added in `react-addons-pure-render-mixin` at the matching version for React. I will file the ticket for adding enzyme

k88hudson:
Filed https://github.com/mozilla/activity-stream/issues/1929

content-src/components/Snippet/Snippet.scss
```diff
@@ -0,0 +1,77 @@
+.snippet {
+  $snippet-padding: 26px;
+  $snippet-bg-color: rgba($white, 0.95);
+  $snippet-divider-color: rgba($black, 0.1);
+  $snippet-width: 500px;
+  $snippet-base-font-size: 12px;
+  $snippet-title-size: 14px;
```
k88hudson: I'd be happy to have a broader discussion about `rem`s – I personally think it's OK, maybe even better to use pixels – but for this case, probably best to stick with what the rest of the project uses (i.e. pixels)

dmose:
Looks good to me; r=dmose!  Thanks for the changes.

PR #1924, opened 2017-01-05
fix (addon): #1867 fix ExperimentProvider exception on unload.

### LINKED ISSUE #1867: Experiment Provider unhandled exceptions
Version: 1.1.9-dev-a2a18d5

Experiment Provider is causing the following unhandled exceptions at install time:

screenshot 2016-12-13 15 00 31

rlr:
Switched to using Provider.jsm instead of sdk/preferences/service.

That fixed the issue. You can repro/verify fix by disabling and enabling from about:addons looking at your browser console.

Maybe we should make this switch everywhere in the codebase?

r? @jaredkerim 

Fixes #1867

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1924)

### DIFF
--- a/addon/ExperimentProvider.js
+++ b/addon/ExperimentProvider.js
@@ -15,12 +15,13 @@ XPCOMUtils.defineLazyGetter(this, "EventEmitter", () => {
 });
 
 exports.ExperimentProvider = class ExperimentProvider {
-  constructor(experiments = require("../experiments.json"), rng) {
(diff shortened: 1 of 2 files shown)


rlr:
oh crap, looks like i cant do that because ExperimentProvider is tested with karma and not jpm test. hmm

rlr:
ok ^^ i moved things around to keep the tests happy

rlr:
nvm, i still see the error in the log now. back to drawing board.

PR #1926, opened 2017-01-05
metrics(metadata): Closes #1897 Track metadata coverage for sites

### LINKED ISSUE #1897: Track metadata coverage in places stats feed
We would like to track what percentage of URLs in places also appear in the metadata table. 

To do this we can add a check in places stats feed here:

https://github.com/mozilla/activity-stream/blob/master/addon/Feeds/PlacesStatsFeed.js#L9

which checks the following 

1) find the earliest insert timestamp in metadata table
2) count the number of URLs P inserted into places since that timestamp
3) count the number of URLs M inserted into metadata store since that timestamp
4) fire a telemetry performance ping of the ratio of M/P

The goal of this tracking is help us track how local metadata parsing will improve this ratio and eventually push it towards 100%.

sarracini:
* PlacesStats Feed now collects metadata coverage and sends a ping
* MetadataStore has 2 new functions - one which gets the oldest entry in the database and the other which counts the rows in the database
* PlacesProvider has a new function to get all the places entries since the earliest timestamp from MetadataStore
* Write some more tests

@jaredkerim r? cc @ncloudioj

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1926)

### DIFF
--- a/addon/Feeds/PlacesStatsFeed.js
+++ b/addon/Feeds/PlacesStatsFeed.js
@@ -1,4 +1,5 @@
 const {PlacesProvider} = require("addon/PlacesProvider");
+const {Cu} = require("chrome");
 const Feed = require("addon/lib/Feed");
 const am = require("common/action-manager");
 const {PlacesStatsUpdate} = am.actions;
@@ -7,6 +8,7 @@ const UPDATE_TIME = 24 * 60 * 60 * 1000; // 24 hours
 module.exports = class PlacesStatsFeed extends Feed {
   // Used by this.refresh
   getData() {
+    this.sendStatsPing();
     return Promise.all([
       PlacesProvider.links.getHistorySize(),
       PlacesProvider.links.getBookmarksSize()
@@ -14,6 +16,20 @@ module.exports = class PlacesStatsFeed extends Feed {
     .then(([historySize, bookmarksSize]) => (
       PlacesStatsUpdate(historySize, bookmarksSize)
     ));
+  }
+  sendStatsPing() {
+    this.options.metadataStore.asyncGetOldestInsert().then(timestamp => {
+      if (timestamp) {
+        Promise.all([
+          PlacesProvider.links.getHistorySizeSince(timestamp),
+          this.options.metadataStore.asyncCountAllItems()
+        ]).then(([placesCount, metadataCount]) => {
+          let event = this.options.tabTracker.generateEvent({source: "PLACES_STATS_FEED"});
+          this.options.tabTracker.handlePerformanceEvent(event, "countHistoryURLs", placesCount);
+          this.options.tabTracker.handlePerformanceEvent(event, "countMetadataURLs", metadataCount);
+        }).catch(e => Cu.reportError(e));
+      }
+    }).catch(e => Cu.reportError(e));
   }
   onAction(state, action) {
     switch (action.type) {
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -525,6 +525,38 @@ MetadataStore.prototype = {
   }),
 
   /**
+   * Find the oldest entry in the database
+   *
+   * Returns the timestamp of the oldest entry in the database
+   */
+  asyncGetOldestInsert: Task.async(function*() {
+    let timestamp = null;
+    try {
+      const entry = yield this.asyncExecuteQuery("SELECT min(created_at) FROM page_metadata");
+      if (entry && entry.length) {
+        timestamp = entry[0][0];
+      }
+    } catch (e) {
+      throw e;
+    }
+    return timestamp;
+  }),
+
+  /**
+   * Counts all the items in the database
+   *
+   * Returns a promise with the array of the retrieved metadata records
+   */
+  asyncCountAllItems: Task.async(function*() {
+    try {
+      const result = yield this.asyncExecuteQuery("SELECT count(*) FROM page_metadata");
+      return result[0][0];
+    } catch (e) {
+      throw e;
+    }
+  }),
+
+  /**
   * Enables the data expiry job. The database connection needs to
   * be established prior to calling this function. Once it's triggered,
   * any following calls will be ignored unless the user disables
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -643,6 +643,25 @@ Links.prototype = {
   }),
 
   /**
+   * Gets History size since a certain timestamp
+   *
+   * @param {String} timestamp in milliseconds
+   *
+   * @returns {Promise} Returns a promise with the count of moz_places records
+   *                    that have been entered since the timestamp provided
+   */
+  getHistorySizeSince: Task.async(function*(timestamp) {
+    let sqlQuery = `SELECT count(*)
+                    FROM moz_places WHERE id IN
+                    (SELECT DISTINCT place_id FROM moz_historyvisits
+                    WHERE datetime(visit_date / 1000 / 1000, 'unixepoch') >= :timestamp)
+                    AND hidden = 0 AND last_visit_date NOT NULL`;
+
+    let result = yield this.executePlacesQuery(sqlQuery, {params: {timestamp}});
+    return result[0][0];
+  }),
+
+  /**
    * Gets Bookmarks count
    *
    * @returns {Promise} Returns a promise with the count of bookmarks
--- a/test/test-MetadataStore.js
+++ b/test/test-MetadataStore.js
@@ -181,6 +181,52 @@ exports.test_async_insert_all = function*(assert) {
   assert.deepEqual(items[7], [3, 5]);
 };
 
+exports.test_get_oldest_insert = function*(assert) {
+  // insert metadata at 3 different times
+  for (let metadata of metadataFixture) {
+    yield gMetadataStore.asyncInsert([metadata]);
+  }
+  // get all the timestamps, pick an entry, set that timestamp to be an hour earlier
+  const metadataIdRows = yield gMetadataStore.asyncExecuteQuery("SELECT id FROM page_metadata LIMIT 1");
+  const metadataId = metadataIdRows[0][0];
+  const expectedOldestTimestamp = Math.floor((Date.now() - (60 * 60 * 1000)) / 1000);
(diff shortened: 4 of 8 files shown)


addon/Feeds/PlacesStatsFeed.js
```diff
@@ -15,6 +17,21 @@ module.exports = class PlacesStatsFeed extends Feed {
...
   }
+  sendStatsPing() {
+    this.options.metadataStore.asyncGetOldestInsert().then(timestamp => (
+      Promise.all([
+        PlacesProvider.links.getHistorySizeSince(timestamp),
+        this.options.metadataStore.asyncCountAllItems()
+      ]).then(([placesCount, metadataCount]) => {
+        let metadataCoverage = 0;
+        if (placesCount !== 0 && metadataCount !== 0) {
+          metadataCoverage = metadataCount / placesCount * 100;
```
jaredkerim: Just send both counts as integers in two separate pings

countHistoryURLs
countMetadataURLs

addon/PlacesProvider.js
```diff
@@ -643,6 +643,28 @@ Links.prototype = {
...
+      return Promise.resolve(0);
+    }
+    let sqlQuery = `SELECT count(1)
+                    FROM moz_places WHERE id IN
+                    (SELECT DISTINCT place_id FROM moz_historyvisits
+                    WHERE visit_date /1000 >= ${timestamp})
+                    AND hidden = 0 AND last_visit_date NOT NULL`;
+
+    let result = yield this.executePlacesQuery(sqlQuery);
+    return result[0][0];
```
jaredkerim: We should just check that there's a result in there before doing this.

test/test-MetadataStore.js
```diff
@@ -181,6 +186,42 @@ exports.test_async_insert_all = function*(assert) {
   assert.deepEqual(items[7], [3, 5]);
 };
 
+exports.test_get_oldest_insert = function*(assert) {
+  // insert metadata at 3 different times
+  for (let metadata of metadataFixture) {
+    yield sleep(3000);
```
jaredkerim: Is there no better way to just manually specify a timestamp or something rather than sleeping for 3 seconds?

addon/Feeds/PlacesStatsFeed.js
```diff
@@ -15,6 +17,21 @@ module.exports = class PlacesStatsFeed extends Feed {
       PlacesStatsUpdate(historySize, bookmarksSize)
     ));
   }
+  sendStatsPing() {
+    this.options.metadataStore.asyncGetOldestInsert().then(timestamp => (
```
ncloudioj: Please add the exception handling code here. i.e. `when asyncGetOldestInsert() resolves with timestamps === null`

addon/Feeds/PlacesStatsFeed.js
```diff
@@ -15,6 +17,21 @@ module.exports = class PlacesStatsFeed extends Feed {
...
+      Promise.all([
+        PlacesProvider.links.getHistorySizeSince(timestamp),
+        this.options.metadataStore.asyncCountAllItems()
```
ncloudioj: asyncCountAllItems() might return null as well. Instead of returning nulls from metadata store, consider just raising exceptions from there and letting the client code decide how to handle those exceptions.

addon/Feeds/PlacesStatsFeed.js
```diff
@@ -15,6 +17,21 @@ module.exports = class PlacesStatsFeed extends Feed {
...
+      ]).then(([placesCount, metadataCount]) => {
+        let metadataCoverage = 0;
+        if (placesCount !== 0 && metadataCount !== 0) {
```
ncloudioj: It's fine if `metadataCount === 0`.

addon/MetadataStore.js
```diff
@@ -525,6 +525,39 @@ MetadataStore.prototype = {
...
+   */
+  asyncGetOldestInsert: Task.async(function*() {
+    let timestamp = null;
+    try {
+      const entry = yield this.asyncExecuteQuery("SELECT created_at FROM page_metadata ORDER BY created_at ASC LIMIT 1");
+      if (entry.length) {
+        timestamp = Date.parse(entry[0][0]);
+      }
+    } catch (e) {
+      Cu.reportError(e);
```
ncloudioj: I recall that we used to have a PR to prevent MetadataStore from logging those lengthy errors/exceptions, perhaps just keep silent here, and simply propagate the exception to the client instead.

addon/MetadataStore.js
```diff
@@ -525,6 +525,39 @@ MetadataStore.prototype = {
...
 
   /**
+   * Find the oldest entry in the database
+   *
+   * Returns the timestamp of the oldest entry in the database
+   */
+  asyncGetOldestInsert: Task.async(function*() {
+    let timestamp = null;
+    try {
+      const entry = yield this.asyncExecuteQuery("SELECT created_at FROM page_metadata ORDER BY created_at ASC LIMIT 1");
```
ncloudioj: `ASC` is by default, usually you don't have to specify it.

addon/MetadataStore.js
```diff
@@ -525,6 +525,39 @@ MetadataStore.prototype = {
...
+   * Counts all the items in the database
+   *
+   * Returns a promise with the array of the retrieved metadata records
+   */
+  asyncCountAllItems: Task.async(function*() {
+    try {
+      const result = yield this.asyncExecuteQuery("SELECT count(*) FROM page_metadata");
+      return result;
+    } catch (e) {
+      Cu.reportError(e);
```
ncloudioj: Same here, maybe just re-throw the exception and keep silient?

addon/PlacesProvider.js
```diff
@@ -643,6 +643,28 @@ Links.prototype = {
...
+   *                    that have been entered since the timestamp provided
+   */
+  getHistorySizeSince: Task.async(function*(timestamp) {
+    if (!timestamp) {
+      return Promise.resolve(0);
+    }
+    let sqlQuery = `SELECT count(1)
+                    FROM moz_places WHERE id IN
+                    (SELECT DISTINCT place_id FROM moz_historyvisits
+                    WHERE visit_date /1000 >= ${timestamp})
```
ncloudioj: Let's use parameterized query here, please reference the function `asyncCacheKeyExists`

addon/PlacesProvider.js
```diff
@@ -643,6 +643,28 @@ Links.prototype = {
...
   /**
+   * Gets History size since a certain timestamp
+   *
+   * @param {String} timestamp in milliseconds
+   *
+   * @returns {Promise} Returns a promise with the count of moz_places records
+   *                    that have been entered since the timestamp provided
+   */
+  getHistorySizeSince: Task.async(function*(timestamp) {
+    if (!timestamp) {
```
ncloudioj: This if statement will be unnecessary if `asyncGetOldestInsert` either returns a valid timestamp, or raises an exception.

addon/PlacesProvider.js
```diff
@@ -643,6 +643,28 @@ Links.prototype = {
...
+      return Promise.resolve(0);
+    }
+    let sqlQuery = `SELECT count(1)
```
ncloudioj: s/count(1)/count(*)/

test/test-MetadataStore.js
```diff
@@ -51,6 +52,10 @@ function waitForAsyncReset() {
   }, 10);
 }
 
+function sleep(ms) {
```
ncloudioj: Consider using `waitUntil` from `sdk/test/utils` if you need a timeout.

addon/PlacesProvider.js
sarracini: I believe since we're doing a count here, if there are no results it'll still return a 0. Was going off of how ```getHistorySize``` was doing it (just above this function)

test/test-MetadataStore.js
sarracini: 😭 😭

ncloudioj: Heh, agreed! Better specify the "created_at" manually to save us 9 seconds here 😃. Unfortunately, we need to change the function `getMetadataParameters` to get this field from the argument and use `Date.now()` if it's absent.

sarracini: Is it worth it? I can change the sleep to 1s instead of 3s to save us some time

ncloudioj: I'd say let it sleep for now. We can make this change if manual specifying `created_at` becomes ubiquitous. What do you think? @sarracini

sarracini: Sounds good to me :)

addon/MetadataStore.js
```diff
@@ -525,6 +525,38 @@ MetadataStore.prototype = {
...
 
   /**
+   * Find the oldest entry in the database
+   *
+   * Returns the timestamp of the oldest entry in the database
+   */
+  asyncGetOldestInsert: Task.async(function*() {
+    let timestamp = null;
+    try {
+      const entry = yield this.asyncExecuteQuery("SELECT created_at FROM page_metadata ORDER BY created_at LIMIT 1");
```
ncloudioj: Although they are equivalent,  let's use this more obvious one `SELECT min(created_at) FROM page_metadata`

addon/MetadataStore.js
```diff
@@ -525,6 +525,38 @@ MetadataStore.prototype = {
...
+
+  /**
+   * Counts all the items in the database
+   *
+   * Returns a promise with the array of the retrieved metadata records
+   */
+  asyncCountAllItems: Task.async(function*() {
+    try {
+      const result = yield this.asyncExecuteQuery("SELECT count(*) FROM page_metadata");
+      return result;
```
jaredkerim: We have to unpack this array of arrays here:

 {"source":"PLACES_STATS_FEED","event_id":"{a7e1b003-00d4-f44e-91cd-46459da20e8a}","action":"activity_stream_performance","tab_id":"-5-1","event":"countMetadataURLs","value":[[310]],"client_id":"c3a2a51b-91cf-0243-b882-03763cf9761d","addon_version":"1.1.8","locale":"en-US","page":"NEW_TAB"}

PR #1927, opened 2017-01-05
fix(addon): #1877 Attach number of highlights in session ping

### LINKED ISSUE #1877: Attach number of highlights in session ping
Send the number of highlights of each type that was shown in a session, the total number of highlights, the number of history highlights, the number of bookmark highlights, etc.

ncloudioj:
This closes #1877. 

The related change for the data pipeline is already deployed. We only track the total number of highlights as the first step instead of the breakdown as specified in #1877. The more detailed information could be implemented in the follow-up PRs or possibly by a new ping.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1927)

### DIFF
--- a/addon/TabTracker.js
+++ b/addon/TabTracker.js
@@ -167,6 +167,25 @@ TabTracker.prototype = {
     // in order to provide the more sepcific reasons other than "navigation"
     this._tabData.unload_reason = this._tabData.unload_reason || reason;
 
+    // Attach the number of highlights to the payload. Fetching the highlight count
+    // here other than doing so in onOpen to make it cover all the cases, such as
+    // new tab, refresh, back, and re-activate
+    //
+    // Note: the selector "selectAndDedup" on the content side might filter out some
+    // highlight links if they also belong to the Top Sites. The highlight count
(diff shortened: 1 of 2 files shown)


addon/TabTracker.js
```diff
@@ -167,6 +167,21 @@ TabTracker.prototype = {
...
+    // here other than doing so in onOpen to make it cover all the cases, such as
+    // new tab, refresh, back, and re-activate
+    //
+    // Note: the selector "selectAndDedup" on the content side might filter out some
+    // highlight links if they also belong to the Top Sites. The highlight count
+    // would be greater than the actual value in this case. This discrepancy will
+    // eventually go away as we are phasing out the filtering in the selectAndDedup
+    if (this._store) {
+      const currentState = this._store.getState();
+      this._tabData.highlights_size = currentState.Highlights.error ? -1 : currentState.Highlights.rows.length;
```
rlr: I think you want to also check on `init`. So something like:

```
this._tabData.highlights_size = currentState.Highlights.error || !currentState.Highlights.init ? -1 : currentState.Highlights.rows.length;
```

PR #1931, opened 2017-01-09
feat(places): Fixes #1901 Refresh after clearing history

### LINKED ISSUE #1901: Top sites and Highlights tiles are still populated after deleting all Firefox history
[Affected versions]:
- Nightly 53.0a1
- Activity Stream 1.1.8-dev

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Latest Activity Stream add-on version (1.1.8-dev) was previously installed on a clean profile.
- Have enough history so that Top Sites and Highlights tiles are populated.

[Steps to reproduce]:
1. Start the browser with the profile from prerequisites.
2. Open a new tab and observe the Activity Stream history.
3. Clear all history (Ctrl+Shift+Delete shortcut) and refresh the page.
4. Observe the Activity Stream Top Sites and Highlights tiles.

[Expected results]:
- The default Activity Stream new tab page is displayed.

[Actual results]:
- Activity Stream Top Sites and Highlights tiles are still populated.

[Notes]:
- If you visit a new website, the Highlights tiles will disappear but the Top Sites remain populated.
- The default Activity Stream new tab page is displayed if the browser is restarted.
- Here is a screen recording with the issue: https://goo.gl/UYUURm.

sarracini:
Triggers a refresh when onManyLinksChanged fires - which is an event fired when either the frecencies are invalidated, or when onClearHistory is called. In both cases it would be good to refresh.

Note: There's a clearHistory event which is also triggered when you clear history, but onManyLinksChanged encapsulates the broader case that the frecencies change as well as the clear history. So by watching this event we get both for free.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1931)


addon/ActivityStreams.js
```diff
@@ -293,6 +292,9 @@ ActivityStreams.prototype = {
       case "bookmarkRemoved":
         this.broadcast(am.actions.Response("RECEIVE_BOOKMARK_REMOVED", data));
         break;
+      case "manyLinksChanged":
+        this.broadcast({type: "CLEAR_HISTORY"});
```
k88hudson: I vaguely remember there were a few other reason this could be triggered, like a sync. That's fine since it would also be a valid refresh case, but might be better to name this something more generic and add a comment in the feed (i.e. this gets triggered by .... etc.)

PR #1935, opened 2017-01-09
fix(highlights): Fixes #1932 Cap the retry logic for HIGHLIGHTS_AWAITING_METADATA

### LINKED ISSUE #1932: Cap the retry logic for HIGHLIGHTS_AWAITING_METADATA
Let's cap this to something like 5 times before giving up - or else we run into the possibility of having an infinite loop

I saw this happen when I was doing issue #1901 - Clearing history then visiting a new site causes this event to fire an insane amount.

sarracini:
Cap the number of times we retry to get metadata after a re-install to > 5

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1935)

### DIFF
--- a/content-test/addon/Feeds/HighlightsFeed.test.js
+++ b/content-test/addon/Feeds/HighlightsFeed.test.js
@@ -107,6 +107,22 @@ describe("HighlightsFeed", () => {
           assert.equal(action.type, "HIGHLIGHTS_AWAITING_METADATA");
         });
     });
+    it("should resolve with a HIGHLIGHTS_RESPONSE action if we've tried more than 5 times", () => {
+      instance.retryHighlightsCount = 5;
+      instance.baselineRecommender = {scoreEntries: () => []};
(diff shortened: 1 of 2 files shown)


content-test/addon/Feeds/HighlightsFeed.test.js
```diff
@@ -107,6 +107,15 @@ describe("HighlightsFeed", () => {
           assert.equal(action.type, "HIGHLIGHTS_AWAITING_METADATA");
         });
     });
+    it("should resolve with a HIGHLIGHTS_RESPONSE action if we've tried more than 5 times", () => {
+      instance.retryHighlightsCount = 6;
```
jaredkerim: If you set this to 4 then call getData once to show it returns AWAITING_METADATA and then again to show that it switches to HIGHLIGHTS_RESPONSE that'll test the transition more rigorously.

Issue #1916, opened 2017-01-04
Duplicate tiles in the Topsites


ncloudioj:
Having two exact same tiles (with the same urls) in the Topsites. You can reproduce this by clicking on the  default tiles in the Topsites, and then open another new tab page.

<img width="712" alt="screen shot 2017-01-04 at 10 46 04 am" src="https://cloud.githubusercontent.com/assets/1455254/21648185/0cc9d27c-d26b-11e6-85ea-5ed082f797a5.png">

Mardak:
Looks like `selectAndDedupe` just appends the defaults:

https://github.com/mozilla/activity-stream/blob/05fcc35a6a06edf025be87ad4e844e7dd6dd2f99/common/selectors/selectAndDedupe.js#L22-L23

PR #1938, opened 2017-01-09
fix(content) #1916 Dedupe against defaults in selectAndDedupe

k88hudson:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1938)

### DIFF
--- a/common/selectors/selectAndDedupe.js
+++ b/common/selectors/selectAndDedupe.js
@@ -13,14 +13,12 @@ const dedupe = require("lib/dedupe");
 module.exports = function selectAndDedupe(group) {
   return group.reduce((result, options, index, arr) => {
     let current;
+    let sites = options.defaults ? options.sites.concat(options.defaults) : options.sites;
     if (result.length) {
       const previous = result.reduce((prev, current) => prev.concat(current), []);
-      current = dedupe.group([previous, options.sites])[1];
+      current = dedupe.group([previous, sites])[1];
     } else {
-      current = dedupe.one(options.sites);
-    }
-    if (options.defaults) {
-      current = current.concat(options.defaults);
+      current = dedupe.one(sites);
     }
     if (options.max && current.length > options.max) {
       current = current.slice(0, options.max);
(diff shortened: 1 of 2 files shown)


common/selectors/selectAndDedupe.js
```diff
@@ -13,14 +13,12 @@ const dedupe = require("lib/dedupe");
 module.exports = function selectAndDedupe(group) {
   return group.reduce((result, options, index, arr) => {
     let current;
+    let sites = options.defaults ? options.sites.concat(options.defaults) : options.sites;
```
ncloudioj: I assume `options.defaults` is null or an empty array if `optons.sites` has enough links, correct? Otherwise, the defaults might exclude the actual topsites in some cases.

Mardak: It shouldn't exclude top sites as the default dupes are after the actual top sites via `concat`.

ncloudioj: That's correct. Thanks for clarifying it! @Mardak 

As `group.defaults` is of the array type, I think we can replace the ternary
`let sites = options.defaults ? options.sites.concat(options.defaults) : options.sites;`
with
`let sites = options.sites.concat(options.defaults);`

Also, looks like `group.dedupe` is an unused argument. Let's update the doc for the better readability.

k88hudson: Unfortunately because `option.defaults` is allowed to be `undefined` (for example, there are no defaults for Highlights) `options.sites.concat(options.defaults)` would result in `undefined` being added to the array, so I think the ternary is necessary unfortunately

ncloudioj: Cool. Let's make sure the doc is up-to-date please.

It took me some time to understand this function :). For example, initially, I was assuming the `dedupe.group()` function was implemented as "sort | uniq", which might include the defaults incorrectly in some cases. Fortunately, that's not the case 😄 !

k88hudson: I agree, hopefully we can get rid of this function eventually, in this case I just did the small fix so that the bug can be fixed

ncloudioj: Makes sense! Go ahead merge this fix in :shipit:

Mardak: Could have done this ;) `options.sites.concat(options.defaults || [])`

ncloudioj: Heh, that's idiomatic! Shall we reopen this case? Just kidding 😜

Issue #1929, opened 2017-01-06
Add Enzyme for testing pure components


k88hudson:
We should consider using Enzyme to allow for a way to test stateless functional components, so that we can actually use them.

We might also consider switching to the class-based syntax for components that do require state.

PR #1944, opened 2017-01-11
chore(testing): #1929 Add Enzyme and add functional components

dmose:
This change adds Enzyme so it's straightforward to test stateless, functional components.  

It also rewrites a few test sets to use Enzyme, and changes those components to be functional ones.

Using Enzyme in new tests also means that its possible to rewrite existing components into functional ones without (necessarily) having to change the tests.  As an added bonus, it can also make tests more readable, and making it easy to use shallow rendering is great because it can enforce that tests are actually unit-tests rather than inadvertent integration tests.

See https://medium.com/@housecor/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc for info on why functional components are worthwhile...

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1944)

### DIFF
--- a/content-src/components/HighlightContext/HighlightContext.js
+++ b/content-src/components/HighlightContext/HighlightContext.js
@@ -33,26 +33,17 @@ HighlightContext.propTypes = {
 /**
  * Only display a placeholder version (ie just outlines/shapes), for use
  * before sufficient data is available to display.
- *
- * This should be a stateless, functional component.  Unfortunately, testing
- * these kinda sucks; see
- * http://stackoverflow.com/questions/36682241/testing-functional-components-with-renderintodocument)
- *
- * The bottom line: we should start using Enzyme and switch this over.
  */
-const PlaceholderHighlightContext = React.createClass({
-  mixins: [PureRenderMixin],
-  render() {
-    return (
-      <div className="highlight-context placeholder">
-        <div className="hc-icon">
-          <div className="icon" />
-        </div>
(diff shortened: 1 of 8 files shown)


dmose:
r=@sarracini?  It'll likely be easiest/fastest to do this review one commit at a time.

dmose:
@sarracini Should be ready for review now.  It may also be helpful to view the entire diff without whitespace changes, so that you can see what the semantic changes are:

https://github.com/mozilla/activity-stream/pull/1944/files?w=1

?w=1 also works on the end of individual commit diffs

content-src/components/HighlightContext/HighlightContext.js
```diff
@@ -2,27 +2,23 @@ const React = require("react");
...
+  let timestamp;
+  const type = props.type;
+  const {icon, label, showTimestamp, tooltip} = Object.assign({}, highlightTypes[type], props);
+  if (props.date && showTimestamp !== false) {
+    timestamp = getRelativeTime(props.date);
   }
-});
+  return (<div className="highlight-context tooltip-container">
+    <span className="hc-icon">
+      <span className={`icon icon-${icon}`} /><span className="sr-only">{props.type}</span>
```
sarracini: I believe ```props.type``` can just be ```type``` here since you've pulled it out above

dmose: Fixed!

PR #1946, opened 2017-01-11
fix(topsites): Combine frecency of the top pages of with/without www instead of all pages

### LINKED ISSUE #1898: Recently visited page showed up in the top sites
A recently visited google page (and visited just once) took a slot in my top sites. It's a bit confusing as I just dismissed a few other google pages from the top sites. Two potential issues that we may want to consider:

* Should we treat a random google search page as one of the top sites?
* Should we dedup the dismissed sites?

Mardak:
Fix #1898. Make sure we combine just the top page instead of all pages with a test. r?@ncloudioj

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1946)

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -348,7 +348,8 @@ Links.prototype = {
 
     let blockedURLs = ignoreBlocked ? [] : this.blockedURLs.items().map(item => `"${item}"`);
 
-    // this query does "GROUP BY rev_nowww" (rev_host without www) to remove urls from same domain.
+    // GROUP first by rev_host to get the most-frecent page of an exact host
+    // then GROUP by rev_nowww to dedupe between top two pages of nowww host.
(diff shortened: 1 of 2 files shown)


addon/PlacesProvider.js
```diff
@@ -348,7 +348,8 @@ Links.prototype = {
 
     let blockedURLs = ignoreBlocked ? [] : this.blockedURLs.items().map(item => `"${item}"`);
 
-    // this query does "GROUP BY rev_nowww" (rev_host without www) to remove urls from same domain.
+    // GROUP first by rev_host to get the most-frecent page of an exact host
+    // then GROUP by rev_nowww to dedupe between top two pages of nowww host.
```
ncloudioj: This comment is also a piece of good tongue twister 🥇

PR #1948, opened 2017-01-11
fix(pageScraper): Fixes #1947 Handle image load failures

### LINKED ISSUE #1947: Handle images that won't load in PageScraper
When we parse metadata for pages locally, we also attempt to parse the image size.  However we don't handle the case that the network call for the image fails:

https://github.com/mozilla/activity-stream/blob/master/addon/PageScraper.js#L68-L75

Here we attach an event listener for the "load" event, but if that event doesn't fire then this promise won't resolve and the parsing will be blocked.  So we should find out what other events this might fire are, like failed to load, network error, etc and then reject the promise on those.

sarracini:
Handle the case that the image fails to load

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1948)

### DIFF
--- a/addon/PageScraper.js
+++ b/addon/PageScraper.js
@@ -60,7 +60,7 @@ PageScraper.prototype = {
    * Locally compute the size of the preview image
    */
   _computeImageSize(metadata) {
-    return new Promise(resolve => {
+    return new Promise((resolve, reject) => {
       if (metadata.images.length) {
         const metadataImage = metadata.images[0];
         let image = new Services.appShell.hiddenDOMWindow.Image();
@@ -73,6 +73,7 @@ PageScraper.prototype = {
           };
           resolve([imageWithSize]);
         });
+        image.addEventListener("error", () => reject());
       } else {
         resolve([]);
       }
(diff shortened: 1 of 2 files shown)


addon/PageScraper.js
```diff
@@ -73,6 +73,7 @@ PageScraper.prototype = {
           };
           resolve([imageWithSize]);
         });
+        image.addEventListener("error", () => reject());
```
Mardak: Do we want to `reject` here? We seem to `resolve` with a default of 500x500 above. Also, if we do `reject`, it will `reportError` https://github.com/sarracini/activity-stream/blob/ff85a9960a53704f3c10a700ab04852b1ff0da8e/addon/PageScraper.js#L47 and is that desired?

sarracini: I think this is what we want, since ```the error``` event will fire only fire if something goes wrong while loading, which we probably want to capture. But we could also resolve with an empty array. What do you think? @Mardak

Mardak: I suppose this should be fine for now, and we'll get reports of the error now to better understand when it's failing.

PR #1954, opened 2017-01-12
feat(experiments): Closes #1952 Use orig tiles algorithm for top sites

### LINKED ISSUE #1952: Use the original tiles algorithm to get the top sites
So in the SHIELD study we saw that our top sites were not as successful as the original tiles in terms of click through rates. An easy experiment we can run is the use the original new tab sites' algorithm to power our top sites (the first 6, anyways). That way we can see if our click through rates improve at all. If they don't, it's a good indication that there's something else contributing to the low click through rate for top sites compared to tiles.

cc @emtwo

sarracini:
Use the newtab sites that are returned by ```NewtabUtils.jsm``` which are the same links that about:newtab shows on it's tiles

cc @emtwo 

dxr links: 
https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/NewTabUtils.jsm#920
https://dxr.mozilla.org/mozilla-central/source/browser/base/content/newtab/grid.js#139

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1954)

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -1,4 +1,4 @@
-/* globals XPCOMUtils, Services, gPrincipal, EventEmitter, PlacesUtils, Task, Bookmarks, SyncedTabs */
+/* globals XPCOMUtils, Services, gPrincipal, EventEmitter, PlacesUtils, Task, Bookmarks, SyncedTabs, NewTabUtils */
 
 "use strict";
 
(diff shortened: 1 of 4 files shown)


addon/PlacesProvider.js
```diff
@@ -321,6 +322,12 @@ Links.prototype = {
     }
   },
 
+  getTopNewTabSites() {
```
ncloudioj: Consider using `Task.async(function*())`?

experiments.json
```diff
@@ -103,5 +103,20 @@
...
+    "active": true,
+    "description": "Use the sites from the original about:newtab",
+    "control": {
+      "value": false,
+      "description": "Use sites from tiles"
+    },
+    "variant": {
+      "id": "exp-008-original-newtab-sites",
+      "value": true,
+      "threshold": 0.1,
```
ncloudioj: Just confirmed with @emtwo that we might want to recruit 20% of the users for this experiment.

addon/PlacesProvider.js
sarracini: The only reason I'm using a promise here is because ```getData()``` explicitly needs a promise handed back to it and since ```NewTabUtils.links.getLinks()``` is sync I didn't see a point in using ```Task.async``` since a regular promise is good enough

ncloudioj: Fair enough :). How about let's call it "asyncGetTopNewTabSites" like other functions in this module?

sarracini: sure thing 👍

PR #1955, opened 2017-01-12
chore(prefs): Remove duplicate pref logic

### LINKED ISSUE #1956: Clean up pref stuff
This stuff is a mess

sarracini:
Closes #1956. This was annoying me - there was a bunch of logic for pref listening and stuff that was either 1. old and not being used anymore or 2. in the wrong place and therefore was there twice

This makes things a little easier to understand

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1955)


common/reducers/Prefs.js
```diff
@@ -22,9 +22,6 @@ const INITIAL_STATE = {
 module.exports = function Prefs(prevState = INITIAL_STATE, action) {
   const state = Object.assign({}, prevState);
   switch (action.type) {
-    case am.type("PREFS_REQUEST"):
-      state.isLoading = true;
```
Mardak: Looks like `isLoading` is always `false` now and is effectively unused... And a quick glance seems to show `init` is also unused?

addon/PrefsProvider.js
```diff
@@ -13,51 +13,37 @@ exports.PrefsProvider = class PrefsProvider {
...
-   * @param  {obj} options
-   *         {obj} options.msg     This is the message from content (i.e. the 'action')
-   *         {obj} options.worker  This is the worker to which this.send responds
+   * destroy - Removes the event listener on prefs
    */
-  actionHandler(options = {}) {
-    const {msg: action, worker} = options;
-    switch (action.type) {
-      case "PREFS_REQUEST":
-        this.send(am.actions.Response("PREFS_RESPONSE", simplePrefs.prefs), worker);
```
Mardak: To be clear, content should access prefs by trusting `Prefs` state is correct and just accessing the value? It seems that `Prefs` will be updated whenever we `ActivityStreams._refreshAppState` which happens on each new tab page bind?

content-test/addon/PrefsProvider.test.js
```diff
@@ -4,31 +4,32 @@ const {SimplePrefs} = require("shims/sdk/simple-prefs");
...
   }
 
   beforeEach(() => setup());
 
-  it("should set instance properties", () => {
-    assert.property(prefsProvider, "broadcast");
-    assert.property(prefsProvider, "send");
-  });
+  it("should set instance properties", () => (
+    assert.property(prefsProvider, "broadcast")
```
Mardak: nit: why change this from `{ statement; }` to `(expression)`? Seems like it'll be easy to run into errors/confusion.

addon/PrefsProvider.js
sarracini: Yep, that sounds right. We were never hitting this action manager anyways since we were sending PREFS_RESPONSE down by refreshAppState

common/reducers/Prefs.js
sarracini: Good catch - we can remove this too

content-test/addon/PrefsProvider.test.js
sarracini: The linter :( Since I removed the send test it didn't want me to wrap a one line statement in ```{}```

Mardak: Which rule was this failing? I can't seem to reproduce it. We have plenty of one line statement blocks. https://github.com/mozilla/activity-stream/blob/master/content-test/addon/PerfMeter.test.js#L161-L166

PR #1962, opened 2017-01-16
feat(metadata): Closes #1940 Deprecate Embedly

### LINKED ISSUE #1940: Go away Embedly 💀
Once MetadataService experiment comes back with some good data we can graduate it and get rid of Embedly once and for all!!!

sarracini:
Get rid of Embedly!

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/1962)

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -191,7 +191,7 @@ MetadataStore.prototype = {
       description: aRow.description,
       media_url: aRow.media && aRow.media.url,
       expired_at: aRow.expired_at,
-      metadata_source: aRow.metadata_source || "Embedly",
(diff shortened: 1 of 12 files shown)


addon/MetadataStore.js
```diff
@@ -191,7 +191,7 @@ MetadataStore.prototype = {
       description: aRow.description,
       media_url: aRow.media && aRow.media.url,
       expired_at: aRow.expired_at,
-      metadata_source: aRow.metadata_source || "Embedly",
+      metadata_source: aRow.metadata_source || "MetadataService",
```
jaredkerim: I'd really rather not have this here.  I know it's just replacing the one we already had but MetadataStore shouldn't know about this, this should be in preview provider.

PR #2047, opened 2017-01-19
fix (content): add system fonts for Windows and Ubuntu

rlr:
r? @bryanbell

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2047)


bryanbell:
We definitely need to utilize the host system's default font.

PR #2056, opened 2017-01-19
refactor (addon): reduce some code duplication in PlacesProvider

rlr:
r? @Mardak

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2056)

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -435,9 +435,7 @@ Links.prototype = {
       params: {limit}
     });
 
-    links = this._faviconBytesToDataURI(links);
-    links = this._addETLD(links);
-    return links.filter(link => LinkChecker.checkLoadURI(link.url));
+    return this._processLinks(links);
   }),
 
   /**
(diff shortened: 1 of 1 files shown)


addon/PlacesProvider.js
```diff
@@ -684,6 +676,20 @@ Links.prototype = {
...
+   *
+   * @param {Array} links
+   *            an array containing link objects
+   *
+   * @returns {Array} an array of checked links with favicons and eTLDs added
+   */
+  _processLinks(links) {
+    return this._addETLD(
+      this._faviconBytesToDataURI(links)
+    ).filter(link => LinkChecker.checkLoadURI(link.url));
```
Mardak: This is an odd combination of nesting and chaining, so keeping the original sequential style would be easier to parse. The ordering doesn't quite matter too much right now (although I suppose the filter could remove links first to not even bother with favicon or eTLD), and the sequential statements would keep that clear.

Issue #1988, opened 2017-01-18
Top Sites icon improvement


tspurway:
https://trello.com/c/JzwG7YNu/126-make-top-sites-more-special

The feature will improve the Top Site presentation/visualization.  The Top Site rendering will choose the best quality (high res, min 128px) icon available for the site.  Failing the availability of this quality, fall back on a screenshot of the site, plus any available favicon, which will be overlayed on the top left corner as per spec.

jaredkerim:
TabCenter captures screenshots of tabs using this method:

https://github.com/bwinton/TabCenter/blob/master/vertical-tabbrowser.xml#L791-L831

nchapman:
Icon rules:

1. Top site visuals should fit inside a 96x96px square
2. Large icons with white or transparent borders should add a border
3. Small icons should follow the same rule as small icons in Highlights

**Root Domains**

1. Use a Tippy Top icon centered on the provided background
2. Else use the display icon (apple touch icon) full width and height (96px)
3. Else use high res favicon (96px minimum) full width and full height (96px)
4. Else use screenshot with small favicon in the corner (32px)
5. Else use screenshot with letter + color in the corner

**Non Root Domains**

1. Use screenshot with favicon in the corner (32px)
2. Else use screenshot with letter + color in the corner

PR #2058, opened 2017-01-20
Add screenshots

k88hudson:
This patch adds a `screenshots` experiment, in which screenshots are fetched for some Top Sites and UI is added to display them.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2058)

### DIFF
--- a/addon/Feeds/TopSitesFeed.js
+++ b/addon/Feeds/TopSitesFeed.js
@@ -1,21 +1,42 @@
+/* globals Task */
+const {Cu} = require("chrome");
 const {PlacesProvider} = require("addon/PlacesProvider");
 const Feed = require("addon/lib/Feed");
 const {TOP_SITES_LENGTH} = require("common/constants");
 const am = require("common/action-manager");
-const simplePrefs = require("sdk/simple-prefs");
 const UPDATE_TIME = 15 * 60 * 1000; // 15 minutes
+const getScreenshots = require("addon/lib/getScreenshots");
+
+Cu.import("resource://gre/modules/Task.jsm");
 
 module.exports = class TopSitesFeed extends Feed {
   // Used by this.refresh
   getData() {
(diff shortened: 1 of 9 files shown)


rlr:
There are some travis issues.

addon/Feeds/TopSitesFeed.js
```diff
@@ -1,21 +1,42 @@
...
+      let links = yield experiments.originalNewTabSites ? PlacesProvider.links.asyncGetTopNewTabSites() : PlacesProvider.links.getTopFrecentSites();
+
+      // Get metadata from PreviewProvider
+      links = yield this.options.getCachedMetadata(links, "TOP_FRECENT_SITES_RESPONSE");
+
+      // Get screenshots if the favicons are too small
+      if (experiments.screenshots) {
+        try {
+          links = yield getScreenshots(links, site => {
+            if (site.favicon_height >= 64 && site.favicon_width >= 64) {
```
rlr: We also want to get the screenshots if the URL path isn't `/` (non root domain). see https://github.com/mozilla/activity-stream/issues/1988#issuecomment-273616738

addon/lib/getScreenshots.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = Task.async(function*getScreenshots(sites, condition) {
+  const result = {};
+  for (let site of sites) {
+    // Don't fetch screenshots if the site doesn't meet the conditions
+    if (condition && !condition(site)) {
+      break;
+    }
+
+    try {
+      const dataURI = yield PreviewProvider.getThumbnail(site.url);
```
rlr: Doesn't this mean that each thumbnail will be get serially instead of in parallel? Is there a way to get them all in parallel? Does it even matter?

rlr:
I cherry-picked the first two commits and left the parallel commit out because it wasn't working right. We can add that in later.

Issue #2069, opened 2017-01-24
Generate default site data from tippy-top


k88hudson:
Right now our colors etc. for default sites (in `first-run-data`) are hard-coded, we should probably use `tippy-top-sites` instead. We might want to move the defaults to the add-on side

PR #2070, opened 2017-01-24
fix(content): Bumped tippy-top-sites for better icons

k88hudson:
This patch bumps `tippy-top-sites` to 1.1.1, which includes

- larger icons
- a constant for `IMAGE_SIZE`

I filed https://github.com/mozilla/activity-stream/issues/2069 as a follow-up to change the first-run sites to be generated from tippy-top sites instead of being hardcoded, since this requires somewhat of a larger change.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2070)

### DIFF
--- a/addon/TippyTopProvider.js
+++ b/addon/TippyTopProvider.js
@@ -2,10 +2,9 @@
 "use strict";
 
 const data = require("sdk/self").data;
-const getSiteData = require("common/vendor")("tippy-top-sites").getSiteData;
+const {IMAGE_SIZE, getSiteData} = require("common/vendor")("tippy-top-sites");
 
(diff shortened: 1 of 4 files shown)


pdehaan:
Is this also something we need to tweak? [/addon/ActivityStreams.js:79](https://github.com/mozilla/activity-stream/blob/a28c098498cfd971be978621b2967ad1f77e29db/addon/ActivityStreams.js#L79)

```js
Services.prefs.setIntPref("places.favicons.optimizeToDimension", 64);
```

sarracini:
@pdehaan i don't even know what that pref does!

pdehaan:
Ref: #731

PR #2085, opened 2017-01-26
feat(rtl): Closes #2023 Make Activity Stream RTL

### LINKED ISSUE #2023: RTL CSS
Split from #1981. Make sure our CSS works with right-to-left languages. Check label positions, context menus, icon directions, etc.

sarracini:
One step closer to locales! 
Things to check:
* The loader should be flipped
* The highlight context stuff should be flipped
* The search suggestions (icons, text, etc...) should be flipped 
* The link menu button should be flipped
* The context menu items (icons, text, the way the share panel opens) should be flipped
* Snippets should be flipped
* Flipped ```debug``` button 
* Top sites UI experiment should be all fine
* Anything else I've forgotten
* Added another commit to clean up some unused CSS while I was there

This patch doesn't include stuff in the debug page... but it can

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2085)


k88hudson:
Looks great! The only thing I noticed were the sub-menu for `ContextMenu` under Share still goes out to the left. I think it would be a good idea to add the button to get to the debug page on the opposite side, but it's not necessary to rtl-ize the actualy debug page.

PR #2088, opened 2017-01-26
fix (addon): #2081 use asyncGetTopNewTabSites in screenshots experiment

### LINKED ISSUE #2081: Top Sites don't have existing pinned tiles
We were hoping that by using the Tile's function for fetching Top Sites that we would get pinned tiles for free. So far that doesn't seem to be happening:

Activity Stream with `screenshot` experiment enabled

image

Tiles

image

rlr:
small r? @sarracini 

Fixes #2081

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2088)


rlr:
True. It's like the other experiment needs to be the control group of this one. or something.

PR #2103, opened 2017-01-30
fix (experiments): turn off screenshots experiment

rlr:
r? @sarracini

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2103)

### DIFF
--- a/experiments.json
+++ b/experiments.json
@@ -130,7 +130,7 @@
     "variant": {
       "id": "exp-009-screenshots",
       "value": true,
-      "threshold": 0.2,
+      "threshold": 0,
       "description": "Show screenshots"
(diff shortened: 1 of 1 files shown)


rlr:
@sarracini yep, I did that first but then notice that the thing disappears from `about:config` which kind of sucks because then nobody will be able to enable it for testing.

rlr:
Or maybe do that but only in the release branch?

sarracini:
@rlr i wonder if maybe you commit this directly to the release branch instead of master? and that way we can keep it on on master?

rlr:
ok. let me redo this against release branch 👍

PR #2109, opened 2017-01-31
Version 1.4.0 in package.json

jaredkerim:
Update package.json to latest version in master.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2109)


sarracini:
@jaredkerim this happens automatically when you tag the new version during release, why do we need to do it here also?

PR #2110, opened 2017-01-31
feat(l10n): Closes #2016 Create strings file and implement i18n in React

### LINKED ISSUE #2016: Create strings file and implement i18n in React
Split from #1981. We need to implement i18n in React and move strings to a strings file so that they can be localized.

sarracini:
L10n:
* Removes search strings hard coded in addon and move them into a ```strings.json``` file
* Creates a ```strings.json``` file in ```locales/en-US/``` for all the strings
* Uses ```react-intl``` to format messages that need to be localized
* Updates tests to use Enzyme when possible 
* Uses two helper functions to make Enzyme possible with ```InjectIntl``` for tests that DO NOT use ```renderWithProvider``` or ```mountWithProvider``` since those two functions need a mock store
* Updates ```renderWithProvider``` and ```mountWithProvider``` to include an ```IntlWrapper```

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2110)


dmose:
I'm a bit confused here.  I believe both Pontoon, the existing firefox process, and the hello and devtools processes used .properties files.  Using a .json file may be ok (eg with https://github.com/mozilla/pontoon-to-json ), but it might have consequences on localizer workflow as well as who we can easily absorb data from.

@Standard8, did we get all of our l10n data from the web-l10n group via Pontoon only, or did folks checkin directly to the github repo, and/or did we get some locales from the regular Firefox process too?

One of the side-consequences here is that we need to know what the minimal l10n bar (i.e. how many/which locales) for shipping the new tab page is. With Hello, it wasn't hard to get away with (eg) just the top-10 locales, but I bet we're going to be held to a higher standard for the new tab page.  I'm not sure exactly who sets that standard, though.  Maybe it's @nchapman or someone else on the product team?  Maybe dbolter has some say here?

cc @Mardak

k88hudson:
@dmose we figured we'd do the conversion process / adding pontoon-to-json in a separate process. Someone suggested that we check with the pontoon folks to see if they do support this format before setting up the build process, since it is a possibility. Likely we'll probably use the `.properties` format

Standard8:
> @Standard8, did we get all of our l10n data from the web-l10n group via Pontoon only, or did folks checkin directly to the github repo, and/or did we get some locales from the regular Firefox process too?

All the l10n was via the l10n repo - mostly pontoon although there was one or two locales checking in directly. The repo needs to be based in the special l10n organisation on github, so that the locales have access to it.

We didn't get any locales from the regular Firefox process.

content-src/components/Search/Search.js
```diff
@@ -270,7 +271,7 @@ const Search = React.createClass({
       <div className="search-container" role="presentation" hidden={!this.getDropdownVisible()}>
       <section className="search-title" hidden={!formHistory.concat(suggestions).length}>
         <img id="current-engine-icon" src={currentEngine.iconBuffer} alt={currentEngine.name.charAt(0)} width="16px" height="16px" />
-        {this.props.searchHeader.replace("%S", currentEngine.name)}
+        {this.props.intl.formatMessage({id: "search_header"}).replace("%S", currentEngine.name)}
```
k88hudson: Not sure if we want to maintain consistency with another format for these strings, but couldn't you just use variables for this?

content-src/components/Search/Search.js
```diff
@@ -305,7 +306,7 @@ const Search = React.createClass({
         </ul>
       </section>
       <section className="search-title">
-        <p>{this.props.searchForSomethingWith}<span><b> {searchString} </b></span>with:</p>
+        <p>{this.props.intl.formatMessage({id: "search_for"}).replace("%S", searchString)}</p>
```
k88hudson: same here

package.json
```diff
@@ -74,6 +74,7 @@
     "npm-run-all": "3.1.1",
     "raw-loader": "0.5.1",
     "react-addons-test-utils": "15.3.2",
+    "react-intl": "^2.2.3",
```
k88hudson: can you lock this version?

Issue #2117, opened 2017-02-02
Set-up pontoon-to-json


k88hudson:
After discussing with @mathjazz we've decided it would be best to use the `.properties` string format (so we can add comments, for example), and convert to our messages json format with https://github.com/mozilla/pontoon-to-json. 

This can be added to the `bundle:locales`build task.

sarracini:
Done with #2120

PR #2120, opened 2017-02-02
feat(intl): #2117 convert strings to .properties and add pontoon-to-json

k88hudson:
This patch converts the `strings.json` file to `.properties` and converts them to json using the pontoon-to-json package. Note that right now, all languages are being included in builds.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2120)


locales/en-US/strings.properties
```diff
@@ -0,0 +1,25 @@
...
+dismiss=Dismiss
+delete=Delete from History
+newtab=New Tab
+search_web=Search the Web
+search_for=Search for {search_term} with
+search_settings=Change Search Settings
+search_header={search_engine_name} Search
+loading=Loading...
+welcome_title=Welcome to new tab
+welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily.
```
sarracini: I think the single quote in you've is making this go funky

sarracini: or maybe github is just having a hard time

k88hudson: It's just the syntax highlighting

yamscripts.yml
```diff
@@ -19,7 +19,7 @@ scripts:
     pre: =>clean && =>copyTestImages
     static: cpx "content-src/static/**/*" data/content
   # TODO: make the number of locales configurable. pontoon-to-json has an option for this
-    locales: cpx "locales/**/*" data/locales
+    locales: pontoon-to-json --src locales --dest data/locales
```
Mardak: `pontoon-to-json` generates a single `json` with all `SUPPORTED_LOCALES`, so I suppose one approach is at add-on build time, for each locale, package up the locale with "en-US" as the default, e.g,. "fr" + "en-US" in `fr.json`. This way when building "fr" Firefox, it'll just package `fr.json` and ignore all the other locale files. A slight tweak on this is we could have a smarter `fr.json` at add-on build time by detecting if strings are missing and selectively use "en-US" strings, so there's only one "locale" in a given `json` file, but it's potentially a mix of translated and default strings.

But on the flip side of just packaging all locales and all strings... there's currently ~30 strings and `loop-client-l10n` had ~80 locales, so that's 2400 strings in a combined `locales.json`.

PR #2121, opened 2017-02-03
Implement local execution of mochitests

dmose:
This is adapted from the debugger.html infrastructure, and contains a single browser chrome (mochi)test that checks the URL opened by creation of a new tab.  

Because the activity-stream trunk has a broken export target, if this test had been in place when that change set landed, it would have caught the problem.  :-)

Once that's fixed, it may still be (correctly) broken, because I think our code doesn't set up its URL override quickly enough.  That, however, is a valid failure as well.

To use, review docs/mochitests.md and follow the directions there.  If they're insufficient, I'll fix any problems with them before landing.

r=@k88hudson

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2121)


bin/download-firefox-artifact
```diff
@@ -0,0 +1,56 @@
...
+    hg clone https://hg.mozilla.org/mozilla-central/ "$FIREFOX_PATH"
+    # if somebody cancels (ctrl-c) out of the long download don't continue
+    if [ $? -ne 0 ]; then
+        exit 1;
+    fi
+    cd "$FIREFOX_PATH"
+
+    # Make an artifact build so it builds much faster
+    echo "
+ac_add_options --enable-artifact-builds
```
k88hudson: Since this only runs when you don't have a `mozilla-central` repo yet, do you maybe want to add a comment in the docs perhaps about how to manually add this to a `.mozconfig` file?

docs/mochitests.md
```diff
@@ -0,0 +1,75 @@
...
+
+If you haven't set up the mochitest environment yet, just run this:
+
+```
+./bin/prepare-mochitests-dev
+```
+
+This will set up everything you need. You should run this *every time* to start working on mochitests, as it makes sure your local copy of Firefox is up-to-date.
+
+On the first run, this will download a local copy of Firefox and set up an [artifact build](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Artifact_builds) (just think of a super fast Firefox build). It may take a while (10-15 minutes) to download and build Firefox.
```
k88hudson: Maybe add a note about the fact this downloads to a directory called `mozilla-central` and also what happens if you already have it downloaded

docs/mochitests.md
```diff
@@ -0,0 +1,75 @@
...
+
+Now, you can run the mochitests like this:
+
+```
+cd firefox
```
k88hudson: `cd mozilla-central`

dmose: Fixed.

bin/download-firefox-artifact
dmose: I've add more verbiage explaining what exactly the scriptage does.  I think it's reasonable to assume that if one wants to switch an existing repo over to artifact building, they'll just click on the link to artifact builds already in the doc.

docs/mochitests.md
dmose: Done.

docs/mochitests.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+./bin/prepare-mochitests-dev
+```
+
+This will set up everything you need. You should run this *every time* you start working on mochitests, as it makes sure your local copy of Firefox is up-to-date.
+
+On the first run, if you don't already have a mozilla-central repo as a sibling of your activity stream repo, this will download one and set up an [artifact build](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Artifact_builds) (just think of a super fast Firefox build). It may take a while (10-15 minutes) to download and build Firefox.
+
+If you do already have a mozilla-central repo, the script ask you if you're ok with losing any local changes in that repo, and, if so, it will merely update to  the latest bits and then export your current activity-stream repo to that
```
k88hudson: Awesome, this is very clear 👍

Issue #2027, opened 2017-01-18
i18n-related QA


k88hudson:
Split from #1981. Review UI for: 
- quality of template strings
- languages with longer strings  (e.g. German)
- languages with shorter strings (e.g. Mandarin)
- RTL languages

flodolo:
A few montes on the strings landed in #2110 

```
"bookmark": "Bookmark",
```
This is a tricky one: is it a verb (e.g. a button label, "add to bookmark") or a noun ("a bookmark")? We'll need comments for cases like this, and it looks like this format doesn't provide them.

```
"email_link": "Email Link...",
"loading": "Loading...",
```
You should use the unicode … (that's the character used in Firefox), unless there are technical limitations.

```
"search_for": "Search for {search_term} with",
```
This assumes that all languages can use this same grammatical structure as English. The name of the search provider should be a placeholder.

For reference, this is the string in Firefox
https://hg.mozilla.org/releases/comm-aurora/file/default/suite/locales/en-US/chrome/browser/navigator.properties#l45

If you’re doing something fancy, like using an image for the search engine, there should be a clear explanation in a comment.

k88hudson:
Thanks for the feedback @flodolo!

k88hudson:
Feedback addressed in  f1e82ec3cebb811d2551e067ecf59dafd5d134b8 @sarracini sarracini committed 5 days ago

Mardak:
No formal review step needed, so we'll untrack this issue.

PR #2124, opened 2017-02-03
feat(l10n): Closes #2122 Add comments to strings.properties

### LINKED ISSUE #2122: Add comments to strings.properties
Some strings could use some contextual hints for our localizers. We should add some comments to help with that

sarracini:
Addresses the first few i18n comments mentioned here: https://github.com/mozilla/activity-stream/issues/2027#issuecomment-277087062
* Adds comments describing the strings for translators
* Uses unicode … rather than ```...```

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2124)

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -1,29 +1,57 @@
-top_sites=Top Sites
-highlights=Highlights
-visited=Visited
-bookmarked=Bookmarked
-synced=Synced from another device
-open=Open
-topic=Topic
-bookmark=Bookmark
-remove_bookmark=Remove Bookmark
-share=Share
-copy_address=Copy Address
-email_link=Email Link...
-open_new_window=Open in a New Window
-open_private_window=Open in a Private Window
-dismiss=Dismiss
-delete=Delete from History
-newtab=New Tab
-search_web=Search the Web
-search_for=Search for {search_term} with
-search_settings=Change Search Settings
+newtab_page_title=New Tab
+default_label_loading=Loading…
+
+header_top_sites=Top Sites
+header_highlights=Highlights
+
+# LOCALIZATION NOTE(type_label_*): These labels are shown with a page to give
+# context on how the page is related to the user
+# e.g. The type indicates that the page is bookmarked, or is currently open on
+# another device
+type_label_visited=Visited
+type_label_bookmarked=Bookmarked
+type_label_synced=Synced from another device
+type_label_open=Open
+type_label_topic=Topic
+
+# LOCALIZATION NOTE(menu_action_*): These shown in a context menu and are meant
+# as a call to action for a given page
+menu_action_bookmark=Bookmark
+menu_action_remove_bookmark=Remove Bookmark
+menu_action_share=Share
+menu_action_copy_address=Copy Address
+menu_action_email_link=Email Link…
+menu_action_open_new_window=Open in a New Window
+menu_action_open_private_window=Open in a Private Window
+menu_action_dismiss=Dismiss
+menu_action_delete=Delete from History
+
+# LOCALIZATION NOTE (search_for_something_with): {search_term} is a placeholder for
+# what the user has typed in the search input field.
+# e.g. 'Search for ' + search_term + 'with:' becomes 'Search for abc with:'
+# The search engine name is displayed as an icon and does not need a translation
+search_for_something_with=Search for {search_term} with:
+
+# LOCALIZATION NOTE (search_header): Displayed at the top of the panel
+# showing search suggestions. {search_term} is replaced with the name of the current
+# default search engine. e.g. 'Google Search'
 search_header={search_engine_name} Search
-loading=Loading...
+
+# LOCALIZATION NOTE (search_web_placeholder): This is shown in the searchbox when
+# the user hasn't typed anything yet.
+search_web_placeholder=Search the Web
+search_settings=Change Search Settings
+
+# LOCALIZATION NOTE (welcome_*): This is shown as a modal dialog typically on a
+# first-run experience when the data is not ready to show yet
 welcome_title=Welcome to new tab
-welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily.
+welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you’ve recently visited, so you can get back to them easily.
 welcome_label=Identifying your Highlights
-less_than_minute_label=<1m
-minute_label={number}m
-hour_label={number}h
-day_label={number}d
+
+# LOCALIZATION NOTE (time_label_*): {number} is a placeholder for a number which reprents a
+# shortened timestamp format e.g. '10m' represents '10 minutes ago'. Do not translate 'm'
+# into 'minutes ago'
+time_label_less_than_minute=<1m
+time_label_minute={number}m
+time_label_hour={number}h
+time_label_day={number}d
--- a/content-src/components/HighlightContext/types.js
+++ b/content-src/components/HighlightContext/types.js
@@ -1,22 +1,22 @@
 module.exports = {
   history: {
-    intlID: "visited",
+    intlID: "type_label_visited",
     icon: "historyItem"
   },
   open: {
-    intlID: "open",
+    intlID: "type_label_open",
     icon: "tab"
   },
   synced: {
-    intlID: "synced",
+    intlID: "type_label_synced",
     icon: "sync"
   },
   bookmark: {
-    intlID: "bookmarked",
+    intlID: "type_label_bookmarked",
     icon: "bookmark"
   },
   topic: {
-    intlID: "topic",
+    intlID: "type_label_topic",
     icon: "topic"
   }
 };
--- a/content-src/components/LinkMenu/LinkMenu.js
+++ b/content-src/components/LinkMenu/LinkMenu.js
@@ -50,13 +50,13 @@ const LinkMenu = React.createClass({
     return [
       {
         ref: "copyAddress",
-        label: this.props.intl.formatMessage({id: "copy_address"}),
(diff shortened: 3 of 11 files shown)


locales/en-US/strings.properties
```diff
@@ -5,25 +5,35 @@ bookmarked=Bookmarked
...
 search_for=Search for {search_term} with
 search_settings=Change Search Settings
+# LOCALIZATION NOTE: {search_engine_name} is a placeholder for their current search
+# engine name: search_engine_name + ' Search'
+# e.g. 'Google Search'
 search_header={search_engine_name} Search
-loading=Loading...
+loading=Loading…
 welcome_title=Welcome to new tab
 welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily.
```
Mardak: The apostrophe for `you're` are usually with `’`, e.g., https://dxr.mozilla.org/mozilla-central/source/browser/locales/en-US/chrome/browser/browser.properties#24

locales/en-US/strings.properties
```diff
@@ -5,25 +5,35 @@ bookmarked=Bookmarked
...
 open_private_window=Open in a Private Window
 dismiss=Dismiss
 delete=Delete from History
 newtab=New Tab
 search_web=Search the Web
+# LOCALIZATION NOTE: {search_term} is a placeholder for what the user has inputed
+# in the search input field: 'Search for ' + search_term + 'with'. The search engine
+# name is displayed as an icon and does not need a translation
+# e.g. 'Search for abc with'
 search_for=Search for {search_term} with
```
Mardak: In particular for these search related strings, we can build off existing localization notes. Also, the original string seems to end with a colon `:` (and that makes it seem more clear the whole string is a phrase as opposed to a partial sentence that looks like it would be concatenated with other words): https://dxr.mozilla.org/mozilla-central/source/browser/locales/en-US/chrome/browser/search.properties#36-40

locales/en-US/strings.properties
```diff
@@ -5,25 +5,35 @@ bookmarked=Bookmarked
 synced=Synced from another device
 open=Open
 topic=Topic
+# LOCALIZATION NOTE: 'bookmark' is an action meant as 'add to bookmarks'
```
Mardak: Typically localization notes look like `LOCALIZATION NOTE (identifier):` to be clear what it's referring to. Also, we could probably avoid some of these types of notes with more explicit identifiers, e.g., `menu_action.bookmark` vs `type_label.synced`. Multi-word identifiers also make it much easier to search for the string in the code, e.g., "open" probably will have a bunch of false positive matches from a simple `grep`.

locales/en-US/strings.properties
```diff
@@ -5,25 +5,35 @@ bookmarked=Bookmarked
...
+loading=Loading…
 welcome_title=Welcome to new tab
 welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily.
 welcome_label=Identifying your Highlights
 less_than_minute_label=<1m
+# LOCALIZATION NOTE: {number} is a place holder for a number which reprents a timestamp
+# e.g. '10m' represents '10 minutes ago'
 minute_label={number}m
 hour_label={number}h
 day_label={number}d
```
Mardak: These identifiers could be better grouped together with some prefix, e.g., `time_label.minute`, and then referring to them all with a `LOCALIZATION NOTE (time_label.*)`

I'm not sure if it's specifically designed this way, but if it is, we should probably note that this should be a short string to avoid accidental localization to "{number}minutes"

locales/en-US/strings.properties
```diff
@@ -5,25 +5,35 @@ bookmarked=Bookmarked
 synced=Synced from another device
```
Mardak: As noted elsewhere, a more descriptive identifier would better identify the usage of these strings, e.g., `section_header.top_sites`, `type_label.visited`

locales/en-US/strings.properties
sarracini: @Mardak what do you think of this:
```.properties
# LOCALIZATION NOTE (time_label_*): {number} is a placeholder for a number which reprents a
# shortened timestamp format e.g. '10m' represents '10 minutes ago'. Do not translate 'm'
# into 'minutes ago'
time_label_less_than_minute=<1m
time_label_minute={number}m
time_label_hour={number}h
time_label_day={number}d
```

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
-share=Share
-copy_address=Copy Address
-email_link=Email Link...
-open_new_window=Open in a New Window
-open_private_window=Open in a Private Window
-dismiss=Dismiss
-delete=Delete from History
+header_top_sites=Top Sites
+header_highlights=Highlights
+type_label_visited=Visited
```
Mardak: At minimum add a empty line here to separate `header` from `type_label`. We should probably have a note describing what a "type_label" is used for, e.g., "These labels are shown with a page to explain how it's related to the user."

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
+type_label_bookmarked=Bookmarked
+type_label_synced=Synced from another device
+type_label_open=Open
```
Mardak: "Open" does seem a bit ambiguous here especially that the previous ones weren't "Visit" and "Bookmark" so this could have been "Opened". Perhaps we could just include it as an example in the above "LOCALIZATION NOTE (type_label_*): ...<general description>... E.g., the type indicates that the page is bookmarked or is currently open."

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
+type_label_topic=Topic
+menu_action_bookmark=Bookmark
```
Mardak: Similarly empty line here to separate out and probably good to have a grouping description. "These following actions are shown in a menu for a given page."

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
+menu_action_remove_bookmark=Remove Bookmark
+menu_action_share=Share
+menu_action_copy_address=Copy Address
+menu_action_email_link=Email Link…
+menu_action_open_new_window=Open in a New Window
+menu_action_open_private_window=Open in a Private Window
+menu_action_dismiss=Dismiss
+menu_action_delete=Delete from History
 newtab=New Tab
```
Mardak: Might as well touch this identifier too! 😃 This string is used for the page title, so maybe `newtab_page_title`?

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
 search_web=Search the Web
```
Mardak: This is used as a placeholder, so maybe `search_web_placeholder` with a NOTE similar to the existing https://dxr.mozilla.org/mozilla-central/source/browser/locales/en-US/chrome/browser/search.properties#7-8

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
+# e.g. 'Search for ' + search_term + 'with:' becomes 'Search for abc with:'
+# The search engine name is displayed as an icon and does not need a translation
+search_for_something_with=Search for {search_term} with:
 search_settings=Change Search Settings
+# LOCALIZATION NOTE (search_header): displayed at the top of the panel
+# showing search suggestions. {search_term} is replaced with the name of the current
+# default search engine. e.g. 'Google Search'
 search_header={search_engine_name} Search
-loading=Loading...
+loading=Loading…
```
Mardak: Where does this get used? I think it's for the page title, but I can't seem to find the reference? If it is, perhaps move it next to the `newtab_page_title` and update the identifier.

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
 welcome_title=Welcome to new tab
```
Mardak: Empty line and note about this "welcome" section describing when and how the welcome message dialog is shown, i.e., typically on first run when there hasn't been enough data processed yet?

locales/en-US/strings.properties
```diff
@@ -1,29 +1,39 @@
...
-welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you've recently visited, so you can get back to them easily.
+welcome_body=Firefox will use this space to show your most relevant bookmarks, articles, videos, and pages you’ve recently visited, so you can get back to them easily.
 welcome_label=Identifying your Highlights
-less_than_minute_label=<1m
-minute_label={number}m
-hour_label={number}h
-day_label={number}d
+# LOCALIZATION NOTE (time_label_*): {number} is a placeholder for a number which reprents a
```
Mardak: newline before! :)

locales/en-US/strings.properties
sarracini: I think "Open" was meant to indicate a currently opened tab on another device. Though I don't think we're ever actually checking for an open tabs on other devices so maybe it doesn't really matter right now?

locales/en-US/strings.properties
sarracini: Oh this was meant as a default prop for Loader.js but I forgot to put a reference to it, will do that!

locales/en-US/strings.properties
```diff
@@ -1,29 +1,57 @@
...
+menu_action_dismiss=Dismiss
+menu_action_delete=Delete from History
+
+# LOCALIZATION NOTE (search_for_something_with): {search_term} is a placeholder for
+# what the user has typed in the search input field.
+# e.g. 'Search for ' + search_term + 'with:' becomes 'Search for abc with:'
+# The search engine name is displayed as an icon and does not need a translation
+search_for_something_with=Search for {search_term} with:
+
+# LOCALIZATION NOTE (search_header):Displayed at the top of the panel
```
Mardak: nit: Space after `:`

Issue #2123, opened 2017-02-03
Add docs for localization process


k88hudson:
This will be helpful for others implementing new strings, as well as if anyone needs to debug any step in the process.

sarracini:
Closed with #2130

PR #2130, opened 2017-02-06
docs(l10n): #2123 Add docs for localization

k88hudson:
This patch adds documentation for how to localize new strings, as well as how to build a version of the add-on with different locales.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2130)

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -1,0 +1,52 @@
+# Localization
+
+## When should you localize strings?
+
+Any time you add human-readable text to the add-on, whether it is plain text, labels, or text only available to screen-readers, you should localize those strings.
+
+## How update localized strings
+
+If you just are fixing a typo in an existing English string, you can go ahead and make the change in `locales/en-US/strings.properties`. If you are changing the *meaning* of an existing string such that other translations need to be re-done, you should also change the *id* of that string in order to invalidate it in other languages. (see [this gudie](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings) for more info).
+
+## How to localize strings
+
+First, add the new string to `locales/en-US/strings.properties`. You do NOT need to add strings to other language files. Try to use an id that is descriptive of the function/role in the application, and consider adding comments for localizers when necessary. See [this guide](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices
+) for more tips on localization best practices.
+
+Note that you must **re-build the add-on** every time you add new strings (the actual build task that processes the strings.properties file is `bundle:locales`, so you need to re-run `npm start`).
+
+On the add-on, you can access strings directly on the redux state object. On the React side, You have two options for implementing the strings:
+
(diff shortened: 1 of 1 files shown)


docs/localization.md
```diff
@@ -0,0 +1,47 @@
...
+
+By default, running `npm run package`, it will build the add-on with *all* locales. This might be useful if you want to QA several locales with a single build.
+
+If you would like to build the add-on with a specific set of locales, you can add a `SUPPORTED_LOCALES` variable to your environment. For example:
+
+```
+SUPPORTED_LOCALES=en-US,de npm run package
+```
+
+will run a build of the add-on with German and English.
```
Mardak: Not sure if it's needed to note that "en-US" should be included as it's used as a default. Or this might change, so not worth noting now?

docs/localization.md
```diff
@@ -0,0 +1,47 @@
...
+
+Any time you add human-readable text to the add-on, whether it is plain text, labels, or text only available to screen-readers, you should localize those strings.
+
+## How to localize strings
+
+First, add the new string to `locales/en-US/strings.properties`. You do NOT need to add strings to other language files.
+
+Note that you must **re-build the add-on** every time you add new strings (the actual build task that processes the strings.properties file is `bundle:locales`, so you need to re-run `npm start`).
+
+On the React side, You have two options for implementing the strings:
```
Mardak: Do we have access to strings from the add-on side? Seems like it gets stashed on the state object.

docs/localization.md
```diff
@@ -0,0 +1,47 @@
+# Localization
+
+## When should you localize strings?
+
+Any time you add human-readable text to the add-on, whether it is plain text, labels, or text only available to screen-readers, you should localize those strings.
+
+## How to localize strings
+
+First, add the new string to `locales/en-US/strings.properties`. You do NOT need to add strings to other language files.
```
Mardak: Maybe at minimum a note about picking a string identifier as preferring more verbose with a "LOCALIZATION NOTE" as localizers might not be able to run the add-on to see how the strings are used, so whatever context you can provide will be helpful. Link to a lot more details: https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices

docs/localization.md
k88hudson: Yes, good idea

docs/localization.md
k88hudson: yep! it's in the state

docs/localization.md
```diff
@@ -0,0 +1,52 @@
+# Localization
+
+## When should you localize strings?
+
+Any time you add human-readable text to the add-on, whether it is plain text, labels, or text only available to screen-readers, you should localize those strings.
+
+## How update localized strings
+
+If you just are fixing a typo in an existing English string, you can go ahead and make the change in `locales/en-US/strings.properties`. If you are changing the *meaning* of an existing string such that other translations need to be re-done, you should also change the *id* of that string in order to invalidate it in other languages. (see [this gudie](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings) for more info).
```
pdehaan: typo: "guide"

PR #2135, opened 2017-02-07
String change to be consistent with desktop

da2x:
Firefox desktop uses the strings `Open Link in a New Window` and `Open Link in a New Private Window`. Change to make string consistent with desktop.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2135)


Mardak:
I believe typically string changes need a id bump, e.g., `menu_action_open_private_window2`. As localizers haven't started localizing yet, we should be okay for now if we push the new string to the strings repository.

Issue #2129, opened 2017-02-06
Support centralized fallbacks for different locales


k88hudson:
We should fall back to `en` strings if any are missing/untranslated.

sarracini:
Fixed with #2139

PR #2139, opened 2017-02-07
fix(l10n): #2129 Fall back to en-US strings if some are missing

k88hudson:
This patch allows `en-US`strings to be fallbacks for the selected locale.

This is kind of hard to manually test right now – check out the unit tests for an example for how it works.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/activity-stream/2139)

### DIFF
--- a/addon/Feeds/LocalizationFeed.js
+++ b/addon/Feeds/LocalizationFeed.js
@@ -2,10 +2,18 @@ const {PrefsTarget} = require("sdk/preferences/event-target");
 const {findClosestLocale, getPreferedLocales} = require("sdk/l10n/locale");
 const Feed = require("../lib/Feed");
 const am = require("../../common/action-manager");
-
+const DEFAULT_LOCALE = "en-US";
 const STRINGS = require("../../data/locales/locales.json");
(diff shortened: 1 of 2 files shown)


addon/Feeds/LocalizationFeed.js
```diff
@@ -2,10 +2,17 @@ const {PrefsTarget} = require("sdk/preferences/event-target");
...
 const STRINGS = require("../../data/locales/locales.json");
 const AVAILABLE_LOCALES = Object.keys(STRINGS);
-const DEFAULT_LOCALE = "en-US";
+
+function getLocalizedStrings(locale, allStrings = STRINGS) {
+  if (locale === DEFAULT_LOCALE) {
+    return allStrings[DEFAULT_LOCALE];
+  }
+  const strings = allStrings[locale];
+  return Object.assign({}, allStrings[DEFAULT_LOCALE], strings || {});
```
sarracini: maybe add a comment to this line that says this is in order to merge on missing strings?

k88hudson: yep!

PR #2143, opened 2017-02-08
Improve localization comments in strings.properties

flodolo:


locales/en-US/strings.properties
```diff
@@ -4,18 +4,20 @@ default_label_loading=Loading…
...
-# context on how the page is related to the user
-# e.g. The type indicates that the page is bookmarked, or is currently open on
-# another device
+# LOCALIZATION NOTE (type_label_*): These labels are associated to pages to give
+# context on how the element is related to the user, e.g. type indicates that
+# the page is bookmarked, or is currently open on another device
 type_label_visited=Visited
 type_label_bookmarked=Bookmarked
 type_label_synced=Synced from another device
+# LOCALIZATION NOTE(type_label_open): Open is an adjective, as in "page is open"
```
flodolo: Rationale for adding a comment before the problematic string is that tools will only display the general comment with the following string, not the entire group.

locales/en-US/strings.properties
```diff
@@ -4,18 +4,20 @@ default_label_loading=Loading…
...
 type_label_open=Open
 type_label_topic=Topic
 
-# LOCALIZATION NOTE(menu_action_*): These shown in a context menu and are meant
-# as a call to action for a given page
+# LOCALIZATION NOTE (menu_action_*): These strings are displayed in a context
+# menu and are meant as a call to action for a given page.
+# LOCALIZATION NOTE (menu_action_bookmark): Bookmark is a verb, as in "Add to
+# bookmarks"
```
flodolo: Same here

locales/en-US/strings.properties
k88hudson: Ah, makes sense. I will add this to our docs

k88hudson:
Thank you!

PR #2149, opened 2017-02-08
feat(l10n): Closes #2136 Determine dir of html

### LINKED ISSUE #2136: Determine dir of html
Firefox probably has a way to check if the locale being used is an RTL or LTR language

sarracini:
* Determine which language is an RTL language and set the html's dir based on that


addon/Feeds/LocalizationFeed.js
```diff
@@ -15,6 +16,10 @@ function getLocalizedStrings(locale, allStrings = STRINGS) {
   return Object.assign({}, allStrings[DEFAULT_LOCALE], strings || {});
 }
 
+function getDirection(locale) {
+  return (RTL_LIST.indexOf(locale.substr(0, 2)) >= 0) ? "rtl" : "ltr";
```
Mardak: We should split on "-" as some locales are actually 3 characters. (Although l10.js seems to compare the full locale string.)

PR #2150, opened 2017-02-08
feat(tests): Closes #2145 Migrate private browsing test to mochitest

### LINKED ISSUE #2145: Migrate jpm tests: Private Browsing
Split from #2000. Convert ```test-ActivityStreams-private-browsing.js``` to a mochitest

sarracini:
A mochitest for private browsing. Funnily enough, the actual test ```test-ActivityStreams-private-browsing.js``` didn't actually test anything.


test/functional/mochitest/browser_as_private_browsing.js
```diff
@@ -0,0 +1,30 @@
...
+add_task(function*doesnt_load_in_private_windows() {
+  const asURL = "resource://activity-streams/data/content/activity-streams.html#/";
+  const privateBrowsingURL = "about:privatebrowsing";
+
+  // create a private browsing window
+  let privateWindow = yield BrowserTestUtils.openNewBrowserWindow({private: true});
+  let tab = privateWindow.gBrowser.selectedBrowser;
+
+  // try to load activity stream's url into the private window
+  tab.loadURI(asURL);
```
dmose: https://wiki.mozilla.org/Electrolysis/e10s_test_tips seems to suggest that the right thing to do is use BrowserTestUtils.loadURI on the XUL browser element so that we don't run in to synchronicity problems which could cause intermittent failures.

test/functional/mochitest/browser_as_private_browsing.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Tests that loading activity streams in private browsing mode doesn't actually
+ * load the activity streams URL and instead loads about:privatebrowsing
+ */
+add_task(function*doesnt_load_in_private_windows() {
+  const asURL = "resource://activity-streams/data/content/activity-streams.html#/";
+  const privateBrowsingURL = "about:privatebrowsing";
+
+  // create a private browsing window
+  let privateWindow = yield BrowserTestUtils.openNewBrowserWindow({private: true});
```
dmose: https://dxr.mozilla.org/mozilla-beta/source/testing/mochitest/BrowserTestUtils/BrowserTestUtils.jsm#471 makes me wonder if it's worth running this test both with and without remote set to true so that we guarantee correctness both with and without e10s.

test/functional/mochitest/browser_as_private_browsing.js
```diff
@@ -0,0 +1,30 @@
...
+  let tab = privateWindow.gBrowser.selectedBrowser;
```
dmose: https://dxr.mozilla.org/mozilla-beta/source/testing/mochitest/BrowserTestUtils/BrowserTestUtils.jsm#471 makes me wonder if gBrowser could be null and that we should use firstBrowserLoaded just to be safe here.  Admittedly, I'd kinda expect the previous yield to have avoided that problem, but...

test/functional/mochitest/browser_as_private_browsing.js
```diff
@@ -0,0 +1,30 @@
...
+
+  // try to load activity stream's url into the private window
+  tab.loadURI(asURL);
+  yield BrowserTestUtils.browserLoaded(tab);
+
+  // check what the content task thinks has been loaded
+  yield ContentTask.spawn(tab, {url: privateBrowsingURL}, args => {
+    Assert.equal(content.location.href, args.url, "Correctly loaded about:privatebrowsing instead of activity-streams");
```
dmose: From poking around, it looks like content.location.href should reference the right thing, but I haven't yet found where that's documented or implemented.  Do you happen to know?

test/functional/mochitest/browser_as_private_browsing.js
dmose: It looks to me like all the various tranches of tests get run both with and without e10s, so this probably isn't necessary.  Might be worth asking mconley just to be sure.

test/functional/mochitest/browser_as_private_browsing.js
sarracini: I haven't seen it formally documented anywhere either...just kind of went off what other tests were doing in dxr

test/functional/mochitest/browser_as_private_browsing.js
sarracini: When I add an extra ```yield firstBrowserLoaded(privateWindow)``` the test hangs, I'm assuming because the load event has already fired so the promise just hangs

PR #2153, opened 2017-02-09
feat(a11y): Closes #1993 Bring '...' button to foreground

### LINKED ISSUE #1993: "..." button is not brought in foreground when it is selected using the "Tab" key
**[Affected versions]:**
-Firefox 45.0 and up
-Activity Stream Dev 1.3.0-dev-1484695187.

**[Affected Platforms]:**
-All Windows

**[Prerequisites]:**
-Latest Activity Stream add-on version (1.3.0-dev) was previously installed on a profile.
-Have more history so that the "Highlights" area is populated with visited websites.

**[Steps to reproduce]:**
2. Open a New Tab so that Activity Stream page is displayed.
3. Drag&Drop one of the Highlights tiles.
4. Press "Tab" key from keyboard.

**[Expected result]:**
-"..." button is brought in foreground and it is selected.

**[Actual result]:**
-"..." button is selected but is hidden.

**[Notes]:**
- [Here](https://drive.google.com/open?id=0B6QogV2AJFKiY0tuT2dGaGNteVU) is a video of the issue.
- If the user cycles through Highlight tiles using the "Tab" key, then each selected Highlight tile will have a text cursor displayed in the left part of the favicon. (can be seen in the video)
- If the context menu is display, it cannot be closed with the "Esc" key.
- These scenarios are reproducible for Top Sites and Highlights.
- Related with #1978 .

sarracini:
This diff is really hard to read but basically what I did was:
1. Move ```.tile``` so that it's nested in ```.tile-outer```. This allowed for some of the focus stuff to work properly
2. Add a new mixin for these "inner" elements, ```.tile``` and ```.spotlight-inner``` which keeps the border and shows the link menu button when you tab through it

The best way to test this is to just run this and start tabbing through the top sites and highlights. What happens is when you tab through it, it'll show focus in 3 ways: 1. the border shows up 2. background colour of ```.site-icon-title``` turns white and 3. the menu link button shows up. Another tab brings the menu link button into focus and if you press enter it'll open the context menu. This works for both top sites and highlights.

Note: In the original ticket #1993 it was mentioned that if you press "ESC" then the context menu doesn't close. That behaviour is still happening here, I figured we could fix that when we make the context menu accessible in general

### DIFF
--- a/content-src/components/LinkMenuButton/LinkMenuButton.scss
+++ b/content-src/components/LinkMenuButton/LinkMenuButton.scss
@@ -18,4 +18,10 @@
   transition-property: transform, opacity;
   transition-duration: 200ms;
   z-index: 399;
+
+  &:focus,
+  &:active {
+    transform: scale(1);
+    opacity: 1;
+  }
 }
--- a/content-src/components/Spotlight/Spotlight.js
+++ b/content-src/components/Spotlight/Spotlight.js
@@ -46,7 +46,7 @@ const SpotlightItem = React.createClass({
       }));
     }
     return (<li className={classNames("spotlight-item", {active: this.state.showContextMenu})}>
-      <a onClick={this.props.onClick} href={site.url} ref="link">
+      <a onClick={this.props.onClick} className="spotlight-inner" href={site.url} ref="link">
         <div className={classNames("spotlight-image", {portrait: isPortrait})} style={style} ref="image">
           <SiteIcon className="spotlight-icon" height={40} width={40} site={site} ref="icon" showBackground={true} border={false} faviconSize={32} />
         </div>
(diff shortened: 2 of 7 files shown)


content-src/components/LinkMenuButton/LinkMenuButton.scss
```diff
@@ -18,4 +18,10 @@
   transition-property: transform, opacity;
   transition-duration: 200ms;
   z-index: 399;
+
+  &:focus,
+  &:active {
+    transform: scale(1);
```
k88hudson: what is this for?

k88hudson:
Cool, looks good to me.

content-src/components/LinkMenuButton/LinkMenuButton.scss
sarracini: when the link-menu-button is activated through tabbing it disappears without this little bit of css

k88hudson: ah, ok

PR #2156, opened 2017-02-09
Force 1 content process in dev prefs fixes #2155

### LINKED ISSUE #2155: Force 1 content process in dev prefs
In Firefox 54 Nightly, there are now 2 content processes by default (previously all browsers used 1).  Because of a bug https://bugzilla.mozilla.org/show_bug.cgi?id=1285898 the localstorage across content processes is out of sync which causes activity stream to appear inconsistently broken because we store all of our state in local storage.  We can temporarily force nightly to use 1 content process in dev prefs for our local testing so that people do not accidentally rediscover this bug and go hunting around for it when it is already a known issue.

jaredkerim:
https://github.com/mozilla/activity-stream/issues/2155

### DIFF
--- a/dev-prefs.json
+++ b/dev-prefs.json
@@ -1,4 +1,5 @@
 {
+  "dom.ipc.processCount": 1,
   "javascript.options.strict": false,
   "xpinstall.signatures.required": false,
   "extensions.@activity-streams.sdk.console.logLevel": "info",


jaredkerim:
@dmose it is so filed! https://github.com/mozilla/activity-stream/issues/2157

Issue #2113, opened 2017-02-02
action.data is null


k88hudson:
I noticed that my Activity Stream on Windows last night was missing highlights for awhile. The Highlights property on redux state was an empty array, and I found the following error in my browser chrome console:

```
TypeError: action.data is null: setRowsOrError/</state.rows<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/reducers/SetRowsOrError.js:41:1
setRowsOrError/<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/reducers/SetRowsOrError.js:40:22
combination@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/vendor.js:971:30
_mergeStateReducer/<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/create-store.js:84:12
dispatch@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/vendor.js:510:23
FeedController/this.reduxMiddleware/</<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/addon/lib/FeedController.js:11:7
module.exports/</<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/parse-url-middleware.js:6:12
module.exports</createThunkMiddleware/</</<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/common/vendor.js:1193:17
broadcast@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/addon/ActivityStreams.js:164:5
_handlePlacesChanges@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/addon/ActivityStreams.js:290:9
emit@resource://devtools/shared/event-emitter.js:191:13
onItemAdded/<@resource://gre/modules/commonjs/toolkit/loader.js
-> resource://activity-streams/addon/PlacesProvider.js:226:11
Handler.prototype.process@resource://gre/modules/Promise.jsm
-> resource://gre/modules/Promise-backend.js:937:23
this.PromiseWalker.walkerLoop@resource://gre/modules/Promise.jsm
-> resource://gre/modules/Promise-backend.js:816:7
this.PromiseWalker.scheduleWalkerLoop/<@resource://gre/modules/Promise.jsm
[...]

PR #2173, opened 2017-02-14
fix(events): #2113 do not fire bookmark events if no data in event

k88hudson:
This fixes an issue where if bookmark-related event is fired with no data (from a test, for example) it will not propagate through our system.

It was causing `data.action is undefined` errors

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -260,9 +260,17 @@ ActivityStreams.prototype = {
   _handlePlacesChanges(eventName, data) {
     switch (eventName) {
       case "bookmarkAdded":
+        // Note: Firefox sometimes fires bookmarkAdded and bookmarkRemoved events
+        // that have no data associated with them
+        if (!data) {
+          return;
+        }
         this.broadcast(am.actions.Response("RECEIVE_BOOKMARK_ADDED", data));
         break;
       case "bookmarkRemoved":
+        if (!data) {
+          return;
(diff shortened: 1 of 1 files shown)


dmose:
I'm wondering if this is the right fix here, it feels like more of a band-aid, but that could be because I don't understand the problem well enough.

That is, if something in our regular code is sending such events, presumably we want to throw, or log an error, or something, correct?

k88hudson:
In this case, the empty events are actually being triggered on the Firefox side – i.e. Firefox is triggering places events unexpectedly (i.e. a bookmark event is triggered, but no data was attached). We could log an error, but that error is not really within the scope of our code

dmose:
OK, so it sounds like this is a band-aid fix over something that, indeed, is out of scope for us to fix.

Here's what I'd suggest:

* File and backlog an issue about either backing out this change or making it conditionally warn.
* File a bugzilla bug against Firefox for the creation of these events, and CC several of issue and mention the github issue
* Add a comment along the lines of 

```XXX Not currently logging these problems as documented in #2173.  Once Bugzilla bug 012345 is fixed, we should either revert this patch, or at least add a warning when these events occur.```

next to each of the two fixes.

r=dmose with these fixes.  I'm going to go turn back into a vacation pumpkin after writing this.  If you'd rather take a different tactic than proposed above and want to land this week, feel free to shift to another reviewer.

PR #2175, opened 2017-02-14
Update feeds on metadata change fixes #1969

### LINKED ISSUE #1969: Bookmark Highlights missing metadata
I am on the latest dev version: 1.2.0-dev-1484268155

This is my normal profile that I use daily. But I haven't browsed much in the last ~3-4 days.

screen shot 2017-01-17 at 10 00 57 am

jaredkerim:


rlr:
```
1115 of 1116 tests passed.

There were test failures...

The following tests failed:

  ./test/test-PreviewProvider.test_process_and_insert_links: exception
```

Was that one of the tests that was failing locally for you?

jaredkerim:
What no that should pass, the two failing ones are in TabTracker and PlacesProvider, one sec I'll check.

test/test-PreviewProvider.js
```diff
@@ -199,6 +199,9 @@ exports.test_process_links = function(assert) {
 exports.test_process_and_insert_links = function(assert) {
   const fakeData = {"url": "http://example.com/1", "title": "Title for example.com/1"};
 
+  const mockStoreActions = [];
+  gPreviewProvider._store = {dispatch: action => mockStoreActions.push(action)};
```
rlr: this is the problem. `[].push(action)` returns action and not the array.

rlr: or ehm. nvm. misread that

jaredkerim: The return doesn't matter, we just need to mutate the array and then check membership.

jaredkerim: This passes locally on my machine, I suspect it's because of some JS behaviour difference with FFox 46

test/test-PreviewProvider.js
```diff
@@ -438,6 +442,7 @@ exports.test_copy_over_correct_data_from_firefox = function*(assert) {
 exports.test_compute_image_sizes = function*(assert) {
   let mockExperimentProvider = {data: {metadataService: false}};
   gPreviewProvider = new PreviewProvider(gMockTabTracker, gMockMetadataStore, mockExperimentProvider, {initFresh: true});
+  gPreviewProvider._store = {dispatch: () => {}};
```
rlr: oooh. why not just reset it back at the end of `test_process_and_insert_links`? It should probably clean up for itself

rlr: oops. nvm. This is a different `gPreviewProvider` instance and you are just mocking out the `store`. :shipit:

PR #2177, opened 2017-02-15
feat(prefs): Closes #2170 Global pref to turn experiments off

### LINKED ISSUE #2170: Create a global pref which turns off experiments
When running mochitests, we're not sure if the failures are coming from experiments or just in general with AS, so for testing purposes we should have a pref which turns off all experiments. This also includes switching it off in prefs_general.diff

sarracini:
* Creates a pref which, if set to false, will not initialize the experiment provider.
* Turns that pref to false in ```prefs_general.diff``` so we can turn them off during mochitests
* Adds a test

### DIFF
--- a/addon/ExperimentProvider.js
+++ b/addon/ExperimentProvider.js
@@ -53,6 +53,20 @@ exports.ExperimentProvider = class ExperimentProvider {
     this._experimentId = null;
   }
 
+  /**
+   * This is used to disable all experiments, i.e set all their
+   * values to their original control value.
+   */
+  disableAllExperiments() {
+    Object.keys(this._experiments).forEach(key => {
+      const experiment = this._experiments[key];
+      const {active, control} = experiment;
+      if (active) {
+        prefService.set(PREF_PREFIX + key, control.value);
+      }
+    });
+  }
+
   setValues() {
     if (ss.storage.overrideExperimentProvider) {
       console.log(`The following experiments were turned on via overrides:\n`); // eslint-disable-line no-console
@@ -64,6 +78,12 @@ exports.ExperimentProvider = class ExperimentProvider {
           prefService.set(PREF_PREFIX + experimentName, control.value);
         }
       });
+      return;
+    }
+
+    // if the global pref to disable experiments is on, disable experiments
+    if (!prefService.get(`extensions.${preferencesBranch}.activateExperiments`)) {
+      this.disableAllExperiments();
       return;
     }
 
--- a/content-test/addon/ExperimentProvider.test.js
+++ b/content-test/addon/ExperimentProvider.test.js
@@ -41,6 +41,7 @@ describe("ExperimentProvider", () => {
   }
 
   beforeEach(() => {
+    prefService.set(`extensions.${preferencesBranch}.activateExperiments`, true);
     global.EventEmitter = {
       decorate(ctx) {
         ctx.emit = sinon.spy();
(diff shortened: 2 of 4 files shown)


mozilla-central/prefs_general.diff
```diff
@@ -5,6 +5,7 @@ diff --git a/testing/profiles/prefs_general.js b/testing/profiles/prefs_general.
 
 +// Make sure that activity stream doesn't hit the network.
 +user_pref("extensions.@activity-streams.telemetry", false);
++user_pref("extensions.@activity-streams.activateExperiments", false);
```
Mardak: Ah the dangers of manually editing patches.. `patch: **** malformed patch at line 12:  // Increase the APZ content response timeout in tests to 1 minute.`

You need to indicate there's 1 additional line added on line 4: `+268,7` to `+268,8`

addon/ActivityStreams.js
```diff
@@ -113,7 +113,7 @@ ActivityStreams.prototype = {
     this._initializePerfMeter();
     this._initializeAppURLHider();
 
-    if (!this.options.shield_variant) {
+    if (!this.options.shield_variant && simplePrefs.prefs.activateExperiments) {
```
Mardak: I suppose if this worked for shield, it might be okay here. In particular, I thought we discussed yesterday about an approach that forces each experiment's pref to be set to the `control` value. I believe we happen to be okay right now because all the control values are `false`, so any pref check happens to get a `false`-y value.

Probably to avoid accidental breakage of non-`false` control values, we should shift where we "disable" things.

test/test-ExperimentProvider.js
```diff
@@ -36,5 +38,32 @@ exports["test ActivityStreams doesn't init experimentProvider"] = assert => {
   assert.deepEqual(as._experimentProvider.data, {}, "should have empty .data");
   as.unload();
 };
+exports["test ActivityStreams doesn't init experimentProvider when pref is off"] = assert => {
```
Mardak: Probably describe the test as making sure the user is in the control group as opposed to just being "doesn't init."

mozilla-central/prefs_general.diff
sarracini: oooops

addon/ExperimentProvider.js
```diff
@@ -67,6 +81,12 @@ exports.ExperimentProvider = class ExperimentProvider {
       return;
     }
 
+    if (prefService.get(`extensions.${preferencesBranch}.activateExperiments`) === false) {
```
Mardak: I don't think we usually explicitly check against `true` or `false`, so a plain `!` check should be good.

addon/ExperimentProvider.js
```diff
@@ -67,6 +81,12 @@ exports.ExperimentProvider = class ExperimentProvider {
...
+      // if the global pref to disable experiments is on, disable experiments
```
Mardak: nit: comment before the `if` block instead of inside.

content-test/addon/ExperimentProvider.test.js
```diff
@@ -374,6 +374,29 @@ describe("ExperimentProvider", () => {
       assert.isTrue(experimentProvider.data.foo);
       assert.isTrue(ss.storage.overrideExperimentProvider);
     });
+    it("should disable all active experiments if pref says so", () => {
```
Mardak: Sorry I wasn't explicit earlier, but we can just file an issue instead of fixing it here. Shield studies have the same bug that this was fixing (i.e., not setting control values), so we'll eventually probably want to have some parameter to `init()` to disable/enable.

PR #2222, opened 2017-03-07
feat (experiment): #2065 screenshots in highlights experiment

### LINKED ISSUE #2065: Use screenshots for highlights with no metadata
There are times when we may show a highlight without metadata (for instance if it's a bookmark that we want to show because it's a bookmark but it does not have a suitable preview image) should we consider using a screenshot for that image rather than showing no image (ie a big block of colour)

rlr:
r? @jaredkerim 

Fixes #2065


rlr:
closing this to redo it using the new non-blocking screenshots @jaredkerim just landed.

PR #2237, opened 2017-03-09
Add importmc task

k88hudson:
This patch adds a task which simply copies changes from `mozilla-central` back to this repository.

It's needed because there were a couple of (small) review changes as per https://bugzilla.mozilla.org/show_bug.cgi?id=1344319

What do you think? Not exactly a sophisticated approach, but works for now maybe?


k88hudson:
@sarracini Yeah, I agree – we should be careful. Also can you change the review status to approved 😸?

PR #2240, opened 2017-03-10
Refresh Topsites if metadata or screenshots are missing fixes #2115

### LINKED ISSUE #2115: Refresh Topsites if metadata is missing
In the new screenshots experiment, top sites now depend on metadata, however they do not refresh when metadata is finished being processed.  We need to refresh top sites when metadata is processed the same way we do with highlights right now.

jaredkerim:
- Make getScreenshot fully async
- Send a message when a screenshot is computed
- Refresh topsites if we're missing rich icons or screenshots and they become available


addon/Feeds/TopSitesFeed.js
```diff
@@ -23,28 +26,27 @@ module.exports = class TopSitesFeed extends Feed {
...
-            if (site.favicon_height >= 96 && site.favicon_width >= 96 && isRootDomain(site.url)) {
-              // If we are at the "root domain path" and the icon is big enough,
-              // we don't show a screenshot.
-              return false;
-            }
-            return true;
-          });
-          links = links.map(link => {
-            if (link.screenshot) {
+        for (const link of links) {
```
rlr: you can't use `const` here since it is going to change every loop. Travis caught it:

```
Full Message: SyntaxError: missing = in const declaration

Full Stack: @resource://activity-streams/addon/Feeds/TopSitesFeed.js:33:NaN
```

k88hudson: Sorry for the random drop-in:

This is actually totally valid syntax for `for...of`, since `const link` is scoped to the inner block of the loop and not the outer block.

However, because of a bug in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1101653)  this throws a syntax error. It's fixed in 51+, but anything lower than that still hast the bug...

... so, maybe you might what to change it to `let` anyway. 🤓

rlr: ah. the more you know... 🤓

addon/Feeds/TopSitesFeed.js
```diff
@@ -23,28 +26,27 @@ module.exports = class TopSitesFeed extends Feed {
...
+          const isMissingIcon = !link.favicon_width && !link.favicon_height;
+          const hasSmallIcon = !isMissingIcon && ((link.favicon_width < 64) || (link.favicon_height < 64));
+          const badIcon = isMissingIcon || hasSmallIcon;
```
rlr: nice way to make this all readable 👍

addon/lib/getScreenshot.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  } else {
+    screenshots.set(url, SCREENSHOT_IN_FLIGHT);
+    PreviewProvider.getThumbnail(url).then(dataURI => {
+      screenshots.set(url, dataURI);
+      store.dispatch({type: "SCREENSHOT_UPDATED"});
+    }).catch(e => {
+      screenshots.set(url, null);
+    });
+  }
```
rlr: clever 👍  me likes

rlr:
6 tests failed on travis FYI

addon/Feeds/MetadataFeed.js
```diff
@@ -18,9 +18,16 @@ module.exports = class MetadataFeed extends Feed {
    * the state
    */
   getInitialMetadata(reason) {
-    return PlacesProvider.links.getRecentlyVisited().then(links => {
+    // Get initial topsites metadata
+    return PlacesProvider.links.getTopFrecentSites().then(links => {
```
rlr: oooh, was it missing this? that explains a lot

k88hudson: might be worth renaming that `getRecentlyVisited` query, it's really easy to confuse it. maybe something like `getPossibleHighlights` or something?

content-test/addon/Feeds/TopSitesFeed.test.js
```diff
@@ -28,6 +28,40 @@ describe("TopSitesFeed", () => {
...
+      testSite = {
+        hasMetadata: true,
+        url: "http://www.example.com/",
+        favicon_width: TopSitesFeed.MIN_ICON_SIZE,
+        favicon_height: TopSitesFeed.MIN_ICON_SIZE
+      };
+    });
+    it("should not get a screenshot for sites missing metadata", () => {
+      testSite.hasMetadata = false;
+      assert.equal(false, instance.shouldGetScreenshot(testSite));
```
rlr: nit: `testSite` already will return `false` as is (as seen below), so changing `hasMetadata` to false doesn't really prove that the condition is working. So I think changing more about the `testSite` (url or favicon dimensions) will make it a better test

as-pine-proxy:
[Treeherder: 1c9cf05afab07f505e360d404dfbaba81a1b4721](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1c9cf05afab07f505e360d404dfbaba81a1b4721)

Issue #2244, opened 2017-03-13
Missing session IDs in events table


emtwo:

emtwo:
Here are some daily counts of this:

https://sql.telemetry.mozilla.org/queries/3578#7038

ncloudioj:
This was fixed by 0f5a17

PR #2262, opened 2017-03-16
fix(metrics): #2244 - fix the missing session_id in the event pings

ncloudioj:
This issue was also found in performance/masga pings.

The fix is a workaround that checks the `session_id` field before each assignment statement (i.e. `session_id` as a left value).


as-pine-proxy:
[Treeherder: bc5fff53646952aad7049f86e0340f919a8b5719](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=bc5fff53646952aad7049f86e0340f919a8b5719)

Issue #2249, opened 2017-03-14
Port Redux/Redux store


k88hudson:
We need a working redux store for the graduated version of the add-on, in order to hook up all of our other pieces (like the remote page manager, etc.)

sarracini:
Fixed with #2263

PR #2263, opened 2017-03-16
feat(mc): #2249 Add Store, Actions, Reducers, and Redux

k88hudson:
This patch adds a store, actions, reducers, and Redux to the system addon. 

## Code
An instance of`lib/Store.jsm` when `lib/ActivityStream.jsm` starts up. It inititializes a redux store, as well as a list of `Feed`s.

I've included a really simple one for now called `NewTabInit`, that just logs the initial state. When the `MessageManager` lands, it can properly be hooked up to send the initial state to the content page.

## Tests
I had to add a few things to help with tests, including:
- `rewire`, a babel plugin that allows us to rewrite `import` statements;
- `GlobalOverrider`, a utility to set and restore global variables (e.g. `Components.utils`). unfortunately `rewire` doesn't support globals.

As usual, run `testmc` to run the tests.

## Manually testing

Try running the code in Firefox and checking your console output. You should see the redux state logged by the `NewTabInit` Action subscriber.

### DIFF
--- a/karma.mc.config.js
+++ b/karma.mc.config.js
@@ -26,6 +26,9 @@ module.exports = function(config) {
     files: [PATHS.testEntryFile],
     preprocessors,
     webpack: {
+      devtool: "eval",
+      // This loader allows us to override required files in tests
+      resolveLoader: {alias: {inject: path.join(__dirname, "loaders/inject-loader")}},
       // This resolve config allows us to import with paths relative to the system-addon/ directory, e.g. "lib/ActivityStream.jsm"
       resolve: {
         modules: [
@@ -41,7 +44,12 @@ module.exports = function(config) {
             exclude: [/node_modules/],
             use: [{
               loader: "babel-loader",
-              options: {plugins: [["jsm-to-esmodules", {basePath: PATHS.resourcePathRegEx, replace: true}]]}
+              options: {
+                plugins: [
+                  // Converts .jsm files into common-js modules
+                  ["jsm-to-commonjs", {basePath: PATHS.resourcePathRegEx, replace: true}]
+                ]
+              }
             }]
           }
         ]
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "babel": "6.5.2",
     "babel-core": "6.18.2",
     "babel-loader": "6.2.7",
-    "babel-plugin-jsm-to-esmodules": "0.2.0",
+    "babel-plugin-jsm-to-commonjs": "0.1.1",
     "babel-plugin-transform-es2015-destructuring": "6.18.0",
     "babel-plugin-transform-es2015-parameters": "6.18.0",
     "babel-plugin-transform-strict-mode": "6.18.0",
--- a/system-addon/lib/NewTabInit.jsm
+++ b/system-addon/lib/NewTabInit.jsm
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+const {utils: Cu} = Components;
+const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
+
+/**
+ * NewTabInit - A placeholder for now. This will send a copy of the state to all
+ *              newly opened tabs.
+ */
+class NewTabInit {
+  onAction(action) {
+    switch (action.type) {
+      // TODO: Replace with sending a copy of the state when a NEW_TAB_LOAD action is received
+      case at.INIT:
+        dump(`\n${JSON.stringify(this.store.getState(), 0, 2)}\n`);
+    }
+  }
+}
+
+this.NewTabInit = NewTabInit;
+this.EXPORTED_SYMBOLS = ["NewTabInit"];
--- a/system-addon/test/unit/lib/Store.test.js
+++ b/system-addon/test/unit/lib/Store.test.js
@@ -1,0 +1,112 @@
+const {Store} = require("lib/Store.jsm")
+const {actionTypes: at} = require("common/Actions.jsm");
+const {createStore} = require("redux");
+const {addNumberReducer} = require("test/unit/utils");
+
+describe("Store", () => {
+  let sandbox;
+  let store;
+  before(() => {
+    sandbox = sinon.sandbox.create();
+  });
+  beforeEach(() => {
+    store = new Store();
+  });
+  afterEach(() => sandbox.restore());
+  it("should have an empty feeds Set", () => {
+    assert.instanceOf(store.feeds, Set);
+    assert.equal(store.feeds.size, 0, ".feeds.size");
+  });
+  it("should have a redux store at ._store", () => {
+    assert.ok(store._store);
+    assert.property(store, "dispatch");
+    assert.property(store, "getState");
+  });
+  describe("#init", () => {
+    it("should dispatch an init action", () => {
+      sandbox.spy(store, "dispatch");
+
+      store.init();
+      
+      assert.calledWith(store.dispatch, {type: at.INIT});
+    });
+    it("should dispatch an init action", () => {
+      sandbox.spy(store, "dispatch");
+
+      store.init([]);
+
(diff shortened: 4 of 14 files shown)


package.json
```diff
@@ -29,7 +29,8 @@
     "babel": "6.5.2",
     "babel-core": "6.18.2",
     "babel-loader": "6.2.7",
-    "babel-plugin-jsm-to-esmodules": "0.2.0",
+    "babel-plugin-jsm-to-esmodules": "0.2.1",
```
k88hudson: I had to fix a small bug in `0.2.1`

karma.mc.config.js
```diff
@@ -26,6 +26,7 @@ module.exports = function(config) {
     files: [PATHS.testEntryFile],
     preprocessors,
     webpack: {
+      devtool: "inline-eval",
```
dmose: The webpack 2.0 docs don't mention inline-eval.  Is this the same as just "eval"?  If so, I'd suggest switching to the canonical name.  If not, a comment would be helpful.

karma.mc.config.js
```diff
@@ -41,7 +42,12 @@ module.exports = function(config) {
             exclude: [/node_modules/],
             use: [{
               loader: "babel-loader",
-              options: {plugins: [["jsm-to-esmodules", {basePath: PATHS.resourcePathRegEx, replace: true}]]}
+              options: {
+                plugins: [
+                  ["jsm-to-esmodules", {basePath: PATHS.resourcePathRegEx, replace: true}],
+                  "rewire"
```
dmose: It would be nice if each of the plugins had a short comment explaining what it was doing and why...

system-addon/lib/NewTabInit.jsm
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * NewTabInit - A placeholder for now. This will send a copy of the state to all
+ *              newly opened tabs.
+ */
+class NewTabInit {
+  onAction(action) {
+    switch (action.type) {
+      case at.INIT:
+        dump(`\n${JSON.stringify(this.store.getState(), 0, 2)}\n`);
```
k88hudson: This will go away as soon as the `MessageManager` lands. I will add a comment. I just added it for manually testing the patch

k88hudson:
Ok, I've converted everything to common-js and added some comments. I still need to add common-js support to the `jsm-to-esmodules` plugin

k88hudson:
Alright, everything should be working with injector and common-js!

system-addon/test/unit/utils.js
```diff
@@ -0,0 +1,52 @@
...
+ * GlobalOverrider - Utility that allows you to override properties on the global object.
+ *                   Seee unit-entry.js for example usage.
+ */
+class GlobalOverrider {
+  constructor() {
+    this.originalGlobals = new Map();
+    this.addedGlobals = new Set();
+    this.sandbox = sinon.sandbox.create();
+  }
+  _override(key, value) {
```
dmose: I think each of the methods in this class want a bit of JSDoc, so that it's possible to easily tell how they should be used.

system-addon/test/unit/utils.js
```diff
@@ -0,0 +1,52 @@
...
+    // Components will throw an error in FF if you override it normal.
+    // So, we have to force override it. Oh well.
+    if (key === "Components") {
+      try {global[key] = value} catch (e) {}
+      return;
+    }
+    if (!this.originalGlobals.has(key)) {
+      this.originalGlobals.set(key, global[key]);
+    }
```
dmose: So if it's been overridden once, that override itself can't be overridden?  In the comments for this method, please include the rationale for this.

system-addon/test/unit/lib/Store.test.js
```diff
@@ -0,0 +1,83 @@
...
+  beforeEach(() => {
+    store = new Store();
+  });
+  afterEach(() => sandbox.restore());
+  it("should have an empty feeds Set", () => {
+    assert.instanceOf(store.feeds, Set);
+    assert.equal(store.feeds.size, 0, ".feeds.size");
+  });
+  it("should have a redux store at ._store", () => {
+    assert.ok(store._store);
```
dmose: This is really just asserting that store is non-falsy.  Maybe assert.instanceOf is would be a better fit?  Otherwise, I'd suggest changing the description to more accurately describe what's being tested.

system-addon/test/unit/lib/Store.test.js
```diff
@@ -0,0 +1,83 @@
...
+      sandbox.spy(store, "dispatch");
+      store.init();
+      assert.calledWith(store.dispatch, {type: at.INIT});
+    });
+    it("should dispatch an init action", () => {
+      sandbox.spy(store, "dispatch");
+      store.init([]);
+      assert.calledWith(store.dispatch, {type: at.INIT});
+    });
+    it("should add each subscriber to .actionScribers and .store to each subscriber", () => {
```
dmose: The english here doesn't seem to exactly characterize what's going on.

system-addon/test/unit/lib/Store.test.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+  });
+  describe("#dispatch", () => {
+    it("should call .onAction of each feed", () => {
+      const sub = {onAction: sinon.spy()};
+      const action = {type: "FOO"};
+      store.init([sub]);
+      store.dispatch(action);
+      assert.calledWith(sub.onAction, action);
+    });
```
dmose: One convention that I've found makes non-trivial tests substantially easier to skim is to split them up with whitespace into three sections: initialize, execute, verify.  In this case, that would look like:

```js
      const sub = {onAction: sinon.spy()};
      const action = {type: "FOO"};
      store.init([sub]);

      store.dispatch(action);

      assert.calledWith(sub.onAction, action);
```

system-addon/test/unit/lib/Store.test.js
```diff
@@ -0,0 +1,83 @@
...
+    it("should call the reducers", () => {
+      store._store = createStore(addNumberReducer);
+      store.dispatch({type: "ADD", data: 14});
+      assert.equal(store.getState(), 14);
+    });
+  });
+  describe("#subscribe", () => {
+    it("subscribe to changes to the store", () => {
```
dmose: Needs the word "should", I think.

system-addon/test/unit/utils.js
k88hudson: It can be overridden more than once, but the `originalGlobals` is what the *original*, unaltered value was. That value should not be re-cached the second time you override it, because it's already been reset once.

package.json
```diff
@@ -29,7 +29,7 @@
     "babel": "6.5.2",
     "babel-core": "6.18.2",
     "babel-loader": "6.2.7",
-    "babel-plugin-jsm-to-esmodules": "0.2.0",
+    "babel-plugin-jsm-to-commonjs": "0.1.0",
```
k88hudson: Node 4 doesn't have `array.includes`, so I had to polyfill it.

k88hudson: Ok, I added a license to `0.1.1` and I will update to that 👍

system-addon/test/unit/utils.js
dmose: Ah, indeed, thanks!  As long the specific semantics are described in a JSDoc block, I think we're good, because at some level, either semantic could be considered reasonable (though I think you've made the better choice).

package.json
dmose: I'd suggest adding an XXX comment saying "let's get rid of this when we stop supporting Node 4", or some such.

k88hudson:
@dmose Thanks! I'll file a follow-up issue to deal with the requirements for adding Redux to the tree.

k88hudson:
@dmose follow-up issue is #2291

system-addon/test/unit/lib/Store.test.js
k88hudson: Unfortunately stores are not really instances of anything  – I added some checks on the shape of the object. Does that sound reasonable?

as-pine-proxy:
[Treeherder: 09dd29795f1878f488a59f5a7c6d12b1b91a68b4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=09dd29795f1878f488a59f5a7c6d12b1b91a68b4)

Issue #2287, opened 2017-03-20
Add coverage reporting to system-addon


k88hudson:
Let's check coverage reporting and fail the tests if the coverage drops below a certain level.

PR #2289, opened 2017-03-20
chore(mc): #2287 Add coverage reporting

k88hudson:
This is blocked on #2249 landing to pass tests. 

It adds local coverage reporting for system-addon code, which will fail the tests if the code drops below a certain threshold. I have it set to `100%` right now but we could drop that to something else.

Try running `testmc` to see the coverage reporting


yamscripts.yml
```diff
@@ -65,9 +65,9 @@ scripts:
   # importmc: Import changes from mc to github repo
   importmc: cpx "../mozilla-central/browser/extensions/activity-stream/**/*" system-addon/
 
-  testmc: karma start karma.mc.config.js
-
-  tddmc: =>testmc -- --no-single-run
+  testmc: karma start karma.mc.config.js || (cat logs/system-addon/text.txt && exit 2)
+  posttestmc: cat logs/system-addon/text-summary.txt
```
Mardak: I suppose this `posttestmc` doesn't actually need to happen as anything under 100% will cause the full text report to show up, but it can be reassuring that the output does show 100% for each.

karma.mc.config.js
```diff
@@ -18,11 +18,29 @@ const preprocessors = {};
...
+  const isTDD = config.tdd;
   config.set({
-    singleRun: true,
+    singleRun: !isTDD,
     browsers: ["Firefox"],
     frameworks: ["mocha", "sinon", "chai"],
-    reporters: ["mocha"],
+    reporters: ["mocha", "coverage"],
+    coverageReporter: {
+      dir: "logs/system-addon",
```
Mardak: Any reason why the non-`system-addon` puts files under a subdirectory of `coverage`? Do we want to match that for potentially other logs or simplify the other?

k88hudson: Sure, that's probably a good idea. No harm in putting it under `coverage/` just in case

as-pine-proxy:
[Treeherder: ca70726efd142fc7b8d7658b528b3d8da0310f0e](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=ca70726efd142fc7b8d7658b528b3d8da0310f0e)

PR #2297, opened 2017-03-22
fix(lint): Enable eslint for system-addon and its export to mozilla-central

### LINKED ISSUE #2229: Enable eslint for system-addon/
Right now we don't have eslint targeting `.jsm`s. There are a few issues we would need to sort out (like `no-unused-vars` conflicting with jsm-style exports, for example) to get this to work

Mardak:
Fix partially #2229 for this repository `eslint`ing and while exporting to `mozilla-central` where the top-level `.eslintrc.js` is only used for this repository and falling back to whatever is used for `browser/` (currently, generally less strict than here). I switched to the `eslintrc` `.js` format instead of `json` to match.

Is the approach of  `this.ActivityStream = class ActivityStream { ` acceptable? I suppose the class could be anonymous at that point? Alternatively is the `/* exports */` pattern.

Linting was partially failing on pine because the dot files weren't being copied over. Four rules were enabled to match mozilla-central and were autofixed except for `no-shadow`.

r? @k88hudson


yamscripts.yml
```diff
@@ -87,7 +87,7 @@ scripts:
 
 # lint: Run eslint and sass-lint
   lint:
-    eslint: esw --ext=.js,.jsx,.json . .storybook
+    eslint: esw --ext=.js,.jsm,.json,.jsx . .storybook
```
k88hudson: I don't actually think we actually use any `.jsx` files

.eslintignore
```diff
@@ -9,5 +8,3 @@ firefox/
 logs/
 stats.json
 test/test-ColorAnalyzer.js
-test/test-SimpleStorage.js
-system-addon/
```
k88hudson: Should we maybe add `system-addon/vendor/` here?

yamscripts.yml
Mardak: Removed `.jsx`

.eslintignore
Mardak: Added

as-pine-proxy:
[Treeherder: e331e030d82a60746740141a45ff893a03890295](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=e331e030d82a60746740141a45ff893a03890295)

dmose:
From the pine treeherder run, it looks like it's trying to lint the bundle itself.  I wonder how hard it would be to make eslint errs for the bootstrapped add-on show up in travis too, so that we'd catch them before merging...

PR #2303, opened 2017-03-22
Test merges fixups

dmose:
Running a bunch of pushes to pine in preparation for setting this up in cron, and needing to debug various things, it became clear that various things weren't aborting on errors like they should have. There are several small bugs in play here; this PR fixes them, and renames some variables for clarity.

### DIFF
--- a/bin/download-firefox-artifact
+++ b/bin/download-firefox-artifact
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env bash -x
 
 # Forked from https://github.com/devtools-html/debugger.html/blob/master/bin/download-firefox-artifact
 #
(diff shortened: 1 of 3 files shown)


bin/download-firefox-artifact
```diff
@@ -1,13 +1,28 @@
...
 # This looks for a mozilla-central artifact build as a sibling of the
 # activity-stream tree.  If it's not there, it creates it.  If it is there, it
 # updates it.
 
-ROOT=`dirname $0`
-FIREFOX_PATH="$ROOT/../../mozilla-central"
+# If AS_GIT_BIN_REPO (the git repo from which prepare-mochitests-dev and
+# friends will be executed) isn't set in the environment, just use the repo
+#  we're running from.
+if [ -z ${AS_GIT_BIN_REPO+x} ]; then
```
Mardak: Eh, I suppose if we tried to refactor to share this with `prepare-mochitest-dev`, that shared script would also need its own set of detecting which version to use. I guess this is fine for now.

Mardak: Also just making sure, these environment variables are only indirectly set when `prepare-mochitest-dev` is called where `test-merges` is passing it along?

bin/test-merges.js
```diff
@@ -155,8 +155,8 @@ function exportToLocalMC(commitId) {
     // use echo.
     shelljs.exec(`
       echo yes | \
-      env AS_PMD_GIT_DIR=${AS_PMD_GIT_DIR} SYMLINK_TESTS=false ENABLE_MC_AS=1 \
-      bash ${PREPARE_MOCHITESTS_DEV}`,
+      env AS_GIT_BIN_REPO=${AS_GIT_BIN_REPO} SYMLINK_TESTS=false ENABLE_MC_AS=1 \
+      ${PREPARE_MOCHITESTS_DEV}`,
```
Mardak: nit: could you indent these to better indicate they're part of one big command even though the `\` and `exec` imply that?

bin/download-firefox-artifact
dmose: Usually, AS_GIT_BIN_REPO is initialized to TESTING_LOCAL_GIT in test-merges.js.  That said, if you want to debug from your own git repo instead, you'll be setting AS_GIT_BIN_REPO in the environment that test-merges.js is started under.   Whatever the value is gets used to find prepare-mochitest-dev, and is passed down via the environment to prepare-mochitest-dev

bin/test-merges.js
dmose: Sure.

as-pine-proxy:
[Treeherder: 027dbbc70081da98012e36237440df2207116ec4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=027dbbc70081da98012e36237440df2207116ec4)

PR #2310, opened 2017-03-23
fix(topsites): Fixes #2228 Links in Top Sites are sometimes incorrect

### LINKED ISSUE #2228: Links in the Top Sites portion of Activity Stream sometimes are not correctly calculated
Running Firefox Developer Edition (Mozilla/5.0 (Windows NT 10.0; WOW64; rv:53.0) Gecko/20100101 Firefox/53.0)

On the Activity Stream new tab page, the Top Sites section lists some sites I visit regularly. That includes Hacker News (https://news.ycombinator.com). This might be a bit of a difficult bug to reproduce, but the link on the Activity Stream page is listed as **https://www.news.ycombinator.com**. The www portion seems to be added erroneously.

Would this happen to be because HN is my home page?

sarracini:
This is an experiment which changes the top sites query in a couple ways:
1. uses ```WITH``` statements to improve readability
2. sums up the frecencies for hosts that are duplicates (www and non www), then adds it as a column to each entry, then picks the highest frecency entry for the deduped entries, and shows this instead of arbitrarily picking the one that the ```SUM``` function that sqlite returns

The use case that this fixes is as follows:

### STR:
Visit ```news.ycombinator.com``` (good link - we want this one)
Visit ```www.news.ycombinator.com``` (bad broken link - don't show this one)
Visit ```news.ycombinator.com``` again to bump up it's frecency against the broken one
Open a new tab page and observe the hacker news top site and what it points to

### Without this patch:
The top site will point to ```www.news.ycombinator.com```, forever

### With this patch:
Since ```news.ycombinator.com``` has a now higher frecency, the top site will point to ```news.ycombinator.com```

### Caveat:
If the user *keeps typing ```www.news.ycombinator.com``` into the URL bar* and continues to visit the broken URL, increasing it's frecency as it happens, then the top site will show the broken URL because from a database level we can't know if the site is a "good one" or a "bad one". All we can really do is try and return the highest frenceny while deduping on www and non www sites. 

This is a very edge case situation - we should switch this experiment on and see that top sites in general isn't affected by changing the query this way (except for these edge case scenarios). Specifically @nchapman if you're still seeing ```www.news.ycombinator.com``` appear in your top sites then there's another issue going on here.

@nanj has also suggested maybe using a lack of metadata (no title, and no favicon and a low frecency) as a good signal that the site is a bad top site, and we should just filter those out. There are advantages and disadvantages of this approach and we should have some more discussions around how to best fix this "broken top site" issue.

cc @Mardak @k88hudson @jaredkerim @emtwo

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -338,6 +338,123 @@ Links.prototype = {
     return null;
   }),
 
+  largestFrecencyQuery(blockedURLs) {
+    return `WITH all_history_items as (
+                      SELECT
+                        rev_host,
+                        CASE SUBSTR(rev_host, -5)
+                        WHEN ".www." THEN SUBSTR(rev_host, -4, -999)
+                        ELSE rev_host
+                        END AS rev_nowww,
+                        moz_places.url,
+                        moz_favicons.data AS favicon,
+                        mime_type AS mimeType,
+                        moz_places.title,
+                        frecency,
+                        last_visit_date,
+                        moz_places.guid AS guid,
+                        moz_bookmarks.guid AS bookmarkGuid
+                      FROM moz_places
+                      LEFT JOIN moz_favicons
+                      ON favicon_id = moz_favicons.id
+                      LEFT JOIN moz_bookmarks
+                      ON moz_places.id = moz_bookmarks.fk
+                      WHERE hidden = 0
+                      AND last_visit_date NOTNULL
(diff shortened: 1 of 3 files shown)


sarracini:
I'm going to close this PR, clearly the issue hasn't been prominent enough that there's a need for this. We can re-open this discussion in the future if we need to

PR #2326, opened 2017-03-23
fix(telemetry): #2301 - compute the newtab stats in content

### LINKED ISSUE #2301: The current highlights_size and topsites_size counters are not accurate
The current way of counting tiles in the highlights and topsites is *not* accurate due to the extra tile processing on the content side (selectAndDedupe).

In order to get the correct stats, we need to do the counting in the content. However, there is already a plan to remove the `selectAndDedupe`, see details #2064 . We need to figure out a better solution to get the correct stats.

ncloudioj:
This fixes #2301

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -298,6 +298,10 @@ ActivityStreams.prototype = {
     this._tabTracker.handleUndesiredEvent(msg.data);
   },
 
+  _handleNewTabStats({msg}) {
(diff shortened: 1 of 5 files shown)


as-pine-proxy:
[Treeherder: 5117704c9361e4b9a3eee82d1ce4a9d93222e6be](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5117704c9361e4b9a3eee82d1ce4a9d93222e6be)

PR #2338, opened 2017-03-27
feat(testing): Include original git commit message in hg test-merges version

Mardak:
Was going through treeherder/perfherder and it was quite obtuse tracking down which revisions were what, so adding the original git commit message to hg too.

r?@dmose

### DIFF
--- a/bin/test-merges.js
+++ b/bin/test-merges.js
@@ -123,7 +123,7 @@ function getPRMergeCommitId(prNumber) {
  * Checks out the given commit into ${TESTING_LOCAL_GIT}
  *
  * @param  {String} commitId
- * @return {Promise<String|?>}    resolves with commitId on checkout completion,
+ * @return {Promise<String[]|?>} Resolves with commit [id, message] on checkout, or
  *                      rejects with error
  */
 function checkoutGitCommit(commitId) {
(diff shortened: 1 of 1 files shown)


bin/test-merges.js
```diff
@@ -141,7 +141,15 @@ function checkoutGitCommit(commitId) {
...
         }
-        resolve(commitId);
+
+        // Pass along the original commit message
+        git.show(["-s", "--format=%B"], (err3, data3) => {
+          if (err3) {
+            reject(err3);
+            return;
+          }
+          resolve([commitId, data3.trim()]);
```
dmose: Please update the JSDoc at the beginning of this function, since we're now resolving with something different.
I'd also suggest adding a return here, just to match prevailing style in this file.

Mardak: Fixed JSDoc. The return isn't allowed with m-c recommended eslint rules: https://hg.mozilla.org/mozilla-central/file/tip/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#l236

as-pine-proxy:
[Treeherder: f7f7ff518802ccc22f984cad4d2bbde47666d88c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f7f7ff518802ccc22f984cad4d2bbde47666d88c)

PR #2341, opened 2017-03-28
fix(tests) disable tightly coupled search tests on pine; closes #2339

### LINKED ISSUE #2339: temporarily disable newtab-coupled search tests on pine
This is part of #2276.  There are some tests in browser/components/search/test that are tightly couple to the newtab implementation.   When run with the bootstrapped add-on turned on (rather than the default tiles), they fail like this:

uncaught exception - TypeError: win.gSearch is undefined at load@chrome://mochitests/content/browser/browser/components/search/test/browser_amazon_behavior.js:98:1

There are various options for dealing with them, but after discussion with @k88hudson, we decided we probably don't want to block on them greening up the tree until our own search code is landed.  So for now, I'll disable them with a patch, and file another issue to reconsider what to do when the search code lands.

dmose:
This just adds a patch disabling some tests that are a bit too tightly coupled to help green up pine.  To test:

```
./bin/prepare-mochitest-dev
cd ../mozilla-central
./mach mochitest -f browser browser/components/search/test
```

Should no longer have 15 errors.

### DIFF
--- a/mozilla-central-patches/disable-search-tests.diff
+++ b/mozilla-central-patches/disable-search-tests.diff
@@ -1,0 +1,49 @@
+diff --git a/browser/components/search/test/browser.ini b/browser/components/search/test/browser.ini
+--- a/browser/components/search/test/browser.ini
++++ b/browser/components/search/test/browser.ini
+@@ -14,40 +14,44 @@ support-files =
+   tooManyEnginesOffered.html
+   webapi.html
+ 
+ [browser_426329.js]
+ [browser_483086.js]
+ [browser_addEngine.js]
+ [browser_amazon.js]
+ [browser_amazon_behavior.js]
++skip-if = true # mozilla/activity-stream#2339
+ [browser_bing.js]
+ [browser_bing_behavior.js]
++skip-if = true # mozilla/activity-stream#2339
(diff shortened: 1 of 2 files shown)


mozilla-central-patches/disable-search-tests.diff
```diff
@@ -0,0 +1,49 @@
...
+@@ -14,40 +14,44 @@ support-files =
+   tooManyEnginesOffered.html
+   webapi.html
+ 
+ [browser_426329.js]
+ [browser_483086.js]
+ [browser_addEngine.js]
+ [browser_amazon.js]
+ [browser_amazon_behavior.js]
++skip-if = true # mozilla/activity-stream#2339
```
Mardak: 8 lines of context seems likely to conflict with any `browser/components/search/test/browser.ini` changes. Just one line of context should be good enough to skip the specific next test.

as-pine-proxy:
[Treeherder: 26caa5a3f3687ef76e64baeac92166b9cbfe91c4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=26caa5a3f3687ef76e64baeac92166b9cbfe91c4)

Issue #2362, opened 2017-04-03
disable tightly-coupled migration undo tests


dmose:
pine is seeing failures of the form:

TEST-UNEXPECTED-FAIL | browser/components/migration/tests/browser/browser_undo_notification.js | Test timed out -
TEST-UNEXPECTED-FAIL | browser/components/migration/tests/browser/browser_undo_notification.js | Found a tab after previous test timed out: about:newtab -
TEST-UNEXPECTED-FAIL | browser/components/migration/tests/browser/browser_undo_notification_wording.js | Test timed out -
TEST-UNEXPECTED-FAIL | browser/components/migration/tests/browser/browser_undo_notification_wording.js | Found a tab after previous test timed out: about:newtab -

because those tests do things with about:newtab and about:home in them.
These need to be temporarily disabled, and I'll also file a bug about re-enabling later.

dmose:
The actual failure here is that the current versions of about:newtab and about:home know about some browser migration testing, and these tests are verifying that they actually handle it correctly.  To turn it back on, we'll presumably need to implement these things as well.

https://dxr.mozilla.org/mozilla-central/search?q=MaybeShowAutoMigration&redirect=false

dmose:
Closed by #2362

PR #2364, opened 2017-04-03
fix(tests): #2362 disable m-c undo notification tests

dmose:
This disables a couple of tests which expect about:home and about:newtab to handle various migration undo related messages.  These will need to be implemented before we can pref on A-S in mozilla-central.


as-pine-proxy:
[Treeherder: 29799ff6c8a5578fb2939f66815668b18cb15e3b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=29799ff6c8a5578fb2939f66815668b18cb15e3b)

PR #2367, opened 2017-04-04
Create Screenshots Long Metadata Cache Experiment #2263

jaredkerim:
Create a new topsites screenshots experiment that includes a long metadata timeout so that high res icons don't disappear after 3 days.

### DIFF
--- a/addon/PreviewProvider.js
+++ b/addon/PreviewProvider.js
@@ -31,7 +31,6 @@ Cu.import("resource://gre/modules/Task.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 
 const DEFAULT_OPTIONS = {
-  metadataTTL: 30 * 24 * 60 * 60 * 1000, // 30 days for the metadata to live
   proxyMaxLinks: 25, // number of links proxy accepts per request
   initFresh: false
 };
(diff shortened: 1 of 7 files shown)


jaredkerim:
This will be uplifted into 1.8.0.

addon/PreviewProvider.js
```diff
@@ -370,7 +378,7 @@ PreviewProvider.prototype = {
    */
   insertMetadata(links, metadataSource) {
     const linksToInsert = links.map(link => Object.assign({}, link, {
-      expired_at: (this.options.metadataTTL) + Date.now(),
+      expired_at: (this._getMetadataTTL()) + Date.now(),
```
ncloudioj: nit: calling this function for each insert seems to be unnecessary. I think it's fine as it's currently just running as an experiment, we may want to call it once and store the TTL at somewhere once the experiment is done.

ncloudioj:
LGTM.

@emtwo Just wanted to confirm with you that this experiment will disable the "screenshotAsync" one in the next release(1.8.0), do we have enough data to draw the conclusion for that one?

ncloudioj:
@jaredkerim Looks like Travis has captured some test failures.

addon/PreviewProvider.js
jaredkerim: @ncloudioj I agree but for the purposes of the experiment this is most consistent than statefully mirroring the experiment state and setting a local constant.  When our experimentation is complete we will move it back to being a hard coded constant.

jaredkerim:
Yep you're right I have to update a test fixture...

emtwo:
@jaredkerim code looks good to me aside from the tests. @ncloudioj yes, the conclusion for screenshotAsync is that click rates still aren't good enough and search is down too.

Ideally I would have wanted to keep it for another couple of weeks for confirmation but we are tight on amount of room for experiments (we have too many experiments!) so I think it's ok to turn this off and create a new variant of it with the cache change in the hopes that it will improve click rates.

as-pine-proxy:
[Treeherder: 676eada004682d0428f08545e731f6a414d6bcc2](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=676eada004682d0428f08545e731f6a414d6bcc2)

PR #2372, opened 2017-04-04
fix(addon): #2264 - fix query for getTopFrecentSites

### LINKED ISSUE #2264: I really like Jenn, but not that much. (I swear)
image

So, this is something I clicked on today, FOR THE FIRST TIME, I totally swear. Why on earth would this be considered a topsite?

ncloudioj:
This fixes #2264


addon/PlacesProvider.js
```diff
@@ -391,7 +405,8 @@ Links.prototype = {
                       WHERE hidden = 0 AND last_visit_date NOTNULL
                       AND moz_places.url NOT IN (${blockedURLs})
                       ORDER BY frecency, last_visit_date, moz_places.url DESC
-                    ) GROUP BY rev_host)
+                    ) GROUP BY rev_host
+                      ORDER BY frecency, last_visit_date, url DESC)
```
Mardak: As noted on irc, we'll probably end up implementing this type of custom combining/sorting logic in js so that the query is just a frecency-indexed `SELECT` to be fast. Having the unit tests will help make sure we maintain the desired logic while we rewrite.

test/lib/PlacesTestUtils.js
```diff
@@ -32,9 +32,11 @@ const PlacesTestUtils = Object.freeze({
    *            [optional] visitDate: visit date in microseconds from the epoch
    *            [optional] referrer: nsIURI of the referrer for this visit
    *          }
-   *
+   * @param {int} updateCount
```
Mardak: nit: I think the prevailing style is either `{Integer}` or `{Number}`

test/test-PlacesProvider.js
```diff
@@ -104,6 +104,14 @@ exports.test_Links_getTopFrecentSites_dedupeWWW = function*(assert) {
   links = yield provider.getTopFrecentSites();
   assert.equal(links.length, 1, "adding both www. and no-www. yields one link");
   assert.equal(links[0].frecency, 200, "frecency scores are combined ignoring extra pages");
+
+  // add another nowww page visit
```
Mardak: nit: no-www to avoid initial parsing of a typo

addon/PlacesProvider.js
ncloudioj: Agreed. I have copied this comment to the function so that other folks know the game plan :)

as-pine-proxy:
[Treeherder: 943472295ee61812ffea25699d548d53ca7bcb22](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=943472295ee61812ffea25699d548d53ca7bcb22)

Issue #2369, opened 2017-04-04
Remove license files from .jar


k88hudson:
We don't need to be building license files inside `system-addon/vendor`; this is also making our build fail on pine

PR #2375, opened 2017-04-04
fix(mc): #2369 Only include .js and .jsm files from vendor dir

k88hudson:
I'm not sure if there's a better way to do this, but this should fix it


as-pine-proxy:
[Treeherder: 859544c286eb85cc99ae95f6b4e58daac3e088b0](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=859544c286eb85cc99ae95f6b4e58daac3e088b0)

PR #2380, opened 2017-04-05
feat(places): Closes #2376 Ditch the left join on moz_favicons

### LINKED ISSUE #2376: Ditch the left join on moz_favicons
This has to go or else it's gonna be broken very soon - we will replace this with a call to the favicons API ```getFaviconDataForPage```

sarracini:
Gets the favicons from the favicons service API. fix #2376

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -1,4 +1,4 @@
-/* globals XPCOMUtils, Services, gPrincipal, EventEmitter, PlacesUtils, Task, Bookmarks, SyncedTabs */
+/* globals XPCOMUtils, Services, gPrincipal, EventEmitter, PlacesUtils, Task, Bookmarks, SyncedTabs, NetUtil */
 
 "use strict";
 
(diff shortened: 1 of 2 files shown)


addon/PlacesProvider.js
```diff
@@ -420,21 +402,42 @@ Links.prototype = {
...
+    links = yield this._getFavicons(links);
     return this._processLinks(links);
   }),
 
+  _getFavicons: Task.async(function*(links) {
+    if (links.length) {
+      let faviconPromises = [];
+      links.forEach(item => {
+        faviconPromises.push(new Promise(resolve => {
+          PlacesUtils.favicons.getFaviconDataForPage(NetUtil.newURI(item.url), (iconuri, len, data, mime) => {
```
Mardak: Thinking about the failure cases here.. potentially `getFaviconDataForPage` could throw or `newURI` doesn't like the url passed in (although this seems unlikely that Places would add a bad url..?). It would lead to a rejection in one of these `Promise`s. The value of `Promise.all` would be the rejection but we're not looking at the result of that anyway, so the main issue is that the first `reject` causes `all` to reject even with other pending requests resulting in `links` being returned when `favicon`s aren't fully added.

Mardak: We probably don't need to add a test for the bad uri case, but at least making sure we're actually getting expected favicon data now using `getFaviconDataForPage`

as-pine-proxy:
[Treeherder: b66522c95dc82eb812b70cd2612e9409ac6da799](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b66522c95dc82eb812b70cd2612e9409ac6da799)

Issue #2276, opened 2017-03-20
green up treeherder for bootstrapped add-on on pine


dmose:
This is actually probably several different issues.  https://treeherder.mozilla.org/#/jobs?repo=pine is an example of the bootstrapped add-on test results on pine.  A few of these are actually "normal" intermittent mozilla-central failures.  Other classes include:

* a bunch of identical-looking failures related to the localization set-up on pine
* a set of failures due to the fact that a bunch of tests in the tree are expecting old-school new tab semantics, even though the push to pine is only building and testing the bootstrapped add-on.  We may need to do some things to make both sets of preferences work in the tests, since presumably we're going to be feature-flagged in the tree for a while.
* eslint failure (there's already a github issue open for this

Next steps here are to go through all the failing test (classes) and break them down into separate tickets.

Mardak:
Looks like just #2371 left.

PR #2384, opened 2017-04-07
fix(mc): Selectively not copy content-src (jsx, scss) and webpack config from system-addon to mozilla-central

Mardak:
Address some of #2276 and #2229. r?@k88hudson 

Don't copy over content-src as the jsx files are put in the bundle and scss into the single css. Also move webpack config to the top, so it's not copied. These are to avoid linting issues in m-c. We do want to copy over the bundle, so update the top level `eslintnignore` to ignore it.

Alternative copy that picks out the desired files types (excludes jsx and scss):
`cpx "system-addon/**/{,.}*{.,_}{build,css,html,in,ini,LICENSE,md,mn,js,jsm,json}" ../mozilla-central/browser/extensions/activity-stream`

That alternate is what I started with although it seems less explicit than a list of directories/files. I suppose with the current list in the PR, I could leave the webpack config within the `system-addon` directory as it's not listed to be copied over.


mozilla-central-patches/eslintignore.diff
```diff
@@ -0,0 +1,9 @@
+diff --git a/.eslintignore b/.eslintignore
+--- a/.eslintignore
++++ b/.eslintignore
+@@ -72,3 +72,4 @@ browser/extensions/pocket/content/panels/js/vendor/**
+ browser/locales/**
+-# vendor library files in activity-stream
++# generated or library files in activity-stream
++browser/extensions/activity-stream/data/content/activity-stream.bundle.js
```
Mardak: Are we expecting other bundles? Or js files in that data/content directory? Maybe just ignore that whole directory? The explicit `bundles` subdirectory would make it clearer why it's being ignored though.

webpack.system-addon.config.js
```diff
@@ -18,9 +19,9 @@ module.exports = {
   },
   // This resolve config allows us to import with paths relative to the system-addon/ directory, e.g. "lib/ActivityStream.jsm"
   resolve: {
+    extensions: [".jsx"],
```
Mardak: This would be to include the `lib` files, yeah? I'll add that in.

webpack.system-addon.config.js
```diff
@@ -18,9 +19,9 @@ module.exports = {
...
     modules: [
-      __dirname,
-      "node_modules"
```
Mardak: Okay, restoring `node_modules`

as-pine-proxy:
[Treeherder: 838c5174dfab6ca4fe1a0e70f43ba7aa929746c6](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=838c5174dfab6ca4fe1a0e70f43ba7aa929746c6)

Mardak:
greeeen! https://treeherder.mozilla.org/#/jobs?repo=pine&revision=838c5174dfab6ca4fe1a0e70f43ba7aa929746c6&selectedJob=89723792&filter-searchStr=eslint

k88hudson:
\o/ 🍏 📗 💚 🍏 📗 💚 🍏 📗 💚 🍏 📗 💚

PR #2387, opened 2017-04-07
feat (topsites): #2274 Pinning

### LINKED ISSUE #2274: Edit Top Sites: implement pinning
From #1987:  UI for the panel allows for pinning of Top Site cards

This should be a link/button on the top right of the tile that appears on hover.

screen shot 2017-03-20 at 11 52 10 am

^ Pinned tiles should have an indicator icon on the lower left corner.

We should reuse the pref used by the existing about:newtab or migrate users to the new system.

rlr:
Fixes #2274

### DIFF
--- a/addon/Feeds/TopSitesFeed.js
+++ b/addon/Feeds/TopSitesFeed.js
@@ -1,4 +1,4 @@
-/* globals Task */
+/* globals Task, NewTabUtils */
 const {Cu} = require("chrome");
 const {PlacesProvider} = require("addon/PlacesProvider");
 const Feed = require("addon/lib/Feed");
@@ -10,12 +10,14 @@ const getScreenshot = require("addon/lib/getScreenshot");
 const {isRootDomain} = require("addon/lib/utils");
 
 Cu.import("resource://gre/modules/Task.jsm");
+Cu.import("resource://gre/modules/NewTabUtils.jsm");
 
 module.exports = class TopSitesFeed extends Feed {
-  constructor(...args) {
-    super(...args);
+  constructor(options) {
+    super(options);
     this.getScreenshot = getScreenshot;
     this.missingData = false;
+    this.pinnedLinks = options.pinnedLinks || NewTabUtils.pinnedLinks;
   }
   shouldGetScreenshot(link) {
     const isMissingIcon = !link.favicon_width && !link.favicon_height;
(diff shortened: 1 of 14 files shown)


addon/Feeds/TopSitesFeed.js
```diff
@@ -10,26 +10,60 @@ const getScreenshot = require("addon/lib/getScreenshot");
...
   }
+  pinLink(action) {
+    this.pinnedLinks.pin(action.data.site, action.data.index);
+  }
+  unpinLink(action) {
+    this.pinnedLinks.unpin(action.data.site);
+  }
+  sortLinks(links, pinned) {
+    // Seperate out the pinned from the rest
+    let pinnedLinks = links.filter(link => this.pinnedLinks.isPinned(link));
```
ncloudioj: nit: looks like the `links` was formed as "[pinnedLinks, nonPinnedFrecentLinks]". Can we just slice it into two sub-arrays instead of filtering twice here?

common/selectors/selectAndDedupe.js
```diff
@@ -14,7 +16,21 @@ dedupe.defaults.createKey = site => site.cache_key || site.hostname || site.url;
...
-    let sites = options.defaults ? options.sites.concat(options.defaults) : options.sites;
+    let sites = options.sites;
+
+    if (options.defaults) {
+      // Dedupe the defaults first.
+      let defaults = dedupe.group([sites, options.defaults])[1];
+      // Fill in any empty slots with default sites.
+      sites.forEach((site, i) => {
+        if (!site) {
+          sites[i] = defaults.shift();
```
ncloudioj: Any chance that `defaults.shift()` could return an `undefined`, i.e. running out of the `defaults`? Imaging a user has pinned two links at slot 1 and 12 in the "showMore" experiment, then he clears the whole history. In that case the default links (as currently 6 tiles) won't be enough to populate those 10 slots.

addon/Feeds/TopSitesFeed.js
rlr: good point. 👍

common/selectors/selectAndDedupe.js
rlr: yes, this is possible. Maybe we show the empty placeholder tiles for that case? Otherwise it will show the pin for slot 12 at slot 8 or something like that.

ncloudioj: Sounds good. I think the placeholders will solve this issue.

addon/Feeds/TopSitesFeed.js
```diff
@@ -10,26 +10,60 @@ const getScreenshot = require("addon/lib/getScreenshot");
...
+    return sortedLinks;
+  }
   getData() {
     return Task.spawn(function*() {
       const experiments = this.store.getState().Experiments.values;
 
-      let links;
-      // Get links from places
-      links = yield PlacesProvider.links.getTopFrecentSites();
+      // Get pinned links
```
ncloudioj: What you think for this change? `getTopFrecentSites` already take a `limit` as one of the arguments. We can just add a new argument "excludedLinks". That also somewhat simplifies the code below.

ncloudioj: Or do all those tweaks in a separate PR if needed. Your call :)

rlr: Yeah, I think I'll hold off on adding the `excludedLinks` argument to the query since those queries are being commited to m-c. We should probably do the change over there once it lands.

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -33,9 +33,9 @@ const NewTabPage = React.createClass({
         const topSites = this.props.TopSites.rows;
         stats.topsitesSize = topSites.length;
         topSites.forEach(row => {
-          if (row.screenshot) {
+          if (row && row.screenshot) {
```
ncloudioj: nice guard 👍

as-pine-proxy:
[Treeherder: 16a133b816727452a06a5471acd543556787dbdf](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=16a133b816727452a06a5471acd543556787dbdf)

Issue #2313, opened 2017-03-23
port localization / jar.mn to system addon


tspurway:

sarracini:
Strings used in search for system-addon so far:
search input placeholder: "Search the Web"
search button and search input title (for a11y): "Submit search"

Mardak:
I chatted with Standard8, and it looks like the `screenshots` system-addon has been able to just include all the locales in the xpi that gets built in Nightly. In terms of uncompressed size, all those `messages.json` total about 200KB. Currently our locales.json is about 170KB.

https://hg.mozilla.org/mozilla-central/file/tip/browser/extensions/screenshots/moz.build

Part of the reason is that relying on the build locale of Firefox might not always be appropriate, so having the other locales available to select dynamically allows for a better experience.

Given that these strings will be used in the content space, it might make sense to use `react-intl` as currently used. We'll need to not rely on `sdk/l10n` which turns out to be buggy, and we'll want to use the new fancy `Services.locale.getRequestedLocale`

PR #2304, opened 2017-03-22
chore(lint): Switch to eslint-plugin-mozilla/recommended@0.2.40

Mardak:
Fix #2229. Update the dependency and clean up linting issues and duplicate config.

@Standard8 One oddness I'm running into right now is that I needed to `ln -s ../mozilla-central/tools/` as `convertWorkerExpressionToGlobals@lib/helpers.js` is looking for `modules.json`. Other than that, I was able to `npm install tools/lint/eslint/eslint-plugin-mozilla` from `activity-stream` to develop against `0.2.32`

PR #2427, opened 2017-04-13
feat(system-addon): Closes #2312 Create search feed + base search UI

### LINKED ISSUE #2312: port search feed to system addon

sarracini:
Closes #2312 Adds some base functionality for the Search component:
- Search react component (just the search bar + button)
- CSS for the Search component 
- Functionality to perform a search
- Functionality to get the default engine so the search properly works
- Brand new assets for the new search spec: https://mozilla.invisionapp.com/share/GMBA4GJRA#/screens
- Tests for the reducer, and the feed

Notes: the actual search component isn't tested - though it's simple

To try it out: ```npm run buildmc``` in activity stream -> ```./mach build faster && ./mach run``` in mozilla-central. and voila! we are able to do some searches! (without suggestions for now)

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -33,6 +33,13 @@ const INITIAL_STATE = {
         "url": "https://twitter.com/"
       }
     ]
+  },
+  Search: {
+    currentEngine: {
+      name: "",
+      icon: ""
+    },
+    engines: []
   }
 };
 
@@ -60,7 +67,23 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
   }
 }
 
+function Search(prevState = INITIAL_STATE.Search, action) {
+  switch (action.type) {
+    case at.SEARCH_STATE_UPDATED: {
+      if (!action.data) {
+        return prevState;
+      }
+      let {currentEngine, engines} = action.data;
+      return Object.assign({}, prevState, {
+        currentEngine,
+        engines
+      });
+    }
+    default:
+      return prevState;
+  }
+}
 this.INITIAL_STATE = INITIAL_STATE;
-this.reducers = {TopSites};
+this.reducers = {TopSites, Search};
 
 this.EXPORTED_SYMBOLS = ["reducers", "INITIAL_STATE"];
--- a/system-addon/content-src/components/Search/Search.jsx
+++ b/system-addon/content-src/components/Search/Search.jsx
@@ -1,0 +1,39 @@
+"use strict";
+const React = require("react");
+const {connect} = require("react-redux");
+const {actionTypes, actionCreators} = require("common/Actions.jsm");
+
+const Search = React.createClass({
+  getInitialState() {
+    return {searchString: ""};
+  },
+  performSearch(options) {
+    let searchData = {
+      engineName: options.engineName,
+      searchString: options.searchString,
+      searchPurpose: "newtab",
+      healthReportKey: "newtab"
+    };
+    this.props.dispatch(actionCreators.SendToMain({type: actionTypes.PERFORM_SEARCH, data: searchData}));
+  },
+  onClick(event) {
+    const {currentEngine} = this.props.Search;
+    event.preventDefault();
+    this.performSearch({engineName: currentEngine.name, searchString: this.state.searchString});
+  },
+  onChange(event) {
+    this.setState({searchString: event.target.value});
+  },
+  render() {
+    return (<form className="search-wrapper">
+      <span className="search-label" />
+      <input value={this.state.searchString} type="search"
+        onChange={this.onChange}
+        maxLength="256" title="Submit search"
+        placeholder="Search the Web" />
+        <button onClick={this.onClick} />
+        </form>);
+  }
+});
+
+module.exports = connect(state => ({Search: state.Search}))(Search);
--- a/system-addon/content-src/components/Search/Search.scss
+++ b/system-addon/content-src/components/Search/Search.scss
@@ -1,0 +1,242 @@
+.search-wrapper {
+  $search-bg-color: #FFF;
+  $search-border-radius: 4px;
+  $search-margin-top: 52px;
+  $search-margin-bottom: 48px;
+  $search-height: 36px;
+  $search-input-left-label-width: 35px;
+  $search-input-border-color: #BFBFBF;
+  $search-blue: #0996F8;
+  $search-hover-shadow: rgba(0.11, 0.34, 0.69, 0.5);
+  $search-focus-shadow: #0996F8;
+  $search-button-grey: #FBFBFB;
+  $search-button-border-color: #D4D4D4;
+  $search-button-width: 36px;
+  $search-glyph-image: url('#{$image-path}glyph-search-16.svg');
+  $search-glyph-size: 20px;
+  $search-text: #666;
+  $search-section-background: #F7F7F7;
+  $search-border-color: #EAEAEA;
+
+  @mixin list-items {
+    padding: 0;
+    margin: 0;
+    list-style: none;
+  }
+
+  @mixin list-link {
+    cursor: default;
+    color: $black;
+    display: block;
+
+    &:hover,
+    &.active {
+      background: $search-blue;
+      color: $white;
+    }
+  }
+
+  @mixin search-input {
+    border: 0;
+    box-shadow: 0 1px 0 0 $faintest-black;
+    flex-grow: 1;
+    margin: 0;
+    outline: none;
+    padding: 0 12px 0 $search-input-left-label-width;
+    height: 100%;
+
+    &:focus {
+      border-color: $search-blue;
+      box-shadow: 0 0 0 2px $search-focus-shadow;
+      transition: box-shadow 150ms;
+      z-index: 1;
+    }
+  }
+
+  cursor: default;
+  display: flex;
+  position: relative;
+  margin: 0 0 $search-margin-bottom;
+  width: 100%;
+  height: $search-height;
+
+  .search-container {
+    z-index: 1001;
+    background: $search-bg-color;
+    position: absolute;
+    left: 0;
+    right: 0;
+    top: 100%;
+    margin-top: -2px;
+    border: 1px solid $search-input-border-color;
+    font-size: 12px;
+    box-shadow: 0 5px 10px $faintest-black;
+    overflow: hidden;
+
+    .search-title {
+      color: $search-text;
+      padding: 5px 10px;
+      background-color: $search-section-background;
+      display: flex;
+      align-items: center;
+      word-break: break-all;
+
+      p {
+        margin: 0;
+      }
+
+      #current-engine-icon {
+        margin-inline-end: 8px;
+      }
+    }
+
+    section {
+      border-bottom: 1px solid $search-border-color;
+      margin-bottom: 0;
+    }
+
+    .search-suggestions {
+      ul {
+        @include list-items;
+
+        li a {
+          @include list-link;
+          padding: 4px 36px;
+        }
+      }
+    }
+
+    .history-search-suggestions {
+      border-bottom: 0;
+
+      ul {
+        @include list-items;
+
+        li a {
+          @include list-link;
+          padding: 4px 10px;
+
+          &:hover > #historyIcon,
+          &.active > #historyIcon {
+            background-image: url('#{$image-path}glyph-search-history.svg#search-history-active');
+          }
+        }
+      }
+    }
+
+    .history-search-suggestions #historyIcon {
+      width: 16px;
+      height: 16px;
+      display: inline-block;
+      margin-inline-end: 10px;
+      margin-bottom: -3px;
+      background-image: url('#{$image-path}glyph-search-history.svg#search-history');
+    }
+
+    .search-partners {
+      ul {
+        @include list-items;
+
+        li {
+          display: inline-block;
+          padding: 5px 0;
+
+          a {
+            display: block;
+            padding: 3px 16px;
+            border-right: 1px solid $search-input-border-color;
+          }
+
+          &:hover,
+          &.active {
+            background: $search-blue;
+            color: $white;
+
+            a {
+              border-color: transparent;
+            }
+          }
+        }
+      }
+    }
+
+    .search-settings {
+      button {
+        color: $search-text;
+        margin: 0;
+        padding: 0;
+        height: 32px;
+        text-align: center;
+        width: 100%;
+        border-style: solid none none;
+        border-radius: 0;
+        background: $search-section-background;
+        border-top: 0;
+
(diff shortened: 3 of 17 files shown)


system-addon/content-src/components/Search/Search.jsx
```diff
@@ -0,0 +1,34 @@
...
+  },
+  render() {
+    const {currentEngine, engines} = this.props.Search;
+    return (<form className="search-wrapper">
+      <span className="search-label" />
+      <input value={this.state.searchString} type="search"
+        onChange={e => this.setState({searchString: e.target.value})}
+        maxLength="256"
+        placeholder="Search the Web"/>
+        <button onClick={e => {e.preventDefault(); this.performSearch({engineName: currentEngine.name, searchString: this.state.searchString});}} />
```
Mardak: I thought we turned on `eslint` `jsx-no-bind` but I guess not. The `onChange` and `onClick` here are creating new functions each time `render` is called, so this will prevent any useful caching of any previously rendered `Search`. These should be able to be plain methods on the class to avoid new functions from `render`.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -0,0 +1,34 @@
...
+    this.props.dispatch(actionCreators.SendToMain({type: actionTypes.PERFORM_SEARCH, data: searchData}));
+  },
+  render() {
+    const {currentEngine, engines} = this.props.Search;
+    return (<form className="search-wrapper">
+      <span className="search-label" />
+      <input value={this.state.searchString} type="search"
+        onChange={e => this.setState({searchString: e.target.value})}
+        maxLength="256"
+        placeholder="Search the Web"/>
```
Mardak: I see the plain string in `TopSites`, so do we have a good place to track these strings to replace with #2313?

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -0,0 +1,34 @@
...
+  getInitialState() {
+    return {
+      searchString: ''
+    };
+  },
+  performSearch(options) {
+    let searchData = {
+      engineName: options.engineName,
+      searchString: options.searchString,
+      searchPurpose: "d",
```
Mardak: This was copied over, but should we be putting something more useful for `purpose`? It looks like existing Firefox usage might be "newtab" and "homepage"

system-addon/content-src/components/Search/Search.scss
```diff
@@ -0,0 +1,56 @@
...
+  button {
+    margin-inline-start: -50px;
+    border: none;
+    width: $search-button-width;
+    transition: box-shadow 150ms;
+    background: transparent url('#{$image-path}glyph-forward.svg') no-repeat center center;
+    background-size: 16px 16px;
+    border-radius: 0 $search-border-radius $search-border-radius 0;
+
+    &:hover {
```
Mardak: Is there a better way to declare these same styles as the `input:focus + button {` above? Also, because this seems to be exactly the same, there's no visual distinction from when the input is focused with button unhovered to the button getting hovered.

system-addon/content-src/components/Search/Search.scss
```diff
@@ -0,0 +1,56 @@
...
+      border-color: $search-blue;
+      transition: box-shadow 150ms;
+      box-shadow: $search-shadow;
+      background-color: $search-blue;
+      background-image: url('#{$image-path}glyph-forward-white.svg');
+      color: $white;
+    }
+  }
+
+  button {
```
Mardak: I noticed the button doesn't have a pointer cursor that the existing about:home/newtab have. A minor parity issues to check with UX?

Mardak: Also parity is the existing pages have a title "Submit search" that we might want to keep for a11y?

system-addon/lib/SearchFeed.jsm
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+  async getState() {
+    const state = await ContentSearch.currentStateObj(true);
+    const action = {type: at.SEARCH_STATE_UPDATED, data: {engines: state.engines, currentEngine: state.currentEngine}};
+    this.store.dispatch(ac.BroadcastToContent(action));
+  }
+  performSearch(action) {
+    let gBrowser = this.getCurrentBrowser();
+    ContentSearch.performSearch({target: gBrowser}, action.data);
```
Mardak: This seems like a fragile way to do the search especially with testing/automation where the most recent window might not be the actual window we want. Is there a good way for `PERFORM_SEARCH` to pass along the browser that triggered the search?

On a side note, `gBrowser` is a confusing name as it's not actually a global and not the usual expected "gBrowser" (which is normally the top level Firefox whole window browser as opposed to here it's a single tabbrowser).

system-addon/lib/SearchFeed.jsm
```diff
@@ -0,0 +1,63 @@
...
+  "nsIWindowMediator");
+
+this.SearchFeed = class SearchFeed {
+  getCurrentBrowser() {
+    const win = windowMediator.getMostRecentWindow("navigator:browser");
+    const browser = win.getBrowser().selectedBrowser;
+    return browser;
+  }
+  addObservers() {
+    Services.obs.addObserver(this, SEARCH_ENGINE_TOPIC, false);
```
Mardak: We'll want a matching `removeObserver`

system-addon/lib/SearchFeed.jsm
```diff
@@ -0,0 +1,63 @@
...
+      case "engine-changed":
+      case "engine-added":
+      case "engine-removed":
+        this.getState();
+        break;
+    }
+  }
+  async getState() {
+    const state = await ContentSearch.currentStateObj(true);
+    const action = {type: at.SEARCH_STATE_UPDATED, data: {engines: state.engines, currentEngine: state.currentEngine}};
```
Mardak: As noted in the reducer, we should do the transformation here to not bake the `iconBuffer` into the "action API"

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -0,0 +1,34 @@
+"use strict";
+const React = require("react");
+const {connect} = require("react-redux");
+const {actionTypes, actionCreators} = require("common/Actions.jsm");
+
+const Search = React.createClass({
+  getInitialState() {
+    return {
+      searchString: ''
```
Mardak: nit: Use double quotes as used everywhere else (thought there was an eslint rule for this??)

system-addon/common/Reducers.jsm
```diff
@@ -60,7 +67,27 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
...
+        return prevState;
+      }
+      return Object.assign({}, prevState, {
+        currentEngine: {
+          name: action.data.currentEngine.name,
+          icon: action.data.currentEngine.iconBuffer
+        },
+        engines: action.data.engines.map(engine => ({
+          name: engine.name,
+          icon: engine.iconBuffer
```
sarracini: Just so i understand: construct the engines and current engines object in the search feed to look how we want it to look and pass it down to the reducer so all the reduce has to do is:
```js
  return Object.assign({}, prevState, {
        currentEngine: action.data.currentEngine,
        engines: action.data.engines
      });
``` 
Right?

system-addon/content-src/components/Search/Search.jsx
sarracini: Not that I know of, but as a temporary solution I can post the strings that I used in this patch in a comment in #2313 ?

system-addon/content-src/components/Search/Search.scss
sarracini: this seems to work:
```css
  input {
    @include search-input;
    padding-inline-start: 35px;
  }

  button {
    margin-inline-start: -50px;
    border: none;
    width: $search-button-width;
    transition: box-shadow 150ms;
    background: transparent url('#{$image-path}glyph-forward.svg') no-repeat center center;
    background-size: 16px 16px;
    border-radius: 0 $search-border-radius $search-border-radius 0;
    cursor: pointer;
  }
  
  input:focus + button, button:hover {
    z-index: 1;
    border-color: $search-blue;
    transition: box-shadow 150ms;
    box-shadow: $search-shadow;
    background-color: $search-blue;
    background-image: url('#{$image-path}glyph-forward-white.svg');
    color: $white;
  }
```

what do you think?

system-addon/content-src/components/Search/Search.jsx
Mardak: Oh duh, we didn't include `jsx` in `esw`

Mardak: I'll fix the esw issue in #2304

system-addon/lib/SearchFeed.jsm
```diff
@@ -0,0 +1,63 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ /* globals ContentSearch, XPCOMUtils, windowMediator, Services */
+"use strict";
+
+const {utils: Cu} = Components;
+const {actionTypes: at, actionCreators: ac} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
+const SEARCH_ENGINE_TOPIC = "browser-search-engine-modified";
+
+Cu.import("resource:///modules/ContentSearch.jsm");
```
Mardak: I see @k88hudson switching things over to lazy import even for things called from `init`, so I guess we should do that here too?

system-addon/content-src/components/Search/Search.scss
Mardak: Looks good. Add a newline separating `input:focus + button` and `button:hover` to make it clear it's selecting two different things as opposed to a long selector for one thing.

system-addon/common/Reducers.jsm
Mardak: Sounds good. Doesn't look like this would save much:
```js
let {currentEngine, engines} = action.data;
return Object.assign({}, prevState, {
  currentEngine,
  engines
});
```

sarracini: ^^ I would do this but I get a "Unexpected lexical declaration in case block" linting error - can't put the ```let``` inside the case statement :(

Mardak: The trick there is to create a block inside the case, e.g.,
```js
case foo: {
  let blah;
  break;
}
```

Mardak: Or alternatively, flip the `if` condition so the if block creates the block for the `let`

system-addon/lib/SearchFeed.jsm
sarracini: Should I do it for ```Services``` too or just ```ContentSearch``` ?

Mardak: After looking through some other m-c files, looks like some aggressively lazily load everything that isn't used immediately for the first execution on loading the module. This would also include the `Actions.jsm` import. @k88hudson any suggestions?

E.g.,
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/History.jsm
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/PlacesUtils.jsm

But not always consistent:
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/BookmarkHTMLUtils.jsm
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/PlacesBackups.jsm

system-addon/lib/SearchFeed.jsm
sarracini: Yup, can do. I just have to change the message manager  to send up the entire message instead of just msg.data: https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/MessageManager.jsm#L145
That way I can do msg.target.browser, and I can access the browser it came from

sarracini:
Blocked on https://reviewboard.mozilla.org/r/130898/diff/1#index_header landing in central
Bugzilla ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1357174

system-addon/lib/SearchFeed.jsm
```diff
@@ -0,0 +1,73 @@
...
+    Services.obs.addObserver(this, CONTENT_SEARCH_READY, false);
+  }
+  removeObservers() {
+    Services.obs.removeObserver(this, SEARCH_ENGINE_TOPIC);
+    Services.obs.removeObserver(this, CONTENT_SEARCH_READY);
+  }
+  observe(subject, topic, data) {
+    switch (topic) {
+      case CONTENT_SEARCH_READY:
+        this.getState();
```
Mardak: As noted in bug 1357174, would just watching for "browser-ui-startup-complete" work as `ContentSearch.init` would happen before that notification? Although does immediately calling `getState` result in the correct values as it's unclear what from `ContentSearch.init` you're actually depending on.

sarracini: Basically when you get the state (specifically the current engine) it tries to get a placeholder for the current engine name by trying to access ```this._stringBundle```, which gets initialized in ```ContentSearch.init```. It throws an error saying ```this._stringBundle``` is undefined at that point

sarracini: So here we can listen to ```browser-ui-startup-complete``` and it will work in the case that we get initialized before ContentSearch gets initialized from nsBrowserGlue, but in the case where ContentSearch.init happens first and we haven't had time to set up our observer for that notification yet, we still need to check, on our init, if ContentSearch is ready to be used. So we still need to expose a ```isInit``` value in ContentSearch.

Mardak: @k88hudson The issue we're running into is the add-on when installed and enabled as a system add-on runs its init related code before the rest of the browser -- in this specific instance depending on `ContentSearch.init()`, which runs from `finalUIStartup` of `nsBrowserGlue`, which also includes other new tab module inits:

https://dxr.mozilla.org/mozilla-central/source/browser/components/nsBrowserGlue.js#507-512

One potential solution is to have the add-on only trigger `bootstrap.js`'s `ActivityStream.init` after it has observed `browser-ui-startup-complete`, which is notified at the end of `finalUIStartup`. However by delaying initialization, there's a risk of being too late?

I suppose the usual way if activity stream was just another module in Firefox, it would get its init triggered directly from `finalUIStartup`.

k88hudson: That seems fine to me, as long as that event is reliable – especially since that's where the other NewTab stuff is initialized. Might also be worth adding a comment why we're using that event

as-pine-proxy:
[Treeherder: b8396ead7433003cb9cb7be09aba35d327a91834](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b8396ead7433003cb9cb7be09aba35d327a91834)

PR #2433, opened 2017-04-17
feature(telemetry): fixes #2284, port TelemetrySender to system addon

### LINKED ISSUE #2284: Port TelemetrySender to system addon
This is for session and event pings

dmose:
Port of TelemetrySender file, with fresh unit tests.  It's not actually hooked up to anything yet, so the only thing to do to test is it to run the unit tests.

r=@sarracini?

### DIFF
--- a/system-addon/lib/TelemetrySender.jsm
+++ b/system-addon/lib/TelemetrySender.jsm
@@ -1,3 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 /* globals Preferences, Services, XPCOMUtils */
 
 const {interfaces: Ci, utils: Cu} = Components;
--- a/system-addon/lib/TelemetrySender.jsm
+++ b/system-addon/lib/TelemetrySender.jsm
@@ -9,6 +9,7 @@ Cu.import("resource://gre/modules/Preferences.jsm");
 Cu.importGlobalProperties(["fetch"]);
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/Console.jsm"); // eslint-disable-line no-console
 
 const ENDPOINT_PREF = "telemetry.ping.endpoint";
 const TELEMETRY_PREF = "telemetry";
@@ -110,7 +111,7 @@ TelemetrySender.prototype = {
     if (this.logging) {
       // performance related pings cause a lot of logging, so we mute them
       if (JSON.parse(data).action !== "activity_stream_performance") {
-        Cu.reportError(`TELEMETRY PING: ${data}`);
+        console.log(`TELEMETRY PING: ${data}`); // eslint-disable-line no-console
       }
     }
 
--- a/system-addon/lib/TelemetrySender.jsm
+++ b/system-addon/lib/TelemetrySender.jsm
@@ -24,6 +24,16 @@ const UNDESIRED_NOTIF = "undesired-event";
 // installed.  Though maybe we should just forcibly disable the old add-on?
 const PREF_BRANCH = "browser.newtabpage.activity-stream";
 
+/**
+ * Observe various notifications and send them to a telemetry endpoint.
+ *
+ * @param {Object} args - optional arguments
+ * @param {Function} args.prefInitHook - if present, will be called back
+ *                   inside the Prefs constructor. Typically used from tests
+ *                   to save off a pointer to a fake Prefs instance so that
+ *                   stubs and spies can be inspected by the test code.
+ *
+ */
 function TelemetrySender(args) {
   let prefArgs = {branch: PREF_BRANCH};
   if (args) {
--- a/system-addon/lib/TelemetrySender.jsm
+++ b/system-addon/lib/TelemetrySender.jsm
@@ -1,0 +1,126 @@
+/* globals Preferences, Services, XPCOMUtils */
+
+const {interfaces: Ci, utils: Cu} = Components;
+
+Cu.import("resource://gre/modules/Preferences.jsm");
+Cu.importGlobalProperties(["fetch"]);
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+
+const ENDPOINT_PREF = "telemetry.ping.endpoint";
+const TELEMETRY_PREF = "telemetry";
+const LOGGING_PREF = "performance.log";
+
+const ACTION_NOTIF = "user-action-event";
+const PERFORMANCE_NOTIF = "performance-event";
+const COMPLETE_NOTIF = "tab-session-complete";
+const UNDESIRED_NOTIF = "undesired-event";
+
+// This is intentionally a different pref-branch than the SDK-based add-on
+// used, to avoid extra weirdness for people who happen to have the SDK-based
+// installed.  Though maybe we should just forcibly disable the old add-on?
+const PREF_BRANCH = "browser.newtabpage.activity-stream";
+
+function TelemetrySender(args) {
+  let prefArgs = {branch: PREF_BRANCH};
+  if (args) {
+    if ("prefInitHook" in args) {
+      prefArgs.initHook = args.prefInitHook;
+    }
+  }
+
+  this._prefs = new Preferences(prefArgs);
+
+  this.enabled = this._prefs.get(TELEMETRY_PREF);
+  this._onTelemetryPrefChange = this._onTelemetryPrefChange.bind(this);
+  this._prefs.observe(TELEMETRY_PREF, this._onTelemetryPrefChange);
+
+  this.logging = this._prefs.get(LOGGING_PREF);
+  this._onLoggingPrefChange = this._onLoggingPrefChange.bind(this);
+  this._prefs.observe(LOGGING_PREF, this._onLoggingPrefChange);
+
+  this._pingEndpoint = this._prefs.get(ENDPOINT_PREF);
+  this._onEndpointPrefChange = this._onEndpointPrefChange.bind(this);
+  this._prefs.observe(ENDPOINT_PREF, this._onEndpointPrefChange);
+
+  if (this.enabled) {
+    Services.obs.addObserver(this, COMPLETE_NOTIF, true);
+    Services.obs.addObserver(this, ACTION_NOTIF, true);
+    Services.obs.addObserver(this, PERFORMANCE_NOTIF, true);
+    Services.obs.addObserver(this, UNDESIRED_NOTIF, true);
+  }
+}
+
+TelemetrySender.prototype = {
+  /* We need this to be multi-process safe while using the observer service
+   * (though it's not obvious to me how the weak reference helps that). */
+  QueryInterface: XPCOMUtils.generateQI([
+    Ci.nsIObserver,
+    Ci.nsISupportsWeakReference
+  ]),
+
+  observe(subject, topic, data) {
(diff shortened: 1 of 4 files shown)


system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+TelemetrySender.prototype = {
+  /* We need this to be multi-process safe while using the observer service
+   * (though it's not obvious to me how the weak reference helps that). */
+  QueryInterface: XPCOMUtils.generateQI([
+    Ci.nsIObserver,
+    Ci.nsISupportsWeakReference
+  ]),
+
+  observe(subject, topic, data) {
+    if (topic === COMPLETE_NOTIF || topic === ACTION_NOTIF || topic === PERFORMANCE_NOTIF || topic === UNDESIRED_NOTIF) {
```
sarracini: maybe this would be nicer as a switch to allow expanding the list of notifications in the future
```js
switch (topic) {
  case COMPLETE_NOTIF:
  case ACTION_NOTIF:
  case PERFORMANCE_NOTIF:
  case UNDESIRED_NOTIF:
    this._sendPing(data);
    break;
}
```

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+    this._pingEndpoint = prefVal;
+  },
+
+  _onLoggingPrefChange(prefVal) {
+    this.logging = prefVal;
+  },
+
+  _onTelemetryPrefChange(prefVal) {
+    if (this.enabled && !prefVal) {
+      Services.obs.removeObserver(this, COMPLETE_NOTIF);
```
sarracini: maybe factor out the ```Services.obs.removeObserver``` and the ```Services.obs.addObserver``` into ```removeObservers``` and ```addObservers``` functions, respectively. That way ```_onTelemetryPrefChange```, ```uninit``` and the constructor can just call into those methods instead of duplicating the code

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+      if (!response.ok) {
+        Cu.reportError(`Ping failure with HTTP response code: ${response.status}`);
+      }
+    }).catch(e => {
+      Cu.reportError(`Ping failure with error: ${e}`);
+    });
+  },
+
+  uninit() {
+    try {
```
sarracini: just out of curiosity - i know this code was copied, but why is this in a try catch?

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+      this._prefs.ignore(TELEMETRY_PREF, this._onTelemetryPrefChange);
+      this._prefs.ignore(ENDPOINT_PREF, this._onEndpointPrefChange);
+    } catch (e) {
+      Cu.reportError(e);
+    }
+  }
+};
+
+this.TelemetrySender = TelemetrySender;
+this.EXPORTED_SYMBOLS = ["TelemetrySender"];
```
sarracini: This is just a nit but I think the pattern we've adopted for the ```jsm```'s in ```system-addon/lib/``` is: 

```js
this.ClassName = class ClassName {
    constructor() {}
    someFunction() {}
    ...
};
this.EXPORTED_SYMBOLS = ["ClassName"];
```

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+
+// This is intentionally a different pref-branch than the SDK-based add-on
+// used, to avoid extra weirdness for people who happen to have the SDK-based
+// installed.  Though maybe we should just forcibly disable the old add-on?
+const PREF_BRANCH = "browser.newtabpage.activity-stream";
+
+function TelemetrySender(args) {
+  let prefArgs = {branch: PREF_BRANCH};
+  if (args) {
+    if ("prefInitHook" in args) {
```
sarracini: Since this isn't hooked up to anything it's hard to tell what this is used for. Can you explain this?

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+
+const ENDPOINT_PREF = "telemetry.ping.endpoint";
+const TELEMETRY_PREF = "telemetry";
+const LOGGING_PREF = "performance.log";
+
+const ACTION_NOTIF = "user-action-event";
+const PERFORMANCE_NOTIF = "performance-event";
+const COMPLETE_NOTIF = "tab-session-complete";
```
sarracini: We should sync up with @k88hudson to see how she's planning to handle session start and session complete pings - it's possible this notification will look different, or we may not need it if we use the RemotePageManager unload events, or we may have to add a ```tab-sessions-start``` event since we're no longer using tab open as an indicator

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+    Services.obs.addObserver(this, COMPLETE_NOTIF, true);
+    Services.obs.addObserver(this, ACTION_NOTIF, true);
+    Services.obs.addObserver(this, PERFORMANCE_NOTIF, true);
+    Services.obs.addObserver(this, UNDESIRED_NOTIF, true);
+  }
+}
+
+TelemetrySender.prototype = {
+  /* We need this to be multi-process safe while using the observer service
+   * (though it's not obvious to me how the weak reference helps that). */
```
sarracini: Well this is interesting - I don't quite get how the weak reference helps either. Can you expand on this?

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
+/* globals Preferences, Services, XPCOMUtils */
```
sarracini: I believe this needs a Mozilla License!

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+  this._onTelemetryPrefChange = this._onTelemetryPrefChange.bind(this);
+  this._prefs.observe(TELEMETRY_PREF, this._onTelemetryPrefChange);
+
+  this.logging = this._prefs.get(LOGGING_PREF);
+  this._onLoggingPrefChange = this._onLoggingPrefChange.bind(this);
+  this._prefs.observe(LOGGING_PREF, this._onLoggingPrefChange);
+
+  this._pingEndpoint = this._prefs.get(ENDPOINT_PREF);
+  this._onEndpointPrefChange = this._onEndpointPrefChange.bind(this);
+  this._prefs.observe(ENDPOINT_PREF, this._onEndpointPrefChange);
```
sarracini: Why do we even want to watch for endpoint changes? Isn't that not a great idea, if we leave the endpoint open to a pref then in theory someone could change it and re-route where their data is going... if we set it once it should be enough to keep it that way. Besides, you've got the fake prefs for testing so there really isn't a need to watch for this pref change
I vote to remove this and just set it once in the constructor 🙋

system-addon/test/unit/lib/TelemetrySender.test.js
```diff
@@ -0,0 +1,295 @@
+const {GlobalOverrider, FakePrefs} = require("test/unit/utils");
```
sarracini: A Mozilla License again

system-addon/test/unit/lib/TelemetrySender.test.js
```diff
@@ -0,0 +1,295 @@
...
+  fakePrefs = this; // eslint-disable-line consistent-this
+};
+const tsArgs = {prefInitHook};
+
+describe("TelemetrySender", () => {
+  let globals;
+  let tSender;
+  let fetchStub;
+  const observerTopics = ["user-action-event", "performance-event",
+    "tab-session-complete", "undesired-event"];
```
sarracini: maybe export these in ```TelemetrySender.jsm``` and import them here so we can stay consistent if these ever change

system-addon/test/unit/lib/TelemetrySender.test.js
```diff
@@ -0,0 +1,295 @@
...
+  const fakeEndpointUrl = "http://127.0.0.1/stuff";
+  const fakePingJSON = JSON.stringify({action: "fake_action", monkey: 1});
```
sarracini: 🐵 🐵 🐵

system-addon/test/unit/lib/TelemetrySender.test.js
```diff
@@ -0,0 +1,295 @@
...
+    globals = new GlobalOverrider();
+
+    fetchStub = globals.sandbox.stub();
+
+    globals.set("Preferences", FakePrefs);
+    globals.set("fetch", fetchStub);
+  });
+
+  beforeEach(() => {
+  });
```
sarracini: can you put ```tSender = new TelemetrySender(tsArgs);``` in this beforeEach so you don't have to call it in each test? Or is it a timing issue with setting the fake prefs?

system-addon/test/unit/lib/TelemetrySender.test.js
```diff
@@ -0,0 +1,295 @@
...
+    beforeEach(() => {
+      FakePrefs.prototype.prefs = {
+        "telemetry": true,
+        "telemetry.ping.endpoint": fakeEndpointUrl
+      };
+      tSender = new TelemetrySender(tsArgs);
+    });
+
+    it("should POST given ping data to telemetry.ping.endpoint pref w/fetch",
+    async () => {
```
sarracini: neat!! this is great!

system-addon/lib/TelemetrySender.jsm
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    this.enabled = prefVal;
+  },
+
+  async _sendPing(data) {
+    if (this.logging) {
+      // performance related pings cause a lot of logging, so we mute them
+      if (JSON.parse(data).action !== "activity_stream_performance") {
+        Cu.reportError(`TELEMETRY PING: ${data}`);
```
sarracini: Do we want to report an error here, or just log it to console?

system-addon/lib/TelemetrySender.jsm
dmose: Given the generateQI issues we saw while pairing, I'm going to leave this as-is to keep risk down.

system-addon/test/unit/lib/TelemetrySender.test.js
dmose: This is actually intentional: while it's good practice to keep the code itself DRY, it's useful to have a separate source of truth for testing.  Without this, a typo in the source code won't get flagged by the tests.

system-addon/lib/TelemetrySender.jsm
dmose: I'd propose we land as-is and adjust when we do our performance event audit in or after that meeting.  Or Kate can adjust as necessary in her TabSender patch.

system-addon/lib/TelemetrySender.jsm
dmose: Yeah, logging is the way to go.  I've switched to use Console.jsm.

system-addon/test/unit/lib/TelemetrySender.test.js
dmose: It's both the fake prefs, and possibly wanting to spy on the Preferences constructor that sometimes need to happen first.

system-addon/test/unit/lib/TelemetrySender.test.js
dmose: :-)

as-pine-proxy:
[Treeherder: 19aa997281c53eb3137a5d491f24c9d30aa8b82d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=19aa997281c53eb3137a5d491f24c9d30aa8b82d)

Issue #2423, opened 2017-04-13
Add mechanism for disabling feeds, disable during pine tests


k88hudson:
We need this to green up pine.

sarracini:
Fixed with #2434

PR #2434, opened 2017-04-17
feat(mc) #2423 Add pref-based initialization for feeds

k88hudson:
This patch adds a pref-based lazy-loading of method of turning feeds off and on.

Prefs can be stand-alone or experiment prefs depending on what is needed.
 
I still need to add the pref flipping for tests.

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -16,6 +16,18 @@ XPCOMUtils.defineLazyModuleGetter(this, "NewTabInit",
 XPCOMUtils.defineLazyModuleGetter(this, "TopSitesFeed",
   "resource://activity-stream/lib/TopSitesFeed.jsm");
 
+const feeds = {
+  // When you add a feed here:
+  // 1. The key in this object should directly refer to a pref, not including the
+  //    prefix (so "feeds.newtabinit" refers to the
+  //    "browser.newtabpage.activity-stream.feeds.newtabinit" pref)
+  // 2. The value should be a function that returns a feed.
+  // 3. You should use XPCOMUtils.defineLazyModuleGetter to import the Feed,
+  //    so it isn't loaded until the feed is enabled.
+  "feeds.newtabinit": () => new NewTabInit(),
+  "feeds.topsites": () => new TopSitesFeed()
+};
+
 this.ActivityStream = class ActivityStream {
 
   /**
@@ -30,13 +42,11 @@ this.ActivityStream = class ActivityStream {
     this.initialized = false;
     this.options = options;
     this.store = new Store();
+    this.feeds = feeds;
   }
   init() {
     this.initialized = true;
-    this.store.init([
-      new NewTabInit(),
-      new TopSitesFeed()
-    ]);
+    this.store.init(this.feeds);
     this.store.dispatch({type: at.INIT});
   }
   uninit() {
--- a/system-addon/lib/Store.jsm
+++ b/system-addon/lib/Store.jsm
@@ -1,19 +1,22 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+/* global Preferences */
 "use strict";
 
 const {utils: Cu} = Components;
 
 const {redux} = Cu.import("resource://activity-stream/vendor/Redux.jsm", {});
-const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
 const {reducers} = Cu.import("resource://activity-stream/common/Reducers.jsm", {});
 const {MessageManager} = Cu.import("resource://activity-stream/lib/MessageManager.jsm", {});
 
+const PREF_PREFIX = "browser.newtabpage.activity-stream.";
+Cu.import("resource://gre/modules/Preferences.jsm");
+
 /**
- * Store - This has a similar structure to a redux store, but includes some extra
- *         functionality to allow for routing of actions between the Main processes
- *         and child processes via a MessageManager.
+ * Store - This has a similar structure to a redux store, but includes some
+ *         extra functionality to allow for routing of actions between the Main
+ *         processes and child processes via a MessageManager.
  *         It also accepts an array of "Feeds" on inititalization, which
  *         can listen for any action that is dispatched through the store.
  */
@@ -32,7 +35,9 @@ this.Store = class Store {
         return this._store[method](...args);
       }.bind(this);
     });
-    this.feeds = new Set();
+    this.feeds = new Map();
+    this._feedFactories = null;
+    this._prefHandlers = new Map();
     this._mm = new MessageManager({dispatch: this.dispatch});
     this._store = redux.createStore(
       redux.combineReducers(reducers),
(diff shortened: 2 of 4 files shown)


system-addon/test/unit/lib/Store.test.js
```diff
@@ -1,28 +1,39 @@
...
-  it("should have an empty feeds Set", () => {
-    assert.instanceOf(store.feeds, Set);
+  afterEach(() => {
+    Preferences.clear();
+    globals.reset();
+  });
+  after(() => globals.restore());
+  it("should have an .feeds property that is a Map", () => {
+    assert.instanceOf(store.feeds, Map);
     assert.equal(store.feeds.size, 0, ".feeds.size");
```
dmose: Switch this to ```assert.propertyVal(store.feeds, "size", 0);``` for readability and clearer output?

system-addon/test/unit/lib/Store.test.js
```diff
@@ -38,55 +49,129 @@ describe("Store", () => {
     store.dispatch({type: "FOO"});
     assert.calledOnce(store._mm.middleware);
   });
-  describe("#init", () => {
-    it("should dispatch an init action", () => {
-      sandbox.spy(store, "dispatch");
+  describe("#initFeed", () => {
+    it("should add an instance of the feed to .feeds", () => {
+      class Foo {}
+      store._feedConstructors = {foo: () => new Foo()};
```
dmose: Aren't these really factories rather than constructors?  Changing the name here would make the code a bit easier to read, I think.

dmose: Also, consider switching the tests to setup/execute/verify blocks for easier skimming.

system-addon/test/unit/lib/Store.test.js
```diff
@@ -38,55 +49,129 @@ describe("Store", () => {
...
+    it("should add a .store property to the feed", () => {
+      class Foo {}
+      store._feedConstructors = {foo: () => new Foo()};
+      store.initFeed("foo");
 
-      store.init([]);
-
-      assert.calledWith(store.dispatch, {type: at.INIT});
+      assert.property(store.feeds.get("foo"), "store");
+      assert.equal(store.feeds.get("foo").store, store);
```
dmose: assert.propertyVal might be a slight win here too.

system-addon/lib/Store.jsm
```diff
@@ -53,20 +58,73 @@ this.Store = class Store {
   }
 
   /**
+   * initFeed - Initializes a feed by calling its constructor function
+   *
+   * @param  {string} feedName The name of a feed, as defined in the object passed to Store.init
+   */
+  initFeed(feedName) {
+    const subscriber = this._feedConstructors[feedName]();
```
dmose: Subscriber seems like a confusing name for this variable.  How about "feed"?

system-addon/lib/Store.jsm
```diff
@@ -53,20 +58,73 @@ this.Store = class Store {
...
+    this.feeds.set(feedName, subscriber);
+  }
+
+  /**
+   * uninitFeed - Removes a feed and calls its uninit function if defined
+   *
+   * @param  {string} feedName The name of a feed, as defined in the object passed to Store.init
+   */
+  uninitFeed(feedName) {
+    const subscriber = this.feeds.get(feedName);
```
dmose: Same comment re the subscriber name as above...

system-addon/test/unit/lib/Store.test.js
```diff
@@ -38,55 +49,129 @@ describe("Store", () => {
...
 
-      assert.calledWith(store.dispatch, {type: at.INIT});
+      assert.isTrue(store.feeds.has("foo"), "foo is set");
+      assert.instanceOf(store.feeds.get("foo"), Foo);
     });
-    it("should dispatch an init action", () => {
-      sandbox.spy(store, "dispatch");
+    it("should add a .store property to the feed", () => {
+      class Foo {}
+      store._feedConstructors = {foo: () => new Foo()};
```
dmose: Probably not worth fixing in these tests, but for the future, it seems like we could stick to public APIs by simply creating a new store and adding Foo to calling store.init().  This will make the tests more useful for refactorings.

system-addon/test/unit/lib/Store.test.js
```diff
@@ -38,55 +49,129 @@ describe("Store", () => {
...
+      });
+    });
+  });
+  describe("#init", () => {
+    it("should set ._feedConstructors to the object passed to init", () => {
+      class Foo {}
+      const feeds = {foo: () => new Foo(), bar: () => new Foo()};
+      store.init(feeds);
+      assert.equal(store._feedConstructors, feeds);
+    });
```
dmose: I think we could test public rather than private APIs, in this case by simply calling initFeed and seeing whether it makes the right call...

system-addon/test/unit/lib/Store.test.js
```diff
@@ -38,55 +49,129 @@ describe("Store", () => {
...
+    it("should call .maybeStartFeedAndListenForPrefChanges with each key", () => {
+      sinon.stub(store, "maybeStartFeedAndListenForPrefChanges");
```
dmose: Should this be sandbox.stub?

system-addon/test/unit/lib/Store.test.js
```diff
@@ -38,55 +49,129 @@ describe("Store", () => {
...
 
       assert.equal(store.feeds.size, 0);
-    });
-    it("should dispatch an uninit action", () => {
-      sandbox.spy(store, "dispatch");
-
-      store.uninit();
-
-      assert.calledWith(store.dispatch, {type: at.UNINIT});
+      assert.equal(store._prefHandlers.size, 0);
```
dmose: These two could be assert.propertyVal, I think.

system-addon/lib/Store.jsm
```diff
@@ -53,20 +58,73 @@ this.Store = class Store {
...
+      subscriber.uninit();
+    }
+    this.feeds.delete(feedName);
+  }
+
+  /**
+   * maybeStartFeedAndListenForPrefChanges - Listen for pref changes that turn a feed off/on,
+   *                                         and as long as that pref was not explicitly set to
+   *                                         false, initialize the feed immediately.
+   *
```
dmose: 80 columns wrapping for this and other comments would be nice.

system-addon/lib/Store.jsm
```diff
@@ -53,20 +58,73 @@ this.Store = class Store {
...
+
+    // If the pref is true, start the feed immediately.
+    if (Preferences.get(prefName)) {
+      this.initFeed(name);
+    }
+  }
+
+  /**
    * init - Initializes the MessageManager channel, and adds feeds.
    *        After initialization has finished, an INIT action is dispatched.
```
dmose: It looks to me like the  INIT action is no longer dispatched.

system-addon/lib/Store.jsm
```diff
@@ -76,10 +134,14 @@ this.Store = class Store {
    * @return {type}  description
```
dmose: Looks like the UNINIT dispatching is gone, so updating the comment would be good.

system-addon/lib/ActivityStream.jsm
```diff
@@ -16,6 +16,17 @@ XPCOMUtils.defineLazyModuleGetter(this, "NewTabInit",
 XPCOMUtils.defineLazyModuleGetter(this, "TopSitesFeed",
   "resource://activity-stream/lib/TopSitesFeed.jsm");
```
k88hudson: Yep! 😎

system-addon/test/unit/lib/Store.test.js
k88hudson: In this case `store` is created for each individual test, so it's not really necessary

system-addon/test/unit/lib/Store.test.js
k88hudson: Wow, I didn't know that existed... is that a newer one or something?

system-addon/test/unit/lib/Store.test.js
k88hudson: In this case don't think it's really necessary, since `size` is a standard property of all `Set`s/`Map`s

system-addon/test/unit/lib/Store.test.js
dmose: I'm not sure.  I've just sort of gotten used to current reality that there are lots of good, specific assertions, so typically when I notice a .equal, I immediately wonder if there isn't something a bit clearer... :-)

system-addon/test/unit/lib/Store.test.js
dmose: True.  I guess I tend to bias towards always using sandbox except when it's super obvious (eg an anonymous stub is assigned directly to a property), for two reasons: if I always do the conservative thing, it's impossible for me to make that particular thinko, and it tends to be more a bit more robust against code changes or copy-pasting of code.

system-addon/test/unit/lib/Store.test.js
dmose: Fair enough.

as-pine-proxy:
[Treeherder: 5d6e88c784ca72dd35b6f0a15701d867f8bd0061](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5d6e88c784ca72dd35b6f0a15701d867f8bd0061)

PR #2446, opened 2017-04-19
TelemetryFeed replacing tab tracker

k88hudson:
Still need a bit of feedback for this and to write some tests.


docs/v2 (system-addon)/data_dictionary.md
```diff
@@ -1,87 +1,82 @@
 # Activity Stream Pings
```
k88hudson: @emtwo Can you review 54bef6d4d5b78ac90e272d69f90d54d1ef789e6b to make sure everything looks reasonable?

dmose:
I'd suggest changing the dir names to not use spaces or parens so that they're easier to poke at in a shell (eg change them to v1-test-pilot and v2-system-addon).

system-addon/common/Reducers.jsm
```diff
@@ -6,6 +6,11 @@
 const {actionTypes: at} = Components.utils.import("resource://activity-stream/common/Actions.jsm", {});
 
 const INITIAL_STATE = {
+  App: {
+    init: false,
```
dmose: I'm guessing this wants to be "initialized"?  If not, a different name or a JSDoc comment would help.

Also, the changes to this file want unit tests.

system-addon/test/unit/lib/TelemetrySender.test.js
```diff
@@ -122,87 +96,49 @@ describe("TelemetrySender", () => {
...
-
-    it("should not call this._sendPing for 'nonexistent-topic'", () => {
-      const fakeSubject = "fakeSubject";
-      tSender = new TelemetrySender(tsArgs);
-
-      tSender.observe(fakeSubject, "nonexistent-topic", fakePingJSON);
+      await tSender.sendPing(fakePingJSON);
 
-      assert.notCalled(TelemetrySender.prototype._sendPing);
+      assert.called(global.dump);
```
k88hudson: Whoops, I didn't even see the Console.jsm import! I didn't even know this existed... ha!

In that case, do you mind if I just change the global override of `console`? I was stumped for quite a while trying to figure out why console wasn't working in test debugging...

dmose: What global override of console?

k88hudson: In `unit-entry.js`, there's an override that stubs console

k88hudson:
Ok, I addressed most of your changes here – can you maybe take a look at the visibility changes and see if that makes it clear enough? https://github.com/mozilla/activity-stream/pull/2446/commits/26fb80247a2ae7066df9dfa74814b8ddd180e6cd

system-addon/test/unit/lib/TelemetrySender.test.js
dmose: What do you want to change it to?  It kinda seems like the right thing to do is to simply make it spy on the real console.log, rather than stubbing it using an anonymous spy.

k88hudson: I just added a stub on the individual test instead of globally, although spying would work too

k88hudson:
@dmose, ok I think I got all of review changes you suggested

pdehaan:
Note that renaming the data_dictionary.md file will break the link from Test Pilot at http://testpilot.dev.mozaws.net/experiments/activity-stream

Re: "You can learn more about the data collection for ⁨Activity Stream⁩ here." link at the bottom which currently points to https://github.com/mozilla/activity-stream/blob/master/docs/data_dictionary.md

k88hudson:
@pdehaan ok, I'll file a follow-up ticket to fix that

system-addon/test/unit/lib/TelemetryFeed.test.js
```diff
@@ -0,0 +1,262 @@
...
+      addSession("foo");
+      assert.isTrue(instance.sessions.has("foo"));
+    });
+    it("should set the start_time", () => {
+      sandbox.spy(Components.utils, "now");
+      const session = addSession("foo");
+      assert.calledOnce(Components.utils.now);
+      assert.equal(session.start_time, Components.utils.now.firstCall.returnValue);
+    });
+    it("should se the session_id", () => {
```
dmose: s/se/set/

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * addSession - Start tracking a new session
+   *
+   * @param  {string} id the portID of the open session
+   */
+  addSession(id) {
+    this.sessions.set(id, {
+      start_time: Components.utils.now(),
+      session_id: String(gUUIDGenerator.generateUUID()),
+      page: "NEW_TAB" // TODO: Handle about:home
```
dmose: File a spinoff issue for this?

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -0,0 +1,162 @@
...
+
+  createPerformanceEvent(action) {
+    return Object.assign(
+      this.createPing(au.getPortIdOfSender(action)),
+      action.data,
+      {action: "activity_stream_performance"}
+    );
+  }
+
+  createSessionEndEvent(session) {
```
dmose: The name is a bit non-intuitive here, since this is really a SessionDurationEvent, not a SessionEndEvent.

system-addon/test/unit/lib/TelemetryFeed.test.js
```diff
@@ -0,0 +1,262 @@
...
+      const session = addSession("foo");
+      instance.endSession("foo");
+
+      // Did we call sendEvent with the result of createSessionEndEvent?
+      assert.calledWith(instance.createSessionEndEvent, session);
+      assert.calledWith(instance.sendEvent, instance.createSessionEndEvent.firstCall.returnValue);
+    });
+  });
+  describe("ping creators", () => {
+    beforeEach(async () => await instance.init());
```
dmose: I'm confused.  This looks like it declares an anonymous function which can't ever be called, since there's no name or reference to call it by.  Am I missing something?

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -0,0 +1,162 @@
...
+
+    session.session_duration = Math.round(Components.utils.now() - session.start_time);
+    this.sendEvent(this.createSessionEndEvent(session));
+    this.sessions.delete(portID);
+  }
+
+  /**
+   * createPing - Create a ping with common properties
+   *
+   * @param  {string} id The portID of the session
```
dmose: This appears to be optional; it's probably worth marking as such and describing the semantics of not passing it.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -0,0 +1,162 @@
...
+      {
+        session_id: session.session_id,
+        page: session.page,
+        session_duration: session.session_duration,
+        action: "activity_stream_session"
+      }
+    );
+  }
+
+  sendEvent(event) {
```
dmose: Does this really want to be a public API?

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    session.session_duration = Math.round(Components.utils.now() - session.start_time);
+    this.sendEvent(this.createSessionEndEvent(session));
+    this.sessions.delete(portID);
+  }
+
+  /**
+   * createPing - Create a ping with common properties
+   *
```
dmose: Do all of these create* methods want to be public APIs?

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -0,0 +1,162 @@
...
+      {action: "activity_stream_event"}
+    );
+  }
+
+  createUndesiredEvent(action) {
+    return Object.assign(
+      this.createPing(au.getPortIdOfSender(action)),
+      {value: 0}, // Default value
+      action.data,
+      {action: "activity_stream_masga_event"}
```
dmose: "masga" is a singularly unenlightening descriptor.  I don't know that you need to fix it here, but...

dmose:
r=dmose for everything except the data_events.md piece.  Note there is also a conflict reported by github here.

system-addon/test/unit/lib/TelemetryFeed.test.js
k88hudson: What do you mean by it can't ever be called? `beforeEach` calls it before every test

system-addon/lib/TelemetryFeed.jsm
k88hudson: I agree, I think that would be something we have to change on the data pipeline side though (or at least talk to marina about)

system-addon/lib/TelemetryFeed.jsm
k88hudson: I mean... you could argue the semantics here, right now duration is a major feature of the session end event, but the purpose of it is not for duration exclusively – it does contain lots of other information about the session.

system-addon/lib/TelemetryFeed.jsm
k88hudson: None of the feed functions (except `onAction`) are really public, so I feel like it would be redundant to make them all `_`.

system-addon/common/Actions.jsm
```diff
@@ -3,19 +3,29 @@
...
 "use strict";
 
-const MAIN_MESSAGE_TYPE = "ActivityStream:Main";
-const CONTENT_MESSAGE_TYPE = "ActivityStream:Content";
+this.MAIN_MESSAGE_TYPE = "ActivityStream:Main";
+this.CONTENT_MESSAGE_TYPE = "ActivityStream:Content";
+this.UI_CODE = 1;
+this.BACKGROUND_PROCESS = 2;
+
+// Are we in UI code (i.e. react, a dom) or some kind of background process?
```
k88hudson: I just exported it so that tests can check it if necessary. I can add a comment about that

system-addon/test/unit/common/Actions.test.js
```diff
@@ -1,10 +1,26 @@
 const {
+  actionTypes: at,
   actionCreators: ac,
   actionUtils: au,
   MAIN_MESSAGE_TYPE,
-  CONTENT_MESSAGE_TYPE
+  CONTENT_MESSAGE_TYPE,
+  UI_CODE,
+  BACKGROUND_PROCESS,
```
k88hudson: I disagree in this case.

When constants refer to something external to the module, such as a DOM event or a URL, having a separate source of truth is useful to test for errors/typos in the constants themselves.

These constants, on the other hand, do not refer to anything external to the module; they exist only in the context of the module itself. They're only created for *internal* consistency.  For that reason, there is no benefit to creating a separate, external source of a truth.

system-addon/test/unit/content-src/lib/detect-user-session-start.test.js
```diff
@@ -0,0 +1,56 @@
...
+describe("detectUserSessionStart", () => {
+  describe("#init", () => {
+    it("should call ._sendEvent immediately if the document is visible", () => {
+      const mockDocument = {visibilityState: "visible"};
+      const instance = new DetectUserSessionStart({document: mockDocument});
+      sinon.stub(instance, "_sendEvent");
+
+      instance.init();
+
+      assert.calledOnce(instance._sendEvent);
```
k88hudson: personally I think individual units/functions should be tested separately whether they are public or private since isolating breakages during refactors is easier, and changing private apis does cause (private) breakages anyway

system-addon/lib/TelemetryFeed.jsm
k88hudson: https://github.com/mozilla/activity-stream/issues/2510

system-addon/content-src/activity-stream.jsx
```diff
@@ -1,10 +1,12 @@
-/* globals addMessageListener, removeMessageListener */
 const React = require("react");
 const ReactDOM = require("react-dom");
 const Base = require("content-src/components/Base/Base");
 const {Provider} = require("react-redux");
 const initStore = require("content-src/lib/init-store");
 const {reducers} = require("common/Reducers.jsm");
+const DetectUserSessionStart = require("content-src/lib/detect-user-session-start");
+
+new DetectUserSessionStart().init();
```
k88hudson: I changed `init` to be more descriptive: `sendEventOrAddListener`

system-addon/test/unit/lib/TelemetryFeed.test.js
dmose: Oh, I see!  My brain had magically interpolated a set of curly braces around the function, where there actually are none.  Never mind!

system-addon/test/unit/common/Actions.test.js
dmose: Ah, I had assumed that because they were exported, it was fair game for external callers to use them too.   Which is one reason it may be worth doing the _ stuff.  *shrug*

system-addon/content-src/activity-stream.jsx
dmose: Sounds good.

system-addon/test/unit/content-src/lib/detect-user-session-start.test.js
dmose: Yeah, it's always a judgement call.  I tend to bias towards using public APIs, but doing it the way you've done it isn't unreasonable either.

system-addon/lib/TelemetryFeed.jsm
dmose: Yeah, fair enough.

system-addon/lib/TelemetryFeed.jsm
dmose: The intent is to make it harder for clients of the API to make mistakes, and new readers of the code itself to have a simple clarity.

dmose:
Feel free to fix any of my nits on this iteration as you see fit, or not.  :-)  r=dmose

PR #2456, opened 2017-04-20
feat(system-addon): Closes #2455 start AS on browser-ui-startup-complete

### LINKED ISSUE #2455: Only start system add-on on browser-ui-startup-complete notification
This way we don't have to check if a bunch of modules that we need are ready yet. They will be because this notification fires in nsBrowserGlue after they've been initialized. That's a good time for us to be sure that we can initialize Activity Stream

cc @Mardak @k88hudson @dmose

sarracini:
Start the system-addon once the browser is ready

cc @Mardak

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -1,16 +1,18 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-/* globals Components, XPCOMUtils, Preferences, ActivityStream */
+/* globals Components, XPCOMUtils, Preferences, Services, ActivityStream */
 "use strict";
 
 const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Preferences.jsm");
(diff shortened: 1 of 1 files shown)


system-addon/bootstrap.js
```diff
@@ -69,22 +84,22 @@ function onPrefChanged(isEnabled) {
...
+  // Only start Activity Stream up when the browser UI is ready
+  Services.obs.addObserver(observe, BROWSER_READY_NOTIFICATION);
+
   // Cache startup data which contains stuff like the version number, etc.
   // so we can use it when we init
   startupData = data;
+  startupReason = reason;
 
   // Listen for changes to the pref that enables Activity Stream
   Preferences.observe(ACTIVITY_STREAM_ENABLED_PREF, onPrefChanged);
```
k88hudson: To be extra safe, should we maybe return early in the pref handler if the browser isn't ready yet?

sarracini: Sure, is just adding a ```browserReady``` bool which is set to ```true``` when we've observed the notification enough?

Mardak: Alternatively only add the pref observer after `BROWSER_READY`.

as-pine-proxy:
[Treeherder: 159d463e4c456d50d045a25f05805dd92481dd35](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=159d463e4c456d50d045a25f05805dd92481dd35)

PR #2459, opened 2017-04-21
chore(build): Read pocket api key from ENV

k88hudson:
This patch allows us to pick up an API key from `process.env` for pocket


k88hudson:
Whoever has time to review

jaredkerim:
I think we decided that we will store the API key in the proxy the way we do now, so the client doesn't need to know the API key.

k88hudson:
@jaredkerim happy to refactor next week, but is it cool if we just do this for now so dev will work?

jaredkerim:
@k88hudson yep sounds good!

as-pine-proxy:
[Treeherder: aef161d199fcee3b8e5b2fb6fb7eb5a45cb822f9](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=aef161d199fcee3b8e5b2fb6fb7eb5a45cb822f9)

PR #2465, opened 2017-04-21
feat(content): Refresh pocket feed on pref change

csadilek:
Triggering a feed refresh on pref. change so that when experiment is turned on, firefox doesn't have to be restarted to see data.


as-pine-proxy:
[Treeherder: 73d3f203c63c5aec21b15d10e36783c592c18d13](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=73d3f203c63c5aec21b15d10e36783c592c18d13)

Issue #2467, opened 2017-04-24
Dismissing a Pocket story doesn't add a new one


jaredkerim:
AS Version: Build: 1.10.0-dev-1492813147-7478ed9

If I miss a Pocket story, it should fill in another one, I know the API returns more than 3.

![screenshot 2017-04-24 10 25 50](https://cloud.githubusercontent.com/assets/119884/25341905/6ac0a432-28d8-11e7-952d-071a37a97006.png)

Issue #2463, opened 2017-04-21
Fix string or add comment for "Trending Stories by [pocket]"


Mardak:
@csadilek added a string in #2449
```
header_stories=Trending Stories by
```
https://github.com/mozilla/activity-stream/pull/2449/files#diff-906742e1af7418eb557312dc9217f2f1R6

The string is used in a context where an image is placed after it, specifically right now the pocket logo.

We might be able to fix this by making the string be something like "Trending Stories by {trending_logo}" although that might need a dangerously_set_html thing?

At minimum we should probably have a LOCALIZATION_NOTE explaining how the string is being used, so some locales can work around the issue with something like "Trending Stories by:" (note the colon) if grammar rules allow…

@flodolo suggestions?

flodolo:
We definitely need a localization note. 

As you said, we could use a placeholder and replace it with the proper HTML code, but would layout support that, e.g. image at the beginning or in the middle?

Mardak:
Sounds like the experiment will be en-US only, so I guess when we export strings to localize, I'll explicitly ignore the pocket commits for now.

jaredkerim:
@k88hudson and @Mardak need to discuss an effective strategy for this

Mardak:
Here's the list of recently added pocket-related strings:
```ini
header_stories=Trending Stories by 
type_label_recommended=Recommended
menu_action_save_to_pocket=Save to Pocket
settings_pane_pocketstories_header=Trending Stories
settings_pane_pocketstories_body=Pocket™, a part of the Mozilla family, will help connect you to high-quality content that you may not have found otherwise.
pocket_read_more=Read More
pocket_read_even_more=View More Stories
```

Which of these will users see even if the experiment/feature is turned off? And which should only be available to en-US?

csadilek:
@Mardak The new plan is that the experiment will be active for everyone but only prefed on for 20% of the users. So, everyone will see the preference pane and everyone might see the remaining strings if they decide to turn on the feature. If possible, I would like to keep the remaining localization strings if the don't cause any concerns (these will be useful in other locales later.)

csadilek:
@Mardak we've added a localization note now. Can this be closed?

Mardak:
Let's wait for the sign off in https://github.com/mozilla/activity-stream-l10n/pull/1

Mardak:
There was just one issue noted: "This sounds more like a developer note. Can you explain what "PocketStories component" is?"

[...]

PR #2479, opened 2017-04-24
Minor fixes to Pocket experiment issues (see below)

csadilek:
https://github.com/mozilla/activity-stream/issues/2467 Dismissing a Pocket story doesn't add a new one

https://github.com/mozilla/activity-stream/issues/2463 Fix string or add comment for "Trending Stories by [pocket]"

### DIFF
--- a/addon/Feeds/PocketStoriesFeed.js
+++ b/addon/Feeds/PocketStoriesFeed.js
@@ -1,7 +1,6 @@
 /* globals Task */
 const {Cu} = require("chrome");
 const PocketFeed = require("./PocketFeed");
-const {POCKET_STORIES_LENGTH} = require("common/constants");
 const am = require("common/action-manager");
 const {pocket_story_endpoint, pocket_consumer_key} = require("../../pocket.json");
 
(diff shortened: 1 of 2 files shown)


csadilek:
The pocket experiment is targeting EN only. So we can get rid of this problematic localization string for now.

content-src/components/PocketStories/PocketStories.js
```diff
@@ -59,7 +59,7 @@ const PocketStories = React.createClass({
 
     return (<section className="pocket-stories spotlight">
       <h3 className="section-title">
-        <FormattedMessage id="header_stories" />
+        Trending Stories By
```
Mardak: From #2463, it sounds like this will potentially be shown to any user including non-en-US, so we would want this localized.

csadilek: @Mardak we just had some wires crossed here. I will change this commit to add a localization note to header_stories..

csadilek:
@Mardak @jaredkerim Commits updated. Thanks.

as-pine-proxy:
[Treeherder: 5f82aa95adb68c0a951dd25658c4b909fe614502](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5f82aa95adb68c0a951dd25658c4b909fe614502)

PR #2480, opened 2017-04-24
 Remove metadata cache expiry fixes #2352

### LINKED ISSUE #2352: Remove metadata cache expiry
After discussion, we've decided that expiring metadata goes against our current approach of trying to maximize metadata coverage for history so we will expire all logic for removing metadata that has expired.

jaredkerim:
We no longer need to expire metadata because we are making assumptions that it will be long lived in our future designs for higlights.

### DIFF
--- a/addon/MetadataStore.js
+++ b/addon/MetadataStore.js
@@ -28,8 +28,7 @@ const SQL_DDLS = [
       type VARCHAR(32),
       description TEXT,
       media_url LONGVARCHAR,
-      created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
-      expired_at LONG
+      created_at DATETIME DEFAULT CURRENT_TIMESTAMP
   )`,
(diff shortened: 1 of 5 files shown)


ncloudioj:
I'd recommend to simply disable the data expiry job or use a longer `expired_at` for now to fix #2352, other than completely removing the expiry functionality.

My main concern is that letting the database grow without bound will almost certainly impact the query performance. Places used to suffer from the same issue, and it's currently using an empirical hard limit (60MB or 100K places) to trigger the data expiry, in order to maintain a decent performance. As an example, my `metadata.sqlite` is already 10MB since the beginning of `metadataLongCache` experiment, and it'll grow pretty fast over time, which could make the query run much slower if we don't do any data expiry.

jaredkerim:
@ncloudioj Okay, you're right let's just increase the cap to 90 days for all users.

PR #2485, opened 2017-04-25
feat (experiment): #2300 add experiment for two rows of top sites by default

### LINKED ISSUE #2300: Experiment showing two rows of Top Sites
Now that we can switch to two rows of top sites via a pref, it should be pretty easy to create an experiment where we show two rows of top sites by default.

The UI for customization isn't rolled out yet and won't until it's all ready, so the users would be stuck with two rows and won't be able to switch it back to one if they don't like it.

rlr:
r? @jaredkerim is this how you envisioned this being done?

Fixes #2300

### DIFF
--- a/addon/ExperimentProvider.js
+++ b/addon/ExperimentProvider.js
@@ -72,6 +72,8 @@ exports.ExperimentProvider = class ExperimentProvider {
     prefService.set(PREF_PREFIX + experimentId, variant.value);
     if (experimentId === "pocket") {
       simplePrefs.prefs.showPocket = true;
+    } else if (experimentId === "topSitesTwoRowsDefault") {
+      simplePrefs.prefs.showMoreTopSites = true;
     }
     this.emit("experimentEnrolled", {id: experimentId, variant});
   }
--- a/content-test/addon/ExperimentProvider.test.js
+++ b/content-test/addon/ExperimentProvider.test.js
@@ -421,5 +421,13 @@ describe("ExperimentProvider", () => {
       experimentProvider.enroll(experimentId, variant);
       assert.isTrue(simplePrefs.prefs.showPocket);
     });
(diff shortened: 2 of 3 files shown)


addon/ActivityStreams.js
```diff
@@ -415,6 +418,21 @@ ActivityStreams.prototype = {
...
+  },
+
+  _unsetupTopSitesTwoRowsExperiment() {
+    this._experimentProvider.off("experimentEnrolled", this._handleExperimentEnrollment);
+  },
+
+  _handleExperimentEnrollment(eventName, experiment) {
+    // If we are enrolling into the topSitesTwoRows experiment, turn on the showMoreTopSites pref.
+    if (experiment.id === "topSitesTwoRowsDefault") {
+      simplePrefs.prefs.showMoreTopSites = true;
```
jaredkerim: Just move this right to here:

https://github.com/mozilla/activity-stream/blob/master/addon/ExperimentProvider.js#L73

instead of having it in ActivityStream.js

Christian ran into annoying timing issues and we don't need to be fancy with it, it's fine if it just lives right in the experiment provider.

test/test-ActivityStream-experiment.js
```diff
@@ -0,0 +1,32 @@
+/* globals Services */
```
jaredkerim: Just test this right on Experiment Provider.

rlr: oh nice. that should be easier!

as-pine-proxy:
[Treeherder: 6fda76d96a51fdc876ba07639fc164d56ae034a1](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6fda76d96a51fdc876ba07639fc164d56ae034a1)

Issue #2466, opened 2017-04-24
Enable pocket pref for all users


jaredkerim:
Right now the ability to turn pocket on or off is hidden behind the experiment pref, but I think we should enable the pref for all users, and then we can have:

Control: Pocket default off
Experiment: Pocket default on

And this will enable us to

1) Roll out pocket more quickly to everybody
2) Measure how many people explicitly opt in
3) Measure how many people explicitly opt out

I think this will let us roll out the feature more quickly and measure more in less time.

k88hudson:
After discussing the details of this, it might be better to just stick with the original plan here (i.e. to only add the settings pref on by default for the variant group, and remove it entirely for the control) for a few reasons:

- flipping the pref default states and accounting for user-set states turns out to involve more work than we anticipated, for not a lot of additional useful data
- it's not clear to do with users in control who decide to turn it on – should we remove their data from the control set? etc.

What do you think @tspurway @emtwo @jaredkerim

PR #2492, opened 2017-04-25
feat(content): #2466 Enable pocket pref for all users

csadilek:
This makes sure pocket is available to all users, but prefed off for the control group. The preference pane allows to opt-in to the experiment.

Had to put the condition to set the simplePrefs.prefs.showPocket into enroll() directly, as discussed, as the "experimentEnrolled" event is firing too late in this case.

### DIFF
--- a/addon/ExperimentProvider.js
+++ b/addon/ExperimentProvider.js
@@ -70,6 +70,9 @@ exports.ExperimentProvider = class ExperimentProvider {
   enroll(experimentId, variant) {
     this._experimentId = variant.id;
     prefService.set(PREF_PREFIX + experimentId, variant.value);
+    if (experimentId === "pocket") {
+      simplePrefs.prefs.showPocket = true;
+    }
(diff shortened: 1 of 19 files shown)


addon/ExperimentProvider.js
```diff
@@ -70,6 +70,9 @@ exports.ExperimentProvider = class ExperimentProvider {
   enroll(experimentId, variant) {
     this._experimentId = variant.id;
     prefService.set(PREF_PREFIX + experimentId, variant.value);
+    if (experimentId === "pocket") {
```
k88hudson: haha, ok that works

as-pine-proxy:
[Treeherder: 77517ec1d94496f37774a4598f39cf75aee06256](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=77517ec1d94496f37774a4598f39cf75aee06256)

PR #2495, opened 2017-04-26
fix(content): Improve localization values and notes for Pocket (read more)

### LINKED ISSUE #2463: Fix string or add comment for "Trending Stories by [pocket]"

csadilek:
Fix #2463. See https://github.com/mozilla/activity-stream-l10n/pull/1 and https://github.com/mozilla/activity-stream/issues/2463#issuecomment-297171710 for details.

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -92,6 +92,9 @@ edit_topsites_unpin_button=Unpin this site
 edit_topsites_edit_button=Edit this site
 edit_topsites_dismiss_button=Dismiss this site
 
-# LOCALIZATION NOTE (pocket): This is shown in the PocketStories component.
-pocket_read_more=Read More
+# LOCALIZATION NOTE (pocket_read_more): This is shown at the bottom of the
+# trending stories section and precedes a list of links to trending topics.
+pocket_read_more=Read More:
+# LOCALIZATION NOTE (pocket_read_even_more): This is shown as a link at the
+# end of the list of trending topic links.
 pocket_read_even_more=View More Stories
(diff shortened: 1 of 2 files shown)


locales/en-US/strings.properties
```diff
@@ -92,6 +92,9 @@ edit_topsites_unpin_button=Unpin this site
 edit_topsites_edit_button=Edit this site
 edit_topsites_dismiss_button=Dismiss this site
 
-# LOCALIZATION NOTE (pocket): This is shown in the PocketStories component.
-pocket_read_more=Read More
+# LOCALIZATION NOTE (pocket_read_more): This is shown below the Trending Stories
+# by Pocket row, followed by a list of trending topic links.
```
Mardak: The "Trending Stories by Pocket row" seems a bit specific and dependent on the other string. Maybe just
"This is shown at the bottom of the trending section and precedes a list of trending topic text links."

as-pine-proxy:
[Treeherder: a54d28483a6dadc4acde9ceac1ca8241435a6818](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a54d28483a6dadc4acde9ceac1ca8241435a6818)

PR #2516, opened 2017-04-28
Metadata Local Refresh Experiment fixes #2358

### LINKED ISSUE #2358: Local Metadata Refresh Experiment
This experiment will test the configuration of having long metadata cache timeout, no remote service calls, and a background periodic metadata refresh task which queries the PlacesDB for all URLs which do not have metadata and fetches them in the background.

This configuration consists of:

1) Long metadata cache timeout
2) Local Metadata Parser
3) No Metadata Service
4) Periodic background refresh task

jaredkerim:
This experiment will test the configuration of having long metadata cache timeout, no remote service calls, and a background periodic metadata refresh task which queries the PlacesDB for all URLs which do not have metadata and fetches them in the background.

This configuration consists of:

    Long metadata cache timeout
    Local Metadata Parser
    No Metadata Service
    Periodic background refresh task

### DIFF
--- a/addon/Feeds/MetadataFeed.js
+++ b/addon/Feeds/MetadataFeed.js
@@ -5,13 +5,14 @@ const {PlacesProvider} = require("addon/PlacesProvider");
 const Feed = require("addon/lib/Feed");
 const am = require("common/action-manager");
 const MAX_NUM_LINKS = 5;
+const UPDATE_TIME = 24 * 60 * 60 * 1000; // Update once per day
 
 Cu.import("resource://gre/modules/NewTabUtils.jsm");
 
 module.exports = class MetadataFeed extends Feed {
   constructor(options) {
     super(options);
-    this.linksToFetch = new Map();
+    this.urlsToFetch = [];
     this.pinnedLinks = options.pinnedLinks || NewTabUtils.pinnedLinks;
   }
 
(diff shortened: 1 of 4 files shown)


addon/Feeds/MetadataFeed.js
```diff
@@ -54,28 +57,37 @@ module.exports = class MetadataFeed extends Feed {
...
       case am.type("RECEIVE_PLACES_CHANGES"):
-        this.linksToFetch.set(action.data.url, Date.now());
-        if (this.linksToFetch.size > MAX_NUM_LINKS) {
+        this.urlsToFetch.push(action.data.url);
+
+        if (experiments.metadataLocalRefresh && ((Date.now() - this.lastRefreshed) >= UPDATE_TIME)) {
+          this.getInitialMetadata("periodically refreshing metadata");
+        }
+
+        if (this.urlsToFetch.length > MAX_NUM_LINKS) {
```
rlr: does the order of these `if` statements matter? I can see the `getInitialMetadata` above would cause this refresh below to be way down at the end of the queue. But it probably doesn't matter for most cases.

jaredkerim: Yeah actually I had to do it in this order, it was finicky and tricky but this was the way that worked.

as-pine-proxy:
[Treeherder: 2bdcf8401b7575e5bc97fd3cdff97794b7585672](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2bdcf8401b7575e5bc97fd3cdff97794b7585672)

PR #2520, opened 2017-05-01
fix(mc): closes #2519, fix err message/suggest workaround for test-merges

### LINKED ISSUE #2519: test-merges fails in hard-to-understand way when code is rebased rather than merged in
When github returns no merge events for a PR that has actually merged, test-merges emits a confusing message.  I have a simple patch that makes the problem clear and suggests a workaround.

dmose:
A simple fix that changes an error message for a github issue to clearly identify the problem and suggest a workaround.  r=@k88hudson?

### DIFF
--- a/bin/test-merges.js
+++ b/bin/test-merges.js
@@ -109,14 +109,17 @@ function getPRMergeCommitId(prNumber) {
     let mergeEvents = data.filter(item => item.event === "merged");
     if (mergeEvents.length > 1) {
       throw new Error("more than one merge event, aborting");
+    } else if (!mergeEvents.length) {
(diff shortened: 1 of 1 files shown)


k88hudson:
Looks good to me. We might want to flip back that setting so we do actually get merge commits again, if that's what the integration relies on

as-pine-proxy:
[Treeherder: 0e3fb6f0c6327c7327b9105e3a320616842e7bcd](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0e3fb6f0c6327c7327b9105e3a320616842e7bcd)

Issue #2518, opened 2017-05-01
Implement Places-related APIs for system addon


k88hudson:
This includes:
- relaying events (including history changes, clearing history, etc.)
- bookmarking API
- deleting API
- blocking API

These can be split into separate tickets if necessary.

k88hudson:
I'm going to grab the relaying events part if that's cool

sarracini:
The events are prefixed with "newtab-": https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/NewTabUtils.jsm#949

To call into the actions use this API: https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/NewTabUtils.jsm#1214
So it would be ```NewTabUtils.activityStreamLinks.deleteHistoryEntry(url);```

PR #2550, opened 2017-05-04
feat(mc) #2518 Add PlacesFeed with observers

k88hudson:
This patch adds observers on various places events we need for bookmarking, blocking, etc.

Note that the block event won't work until https://bugzilla.mozilla.org/show_bug.cgi?id=1361797 gets merged to master (which should be done by tomorrow)


system-addon/lib/PlacesFeed.jsm
```diff
@@ -0,0 +1,192 @@
...
+
+const LINK_BLOCKED_EVENT = "newtab-linkBlocked";
+
+/**
+ * Observer - a wrapper around history/bookmark observers to add the QueryInterface.
+ */
+class Observer {
+  constructor(dispatch, observerInterface) {
+    this.dispatch = dispatch;
+    this.QueryInterface = XPCOMUtils.generateQI([observerInterface, Ci.nsISupportsWeakReference]);
```
k88hudson: @dmose you had a problem with something similar pretty recently – is there any reason why there would be a problem with this approach?

dmose: The main issue I'm aware of is that generateQI will be called for each instance construction, rather than once for the class, which would be a tiny perf hit.  I wouldn't block on that, however.

k88hudson: Ah ok that makes sense

k88hudson:
Ok, I added the deleting/blocking/bookmarking apis in there as well so the context menu can use them

system-addon/lib/PlacesFeed.jsm
```diff
@@ -0,0 +1,204 @@
...
+ /* globals ContentSearch, XPCOMUtils, PlacesUtils, NewTabUtils, Services */
+"use strict";
+
+const {utils: Cu, interfaces: Ci} = Components;
+const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "ContentSearch",
+  "resource:///modules/ContentSearch.jsm");
```
k88hudson: lol oops

system-addon/lib/PlacesFeed.jsm
```diff
@@ -0,0 +1,204 @@
...
+   * @param  {int} parent
+   * @param  {str} guid         The unique id of the bookmark
+   */
+  async onItemChanged(...args) {
+    const property = args[1];
+    const type = args[5];
+    const guid = args[7];
+
+    // Only process this event if it is a TYPE_BOOKMARK, and uri or title was the property changed.
+    if (type !== PlacesUtils.bookmarks.TYPE_BOOKMARK || !["uri", "title"].includes(property)) {
```
k88hudson: tags and folders can cause updates, but we don't care about those – another one I noticed is `favicon`, but it seems that it weirdly gets triggered right after every time a bookmark is added (which seems wrong to me)

sarracini: does changing the uri or the title change the frecency? is that why we only care about those?

k88hudson: Hm, I guess we don't actually display the title either... but if the URI changes, I guess we'll want to remove the bookmark on the old URI? That is kind of edge-casey though

as-pine-proxy:
[Treeherder: 270ca3ef1f3bf7f8963649f6771bc9a116966bb4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=270ca3ef1f3bf7f8963649f6771bc9a116966bb4)

PR #2555, opened 2017-05-05
feat(mc): Closes #2313 Port localization code using react-intl and reusing strings

### LINKED ISSUE #2313: port localization / jar.mn to system addon

Mardak:
Fix #2313. r?@sarracini Main differences is stashing `locale`/`strings` on `App` instead of `Intl` and computing `direction` directly in `Base`. Keeps `LocalizationFeed` but now it's the one place that reads the `locales.json` file.

image
image


system-addon/content-src/components/Base/Base.jsx
```diff
@@ -1,10 +1,61 @@
...
+    addEventListener("visibilitychange", this);
+  }
+  componentWillUpdate({App}) {
+    if (App.locale !== this.props.App.locale) {
+      addLocaleDataForReactIntl(App);
+      this.updateTitle(App);
+    }
+  }
+
+  handleEvent({type}) {
```
Mardak: Oh I think this should be okay. I was thinking of the jsx-no-bind rule, but that's more of creating callbacks from code related to `render`. Here the callback is created and used after first mount. I'll fix.

as-pine-proxy:
[Treeherder: 8ae738e940449332432ed3a725de03451ca43251](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8ae738e940449332432ed3a725de03451ca43251)

PR #2558, opened 2017-05-05
fix(mc): Closes #2548 Send a dummy content message to trigger ContentSearch initialization

### LINKED ISSUE #2548: ContentSearch not initialized
Full message: TypeError: this._stringBundle is undefined
Full stack: this.ContentSearch._currentEngineObj<@resource:///modules/ContentSearch.jsm:508:9

@sarracini This was broken today by https://hg.mozilla.org/mozilla-central/rev/84909c2a4dbb
https://bugzilla.mozilla.org/show_bug.cgi?id=1358921

Instead of `init`ing in `_finalUIStartup`, it only gets loaded when it receives a message.

Mardak:
Fix #2548. @sarracini @k88hudson There must be a better way to do this…?

note: there's two commits with different approaches. The former eagerly inits `ContentSearch` on activity stream init and the latter only inits when the search is rendered/mounted.

### DIFF
--- a/system-addon/content-src/components/Search/Search.jsx
+++ b/system-addon/content-src/components/Search/Search.jsx
@@ -11,6 +11,12 @@ class Search extends React.Component {
     this.onClick = this.onClick.bind(this);
     this.onChange = this.onChange.bind(this);
   }
+
+  componentWillMount() {
+    // Trigger initialization of ContentSearch in case it hasn't happened yet
+    dispatchEvent(new CustomEvent("ContentSearchClient", {detail: {}}));
+  }
+
   performSearch(options) {
     let searchData = {
       engineName: options.engineName,
--- a/system-addon/lib/SearchFeed.jsm
+++ b/system-addon/lib/SearchFeed.jsm
@@ -18,10 +18,19 @@ XPCOMUtils.defineLazyModuleGetter(this, "Services",
 this.SearchFeed = class SearchFeed {
   addObservers() {
     Services.obs.addObserver(this, SEARCH_ENGINE_TOPIC);
+
+    // Notice when ContentSearch.init would be lazily loaded from nsBrowserGlue
+    this.contentSearch = new Promise(resolve => Services.mm.addMessageListener(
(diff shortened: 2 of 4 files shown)


k88hudson:
@Mardak Personally I prefer the first approach because it keeps the code in a single place and will survive front-end refactors etc, but either seems pretty reasonable.

sarracini:
Of the two commits I like the second approach because it's clearer. But really there must be a better way to do this...
Maybe we could change ```ContentSearch``` to listen for an ```init``` message which would call into ```ContentSearch.init()``` and we could send that message on our ```APP_INIT```?
Or maybe it would be simpler to add a guard in ContentSearch's ```init``` and just explicitly call ```ContentSearch.init``` ourselves?

k88hudson:
I think it would be good some kind of fix in before https://bugzilla.mozilla.org/show_bug.cgi?id=1363085 to unbreak us

dmose:
From what I know now, the guard approach sounds simplest and most contained.

Mardak:
> Maybe we could change ```ContentSearch``` to listen for an ```init``` message which would call into ```ContentSearch.init()``` and we could send that message on our ```APP_INIT```?

This is unreliable by itself as `init` will only be called if a page with search is loaded. So in addition to this m-c change, we would need to dispatch the custom event from our page.

> Or maybe it would be simpler to add a guard in ContentSearch's ```init``` and just explicitly call ```ContentSearch.init``` ourselves?

If we were to naively do this, it would be similar to partially backing out the performance gains patch that broke us in the first place albeit just for `ContentSearch`, I would assume this would be `r-`ed. So to make it better, we would still need some way to delay calling `init` (probably on page load) instead of immediately on add-on `INIT`.

as-pine-proxy:
[Treeherder: 4f913a2c0652de767e1374b4d1deae3c07e6477a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4f913a2c0652de767e1374b4d1deae3c07e6477a)

PR #2574, opened 2017-05-09
vendor(mc): Use minified vendor react-intl@2.2.3 for system-addon

Mardak:
r?@k88hudson More followup to #2313 so that react-intl isn't part of the bundle. In preparation for https://bugzilla.mozilla.org/show_bug.cgi?id=1363085#c4 to export as Part 2.

I grabbed the UMD prod link from https://github.com/yahoo/react-intl/wiki#the-react-intl-package for 2.2.3 which is the version in `package.json`, but the latest is 2.3.0. I added the comment block to react-intl.js for the version and license info copied from `node_modules/react-intl/dist/react-intl.js`


Mardak:
Looks like the `unpkg` links might just work for the non-dev `npm` files?

https://unpkg.com/react@latest/dist/react.min.js
https://unpkg.com/react-dom@latest/dist/react-dom.min.js
https://unpkg.com/react-intl@latest/dist/react-intl.min.js
https://unpkg.com/react-redux@latest/dist/react-redux.min.js
https://unpkg.com/redux@latest/dist/redux.min.js

as-pine-proxy:
[Treeherder: 787c1dacdadf8f91ef525167da1610dafc4d536f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=787c1dacdadf8f91ef525167da1610dafc4d536f)

PR #2575, opened 2017-05-09
feat(systemaddon): Closes #2546 Hook in events from places with context menu

### LINKED ISSUE #2533: Bring context menu to system-addon
doing this in 3 tickets:
- [x] bring the UI over which won't do much but it'll be accessible #2544
- [ ] wait for places change stuff to land so i can hook up the proper events #2546
- [x] test for the UI #2545

### LINKED ISSUE #2546: Hook in events from places with context menu
split from #2533

sarracini:
Closes #2546. This is the last part to fix #2533 
I'm going to add some more tests to the context menu to see that their onClick actions get properly dispatched - but right now it shouldn't block the landing of it into m-c


system-addon/common/Reducers.jsm
```diff
@@ -70,6 +70,31 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
...
+          const {bookmarkGuid, bookmarkTitle, lastModified} = action.data;
+          return Object.assign({}, site, {bookmarkGuid, bookmarkTitle, bookmarkDateCreated: lastModified});
+        }
+        return site;
+      });
+      return Object.assign({}, prevState, {rows: newRows});
+    case at.PLACES_BOOKMARK_REMOVED:
+      newRows = prevState.rows.map(site => {
+        if (site.url === action.data.url) {
+          const newSite = site;
```
Mardak: You'll want to `Object.assign` if the intention is to not modify the original `site` object.

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -24,9 +24,23 @@ class ContextMenu extends React.Component {
...
+    switch (event.key) {
+      case "Tab":
+        if (event.shiftKey) {
+          if (option.first) {
+            this.hideContext();
+          }
+          return;
+        }
+        if (option.last) {
+          this.hideContext();
```
Mardak: Should be able to combine this check and body with above:
```js
if (event.shiftKey && option.first || option.last) {
  this.hideContext()
```
With a comment describing that this is to handle tabbing past the front or end

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -24,9 +24,23 @@ class ContextMenu extends React.Component {
...
+        }
+        break;
+      case "Enter":
+        this.hideContext();
+        option.onClick();
```
Mardak: Should we be concerned about the ordering of this logic? I.e., `hideContext` causing the removal instead of just visibility change?

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
```diff
@@ -1,52 +1,98 @@
...
   }
   getOptions() {
-    return [
+    const {dispatch, site} = this.props;
+    const defaultOptions = this.getDefaultContextMenu(dispatch, site);
+
+    // default top sites have a limited set of context menu options
+    if (site.isDefault) {
+      return this.getFirstAndLast(defaultOptions).map(option => {
+        // Convert message ids to localized labels
```
Mardak: Instead of duplicating the `intl` code or needing the `getFirstAndLast` helper function, this code will probably flow better without the short circuit:
```js
getOptions() {
  let options = getDefaultContextMenu();
  // Add more options for regular sites
  if (!site.isDefault) {
    options = [
      { bookmarkStuff },
      { separator },
      ...options, // inserts all default options here with spread operator
      { separator },
      …
    ];
  }
  // convert ids to labels and hook up onClick action
  options.map(option => …);
  // annotate the first and last options for a11y
  options[0].first = true;
  …
}
```

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
```diff
@@ -1,52 +1,98 @@
...
+      action: actionTypes.BOOKMARK_URL,
+      data: site.url
+    });
+  }
+  getDefaultContextMenu(dispatch, site) {
+    return [{
+      id: "menu_action_open_new_window",
+      icon: "new-window",
+      onClick: () => {
+        dispatch(actionCreators.SendToMain({type: actionTypes.OPEN_NEW_WINDOW, data: {url: site.url}}));
```
Mardak: It looks like all of the options have an onClick of `dispatch`ing `SendToMain`, so it each of the options could probably just have an `action` and `data` that gets hooked up for all options.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -73,6 +77,10 @@ this.TopSitesFeed = class TopSitesFeed {
           this.refresh(action);
         }
         break;
+      case at.OPEN_NEW_WINDOW:
+      case at.OPEN_PRIVATE_WINDOW:
+        this.openNewWindow(action);
```
Mardak: This seems a little dangerous to rely on the caller to set `isPrivate`. We can just have the `PRIVATE` case `action = Object.assign({}, action, {isPrivate: true})` and add a comment to explicitly fall through.

system-addon/test/unit/content-src/components/LinkMenu.test.js
```diff
@@ -4,26 +4,36 @@ const {_unconnected: LinkMenu} = require("content-src/components/LinkMenu/LinkMe
...
-    const contextMenuProps = wrapper.find(ContextMenu).props();
-    assert.equal(contextMenuProps.tabbableOptionsLength, 2);
+
+    const options = component.getFirstAndLast(component.getOptions());
+    const firstItem = options[0];
+    const lastItem = options[options.length - 1];
+    assert.equal(firstItem.label, firstItemLabel);
+    assert.isTrue(firstItem.first);
+    assert.equal(lastItem.label, lastItemLabel);
+    assert.isTrue(lastItem.last);
```
Mardak: A negative test that the first item `isTrue(!firstItem.last)`, and similarly for last, but also that the middle item has neither.

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
Mardak: Oh, is this ordered this way because separators don't have an `onClick`?

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
sarracini: I had this originally, but if you're on the last option and you press Shift + Tab, it closes the context menu because it'll satisfy the ```option.last``` criteria, when really what you want is for it to start going back up :(

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
sarracini: Yeah I'm not sure about this, but I think it's ok - the ```onClick``` has the same ordering and so far it hasn't been a problem. I think since it's calling into a function that's defined by it's props, it should still execute the function regardless of visibility

sarracini: You can't tab over a separator so you'd never be able to have a keydown event on it

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
```diff
@@ -1,52 +1,98 @@
...
 
 class LinkMenu extends React.Component {
-  constructor(props) {
-    super(props);
-    this.getOptions = this.getOptions.bind(this);
+  getBookmarkStatus(site) {
+    return (site.bookmarkGuid ? {
+      id: "menu_action_remove_bookmark",
+      icon: "bookmark-remove",
+      action: actionTypes.DELETE_BOOKMARK_BY_ID,
```
Mardak: Not sure if this will be more confusing vs simpler, but the action can be a string "DELETE_BOOKMARK_BY_ID" and let the postprocessing do `let {action, data, id} = option; option.onClick = () => …({type: actionTypes[action], data})` ;)

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
Mardak: Oh oops. shift && first || !shift && last. Some reason looks like there should be a simple binary logic for that..

system-addon/lib/TopSitesFeed.jsm
sarracini: How about doing ``` openNewWindow(action, isPrivate = false)``` and then in the ```PRIVATE``` case call  ```this.openNewWindow(action, true);```. So we're covered both ways, and by default it just opens a regular window

Mardak: Sounds good.

system-addon/test/unit/content-src/components/LinkMenu.test.js
```diff
@@ -4,26 +4,34 @@ const {_unconnected: LinkMenu} = require("content-src/components/LinkMenu/LinkMe
...
     const contextMenuProps = wrapper.find(ContextMenu).props();
-    ["visible", "onUpdate", "options", "tabbableOptionsLength"].forEach(prop => assert.property(contextMenuProps, prop));
+    ["visible", "onUpdate", "options"].forEach(prop => assert.property(contextMenuProps, prop));
   });
   it("should give ContextMenu the correct tabbable options length for a11y", () => {
-    const wrapper = shallowWithIntl(<LinkMenu />);
+    const options = wrapper.find(ContextMenu).props().options;
+    const firstItem = options[0];
+    const lastItem = options[options.length - 1];
+    const middleItem = options[Math.ceil(options.length / 2)];
```
Mardak: ha. I didn't mean the middle item had to be right in the middle ;) 👍

as-pine-proxy:
[Treeherder: 8710fb4ef5ad0359e490c5242d7a9054c0a33f71](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8710fb4ef5ad0359e490c5242d7a9054c0a33f71)

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -73,6 +77,13 @@ this.TopSitesFeed = class TopSitesFeed {
           this.refresh(action);
         }
         break;
+      case at.OPEN_NEW_WINDOW:
+        this.openNewWindow(action);
+        break;
+      case at.OPEN_PRIVATE_WINDOW: {
```
Mardak: Oh, I only noticed this when looking at the export to m-c https://reviewboard.mozilla.org/r/137712/diff/1#index_header but these handlers might be more appropriate somewhere outside of `TopSitesFeed`… although `NewTabInit` doesn't seem quite right either. @k88hudson suggestions?

PR #2581, opened 2017-05-10
Initialize TopSites and Highlights feeds before others fixes #2580

### LINKED ISSUE #2580: Initialize TopSites and Highlights Feeds before any others
We recently introduced new feeds for Pocket and are initializing those before we initialize all of the feeds which block the new tab page from being available (ie not showing Hang Tight).  Even though we do so asynchronously, it's possible that this is still accounting for the increase in Hang Tights have measured for those users which have Pocket enabled.  A simple mitigation strategy is to initialize all feeds required for Activity Stream to be fully loaded (TopSites and Highlights) before we initialize any other feeds.

jaredkerim:


csadilek:
Thanks, I will cherry-pick to the release branch as well.

as-pine-proxy:
[Treeherder: 0f8dbf69318e0bec3122bfe7713f8f8fd6d7a970](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0f8dbf69318e0bec3122bfe7713f8f8fd6d7a970)

Issue #2451, opened 2017-04-19
move most test setup/teardown to {before, after}Each for better test isolation


dmose:
Right now, in the system add-on tests, there are lots of examples of this pattern: 

```javascript
before(() => {
  sandbox = sinon.sandbox.create();
  fetchStub = sandbox.stub(Window.fetch);
});

after(() => {
  sandbox.restore();
});
```

Individual tests often want to modify some stubs in different ways, e.g.

```javascript
it("should handle success well", async () => {
  fetchStub.resolves({ok: true, status: 200});
  ...
});

it("should handle failure well", async () => {
  fetchStub.rejects({ok: false, status: 404});
  ...
});

Particularly now that we are doing more async tests, this is likely to cause shared-state collisions, which will likely show up as intermittent test failures.  I believe we should (whenever practical) create new objects for each test that is run rather than sharing objects, and fix the existing code.

k88hudson:
I agree with this, just a point of clarification: mocha tests (including async ones) will not run simultaneously as long as the test function returns a `Promise` or calls the `done` method appropriately

PR #2585, opened 2017-05-11
tests(mc): #2451 Move test setup to beforeEach instead of before

k88hudson:


dmose:
This looks good; r=dmose.  Note that to close the bug, we also need to tweak the user of overrider in unit-entry.js in the same way as you've done here, so please do that too.

k88hudson:
@dmose that part is actually a bit trickier than I thought it would be since there are some globals that are required at import time... let me see what I can do

Issue #2561, opened 2017-05-08
Persist Pocket dismissals


jaredkerim:
When a user dismisses a Pocket tile, we need to persist which URL the user dismissed so that it does not appear again.  This helps us maintain a consistent user experience between Pocket and Highlights.

PR #2588, opened 2017-05-11
fix(content): #2561 Persist Pocket dismissals

csadilek:
The urls were already persisted (using the same functionality as Highlights), but there was no check in place to filter out these urls so they won't reappear.

### DIFF
--- a/addon/Feeds/PocketStoriesFeed.js
+++ b/addon/Feeds/PocketStoriesFeed.js
@@ -3,6 +3,7 @@ const {Cu} = require("chrome");
 const PocketFeed = require("./PocketFeed");
 const am = require("common/action-manager");
 const {pocket_story_endpoint, pocket_consumer_key} = require("../../pocket.json");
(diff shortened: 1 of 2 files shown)


addon/Feeds/PocketStoriesFeed.js
```diff
@@ -52,7 +53,9 @@ module.exports = class PocketStoriesFeed extends PocketFeed {
         "url": s.dedupe_url,
         "lastVisitDate": s.published_timestamp,
         "pocket": true
-      }));
+      }))
+      .filter(s => !PlacesProvider.links.blockedURLs.has(s.url));
```
ncloudioj: Nit: I'd recommend to do the filtering prior to the map function. That might save some cycles for us.

csadilek: @ncloudioj Thanks, that's a good point. I will update the commit.

PR #2602, opened 2017-05-15
Replace performance ping values with a valid 32 bit integer fixes #2601

### LINKED ISSUE #2601: Change performance ping values to a 32 bit integer
In a few performance pings we use Date.now() which is a 64bit float but we need to constrain that to a 32 bit integer, so instead of sending Date.now() (which isn't a value we really care about) just hard code a value of 1.

jaredkerim:
We can't use Date.now() for this because the ETL job in Onyx will fail to load it into Redshift.

### DIFF
--- a/addon/PageScraper.js
+++ b/addon/PageScraper.js
@@ -46,13 +46,13 @@ PageScraper.prototype = {
    * Parse the HTML and attempt to insert it in the metadata database
    */
   _parseAndSave(rawHTML, url, event) {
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: 28a45a01ce22755836ca48da86633dc99a784983](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=28a45a01ce22755836ca48da86633dc99a784983)

Issue #2605, opened 2017-05-16
Use chrome:// urls for Activity Stream content files 


k88hudson:
We should update Activity Stream to use chrome:// URLs for content files so they cannot be loaded in regular web pages (resource:// URLs can).

This needs to be done in MC first because it requires a change to aboutNewTabService.js, we can port back to Github after that lands (https://bugzilla.mozilla.org/show_bug.cgi?id=1365332)

k88hudson:
After some discussion we decided this isn't necessary (given that we Activity-Stream will be unprivileged and can handle other security issues with CSP)

PR #2606, opened 2017-05-16
chore(mc): #2605 Use chrome:// urls for Activity Stream content files

k88hudson:
This needs https://bugzilla.mozilla.org/show_bug.cgi?id=1365332 to be landed before this lands since it requires a change in MC.


system-addon/content/activity-stream.html
```diff
@@ -12,6 +12,6 @@
     <script src="resource://activity-stream/vendor/react-intl.js"></script>
     <script src="resource://activity-stream/vendor/redux.js"></script>
     <script src="resource://activity-stream/vendor/react-redux.js"></script>
-    <script src="resource://activity-stream/data/content/activity-stream.bundle.js"></script>
+    <script src="chrome://activity-stream/content/activity-stream.bundle.js"></script>
```
Mardak: So now we have an odd mix of `chrome` and `resource` URIs. Couldn't this actually all be resource URIs? And only have this file (`activity-stream.html`) as content to get it loaded via `chrome:`?

This means the `css` is loaded via `resource:` and all the related/relative assets can stay where they are?

system-addon/jar.mn
```diff
@@ -3,13 +3,16 @@
...
+  res/common/ (./common/*)
+  res/vendor/Redux.jsm (./vendor/Redux.jsm)
+  res/vendor/react.js (./vendor/react.js)
+  res/vendor/react-dom.js (./vendor/react-dom.js)
+  res/vendor/react-intl.js (./vendor/react-intl.js)
+  res/vendor/redux.js (./vendor/redux.js)
+  res/vendor/react-redux.js (./vendor/react-redux.js)
+  res/data/locales.json (./data/locales.json)
+
+% content activity-stream %content/
```
Mardak: Actually, would it be sufficient to just have `% content activity-stream %content/data/content/` without touching anything else?

PR #2611, opened 2017-05-18
feat(mc): Implement search suggestions by reusing contentSearchUI

### LINKED ISSUE #2517: Implement search suggestions in Search
We need to implement suggestions and related UI for graduation. Let's work with Bryan on this one to make sure the UI matches up with the style of the new about:newtab search input.

I also noticed there were some changes recently to the search box to the right of the awesomebar in nightly... perhaps we should take that into account?

Mardak:
Fix #2517. I was looking into implementing suggestions, and noticed `contentSearchUI` has all sorts of fancy/featureful/accessible/etc code. Any particular reason we shouldn't reuse all that logic that powers the existing newtab and home pages? It's not the UI as designed for activity-stream and just plain html as opposed to React. ni?@k88hudson

Here's a webm/gif video of a search in activity-stream that can be hooked up with just a few lines adding the script, css, and calling the constructor:

https://people-mozilla.org/~elee/activity-stream/activity-stream.search.webm

If you're not familiar with the keyboard behavior, you can press down to select items (form history as well as search suggestions) and also press alt-down to select a one-off search.

Updated PR with UI tweaks and tests and other cleanup:
image


system-addon/content-src/components/Search/Search.jsx
```diff
@@ -1,39 +1,24 @@
...
-    this.state = {searchString: ""};
     this.onClick = this.onClick.bind(this);
-    this.onChange = this.onChange.bind(this);
   }
 
-  componentWillMount() {
-    // Trigger initialization of ContentSearch in case it hasn't happened yet
-    dispatchEvent(new CustomEvent("ContentSearchClient", {detail: {}}));
+  componentDidMount() {
+    let searchText = document.getElementById("search-input");
```
k88hudson: I would use a `ref` for this instead of an `id`

k88hudson:
This is great! I am all for avoiding duplication of code 👍 

I think the only thing needed here is a bit of polish on the edge where the suggestions touch the bottom of the input (it looks a little rough at the moment)

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -184,7 +184,6 @@
   input {
     @include search-input;
     border-top-left-radius: $search-border-radius;
-    border-bottom-left-radius: $search-border-radius;
```
k88hudson: This should be rounded, and only `0`  when suggestions are visible

Mardak: Oh indeed! Nice catch.

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -184,7 +184,6 @@
...
     padding-inline-start: 35px;
 
     &:focus + button {
```
Mardak: Ah doh. I thought I didn't need to change `button` to `.search-button` here as it's effectively scoped to just the actual search button, but CSS specificity calculations with the element caused styling to get messed up relative to the lower specificity of the class selector.

system-addon/content-src/components/Search/_Search.scss
Mardak: I changed this to use the 4-in-1 `border-radius` to avoid issues with accidental ltr partial border styles affecting rtl partial border styles. In this case it would have been okay as the ltr "expanded" bottom-left being 0 happens to match the rtl non-"expanded" bottom-left already being 0.

as-pine-proxy:
[Treeherder: eba6dded66c6cf247c7493c3c35e2868a22521cb](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=eba6dded66c6cf247c7493c3c35e2868a22521cb)

Issue #2600, opened 2017-05-15
Add telemetry on the "Read More" links for Pocket


emtwo:

emtwo:
This is for any clicks on the links shown in the screenshot here
<img width="783" alt="screen shot 2017-05-15 at 2 09 29 pm" src="https://cloud.githubusercontent.com/assets/784781/26072042/2c46c56e-3978-11e7-9973-2c65eaa69dcd.png">

csadilek:
Fixed with https://github.com/mozilla/activity-stream/commit/6f8f52cde358d9bae6061071f02c452d85aad3ae

PR #2629, opened 2017-05-24
feat(content): #2600 Add telemetry on the 'Read More' links for Pocket

csadilek:
After discussion, we decided the simplest way is to use default events as "impression stats" won't add much value (trending topic don't seem to change much.)


as-pine-proxy:
[Treeherder: 6a85b170a2e3ea33b70797ee679d7427d4f93082](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6a85b170a2e3ea33b70797ee679d7427d4f93082)

Issue #2507, opened 2017-04-27
Add a preferences manager to system-addon


k88hudson:
We need something to manage preferences for the system-addon that replaces the functionality of package.json. It should handle adding and removing preferences on install/uninstall.

It would be nice to be able to:

- manage the pref branch 
- set different defaults (e.g. for nightly v.s. release v.s. testing)

PR #2631, opened 2017-05-24
feat(mc): #2507 Add ActivityStreamPrefs.jsm to manage default prefs

k88hudson:
This patch adds default pref setting to Activity Stream, including for feeds. The pref configuration is stored in `ActivityStream.jsm` although theoretically this could be moved to a separate file.

The schema for prefs is pretty much the same as in `package.json`, although it does allow for different values to be set in different contexts:

```
{
  {
    name: "telemetry",
    title: "Enable system error and usage data collection",
    value: false
  },
  {
    name: "pocket",
    title: "Some other pref about pocket",
    values: {
      nightly: true, // In nightly, the default value will be true
      default: false // In other builds, the default will be false
    }
  },
}
```

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -75,7 +75,6 @@ function init(reason) {
 function uninit(reason) {
   if (activityStream) {
     activityStream.uninit(reason);
-    activityStream = null;
   }
 }
 
@@ -156,4 +155,9 @@ this.shutdown = function shutdown(data, reason) {
   modulesToUnload.forEach(Cu.unload);
 };
 
-this.uninstall = function uninstall(data, reason) {};
+this.uninstall = function uninstall(data, reason) {
+  if (activityStream) {
+    activityStream.uninstall(reason);
+    activityStream = null;
+  }
+};
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -8,6 +8,11 @@ Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 const {Store} = Cu.import("resource://activity-stream/lib/Store.jsm", {});
 const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
 
+const REASON_ADDON_UNINSTALL = 6;
+
+XPCOMUtils.defineLazyModuleGetter(this, "DefaultPrefs",
+  "resource://activity-stream/lib/ActivityStreamPrefs.jsm");
+
 // Feeds
 XPCOMUtils.defineLazyModuleGetter(this, "LocalizationFeed",
   "resource://activity-stream/lib/LocalizationFeed.jsm");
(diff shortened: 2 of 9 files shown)


system-addon/bootstrap.js
```diff
@@ -156,4 +157,11 @@ this.shutdown = function shutdown(data, reason) {
   modulesToUnload.forEach(Cu.unload);
 };
 
-this.uninstall = function uninstall(data, reason) {};
+this.uninstall = function uninstall(data, reason) {
+  // Uninstall is also called before an upgrade, so we
+  // need to make sure this is an actual uninstall
+  if (activityStream && reason === REASON_ADDON_UNINSTALL) {
+    activityStream.uninstall(reason);
```
Mardak: Any particular reason to skip upgrades (and downgrades)? The list of prefs can be reduced when installing a different version, although the impact isn't too bad as it'll just appear as if there's a default pref when it's most likely not being used.

system-addon/lib/ActivityStream.jsm
```diff
@@ -27,21 +30,74 @@ XPCOMUtils.defineLazyModuleGetter(this, "TelemetryFeed",
...
+    name: "feeds.places",
+    title: "Listens for and relays various Places-related events",
+    value: true,
+    init: () => new PlacesFeed()
+  },
+  {
+    name: "feeds.search",
+    title: "Search stuff",
+    value: true,
+    init: () => new SearchFeed()
```
Mardak: Sorry! bitrotted

system-addon/lib/ActivityStream.jsm
```diff
@@ -58,9 +114,11 @@ this.ActivityStream = class ActivityStream {
     this.options = options;
     this.store = new Store();
     this.feeds = feeds;
+    this._defaultPrefs = new DefaultPrefs(DEFAULT_PREFS);
```
Mardak: This reads kinda odd (assign default prefs with a new default prefs using default prefs), but I suppose they are indeed distinguishable, so we can just leave it. ;)

system-addon/lib/ActivityStreamPrefs.jsm
```diff
@@ -0,0 +1,58 @@
...
+    super({branch});
+    this._branchName = branch;
+  }
+  get branchName() {
+    return this._branchName;
+  }
+};
+
+this.DefaultPrefs = class DefaultPrefs {
+  constructor(config, branch = ACTIVITY_STREAM_PREF_BRANCH) {
```
Mardak: Some comments about what `config` should look like would be useful.

system-addon/lib/ActivityStreamPrefs.jsm
```diff
@@ -0,0 +1,58 @@
...
+        break;
+      case "number":
+        this.branch.setIntPref(key, val);
+        break;
+      case "string":
+        this.branch.setCharPref(key, val);
+        break;
+    }
+  }
+  init(context) {
```
Mardak: I see tests are using this context for channel, e.g., nightly, but potentially this could be anything. Are we planning on using it for just update channel?

system-addon/lib/ActivityStreamPrefs.jsm
```diff
@@ -0,0 +1,58 @@
...
+  _setDefaultPref(key, val) {
+    switch (typeof val) {
+      case "boolean":
+        this.branch.setBoolPref(key, val);
+        break;
+      case "number":
+        this.branch.setIntPref(key, val);
+        break;
+      case "string":
+        this.branch.setCharPref(key, val);
```
Mardak: This will fail in some way with non-ascii/unicode strings. Apparently there's some new fancy `setStringPref` in 55:
https://bugzilla.mozilla.org/show_bug.cgi?id=1345294
(As opposed to the old `setComplexValue`)

system-addon/lib/ActivityStream.jsm
```diff
@@ -27,21 +30,74 @@ XPCOMUtils.defineLazyModuleGetter(this, "TelemetryFeed",
...
   "resource://activity-stream/lib/TopSitesFeed.jsm");
 
-const feeds = {
-  // When you add a feed here:
-  // 1. The key in this object should directly refer to a pref, not including the
-  //    prefix (so "feeds.newtabinit" refers to the
-  //    "browser.newtabpage.activity-stream.feeds.newtabinit" pref)
-  // 2. The value should be a function that returns a feed.
+// Preferences configuration
+const DEFAULT_PREFS = [
```
Mardak: Not sure if this is the right place to list non-default add-on prefs.. although the only one should be "browser.newtabpage.activity-stream.enabled" right now? But if there are others, this list could easily diverge…

system-addon/bootstrap.js
k88hudson: The `reset` stuff removes *user* set prefs, which we don't want to do on an upgrade/downgrade – default prefs get cleared on their own on a browser shut down, since they aren't persisted anywhere

system-addon/lib/ActivityStream.jsm
k88hudson: I could changed it to `PREFS_CONFIG` or something

system-addon/bootstrap.js
k88hudson: I can add some comments about that

system-addon/lib/ActivityStreamPrefs.jsm
k88hudson: I was thinking it would be useful for release v.s. nightly etc. but I might just remove it for now (it's easy to add in later)

system-addon/lib/ActivityStreamPrefs.jsm
```diff
@@ -0,0 +1,58 @@
...
+        value = pref.values[context] || pref.values.default;
+      } else {
+        value = pref.value;
+      }
+      this._setDefaultPref(pref.name, value);
+    }
+  }
+  reset() {
+    for (const pref of this._config) {
+      this.branch.clearUserPref(pref.name);
```
Mardak: Oh hrm. Manual testing seems to show that this works on the default branch. I was confused as I thought these changes on the default branch would only affect the defaults. See idl:
"This method does nothing if this object is a default branch."
https://dxr.mozilla.org/mozilla-central/source/modules/libpref/nsIPrefBranch.idl#234-237

k88hudson: Huh... yeah, it definitely works. I will add some comments 👍

as-pine-proxy:
[Treeherder: 06dc87bc52f2f474d8470cdaab80d945dc3217d8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=06dc87bc52f2f474d8470cdaab80d945dc3217d8)

PR #2659, opened 2017-06-04
Bug 2621 bookmarks section

piatra:
Bookmarks component. Just a container over the `SpotlightItem` component + a new method in placesProvider to fetch these recent items.
I've also added a storybook component but the build task for storybook fails and I was unable to test the output.

Both `Spotlight` and `Bookmarks` component could benefit from a refactoring. Have one single component that displays card like items (currently `SpotlightItems`). `Spotlight` and `Bookmarks` would extend this component and just provide different sources for data & different event names.
Another potential improvement would be in `Feeds`. Again Bookmarks and Spotlight are similar: they fetch data from `PlacesProvider` and add metadata to the links.
I suggest deferring this in another issue as this one has become quite big.

I've tried adding a test for the new places provider method but any attempt I made seems to time out. It isn't straight forward because links returned by `.getBookmarks` have to be bookmarks that have also been visited. Here is an example of an attempt I made https://gist.github.com/piatra/de9e605daef1523c887ac625ca0cd950


addon/Feeds/BookmarksFeed.js
```diff
@@ -0,0 +1,106 @@
...
+      case am.type("APP_INIT"):
+        // When the app inititalizes refresh the data. TODO
+        this.refresh("app was initializing");
+        break;
+      case am.type("RECEIVE_BOOKMARK_ADDED"):
+        // We always want new bookmarks
+        this.refresh("a bookmark was added");
+        break;
+      case am.type("RECEIVE_PLACES_CHANGES"):
+        this.refresh("a bookmark was removed");
```
rlr: is that right? `RECEIVE_PLACES_CHANGES` action means a bookmark was removed? I grep'd and there is a `RECEIVE_BOOKMARK_REMOVED` but not sure...

addon/Feeds/BookmarksFeed.js
```diff
@@ -0,0 +1,106 @@
...
+          this.missingData = true;
+        }
+      }
+
+      return am.actions.Response("BOOKMARKS_RESPONSE", links);
+    }.bind(this));
+  }
+
+  onAction(state, action) {
+    // console.log("BOOKMARKS FEED", action.type);
```
rlr: this can be removed

addon/PlacesProvider.js
```diff
@@ -749,6 +749,53 @@ Links.prototype = {
...
+                    FROM moz_bookmarks b, moz_places p
+                    WHERE type = :type
+                    AND b.fk = p.id
+                    AND p.last_visit_date IS NOT NULL
+                    ORDER BY b.lastModified DESC
+                    LIMIT ${limit}`;
+
+    let result = yield this.executePlacesQuery(sqlQuery, {columns, params: {type: Bookmarks.TYPE_BOOKMARK}});
+
+    return result;
```
rlr: Comparing to some of the other queries, they have this after the query instead of returning the query results directly. Do we not need this?

```
links = yield this._addFavicons(links);
return this._processLinks(links);
```

content-src/components/Bookmarks/Bookmarks.js
```diff
@@ -0,0 +1,95 @@
...
+        page={this.props.page}
+        source="BOOKMARKS"
+        bestImage={getBestImage(site.images)}
+        onClick={this.onClickFactory(i, site)}
+        dispatch={this.props.dispatch}
+        {...site}
+        prefs={this.props.prefs} />
+    );
+  },
+  handleHeaderClick() {
```
rlr: Yeah, so this collapsing stuff can be removed if you use the new `<CollapsibleSection/>` instead.

rlr:
I did a first pass and things look pretty awesome. I'll do some manual testing here and then give it another pass.

addon/PlacesProvider.js
piatra: I'm fetching all the metadata for the `Bookmarks` in the [`BookmarksFeed`](https://github.com/piatra/activity-stream/blob/bug-2621-bookmarks-section/addon/Feeds/BookmarksFeed.js#L45) I assume those methods to something similar but still they don't provide the preview image. The `getCachedMetadata` seems to provide all the information needed. But I might be missing something. @ncloudioj can you give more context here?

addon/Feeds/BookmarksFeed.js
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * getData
+   *
+   * @return Promise  A promise that resolves with the "BOOKMARKS_RESPONSE" action
+   */
+  getData() {
+    return Task.spawn(function*() {
+      let links;
+      // Get links from places
+      links = yield PlacesProvider.links.getBookmarks({limit: BOOKMARKS_LENGTH});
```
rlr: after playing with it locally, i think we should request some extras for the case that the user removes a bookmark from the context menu. that way the row stays filled in with 3 if some are removed.

piatra: Good point. 👍

content-src/components/NewTabPage/NewTabPage.js
```diff
@@ -114,6 +119,13 @@ const NewTabPage = React.createClass({
               length={HIGHLIGHTS_LENGTH} sites={props.Highlights.rows}
               prefs={props.Prefs.prefs} />
           }
+          {showBookmarks &&
+            <section>
```
rlr: this wrapper `<section/>` isn't really needed. I removed it from the other components above.

rlr:
I think we should be good to go once you rebase and fix the couple of nits.

addon/Feeds/BookmarksFeed.js
piatra: This has been implemented. Updated `BOOKMARKS_LENGTH` to 6 and added a new constant `BOOKMARKS_DISPLAYED_LENGTH`

piatra:
all green 👏 @rlr please review

rlr:
please squash before merge 😄

as-pine-proxy:
[Treeherder: 21f73cad9d412d5cb4618acfedb6c71191f86915](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=21f73cad9d412d5cb4618acfedb6c71191f86915)

tspurway:
good work @piatra and @rlr  🎉

Issue #2653, opened 2017-06-01
Define "activity-stream-session" payload&schema for the system addon 


ncloudioj:
Let's use this bug to kick off the discussion about "activity-stream-session" schema&payload definition.

We can start from the current version defined [here](https://github.com/mozilla/activity-stream/blob/master/docs/v2-system-addon/data_events.md#session-end-pings)
```js
{
  "action": "activity_stream_session",
  "client_id": "26288a14-5cc4-d14f-ae0a-bb01ef45be9c",
  "addon_version": "1.0.12",
  "locale": "en-US",
  "page": ["about:newtab" | "about:home"],
  "session_duration": 4199
}
```

@dmose Could you add those timestamp fields you mentioned earlier for calculating the "session_duration"?

Comparing to the Test Pilot counterpart, I've also listed out all the missing fields as follows,
* `load_reason` and `unload_reason`
* `load_latency`
* `user_prefs`
* `history_size` and `total_bookmarks`
* `topsites` and `highlights` related stats (size/#tippytop/#screenshot)

Do we want to include any of those fields in the session pings?

@emtwo @tspurway @k88hudson

k88hudson:
`load_reason` and `unload_reason` are no longer being collected, so they can be omitted.

I think `load_latency` will be replaced with whatever values @dmose is adding for performance

k88hudson:
`history_size` and `total_bookmarks` might be useful but they haven;t been implemented yet

emtwo:
`topsites` and `highlights` metrics are also useful but likely not implemented yet.

Since there are a bunch of stuff that's not yet implemented, @ncloudioj I suppose we could either add those fields and populate them on the server or just add them as needed. What do you think is better?

ncloudioj:
@emtwo For those fields that we know we're going to include sooner or later, I'd rather have them in the database first and use default values to populate them on the server. I think this applies to `topsites` and `highlights` metrics. For `history_size` and `total_bookmarks`, we can add them later if needed.

How about the `user_prefs`? I wonder if it's too messy to decode the bitmask back to the actual pref strings in SQL. Your call :) @emtwo

dmose:
As Kate said, load_latency will go away.  Anything not otherwise noted is intended to be a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) (aka a double of milliseconds, with a .5 microsecond precision). 

[...]

PR #2660, opened 2017-06-04
Performance telemetry roadmap scratchpad

dmose:
This is a WIP so that we'll have a working document to point as we talk through #2653.

### DIFF
--- a/docs/v2-system-addon/data_events.md
+++ b/docs/v2-system-addon/data_events.md
@@ -266,8 +266,7 @@ An example might look like this:
     "toolbar_button" | // home only
     "session_restore" | // home or newtab
     "url_bar" | // home or newtab
-    "refresh" | // home or newtab
-    "other"], // home or newtab
+    "refresh"] // home or newtab
 
   // when the page itself receives an event that document.visibilityState=visible
   "visibility_event_rcvd_ts": 2,
--- a/docs/v2-system-addon/data_events.md
+++ b/docs/v2-system-addon/data_events.md
@@ -76,7 +76,7 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as 
   "event": "DELETE",
   "source": "TOP_SITES",
   "action_position": 2,
-  
+
   // Basic metadata
   "action": "activity_stream_event",
   "page": ["about:newtab" | "about:home"],
@@ -148,7 +148,7 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as 
   "event": "OPEN_NEW_WINDOW",
   "source": "TOP_SITES",
   "action_position": 2,
-  
+
   // Basic metadata
   "action": "activity_stream_event",
   "page": ["about:newtab" | "about:home"],
@@ -238,29 +238,81 @@ Here are different scenarios that cause a session end event to be sent:
 5. Refreshing
 6. Navigating to a new URL via the url bar or file menu
 
-
 ### Session performance data
-
-This data is held in a child object of the `activity_stream_session` event called `perf`.  All fields suffixed by `_ts` are type `DOMHighResTimeStamp` (aka a double of milliseconds, with a 5 microsecond precision) with 0 being the [timeOrigin](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp#The_time_origin) of the browser's hidden chrome window.
+This data is held in a child object of the `activity_stream_session` event called `perf`.  All fields suffixed by `_ts` are of type (approximately, talk to @ncloudioj for details) `DOMHighResTimeStamp` (aka a double of milliseconds, with a 5 microsecond precision) with 0 being the [timeOrigin](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp#The_time_origin) of the browser's hidden chrome window.
 
 An example might look like this:
 
-```javascript
-perf: {
+```js
+{
   // Timestamp of the action perceived by the user to trigger the load
   // of this page.
   //
   // Not required at least for error cases where the
   // observer event doesn't fire
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2658
+
   "load_trigger_ts": 1,
 
   // What was the perceived trigger of the load action:
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2658
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2685
+
   "load_trigger_type": [
     "menu_plus_or_keyboard" | // newtab only
     "unexpected" // sessions lacking actual start times
-  ],
+    "first_window_open" | // home only
+    "subsequent_window_open" | // home only
+    "toolbar_button" | // home only
+    "session_restore" | // home or newtab
+    "url_bar" | // home or newtab
+    "refresh" | // home or newtab
+    "other"], // home or newtab
 
-  // when the page itself receives an event that document.visibilityStat=visible
+  // when the page itself receives an event that document.visibilityState=visible
   "visibility_event_rcvd_ts": 2,
+
+  // As of this writing, this will be false for the first tab in every window,
+  // and true for every subsequent tab in that window.
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2539
+  "new_tab_preloaded": false,
+
+  // first thing a user might want to interact with
+  "search_box_painted_ts": 3,
+
+  // Most likely thing (based on previous telemetry) that the user is going
+  // to interact with (i.e. Hero element).
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2661
+  "top_sites_painted_ts": 5,
+
+  // When the entire page has been painted (not including stuff like screenshots
+  // showing up later).
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2662
+  "display_done_ts": 7,
+
+  // XXX below here, things are more up in the air.  We may not want to use
+  // the schema proposed below, or even implement all of the stuff to be
+  // collected here.
+
+  // See when the data was ready in case if and only if it happened
+  // after the data was required to render.
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2672
+  "top_sites_data_ready_ts": 9,
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2663
+  "search_keystroke_latency": ["TBD"],
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2664
+  "search_keystroke_autocomplete_latency": ["TBD"],
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2673
+  "slow_event_handlers": [{stack: "", approxTime: 50}],
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2526
+  "time running on main thread": "[TBD]"
 }
 ```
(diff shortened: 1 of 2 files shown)


docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,81 @@ Here are different scenarios that cause a session end event to be sent:
 3. Closing the browser
 5. Refreshing
 6. Navigating to a new URL via the url bar or file menu
+
+### Session performance data
+
+This data is held in a child object of the ```activity_stream_session``` event called ```perf```.  All fields suffixed by ```_ts``` are type ```DOMHighResTimeStamp``` (aka a double of milliseconds, with a .5 microsecond precision) with 0 being the timeOrigin of the browser's hidden
```
k88hudson: These can be `activity_stream_session`

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,81 @@ Here are different scenarios that cause a session end event to be sent:
...
+chrome window.
```
k88hudson: Is there maybe a slightly clearer way to describe "the timeOrigin of the browser's hidden chrome window"? Is timeOrigin (`timeOrigin`?) a specific thing in this context? If so would a link/small description help?

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,81 @@ Here are different scenarios that cause a session end event to be sent:
...
+
+  // XXX below here, things are more up in the air.  We may not want to use
+  // the schema proposed below, or even implement all of the stuff to be
+  // collected here.
+
+  // See when the data was ready in case if and only if it happened
+  // after the data was required to render.
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2672
+  "top_sites_data_ready_ts": 9,
```
k88hudson: What's the difference between `top_sites_painted` and `top_sites_data_ready`?

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,81 @@ Here are different scenarios that cause a session end event to be sent:
...
+  "highlights_data_ready_ts": 10,
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2663
+  "search_keystroke_latency": ["TBD"],
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2664
+  "search_keystroke_autocomplete_latency": ["TBD"],
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2673
+  "slow_event_handlers": [{stack: "", approxTime: 50}],
```
k88hudson: How is this intended to be translated into the DB?

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,81 @@ Here are different scenarios that cause a session end event to be sent:
...
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2672
+  "top_sites_data_ready_ts": 9,
+  "highlights_data_ready_ts": 10,
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2663
+  "search_keystroke_latency": ["TBD"],
+
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2664
+  "search_keystroke_autocomplete_latency": ["TBD"],
```
k88hudson: I'm not convince we need to spend time tracking this because we're using existing code paths, but I guess if it's easy it can't hurt

docs/v2-system-addon/data_events.md
dmose: Ah, good to know; fixed.

dmose: I've just hyperlinked "timeOrigin" to a more useful description.

docs/v2-system-addon/data_events.md
dmose: LInked.

docs/v2-system-addon/data_events.md
dmose: In the old add-on, top_sites_data_ready was when all the metadata from the feed was ready, so that things could be render.  If the data wasn't ready at render time, we'd show hang tight.  I thought I understood you to say that we wanted to track this as a bad state, even in the system-addon.  

top_sites_painted is after render has successfully been called, and all of the rendered content is known to have painted to the screen.

docs/v2-system-addon/data_events.md
dmose: This is really just a speculative schema.  It may want to morph some as work on #2673 happens, or for some other reason (eg we want to use something that has a more natural mapping to the DB).

docs/v2-system-addon/data_events.md
dmose: Can you link me to the existing code?  I suspect that we don't have a telemetry measurement for this latency yet, and I suspect I can figure that out by starting from the code in dxr...

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,80 @@ Here are different scenarios that cause a session end event to be sent:
 3. Closing the browser
 5. Refreshing
 6. Navigating to a new URL via the url bar or file menu
+
+### Session performance data
+
+This data is held in a child object of the `activity_stream_session` event called `perf`.  All fields suffixed by `_ts` are type `DOMHighResTimeStamp` (aka a double of milliseconds, with a .5 microsecond precision) with 0 being the [timeOrigin](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp#The_time_origin) of the browser's hidden chrome window.
```
ncloudioj: I think it should be 5 microsecond precision.

> The unit is milliseconds and should be accurate to 5 µs

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,80 @@ Here are different scenarios that cause a session end event to be sent:
...
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2658
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2539
+
+  "load_trigger_ts": 1,
+
+  // What was the perceived trigger of the load action:
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2658
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2539
+
+  "load_trigger_type": [
```
ncloudioj: We have "refresh" as one of the load triggers in T-P, do we want it here?

ncloudioj:
LGTM. Just to be clear,

* Does any field in `perf`need extra processing on the server side? Haven't seen that in the doc, or are we going to add more details for that later on?
* Are those fields in `perf` all not-null fields? If not, is it OK to populate those missing fields with default value (e.g. -1)? Having a lot of nulls in DB could hurt the query performance.

docs/v2-system-addon/data_events.md
dmose: Good catch; fixed.

docs/v2-system-addon/data_events.md
dmose: Sure, sounds worth differentiating from "other".  Fixed.

dmose:
>     Does any field in perf need extra processing on the server side? Haven't seen that in the doc, or are we going to add more details for that later on?

Since these are absolute times, and we're going to have to compute durations for things we care about, pretty much anything timestamp related is going to want math.

Is this the right place to document that stuff?  I mean, I guess at some level we really want a list of the dashboards that we want to build to work off of (I probably should have started with this).

@emtwo Would love your thoughts on this comment, as well as on the data schema strawman itself...

>     Are those fields in perf all not-null fields?

Doubtful; I think this will to some degree become more clear as we implement them one by one.
In particular, there could be errors collecting the load_trigger_ts.  I suspect we'd want to continue to send the rest of the timestamps in the those cases, in case we decided after the fact that we'd like to compare them to each other, but I'm not really sure.

Other things, like keystroke latency, presumably don't want to be sent if the user doesn't type any keystrokes.  That said, we could also just send (eg) a default value in that case, but I'm assuming having the default be on the server side is the better plan.

> If not, is it OK to populate those missing fields with default value (e.g. -1)? Having a lot of nulls in DB could hurt the query performance.

I would assume so, but I'm not sure I really have enough context to answer that.  It may only become clear perf-field as we implement.  I also suspect that this is partly a question for dashboard authors (eg @emtwo), since it effects the code they would want to write.

ncloudioj:
> Is this the right place to document that stuff? I mean, I guess at some level we really want a list of the dashboards that we want to build to work off of (I probably should have started with this).

Indeed, we should find a home for all the telemetry related details. For the Test Pilot version, there is a `data_dictionary.md` which serves as *the* reference for all the data crunching stuff on the server side. 

> Other things, like keystroke latency, presumably don't want to be sent if the user doesn't type any keystrokes. That said, we could also just send (eg) a default value in that case, but I'm assuming having the default be on the server side is the better plan.

Yup, we can definitely do this on the server side. We just need to label those required fields in the document so that ETL jobs can verify the payload accordingly. Otherwise, we treat them as optional, and will populate them with default values if missing.

emtwo:
If I understand correctly, @dmose, only raw timestamps will be collected. I think that's great because it allows us to choose any two timestamps and subtract them at analysis time. I think this subtraction would be fairly simple to do in a query. So I don't think the server needs to do this extra bit of processing.

I also think it sounds fine to me that timestamps fields default to -1 on the server side. We can filter them out in analysis.

docs/v2-system-addon/data_events.md
```diff
@@ -195,3 +195,82 @@ Here are different scenarios that cause a session end event to be sent:
...
+  // XXX below here, things are more up in the air.  We may not want to use
+  // the schema proposed below, or even implement all of the stuff to be
+  // collected here.
+
+  // See when the data was ready in case if and only if it happened
+  // after the data was required to render.
+  //
+  // TO BE IMPLEMENTED: https://github.com/mozilla/activity-stream/issues/2672
+  "top_sites_data_ready_ts": 9,
+  "highlights_data_ready_ts": 10,
```
ncloudioj: Since "highlights" won't make it to the system addon, shall we just take it out here? For other feeds (bookmarks, pocket, visit again), we can add them in later.

dmose: Good idea; removed.

dmose:
At this point, my thinking is that makes the most sense to use this PR as a way to discuss the overall schema, and not actually ever land it.  Instead, the schema in the tree would be updated as the pieces of it are landed.  The reasoning here is that a bunch of the stuff you don't necessarily discover until implementation time.

ncloudioj:
@dmose @k88hudson Have we reached some agreement on those TBD fields? If still unsure, I'd recommend not to include them for now as the first version, we can always add more fields in later.

As the new endpoint will be up and running shortly, let's get the first version of schema ready, and start collecting telemetries ASAP.

dmose:
@emtwo, yes only raw timestamps will be collected.  Note that they can't be compared across machines, or even browser restarts, but for handling individual sessions, at least those that don't spawn a browser restart, subtraction should be just fine.

Do we need to do anything special about sessions that cross browser restarts?  I'm betting we don't handle them correctly, if at all, right now...

ncloudioj:
@emtwo Do we still want to include a "session_id" field in both "session" and "event" pings to associate them with each other?

emtwo:
@dmose I believe if the browser restarts then existing sessions will end. At least that's is how it used to work in test pilot. I'd say for adding perf telemetry we probably don't need to make any modifications to that.

@ncloudioj yes. we should definitely keep `session_id` for both `session`and `event` pings because many computations join the two tables.

ncloudioj:
@emtwo Yup, I've added it to the schema. Thanks for confirming this!

k88hudson:
@dmose can we close this?

dmose:
Done.  Since our strategy has shifted, this is no longer so relevant. And the PR will still exist if/when we need to go back review earlier thinking on this.

PR #2680, opened 2017-06-07
Bug 2649 shield study

### LINKED ISSUE #2649: Revive Shield Study addon wrapper code
We need to blow the dust off of the shield study code which wraps the addon in the shield study layer for:

- starting the addon
- selecting variants (the exact variants right now are unclear, so to start just have two variants: control (tiles with monkeypatched telemetry) and activity stream
- uninstalling it after the study is complete
- monkey patching tiles for rich telemetry

The commit that has all the relevant code is here:

https://github.com/mozilla/activity-stream/commit/1258e6c9b292a0e4b22985fb5a543cf797e193b2

So this needs to be rebased onto master, updated, and then tested in the following versions of Firefox: 54, 55.

We also need to make sure that this study can survive an update from 54 to 55, so figure out if there's any way to get 54, install the shield study addon on it, then force it to update to 55 somehow, and make sure the study continues to function as expected.  

The tiles telemetry monkey patching will only work in 54 and not 55, so we need to make sure that if someone has the study addon installed in 54 and the tiles telemetry is monkey patched, and then their browser updates, that the telemetry just quietly goes away but that the study continues to exist and not crash horribly.

@sarracini is available for mentorship on this because she wrote most of this code, but her time is precious so be sparing in your requests.

piatra:
Tested in nightly. Tried testing in beta but fails due to addon signing process:
```
addons.xpi	WARN	Add-on @activity-stream-shield-study is not correctly signed.
```

**UPDATE** Works in [unbranded builds](https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds)

Fixes #2649

### DIFF
--- a/SHIELD.md
+++ b/SHIELD.md
@@ -1,0 +1,123 @@
+#Activity Stream SHIELD Study
+
+## How to run this SHIELD study
+
+tldr; Run the command ```npm run once``` and get a randomly assigned variant.
+
+The SHIELD study work currently lives on branch ```shield_study``` in this repo. To run the study: install [shield-cli](https://github.com/mozilla/shield-study-cli), checkout the branch, run command ```npm start``` in one terminal window, and in another terminal window run command ```shield run . -- -b nightly```. This will randomly pick one of the two variants provided in the study and start it in Nightly. At this point you can see which variant it chose. You can click around and print the ping payloads that get sent. If you want to see one variant specifically, run either ```shield run . Tiles -- -b nightly``` or ```shield run . ActivityStream -- -b nightly``` instead to load up the variant specified.
+
+## How it works
+
+#### Specifics of this study
+In this [SHIELD study](https://wiki.mozilla.org/Firefox/Shield/Shield_Studies) a user will be randomly assigned into either the control group or the variant group. If they are in the control group, Activity Stream will **not** be loaded into their browser, and instead they will have the original about:newtab page as their default.
+#### About the metrics
+At this point any interaction that the user makes with about:newtab will send a ping to our custom data pipeline where it will be recorded for analysis. If they are in the variant group, normal Activity Stream will be loaded and all user events will be collected and recorded as normal. The pings belonging to a SHIELD study user will contain an extra field named ```shield_variant``` which will contain either the name of the control group or the name of the variant group. This will be what we use to distinguish which group each unique user belongs to.
+
+**Sample pings:**
+```json
+{
+	"event":"CLICK",
+	"action_position":0,
+	"source":"AFFILIATE",
+	"shield_variant":"shield-study-01-Tiles",
+	"action":"activity_stream_event",
+	"tab_id":"-3-2",
+	"client_id":"ee88c4ef-4f1d-8149-9582-b727dc92e89f",
+	"addon_version":"1.1.5",
+	"locale":"en-US",
+	"page":"NEW_TAB",
+	"session_id":"{ab6f3b3e-dfc6-f944-b082-6f4bbbc21e72}"
+}
+```
+```json
+{
+	"event":"SEARCH",
+	"action":"activity_stream_event",
+	"tab_id":"-3-2",
+	"client_id":"daf5dc62-08e4-e840-837d-c2adb3b62463",
+	"addon_version":"1.1.5",
+	"locale":"en-US",
+	"page":"NEW_TAB",
+	"session_id":"{2426a185-b9bf-a84d-b359-61eb11452382}",
+	"shield_variant":"shield-study-01-Tiles"}
+```
+```json
+{
+	"tab_id":"-3-2",
+	"session_id":"{2426a185-b9bf-a84d-b359-61eb11452382}",
+	"total_history_size":100,
+	"total_bookmarks":10,
+	"load_reason":"newtab",
+	"url":"about:newtab",
+	"unload_reason":"search",
+	"client_id":"daf5dc62-08e4-e840-837d-c2adb3b62463",
+	"addon_version":"1.1.5",
+	"locale":"en-US",
+	"page":"NEW_TAB",
+	"shield_variant":"shield-study-01-Tiles",
+	"action":"activity_stream_session",
+	"session_duration":6242
+}
+```
+
+## What we hope to achieve
+
+The success criteria for the SHIELD study is as follows:
+- Increase the average number of top sites clicks per user per day
+- Increase the average number of newtab searches per user per day
+- Increase the average number of newtab sessions per user per day
+- Maintain search volume
+
+
+## Studies
+###shield-study-01
+
+Dates of study: November xx 2016 - November xx 2016
+
+Duration: 14 days
+
+Activity Stream addon version: 1.1.x
+
+Metrics measured: Clicks and blocks on top sites, search, sessions, performance related events
+
+Total number of users: 20,000
+
+Total number of users per arm: 10,000
+
+Control group: 'Tiles'
+
+Variant group: 'Activity Stream'
+
+User eligibility criteria:
+
+- Must have about:newtab enabled
+- Must have local set to en-US
+- Must not have Test Pilot add on installed
+
+Findings: *links to dashboards go here*
+
+###shield-study-02
+
+Dates of study: June 26 2017 - August xx 2017
+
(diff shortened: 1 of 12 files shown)


piatra:
In order to test you cannot use a release version of Firefox 54 because changing `xpinstall.whitelist.required` does nothing in a release version.
You need to [install an unbranded version of Firefox 54](https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds) and run shield-cli experiment with that.
I added a pref to enable more logging in case something fails but we can take that out.

addon/shield.js
```diff
@@ -0,0 +1,210 @@
...
+  _connectRetried: 0,
+  _reconnectTimeoutID: null,
+  _initializedTiles: false,
+  _doesHaveTestPilot: false,
+  _eligibilityFilters: [],
+  _placesQueries: null,
+  _originalPreloadPref: null,
+
+  setVariant(variant) {
+    this._variant = `shield-study-01-${variant}`;
```
sarracini: This should be 
```js
`shield-study-02-${variant}`
```

addon/shield.js
```diff
@@ -0,0 +1,210 @@
...
+
+  setUp: Task.async(function*() {
+    this._originalPreloadPref = prefs.get("browser.newtab.preload");
+    prefs.set("browser.newtab.preload", false);
+    PlacesProvider.links.init();
+    const clientID = yield ClientID.getClientID();
+    this._options.clientID = clientID;
+    this._options.shield_variant = this._variant;
+    this._options.telemetry = false;
+    this._options.tp_version = "1.1.7";
```
sarracini: i *think* this should reflect the version of the addon that is in test pilot right now, so i guess 1.1.12 ?
@emtwo can you remember why we had this and what this was again?

sarracini:
(Not sure if you know this but) normally our review process is:
1. set the label to "Needs Review"
2. tag a reviewer
3. assign the PR to the person you've assigned to be the reviewer
4. if the reviewer has requested changes they will assign you back to the PR with the label "Needs work"
5. once you've done the changes, repeat from step 1

Also remember to smoosh your commits and change the commit message to something more descriptive

addon/shield.js
piatra: Maybe `"version": "1.12.1",` to reflect the number in the package.json?

piatra:
@sarracini Tested multiple times in 54, no problems getting both variants + always get the survey on uninstalls.

addon/shield.js
sarracini: If yes, then yeah 1.13.0 but i want to double check with @emtwo because there's also an addon version so........

sarracini:
@piatra did you check with @ncloudioj to see if the pings were going through?

piatra:
Added some extra people for review because there are some questions to be answered.
@ncloudioj can you just confirm if there are any other pings that need to be checked
@jaredkerim is there any way to know when the experiment will run. regarding: https://github.com/mozilla/activity-stream/pull/2680#issuecomment-308065087

ncloudioj:
Hey @piatra , let's verify following user events are captured and processed properly by both activity stream and tiles variants. You can trigger those events locally in the Top sites area, and make sure they are stored in the database afterwards.
* search
* click
* dismiss
* bookmark

piatra:
There is no `DISMISS` event right? Looking at the code for the `LinkMenu` I see
```
userEvent: "BLOCK",
```
For the dismiss action. Below are the rest.


screen shot 2017-06-14 at 14 49 09
screen shot 2017-06-14 at 14 49 17
screen shot 2017-06-14 at 14 49 41

screen shot 2017-06-14 at 15 00 57

ncloudioj:
@piatra Yes, we call it `BLOCK` in the reporting. Thanks for confirming this!

LGTM! r=nanj

SHIELD.md
```diff
@@ -0,0 +1,99 @@
...
+
+User eligibility criteria:
+
+- Must have about:newtab enabled
+- Must have local set to en-US
+- Must not have Test Pilot add on installed
+
+Findings: *links to dashboards go here*
+
+###shield-study-02
```
jaredkerim: +###shield-study-02 
+Dates of study: June 26 2017 - August xx 2017
+
+Duration: 14 days
+
+Activity Stream addon version: 1.14.x
+
+Metrics measured: Clicks and blocks on top sites, search, sessions, performance related events
+
+Total number of users: 10,000
+
+Total number of users per arm: 5,000
+
+Control group: 'Tiles'
+
+Variant group: 'Activity Stream'
+
+User eligibility criteria:
+
+- Must have about:newtab enabled
+- Must have local set to en-US
+- Must not have Test Pilot add on installed
+
+Findings: *links to dashboards go here*

addon/index.js
```diff
@@ -0,0 +1,44 @@
...
+const {when: unload} = require("sdk/system/unload");
+const {Feature} = require("./shield");
+const {Cu} = require("chrome");
+const feature = new Feature();
+
+const studyConfig = {
+  name: "ACTIVITY_STREAM",
+  days: 14,
+  surveyUrls: {
+    "end-of-study": "https://qsurvey.mozilla.com/s3/Shield-Activity-Stream",
```
jaredkerim: Let's make sure this is the correct link to the new survey.

addon/shield.js
```diff
@@ -0,0 +1,210 @@
...
+
+  setUp: Task.async(function*() {
+    this._originalPreloadPref = prefs.get("browser.newtab.preload");
+    prefs.set("browser.newtab.preload", false);
+    PlacesProvider.links.init();
+    const clientID = yield ClientID.getClientID();
+    this._options.clientID = clientID;
+    this._options.shield_variant = this._variant;
+    this._options.telemetry = false;
+    this._options.tp_version = "1.13.0";
```
jaredkerim: This will be 1.14.0

dev-prefs.json
```diff
@@ -2,6 +2,8 @@
   "dom.ipc.processCount": 1,
   "javascript.options.strict": false,
   "xpinstall.signatures.required": false,
+  "extensions.logging.enabled": true,
   "extensions.@activity-streams.sdk.console.logLevel": "info",
-  "extensions.@activity-streams.telemetry.ping.endpoint": "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream"
+  "extensions.@activity-streams.telemetry.ping.endpoint": "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream",
+  "extensions.@activity-stream-shield-study.telemetry.ping.endpoint": "https://onyx_tiles.stage.mozaws.net/v3/links/activity-stream"
```
jaredkerim: @ncloudioj Is this the right endpoint to send the shield study telemetry to?

jaredkerim: @ncloudioj Oh I see this is dev prefs, for the production one do we send it to the regular production activity stream telemetry endpoint and then separate the shield data using an infernyx job?

jaredkerim: In which case we don't want a separate pref I think?

ncloudioj: Yes, it shares the same endpoint with the test pilot addon. Each payload of Shield has a "shield_variant" field to identify itself as a shield specific ping. So, we don't need a separate pref here for the endpoint.

piatra:
@jaredkerim 
screen shot 2017-06-16 at 14 17 46
Session events are coming in from all versions.

Click and block events that depend on `DirectoryLinksProvider` do not work in FFx >= 55. 
Using sdk/page-mod you can [listen for click events and it's not that difficult](https://gist.github.com/piatra/15e0b3e145ebc4a147fa679a15281d09) and blocking should be similar or even less code.

I am not sure of the reason but I'm not seeing any events from the past 2 days showing up in staging. But I know it works because I previously provided [screenshots testing for them](https://github.com/mozilla/activity-stream/pull/2680#issuecomment-308424200).
screen shot 2017-06-16 at 14 23 03

addon/index.js
jaredkerim: Correct survey link:

https://qsurvey.mozilla.com/s3/Shield-Activity-Stream-2

jaredkerim:
I am closing this PR and moving it to https://github.com/mozilla/activity-stream/pull/2747

PR #2687, opened 2017-06-08
Group duplicate bookmarks as 1 item

piatra:
When you add labels to bookmarks it saves 1 bookmark per label (in the places.db) When displaying spotlight items the key prop used in React is places.guid (which is the same for all items) so it only gets rendered once. But the .slice(0, 3) call on `sites` already happened so you end up actually showing `3 - (number of duplicates)` items.

### DIFF
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -784,6 +784,7 @@ Links.prototype = {
                     AND b.fk = p.id
                     AND p.last_visit_date IS NOT NULL
                     ORDER BY b.lastModified DESC
+                    GROUP BY p.guid
                     LIMIT ${limit}`;
 
     let links = yield this.executePlacesQuery(sqlQuery, {columns, params: {type: Bookmarks.TYPE_BOOKMARK}});
--- a/addon/PlacesProvider.js
+++ b/addon/PlacesProvider.js
@@ -783,8 +783,8 @@ Links.prototype = {
                     WHERE type = :type
                     AND b.fk = p.id
                     AND p.last_visit_date IS NOT NULL
-                    ORDER BY b.lastModified DESC
                     GROUP BY p.guid
+                    ORDER BY b.lastModified DESC
                     LIMIT ${limit}`;
 
     let links = yield this.executePlacesQuery(sqlQuery, {columns, params: {type: Bookmarks.TYPE_BOOKMARK}});


rlr:
travis: `console.log: activity-streams: [JavaScript Error: "Error(s) encountered during statement execution: near "GROUP": syntax error" {file: "resource://gre/modules/Sqlite.jsm" line: 816}]
handleCompletion@resource://gre/modules/Sqlite.jsm:816:25`

Is that the right sqlite syntax for group by?

as-pine-proxy:
[Treeherder: 3c7171295b388cb0f734d5183abd8dc29cd1b371](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=3c7171295b388cb0f734d5183abd8dc29cd1b371)

Mardak:
This fix doesn't seem right as your `GROUP BY` will not guarantee if the correct bookmark is being selected. The reason why there's duplicates is that the query includes tags instead of just plain bookmarks.

Mardak:
The short story is that this change makes it so that instead of relying on React to dedupe with the `key` attribute, it's using SQL's `GROUP BY` to dedupe. However, the items being deduped are not actually the same because the bookmark items have different `id` and `guid` values (where these "duplicates" are caused by places treating tags as bookmarks).

I believe typically the lowest bookmark id (and probably first of the duplicates) is actually the bookmark and the rest are tags, so if React happens to keep the first item, deleting this bookmark will correctly remove the user bookmark. However, sqlite's GROUP BY doesn't guarantee which duplicate will remain, and in a quick test from my profile, it happened to select the second bookmark id, which is for a tag. The reason why selecting the correct bookmark `id`/`guid` is that deleting a bookmark is by bookmark guid.

I suppose maybe a quick fix might be to only select bookmarks that have `b.title NOT NULL` as tag bookmark entries don't have a title. The more correct fix would probably be to check that the tag bookmark's parent's parent is not the "Tags" folder (which has the `tags________` `guid`). [On a separate note, should we be using the bookmark title instead of the page title?)

rlr:
Wow, super interesting. I think this might be a problem with the other queries too and would explain why sometimes "unbookmarking" doesn't work for people from activity stream.

Mardak:
Oh quite possibly. Removing a bookmark that is actually a tag would just untag that bookmark.

piatra:
@Mardak can you please validate this query fixes the problem as you described? I went with the solution of checking the parent's parent.
 I use SQL editor against `places.db` to test.
```
SELECT p.url as url,
                           p.guid as guid,
                           b.parent as parent,
                           p.title as title,
                           p.frecency as frecency,
                           p.visit_count as visitCount,
                           p.last_visit_date / 1000 as lastVisitDate,
                           b.id as bookmarkId,
                           (
                           	select title FROM moz_bookmarks
                           	where id is (
                           		select parent from moz_bookmarks
                           		where id IS (
                           			select parent from moz_bookmarks
                           			where id is b.id
                           		)
                           	)
                           ) as bookmark_type,
                           "bookmark" as type,
                           b.guid as bookmarkGuid,
                           b.title as bookmarkTitle,
                           b.lastModified / 1000 as lastModified,
                           b.dateAdded / 1000 as bookmarkDateCreated
                    FROM moz_bookmarks b, moz_places p
                    WHERE b.fk = p.id
                    AND p.last_visit_date IS NOT NULL
                    ORDER BY b.lastModified DESC
```

screen shot 2017-06-09 at 11 52 10

It should surface **bookmark_type** as empty string, folder name or *Tags* if it happens to be a tag. Then I can just modify the query to say `AND Tags NOT IN (<same query>)` to filter them out.

Mardak:
The `.getRecent` should be a handy reference:
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/Bookmarks.jsm#1492-1511

Looks like it's checking bookmark.parent.parent to not be the tags folder id as well as matching the bookmark type. I believe the type check will help make sure the parent checks don't start from folders items as opposed to bookmark items.

So it looks like your query is mostly there. Instead of modifying what's being `SELECT`ed, you can just add this to the `WHERE` clause:

`AND (SELECT guid FROM moz_bookmarks WHERE id = (SELECT parent FROM moz_bookmarks WHERE id = b.parent)) != "tags________"`

PR #2688, opened 2017-06-08
fix (sidebar): #2618 fix Done button on bottom and make contents scrollable

### LINKED ISSUE #2618: In New Tab Preferences overlay, the main body should be scrollable, and the "Done" button should be pinned
Because the new design may be arbitrarily long (have many Sections), we should always be able to dismiss the preference pane.  Let's make it so that Done button is always fixed at the bottom of the window.

rlr:
Fixes #2618

### DIFF
--- a/content-src/components/PreferencesPane/PreferencesPane.js
+++ b/content-src/components/PreferencesPane/PreferencesPane.js
@@ -48,56 +48,58 @@ const PreferencesPane = React.createClass({
         </div>
         <div className="prefs-pane">
           <div className={classNames("sidebar", {hidden: !this.state.showPane})} ref="sidebar">
-            <div className="prefs-modal-inner-wrapper">
-              <h1><FormattedMessage id="settings_pane_header" /></h1>
-              <p><FormattedMessage id="settings_pane_body" /></p>
-              <section>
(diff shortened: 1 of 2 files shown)


content-src/components/PreferencesPane/PreferencesPane.js
```diff
@@ -48,56 +48,58 @@ const PreferencesPane = React.createClass({
         </div>
         <div className="prefs-pane">
           <div className={classNames("sidebar", {hidden: !this.state.showPane})} ref="sidebar">
-            <div className="prefs-modal-inner-wrapper">
```
rlr: The only thing that really changed in this file is that I added a `<div className="sidebar-inner-wrapper">` to wrap all this content.

ncloudioj: Thanks @rlr for the comment! Github makes this change look so scary!

as-pine-proxy:
[Treeherder: e683ac5a0f7dd549d432ad6c0c8757d2c0311fb8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=e683ac5a0f7dd549d432ad6c0c8757d2c0311fb8)

Issue #2366, opened 2017-04-04
Fix pine failure browser_google_codes.js


sarracini:
https://treeherder.mozilla.org/logviewer.html#?job_id=87925091&repo=pine&lineNumber=2408

dmose:
As it turns out, this was previously seen in https://bugzilla.mozilla.org/show_bug.cgi?id=1315953, so it's actually not us.  Going to disable.

dmose:
There's some reasonable chance that this won't show up when we do things with the try servers to test landing / turning, so I don't think it's worth filing a reminder bug for re-enabling.  If it's a problem when we use the try servers before turning on by default, we'll hit it then, and can re-open this.

PR #2689, opened 2017-06-08
fix(mc): Re-enable tightly-coupled search tests and correctly use newtab as the searchPurpose

### LINKED ISSUE #2340: Re-enable/fix tightly-coupled search tests
Over in #2339, we disabled some of the browser search tests that were tightly coupled to the existing tiles implementation.

As we land search in the bootstrapped add-on (probably around the time #2314 lands), we need to figure out what to do about that.  Assuming we're planning on a gradual roll-out of A-S (meaning that the search tests need to function with A-S preffed on or off), we'd ideally run the tests once under each setting.

We might be able to just assert that pine should stay alive to keep this coverage until we're done with the rollout.

We could also write some utility for running any given test with both pref settings (assuming such a thing doesn't already exist), and wrap the tightly coupled tests with that as we turn them back on...

Mardak:
Fix #2340. Don't land this until https://bugzilla.mozilla.org/show_bug.cgi?id=1371479 lands.

First change removes the disabling of search tests except for the one removed by #2366 because we're using artifact builds on `pine`.

Second change is to correctly use "newtab" as the `searchPurpose` and "activity" as `healthReportKey`. See: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/contentSearchUI.js#32-35

r?@sarracini

### DIFF
--- a/mozilla-central-patches/disable-search-tests.diff
+++ b/mozilla-central-patches/disable-search-tests.diff
@@ -1,28 +1,8 @@
 diff --git a/browser/components/search/test/browser.ini b/browser/components/search/test/browser.ini
 --- a/browser/components/search/test/browser.ini
 +++ b/browser/components/search/test/browser.ini
-@@ -21,4 +21,6 @@ support-files =
- [browser_amazon_behavior.js]
-+skip-if = true # mozilla/activity-stream#2339
- [browser_bing.js]
- [browser_bing_behavior.js]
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: ba43dfe69907105c36c2edc3db76c5c5756e4423](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=ba43dfe69907105c36c2edc3db76c5c5756e4423)

PR #2695, opened 2017-06-09
tests(mc): #2344 Reenable tiles tests

### LINKED ISSUE #2344: decide how to handle existing tiles tests
Over in #2342, we temporarily disabled tiles tests, since it seems premature to try and figure out what to do with them until we have something stood up that is closer to replacing them.  Assuming we're doing incremental roll-out, any solution needs to ensure that both tiles and activity-stream are tested while the roll-out is going on.

k88hudson:
Fix #2344. Now that https://bugzilla.mozilla.org/show_bug.cgi?id=1369875 has landed, we should be able to reenable tiles tests

### DIFF
--- a/mozilla-central-patches/disable-tiles-tests.diff
+++ b/mozilla-central-patches/disable-tiles-tests.diff
@@ -1,7 +1,0 @@
-diff --git a/browser/base/moz.build b/browser/base/moz.build
---- a/browser/base/moz.build
-+++ b/browser/base/moz.build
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: b47a66bcddb43827fb0ec402d659c7ebd6f4bb7c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b47a66bcddb43827fb0ec402d659c7ebd6f4bb7c)

PR #2707, opened 2017-06-12
system-addon(tests): Closes #2348 Re-enable browser_UsageTelemetry_content.js

### LINKED ISSUE #2348: decide what to do with tightly-couple usaged telemetry tests
Over in #2345 on pine, we disabled a file full of tests because it depended on a working new tab page search.  Once we get the bootstrapped add-on search up and running, we should decide what to do about that so that, assuming a gradual roll-out, we keep test coverage with both tiles and Activity-Stream.

sarracini:
Change the source to be "newtab" so it can be recognized in BrowserUsageTelemetry.jsm and we get the search events properly. Also change the id on the input to "newtab-search-text" (which is what the test looks for). As a consequence this fixes the browser_UsageTelemetry_content.js test, so we can re-enable it on pine 

cc @Mardak there's no reason this *needs* to be "activity", right?

### DIFF
--- a/system-addon/content-src/components/Search/Search.jsx
+++ b/system-addon/content-src/components/Search/Search.jsx
@@ -23,8 +23,13 @@ class Search extends React.Component {
   }
   onInputMount(input) {
     if (input) {
+      // The first "newtab" parameter here is called the "healthReportKey" and needs
+      // to be "newtab" so that BrowserUsageTelemetry.jsm knows to handle events with
+      // this name, and can add the appropriate telemetry probes for search. Without the
+      // correct name, certain tests like browser_UsageTelemetry_content.js will fail (See
+      // github ticket #2348 for more details)
       this.controller = new ContentSearchUIController(input, input.parentNode,
-        "activity", "newtab");
+        "newtab", "newtab");
       addEventListener("ContentSearchClient", this);
     } else {
       this.controller = null;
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/Search/Search.jsx
```diff
@@ -24,7 +24,7 @@ class Search extends React.Component {
   onInputMount(input) {
     if (input) {
       this.controller = new ContentSearchUIController(input, input.parentNode,
-        "activity", "newtab");
+        "newtab", "newtab");
```
dmose: Please add a comment here describing what this does, along with the side-effecti-ness that we've discovered, as well as what is likely to need to be changed if someone does need to touch this in the future.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -34,11 +34,11 @@ class Search extends React.Component {
 
   render() {
```
dmose: Please add comments here about which of these attributes have magic side-effects and what those are, so that they don't get inadvertently changed in the future.

as-pine-proxy:
[Treeherder: 1a9d9647f4205eed8022774dfad87abcd81fb94a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1a9d9647f4205eed8022774dfad87abcd81fb94a)

PR #2720, opened 2017-06-14
Add bookmarks from pocket to recent bookmarks

### LINKED ISSUE #2718: Top Stories that are bookmarked are not added to Recent Bookmarks section
**[Affected versions]:**
- Firefox 53 and up
- Activity Stream Build: 1.14.0-dev-1497445690-a315655

**[Affected Platforms]:**
- All Windows
- All Linux
- All Mac

**[Prerequisites]:**
- Latest Activity Stream dev add-on version was previously installed on a profile with history.
- Top Stories are enabled.

**[Steps to reproduce]:**
2. Bookmark a Top Stories item.
3. Observe Recent Bookmarks section.
 
**[Expected results]:**
- The bookmarked Top Story item is added to Recent Bookmarks section.

**[Actual results]:**
- The item is marked as bookmarked in Top Stories section but it is not added in Recent Bookmarks.

piatra:
Explanation:
Previous version of the `getBookmarks` query filtered by `last_visit_date` IS NOT NULL. 
This excluded default bookmarks but also Pocket Stories that you would bookmark.
With this fix we try to fetch metadata information when a bookmark is added. If the page is visited it will just hit the Metadata store cache, otherwise (as for Pocket Stories) it will make a network request.
And filtering for the section happens based on metadata & images being present.
If you happen to actually visit a default bookmark page and metadata is present then yes the link will show up in the Bookmarks section.

Alternative solution included `PlacesUtils` module and added a "fake" visit in `moz_places` with current timestamp when you added a bookmark.

Fixes #2718

### DIFF
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -225,7 +225,12 @@ ActivityStreams.prototype = {
   _respondToPlacesRequests({msg, worker}) {
     switch (msg.type) {
       case am.type("NOTIFY_BOOKMARK_ADD"):
-        PlacesProvider.links.asyncAddBookmark(msg.data);
+        PlacesProvider.links.asyncAddBookmark(msg.data.url);
+
+        // Request metadata for the new bookmark if it has been added from PocketStories.
+        if (msg.data.source === "RECOMMENDED") {
+          this._pageScraper.asyncFetchLinks([{url: msg.data.url}], msg.type);
+        }
         break;
       case am.type("NOTIFY_BOOKMARK_DELETE"):
         PlacesProvider.links.asyncDeleteBookmark(msg.data);
--- a/addon/ActivityStreams.js
+++ b/addon/ActivityStreams.js
@@ -229,7 +229,7 @@ ActivityStreams.prototype = {
 
         // Request metadata for the new bookmark if it has been added from PocketStories.
         if (msg.data.source === "RECOMMENDED") {
-          this._pageScraper.asyncFetchLinks([{url: msg.data.url}], msg.type);
+          this._pageScraper.asyncFetchLinks([{url: msg.data.url}], msg.type, true);
         }
         break;
       case am.type("NOTIFY_BOOKMARK_DELETE"):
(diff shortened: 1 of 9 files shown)


piatra:
Make a request to fetch metadata when adding a new bookmark. This should cover Pocket Stories that have never been visited. This should not be a performance issue since [pageScraper does a check before sending out network requests](https://github.com/mozilla/activity-stream/blob/ac68fc52d4915c498562dec98813932336eeac34/addon/PageScraper.js#L100)

piatra:
@k88hudson, @sarracini I need an extra pair of eyes on this solution (👍 / 👎) should I go forward with this?

addon/PlacesProvider.js
```diff
@@ -782,7 +782,6 @@ Links.prototype = {
                     FROM moz_bookmarks b, moz_places p
                     WHERE type = :type
                     AND b.fk = p.id
-                    AND p.last_visit_date IS NOT NULL
```
Mardak: Removing this condition will result in `place:folder=TOOLBAR` type bookmarks to show up in the query, so those will need to be additionally handled if things aren't expecting `place:` `URI`s.

Mardak: Looks like `Bookmarks.jsm` avoids `place:` rows by using `hash()`:
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/places/Bookmarks.jsm#1503-1504

`AND url_hash NOT BETWEEN hash("place", "prefix_lo") AND hash("place", "prefix_hi")`

addon/ActivityStreams.js
```diff
@@ -226,6 +226,8 @@ ActivityStreams.prototype = {
     switch (msg.type) {
       case am.type("NOTIFY_BOOKMARK_ADD"):
         PlacesProvider.links.asyncAddBookmark(msg.data);
+        // Request metadata for the new bookmark.
+        this._pageScraper.asyncFetchLinks([{url: msg.data}], msg.type);
```
sarracini: are you able to check if the bookmark request came from a pocket story and then if it did then do this? i would rather not do a db lookup if we don't have to

piatra:
@sarracini updated with event source so that we only make requests for pocket bookmarks.

as-pine-proxy:
[Treeherder: 1f81629f542f42b622ea7b557606e77bce6ac9a7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1f81629f542f42b622ea7b557606e77bce6ac9a7)

PR #2749, opened 2017-06-22
Add Dialog component

### LINKED ISSUE #2584: Implement confirm dialog for deleting items in system add-on
Implement a dialog for asking users to confirm when items are deleted.

This should be implemented in the system add-on (there is not currently a confirmation of any kind ) and can re-use existing localized strings.

piatra:
Fixes #2584 

@aaronrbenson 
Current behavior: You click the dismiss button the dialog appears. You refresh the page, the dialog remains persists. You open a new tab, the dialog persists. Should the dialog go away after a refresh? Should it not be there on a different new tab?

@k88hudson (related to the q^): How would you implement the behavior? In the reducers I couldn't see any events that notify a new tab being opened or refresh. There was a `@@redux/INIT` but that is probably framework specific and not something we control.


aaronrbenson:
@piatra you mean you click the "Delete from History" button, right? Dismissing an item shouldn't raise the dialog.

k88hudson:
Awesome! I have a couple of meetings/things to get to first but I'll look at this by end of day 👍

k88hudson:
@piatra the event for a new tab being opened is `NEW_TAB_LOAD` (e.g. https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/NewTabInit.jsm#L17)

locales/en-US/strings.properties
```diff
@@ -36,6 +36,8 @@ menu_action_open_new_window=Open in a New Window
 menu_action_open_private_window=Open in a New Private Window
 menu_action_dismiss=Dismiss
 menu_action_delete=Delete from History
+confirm_history_delete=Are you sure you want to delete every instance of this page from your history?
```
k88hudson: I think these can be combined into one string, since they are part of the same UI message

system-addon/content-src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,84 @@
...
+ * Example:
+ * data: {
+ *   // Any sort of data needed to be passed around by actions.
+ *   payload: site.url,
+ *   // Primary button SendToMain action.
+ *   action: "DELETE_HISTORY_URL",
+ *   // Primary button USerEvent action.
+ *   userEvent: "DELETE",
+ *   // Array of locale ids to display.
+ *   message_body: ["confirm_history_delete", "confirm_history_delete_notice"],
```
k88hudson: I think you can limit this to a single id. Also to make this more clear that it's a string id and not plain text, you could name this `message_body_string_id` or something similar.

system-addon/content-src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,84 @@
...
+ *   // Text for primary button.
+ *   confirm_btn_id: "menu_action_delete"
```
k88hudson: maybe `confirm_button_string_id` would be slightly clearer

system-addon/content-src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,84 @@
...
+          <button ref="confirmButton" className="done" onClick={this._handleConfirmBtn}>
+            <FormattedMessage id={props.data.confirm_btn_id} />
+          </button>
+        </section>
+      </div>
+    </div>);
+  }
+});
+
+module.exports = connect(state => ({Dialog: state.Dialog}))(Dialog);
```
k88hudson: I think `connect(state => state.Dialog)(Dialog)` would be better in this case

system-addon/common/Reducers.jsm
```diff
@@ -95,6 +99,19 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
   }
 }
 
+function Dialog(prevState = INITIAL_STATE.Dialog, action) {
+  switch (action.type) {
+    case at.DELETE_HISTORY_URL_CONFIRM:
```
k88hudson: Isn't this action really `DIALOG_OPEN` since Dialog is a multi-purpose component?

system-addon/content-src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,84 @@
...
+ *   action: "DELETE_HISTORY_URL",
+ *   // Primary button USerEvent action.
+ *   userEvent: "DELETE",
+ *   // Array of locale ids to display.
+ *   message_body: ["confirm_history_delete", "confirm_history_delete_notice"],
+ *   // Text for primary button.
+ *   confirm_btn_id: "menu_action_delete"
+ * },
+ */
+const Dialog = React.createClass({
```
k88hudson: I think you should rename this to `ConfirmDialog` since it is specifically used for confirmations

system-addon/content-src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Confirmation Dialog component.
+ * One primary action button, one cancel button.
+ *
+ * Content displayed is controlled by `data` prop the component receives.
+ * Example:
+ * data: {
+ *   // Any sort of data needed to be passed around by actions.
+ *   payload: site.url,
+ *   // Primary button SendToMain action.
```
k88hudson: Maybe `"The action dispatched to the main process if the user clicks the confirmation button"`

system-addon/content-src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,84 @@
...
+ *   action: "DELETE_HISTORY_URL",
+ *   // Primary button USerEvent action.
```
k88hudson: Maybe `"The event name for the UserEvent telemetry action dispatched if the user clicks the confirmation button"`

k88hudson:
Oh yeah, as for not persisting the state when a refresh happens – I think the right way to do this is *not* to propagate this state to the parent reducer, and make sure it stays in the child reducer only. Let take a look to see why this isn't happening already

k88hudson:
One more thing – I think the overlay should probably have cancel the dialog if it is clicked. Right?

system-addon/content-src/components/Dialog/Dialog.jsx
piatra: @k88hudson The design has the two texts separated. As two different paragraphs. Any way to achieve this in an a11y way?
[link to design](https://mozilla.invisionapp.com/share/49C81C6Z7#/screens/238928525_Photon_-_Confirm_Dialog)

piatra:
@k88hudson I rebased on top of your pull request that refactors `LinkMenu` & addressed your comments.

locales/en-US/strings.properties
piatra: Adding this comment again because the other got collapsed by the new pull req.
@k88hudson The design has the two texts separated. As two different paragraphs. Any way to achieve this in an a11y way?
[link to design](https://mozilla.invisionapp.com/share/49C81C6Z7#/screens/238928525_Photon_-_Confirm_Dialog)

k88hudson: Ah ok, that makes sense. You could put html in the string, although that requires `dangerouslySetInnerHTML`.  @Mardak @flodolo do you have any ideas here?

k88hudson:
Ok this looks good to go, with the exception of the localization issue. I would recommend resolving that issue 

Another thing I noticed in the comps is the blur that is applied when the modal is open, but that can be filed as a follow-up.

locales/en-US/strings.properties
k88hudson: Maybe you can leave this as-is with the following changes:

- Rename them to `confirm_history_delete_notice_p1` and `confirm_history_delete_notice_p2`
- Add a comment `# LOCALIZATION NOTE: These two strings appear together in the same dialog `  or something along those lines

flodolo: I think leaving them separate as they are is good, introducing markup in strings is bad (easy to break).

This comment should work, remembering that the real problem is that localizers see only one string at a time.

```
confirm_history_delete=Are you sure you want to delete every instance of this page from your history?
# LOCALIZATION NOTE (confirm_history_delete_notice): this string is displayed in
# the same dialog as confirm_history_delete. "This action" refers to deleting a
# page from history.
confirm_history_delete_notice=This action cannot be undone.
```

piatra:
@k88hudson updated strings. I would defer the blur.

system-addon/common/Actions.jsm
```diff
@@ -28,6 +28,9 @@ for (const type of [
   "BOOKMARK_URL",
   "DELETE_BOOKMARK_BY_ID",
   "DELETE_HISTORY_URL",
+  "DELETE_HISTORY_URL_CONFIRM",
```
Mardak: This `_CONFIRM` seems to be unused?

system-addon/test/schemas/pings.js
```diff
@@ -31,6 +31,9 @@ const UserEventAction = Joi.object().keys({
       "SEARCH",
       "BLOCK",
       "DELETE",
+      "DELETE_CONFIRM",
```
Mardak: Was this supposed to be used somewhere?

piatra: The implementation of the Dialog component changed from being specific to deleting history items to a generic component, those are events that would have made sense in the first implementation. They should be removed.

PR #2755, opened 2017-06-23
Bug 2748 bookmarks

piatra:
Here are some of the issues/fixes:
- [cache key created](https://github.com/mozilla/activity-stream/blob/8b6d6a35a8644488c3e0c2b522de4e18f6a3d3f3/addon/PreviewProvider.js#L138) is different from [cache key used for lookup](https://github.com/mozilla/activity-stream/blob/8b6d6a35a8644488c3e0c2b522de4e18f6a3d3f3/addon/PageScraper.js#L107)
- [when processing links](https://github.com/mozilla/activity-stream/blob/8b6d6a35a8644488c3e0c2b522de4e18f6a3d3f3/addon/PreviewProvider.js#L178) date last modified (which was used to sort items) is not being considered. Actual date shown for bookmark is date created and that way you end up bookmarks which "seem" old.
- [forcing bookmarks to have images](https://github.com/mozilla/activity-stream/blob/8b6d6a35a8644488c3e0c2b522de4e18f6a3d3f3/addon/Feeds/BookmarksFeed.js#L67) doesn't always work to our advantage. If fetching the image preview fails then a lot of bookmarks end up not being show. Switching to filtering by metadata only until a better fix.

A subsequent pull request would look into why the image fetching fails/try to fix it for prettier bookmarks.

### DIFF
--- a/addon/Feeds/BookmarksFeed.js
+++ b/addon/Feeds/BookmarksFeed.js
@@ -19,14 +19,15 @@ module.exports = class BookmarksFeed extends Feed {
 
   _getDefaultBookmarksAge() {
     return Task.spawn(function*() {
-      const pref = parseInt(simplePrefs.prefs.defaultBookmarksAge, 10);
+      const prefValue = parseInt(simplePrefs.prefs.defaultBookmarksAge, 10);
 
-      if (pref && pref !== 0) {
-        return pref;
+      if (prefValue && prefValue !== 0) {
+        return prefValue;
       }
 
       const result = yield PlacesProvider.links.getDefaultBookmarksAge();
       simplePrefs.prefs.defaultBookmarksAge = result.toString();
+      console.log("got default age pl", result);
       return result;
     });
   }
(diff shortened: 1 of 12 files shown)


piatra:
@k88hudson could you have a look over the things described above? Not necessarily do a code review. I'm particularly interested in the cache key issue, from my investigation it looked like most lookups failed because of url mismatches (things like an extra / or having query parameters) because one was sanitized another wasn't.

k88hudson:
@piatra that sounds right to me. As far as I can remember, I think previewProvider should be looking up by cache key, not url.

piatra:
Bugs #2769 and #2760 might be related with this. AS makes additional request which can cause the bandwidth abuse and extra processing.

as-pine-proxy:
[Treeherder: 739d8e2b926fd43d849feff8095d566e0790f1e8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=739d8e2b926fd43d849feff8095d566e0790f1e8)

PR #2757, opened 2017-06-24
fix(mc): Optimize startup init by avoiding overhead and lazy loading

### LINKED ISSUE #2750: Talos sessionrestore and sessionrestore_no_auto_restore 4%+ regression
The regression seems like it can be entirely attributed to activity-stream getting initialized after "browser-delayed-startup-finished" which happens after `SessionStore.init` during "final-ui-startup" but before it reads the sessionstore file and finishes with "sessionstore-windows-restored"

Here's a profile: http://bit.ly/2stCFwh

This profile is 34ms of init after receiving the browser event. There's ~15 activity stream modules to load taking at least 1ms each, and 5 feeds to construct as well as potentially handle the `INIT` dispatch. So at quick glance, that's 15ms + 5ms + 5ms for 25ms of initing. Some modules take more time to load as they load other Firefox modules, e.g., `PlacesUtils`, or do some work at the top level (reading prefs).

### LINKED ISSUE #2759: Backport Bug 1374282 - script generated patch to remove useless bind calls, r=Mossop.
https://bugzilla.mozilla.org/show_bug.cgi?id=1374282
https://hg.mozilla.org/mozilla-central/rev/a65201e69ebf

Mardak:
Fix #2750 and fix #2759. r?@k88hudson I've pushed various separate commits that optimize in various ways. Try runs seem to confirm fixing the `sessionrestore` Talos regression.

https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=99adfb3d3f9edb5de25fb9325a25c9a149a1548c&newProject=try&newRevision=368b0dd9c3e2b68fbd6048f551a24cf53129088e&framework=1

Local profiles show it reducing the running time of `bootstrap.js`'s `observe` from 40ms to 20ms.

Here's a description of commits:

Defer loading until after sessionrestore: Looks like we can wait until later without affecting the ability to restore about:newtab or open Firefox directly to about:newtab

Convert looping to for-of: Various microbenchmarks seem to have .forEach being faster after it gets optimized over millions of runs, but profiling on the initial and small number of iterations seems to be better to avoid the callback.

Load default top sites loading on init instead of on import: This is needed to eagerly load feeds where loading at the top level assumes default prefs are already set, so loading pref on `INIT` guarantees that defaults are ready.

Lazily load sender/id from TelemetryFeed: This avoids starting up TelemetrySender (which gets and observes a number prefs) and the async await ID request until a ping needs to actually be sent.

Eager and lazy import as appropriate: Avoid the overhead of lazy getters if we're going to just load the module anyway on import/`constructor`/`INIT` and defer things that won't be used immediately.

Use prefix string check instead of regex: Avoids the regex.

Delay loading of PlacesUtils with direct services: Sometimes PlacesUtils gets called during restore, but if not, we don't want to be the one to trigger it, so grab the service directly.

Rework CONFIGs as Maps to observe whole branch instead of individual prefs: Instead of setting up observers for every single pref, get notified for anything in the branch and check if it's something we care about for `Store` and `PrefsFeed`. Also spits out `PREFS_CONFIG` and `FEEDS_CONFIG` as `Map`s to optimize some usage within each of `Store` and `PrefsFeed`.

Work around webpack/babel confusing istanbul: Not sure if you have a suggested `karma.mc.config.js` tweak to avoid this, but without, coverage thinks we're missing statements because destructured assignment results in an added helper function with unused code.

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -12,9 +12,11 @@ XPCOMUtils.defineLazyModuleGetter(this, "Preferences",
 
 XPCOMUtils.defineLazyModuleGetter(this, "Services",
   "resource://gre/modules/Services.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "setTimeout",
+  "resource://gre/modules/Timer.jsm");
 
 const ACTIVITY_STREAM_ENABLED_PREF = "browser.newtabpage.activity-stream.enabled";
-const BROWSER_READY_NOTIFICATION = "browser-delayed-startup-finished";
+const BROWSER_READY_NOTIFICATION = "sessionstore-windows-restored";
 const REASON_SHUTDOWN_ON_PREF_CHANGE = "PREF_OFF";
 const REASON_STARTUP_ON_PREF_CHANGE = "PREF_ON";
 const RESOURCE_BASE = "resource://activity-stream";
@@ -113,7 +115,8 @@ function observe(subject, topic, data) {
   switch (topic) {
     case BROWSER_READY_NOTIFICATION:
       Services.obs.removeObserver(observe, BROWSER_READY_NOTIFICATION);
-      onBrowserReady();
+      // Avoid running synchronously during this event that's used for timing
+      setTimeout(() => onBrowserReady());
       break;
   }
 }
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -4,12 +4,11 @@
 "use strict";
 
 const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
-Cu.importGlobalProperties(["fetch"]);
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.importGlobalProperties(["fetch"]);
 
 XPCOMUtils.defineLazyModuleGetter(this, "Preferences",
   "resource://gre/modules/Preferences.jsm");
-
 XPCOMUtils.defineLazyModuleGetter(this, "Services",
   "resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "setTimeout",
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -4,33 +4,20 @@
 "use strict";
 
 const {utils: Cu} = Components;
-Cu.import("resource://gre/modules/XPCOMUtils.jsm");
-const {Store} = Cu.import("resource://activity-stream/lib/Store.jsm", {});
+
+// NB: Eagerly load modules that will be loaded/constructed/initialized in the
+// common case to avoid the overhead of wrapping and detecting lazy loading.
 const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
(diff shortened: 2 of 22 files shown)


system-addon/bootstrap.js
```diff
@@ -113,7 +114,7 @@ function observe(subject, topic, data) {
   switch (topic) {
     case BROWSER_READY_NOTIFICATION:
       Services.obs.removeObserver(observe, BROWSER_READY_NOTIFICATION);
-      onBrowserReady();
+      setTimeout(() => onBrowserReady());
```
k88hudson: Can you add a comment for this use of `setTimeout`

system-addon/lib/PlacesFeed.jsm
```diff
@@ -150,8 +150,14 @@ class PlacesFeed {
   }
 
   addObservers() {
-    PlacesUtils.history.addObserver(this.historyObserver, true);
-    PlacesUtils.bookmarks.addObserver(this.bookmarksObserver, true);
+    // NB: Directly access services without loading PlacesUtils yet.
```
k88hudson: Can you just add a note about PlacesUtils being really big / slow to load

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -201,8 +210,10 @@ this.TelemetryFeed = class TelemetryFeed {
     Services.obs.removeObserver(this.browserOpenNewtabStart,
       "browser-open-newtab-start");
 
-    this.telemetrySender.uninit();
-    this.telemetrySender = null;
+    // Only uninit if the getter has initialized it
+    if (Object.prototype.hasOwnProperty.call(this, "telemetrySender")) {
```
k88hudson: I think you can use `if telemetrySender in this`, but maybe check this

system-addon/lib/ActivityStream.jsm
```diff
@@ -4,113 +4,69 @@
...
+]);
 
-const feeds = {};
-for (const pref of PREFS_CONFIG) {
-  if (pref.name.match(/^feeds\./)) {
-    feeds[pref.name] = pref.init;
-  }
+const FEEDS_CONFIG = new Map();
+for (const config of [
+  ["localization", "Initialize strings and detect locale for Activity Stream",
```
k88hudson: We should still have a way to set a default value to `false` for feeds

Mardak: Turns out webpack only expands `[,,,] = ` with the helper function but `{,,,} =` gets expanded to multiple assignments without helper! 👍

system-addon/lib/TelemetryFeed.jsm
Mardak: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#Description
"unlike the `in` operator, this method does not check down the object's prototype chain."

Maybe we were thinking of the behavior of `for … in` skipping over non-enumerable https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

system-addon/bootstrap.js
Mardak: Added

system-addon/lib/PlacesFeed.jsm
Mardak: Noted that it's **`*BIG*`**

PR #2786, opened 2017-06-28
fix(mc): Turn off processPrelaunch when preffing on activity stream

### LINKED ISSUE #2785: Flip dom.ipc.processPrelaunch.enabled to false when enabling activity-stream
Instead of backing out with #2756, we will do this instead:

flip `dom.ipc.processPrelaunch.enabled` to `false`

This will probably be done as part of https://github.com/mozilla/activity-stream/blob/master/mozilla-central-patches/enable-mc-as.diff

Mardak:
Fix #2785. r?@dmose

This will turn things on for `pine`. And when we actually turn on in mozilla-central, we can get rid of this patch (as it will be the patch being reviewed by mconley to pref on for Nightly.

### DIFF
--- a/mozilla-central-patches/enable-mc-as.diff
+++ b/mozilla-central-patches/enable-mc-as.diff
@@ -8,6 +8,13 @@ diff --git a/browser/app/profile/firefox.js b/browser/app/profile/firefox.js
 +pref("browser.newtabpage.activity-stream.enabled", true);
  
  // Enable the DOM fullscreen API.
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: 5cd620257522c3b9a15d77743cb10b40e6016ef9](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5cd620257522c3b9a15d77743cb10b40e6016ef9)

PR #2816, opened 2017-07-01
feat(mc): #2813 Add react pre-rendering

### LINKED ISSUE #2813: Implement prerendering for Activity Stream
Add prerendering to make the initial load / perceived load time of activity stream faster.

k88hudson:
Fix #2813. This patch adds react prerendering to html, in order to improve the perceived intial load time.

There are a few remaining issues:
- The prerendering assumes `en-US` strings and `ltr`
- The prerendering assumes top sites and search are pref'd on. The case where a user has either pref'd off needs to be handled.


k88hudson:
@sarracini want to take a look?

k88hudson:
@dmose, maybe a second pair of eyes wouldn't hurt?

sarracini:
I don't know if we should land this until we figure out the two issues you mentioned, they seem sort of like big deals

dmose:
It does seem hard to imagine that we could ship without those things handled, for some definition of handled.  It seems like the question of the day is whether that work should be prioritized over the other stuff in "das Boot".

For the l10n stuff, the ideal case would be have pre-rendered pages for every language that we have a locale for, and just select the one to use at package-time, install-time, or (worst-case) run-time.  Doing a simple version that just packages everything for all the locales sounds straightforward; you could ping mossop about whether he'd accept that.

Another possibly simple way forward would be to ship an English pre-rendered page, and a non-prerendered page, so at least the en-* case would get better perf.

I seem to recall us speculating about already knowing the maximum block-sizes of where the text is gonna be, so we could just pre-render without text and have text rendered in at load time.

What would handling the pref work likely involve?

k88hudson:
Closing this in favour of refactoring at a later date

PR #2824, opened 2017-07-06
feat(systemaddon): Closes #2820 Remove dependency on Preview Provider

### LINKED ISSUE #2820: Remove dependency on PreviewProvider in TopSitesFeed.jsm
We're using PreviewProvider.jsm which was a module used in remote newtab's stuff. It's failing a bunch of tests in the tree. I told them to disable them because the code isn't being used right now. So let's remove our dependency on it and then we can see on pine if it's failing tests still

sarracini:
Fix #2820. removes dependency on preview provider so that we don't have to deal with failures from: https://bugzilla.mozilla.org/show_bug.cgi?id=1343150

### DIFF
--- a/system-addon/lib/Screenshots.jsm
+++ b/system-addon/lib/Screenshots.jsm
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+this.EXPORTED_SYMBOLS = ["Screenshots"];
+
+const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "BackgroundPageThumbs",
+  "resource://gre/modules/BackgroundPageThumbs.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PageThumbs",
+  "resource://gre/modules/PageThumbs.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "FileUtils",
+    "resource://gre/modules/FileUtils.jsm");
+XPCOMUtils.defineLazyServiceGetter(this, "MIMEService",
+  "@mozilla.org/mime;1", "nsIMIMEService");
+XPCOMUtils.defineLazyModuleGetter(this, "OS",
+  "resource://gre/modules/osfile.jsm");
+
+this.Screenshots = {
+  async getScreenshotForURL(url) {
+    let screenshot = null;
+    try {
+      await BackgroundPageThumbs.captureIfMissing(url);
+      const imgPath = PageThumbs.getThumbnailPath(url);
+
+      // OS.File object used to easily read off-thread
+      const file = await OS.File.open(imgPath, {read: true, existing: true});
+
+      // nsIFile object needed for MIMEService
+      const nsFile = FileUtils.File(imgPath);
+
+      const contentType = MIMEService.getTypeFromFile(nsFile);
+      const bytes = await file.read();
+      const encodedData = btoa(String.fromCharCode.apply(null, bytes));
+      file.close();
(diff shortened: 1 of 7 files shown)


sarracini:
Either dmose or ed, whoever has time first just grab it, and remove the other reviewer. Thanks!

Mardak:
For reference, the existing `PreviewProvider.jsm`: https://dxr.mozilla.org/mozilla-central/source/browser/components/newtab/PreviewProvider.jsm

And to be clear, this change isn't directly about avoiding failures from the linked bug, but that because of the linked bug and the lack of use of the module in m-c means it's likely to be removed, so we're moving the jsm to here to avoid breakage when the existing jsm is gone.

sarracini:
@Mardak as much as I like the spread way of doing it:
``` js
function* get_chunk(data) {
  let r, g, b, a;
  while (data.length) {
    yield [r, g, b, a, ...data] = data;
  }
}
```

it gives a couple annoying lint errors:
```
Split 'let' declarations into multiple statements                            one-var
'r' is assigned a value but never used                                       no-unused-vars
Variables within the same declaration block should be sorted alphabetically  sort-vars
'g' is assigned a value but never used                                       no-unused-vars
Variables within the same declaration block should be sorted alphabetically  sort-vars
'b' is assigned a value but never used                                       no-unused-vars
Variables within the same declaration block should be sorted alphabetically  sort-vars 
'a' is assigned a value but never used                                       no-unused-vars
Assignment to function parameter 'data'                                      no-param-reassign
```

Mardak:
Oh, I was suggesting the spread operator to get rid of needing to have the named generator:
```diff
diff --git a/system-addon/test/functional/mochitest/browser_getScreenshots.js b/system-addon/test/functional/mochitest/browser_getScreenshots.js
index ebd90f0..09712ff 100644
--- a/system-addon/test/functional/mochitest/browser_getScreenshots.js
+++ b/system-addon/test/functional/mochitest/browser_getScreenshots.js
@@ -45,5 +45,10 @@ add_task(async function test_screenshot() {
   const screenshotAsDataURI = await Screenshots.getScreenshotForURL(TEST_URL);
-  const pixels = await get_pixels_for_data_uri(screenshotAsDataURI, 10, 10);
+  let pixels = await get_pixels_for_data_uri(screenshotAsDataURI, 10, 10);
   let rgbaCount = {r: 0, g: 0, b: 0, a: 0};
-  for (let [r, g, b, a] of get_chunk(pixels)) {
+
+  while (pixels.length) {
+    // break the pixels into arrays of 4 components [red, green, blue, alpha]
+    let [r, g, b, a, ...rest] = pixels;
+    pixels = rest;
+
     // count the number of each coloured pixels
```

dmose:
Yeah, just the dummy one and the skip-if line that follows it.  Sorry that the original comment in the file was confusing.

system-addon/lib/Screenshots.jsm
```diff
@@ -0,0 +1,45 @@
...
+XPCOMUtils.defineLazyModuleGetter(this, "BackgroundPageThumbs",
+  "resource://gre/modules/BackgroundPageThumbs.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PageThumbs",
+  "resource://gre/modules/PageThumbs.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "FileUtils",
+    "resource://gre/modules/FileUtils.jsm");
+XPCOMUtils.defineLazyServiceGetter(this, "MIMEService",
+  "@mozilla.org/mime;1", "nsIMIMEService");
+XPCOMUtils.defineLazyModuleGetter(this, "OS",
+  "resource://gre/modules/osfile.jsm");
```
Mardak: These currently don't need to be lazy as the module is only imported when `getScreenshotForURL` would be used anyway, but it's still good to have if this module expands and has other methods.

as-pine-proxy:
[Treeherder: 1519824594fe01a5f8cdfe9950eba679156b2299](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1519824594fe01a5f8cdfe9950eba679156b2299)

PR #2826, opened 2017-07-07
fix(system-addon): Return a topsite label for localhost. Closes #2814

### LINKED ISSUE #2814: Misalligned tile in the top sites
Firefox Nightly 56.0a1 on OS X

screen shot 2017-06-30 at 9 57 01 am

piatra:
The issue is that localhost has eTLD set to localhost. This results in `shortURL` to slice it completely returning an empty string. This change returns hostname which always exists given a valid URL. Should guard against similar scenarios with custom/special DNS entries.


as-pine-proxy:
[Treeherder: 16f2c31dc82959bd45ef456af12ce72253d5bd2b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=16f2c31dc82959bd45ef456af12ce72253d5bd2b)

PR #2828, opened 2017-07-07
feat(systemaddon): Closes #2821 Implement clear history

### LINKED ISSUE #2821: implement clear history in system-addon
oops

sarracini:
Fix #2821. let's try this again, but this time, with passing tests

### DIFF
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -52,7 +52,7 @@ this.TopSitesFeed = class TopSitesFeed {
 
     return insertPinned([...frecent, ...DEFAULT_TOP_SITES], pinned).slice(0, TOP_SITES_SHOWMORE_LENGTH);
   }
-  async refresh(action) {
+  async refresh(target = null) {
     const links = await this.getLinksWithDefaults();
 
     // First, cache existing screenshots in case we need to reuse them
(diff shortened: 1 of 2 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -76,7 +76,7 @@ this.TopSitesFeed = class TopSitesFeed {
     const newAction = {type: at.TOP_SITES_UPDATED, data: links};
 
     // Send an update to content so the preloaded tab can get the updated content
-    this.store.dispatch(ac.SendToContent(newAction, action.meta.fromTarget));
+    this.store.dispatch(ac.BroadcastToContent(newAction));
```
Mardak: This isn't quite right as we do want to send just to the single content target on NEW_TAB_LOAD but broadcast for history clear. I suppose we could probably check if action.meta.fromTarget exists and assumes if there is none, it's meant to broadcast?

sarracini: we could also just take out the getting of the links part in refresh, have a clearHistory function to call that function, and then just get that to broadcast to content?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -72,13 +72,24 @@ this.TopSitesFeed = class TopSitesFeed {
...
+  }
+  async refresh(action) {
+    let links = await this.getLinks();
     const newAction = {type: at.TOP_SITES_UPDATED, data: links};
 
     // Send an update to content so the preloaded tab can get the updated content
     this.store.dispatch(ac.SendToContent(newAction, action.meta.fromTarget));
     this.lastUpdated = Date.now();
   }
+  async clearHistory(action) {
```
Mardak: This and `refresh` are doing almost the same thing, so unless we expect the logic to diverge much, this will likely cause problems if one is updated and not the other. Also the method name seems to imply it's trying to clear history instead of handling a cleared history.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -72,13 +72,24 @@ this.TopSitesFeed = class TopSitesFeed {
...
     }
-
+    return links;
+  }
+  async refresh(action) {
+    let links = await this.getLinks();
     const newAction = {type: at.TOP_SITES_UPDATED, data: links};
 
     // Send an update to content so the preloaded tab can get the updated content
     this.store.dispatch(ac.SendToContent(newAction, action.meta.fromTarget));
```
Mardak: If the concern of checking if `fromTarget` exists, the method could probably just take an optional target instead of an `action`. If there's a target, `SendToContent` and if not, it's a `BroadcastToContent`

as-pine-proxy:
[Treeherder: 00b72fe02ce0f4274cffbdc20f5ccda589bb50e7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=00b72fe02ce0f4274cffbdc20f5ccda589bb50e7)

PR #2836, opened 2017-07-10
fix(testpilot): Fixes #2613 Typo in Top Sites Editor

### LINKED ISSUE #2613: Top Sites Editor: Change "Show less" to "Show Fewer"
screen shot 2017-05-18 at 1 38 40 pm

Change the button in the Top Sites editor to read “Show Fewer” (note the caps case because it’s a button)

AdamHillier:
Fixed in en-GB and en-US. Also capitalised `Show More` for consistency.

### DIFF
--- a/locales/en-GB/strings.properties
+++ b/locales/en-GB/strings.properties
@@ -77,8 +77,8 @@ settings_pane_done_button=Done
 # dialog.
 edit_topsites_button_text=Edit
 edit_topsites_button_label=Customise your Top Sites section
-edit_topsites_showmore_button=Show more
-edit_topsites_showless_button=Show less
+edit_topsites_showmore_button=Show More
+edit_topsites_showless_button=Show Fewer
 edit_topsites_done_button=Done
 edit_topsites_pin_button=Pin this site
 edit_topsites_edit_button=Edit this site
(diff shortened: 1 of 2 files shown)


Mardak:
If we do take this, I would think the string ids will need to change. And I'm not sure if we should directly edit `en-GB`? @flodolo?

` +edit_topsites_showless_button=Show Fewer`
My guess is this would want to be `edit_topsites_showfewer_button` even if other locales translate to the same thing if there's no distinction between countable/not.

Changing the casing of "m" to "More" would probably want to be matched in other locales that have casing.

flodolo:
> If we do take this, I would think the string ids will need to change. And I'm not sure if we should directly edit en-GB? @flodolo?

(on PTO this week)
String ID should **NOT** change, the meaning was clear enough in English to let localizers translate it properly. 

Definitely do not change IDs if you just update the case, each locale should have its own capitalization rule, no point in invalidating all existing translations
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

As for updating en-GB directly, it would be nice to avoid it and go through Pontoon (but it looks like Adam is already there).
https://pontoon.mozilla.org/en-GB/activity-stream-new-tab/

as-pine-proxy:
[Treeherder: a6469c920972649cbf090c5992bea09130f312c4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a6469c920972649cbf090c5992bea09130f312c4)

PR #2861, opened 2017-07-14
fix(mc): Slightly delay init and fix any earlier about:newtabs

### LINKED ISSUE #2819: resource://activity-stream/data/content/activity-stream.bundle.js changes restored new tab title to "undefined" when switching to another tab
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1382139
I was unable to reproduce this with mozregression nor clean profile, but this still happens even after I have disabled all add-ons, so it's likely to happen for users with used profile when Activity Stream is in Firefox.

STR ([see this video](https://github.com/mozilla/activity-stream/files/1122635/activity-stream-undefined-title.mp4.zip)):
1. Open Firefox. Set preference what to do when Firefox starts to open last tabs and windows.
2. Open about:newtab and have it focused.
3. Close Firefox.
4. Start Firefox.
5. Switch to another tab.
6. Tab title changes to "undefined".

The cause is a call from Base.componentDidMount() on visibilitychange event to updateTitle() with _ref3 variable in uninitialized state. Personally I am not sure why this happens or in which cases should be the tab title changed at all. But the fix should be easy, check the _ref3 state or the string existence before it is actually used for document.title.

### LINKED ISSUE #2856: Perma-orange: browser_startup.js annotation-service and nav-bookmarks-service before handling user events
```
FAIL | browser/base/content/test/performance/browser_startup.js | @mozilla.org/browser/annotation-service;1 is not allowed before handling user events
FAIL | browser/base/content/test/performance/browser_startup.js | @mozilla.org/browser/nav-bookmarks-service;1 is not allowed before handling user events 
```
https://treeherder.mozilla.org/logviewer.html#?job_id=114406810&repo=pine&lineNumber=2438

This change just landed in the last day:
`Bug 1380011 - Make startupRecorder.js' 'before handling user event' phase more reliable`
https://bugzilla.mozilla.org/show_bug.cgi?id=1380011
https://hg.mozilla.org/mozilla-central/rev/914e475abbf4

That commit moved "@mozilla.org/browser/annotation-service;1" and "@mozilla.org/browser/nav-bookmarks-service;1", blacklist from "before first paint" to "before handling user events"

Mardak:
Fix #2856 as well as mostly avoid to fix #2819. This delay seems to be enough to make `browser_startup.js` happy. I tested with `./mach run about:newtab` to try to get it to open as soon as possible, and there were some instances where the page stays blank now that `RemotePages` is also slightly delayed. This seems to be what's causing #2819, so we can catch any early `about:newtab`s and fix them up.

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -63,11 +63,9 @@ function init(reason) {
   }
   const options = Object.assign({}, startupData || {}, ACTIVITY_STREAM_OPTIONS);
   activityStream = new ActivityStream(options);
-  try {
-    activityStream.init(reason);
-  } catch (e) {
-    Cu.reportError(e);
-  }
+  // Slightly delay the rest of initialization from constructing to avoid
+  // loading some services too early.
+  setTimeout(() => activityStream.init(reason));
 }
 
 /**
--- a/system-addon/lib/ActivityStreamMessageChannel.jsm
+++ b/system-addon/lib/ActivityStreamMessageChannel.jsm
@@ -7,6 +7,7 @@
 const {utils: Cu} = Components;
 Cu.import("resource:///modules/AboutNewTab.jsm");
 Cu.import("resource://gre/modules/RemotePageManager.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
 
 const {actionCreators: ac, actionTypes: at, actionUtils: au} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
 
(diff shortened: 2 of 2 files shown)


Mardak:
I closed this PR because it caused a lot of test failures most likely caused by the heavy hammer of reloading any about:newtab pages: https://github.com/mozilla/activity-stream/pull/2861/files#diff-a12f94b26c6ac483de6e2a17547ced6bR145

Various tests are probably opening new tabs and forcing them to reload during a test could break tests by clearing state or triggering unexpected events, etc.

PR #2867, opened 2017-07-17
fix(systemaddon): Fallback to link title for topsite title.

### LINKED ISSUE #2857: Topsite title missing for entries in the local file system
Open a file from disk on a fresh profile to make sure it shows up in Topsites.

screen shot 2017-07-14 at 16 11 14
screen shot 2017-07-14 at 16 11 28

piatra:
screen shot 2017-07-17 at 11 17 16

@aaronrbenson do you think this is acceptable? Showing title when we can't get a URL. This happens when you open a file from your computer instead of accessing a website, there might be other scenarios but edge cases basically.

Fixes #2857

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -23,10 +23,19 @@ class TopSite extends React.Component {
       action_position: this.props.index
     }));
   }
+  _selectTileTitle(link) {
+    const title = link.pinTitle || shortURL(link);
+
+    if (title.length > 0) {
+      return title;
+    }
+
+    return link.title;
+  }
   render() {
     const {link, index, dispatch} = this.props;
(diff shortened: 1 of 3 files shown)


system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -23,10 +23,19 @@ class TopSite extends React.Component {
       action_position: this.props.index
     }));
   }
+  _selectTileTitle(link) {
+    const title = link.pinTitle || shortURL(link);
```
sarracini: can you put this functionality inside the ```shortURL``` function since we want to keep these components as simple as possible

as-pine-proxy:
[Treeherder: a18c7cd0553a9ab4c0b740f285eb43eacab6e9f1](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a18c7cd0553a9ab4c0b740f285eb43eacab6e9f1)

Issue #2870, opened 2017-07-17
Reduce code coverage for system add-on


k88hudson:
A few people have been complaining that maintaining 100% code coverage is quite cumbersome in some cases.

Some options here:

- we could reduce the min coverage threshold to a lower number (90%?)
- we could make use of istanbul-ignore statements when necessary

piatra:
I think we decided in the retrospect to go with the second option. Can this be closed?

PR #2871, opened 2017-07-17
chore(mc): #2870 Relax coverage requirements for lines/branches/statements

k88hudson:
A few people have been complaining that maintaining 100% code coverage is quite cumbersome in some cases. How do we feel about reducing our coverage of lines/branches/statements? Should we consider something else? Suggestions/opinions welcome.

@sarracini @Mardak @dmose @piatra @csadilek @AdamHillier @rlr?

### DIFF
--- a/karma.mc.config.js
+++ b/karma.mc.config.js
@@ -31,8 +31,8 @@ module.exports = function(config) {
       // This will make karma fail if coverage reporting is less than the minimums here
       check: !isTDD && {
         global: {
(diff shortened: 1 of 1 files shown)


piatra:
I would argue against it. I would prefer doing it on a case by case basis and reducing it incrementally if actually needed. Decreasing by a large margin all of a sudden would mean all pull request that follow this will have very tests.

Issue #2872, opened 2017-07-17
Pref on in Nightly


Mardak:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1381569

Mardak:
https://hg.mozilla.org/mozilla-central/rev/07a7f3038e14

PR #2892, opened 2017-07-17
chore(mc): Fix scripts to handle activity-stream enabled on nightly

### LINKED ISSUE #2891: Fix scripts to handle activity-stream enabled on nightly
No need to specially pref on with special conditions. Also backport/unpatch the stuff from https://bugzilla.mozilla.org/show_bug.cgi?id=1381569

Mardak:
Fix #2891. Can't land until https://bugzilla.mozilla.org/show_bug.cgi?id=1381569 sticks from #2872


system-addon/test/functional/mochitest/browser.ini
```diff
@@ -1,9 +1,5 @@
 [DEFAULT]
-# XXX This defaults to forcing activity-stream tests to be skipped in m-c,
-# since, as of this writing, mozilla-central itself is still turned off.
-# The tests can be run locally using 'npm run mochitest' which does various
-# overrides.
-skip-if=!activity_stream
+skip-if=!nightly_build
```
dmose: Why this?  Because we haven't been approved to uplift yet?

Whatever the reason, I think this wants a comment in the ini file about why.

Mardak: It's to match the logic of browser/extensions `moz.build` where activity-stream isn't even built not on nightly, so running tests would definitely fail. I filed #2893 to make sure we have the right settings (most likely build and test every channel but pref on by default only in nightly [or potentially early_beta_or_earlier]).

dmose: OK, r=dmose with a comment in the build.moz to that effect.

dmose: Er, moz.build

Mardak: Actually, I just tested that we don't need to duplicate the logic from `moz.build` and just remove this `skip-if=!nightly_build` line completely. As if activity-stream isn't being built, the test directory won't be processed anyway.

as-pine-proxy:
[Treeherder: a966e2fcdc1abbf359b0ffd88125b4c1301af68c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a966e2fcdc1abbf359b0ffd88125b4c1301af68c)

PR #2895, opened 2017-07-18
Query experiment definitions from experimenter fixes #2100

### LINKED ISSUE #2100: Query remote experiment definitions from Experimenter
Presently, we ship our experiment definition JSON blob inside the addon.  When the new Experimenter service is stood up, it will push experiment definitions into an S3 bucket which the client can read.  When that happens we need to modify ExperimentProvider to read from that S3 bucket rather than reading a local resource.

jaredkerim:
Experiment Provider now has two modes, one where it is fully synchronous during unit tests and receives its experiment definitions from an external caller, and the other where it asynchronously fetches experiment definitions periodically from a remote source and reinitializes itself.

### DIFF
--- a/addon/ExperimentProvider.js
+++ b/addon/ExperimentProvider.js
@@ -1,10 +1,13 @@
 const {Cu} = require("chrome");
 const prefService = require("sdk/preferences/service");
 const simplePrefs = require("sdk/simple-prefs");
-const {PrefsTarget} = require("sdk/preferences/event-target");
+const {setTimeout} = require("sdk/timers");
 const {preferencesBranch} = require("sdk/self");
 const PREF_PREFIX = `extensions.${preferencesBranch}.experiments.`;
+const EXPERIMENTS_ENDPOINT = "experiments.endpoint";
+const EXPERIMENTS_REFRESH_TIMEOUT = 60 * 60 * 1000; // Refresh once per hour
 
+Cu.importGlobalProperties(["fetch"]);
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 
 XPCOMUtils.defineLazyGetter(this, "EventEmitter", () => {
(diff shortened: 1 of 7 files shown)


addon/ExperimentProvider.js
```diff
@@ -13,76 +17,90 @@ XPCOMUtils.defineLazyGetter(this, "EventEmitter", () => {
...
         get() {
-          return prefService.get(PREF_PREFIX + experimentName);
+          return prefService.get(PREF_PREFIX + experimentId);
         },
         enumerable: true
       });
     });
   }
 
   _onPrefChange(prefName) {
```
jaredkerim: I found that taking this out caused tests to fail.

addon/TabTracker.js
```diff
@@ -114,15 +114,11 @@ TabTracker.prototype = {
     }
   },
 
-  set experimentId(experimentId) {
-    this._experimentID = experimentId || null;
-  },
-
```
jaredkerim: Ah yes good catch those need to go too.

addon/ExperimentProvider.js
jaredkerim: Nvm it was only one test, removed.

addon/TabTracker.js
```diff
@@ -114,15 +114,11 @@ TabTracker.prototype = {
...
   isActivityStreamsURL(URL) {
     return this._trackableURLs.indexOf(URL) !== -1;
   },
 
-  handleUserEvent(payload, experimentId) {
```
jaredkerim: Yep.

as-pine-proxy:
[Treeherder: 597f42bdcba94000e68785377017286ca139c74d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=597f42bdcba94000e68785377017286ca139c74d)

PR #2921, opened 2017-07-20
fix(mc): #2917 Only initialize search on focus

### LINKED ISSUE #2917: Delay loading of content search until necessary
From mconley's joy of coding 106 notes:

https://www.evernote.com/shard/s434/sh/4cc743c4-933f-4917-af1e-343acf32a096/eb0c9f60ac08f6f0
> Leaking Blob URI:
contentSearchUI probably leaks Blob URIs if the favicons for the one-off buttons are never shown (like, if the defaultEngine is set, but the user never tries to do a search in about:newtab).

k88hudson:
Fix #2917.

### DIFF
--- a/system-addon/content-src/components/Search/Search.jsx
+++ b/system-addon/content-src/components/Search/Search.jsx
@@ -8,33 +8,45 @@ const {actionCreators: ac} = require("common/Actions.jsm");
 class Search extends React.Component {
   constructor(props) {
     super(props);
+    this.controller = null;
     this.onClick = this.onClick.bind(this);
+    this.onFocus = this.onFocus.bind(this);
     this.onInputMount = this.onInputMount.bind(this);
   }
-
   handleEvent(event) {
     // Also track search events with our own telemetry
     if (event.detail.type === "Search") {
       this.props.dispatch(ac.UserEvent({event: "SEARCH"}));
     }
   }
+  onInputMount(input) {
+    this.input = input;
+  }
+  ensureSearchIsInitialized() {
+    if (this.controller) {
+      return;
+    }
+    // The first "newtab" parameter here is called the "healthReportKey" and needs
+    // to be "newtab" so that BrowserUsageTelemetry.jsm knows to handle events with
+    // this name, and can add the appropriate telemetry probes for search. Without the
+    // correct name, certain tests like browser_UsageTelemetry_content.js will fail (See
+    // github ticket #2348 for more details)
+    this.controller = new ContentSearchUIController(this.input, this.input.parentNode,
+      "newtab", "newtab");
+  }
   onClick(event) {
+    this.ensureSearchIsInitialized();
     this.controller.search(event);
   }
-  onInputMount(input) {
-    if (input) {
-      // The first "newtab" parameter here is called the "healthReportKey" and needs
-      // to be "newtab" so that BrowserUsageTelemetry.jsm knows to handle events with
-      // this name, and can add the appropriate telemetry probes for search. Without the
-      // correct name, certain tests like browser_UsageTelemetry_content.js will fail (See
-      // github ticket #2348 for more details)
-      this.controller = new ContentSearchUIController(input, input.parentNode,
-        "newtab", "newtab");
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/Search/Search.jsx
```diff
@@ -8,33 +8,43 @@ const {actionCreators: ac} = require("common/Actions.jsm");
...
-    }
+  onFocus(e) {
+    this.initializeSearch();
+  }
+  componentDidMount() {
+    // The first "newtab" parameter here is called the "healthReportKey" and needs
+    // to be "newtab" so that BrowserUsageTelemetry.jsm knows to handle events with
+    // this name, and can add the appropriate telemetry probes for search. Without the
+    // correct name, certain tests like browser_UsageTelemetry_content.js will fail (See
+    // github ticket #2348 for more details)
```
Mardak: This comment block should go next to the `new ContentSearchUIController` in `initializeSearch`

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -49,10 +59,11 @@ class Search extends React.Component {
       </label>
       <input
         id="newtab-search-text"
+        ref={input => { this.input = input; }}
```
Mardak: We don't want `bind` in `render`. Can't this be a string to access via `refs`?

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -49,10 +59,11 @@ class Search extends React.Component {
...
         maxLength="256"
         placeholder={this.props.intl.formatMessage({id: "search_web_placeholder"})}
-        ref={this.onInputMount}
         title={this.props.intl.formatMessage({id: "search_web_placeholder"})}
+        onFocus={this.controller ? null : this.onFocus}
```
Mardak: Will we ever `render` when `controller` already exists? Seems like `onFocus` will be called on focus each time anyway and there's guards to prevent reinitializing.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -8,33 +8,43 @@ const {actionCreators: ac} = require("common/Actions.jsm");
...
+    if (this.controller) {
+      return;
+    }
+    this.controller = new ContentSearchUIController(this.input, this.input.parentNode,
+      "newtab", "newtab");
+  }
   onClick(event) {
+    if (!this.controller) {
+      this.initializeSearch();
+    }
```
Mardak: I'm suspicious if this works as async messages take time to happen ? Although maybe the controller is smart enough to delay its `search`? Also, I suppose it's safe to check if `controller` exists but `initializeSearch` does it too. ;) Whatever you think is clearer.

system-addon/content-src/components/Search/Search.jsx
k88hudson: String refs are considered legacy, although i could add this function to the class definition to avoid re-binding it

system-addon/content-src/components/Search/Search.jsx
k88hudson: oh yeah good point.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -49,10 +61,11 @@ class Search extends React.Component {
       </label>
       <input
         id="newtab-search-text"
+        ref={this.onInputMount}
         maxLength="256"
         placeholder={this.props.intl.formatMessage({id: "search_web_placeholder"})}
-        ref={this.onInputMount}
         title={this.props.intl.formatMessage({id: "search_web_placeholder"})}
+        onFocus={this.onFocus}
```
Mardak: but but alphabetical! oh well… ;)

as-pine-proxy:
[Treeherder: eb86619eca3a5ccd1bfbf5dbd2977dccaa9caaa6](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=eb86619eca3a5ccd1bfbf5dbd2977dccaa9caaa6)

PR #2925, opened 2017-07-20
fix(systemaddon): Closes #2924 Add unique key for top sites

### LINKED ISSUE #2924: Top sites need a unique key
screen shot 2017-07-20 at 2 52 37 pm

here "facebook.com" was considered a top site for me, but also it was a default top site which was displayed because i blocked through my top sites but the tile never showed and that's because their key needs to be unique whereas right now the key is the url so the second facebook tile didn't get rendered

sarracini:


as-pine-proxy:
[Treeherder: 4525c6ed92013b1ab1c8002511b7b11c3d1755be](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4525c6ed92013b1ab1c8002511b7b11c3d1755be)

PR #2928, opened 2017-07-21
chore(lint): Turn on jsx-no-bind for system-addon and clean up existing code

Mardak:
As now there's more people touching system-addon code, `jsx-no-bind` should be an easy way to make sure we don't accidentally cause performance/memory regressions by having React unnecessarily re-`render`.

I ended up needing to create a `ContextMenuItem` to hold a reference to `option` so that it can be accessed from callbacks.


as-pine-proxy:
[Treeherder: d859348ff840832ff9ae22f7e027b51a4a523261](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=d859348ff840832ff9ae22f7e027b51a4a523261)

PR #2936, opened 2017-07-23
send topsites_painted_ts (Hero) telemetry (closes #2661)

### LINKED ISSUE #2661: send Hero Element (TopSites) painted telemetry
This will (approximately) be adding ```top_sites_painted_ts``` to the session telemetry.  

Note that doing it will be a bit fun.  My suspicion is that the right way to do it is to look at the technique described in http://stackoverflow.com/a/34999925, but to use two requestAnimationFrame's as suggested by https://github.com/mozilla/activity-stream/issues/2005#issuecomment-298095620 since rAF is less likely to jank than setTimeout.

dmose:
This adds telemetry to send the timestamp with the topsites element has painted to each session.  

@fmarier, if you could do a data review on the changes to data_events.md and data_dictionary.md, that would be wonderful.

@emtwo, I'd love for you to have a quick look at data_events.md, data_dictionary.md, and pings.js, that'd be much appreciated.  Separately from that, a question: before this patch, the session_start time is set shortly after the tab becomes visible.  After this patch, the session_start time is set shortly after the tab is created, which, in the common cause of a prerendered tab could be substantially (minutes, hours, ...) earlier.  Depending on what we use the start time for, maybe this doesn't matter.  On the other hand, if it does matter, assuming that the main reason we care about this is for duration (is it?), I'd propose that we compute the duration with it starting either at the load_trigger_ts or visibility_event_rcvd_ts, and drop the session_start entirely.  Whatever we do, we'll likely not be able to compare new data with old data.  What do you think?

@Mardak, would love a code review.  Reviewing the individual commits (and not squashing before landing) is probably the easiest and most useful way to go.  How to test:

```bash
$ npm run startmc
$ cd ../mozilla-central
$ ./mach build && ./mach run --jsdebugger
# open a new tab
# close the tab
# review the logged telemetry in the js console
```


system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,49 @@ class TopSite extends React.Component {
...
+      requestAnimationFrame(callback);
+    });
+  }
+  _sendPaintedEvent() {
+    this.perfSvc.mark("topsites_painted_ts");
+
+    let topsites_painted_ts = this.perfSvc
+      .getMostRecentAbsMarkStartByName("topsites_painted_ts");
+
+    this.sendAsyncMessage("ActivityStream:ContentToMain", {
```
Mardak: In this context, we should be able to use dispatch the action without needing "ContentToMain" stuff.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,49 @@ class TopSite extends React.Component {
...
+  _onNextFrame(callback) {
+    requestAnimationFrame(() => {
+      requestAnimationFrame(callback);
+    });
+  }
+  _sendPaintedEvent() {
+    this.perfSvc.mark("topsites_painted_ts");
+
+    let topsites_painted_ts = this.perfSvc
+      .getMostRecentAbsMarkStartByName("topsites_painted_ts");
```
Mardak: Why does this need to mark and get the mark in the same scope? Maybe `common/PerfService.jsm` should just have mark return the just-marked time?

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,49 @@ class TopSite extends React.Component {
...
+ * @extends {React.Component}
+ */
+class TopSitesPerfTimer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.perfSvc = this.props.perfSvc || perfSvc;
+    this.sendAsyncMessage = this.props.sendAsyncMessage || window.sendAsyncMessage;
+  }
+  componentDidMount() {
+    this._onNextFrame(this._sendPaintedEvent.bind(this));
```
Mardak: Typically the binding happens in the constructor

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,49 @@ class TopSite extends React.Component {
...
+  }
+  _onNextFrame(callback) {
+    requestAnimationFrame(() => {
+      requestAnimationFrame(callback);
```
Mardak: Maybe some explanation about why the doubling of request. Also, I'm not sure if this is actually the right time as `props.TopSites` might not actually have any data, so it might have attempted to render, but it ends up rendering nothing.

Mardak:
Also, I suppose we didn't touch any `TELEMETRY_PERFORMANCE_EVENT` items here, but we were saying those shouldn't be associated with a particular session, so don't pass in a port, i.e., no session is created?

emtwo:
> @emtwo, I'd love for you to have a quick look at data_events.md, data_dictionary.md, and pings.js, that'd be much appreciated. Separately from that, a question: before this patch, the session_start time is set shortly after the tab becomes visible. After this patch, the session_start time is set shortly after the tab is created, which, in the common cause of a prerendered tab could be substantially (minutes, hours, ...) earlier. Depending on what we use the start time for, maybe this doesn't matter. On the other hand, if it does matter, assuming that the main reason we care about this is for duration (is it?), I'd propose that we compute the duration with it starting either at the load_trigger_ts or visibility_event_rcvd_ts, and drop the session_start entirely. Whatever we do, we'll likely not be able to compare new data with old data. What do you think?

I've looked at data_events.md, data_dictionary.md, and pings.js and they seem good to me. You're right that session_start time is used for computing session duration and so it doesn't make sense to compute duration starting from when a tab is created if it could be much earlier. I'd personally choose using `visibility_event_rcvd_ts` as the timestamp to replace `session_start` since session duration is about how much time a user spends on the page once it becomes visible.

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: Fixed.

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: OK, I've added the comment about the doubling.  I see your point about TopSites.  Hmmm...

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: That sounds like a good idea.  I'll add an XXX comment for now.

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: OK, I've ditched this.

dmose: (In favor of dispatching an action)

dmose:
Three more things to do: 

- [ ] fix up start_session stuff as @emtwo suggests
- [ ] document exactly what we're timing, possibly changing the timestamp name too
- [ ] file a spin-off issue to measure hero element after screenshots are all replaced

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,93 @@ class TopSite extends React.Component {
...
+ *
+ * @class TopSitesPerfTimer
+ * @extends {React.Component}
+ */
+class TopSitesPerfTimer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.perfSvc = this.props.perfSvc || perfSvc;
+
+    this._maybeSendPaintedEvent = this._maybeSendPaintedEvent.bind(this);
```
dmose: Good catch; removed.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,93 @@ class TopSite extends React.Component {
...
+    this._maybeSendPaintedEvent();
+  }
+
+  componentDidUpdate() {
+    this._maybeSendPaintedEvent();
+  }
+
+  /**
+   * Call the given callback when the next animation frame (not the upcoming
+   * one) paints.
```
dmose: > I suppose the confusion comes from the requestAnimationFrame is at the beginning of the frame before it actually happens.

This is indeed exactly the problem.  I've changed the verbiage to include the word subsequent, which maybe helps a bit.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,93 @@ class TopSite extends React.Component {
...
+    // If we've already saved a timestamp for this session, don't do so again.
+    if (this._timestampSent) {
+      return;
+    }
+
+    // We don't want this to ever happen, but sometimes it does.  And when it
+    // does (typically on the first newtab at startup time calling
+    // componentDidMount), the paint(s) we care about will be later (eg
+    // in a subsequent componentDidUpdate).
+    if (!this.props.TopSites.initialized) {
```
dmose: No, because that doesn't mean that the remained of the TopSites have mounted, I don't think.  Even if it does, given our current timelines, I don't think it's worth the time to move things around, because it doesn't really make a big difference in code readability/maintainability.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -80,6 +168,7 @@ const TopSites = props => (<section>
   </ul>
 </section>);
 
-module.exports = connect(state => ({TopSites: state.TopSites}))(TopSites);
+module.exports = connect(state => ({TopSites: state.TopSites}))(TopSitesPerfTimer);
 module.exports._unconnected = TopSites;
```
dmose: I'm confused.  _unconnected is already set to just TopSites.  What would need to be changed?

system-addon/content-src/components/TopSites/TopSites.jsx
Mardak: I suppose if the double requestAnimationFrame is consistent, then it'll be more useful in tracking relative changes and the actual reported value can always be adjusted for whatever the delay between a pair of calls would be.

system-addon/content-src/components/TopSites/TopSites.jsx
Mardak: The connected component is `TopSitesPerfTimer`, so the unconnected one would seem to imply it's the `TopSitesPerfTimer`

dmose: I see what you mean; I've pushed a patch to clarify.

dmose:
@ncloudioj and @emtwo, I'd love your thoughts on my [commit that eliminates `session_start` and redefines `session_duration`](https://github.com/mozilla/activity-stream/pull/2936/commits/c151c9d6f2605c58b92955c4039a9e225e422fb0), both how it's computed, whether it's required, and what type it is.  This will mean a couple of server-side changes, but it makes things more consistent.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -119,7 +118,10 @@ this.TelemetryFeed = class TelemetryFeed {
       return;
     }
 
-    session.session_duration = Math.round(Components.utils.now() - session.start_time);
+    if (session.perf.visibility_event_rcvd_ts) {
+      session.session_duration = perfService.absNow() - session.perf.visibility_event_rcvd_ts;
```
ncloudioj: Is it possible to stick with the millisecond precision for `session_duration`? Unfortunately, our data warehouse doesn't support altering the column type (int -> double) on the fly. To do that regardless would need a series of table migrations.

@emtwo Could you also confirm this?

Making `session_duration` as an optional field is fine, we can handle it in the ETL job.

dmose: @ncloudioj Sure, that's easy enough to do.  Will push an update with that change momentarily.

dmose:
Just chatted with @emtwo, and she agrees that we should keep session_duration as an int.  She has also suggested topsites_first_painted_ts is a good name for thing we're measuring.  I'll make both those changes shortly.

system-addon/lib/TelemetryFeed.jsm
ncloudioj: Thanks for confirming this! @dmose 

r=nanj

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,92 @@ class TopSite extends React.Component {
...
+ * XXX Should probably be split out into a higher-order component to wrap
+ * whatever.
+ *
+ * @class TopSitesPerfTimer
+ * @extends {React.Component}
+ */
+class TopSitesPerfTimer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.perfSvc = this.props.perfSvc || perfSvc;
```
k88hudson: any reason you want to have `this.perfSvc` v.s. `this.props.perfSvc`? This seems a little strange in that if props change, it won't be updated

k88hudson: if it's the case that you never expect that prop to update, it's still potentially bug-causing and i'd recommend at least a comment

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -69,6 +70,92 @@ class TopSite extends React.Component {
...
+
+    this.props.dispatch(ac.SendToMain({
+      type: at.SAVE_SESSION_PERF_DATA,
+      data: {topsites_first_painted_ts}
+    }));
+
+    this._timestampSent = true;
+  }
+  render() {
+    return (<TopSites {...this.props} />);
```
k88hudson: Personally I would probably do

```js
<PerfTimer>
   <TopSites />
</PerfTimer>
```

and render `this.props.children` instead of nesting but this works too.

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: @k88hudson perfSvc is already required at the top of the file; putting it into options is really just for dependency injection testing.  In production, that should never be used.  I'll add a comment.

Mardak: Isn't that the purpose of the `injector`? You can replace the require with custom functionality for testing.

dmose: Yes.  I find injector more awkward to work with, and we're already using the "options"
 style of dependency injection in at least a few other places in the tree.  I did just try to do a quick and dirty switch to injector, and it got messy, consistent with my previous biases, perhaps.  In any case, I'm just going to leave a comment for now.

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: Agreed, that would be cleaner.  I've updated the comment about higher-order component work to suggest this as a possible alternative.

system-addon/content-src/components/TopSites/TopSites.jsx
Mardak: Where are these? A quick grep seems to only have detect-user-session-start, which isn't a component?

dmose: PerfService uses this style two, and I think I reviewed something a little while ago from @k88hudson that did, but I can't recall what it was.  However, if you're speaking specifically of React components, you might well be right, that this is the first case of using that in the tree.

PR #2946, opened 2017-07-24
fix(systemaddon): All links blocked by NewTabUtils. Closes #2941

### LINKED ISSUE #2941: All Pocket links got blocked by "NewTabUtils.blockList.isBlocked"
Nightly 56.0a1 (2017-07-24 64bit) on OS X

screen shot 2017-07-24 at 1 13 58 pm

For some reason, the Pocket recommended articles didn't work for my A-S. After poking around it with @csadilek, we found that all the Pocket links got blocked by `NewTabUtils.blockList.isBlocked`. After another closer look, we noticed that this particular link `"1B2M2Y8AsgTpgAmY7PhCfg==":1` in my `browser.newtabpage.blocked` made `NewTabUtils.blockList.isBlocked` mistakenly report every link from Pocket as blocked.

This sounds more likely a `NewTabUtils` bug other than the A-S one, tracking it here for now as it affected A-S.

csadilek:
The "offending" hash value is caused by blocking undefined. So, at some point the call to ```blockURL``` was implemented incorrectly, passing a string instead of ```{url: value}```. This caused the hash value in the ```browser.newtabpage.blocked``` list. This is no longer a problem in Activity Stream.

Now, I have made a similar mistake when calling ```isBlocked``` passing the url instead of ```{url: value}```. If now undefined was previously blocked, all URLs will appear blocked.

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -78,7 +78,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
         .then(body => {
           let items = JSON.parse(body).list;
           items = items
-            .filter(s => !NewTabUtils.blockedLinks.isBlocked(s.dedupe_url))
+            .filter(s => !NewTabUtils.blockedLinks.isBlocked({"url": s.dedupe_url}))
             .map(s => ({
               "guid": s.id,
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 10187fd8cd0ed8a765dab706d46cd8da9ff7bba0](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=10187fd8cd0ed8a765dab706d46cd8da9ff7bba0)

PR #2949, opened 2017-07-24
fix(l10n): Add better localization notes for section strings

### LINKED ISSUE #2934: Address l10n review comments
https://github.com/mozilla/activity-stream-l10n/pull/4

@AdamHillier the localization note for `section_info_option` needs some more context to be useful

@csadilek is `empty_state_topstories` meant to be generic for top stories because it includes "Pocket" in the string. Or should it use `{provider}` to insert Pocket? A localization note would be useful too. A simple fix might be to just rename it `pocket_empty_state` if we expect other top stories to have different empty messages.

Mardak:
Fix #2934. r?@sarracini Replaced the hardcoded "Pocket" with `{provider}` and added notes.

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -69,8 +69,8 @@ search_header={search_engine_name} Search
 search_web_placeholder=Search the Web
 search_settings=Change Search Settings
 
-# LOCALIZATION NOTE (section_info_option): This is the screenreader only text
-# for the section helper text.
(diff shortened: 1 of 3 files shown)


as-pine-proxy:
[Treeherder: e897e0e4c7b600f247bd191d1dded8537834486f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=e897e0e4c7b600f247bd191d1dded8537834486f)

PR #2953, opened 2017-07-24
fix (systemaddon): #2938 add fallback to shortUrl() for missing title

### LINKED ISSUE #2938: short-url.js needs extra fallback in case of missing title
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1383859
In case link title is empty [short url](https://github.com/mozilla/activity-stream/blob/75e5bec66d83924b3979925bc1e1fc4cee559999/system-addon/content-src/lib/short-url.js#L26) will return null and crash Activity Stream. This only happens for local uris, depends on [the answer in this bug](https://github.com/mozilla/activity-stream/issues/2937#issuecomment-317418324) if we actually need to fix/handle this. An extra `|| link.url` should fix it.

screen shot 2017-07-24 at 15 14 59

rlr:
r? @Mardak 

Fixes #2938

### DIFF
--- a/system-addon/content-src/lib/short-url.js
+++ b/system-addon/content-src/lib/short-url.js
@@ -10,6 +10,7 @@
  *               via Services.eTLD.getPublicSuffix
  *         {str} link.hostname (optional) - The hostname of the url
  *               e.g. for http://www.hello.com/foo/bar, the hostname would be "www.hello.com"
+ *         {str} link.title (optional) - The title of the link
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 3c9827cb21a80c78e49d1af8de389205d332a182](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=3c9827cb21a80c78e49d1af8de389205d332a182)

PR #2959, opened 2017-07-25
feat(systemaddon): Add HTTP Referrer to Top Stories

csadilek:
Adding the referrer header when clicking on a card, as well as "open new window" and "open new private window" of the card's context menu.

Moved the handling of OPEN_NEW_WINDOW to the PlacesFeed as it's used by both TopSites and TopStories.


system-addon/content-src/components/Card/Card.jsx
```diff
@@ -27,6 +29,10 @@ class Card extends React.Component {
       showContextMenu: true
     });
   }
+  onClick(event) {
```
Mardak: nit: Probably something like `onLinkClick` as there's multiple "click" events going on here. See https://github.com/mozilla/activity-stream/blob/f4edbe410dd2b57ac19d337ae080528c25db7649/system-addon/content-src/components/TopSites/TopSites.jsx#L25

system-addon/lib/PlacesFeed.jsm
```diff
@@ -207,9 +215,23 @@ class PlacesFeed {
       case at.DELETE_HISTORY_URL:
         NewTabUtils.activityStreamLinks.deleteHistoryEntry(action.data);
         break;
+      case at.OPEN_NEW_WINDOW:
```
Mardak: I agree that `TopSitesFeed` was an odd place to have had this, but I'm not sure `PlacesFeed` is that much more appropriate… but I suppose that's @k88hudson's call to make!

system-addon/content-src/components/Card/Card.jsx
k88hudson: yeah I agree, that would be clearer. Also can you add a comment somewhere around here about why this exists? (i.e. that we need to override the regular behaviour because of HTTP referrer stuff)

system-addon/lib/PlacesFeed.jsm
k88hudson: I might be inclined to suggest creating a new `WindowFeed` or something like that... but because there is a lot of overhead to creating new jsms, maybe just leave this for now.

system-addon/content-src/components/Card/Card.jsx
csadilek: OK, updated.

as-pine-proxy:
[Treeherder: 7033a6179fa0b54d4e18df240bc8e3109ebe349b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7033a6179fa0b54d4e18df240bc8e3109ebe349b)

PR #2960, opened 2017-07-25
feat(systemaddon): Closes #2943 Pocket on by default for en-{US,CA}

### LINKED ISSUE #2943: Turn pocket on for en-US/en-CA by default in nightly
we've got the go ahead for en-US and en-CA to turn top stories on by default in nightly. we should do this asap so we get as much testing in as possible to ensure an uplift into release 56

sarracini:
Fix #2943. Turns on top stories by default for locales en-US and en-CA

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -4,6 +4,7 @@
 "use strict";
 
 const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/Services.jsm");
 
 // NB: Eagerly load modules that will be loaded/constructed/initialized in the
 // common case to avoid the overhead of wrapping and detecting lazy loading.
@@ -28,7 +29,8 @@ const SECTIONS = new Map([
   ["topstories", {
     feed: TopStoriesFeed,
     prefTitle: "Fetches content recommendations from a configurable content provider",
-    showByDefault: false
+     // for now, we only want to show top stories by default to the following locales
+    showByDefault: ["en-US", "en-CA"].includes(Services.locale.getRequestedLocale())
   }]
 ]);
 


Mardak:
Hmmm....... I'm not sure what's the right way this should be structured, but it seems like `LocalizationFeed` should be in charge of getting locales and detecting changes, which it has been doing, instead of getting a locale passed in.

Should then some other feed.. `PrefsFeed` be watching for `at.LOCALE_UPDATED` to change some default settings of `TopStoriesFeed`? That doesn't really seem right either.

@k88hudson any suggestions?

system-addon/lib/ActivityStream.jsm
```diff
@@ -22,13 +23,16 @@ const {TopSitesFeed} = Cu.import("resource://activity-stream/lib/TopSitesFeed.js
...
+// for now, we only want to show top stories by default to these locale
+const ACCEPTED_POCKET_LOCALES = ["en-US", "en-CA"];
 
 // Sections, keyed by section id
 const SECTIONS = new Map([
   ["topstories", {
     feed: TopStoriesFeed,
     prefTitle: "Fetches content recommendations from a configurable content provider",
-    showByDefault: false
+    showByDefault: ACCEPTED_POCKET_LOCALES.includes(LOCALE) || false
```
Mardak: No need for `|| false` as it would only use the `false` if `includes` returned `false`

Mardak: Actually, just inline all of `LOCALE` and `ACCEPTED_POCKET_LOCALES` to right here to be clear this is special temporary logic for preffing on/off top stories.

system-addon/lib/LocalizationFeed.jsm
```diff
@@ -17,6 +17,9 @@ const LOCALES_CHANGE_TOPIC = "intl:requested-locales-changed";
 const LOCALES_FILE = "resource://activity-stream/data/locales.json";
 
 this.LocalizationFeed = class LocalizationFeed {
+  constructor(options = {}) {
+    this._locale = options.locale;
```
Mardak: Let's skip the passing in locale

Mardak:
Just to summarize the discussion from IRC:

There's a lack of DRY-ness (Don't Repeat Yourself) by additionally getting the locale from outside `LocalizationFeed`. Ideally `ActivityStream.jsm` or `DefaultPrefs` has some way to watch for `LOCALE_UPDATED` to change the default of `feeds.section.topstories` but those currently don't participate in observing actions. `ActivityStream` does have a reference to a `Store`.

Given that this should be relatively short-lived in that Pocket will be enabled for all locales (?), we'll just grab the locale at the top level of `ActivityStream.jsm` when it loads.

sarracini:
Additionally, if the logic of showing sections by default, or updating them based on some event becomes more complicated we should think about a way to handle this kind of case gracefully. This is just one example of a configurable section, but in the future I can see this becoming something we need to handle

as-pine-proxy:
[Treeherder: d9f70e19687d2710a2ae9c02b0551fb5835b8fe7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=d9f70e19687d2710a2ae9c02b0551fb5835b8fe7)

jaredkerim:
I'm not sure if this helps, but Shield supports localization targetting for pref studies, so I think we could accomplish all of this by exposing two prefs, one to enable Activity Stream, and one to enable Pocket, and, then just tell Shield that we want to study Activity Stream + Pocket in locales A, B, C, and Activity Stream without Pocket in locales D, E, F without having to write any locale targetting code directly in Activity Stream?

jaredkerim:
Okay so I think the solution we discussed is that we will do the Pocket activation targeting inside the system addon, and then we will selectively roll out to locales using Shield, excluding ones we do want to use pocket with until their locale specific pocket feed is ready, and then with Shield deploy to those locales.

PR #2987, opened 2017-07-27
fix(systemaddon): Dedupe Topsites

### LINKED ISSUE #2933: Dup tiles in the Top sites
Nightly 56.0a1 (Linux)

screenshot from 2017-07-21 14-40-00

We fixed the same issue in the Test Pilot version by deduping the default top sites against the user's top sites.

piatra:
Fixes #2933 
Dedupe happens in the Feeds. I've added the hostname key to the site objects because it is displayed in the cards. Previously the value returned by `shortURL` was named title (especially in the tests) it but made it confusing since websites already have a title.

### DIFF
--- a/system-addon/common/Dedupe.jsm
+++ b/system-addon/common/Dedupe.jsm
@@ -12,6 +12,11 @@ this.Dedupe = class Dedupe {
     return false;
   }
 
+  /**
+   * Removes duplicates in a list based on createKey.
+   * @param {Array} values
+   * @returns {Array}
+   */
   collection(values) {
     const valueMap = new Map();
     values.forEach(value => {
--- a/system-addon/common/Dedupe.jsm
+++ b/system-addon/common/Dedupe.jsm
@@ -1,0 +1,27 @@
+this.Dedupe = class Dedupe {
+  constructor(createKey, compare) {
+    this.createKey = createKey || this.defaultCreateKey;
+    this.compare = compare || this.defaultCompare;
+  }
+
+  defaultCreateKey(item) {
+    return item;
+  }
+
+  defaultCompare() {
+    return false;
+  }
+
+  one(values) {
+    const valueMap = new Map();
+    values.forEach(value => {
+      const key = this.createKey(value);
+      if (!valueMap.has(key) || this.compare(valueMap.get(key), value)) {
+        valueMap.set(key, value);
+      }
+    });
+    return Array.from(valueMap.values());
+  }
+};
+
+this.EXPORTED_SYMBOLS = ["Dedupe"];
--- a/system-addon/common/Dedupe.jsm
+++ b/system-addon/common/Dedupe.jsm
@@ -22,17 +22,17 @@ this.Dedupe = class Dedupe {
   group(groups) {
     const globalKeys = new Set();
     const result = [];
-    groups.forEach(values => {
+    for (const values of groups) {
       const valueMap = new Map();
-      values.forEach(value => {
+      for (const value of values) {
         const key = this.createKey(value);
         if (!globalKeys.has(key) && (!valueMap.has(key) || this.compare(valueMap.get(key), value))) {
           valueMap.set(key, value);
         }
-      });
+      }
       result.push(valueMap);
       valueMap.forEach((value, key) => globalKeys.add(key));
-    });
+    }
     return result.map(m => Array.from(m.values()));
   }
 };
--- a/system-addon/common/Dedupe.jsm
+++ b/system-addon/common/Dedupe.jsm
@@ -13,19 +13,27 @@ this.Dedupe = class Dedupe {
   }
 
   /**
-   * Removes duplicates in a list based on createKey.
-   * @param {Array} values
+   * Dedupe an array containing groups of elements.
+   * Duplicate removal favors earlier groups.
+   *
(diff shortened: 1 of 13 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -55,7 +61,12 @@ this.TopSitesFeed = class TopSitesFeed {
       frecent = frecent.filter(link => link && link.type !== "affiliate");
     }
 
-    return insertPinned([...frecent, ...DEFAULT_TOP_SITES], pinned).slice(0, TOP_SITES_SHOWMORE_LENGTH);
+    pinned = insertPinned([...frecent, ...DEFAULT_TOP_SITES], pinned);
+
+    // Parse site url and extract hostname.
+    pinned.forEach(site => { site.hostname = shortURL(site); });
+
+    return this.dedupe.collection(pinned).slice(0, TOP_SITES_SHOWMORE_LENGTH);
```
k88hudson: Would this always favour the pinned item? Or is there a possibility it would get removed?

system-addon/common/Dedupe.jsm
```diff
@@ -0,0 +1,32 @@
...
+  defaultCreateKey(item) {
+    return item;
+  }
+
+  defaultCompare() {
+    return false;
+  }
+
+  /**
+   * Removes duplicates in a list based on createKey.
```
Mardak: This should probably say it picks the first item that it finds. So re: @k88hudson's question about pinned, it will remove the pinned if a non-pinned duplicate key item appeared first.

Mardak: Actually, nevermind. it depends on the compare.

system-addon/lib/TopSitesFeed.jsm
Mardak: Currently no, but Dedupe can be called with a compare function that prefers pinned items.

piatra:
@k88hudson If I'm [reading this right](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/TopSitesFeed.jsm#L118-L121) we are already calling `this.refresh` every time until the user has 12 non default topsites (so a lot for new profiles). Is there any reason we don't just have the 15 minutes rule?

piatra:
We could also [after getting this result](https://github.com/mozilla/activity-stream/blob/e5ec32e35cf686b178f49b78eec19ebe545d2a9c/system-addon/lib/TopSitesFeed.jsm#L58) compare the length of the result with the length of `this.store.getState().TopSites.rows` if it's the same we can return early.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -22,6 +24,13 @@ const DEFAULT_TOP_SITES = [];
 this.TopSitesFeed = class TopSitesFeed {
   constructor() {
     this.lastUpdated = 0;
+    this.dedupe = new Dedupe(this._dedupeKey, this._dedupeCompare);
+  }
+  _dedupeKey(site) {
+    return site.hostname;
+  }
+  _dedupeCompare(storedValue, newValue) {
+    return newValue.isPinned;
```
Mardak: Not sure if we want to somehow indicate if both are pinned and neither deduped?

piatra: That's what I'm currently working on, realized just as I wanted to request review.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -25,6 +27,10 @@ this.TopSitesFeed = class TopSitesFeed {
     this.lastUpdated = 0;
     this._tippyTopProvider = new TippyTopProvider();
     this._tippyTopProvider.init();
+    this.dedupe = new Dedupe(this._dedupeKey);
+  }
+  _dedupeKey(site) {
+    return site.hostname;
```
k88hudson: This is throwing an error and crashing the feed when the pinned group contains `null` items, so maybe check `site && site.hostname` and/or filter out `null` values from `pinned` before running dedupe

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -57,7 +63,19 @@ this.TopSitesFeed = class TopSitesFeed {
       frecent = frecent.filter(link => link && link.type !== "affiliate");
     }
 
-    return insertPinned([...frecent, ...DEFAULT_TOP_SITES], pinned).slice(0, TOP_SITES_SHOWMORE_LENGTH);
+    // Group together websites that require deduping.
+    const topsitesGroup = [pinned, frecent, DEFAULT_TOP_SITES];
+    topsitesGroup.forEach(group => group.forEach(site => {
```
k88hudson: Maybe do this to filter out the `null` items and prevent accidental mutation?

```js
for (const group in topsitesGroup) {
  group
    .filter(site => site)
    .map(site => Object.assign({}, site, {hostName: shortURL(site)});
}

```

k88hudson:
🔥 🔥 🔥

as-pine-proxy:
[Treeherder: 866b724e49a45a39c3b3a33961bf97c1f33163b5](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=866b724e49a45a39c3b3a33961bf97c1f33163b5)

PR #2990, opened 2017-07-27
Migration for old profiles

### LINKED ISSUE #2967: Don't show manual migration offer on existing profiles
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1384392

"This is a little weird, because this is the Firefox profile that I've been using for 6+ years now. Why suddenly offer to import data? … seems like this should only be triggering for recently created profiles.

piatra:
Closes #2967

### DIFF
--- a/system-addon/lib/ManualMigration.jsm
+++ b/system-addon/lib/ManualMigration.jsm
@@ -7,11 +7,13 @@ const {utils: Cu} = Components;
 const {actionCreators: ac, actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
 const {Prefs} = Cu.import("resource://activity-stream/lib/ActivityStreamPrefs.jsm", {});
 const MIGRATION_ENDED_EVENT = "Migration:Ended";
+const MS_PER_DAY = 86400000;
 
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 
 XPCOMUtils.defineLazyModuleGetter(this, "MigrationUtils", "resource:///modules/MigrationUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge", "resource://gre/modules/ProfileAge.jsm");
 
 this.ManualMigration = class ManualMigration {
   constructor() {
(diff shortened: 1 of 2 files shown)


system-addon/lib/ManualMigration.jsm
```diff
@@ -23,20 +25,22 @@ this.ManualMigration = class ManualMigration {
...
 
-  isMigrationMessageExpired() {
+  async isMigrationMessageExpired() {
     let migrationLastShownDate = new Date(this._prefs.get("migrationLastShownDate") * 1000);
     let today = new Date();
     // Round down to midnight.
     today = new Date(today.getFullYear(), today.getMonth(), today.getDate());
-
     if (migrationLastShownDate < today) {
+      let profileAge = new ProfileAge(null, null);
```
Mardak: This constructor works with no args. Remove the double `null`

system-addon/lib/ManualMigration.jsm
```diff
@@ -23,20 +25,22 @@ this.ManualMigration = class ManualMigration {
...
+      let profileCreationDate = await profileAge.created;
+      let daysSinceProfileCreation = (Date.now() - profileCreationDate) / MS_PER_DAY;
```
Mardak: Any reason to do this only inside ` if (migrationLastShownDate < today)` ? It seems like all the profile age stuff can be done at the beginning of `isMigrationMessageExpired` and just short circuit the rest of the method? It would also avoid setting the remaining/lastshown prefs below… unless that's desired to set?

system-addon/lib/ManualMigration.jsm
```diff
@@ -51,8 +55,9 @@ this.ManualMigration = class ManualMigration {
    * @param {bool} alreadyExpired Pref flag that is false for the first 3 active days,
    *                              time in which we display the migration message to the user.
    */
-  expireIfNecessary(alreadyExpired) {
-    if (!alreadyExpired && this.isMigrationMessageExpired()) {
+  async expireIfNecessary(alreadyExpired) {
+    const isMigrationExpired = await this.isMigrationMessageExpired();
+    if (!alreadyExpired && isMigrationExpired) {
```
Mardak: adding `await` in the same original line is well within 80 characters on the line:
`    if (!alreadyExpired && await this.isMigrationMessageExpired()) {`

system-addon/lib/ManualMigration.jsm
piatra: The reason I had it like was to reduce the number of `return`s in the function. Makes it easier to reason (at least for me). But I don't feel strongly about it.

piatra:
@Mardak updated & answered your question. Let me know if you still want me to make the change.

Mardak:
The code is doing unnecessary things and will be confusing to others why those changes are being done. E.g., "why are we checking and updating last shown date if profile age is independent of that?"

Move the profile age check and add a comment above that section to make it easier for you to reason.

as-pine-proxy:
[Treeherder: bab389dd80c18fbb65c1f5039c61045cc2e5c215](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=bab389dd80c18fbb65c1f5039c61045cc2e5c215)

PR #2991, opened 2017-07-27
feat(systemaddon): Get Pocket feed for the appropriate locale. Closes…

### LINKED ISSUE #2974: Get Pocket feed for the appropriate locale
@csadilek says accept header might be doable. But if not, the code needs to pass in the locale probably as part of the url query param.

csadilek:
Fix #2974.

Currently requesting other locales than en-* will return the exact same content, so there's no harm in putting this in now. Once other language feeds are ready they should just work (we won't need any client-side changes then).

Relying on accept-lang is possible in a future endpoint version, but in the meantime let's use the locale_lang request parameter. This is also consistent with A-S lang settings, as there we're not relying on accept-lang, but on general.useragent.locale.

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -28,8 +28,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
       const prefs = new Prefs();
       const options = JSON.parse(prefs.get("feeds.section.topstories.options"));
       const apiKey = this._getApiKeyFromPref(options.api_key_pref);
-      this.stories_endpoint = this._produceUrlWithApiKey(options.stories_endpoint, apiKey);
-      this.topics_endpoint = this._produceUrlWithApiKey(options.topics_endpoint, apiKey);
+      const locale = Services.locale.getRequestedLocale();
+      this.stories_endpoint = this._produceFinalEndpointUrl(options.stories_endpoint, apiKey, locale);
+      this.topics_endpoint = this._produceFinalEndpointUrl(options.topics_endpoint, apiKey, locale);
+
       this.read_more_endpoint = options.read_more_endpoint;
       this.stories_referrer = options.stories_referrer;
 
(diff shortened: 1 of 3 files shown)


system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -28,8 +28,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
       const prefs = new Prefs();
       const options = JSON.parse(prefs.get("feeds.section.topstories.options"));
       const apiKey = this._getApiKeyFromPref(options.api_key_pref);
-      this.stories_endpoint = this._produceUrlWithApiKey(options.stories_endpoint, apiKey);
-      this.topics_endpoint = this._produceUrlWithApiKey(options.topics_endpoint, apiKey);
+      const locale = Services.locale.getRequestedLocale();
+      this.stories_endpoint = this._produceFinalEndpointUrl(options.stories_endpoint, apiKey, locale);
```
Mardak: I see that `fetch*` is called later in `init`, but does this actually need to happen during init? We ran into an issue where reftests were failing because network connections were being made aggressively (and denied).

But even ignoring the network connection issue (already fixed for now), instead of using `Services.locale.getRequestedLocale`, `onAction` should watch for `LOCALE_UPDATED` and have a helper method set the endpoint and maybe fetch. The helper would be called from LOCALE_UPDATED and `init` and can get the locale from `this.store.getState().App.locale`

csadilek: Yes, I believe calling fetch* in init is best. We want this to happen as soon as possible to increase the chance we have initial/fresh stories and topics fetched when a new tab is opened.

I will add the refactoring you requested.

Mardak: We don't need to fix the "later" than init network now, but I was thinking not that much long after init. It's just that anything done synchronously during the constructor and INIT counts against one of the talos performance metrics.

as-pine-proxy:
[Treeherder: f294152ad04676ee052335405933f4f254f36902](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f294152ad04676ee052335405933f4f254f36902)

Issue #2787, opened 2017-06-28
Integrate manual migration UI and functionality into Activity Stream


k88hudson:
[Bug 1361294 - show "import from other browser" link on Activity Stream](https://bugzilla.mozilla.org/show_bug.cgi?id=1361294)

We will need to add some UI for manual migration with a "Import Now" button, as well as the API to trigger the import wizard. This will replace the work that was done in #2616.

@gasolin feel free to add more details here

Manual import design:
![manual_import](https://user-images.githubusercontent.com/438537/27658812-1fdf9f7a-5c06-11e7-83f1-c65bfc998246.png)

Hover (adds `[x]` on right side):
![manual_import_hover](https://user-images.githubusercontent.com/438537/27658815-23033fb8-5c06-11e7-9dd4-6b5fc55356ab.png)

Mardak:
There's some logic from the AutoMigrate PR:

```js
+XPCOMUtils.defineLazyModuleGetter(this, "MigrationUtils",
+  "resource:///modules/MigrationUtils.jsm");
+
+this.AutoMigrateFeed = class AutoMigrateFeed {
+  onAction(action) {
+    switch (action.type) {
+      case at.AUTOMIGRATE_MANUAL_IMPORT:
+        // We pass in the type of source we're using for use in telemetry
+        MigrationUtils.showMigrationWizard(action._target.browser.ownerGlobal,
+          [MigrationUtils.MIGRATION_ENTRYPOINT_NEWTAB]);
```
https://github.com/mozilla/activity-stream/pull/2650/files#diff-43cb90fc3065e94a29974775e0f22749R20

gasolin:
can use `Automigrate.isMigratePromptExpired` or `profileAge.created` to check the remain days
http://searchfox.org/mozilla-central/source/toolkit/modules/ProfileAge.jsm#43

Mardak:
From discussing with @gasolin, it'll be safer to just check the `profileAge.created` given the note in https://github.com/mozilla/activity-stream/issues/2787#issuecomment-311779167 about the unexpected external callers of `isMigratePromptExpired`.

bryantmao:
Update UX spec for manual import
![manual import_v1_0711](https://user-images.githubusercontent.com/15391951/28056397-68e8a448-664f-11e7-9a58-5ffba2f29962.png)

piatra:
[...]

PR #3008, opened 2017-07-28
Change margin on Manual Migration message for better RTL support

tomer:
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1385235
issue #2787 introduced a migration request message for new profiles. Unfortunately, this looks worse on RTL builds than on LTR. 

ltr import offer
rtl import offer

The change I am suggesting included using margin-inline-end correctly, so the same rules will apply correctly to both LTR and RTL builds.

fixed wip

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1385235

### DIFF
--- a/system-addon/content-src/components/ManualMigration/ManualMigration.scss
+++ b/system-addon/content-src/components/ManualMigration/ManualMigration.scss
@@ -16,7 +16,8 @@
   }
 
   .icon {
-    margin: 0 12px 0 0;
+    margin: 0;
+    margin-inline-end: 12px;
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: c9ba78f6cff2e50f2ab8721985969f66ac64fa49](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c9ba78f6cff2e50f2ab8721985969f66ac64fa49)

PR #3018, opened 2017-07-29
feat(pocket): Use geo and locale to determine when to enable Pocket

### LINKED ISSUE #3004: Check desired geo and locales list for turning on Pocket in 56
See https://docs.google.com/spreadsheets/d/1hyI166PJFAFkBVaNqIz7MkFApdpo2C_dUwO0-GwnSxA

We need logic to check the geo pref "browser.search.region" It should be a 2 letter uppercase value or blank if search hasn't set it yet (but that should happen pretty soon after first start). If it doesn't exist, we need to observe when it does change so that the appropriate showTopStoriesByDefault default pref gets set. https://github.com/mozilla/activity-stream/blob/0675a44abd1b3954565501df7b71f8f5e9fb32e5/system-addon/lib/ActivityStream.jsm#L28

The logic to turn on is that the geo AND locale need to match the "desired" list.

Mardak:
Fix #3004. r?@k88hudson Some caveats that are mostly noted in the comment blocks:

Most users will already have a geo set, but for those with a new profile, it's possible to need to wait for geo to be set soon after first run. Unfortunately, default prefs are determined when `ActivityStream.jsm` loads, so as a quick hack, I made it detect the pref change and directly set the default for the topstories section feed.

This doesn't update the options, which are needed to remember if the section was on/off by default to show the prefs. This PR currently does not re-set the default for the options and works around this by having the prefs also appear if the section was somehow turned on (in this case, probably because the geo was just set).

Separate from that, when writing the AS.jsm load tests, it ends up trying to load NewTabInit which results in istanbul complaining that it's not tested. So for now I just added the "ignore next" as it isn't really in scope of this PR.

### DIFF
--- a/system-addon/content-src/components/PreferencesPane/PreferencesPane.jsx
+++ b/system-addon/content-src/components/PreferencesPane/PreferencesPane.jsx
@@ -53,6 +53,12 @@ class PreferencesPane extends React.Component {
     const props = this.props;
     const prefs = props.Prefs.values;
     const isVisible = this.state.visible;
+
+    // Show prefs for top stories only if we have some top story option and
+    // either it should be shown by default or it was already turned on
+    const showTopStoriesPref = this.topStoriesOptions &&
+      (this.topStoriesOptions.shownByDefault || prefs["feeds.section.topstories"]);
+
     return (
       <div className="prefs-pane-wrapper" ref="wrapper">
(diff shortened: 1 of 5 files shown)


Mardak:
Eh. It felt too wrong with all the caveats, so I rewrote it with a completely different approach.

Issue #2776, opened 2017-06-27
Land localized default top sites


sarracini:
See https://docs.google.com/spreadsheets/d/15yLsFBkic5DFUSdbvSFlR3Rzjxqwmizq9EmksK28l6I

We need to figure out the mechanism for using default top sites depending on the locale

…or geo? https://github.com/mozilla/activity-stream/issues/2776#issuecomment-318877317

tspurway:
- default top sites themselves broken  down by locale
- actual assets/icons for all topsites
- json doc mapped by locale, code for localized selection

zbraniecki:
L10n-Drivers team can help you get the list of most popular sites per locale if you'd like to use that to influence the defaults.

Mardak:
I believe according to mpopova (activity-stream PM), for the initial release, we'll just do US/China/rest; but having a better way than looking at top sites lists to get these 6 for each locale might be better going forwards.

AdamHillier:
Amazon is listed under "Global" but amazon.com won't automatically redirect to the correct local version amazon.{fr, it, co.uk, jp} so will we (maybe not for 56) have to localise the top sites in a more granular way?

Mardak:
depends on what's needed for shield beta 56

Mardak:
Initial verbal example was en-US and de. Blocked on product decision of which exact locales for shield study.

Mardak:
Development can start to support just en-US and de. Adding more locales or changing sites should be straightforward once the logic is there to support 2.

mariappopova:
The top sites for 56 will include Tier 1 and Tier 2 countries + default (rest of the world):  https://docs.google.com/spreadsheets/d/15yLsFBkic5DFUSdbvSFlR3Rzjxqwmizq9EmksK28l6I/edit#gid=170333837

UX is currently reviewing and coming up with tippytop/icons for those default sites

Mardak:
So it sounds like Canada doesn't actually get its own locale (i.e., en-CA doesn't actually exist). This means canadian users who are using en-US will just get the en-US. mconnor has mentioned how search engines get selected based on geo, but that's not something activity stream is doing currently (but could potentially reuse what search has done ??)

Mardak:
Approach for 56 can probably just grab the locale from Services.locale and later clean up in #2999. 

Use that locale to key into an updated "default.sites" that is an object with keys of locales falling back to a "default"
[...]

PR #3021, opened 2017-07-30
feat(pocket): Use geo and locale to determine when to enable Pocket

### LINKED ISSUE #3004: Check desired geo and locales list for turning on Pocket in 56

Mardak:
Fix #3004. r?@k88hudson This cleans up `AS.jsm` a little bit to not have a separate `SECTIONS` and just inlines it to the rest of `FEEDS_DATA`. Also sorts things alphabetically.

The main core change is that prefs can have a `valueFunc` that will compute a value based on geo/locale. If geo isn't available, it'll observe the pref and update the pref config as well as updating the default.

This should also greatly simplify things for #2776.

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -22,34 +22,45 @@ const {TelemetryFeed} = Cu.import("resource://activity-stream/lib/TelemetryFeed.
 const {TopSitesFeed} = Cu.import("resource://activity-stream/lib/TopSitesFeed.jsm", {});
 const {TopStoriesFeed} = Cu.import("resource://activity-stream/lib/TopStoriesFeed.jsm", {});
 
+const GEO_PREF = "browser.search.region";
 const REASON_ADDON_UNINSTALL = 6;
 
-// For now, we only want to show top stories by default to the following locales
-const showTopStoriesByDefault = ["en-US", "en-CA"].includes(Services.locale.getRequestedLocale());
-// Sections, keyed by section id
-const SECTIONS = new Map([
-  ["topstories", {
-    feed: TopStoriesFeed,
-    prefTitle: "Fetches content recommendations from a configurable content provider",
-    showByDefault: showTopStoriesByDefault
-  }]
-]);
-
-const SECTION_FEEDS_CONFIG = Array.from(SECTIONS.entries()).map(entry => {
-  const id = entry[0];
-  const {feed: Feed, prefTitle, showByDefault: value} = entry[1];
-  return {
-    name: `section.${id}`,
-    factory: () => new Feed(),
-    title: prefTitle || `${id} section feed`,
-    value
-  };
-});
-
+// Configure default Activity Stream prefs with a plain `value` or a `getValue`
+// that computes a value. A `value_local_dev` is used for development defaults.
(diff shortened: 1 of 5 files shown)


system-addon/lib/ActivityStream.jsm
```diff
@@ -22,35 +22,46 @@ const {TelemetryFeed} = Cu.import("resource://activity-stream/lib/TelemetryFeed.
...
+      hidden: !PREFS_CONFIG.get("feeds.section.topstories").valueFunc(args),
+      learn_more_endpoint: "https://getpocket.com/firefox_learnmore?src=ff_newtab",
+      provider_description: "pocket_feedback_body",
+      provider_icon: "pocket",
+      provider_name: "Pocket",
+      read_more_endpoint: "https://getpocket.com/explore/trending?src=ff_new_tab",
+      stories_endpoint: "https://getpocket.com/v3/firefox/global-recs?consumer_key=$apiKey&locale_lang=$locale",
+      stories_referrer: "https://getpocket.com/recommendations",
+      survey_link: "https://www.surveymonkey.com/r/newtabffx",
+      topics_endpoint: "https://getpocket.com/v3/firefox/trending-topics?consumer_key=$apiKey&locale_lang=$locale"
```
Mardak: I suppose with `valueFunc`, at some point we could just insert `args.locale` directly into these endpoints instead of having `TopStoriesFeed` separately get the locale and insert it from #2991. @csadilek 

Theoretically `valueFunc` will be called when locale changes too (it doesn't do it in this PR), but that also means `TopStoriesFeed` wouldn't need to watch for locale changes as its prefs will just update.

system-addon/lib/ActivityStream.jsm
```diff
@@ -22,35 +22,46 @@ const {TelemetryFeed} = Cu.import("resource://activity-stream/lib/TelemetryFeed.
...
-  return {
-    name: `section.${id}`,
-    factory: () => new Feed(),
-    title: prefTitle || `${id} section feed`,
-    value
-  };
-});
-
+// Configure default Activity Stream prefs with a plain `value` or a `valueFunc`
+// that computes a value. A `value_local_dev` is used for development defaults.
```
Mardak: k88: I would probably name this `getValue`, but up to you
Sure.

as-pine-proxy:
[Treeherder: 4a7519f9ee814fa4817989e83eb7a68170bf25a1](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4a7519f9ee814fa4817989e83eb7a68170bf25a1)

Issue #2127, opened 2017-02-06
List and break down key functional tests to write before landing.


dmose:
We want to have a minimal set of mochitests running when we land in tree so that if/when key toolkit/platform APIs change out from under us, the tree breaks if the A-S code is not patched at the same time.

sarracini:
https://public.etherpad-mozilla.org/p/as-mc-failures

Mardak:
These tests will be implemented as part of #2880:

test search box appears
test top sites section appears when that section is on
test top sites section doesn't appear when that section is off

Other tests probably useful:
click interaction on various elements including pref pane
sections gets populated with dummy data including top sites, recommendations

PR #3026, opened 2017-07-31
fix(tests): Add mochitests that run successfully whether enabled or not

### LINKED ISSUE #2880: Add mochitest to ensure correct behavior enabled/disabled on all channels
There's some complexity here in that for initial rollout, activity-stream is enabled on nightly 56 and disabled beta 56. And beta 56 doesn't even build the extension.

This is to make sure tests don't start failing because of a nightly -> beta transition where build/default prefs change.

Mardak:
Fix #2880. r?@dmose Beefs up `as_load_location` to work preloaded or not and activity-stream.enabled or not. Adds a `head.js` that turns on activity-stream with a set of basic actual react rendering tests covering some of those itemized in #2127. Also cleans up eslint to use the fancy "plugin:mozilla/browser-test" that auto-globalizes-head.js helpers (and other test globals).

Try looks good when a-s.enabled: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2b1081f19bfc7b5adbd761e347a26fd264b55783
And disabled: https://treeherder.mozilla.org/#/jobs?repo=try&revision=34f4c5cf8fbe1c0e2933c429e270e03817b55b2c

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -76,8 +76,10 @@ function init(reason) {
  * @param  {type} reason Reason for uninitialization. Could be uninstall, upgrade, or PREF_OFF
  */
 function uninit(reason) {
+  // Make sure to only uninit once in case both pref change and shutdown happen
   if (activityStream) {
     activityStream.uninit(reason);
+    activityStream = null;
   }
 }
 
--- a/system-addon/test/functional/mochitest/browser_as_load_location.js
+++ b/system-addon/test/functional/mochitest/browser_as_load_location.js
@@ -1,34 +1,53 @@
 "use strict";
 
-let Cu = Components.utils;
-Cu.import("resource://gre/modules/Services.jsm");
-
 /**
- * Tests that opening a new tab opens a page with the expected activity stream
- * content.
+ * Helper to test that a newtab page loads its html document.
  *
- * XXX /browser/components/newtab/tests/browser/browser_newtab_overrides in
- * mozilla-central is where this test was adapted from.  Once we get decide on
- * and implement how we're going to set the URL in mozilla-central, we may well
- * want to (separately from this test), clone/adapt that entire file for our
- * new setup.
+ * @param selector {String} CSS selector to find an element in newtab content
+ * @param message {String} Description of the test printed with the assertion
  */
-add_task(async function checkActivityStreamLoads() {
-  const asURL = "resource://activity-stream/data/content/activity-stream.html";
-
+async function checkNewtabLoads(selector, message) {
   // simulate a newtab open as a user would
   BrowserOpenTab();
 
   // wait until the browser loads
   let browser = gBrowser.selectedBrowser;
-  await BrowserTestUtils.browserLoaded(browser);
+  await waitForPreloaded(browser);
 
   // check what the content task thinks has been loaded.
-  await ContentTask.spawn(browser, {url: asURL}, args => {
-    Assert.ok(content.document.querySelector("body.activity-stream"),
-      'Got <body class="activity-stream" Element');
-  });
+  let found = await ContentTask.spawn(browser, selector, arg =>
+    content.document.querySelector(arg) !== null);
+  ok(found, message);
 
   // avoid leakage
   await BrowserTestUtils.removeTab(gBrowser.selectedTab);
+}
+
+// Test with activity stream on
+async function checkActivityStreamLoads() {
+  await checkNewtabLoads("body.activity-stream", "Got <body class='activity-stream'> Element");
+}
+
+// Run a first time not from a preloaded browser
+add_task(async function checkActivityStreamNotPreloadedLoad() {
+  gBrowser.removePreloadedBrowser();
+  await checkActivityStreamLoads();
 });
+
+// Run a second time from a preloaded browser
+add_task(checkActivityStreamLoads);
+
+// Test with activity stream off
+async function checkNotActivityStreamLoads() {
+  await SpecialPowers.pushPrefEnv({set: [["browser.newtabpage.activity-stream.enabled", false]]});
+  await checkNewtabLoads("body:not(.activity-stream)", "Got <body> Element not for activity-stream");
+}
+
+// Run a first time not from a preloaded browser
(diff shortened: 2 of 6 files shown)


system-addon/bootstrap.js
```diff
@@ -78,6 +78,7 @@ function init(reason) {
 function uninit(reason) {
   if (activityStream) {
     activityStream.uninit(reason);
+    activityStream = null;
```
Mardak: This is to allow tests to clear/disable activity-stream and immediately uninstall (which resulted in back-to-back `uninit`s).

dmose: How about adding a comment to the code to this effect?

system-addon/test/functional/mochitest/browser_as_load_location.js
```diff
@@ -1,34 +1,48 @@
...
- * XXX /browser/components/newtab/tests/browser/browser_newtab_overrides in
- * mozilla-central is where this test was adapted from.  Once we get decide on
- * and implement how we're going to set the URL in mozilla-central, we may well
- * want to (separately from this test), clone/adapt that entire file for our
- * new setup.
- */
-add_task(async function checkActivityStreamLoads() {
-  const asURL = "resource://activity-stream/data/content/activity-stream.html";
-
+// Helper to test that a newtab page loads
```
dmose: JSDoc would be nice here

system-addon/test/functional/mochitest/browser_as_load_location.js
```diff
@@ -1,34 +1,48 @@
...
+  gBrowser.removePreloadedBrowser();
+  await checkActivityStreamLoads();
+});
+
+// Run a second time from a preloaded browser
+add_task(checkActivityStreamLoads);
+
+// Test with activity stream off
+async function checkNotActivityStreamLoads() {
+  await SpecialPowers.pushPrefEnv({set: [["browser.newtabpage.activity-stream.enabled", false]]});
```
dmose: Does this need to be popped manually in cleanup somewhere, since it's not in a scopedPushPrefs?

system-addon/test/functional/mochitest/head.js
```diff
@@ -0,0 +1,93 @@
...
+    let tab = await BrowserTestUtils.openNewForegroundTab(gBrowser, "about:newtab", false);
+
+    // Specially wait for potentially preloaded browsers
+    let browser = tab.linkedBrowser;
+    await waitForPreloaded(browser);
+
+    // Wait for React to render something
+    await BrowserTestUtils.waitForCondition(() => ContentTask.spawn(browser, {},
+      () => content.document.getElementById("root").children.length),
+      "Should render activity stream content");
```
dmose: I'm a little concerned that this could race, since just because root has children doesn't necessarily tell us much about how far along rendering is.  That said, I don't have a better suggestion.

system-addon/test/functional/mochitest/head.js
```diff
@@ -0,0 +1,93 @@
...
+      let contentArg = await before({pushPrefs: scopedPushPrefs, tab});
+      let contentResult = await ContentTask.spawn(browser, contentArg, contentTask);
+      await after(contentResult);
+    } finally {
+      // Clean up for next tests
+      await scopedPopPrefs();
+      await BrowserTestUtils.removeTab(tab);
+    }
+  };
+
```
dmose: This is great.  Someday, it (or something like it) should be hoisted somewhere more central in m-c so that it's generally available.

system-addon/test/functional/mochitest/head.js
```diff
@@ -0,0 +1,93 @@
...
+// Activity Stream tests expect it to be enabled, and make sure to clear out any
+// preloaded browsers that might have about:newtab that we don't want to test
+const ACTIVITY_STREAM_PREF = "browser.newtabpage.activity-stream.enabled";
+pushPrefs([ACTIVITY_STREAM_PREF, true]);
+gBrowser.removePreloadedBrowser();
+
+/**
+ * Helper to wait for potentially preloaded browsers to load
+ */
```
dmose: Probably worth pointing out in the comment that this is effectively the "load" event, and contains no guarantee about what has or hasn't been rendered.

system-addon/test/functional/mochitest/browser_as_load_location.js
Mardak: The purpose of the push is so that it does autopop.... at the end of the whole test run it seems.

system-addon/test/functional/mochitest/browser_as_load_location.js
Mardak: Added

system-addon/test/functional/mochitest/head.js
Mardak: Commented

system-addon/test/functional/mochitest/head.js
Mardak: Sooooooo… this was very odd to add in the first place, and it's mostly a sanity check instead of an even odder looks-like-no-op. From my testing on slower debug builds, it looks to always match the condition on the first check.

In fact, by doing a `await ContentTask.spawn(browser, {}, () => {})`, enough time will have passed from sending and waiting for the message. However, without doing that "nothing", tests fail because there's no data for React to `render`. React is actually rendering null resulting in an empty comment, e.g., https://stackoverflow.com/questions/38923805/what-is-react-empty

system-addon/bootstrap.js
Mardak: Commented

system-addon/test/functional/mochitest/head.js
Mardak: The chaining? I suppose it's "overengineered" in that I didn't quite use all the functionality in the actual tests, but it did get me to write some additional actual activity-stream related tests to make sure the helper actually worked. :)

as-pine-proxy:
[Treeherder: b9633b96829d377682ed260d4cd2d93abfdecf5f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b9633b96829d377682ed260d4cd2d93abfdecf5f)

PR #3031, opened 2017-07-31
fix(topstories): Closes #3029 Add ETLDs for top stories

### LINKED ISSUE #3029: Card domain does not cut off all TLDs
Reported via slack by Nate:
pasted image at 2017_07_28 03_03 pm

> Looks like we have logic on these cards that cuts off ".com" from domains (see Wired above), but looks like we may be missing other TLDs (ex .org, see HBR's card)

sarracini:
Fix #3029. @dmose, can you also review this approach?


as-pine-proxy:
[Treeherder: fa2a0ee29a75dd670adcb228d2b55185f3e238ec](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=fa2a0ee29a75dd670adcb228d2b55185f3e238ec)

PR #3034, opened 2017-07-31
feat(systemaddon): Closes #2645 Add <meta> CSP to activity-stream.html

### LINKED ISSUE #2645: add <meta> CSP tag to the system-addon HTML
Since resource: URIs are framable, it come up at the security review that we should add a CSP <meta> tag.   dveditz suggested that at the very least we want no inline scripts and no ancestor framing.  We may be able to use the CSP from the SDK add-on as a starting point.

sarracini:
CSP tag on Report-Only mode which reports to a prod endpoint as per the comment in ticket fix #2645 

@dmose @Mardak @k88hudson @ncloudioj  what do we think?

### DIFF
--- a/system-addon/data/content/activity-stream.html
+++ b/system-addon/data/content/activity-stream.html
@@ -2,6 +2,7 @@
 <html lang="en-us" dir="ltr">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


ncloudioj:
@sarracini LGTM

as-pine-proxy:
[Treeherder: 099c271a3aabe35684d00788d2ad993e7d69eebc](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=099c271a3aabe35684d00788d2ad993e7d69eebc)

PR #3037, opened 2017-07-31
feat(topstories): Get new recommendations on locale|options change

### LINKED ISSUE #3000: Get new recommendations on locale change
This was split out from https://github.com/mozilla/activity-stream/pull/2991 where `TopStoriesFeed` should watch `LOCALE_UPDATED` to update its endpoints.

csadilek:
Follow up to https://github.com/mozilla/activity-stream/pull/3021, fixes #3000, by making sure the feed re-initializes when its options have changed.

This also includes the refactoring to no longer read the locale from within the TopStoriesFeed.

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -16,21 +16,21 @@ const STORIES_UPDATE_TIME = 30 * 60 * 1000; // 30 minutes
 const TOPICS_UPDATE_TIME = 3 * 60 * 60 * 1000; // 3 hours
 const STORIES_NOW_THRESHOLD = 24 * 60 * 60 * 1000; // 24 hours
 const SECTION_ID = "TopStories";
+const FEED_PREF = "feeds.section.topstories";
+const SECTION_OPTIONS_PREF = "feeds.section.topstories.options";
 
 this.TopStoriesFeed = class TopStoriesFeed {
-  constructor() {
-    this.storiesLastUpdated = 0;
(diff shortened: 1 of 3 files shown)


system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -16,21 +16,21 @@ const STORIES_UPDATE_TIME = 30 * 60 * 1000; // 30 minutes
...
 this.TopStoriesFeed = class TopStoriesFeed {
-  constructor() {
-    this.storiesLastUpdated = 0;
-    this.topicsLastUpdated = 0;
-  }
 
   init() {
     try {
+      this.storiesLastUpdated = 0;
+      this.topicsLastUpdated = 0;
```
Mardak: This whole `try` block is a bit odd in that various things that don't need to be inside are. E.g., this assignments really shouldn't be throwing.

Don't need to fix now, but what are we actually expecting to fail? I could see `JSON.parse` failing.. anything else?

as-pine-proxy:
[Treeherder: 500292001a3de3c49a61db9951b9b35a407ca523](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=500292001a3de3c49a61db9951b9b35a407ca523)

system-addon/lib/TopStoriesFeed.jsm
csadilek: @Mardak true, nothing else throwing here anymore. so, the scope is too wide. It's really about catching invalid JSON in .options.

PR #3046, opened 2017-07-31
fix(topstories): Ctrl-click/middle mouse button does not open new tab

### LINKED ISSUE #3033: Ctrl+Click or middle mouse button click does not open the Pocket tile in a new tab
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1385871

This is because Pocket specially handles clicks to insert a referrer. We probably just need to pass in the click event details as the gBrowser helper functions should open appropriately based on event.

Probably regressed by #2959.

csadilek:
Fixes #3033.

### DIFF
--- a/system-addon/content-src/components/Card/Card.jsx
+++ b/system-addon/content-src/components/Card/Card.jsx
@@ -31,7 +31,11 @@ class Card extends React.Component {
   }
   onLinkClick(event) {
     event.preventDefault();
-    this.props.dispatch(ac.SendToMain({type: at.OPEN_LINK, data: this.props.link}));
+    const {altKey, button, ctrlKey, metaKey, shiftKey} = event;
+    this.props.dispatch(ac.SendToMain({
+      type: at.OPEN_LINK,
+      data: Object.assign(this.props.link, {event: {altKey, button, ctrlKey, metaKey, shiftKey}})
+    }));
     this.props.dispatch(ac.UserEvent({
       event: "CLICK",
       source: this.props.eventSource,
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -225,10 +225,12 @@ class PlacesFeed {
         Pocket.savePage(action._target.browser, action.data.site.url, action.data.site.title);
         break;
       case at.OPEN_LINK: {
+        const win = action._target.browser.ownerGlobal;
+        const where = win.whereToOpenLink(action.data.event);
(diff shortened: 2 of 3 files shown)


system-addon/content-src/components/Card/Card.jsx
```diff
@@ -31,7 +31,10 @@ class Card extends React.Component {
   }
   onLinkClick(event) {
     event.preventDefault();
-    this.props.dispatch(ac.SendToMain({type: at.OPEN_LINK, data: this.props.link}));
+    this.props.dispatch(ac.SendToMain({
+      type: at.OPEN_LINK,
+      data: Object.assign(this.props.link, {where: (event.ctrlKey || event.metaKey || event.button === 1) ? "tab" : "current"})
```
Mardak: So this supports tabs but not the other click behaviors, e.g., "window" (on shift) and "save" (on alt) or other behaviors supported in `whereToOpenLink`:
https://searchfox.org/mozilla-central/source/browser/base/content/utilityOverlay.js#111-159

Not sure if there's a good way to just send the event over to call that method on the browser.

csadilek: Yeah, I didn't find one :(. We have support for "window" though via the context menu (not shortcut) (incl. private windows).

I think this is better than not having it for 56, but I will leave it up to you guys. I am fine either way and can dig in more.

k88hudson: Looks like you could call`browser.ownerGlobal.whereToOpenLink` with the event, if you can pass the whole event up

csadilek:
OK, I can give it another try.

system-addon/content-src/components/Card/Card.jsx
Mardak: A fake event should work:
```js
const {altKey, button, ctrlKey, metaKey, shiftKey} = event;
dispatch({…
  data: … {event: {altKey, button, ctrlKey, metaKey, shiftKey}, …}
```

csadilek: @Mardak ah I will try that. Was just looking for ways to get the event across again.

csadilek:
@Mardak @k88hudson thanks, updated.

as-pine-proxy:
[Treeherder: 56a25617de9c4426bcd87ec6eabe0663a41b72f3](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=56a25617de9c4426bcd87ec6eabe0663a41b72f3)

PR #3049, opened 2017-08-01
feat (systemaddon): #2944 add tippy top icons

### LINKED ISSUE #2944: Add TippyTop sites to the System Addon
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1382703
Import the Tippy Top sites list and icons into the System Addon and use them for matching URLs.

rlr:
Fixes #2944

image


rlr:
We still need to cut a limited tippy top release but that can be done as a quick followup.

r? @Mardak and/or @jaredkerim and/or ?

system-addon/lib/TippyTopProvider.jsm
```diff
@@ -0,0 +1,66 @@
...
+
+this.TippyTopProvider = class TippyTopProvider {
+  constructor() {
+    this._sitesByDomain = {};
+  }
+  async init() {
+    // Load the Tippy Top sites from the json manifest.
+    fetch(TIPPYTOP_JSON_PATH)
+      .then(response => response.json())
+      .then(sites => {
```
Mardak: These should be `await`s given that we're inside an `async` function

```js
try {
  for (const site of await (await fetch(PATH)).json()) {
 …
} catch(error) …

File: system-addon/lib/TippyTopProvider.jsm [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+  return domain;
+}
+
+function getPath(url) {
+  return new URL(url).pathname;
+}
+
+this.TippyTopProvider = class TippyTopProvider {
+  constructor() {
+    this._sitesByDomain = {};
```
Mardak: Let's make this a `Map`

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -65,9 +69,15 @@ this.TopSitesFeed = class TopSitesFeed {
       }
     }
 
-    // Now, get a screenshot for every item
+    // Now, get a tippy top icon or screenshot for every item
     for (let link of links) {
       if (!link) { continue; }
+
+      // Check for tippy top icon.
+      link = this._tippyTopProvider.processSite(link);
```
Mardak: So, this seems like it could fail or give wrong results if `TippyTopProvider` hasn't finished adding to `_sitesByDomain`. I would guess the correct fix is to make `processSite` `async`. But we can skip this for now.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -22,8 +23,11 @@ const DEFAULT_TOP_SITES = [];
 this.TopSitesFeed = class TopSitesFeed {
   constructor() {
     this.lastUpdated = 0;
+    this._tippyTopProvider = new TippyTopProvider();
```
Mardak: @k88hudson any thoughts on having this `TippyTopProvider` vs a `TippyTopFeed`? If we do want a feed, we should just refactor it later. I could see the feed dispatching a mapping of domains to urls, but then again, would we want all that data going to stores and the content process?

yamscripts.yml
```diff
@@ -40,9 +40,17 @@ scripts:
   # mochitest somewhere.
   mochitest-debug: (cd ../mozilla-central && ./mach mochitest --jsdebugger browser/extensions/activity-stream/test/functional/mochitest )
 
+  cleanmc: rimraf ../mozilla-central/browser/extensions/activity-stream
+
+  cleanTippyTop: rimraf system-addon/data/content/tippytop
+
+  copyTippyTopManifest: cpx "node_modules/tippy-top-sites/top_sites.json" system-addon/data/content/tippytop
+
+  copyTippyTopImages: cpx "node_modules/tippy-top-sites/images/**/*" system-addon/data/content/tippytop/images
```
Mardak: nit: this seems better organized as
```
tippymc:
  pre: rimraf …
  manifest: cpx …
  images: cpx …
```

yamscripts.yml
```diff
@@ -40,9 +40,17 @@ scripts:
...
+
 # buildmc: Export the bootstraped add-on to mozilla central
   buildmc:
-    pre: rimraf ../mozilla-central/browser/extensions/activity-stream
+    pre: =>cleanmc && =>cleanTippyTop && =>copyTippyTopManifest && =>copyTippyTopImages
```
Mardak: then this would just be `=> cleanmc && => tippymc`

system-addon/lib/TippyTopProvider.jsm
```diff
@@ -0,0 +1,66 @@
...
+    fetch(TIPPYTOP_JSON_PATH)
+      .then(response => response.json())
+      .then(sites => {
+        for (let site of sites) {
+          if ("url" in site) {
+            this._sitesByDomain[getDomain(site.url)] = site;
+          }
+          if ("urls" in site) {
+            for (let url of site.urls) {
+              this._sitesByDomain[getDomain(url)] = site;
```
Mardak: nit: This seems a bit awkward with the repetition. Maybe
```js
for (const url of site.url ? [site.url] : site.urls || []) {
  this._sitesByDomain.set(getDomain(url), site);
```

Or maybe that's just too clever :p It would want a comment ;) Also not quite exactly the same logic as it doesn't allow for both `url` and `urls` to exist..

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -43,12 +43,24 @@ class TopSite extends React.Component {
...
         <a href={link.url} onClick={this.onLinkClick}>
           <div className="tile" aria-hidden={true}>
               <span className="letter-fallback">{title[0]}</span>
-              <div className={screenshotClassName} style={style} />
+              {showScreenshot &&
+                <div className={screenshotClassName} style={style} />
+              }
+              {!showScreenshot &&
+                <div className="tippy-top-icon" style={style} />
+              }
```
Mardak: Or… we could be even fancier :p ?
`<div {...imageAttributes} />`

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -22,8 +23,11 @@ const DEFAULT_TOP_SITES = [];
...
   }
   init() {
+    this._tippyTopProvider.init();
```
Mardak: Sorry for the conflict with removing `TopSitesFeed.init` Maybe just call this just after constructing it from `TopSitesFeed.constructor` ?

system-addon/lib/TopSitesFeed.jsm
rlr: Yeah, figured it should just should just move on if it isn't fully init'd instead of waiting. I guess we would want to dispatch an action when init is done to refresh again. Maybe there should be a tippy top feed like you suggested 🤔

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -43,12 +43,24 @@ class TopSite extends React.Component {
...
     const screenshotClassName = `screenshot${link.screenshot ? " active" : ""}`;
     const topSiteOuterClassName = `top-site-outer${isContextMenuOpen ? " active" : ""}`;
-    const style = {backgroundImage: (link.screenshot ? `url(${link.screenshot})` : "none")};
+    const tippyTopIcon = link.tippyTopIcon;
+    const showScreenshot = !tippyTopIcon;
+    let style;
+    if (showScreenshot) {
+      style = {backgroundImage: (link.screenshot ? `url(${link.screenshot})` : "none")};
+    } else {
+      style = {backgroundImage: `url(${tippyTopIcon})`, backgroundColor: link.backgroundColor || "none"};
```
rlr: that turned out nice like that 👍

system-addon/lib/TippyTopProvider.jsm
rlr: I kind of likey 😉

system-addon/lib/TippyTopProvider.jsm
rlr: Do I need to stub it out differently in tests to do this? I can't get them to work now for some reason.

rlr: oops nvm. there waz a typo

system-addon/lib/TopSitesFeed.jsm
Mardak: Well, Provider doesn't have the usual dispatch, but it async init returns a promise that TopSitesFeed could use to have refresh wait until Provider is `init`ed . E.g.,
```
TopSitesFeed {
constructor() {
  const provider = new TippyTopProvider();
  this._initedProvider = provider.init(); // init needs to return the provider
}
async refresh() {
  …
  (await this._initedProvider).processSite(link);
```
This can be separately handled as a followup as I expect the tests would need to change a bit.

rlr: cool, I'll look into that after doing the tippy top release for default sites

as-pine-proxy:
[Treeherder: 7e3ee5c3f1d47e9f2e42812c0d65f55f05dd26b2](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7e3ee5c3f1d47e9f2e42812c0d65f55f05dd26b2)

PR #3063, opened 2017-08-01
mcbackport bug 1385090; Closes #3060; Closes #2819

### LINKED ISSUE #2819: resource://activity-stream/data/content/activity-stream.bundle.js changes restored new tab title to "undefined" when switching to another tab

### LINKED ISSUE #3060: Backport Bug 1385090 - Pass Remote Pages instance from AboutNewTab on override r=mossop
https://bugzilla.mozilla.org/show_bug.cgi?id=1385090
https://hg.mozilla.org/integration/autoland/rev/c291fa5c4c8a modified `ActivityStreamMessageChannel.jsm` so we need to copy it to this repository before we export to m-c again.

This shouldn't be committed until the above change makes it to m-c.

AdamHillier:
Closes #3060

### DIFF
--- a/system-addon/lib/ActivityStreamMessageChannel.jsm
+++ b/system-addon/lib/ActivityStreamMessageChannel.jsm
@@ -126,11 +126,9 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
    *                 between the main process and child pages
    */
   createChannel() {
-    //  RemotePageManager must be disabled for about:newtab, since only one can exist at once
-    if (this.pageURL === ABOUT_NEW_TAB_URL) {
-      AboutNewTab.override();
-    }
-    this.channel = new RemotePages(this.pageURL);
+    //  Receive AboutNewTab's Remote Pages instance, if it exists, on override
+    const channel = this.pageURL === ABOUT_NEW_TAB_URL && AboutNewTab.override(true);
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 8d25c9d8c0f96ae2cf7206524c4a017b5873d414](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8d25c9d8c0f96ae2cf7206524c4a017b5873d414)

PR #3064, opened 2017-08-01
fix (systemaddon): #3051 update to tippy top v1.3.1 (default sites only)

### LINKED ISSUE #3051: Include tippy-top-sites of only the default sites
Split from #3049 as a followup for #2944 to use some of the original and newly added icons from https://github.com/mozilla/tippy-top-sites/pull/72

The default sites are being added in #3022 for #2776

rlr:
Fix #3051. This version of tippy-top-sites only has the 16 sites that are included in our default top sites list. See https://github.com/mozilla/tippy-top-sites/blob/v1.3.1/top_sites.json

r? @Mardak

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "redux-thunk": "2.1.0",
     "redux-watch": "1.1.1",
     "reselect": "2.5.4",
-    "tippy-top-sites": "1.2.2"
+    "tippy-top-sites": "1.3.1"
   },
   "devDependencies": {
     "@kadira/storybook": "2.29.5",


rlr:
Also see: https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/ActivityStream.jsm#L25

as-pine-proxy:
[Treeherder: 763be035c8518a877f3d4cb06067619a7acc089b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=763be035c8518a877f3d4cb06067619a7acc089b)

PR #3066, opened 2017-08-01
fix(init): Let content know things are initialized and fix "early" issues

### LINKED ISSUE #3054: addMessageListener not defined
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1386445
Every now and then, I get an empty homepage. Found the following message in the console:
```
ReferenceError: addMessageListener is not defined[Weitere Informationen] activity-stream.bundle.js:772:3
	initStore resource://activity-stream/data/content/activity-stream.bundle.js:772:3
	<anonym> resource://activity-stream/data/content/activity-stream.bundle.js:2705:15
	__webpack_require__ resource://activity-stream/data/content/activity-stream.bundle.js:20:12
	<anonym> resource://activity-stream/data/content/activity-stream.bundle.js:66:18
	<anonym> resource://activity-stream/data/content/activity-stream.bundle.js:1:11
```
This is on Nightly.

Mardak:
Fix #3054 with https://bugzilla.mozilla.org/show_bug.cgi?id=1386445 but this can land separately. When the bug lands, it will allow activity stream to load earlier, and some things don't quite work right when running earlier. Those are fixed here mostly small fixes checking if things actually exist before using them.

One meaningful change is `ac.BroadcastToContent({` of `INIT` that causes the page/content's `Reducer` to actually set `initialized` on the `App` state, which also needed a fix to pull in `INITIAL_STATE.App` as `prevState`

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -37,7 +37,7 @@ const INITIAL_STATE = {
 function App(prevState = INITIAL_STATE.App, action) {
   switch (action.type) {
     case at.INIT:
-      return Object.assign({}, action.data || {}, {initialized: true});
+      return Object.assign({}, prevState, action.data || {}, {initialized: true});
     case at.LOCALE_UPDATED: {
       if (!action.data) {
         return prevState;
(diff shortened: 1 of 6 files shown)


Mardak:
Added tests for INIT and Reducer. The pref/json one looks like it'll be fixed separately. Looks like there's no tests for `Base` so filing that.

as-pine-proxy:
[Treeherder: 138c6c26fdb452f5b73ef44005445a9667af0d20](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=138c6c26fdb452f5b73ef44005445a9667af0d20)

PR #3069, opened 2017-08-02
Fix(a11y): closes #2997, add basic a11y support to section-info-option

### LINKED ISSUE #2997: Make section-info-option accessible
From https://github.com/mozilla/activity-stream/issues/2874#issuecomment-318425638

section-info-option is not accessible (for example screen/keyboard reader user will never be able to send feedback):
- it has to be in tab index
- it has also no semantic information (could be role="note")
- sr-only could be eliminated in favour title or aria-label
- it should have aria-haspopup, aria-controls and aria-expanded attributes (I would expand it as soon as it's focused)
- you can probably use aria-describedby attribute here to point at the info-option header and body
- info-option-header should have header semantics.

dmose:
This gets most of the suggestions in fix #2997.  We'll have to spin off a bug to do the aria-describedby work, as well as getting better focus stuff for 57 for the expanded info option.

### DIFF
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -17,9 +17,12 @@ class Section extends React.Component {
   }
 
   onInfoLeave(event) {
+    // If we have a related target, check to see if it is within the current
+    // target (section-info-option) to keep infoActive true. False otherwise.
     this.setState({
-      infoActive: event && event.relatedTarget && (event.relatedTarget.compareDocumentPosition(event.currentTarget) &
-        Node.DOCUMENT_POSITION_CONTAINS)
+      infoActive: event && event.relatedTarget && (
+        event.relatedTarget.compareDocumentPosition(event.currentTarget) &
+          Node.DOCUMENT_POSITION_CONTAINS)
     });
   }
 
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -1,25 +1,56 @@
 const React = require("react");
 const {connect} = require("react-redux");
-const {FormattedMessage} = require("react-intl");
+const {injectIntl, FormattedMessage} = require("react-intl");
 const Card = require("content-src/components/Card/Card");
 const Topics = require("content-src/components/Topics/Topics");
 
 class Section extends React.Component {
+  constructor(props) {
+    super(props);
+    this.onMouseOverSectionInfoOption =
+      this.onMouseOverSectionInfoOption.bind(this);
+    this.onMouseOutSectionInfoOption =
+      this.onMouseOutSectionInfoOption.bind(this);
+    this.state = {isHovering: false};
+  }
+
+  onMouseOverSectionInfoOption() {
+    this.setState({isHovering: true});
+  }
+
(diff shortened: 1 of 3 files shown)


system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -1,25 +1,56 @@
...
+
   render() {
-    const {id, eventSource, title, icon, rows, infoOption, emptyState, dispatch, maxCards, contextMenuOptions} = this.props;
+    const {id, eventSource, title, icon, rows, infoOption, emptyState, dispatch, maxCards, contextMenuOptions, intl} = this.props;
     const initialized = rows && rows.length > 0;
     const shouldShowTopics = (id === "TopStories" && this.props.topics && this.props.read_more_endpoint);
+
+    const infoOptionIconA11yAttrs = {
+      "aria-haspopup": "true",
+      "aria-controls": "info-option",
```
yzen: Ah so the unfortunate part of the aria-spec (IMO) is that attributes for relations such as aria-controls, aria-describedby etc only take a list of 1 or more id refs. Class names unfortunately do not work

yzen:
Sorry if I missed it, I think one thing that's missing is a handler of a focus event that the icon would receive (when keyboard user tabs to it). I guess, we would want to do that same expansion that happens on hover. Next tab should focus on the link ensuring that the info-section does not get collapsed. The tab away from a link should probably collapse it?

system-addon/content-src/components/Sections/Sections.jsx
dmose: yeah, there appears to be some workarounds for this in react, i'll file a spin-off bug to deal with that by 57.

dmose:
@yzen The focus stuff won't get done tonight.  I can either land this patch as-is/with minimal tweaks for 56, or wave it off until 57.

Mardak:
activity-stream info tabfocus

Mardak:
Github doesn't like my webm so here's a tiny gif..
It's hard to tell but I click Pocket text, tab to info, tab to feedback, tab past to story, shift-tab to info, tab to feedback, tab to story, shift-tab to info, shift-tab to twitter context menu.
screenflow webmhd

system-addon/test/unit/content-src/components/Sections.test.jsx
```diff
@@ -12,14 +14,82 @@ describe("<Sections>", () => {
...
+    assert.lengthOf(wrapper.find('.info-option-icon[aria-expanded="true"]'), 1);
+  });
+
+  it('should render info-option-icon w/aria-expanded["false"] when moused out', () => {
+    const wrapper = shallowWithIntl(<Section {...FAKE_SECTION} />);
+    wrapper.find(".section-info-option").simulate("mouseover");
+
+    wrapper.find(".section-info-option").simulate("mouseout");
+
+    assert.lengthOf(wrapper.find('.info-option-icon[aria-expanded="false"]'), 1);
```
Mardak: Thanks for the tests! Made it easy to refactor!

as-pine-proxy:
[Treeherder: 9b5b1cc94409ef003cd1d263e5fe815e14374c8b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=9b5b1cc94409ef003cd1d263e5fe815e14374c8b)

Issue #3071, opened 2017-08-02
Activity Stream crashes in latest autoland 56 Nightly build


tspurway:
This pertains to @Mardak 's *special* build latest / landing Activity Stream changes for 56:

https://queue.taskcluster.net/v1/task/Y071RMJfQUOeU6DRY8lKsw/runs/0/artifacts/public/build/target.dmg

The build works well for a fresh profile, but when opened on my big 'ol profile, it refuses to display my Top Sites, complaining about `title is undefined` in `activity-stream.bundle.js:738:7`.  Subsequent newtabs are blank.

Screenshot and stack dump:

<img width="838" alt="image" src="https://user-images.githubusercontent.com/25276/28879561-779ec792-7770-11e7-91ac-f980e3d28d40.png">

```
Content msg: 
Object { name: "ActivityStream:MainToContent", data: {…} }
 Dispatch error:  
TypeError: title is undefined
Stack trace:
render@resource://activity-stream/data/content/activity-stream.bundle.js:2334:1
_renderValidatedComponentWithoutOwnerOrContext@resource://activity-stream/vendor/react-dom.js:13:10748
_renderValidatedComponent@resource://activity-stream/vendor/react-dom.js:13:10872
performInitialMount@resource://activity-stream/vendor/react-dom.js:13:6691
mountComponent@resource://activity-stream/vendor/react-dom.js:13:5737
mountComponent@resource://activity-stream/vendor/react-dom.js:14:18731
updateChildren@resource://activity-stream/vendor/react-dom.js:13:2956
_reconcilerUpdateChildren@resource://activity-stream/vendor/react-dom.js:14:15091
_updateChildren@resource://activity-stream/vendor/react-dom.js:14:15853
updateChildren@resource://activity-stream/vendor/react-dom.js:14:15751
_updateDOMChildren@resource://activity-stream/vendor/react-dom.js:13:21826
updateComponent@resource://activity-stream/vendor/react-dom.js:13:20050
receiveComponent@resource://activity-stream/vendor/react-dom.js:13:19603
receiveComponent@resource://activity-stream/vendor/react-dom.js:14:19125
updateChildren@resource://activity-stream/vendor/react-dom.js:13:2837
_reconcilerUpdateChildren@resource://activity-stream/vendor/react-dom.js:14:15091
_updateChildren@resource://activity-stream/vendor/react-dom.js:14:15853
updateChildren@resource://activity-stream/vendor/react-dom.js:14:15751
[...]

PR #3073, opened 2017-08-02
Add hostname to pinned links.

piatra:
Should fix https://github.com/mozilla/activity-stream/issues/3071
What I think is happening:
- When we call [`insertPinned`](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/TopSitesFeed.jsm#L74) we call it with `pinned` which do not have the hostname property. If it happens that the pinTitle is missing then we end up with `title = undefined`.


as-pine-proxy:
[Treeherder: 6b6ca6276aa8a27c4afe4037cddb424031a0121c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6b6ca6276aa8a27c4afe4037cddb424031a0121c)

PR #3081, opened 2017-08-03
fix(systemaddon): Limit the number of topsites being rendered.

### LINKED ISSUE #2945: My top sites are tripping out .......
So I had no pinned top sites, and then I pinned a top site and all of a sudden: 
screen shot 2017-07-24 at 2 58 01 pm

And then the extra 3 wouldn't go away no matter how many times I opened a new tab......

cc @dmose @k88hudson @Mardak @rlr any idea what could be going on here?

piatra:
Closes #2945.
Cause of bug seems more of an edge case: if you switch from AS to Tiles you could get different sites presented to you and pin > 12 sites. I managed to get 14 pinned sites on my profile. These show up on `PINNED_SITES_UPDATED` actions because the reducer does not `slice` the results.
I moved the slice in the actual component this way we could catch the issue if it happens from other sources as well.
It should not be a problem for people coming over from Tiles because they only had at most 12 tiles.

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -5,6 +5,8 @@
 
 const {actionTypes: at} = Components.utils.import("resource://activity-stream/common/Actions.jsm", {});
 
+const TOP_SITES_SHOWMORE_LENGTH = 12;
+
 const INITIAL_STATE = {
   App: {
     // Have we received real data from the app yet?
@@ -139,7 +141,7 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       return Object.assign({}, prevState, {rows: newRows});
     case at.PINNED_SITES_UPDATED:
       pinned = action.data;
-      newRows = insertPinned(prevState.rows, pinned);
+      newRows = insertPinned(prevState.rows, pinned).slice(0, TOP_SITES_SHOWMORE_LENGTH);
       return Object.assign({}, prevState, {rows: newRows});
     default:
       return prevState;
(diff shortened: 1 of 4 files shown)


Mardak:
Any reason not to do this from the reducer? I also see there is basically the same `.slice-12` logic in `TopSitesFeed` which would be annoying to make sure they're kept in sync.

piatra:
@Mardak agreed. I initially tried to do this in the reducer and import the 12 constant from TopSitesFeed but that results in a circular dependency. So I decided it made more sense (if I have to duplicate the constant) to do it in the component. Should I create a constants file for this?

Mardak:
@k88hudson any suggestions on this? If we just put the 12 `const` *only* in `Reducer` it would result in more-than-necessary work in `TopSitesFeed.refresh` and in the passing around of data for `TOP_SITES_UPDATED`.

Perhaps a step back, the reason why the extra pinned sites appear is that there's a `PINNED_SITES_UPDATED` that updates the `TopSites` state. And now taking a closer look at `TopSitesFeed._getPinnedWithData`, it seems like there could be logic from `refresh` that should be there too? Maybe instead of `PINNED_SITES_UPDATED` we just call `refresh` to dispatch `TOP_SITES_UPDATED` ?

k88hudson:
The only disadvantage to handling `PINNED_SITES_UPDATED` with a refresh would a lag on slower computers, which is probably something we don't want;

In terms of sharing a constant, why not just define it in `Reducers.jsm`?

Mardak:
After some chatting with @rlr on irc. It sounds like `refresh` should be called infrequently probably to avoid re-querying SQL via `NewTabUtils.activityStreamLinks.getTopSites()`. And there is existing refresh-once-per-15-minutes logic on `NEW_TAB_LOAD`. If we are *only* concerned about the SQL overhead, we can cache it locally, e.g.,

```js
async getFrecent(force) {
  if (force || 15 minutes passed) {
    this._frecent = await NewTabUtils.activityStreamLinks.getTopSites()
  }
  return this._frecent;
}
async getLinksWithDefaults(force) {
  let frecent = await this.getFrecent(force);
}
```

Mardak:
I suppose frequent `refresh`es also trigger more broadcasting..

piatra:
@Mardak updated.

as-pine-proxy:
[Treeherder: 67dfebe927b739952eac628db49ecb70d14f5624](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=67dfebe927b739952eac628db49ecb70d14f5624)

PR #3088, opened 2017-08-03
fix (systemaddon): #3058 refresh TopSitesFeed when TippyTopProvider init is complete

### LINKED ISSUE #3058: Wait for TippyTopProvider to finish processing manifest before processing sites
Split out from #3049 with a couple related issues:

https://github.com/mozilla/activity-stream/pull/3049#discussion_r130511470
> @k88hudson any thoughts on having this TippyTopProvider vs a TippyTopFeed? If we do want a feed, we should just refactor it later. I could see the feed dispatching a mapping of domains to urls, but then again, would we want all that data going to stores and the content process?

https://github.com/mozilla/activity-stream/pull/3049#discussion_r130511147
> So, this seems like it could fail or give wrong results if TippyTopProvider hasn't finished adding to _sitesByDomain. I would guess the correct fix is to make processSite async. But we can skip this for now.
> Yeah, figured it should just should just move on if it isn't fully init'd instead of waiting. I guess we would want to dispatch an action when init is done to refresh again. Maybe there should be a tippy top feed like you suggested 🤔 
> Well, Provider doesn't have the usual dispatch, but it async init returns a promise that TopSitesFeed could use to have refresh wait until Provider is `init`ed . 
Although, it wouldn't allow for the "just move on" if that's desired. E.g.,
```js
TopSitesFeed {
constructor() {
  const provider = new TippyTopProvider();
  this._initedProvider = provider.init(); // init needs to resolve the provider
}
async refresh() {
  …
  (await this._initedProvider).processSite(link);
```

rlr:
Fixes #3058 

r? @Mardak

### DIFF
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -26,7 +26,6 @@ this.TopSitesFeed = class TopSitesFeed {
   constructor() {
     this.lastUpdated = 0;
     this._tippyTopProvider = new TippyTopProvider();
-    this._tippyTopProvider.init();
     this.dedupe = new Dedupe(this._dedupeKey);
   }
   _dedupeKey(site) {
@@ -137,8 +136,12 @@ this.TopSitesFeed = class TopSitesFeed {
       data: this._getPinnedWithData()
     }));
   }
-  onAction(action) {
+  async onAction(action) {
     switch (action.type) {
+      case at.INIT:
+        await this._tippyTopProvider.init();
+        this.refresh();
+        break;
       case at.NEW_TAB_LOAD:
         if (
           // When a new tab is opened, if the last time we refreshed the data
--- a/system-addon/test/unit/lib/TopSitesFeed.test.js
+++ b/system-addon/test/unit/lib/TopSitesFeed.test.js
@@ -10,7 +10,7 @@ const FAKE_SCREENSHOT = "data123";
 
 function FakeTippyTopProvider() {}
 FakeTippyTopProvider.prototype = {
-  init() {},
+  async init() {},
   processSite(site) { return site; }
 };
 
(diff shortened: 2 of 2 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -137,8 +136,12 @@ this.TopSitesFeed = class TopSitesFeed {
       data: this._getPinnedWithData()
     }));
   }
-  onAction(action) {
+  async onAction(action) {
     switch (action.type) {
+      case at.INIT:
```
k88hudson: This could still fail due to a race condition, because `NEW_TAB_LOAD` could be fired before `this._tippyTopProvider` is done initializing. I think it might be better to do something like this inside `refresh`:

```js
if (!this._tippyTopProvider.initialized) await this._tippyTopProvider.init()
// continue refreshing
```

rlr: yeah, I was discussing with @Mardak on IRC: "so adding this refresh means it could refresh earlier than necessary or result in a second refresh after a first NEW_TAB_LOAD"

But we figured that might be better than holding up the refresh when a `NEW_TAB_LOAD` event comes in. Don't you think?

Mardak: @rlr and I did chat about this and decided that if we do have a `NEW_TAB_LOAD` it seemed undesirable for it to be blocked on a potentially unbounded amount of time for `tippyTopProvider.init`

Mardak: And on the flip side, if tippyTop does resolve sooner than `NEW_TAB_LOAD`, the state will already be updated and not refreshed due to the 15-minute check.

Mardak: Also, if we block `NEW_TAB_LOAD`, the section would be collapsed until there were sites. (Although that early, other things aren't quite loaded yet either including strings !)

Check out this 10x slowed-down video of a debug build with this patch (and my load-correctly-on-startup patch):
https://ed.agadak.net/as/activity-stream.startup.10x.webm

rlr: 👀 Cool video. How do you do that?

Mardak: I used Screenflow, but any screen recording software with editing should be good.

Mardak: I suppose one downside of the "go ahead without waiting" is that it'll trigger fetching screenshots.

as-pine-proxy:
[Treeherder: 0fa234ac5b71329660bc684f19629324fcc254cc](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0fa234ac5b71329660bc684f19629324fcc254cc)

PR #3091, opened 2017-08-04
fix(systemaddon): Remove pinTitle from sites, currently not being used.

### LINKED ISSUE #3080: Pinned sites imported from Tiles use page title instead of hostname
This happens for pinned websites coming over from tiles. Removing the pin/adding it back fixes it.
screen shot 2017-08-03 at 12 30 27
screen shot 2017-08-03 at 12 32 28

piatra:
Closes #3080.
When switching your pinned sites from tiles we would store `pinTitle = site.title` which is not what we want. Currently pinTitle is not being used we just ported from Test Pilot side so I went ahead and removed it completely.

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -67,7 +67,6 @@ function insertPinned(links, pinned) {
   newLinks = newLinks.map(link => {
     if (link && link.isPinned) {
       delete link.isPinned;
-      delete link.pinTitle;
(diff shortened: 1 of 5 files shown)


as-pine-proxy:
[Treeherder: daab53637496bbc964fda70fb568470d1bd0f715](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=daab53637496bbc964fda70fb568470d1bd0f715)

PR #3093, opened 2017-08-04
fix(systemaddon): remove title from pinned site pref value

rlr:
This is a followup to pull #3091

Since we aren't using the `pinTitle` right now, we don't need to save it. We will probably need it back when we implement customization, but we should call it `label` or something less confusing at that point.

r? @piatra


piatra:
That's true. They get their title through the shortURL(title) no need to pass that around. r+

as-pine-proxy:
[Treeherder: 2a1a11e28f6d6e91242c8659e7a283c9d7783692](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2a1a11e28f6d6e91242c8659e7a283c9d7783692)

Issue #3007, opened 2017-07-28
Snippets: Read telemetryEnabled from Services.telemetry.canRecordBase.


glogiotatidis:
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1384843
Based on suggestion given by  :gfritzsche in [Bug 1384843](https://bugzilla.mozilla.org/show_bug.cgi?id=1384843) we should get `ervices.telemetry.canRecordBase` instead of directly reading prefs.

/cc @k88hudson

dmose:
We're going to want this for the rest of our telemetry too, I think.  We should grep the code base for other users of telemetry-related prefs.

k88hudson:
Ok thanks!

PR #3095, opened 2017-08-04
Revert #3007 use Services.telemetry.canRecordBase for snippets telemetry

k88hudson:
This reverts commit aeee212d3f0c4ede4647e065af8a328ca6f5bc11.

After some discussion with gfritzsche, it looks like we need to use the datareporting.healthreport.uploadEnabled pref (which refers to the uploading of data) not .canRecordBase, which refers to the recording of data.

### DIFF
--- a/system-addon/lib/SnippetsFeed.jsm
+++ b/system-addon/lib/SnippetsFeed.jsm
@@ -14,6 +14,7 @@ XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge",
 
 // Url to fetch snippets, in the urlFormatter service format.
 const SNIPPETS_URL_PREF = "browser.aboutHomeSnippets.updateUrl";
+const TELEMETRY_PREF = "datareporting.healthreport.uploadEnabled";
 const ONBOARDING_FINISHED_PREF = "browser.onboarding.notification.finished";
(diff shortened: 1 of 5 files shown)


as-pine-proxy:
[Treeherder: fd4763cb45cd4fa3f8fcdc1d5846c4435edc3934](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=fd4763cb45cd4fa3f8fcdc1d5846c4435edc3934)

dmose:
We're going to need to do an uplift to 56 that includes this patch.  What's the right way to go about that?  Cut a branch for 56 and export to beta from there?  CC @Mardak

Mardak:
We could. Depends on how many things we're expecting to uplift to 56. Just making sure, this *really* is needed for the shield study?

dmose:
@Mardak, in the interest of being consistent with our data policy, I think we should.  It turns out that canRecord* is just about whether Fx is allowed to record telemetry, and that that's often set to true even when the upload preference is set to false.  :-(

Mardak:
Uplifted to firefox-56 branch https://github.com/mozilla/activity-stream/commit/5aa2666eef1ba8e959071351dea015d173b37bf6

PR #3097, opened 2017-08-04
fix(testing): make prepare-mochitest-build work even if no patches

dmose:
The pine script blew up this morning because we don't currently have any patches in the mozilla-central-patches directory.  Here's a fix.


as-pine-proxy:
[Treeherder: 76b0aee361440744346ec497bb233a02219e2e4c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=76b0aee361440744346ec497bb233a02219e2e4c)

PR #3108, opened 2017-08-07
Topsites telemetry fixes (#3105)

### LINKED ISSUE #3105: hero element timestamp sent too many times
So I found the bug that's causing the the extra topsites_first_painted_ts in https://bugzilla.mozilla.org/show_bug.cgi?id=1387852

https://github.com/mozilla/activity-stream/blob/3b2f4a3903d2c16cf76beeafbc39d1482b8764ab/system-addon/content-src/components/TopSites/TopSites.jsx#L133-L169

What's happening is that _maybeSendPaintedEvent is relying on this._timestampSent to guard against sending events for all 12 screenshots as they come in, one after another.  Unfortunately, there's a race here: _sendPaintedEvent only set this._timestampSent on the next frame.  Which means that if a bunch of the screenshots come in before the next frame is painted, they'll all get sent too.

We will likely want to uplift this fix to 56 so that we get useful topsites_first_painted_ts data from the shield study.

dmose:
This does two things: 

* avoids extra marks that corrupt topsites_first_painted_ts fix #3105 
* switches to using RFA + Promise instead of double RFA to make painting times more precise.

The part that is easily testable is the extra marks: 

1. start up the browser
2. start the profiler
3. open a new tab
4. capture the profile
5. In the profile, switch to the Markers view
6. Enter "User Timing" in the filter textbox
7. Go through the main and all content processes
8. There should be one, or at most two, topsites_first_painted_ts

There should be exactly two topsites_first_painted_ts.

r? Mardak

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -105,7 +105,7 @@ class TopSitesPerfTimer extends React.Component {
     this.perfSvc = this.props.perfSvc || perfSvc;
 
     this._sendPaintedEvent = this._sendPaintedEvent.bind(this);
-    this._timestampSent = false;
+    this._timestampHandled = false;
   }
 
   componentDidMount() {
@@ -131,11 +131,6 @@ class TopSitesPerfTimer extends React.Component {
   }
 
   _maybeSendPaintedEvent() {
-    // If we've already saved a timestamp for this session, don't do so again.
-    if (this._timestampSent) {
-      return;
-    }
-
     // We don't want this to ever happen, but sometimes it does.  And when it
     // does (typically on the first newtab at startup time calling
     // componentDidMount), the paint(s) we care about will be later (eg
@@ -144,6 +139,19 @@ class TopSitesPerfTimer extends React.Component {
       // XXX should send bad event
       return;
     }
+
+    // If we've already handled a timestamp, don't do it again
+    if (this._timestampHandled) {
+      return;
+    }
+
+    // And if we haven't, we're doing so now, so remember that.
+    // Even if something goes wrong in the callback, we can't
+    // try again, as we'd be sending back the wrong data, and we
+    // have to do it here, so that other calls to this method
+    // while waiting for the next frame won't also try to handle
+    // handle it.
+    this._timestampHandled = true;
 
     this._onNextFrame(this._sendPaintedEvent);
   }
@@ -162,11 +170,10 @@ class TopSitesPerfTimer extends React.Component {
     } catch (ex) {
       // If this failed, it's likely because the `privacy.resistFingerprinting`
       // pref is true.  We should at least not blow up, and should continue
-      // to set this._timestampSent to avoid going through this again.
+      // to set this._timestampHandled to avoid going through this again.
     }
-
-    this._timestampSent = true;
   }
+
   render() {
     return (<TopSites {...this.props} />);
   }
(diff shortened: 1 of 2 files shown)


system-addon/test/unit/content-src/components/TopSites.test.jsx
```diff
@@ -61,51 +61,73 @@ describe("<TopSitesPerfTimer>", () => {
...
+      this.callback = () => done();
+      sandbox.spy(this, "callback");
+      const wrapper = shallow(<TopSitesPerfTimer {...DEFAULT_PROPS} />);
+      const instance = wrapper.instance();
+
+      instance._afterFramePaint(this.callback);
+
+      assert.notCalled(this.callback);
       mockRaf.step({count: 1});
-      assert.calledOnce(callback);
```
Mardak: This was a little confusing to figure out as there's no explicit check that the callback was called except that if it wasn't called, the test would have timed out.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -145,7 +144,20 @@ class TopSitesPerfTimer extends React.Component {
...
+      return;
+    }
+
+    // And if we haven't, we're doing so now, so remember that.
+    // Even if something goes wrong in the callback, we can't
+    // try again, as we'd be sending back the wrong data, and we
+    // have to do it here, so that other calls to this method
+    // while waiting for the next frame won't also try to handle
+    // handle it.
+    this._timestampHandled = true;
```
Mardak: nit: this narrow comment block was bothering me more than it probably should ;)
```js
    // And if we haven't, we're doing so now, so remember that.  Even if
    // something goes wrong in the callback, we can't try again, as we'd be
    // sending back the wrong data, and we have to do it here, so that other
    // calls to this method while waiting for the next frame won't also try to
    // handle handle it.
```

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -117,25 +117,24 @@ class TopSitesPerfTimer extends React.Component {
   }
 
   /**
-   * Call the given callback when the subsequent animation frame
-   * (not the upcoming one) paints.
+   * Call the given callback after the upcoming frame paints.  We're using
+   * a Promise rather than a setTimeout or double rFA, as Promise resolution
+   * is faster, as of this writing (Firefox 57 Nightly) - see #3105 for
```
Mardak: probably note that it's also consistent not just faster in particular when in a not visible browser

system-addon/test/unit/content-src/components/TopSites.test.jsx
dmose: Yeah, I'll fix this.

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: I'll update the stuff referring to Promises to talk about setTimeout (and maybe postMessage).

system-addon/content-src/components/TopSites/TopSites.jsx
dmose: Yeah, I dunno how it ended up that narrow.  As we all know, 80 columns is the One True Width.  :-D

dmose:
@Mardak, can you have another quick look, since I've swapped out the promise for setTimeout...

as-pine-proxy:
[Treeherder: 9aee710671d04d0d7516f90b572c0b1666b77c63](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=9aee710671d04d0d7516f90b572c0b1666b77c63)

PR #3117, opened 2017-08-08
fix (systemaddon): #3111 dont package tippy top assets in m-c

### LINKED ISSUE #3111: Use packaged localized screenshots instead of tippy top icons
This needs to be removed before shield studies can run.

rlr:
@Mardak this seems to do the trick for now. r?

Fixes #3111

### DIFF
--- a/yamscripts.yml
+++ b/yamscripts.yml
@@ -49,7 +49,7 @@ scripts:
 
 # buildmc: Export the bootstraped add-on to mozilla central
   buildmc:
-    pre: =>cleanmc && =>tippymc
+    pre: =>cleanmc && rimraf system-addon/data/content/tippytop
     webpack: webpack --config webpack.system-addon.config.js
     css: node-sass system-addon/content-src/activity-stream.scss -o system-addon/data/content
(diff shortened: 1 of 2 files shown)


Mardak:
I sent an email trying to get an answer about if this should be just for Firefox 56 or everywhere (includes Nightly 57). This current PR is for master which would be for nightly. I just pushed a "firefox-56" branch if this change is 56-only.

https://github.com/mozilla/activity-stream/tree/firefox-56

yamscripts.yml
```diff
@@ -49,7 +49,7 @@ scripts:
 
 # buildmc: Export the bootstraped add-on to mozilla central
   buildmc:
-    pre: =>cleanmc && =>tippymc
+    pre: =>cleanmc && rimraf system-addon/data/content/tippytop
```
Mardak: The last part can just be `&& =>pretippymc`

rlr:
For 56, we will be sending hand rolled screenshots for default sites. In 57, we'll do legit/OG icons. I'm working with @bryanbell on the 56 stuff now. I'll file a PR against that branch when ready.

PR #3124, opened 2017-08-09
fix (systemaddon): #3111 use screenshots instead of icons for default sites

### LINKED ISSUE #3111: Use packaged localized screenshots instead of tippy top icons

rlr:
This only fully covers en-US default top sites. @bryanbell says to ship these and the others will come shortly.

r? @Mardak 

Fixes #3111


rlr:
Screenshot:

screen shot 2017-08-09 at 3 53 31 pm

Mardak:
Sounds like "pantsuit nation" will be removed from the facebook one

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -46,10 +46,7 @@ class TopSite extends React.Component {
     let imageStyle;
     if (tippyTopIcon) {
       imageClassName = "tippy-top-icon";
-      imageStyle = {
-        backgroundColor: link.backgroundColor,
-        backgroundImage: `url(${tippyTopIcon})`
```
Mardak: Looks like tests want this. Could just put empty string (or leave it as `undefined`) and it should work?

rlr: doh. yeah will just add a guard

rlr:
haha, @bryanbell says he has a to fix a screenshot. guessing it's that

system-addon/content-src/components/TopSites/TopSites.jsx
rlr: ah, leaving as `undefined` works fine. react ignores it and doesn't render it out 💃

rlr:
ok, I'll add a [WIP] to the title and a blocked tag we'll wait for the rest.

system-addon/data/content/tippytop/top_sites.json
```diff
@@ -0,0 +1,32 @@
+[
+  {
+    "title": "amazon",
+    "urls": ["https://www.amazon.com/", "https://www.amazon.ca/", "https://www.amazon.de/", "https://www.amazon.co.uk/", "https://www.amazon.fr/"],
```
Mardak: This won't work because the screenshots have english text in them.

system-addon/data/content/tippytop/top_sites.json
```diff
@@ -0,0 +1,32 @@
...
+    "image_url": "amazon-com@2x.png"
+  },
+  {
+    "title": "facebook",
+    "url": "https://www.facebook.com/",
+    "image_url": "facebook-com@2x.png"
```
Mardak: Not sure what we'll do for other languages that use facebook.com, but the screenshot is english.

rlr: oy, this is becoming more work than probably expected. Is the country in the redux state somewhere?

Mardak: @k88hudson Not sure how long term this change would be, but one quick hack to get geo/country code into content would be to set it as a `getValue` pref:
https://github.com/mozilla/activity-stream/blob/007d0c1340cf3b1202533037ab8ef276fe947cdf/system-addon/lib/ActivityStream.jsm#L44
```js
["geo", {
  getValue: ({geo}) => geo
}]
```

And then content can access it via Prefs from the store.

Although, don't you actually want the geo from `TopSitesFeed` or `TippyTopProvider`? Those could just directly access `Services.prefs.getStringPref("browser.search.region")`. But if we do want geo everywhere, it should probably go in `store.App` alongside `locale` ?

rlr: Good point, I would only need it from `TippyTopProvider` for this. 👍

k88hudson: It seems reasonable to access it directly here, but I wonder if we need to solve this issue more broadly for things down the line? adding it to `store.App`makes sense, although ordering would still be an issue I guess

rlr:
@k88hudson this is ready to go now. I manually verified locally that all the regions get a resource:// screenshot. final r?

k88hudson:
Ok this looks great! Let's just wait until https://github.com/mozilla/activity-stream/pull/3186 to land this 👍

as-pine-proxy:
[Treeherder: 06ee05e143cc2dd740b361f5a93a801769d9e560](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=06ee05e143cc2dd740b361f5a93a801769d9e560)

Issue #3090, opened 2017-08-03
Improve startup rendering


Mardak:
This could be partially covered by #2813, but this is specifically about first startup where various data needs to be loaded for the first time, e.g., prefs, strings, sites, stories.

Here's a full speed startup on my machine using a debug build from first window paint:
https://ed.agadak.net/as/activity-stream.startup.webm

And slowed down to 10% from first thing painted on about:newtab:
https://ed.agadak.net/as/activity-stream.startup.10x.webm

Here's a timeline of what happens in the 10x slower video (30fps):
0sec 0frame top left onboarding icon appears with bubble text
0sec 14frame bottom onboarding bar appears with left icon, text, button
0sec 19frame bottom onboarding sync image and close icon appear
3sec 10frame search box, manual migration box, top sites header appear with placeholder text
4sec 0frame recommended stories header and section appear with placeholder text
5sec 15frame scrollbar and default top sites appear with single letter and hostname below
7sec 25frame placeholder strings are updated with actual strings
8sec 10frame tippy top data changes the background color of top sites and hides single letter
10sec 0frame various icons appear (search, migration info, header) and story texts
10sec 10frame tippy top icons appear and story time icons (trending bolt, clock now)
11sec 5frame story images appear

- [x] Render after strings load #3127
- [ ] Prerendering #2813
- [ ] Fixed element positions #2791

Mardak:
@aaronrbenson suggests we should at least have skeleton / scaffolding so things don't jump around. @k88hudson might have a WIP already.

Leaving design feedback needed as there's a bunch of other things possible to address

Mardak:
@bryanbell suggests don't show anything until strings are available. For video above, this happens to hide content that would have shifted around.

Mardak:
With #3127, I made it so rendering happens only after strings are ready. Here's the current rendering video:

https://ed.agadak.net/as/activity-stream.startup.strings.10x.webm

This particular case has top sites appearing after strings (whereas the earlier video has topsites appearing before strings). So the suggestion to have the placeholder is still useful.

Mardak:
Tracking "Performance / Optimization" https://github.com/mozilla/activity-stream/projects/16

PR #3127, opened 2017-08-09
fix(init): Wait until we have strings to render/update

### LINKED ISSUE #3092: When opening a new tab, "undefined" flickers as the tab title
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1386911

Open many tabs quickly and sometimes it'll briefly show "undefined"

I believe this is because strings haven't made it to content and we're getting a visibility change event.

Mardak:
Fix #3092 and address some of #3090.

Here's how startup renders 10x slowed down:
 https://ed.agadak.net/as/activity-stream.startup.strings.10x.webm

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -14,7 +14,7 @@ const INITIAL_STATE = {
     // The locale of the browser
     locale: "",
     // Localized strings with defaults
-    strings: {},
+    strings: null,
     // The version of the system-addon
     version: null
   },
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 300fdf18735b8caea3f113c46c03cbd4b37676bf](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=300fdf18735b8caea3f113c46c03cbd4b37676bf)

PR #3128, opened 2017-08-10
fix(telemetry): #3118 workaround 500s apparent topsites_first_paint_ts

### LINKED ISSUE #3118: Better handle session start load trigger information with preloaded
The core issue right now is a preloading browser will share the marked newtab-start of a just-opened unpreloaded tab.

@dmose and I discussed some ideas of waiting for visibility event to set the load trigger. There was also some discussion of detecting if the browser is preloaded without relying on visibility status as an open background tab is "hidden" but not preloaded.

This piece of code for `ActivityStreamMessageChannel` seems to detect preloaded-ness:
```diff
   onNewTabInit(msg) {
-    this.onActionFromContent({type: at.NEW_TAB_INIT}, msg.target.portID);
+    this.onActionFromContent({
+      type: at.NEW_TAB_INIT,
+      preloaded: msg.target.browser === msg.target.browser.ownerGlobal.gBrowser._preloadedBrowser
+    }, msg.target.portID);
   }
```

However even with that information, if the user is opening up multiple tabs at the same time, the timing of things is a bit convoluted especially when the preloaded browser is loaded from a different content process. Below are the order of things happening after I open a new tab ("0th") from a fresh start causing the preloading of the "1st" browser (I removed the 0th entries):

```
NEW_TAB_INIT (1st) preloaded = true
NEW_TAB_LOAD (1st)
SAVE_SESSION_PERF_DATA (1st) topsites
```

I then opened 3 tabs very quickly as noted by the marks happening before INIT even happens:
```
marking new tab (1st)
SAVE_SESSION_PERF_DATA (1st) visible change from hidden
marking new tab (2nd)
marking new tab (3rd)
NEW_TAB_INIT (2nd) preloaded = false
NEW_TAB_INIT (3rd) preloaded = false
SAVE_SESSION_PERF_DATA (2nd) visible when loaded
NEW_TAB_LOAD (2nd)
SAVE_SESSION_PERF_DATA (3rd) visible when loaded
NEW_TAB_LOAD (3rd)
NEW_TAB_INIT (4th) preloaded = true
SAVE_SESSION_PERF_DATA (3rd) topsites
NEW_TAB_LOAD (4th)
SAVE_SESSION_PERF_DATA (2nd) topsites
SAVE_SESSION_PERF_DATA (4th) topsites
```

I then opened 3 more tabs very quickly:
```
marking new tab (4th)
SAVE_SESSION_PERF_DATA (4th) visible change from hidden
marking new tab (5th)
marking new tab (6th)
NEW_TAB_INIT (7th) preloaded = true
NEW_TAB_INIT (5th) preloaded = false
NEW_TAB_INIT (6th) preloaded = false
SAVE_SESSION_PERF_DATA (5th) visible when loaded
NEW_TAB_LOAD (5th)
SAVE_SESSION_PERF_DATA (6th) visible when loaded
NEW_TAB_LOAD (6th)
NEW_TAB_LOAD (7th)
SAVE_SESSION_PERF_DATA (5th) topsites
SAVE_SESSION_PERF_DATA (6th) topsites
SAVE_SESSION_PERF_DATA (7th) topsites
```

This second set is different from the first set in that the 7th tab (still not visible at the end of this) was preloaded into a different content process and its `INIT` happens before the 5th and 6th tabs.

If we want to set the load trigger correctly (in the cases of the user opening a new tab triggering "browser-open-newtab-start"), it looks like what we want to do is to just consume the marks in queue order whenever we get the visibility `SESSION_PERF_DATA`.

Although with a queue, making sure we match up the event to the appropriate browser will be tricky especially when we don't mark all triggering sources leading to underflow. Or overflow in cases where the user opens up new tabs without actually switching to it (cmd-shift click on the new tab button).

@dmose's ideal solution would be to have an actual mapping from new tab user action to browser.

dmose:
Fix #3118. r?Mardak  This should be the exact fix we already discussed.

### DIFF
--- a/system-addon/lib/TelemetryFeed.jsm
+++ b/system-addon/lib/TelemetryFeed.jsm
@@ -225,7 +225,6 @@ this.TelemetryFeed = class TelemetryFeed {
         break;
       case at.NEW_TAB_INIT:
         this.addSession(au.getPortIdOfSender(action));
-        this.setLoadTriggerInfo(au.getPortIdOfSender(action));
         break;
       case at.NEW_TAB_UNLOAD:
         this.endSession(au.getPortIdOfSender(action));
@@ -265,6 +264,15 @@ this.TelemetryFeed = class TelemetryFeed {
     // XXX should use try/catch and send a bad state indicator if this
     // get blows up.
     let session = this.sessions.get(port);
+
+    // Partial workaround for #3118; avoids the worst incorrect associations of
+    // times with browsers, by associating the load trigger with the visibility
+    // event as the user is most likely associating the trigger to the tab just
+    // shown. This helps avoid associateing with a preloaded browser as those
+    // don't get the event until shown. Better fix for more cases forthcoming.
+    if (data.visibility_event_rcvd_ts) {
+      this.setLoadTriggerInfo(port);
+    }
 
     Object.assign(session.perf, data);
   }
--- a/system-addon/test/unit/lib/TelemetryFeed.test.js
+++ b/system-addon/test/unit/lib/TelemetryFeed.test.js
@@ -355,7 +355,29 @@ describe("TelemetryFeed", () => {
 
       assert.include(instance.sessions.get("port123").perf, data);
     });
+
+    it("should call setLoadTriggerInfo if data has visibility_event_rcvd_ts", () => {
+      sandbox.stub(instance, "setLoadTriggerInfo");
+      instance.addSession("port123");
+      const data = {visibility_event_rcvd_ts: 444455};
+
+      instance.saveSessionPerfData("port123", data);
+
+      assert.calledOnce(instance.setLoadTriggerInfo);
+      assert.calledWithExactly(instance.setLoadTriggerInfo, "port123");
+      assert.include(instance.sessions.get("port123").perf, data);
+    });
+
+    it("shouldn't call setLoadTriggerInfo if data has no visibility_event_rcvd_ts", () => {
+      sandbox.stub(instance, "setLoadTriggerInfo");
+      instance.addSession("port123");
+
+      instance.saveSessionPerfData("port123", {monkeys_ts: 444455});
+
+      assert.notCalled(instance.setLoadTriggerInfo);
+    });
   });
+
   describe("#uninit", () => {
     it("should call .telemetrySender.uninit", () => {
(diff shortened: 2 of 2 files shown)


system-addon/lib/TelemetryFeed.jsm
```diff
@@ -266,6 +265,13 @@ this.TelemetryFeed = class TelemetryFeed {
     // get blows up.
     let session = this.sessions.get(port);
 
+    // Partial workaround for #3118; avoids the worst incorrect associations of
+    // times with browsers, and should result in no longer having apparent load
+    // times of 500,000 ms.  Better fix for more cases forthcoming.
```
Mardak: The 500sec comment doesn't really make any sense by itself. Probably better to just say to associate the load trigger with the visibility event as the user is most likely associating the trigger to the tab just shown. Could maybe note that this is to not associate with a preloaded browser as those don't get the event until shown.

system-addon/test/unit/lib/TelemetryFeed.test.js
```diff
@@ -355,7 +355,20 @@ describe("TelemetryFeed", () => {
...
+      sandbox.stub(instance, "setLoadTriggerInfo");
+      instance.addSession("port123");
+      const data = {visibility_event_rcvd_ts: 444455};
+
+      instance.saveSessionPerfData("port123", data);
+
+      assert.calledOnce(instance.setLoadTriggerInfo);
+      assert.calledWithExactly(instance.setLoadTriggerInfo, "port123");
+      assert.include(instance.sessions.get("port123").perf, data);
+    });
```
Mardak: A negative test would be good sanity check.

system-addon/lib/TelemetryFeed.jsm
dmose: Good catch, the 500sec was pretty context-free.  :-)

system-addon/test/unit/lib/TelemetryFeed.test.js
dmose: Good idea; done.

as-pine-proxy:
[Treeherder: cb1f4dd2146591771548794ef5c31387c97ae2a4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=cb1f4dd2146591771548794ef5c31387c97ae2a4)

PR #3129, opened 2017-08-10
fix(styles): Use message-box for font (including sizing) for Top Sites title

### LINKED ISSUE #2975: Increase size of top site labels
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1384052

"the labels still exist as a hint besides the icons, and that hint needs to be accessible unless it's considered redundant, in which case we should probably remove the labels rather than making them illegibly small."

Mardak:
Fix #2975. r?@sarracini This makes the font bigger on Linux matching a <tab>'s label. The change doesn't affect the look on Mac.

Double checking with @bryanbell that this screenshot looks okay:
screen shot 2017-08-10 at 7 05 48 am

### DIFF
--- a/system-addon/content-src/components/TopSites/_TopSites.scss
+++ b/system-addon/content-src/components/TopSites/_TopSites.scss
@@ -133,10 +133,10 @@
     }
 
     .title {
+      font: message-box;
       height: $top-sites-title-height;
       line-height: $top-sites-title-height;
       text-align: center;
-      font-size: 11px;
       width: $top-sites-size;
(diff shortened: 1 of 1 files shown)


Mardak:
bryan said good!

as-pine-proxy:
[Treeherder: 80db985c4a912e04eafb36d3cd197def720baeac](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=80db985c4a912e04eafb36d3cd197def720baeac)

Issue #3139, opened 2017-08-10
Generalize TelemetrySender into PingCentre


emtwo:
This includes the following tasks:

- [x] Rename TelemetrySender to PingCentre
- [x] Take 'topic' as required param and 'pingEndpoint' as optional param for PingCentre
- [x] Append client ID inside PingCentre
- [x] Separate ping centre telemetry prefs
- [x] Allow toggling stage/prod/a-s endpoints through PingCentre

What this does **not** include but needs to be done in m-c:
- [x] Move PingCentre to browser/modules
- [ ] Port tests from karma to xpcshell
- [x] Set default values for ping centre prefs.

emtwo:
https://github.com/mozilla/activity-stream/commit/cfa5260d22b1db2908189367d55aae4475826fd5

PR #3140, opened 2017-08-10
feat(system-addon): #3139 Generalize TelemetrySender into PingCentre

emtwo:


emtwo:
Note: I removed the default pref settings from ActivityStream.jsm. I believe the more appropriate place for them might be inside browser/app/profile/firefox.js instead where browser pref defaults are set. But that will likely be a separate PR that blocks this one.

system-addon/lib/ActivityStream.jsm
```diff
@@ -81,16 +81,6 @@ const PREFS_CONFIG = new Map([
     title: "Show the Top Sites section on the New Tab page",
     value: true
   }],
-  ["telemetry", {
```
k88hudson: Are we sure we want to make this a global setting for all ping centre clients? Wouldn't we want to control what activity stream does separately?

system-addon/lib/PingCentre.jsm
```diff
@@ -0,0 +1,149 @@
...
+Cu.importGlobalProperties(["fetch"]);
+
+XPCOMUtils.defineLazyModuleGetter(this, "ClientID",
+  "resource://gre/modules/ClientID.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "console",
+  "resource://gre/modules/Console.jsm");
+
+const PREF_BRANCH = "browser.ping-centre.telemetry.";
+
+const TELEMETRY_PREF = `${PREF_BRANCH}telemetry`;
```
k88hudson: You could also create a pref branch to handle these in either of these two ways:

```js
const prefs = Services.prefs.getDefaultBranch(PREF_BRANCH);
// or
const prefs = new Preferences({branch: PREF_BRANCH});
```

system-addon/lib/PingCentre.jsm
```diff
@@ -0,0 +1,149 @@
...
+const LOGGING_PREF = `${PREF_BRANCH}log`;
+const IS_PRODUCTION_PREF = `${PREF_BRANCH}production`;
```
k88hudson: Instead of using this strategy, maybe something like this would be easier?
- define some endpoint(s) for different builds (nightly, release, beta). You can check what kind of built it is with [this file](http://searchfox.org/mozilla-central/source/toolkit/modules/AppConstants.jsm)
- take an optional override pref name

system-addon/lib/PingCentre.jsm
emtwo: I didn't do this because PingCentre also references `FHR_UPLOAD_ENABLED_PREF` from its `this._prefs` object so it would need to access its different prefs in different ways which seemed more verbose to me. Let me know if you disagree.

emtwo:
@k88hudson, I added the code to export to browser/modules when running `npm run buildmc` and I addressed your comments. Here's a link to the associated bugzilla bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1390249

Let me know if there's anything missing. Thanks!

r?

k88hudson:
I still have one more PR to review / merge for the uplift, but I promise I will get to this today

system-addon/lib/PingCentre.jsm
```diff
@@ -0,0 +1,147 @@
...
+    Object.defineProperty(this, "telemetryClientId", {value: ClientID.getClientID()});
+    return this.telemetryClientId;
+  }
+
+  get enabled() {
+    return this._enabled && this._fhrEnabled;
+  }
+
+  _setPingEndpoint(topic, overrideEndpointPref) {
+    this._pingEndpoint = this._prefs.get(overrideEndpointPref) ||
```
k88hudson: I would maybe do

```js
const overrideValue = overrideEndpointPref && this._prefs.get(overrideEndpointPref);
if (overrideValue) {
   this._pingEndpoint = overrideValue;
} else if (AppConstants.MOZ_UPDATE_CHANNEL === "release") {
...
} else {
...
}
```
```

File: system-addon/lib/PingCentre.jsm [hunk 5]
```diff
@@ -0,0 +1,147 @@
...
+ *                  their data on the server side.
+ * @param {string} pingEndpoint - optional pref for URL where the POST is sent.
+ * @param {Object} args - optional arguments
+ * @param {Function} args.prefInitHook - if present, will be called back
+ *                   inside the Prefs constructor. Typically used from tests
+ *                   to save off a pointer to a fake Prefs instance so that
+ *                   stubs and spies can be inspected by the test code.
+ */
+class PingCentre {
+  constructor(topic, overrideEndpointPref, args) {
```
k88hudson: I'd recommend making the first two args (`topic` and `overrideEndpointPref`) an options object maybe, in case you want to start adding more options (it will also make usages a little clearer to read)

yamscripts.yml
```diff
@@ -55,6 +55,8 @@ scripts:
     locales: pontoon-to-json --src locales --dest system-addon/data
     # Copy over all of the system-addon directory
     copy: cpx "system-addon/**/{,.}*" ../mozilla-central/browser/extensions/activity-stream
+    copyPingCentre: cpx "system-addon/lib/PingCentre.jsm" ../mozilla-central/browser/modules
```
k88hudson: Can you add a task to `startmc` as well

yamscripts.yml
```diff
@@ -55,6 +55,8 @@ scripts:
...
+    rmDupePingCentre: rimraf "../mozilla-central/browser/extensions/activity-stream/lib/PingCentre.jsm"
```
k88hudson: Instead of doing this, I think it might be a better idea to put `PingCenter` in its own folder at the top level `i.e. ping-centre/PingCentre.jsm`

emtwo:
@k88hudson I've moved over to Services.prefs by using the `FakePrefs` object and adding to it the functions called by `Services.prefs`. Alternatively this could be a different object entirely. Let me know if you had something different in mind.

system-addon/ping-centre/PingCentre.jsm
```diff
@@ -0,0 +1,140 @@
...
+ * @param {string} options.overrideEndpointPref - optional pref for URL where the POST is sent.
+ */
+class PingCentre {
+  constructor(options) {
+    if (!options.topic) {
+      throw new Error("Must specify topic.");
+    }
+
+    this._topic = options.topic;
+    this._prefs = Services.prefs.getBranch("");
```
k88hudson: Out of curiosity why not `getBranch(PREF_BRANCH)`?

emtwo: this._prefs is being used to access `FHR_UPLOAD_ENABLED_PREF` which is not prepended with `PREF_BRANCH`. Thought it was easier this way to access all from one object.

system-addon/ping-centre/PingCentre.jsm
```diff
@@ -0,0 +1,140 @@
...
+XPCOMUtils.defineLazyModuleGetter(this, "ClientID",
+  "resource://gre/modules/ClientID.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "console",
+  "resource://gre/modules/Console.jsm");
+
+const PREF_BRANCH = "browser.ping-centre.";
+
+const TELEMETRY_PREF = `${PREF_BRANCH}telemetry`;
+const LOGGING_PREF = `${PREF_BRANCH}log`;
+const STAGING_ENDPOINT_PREF = `${PREF_BRANCH}staging.endpoint`;
```
k88hudson: If you want to store the endpoint in a pref, you might as well just have one pref and remove the channel detection in this file (since you can just do an `if` statement in app/profile/firefox.js)

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -305,8 +302,10 @@ this.TelemetryFeed = class TelemetryFeed {
     );
   }
 
-  async sendEvent(eventPromise) {
-    this.telemetrySender.sendPing(await eventPromise);
+  async sendEvent(event_object) {
+    if (this._prefs.get("telemetry")) {
```
k88hudson: Hm, it might not be a good idea to check this pref on every ping – I think you might want to cache this value and add a pref listener for it the way you do in `PingCentre`

k88hudson:
Adding the blocked label since this needs to land in MC first

system-addon/ping-centre/PingCentre.jsm
emtwo: As discussed in IRC, we can't really do this since there is no variable for release only in app/profile/firefox.js

sarracini:
https://hg.mozilla.org/mozilla-central/rev/944ab211fc31  i think this is unblocked now :)

emtwo:
Note: When this goes into m-c, we'll also need to land a patch for this: https://bugzilla.mozilla.org/show_bug.cgi?id=1393604 so that we don't cause failures. I'll add the patch into that bug.

I don't know how you guys

Issue #3076, opened 2017-08-02
Update Top story spacing to match design


ideashower:
![pasted image at 2017_08_02 11_39 am](https://user-images.githubusercontent.com/188963/28891262-c18e78bc-777e-11e7-801d-69b186fdaa82.png)

Final design:
https://mozilla.invisionapp.com/share/UWCKPZTBY#/243608608_New_Tab_-_All

The spacing on the Top Stories section could use some tightening up to match the final design. See attached image for examples, but some specific call-outs:
- Header "Recommended by Pocket" is too far away from cards
- Spacing between image and domain needs to be expanded
- Title size is too big (`14px` vs `inherit`)
- Spacing of context 
- Context string should be vertically centered with lightning bolt icon
- Popular Topics too far away from cards

//cc @aaronrbenson

Mardak:
From #3075:
> https://github.com/mozilla/activity-stream/blob/2f753b7c9be5c7398647c2518dbdf1efacd49019/system-addon/content-src/components/Card/_Card.scss#L79
> Only one that inherits, should be 14px.

Mardak:
Here's an after (left) and before (right) screenshot:
![screen shot 2017-08-11 at 1 10 52 pm](https://user-images.githubusercontent.com/438537/29230196-bd911d7c-7e96-11e7-9fd7-7a8fc9f04006.png)

ideashower:
@Mardak @sarracini Sorry for delay but just seeing this. Looking at the latest screenshot @Mardak posted, it looks like there are still some outstanding issues from the initial comment, in particular:

- Title size is too big (14px vs inherit)
- Context string should be vertically centered with lightning bolt icon

Mardak:
The title font size in that screenshot is already 14px. (The updates haven't made it in to Nightly yet.)

PR #3166, opened 2017-08-14
fix(style): Minor adjustments after #3076 fix

csadilek:
Just minor tweaks after recent changes. Not sure why the ">" need another realignment but fixing it by removing code is always welcome :). Tested on Windows, too.

Before:
before

After:
after

### DIFF
--- a/system-addon/content-src/components/Sections/_Sections.scss
+++ b/system-addon/content-src/components/Sections/_Sections.scss
@@ -10,6 +10,7 @@
 
     .section-info-option {
       float: right;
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: a0d2152f07e96b95bb4fc7a21ba3bf4fe9eb2f2a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a0d2152f07e96b95bb4fc7a21ba3bf4fe9eb2f2a)

PR #3167, opened 2017-08-14
assets(system-addon): Add webextension icon

AdamHillier:
To be default icon for custom sections


Mardak:
Looks like the only difference is the fill="context-fill" vs "#4d4d4d" is that intended? If they are supposed to be the same, then we might get packaging errors

sarracini:
@Mardak I think we've got other svg's that have fill="#4d4d4d" and it doesn't look to be a problem. Is that what you mean?

as-pine-proxy:
[Treeherder: 6837280772bbe18f235d2067226822c586af8308](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6837280772bbe18f235d2067226822c586af8308)

PR #3174, opened 2017-08-15
feat(mc): Only send one user interaction ping for each top stories article

### LINKED ISSUE #3165: Make sure it only sends one ping per article for top stories user events
This includes `click`, `dismiss`, and `save_to_pocket`.

Since currently we can't include `client_id` in the pings, we have to do some extra work in order to calculate the unique #s of those user events for Top stories.

This relates to #2898

ncloudioj:
This closes #3165

### DIFF
--- a/system-addon/lib/TelemetryFeed.jsm
+++ b/system-addon/lib/TelemetryFeed.jsm
@@ -30,10 +30,77 @@ const USER_PREFS_ENCODING = {
   "feeds.section.topstories": 1 << 2
 };
 
+const IMPRESSION_STATS_RESET_TIME = 60 * 60 * 1000; // 60 minutes
+const PREF_IMPRESSION_STATS_CLICKED = "impressionStats.clicked";
+const PREF_IMPRESSION_STATS_BLOCKED = "impressionStats.blocked";
+const PREF_IMPRESSION_STATS_POCKETED = "impressionStats.pocketed";
+
+/**
+ * A pref persistent GUID set
+ */
+class PersistentGuidSet extends Set {
+  constructor(prefs, prefName) {
+    let guids = [];
+    try {
+      guids = JSON.parse(prefs.get(prefName));
+      if (typeof guids[Symbol.iterator] !== "function") {
+        guids = [];
+        prefs.set(prefName, "[]");
+      }
+    } catch (e) {
+      Cu.reportError(e);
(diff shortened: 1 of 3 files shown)


system-addon/lib/TelemetryFeed.jsm
```diff
@@ -242,6 +309,36 @@ this.TelemetryFeed = class TelemetryFeed {
     this.telemetrySender.sendPing(await eventPromise);
   }
 
+  async handleImpressionStats(action) {
```
csadilek: I don't think we need async here. There's no promise returned or long running call that we're awaiting.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -256,8 +353,13 @@ this.TelemetryFeed = class TelemetryFeed {
       case at.SAVE_SESSION_PERF_DATA:
         this.saveSessionPerfData(au.getPortIdOfSender(action), action.data);
         break;
+      case at.SYSTEM_TICK:
+        if (Date.now() - this._impressionStatsLastReset >= IMPRESSION_STATS_RESET_TIME) {
+          this.resetImpressionStats();
+        }
```
csadilek: Is it intentional that we reset after the first system tick (5 mins and then every 60 mins) or should we initialize ```this._impressionStatsLastReset = Date.now();``` instead of ```this._impressionStatsLastReset = 0;```

system-addon/lib/TelemetryFeed.jsm
ncloudioj: Good catch!

system-addon/lib/TelemetryFeed.jsm
ncloudioj: Yes, it is intentional. However, after the second thought, I think initializing it to `Date.now()` would be preferred, since this timestamp isn't persisted in the prefs. Rebooting Firefox frequently (e.g. update restart) may introduce some undesired behaviors. Will change it to "Date.now()". Thanks!

as-pine-proxy:
[Treeherder: c8f64e7766dc83be53df0fd399b0b8c53b1f5058](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c8f64e7766dc83be53df0fd399b0b8c53b1f5058)

Issue #3078, opened 2017-08-03
More app information in gSnippetsmap


glogiotatidis:
Fields required to bring AS in par with about:home gSnippetsmap:

 - `selectedSearchEngine`: Set to the abbreviation for default search engine (e.g. `ddg`)
 - fxaccount: Set to `true` if user has setup a FxAccount, `false` if not, `undefined` if we don't have the information
 - `defaultBrowser`: Set to `true` if this the default browser for the system, `false` if it's not, `undefined` if we don't have the information

/cc @k88hudson

glogiotatidis:
fxaccount here https://github.com/mozilla/activity-stream/commit/b26da1986bb30d70695f76227b20ce3b0cd9f0e6

glogiotatidis:
😄

PR #3176, opened 2017-08-15
snippets(mc): #3078 Add defaultBrowser, selectedSearchEngine to snippets

k88hudson:


system-addon/lib/SnippetsFeed.jsm
```diff
@@ -27,12 +33,22 @@ this.SnippetsFeed = class SnippetsFeed {
...
       .replace("%STARTPAGE_VERSION%", STARTPAGE_VERSION);
     return Services.urlFormatter.formatURL(updateURL);
   }
+
+  isDefaultBrowser() {
+    try {
+      return ShellService.isDefaultBrowser();
+    } catch (e) {}
+    // istanbul ignore next
+    return null;
```
sarracini: nit: indentation is off i think. also, do we want to return false for consistency or is it necessary to return null for some other part of the code to know it went wrong?

docs/v2-system-addon/snippets.md
```diff
@@ -26,28 +26,50 @@ async function setFoo() {
...
+{
+  engines: ["searchEngine-google", "searchEngine-yahoo", "searchEngine-bing"],
+  searchEngineIdentifier : "google"
+}
+```
 
-`appData.telemetryEnabled`: Is telemetry enabled for the user?
+`appData.defaultBrowser`: (bool) Is Firefox the user's default browser?
+If we could not determine the default browser, this value is `null`. Note that
+currently this value is only checked once when the browser is initialized.
```
sarracini: how does this work? is ```_refresh``` only called on browser init?

system-addon/lib/SnippetsFeed.jsm
k88hudson: There needs to be a distinction between `true`, `false`, and we couldn't actually determine the value (this is also copying what the existing API is)

docs/v2-system-addon/snippets.md
k88hudson: Yes that's right

system-addon/lib/SnippetsFeed.jsm
k88hudson: Where's the indentation issue? the return null is outside the catch block

sarracini: whoooops my bad thought it was in the catch

as-pine-proxy:
[Treeherder: 4a12df4c93a2ff9f1127ba861178d040995de02d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4a12df4c93a2ff9f1127ba861178d040995de02d)

PR #3182, opened 2017-08-15
feat (systemaddon): #3133 edit hover menu: pin and dismiss buttons

### LINKED ISSUE #3133: Port the Top Sites pin and dismiss button (edit mode)
When in edit mode, the Top Sites show a pin and dismiss button on hover:

screen shot 2017-08-10 at 10 50 48 am

They also show an edit button, but we'll add that one as part of a separate bug that ports over top site editing.

rlr:
This adds some hover buttons for the Top Sites edit modal:
screen shot 2017-08-15 at 3 06 41 pm

Default (first run) sites only get a Pin button since we don't allow dismissing them:
screen shot 2017-08-15 at 3 07 06 pm

Fixes #3133

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -15,6 +15,8 @@ class TopSite extends React.Component {
     this.onLinkClick = this.onLinkClick.bind(this);
     this.onMenuButtonClick = this.onMenuButtonClick.bind(this);
     this.onMenuUpdate = this.onMenuUpdate.bind(this);
+    this.onDismissButtonClick = this.onDismissButtonClick.bind(this);
+    this.onPinButtonClick = this.onPinButtonClick.bind(this);
   }
   toggleContextMenu(event, index) {
     this.setState({
@@ -22,12 +24,20 @@ class TopSite extends React.Component {
       showContextMenu: true
     });
   }
-  onLinkClick() {
+  userEvent(event) {
     this.props.dispatch(ac.UserEvent({
-      event: "CLICK",
+      event,
       source: TOP_SITES_SOURCE,
       action_position: this.props.index
     }));
+  }
+  onLinkClick(ev) {
+    if (this.props.editMode) {
+      // Ignore clicks if we are in the edit modal.
+      ev.preventDefault();
+      return;
+    }
+    this.userEvent("CLICK");
   }
   onMenuButtonClick(event) {
     event.preventDefault();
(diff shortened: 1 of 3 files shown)


system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -235,7 +303,9 @@ class TopSitesEdit extends React.Component {
                   key={link.guid || link.url}
                   dispatch={this.props.dispatch}
                   link={link}
-                  index={index} />)}
+                  index={index}
+                  intl={this.props.intl}
+                  editMode={true} />)}
```
sarracini: This is out of scope for this ticket, but maybe we can file a follow up ticket to factor our the modal and some of the edit logic into it's own component so the top sites component doesn't get too gnarly. It would be nice if we only had one set of top sites being rendered at a time, and they could either be in edit mode or not, rather than rendering the default top sites and then rendering the editable top sites on top. Can you file a ticket for that?

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -5,6 +5,9 @@
   $top-sites-vertical-space: 18px;
   $screenshot-size: cover;
   $tippy-top-size: 80px;
+  $edit-menu-button-size: 25px;
+  $edit-menu-button-border: 1px solid rgba($black, 0.2);
+  $edit-menu-button-boxshadow: 0 2px 0 rgba($black, 0.1);
```
sarracini: here you can use ```$faint-black``` and ```$faintest-black``` in variables._scss

sarracini:
And I guess a rebase 😓

system-addon/content-src/components/TopSites/TopSites.jsx
rlr: Yeah, I've thought about it that little too but wasn't sure how to do it easily with the way the UI is laid out with the overlay on top of the entire page then the modal on top. But I'm sure there is a way to do it, or at least simplify it a bit if we dedicate some brainpower to it 👍

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -65,17 +106,43 @@ class TopSite extends React.Component {
...
+              onUpdate={this.onMenuUpdate}
+              options={TOP_SITES_CONTEXT_MENU_OPTIONS}
+              site={link}
+              source={TOP_SITES_SOURCE}
+              visible={isContextMenuOpen} />
+          </div>
+        }
+        {editMode &&
+          <div className="edit-menu">
+            {link.isPinned &&
```
rlr: oops, yeah I guess everything else is the same. Oh, and the intl message id but I can do the same thing.

rlr:
^^ @sarracini feedback fixor'd. Thanks for the lessons! 😅

system-addon/content-src/components/TopSites/TopSites.jsx
rlr: Filed #3189

as-pine-proxy:
[Treeherder: f75f358c995be8abf95709b671ca064ffa9df593](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f75f358c995be8abf95709b671ca064ffa9df593)

PR #3185, opened 2017-08-16
fix(sections): Conditionally set the state to avoid unnecessary renders

### LINKED ISSUE #3180: Event handlers in Sections.jsx are causing too many re-renders
Due to a major bug we found (multiple renders being caused by `setState` being called on `mouseMove` events) we need to back out this commit and uplift.

Mardak:
Fix #3180. r?@k88hudson Conditionally set state. This also fixes some unnecessary `setState` by moving the `div` closer to the `span`/icon:

image

The PR is based on a master from before the revert. I suppose a separate commit reverting the revert instead of all together could simplify uplifting.

### DIFF
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -12,18 +12,29 @@ class Section extends React.Component {
     this.state = {infoActive: false};
   }
 
+  /**
+   * Take a truthy value to conditionally change the infoActive state.
+   */
+  _setInfoState(nextActive) {
+    const infoActive = !!nextActive;
+    if (infoActive !== this.state.infoActive) {
+      this.setState(Object.assign({}, this.state, {infoActive}));
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


k88hudson:
Awesome! I added this to https://github.com/mozilla/activity-stream/pull/3186 for rebasing etc.

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -12,18 +12,29 @@ class Section extends React.Component {
     this.state = {infoActive: false};
   }
 
+  /**
+   * Take a truthy value to conditionally change the infoActive state.
+   */
+  _setInfoState(nextActive) {
+    const infoActive = !!nextActive;
+    if (infoActive !== this.state.infoActive) {
+      this.setState(Object.assign({}, this.state, {infoActive}));
```
k88hudson: The`Object.assign` isn't necessary as this is done internally in `setState

PR #3195, opened 2017-08-17
fix(css): Closes #3076 Context string is vertically aligned

### LINKED ISSUE #3076: Update Top story spacing to match design

sarracini:
I think the font size is already 14px, so the only outstanding issue left is the vertical alignment of the icon with the label. Removing this fixes that
vertical align

### DIFF
--- a/system-addon/content-src/components/Card/_Card.scss
+++ b/system-addon/content-src/components/Card/_Card.scss
@@ -114,7 +114,6 @@
     color: $light-grey;
     font-size: 11px;
     display: flex;
-    align-items: center;
   }
 
   .card-context-icon {


as-pine-proxy:
[Treeherder: fc0ddcbf1ae159d0ee256cf9b735d3e9488d41db](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=fc0ddcbf1ae159d0ee256cf9b735d3e9488d41db)

PR #3206, opened 2017-08-21
fix(channel): Always simulate an init for existing ports

### LINKED ISSUE #3163: TelemetryFeed session is undefined for early / restored new tab pages
```
session is undefined TelemetryFeed.jsm:301
	saveSessionPerfData resource://activity-stream/lib/TelemetryFeed.jsm:301:5
	onAction resource://activity-stream/lib/TelemetryFeed.jsm:257:9
```
@dmose Looks like this could happen if `addSession` isn't called. And that happens when `NEW_TAB_INIT` doesn't fire for about:newtab pages inherited by ActivityStreamMessageChannel.

https://github.com/mozilla/activity-stream/blob/0af09806a681ec61b9ed8746c72064f21c8dd8f4/system-addon/lib/ActivityStreamMessageChannel.jsm#L137-L140

I suppose either make it also fake a `onNewTabInit` or do we still need a separate `NEW_TAB_INIT` from `NEW_TAB_LOAD`? Looks like we can get visibility perf before LOAD, so that would require some cleanup... See https://github.com/mozilla/activity-stream/issues/3118

Mardak:
Fix #3163. r?@k88hudson Just call `onNewTabInit` similar to calling `onNewTabLoad` when inheriting pages.

### DIFF
--- a/system-addon/lib/ActivityStreamMessageChannel.jsm
+++ b/system-addon/lib/ActivityStreamMessageChannel.jsm
@@ -136,8 +136,10 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
 
     // Some pages might have already loaded, so we won't get the usual message
     for (const {loaded, portID} of this.channel.messagePorts) {
+      const simulatedMsg = {target: {portID}};
+      this.onNewTabInit(simulatedMsg);
       if (loaded) {
-        this.onNewTabLoad({target: {portID}});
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 07e6515ed6f8a991ac377a8d3b39327dc2b0c8ec](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=07e6515ed6f8a991ac377a8d3b39327dc2b0c8ec)

PR #3208, opened 2017-08-22
fix(systemaddon): Handle internationalized domain names. Closes #3200.

### LINKED ISSUE #3200: Punycode is shown for Internationalized Domain Name
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1391667

The non-eTLD part of `https://xn--c1asce.xn--90a3ac/` is shown when `https://гном.срб/` is a top site:
image

Looks like we want something like:
```js
Cc["@mozilla.org/network/idn-service;1"].
getService(Ci.nsIIDNService).
convertToDisplayIDN("xn--c1asce", {})
```

From https://searchfox.org/mozilla-central/source/browser/base/content/browser.js#7375-7389
https://searchfox.org/mozilla-central/source/netwerk/dns/nsIIDNService.idl#51-57

piatra:


system-addon/common/ShortURL.jsm
```diff
@@ -1,7 +1,10 @@
 Components.utils.importGlobalProperties(["URL"]);
 
+const Cc = Components.classes;
+const Ci = Components.interfaces;
```
Mardak: This seems a bit odd to put `Components` usage under `common/`. Other usage seems to have various guards or is importing an activity-stream jsm so webpack converts it. Did you look at putting the IDNService usage under `lib/` ?

Mardak: @k88hudson says to just move it into `lib/`

system-addon/common/ShortURL.jsm
```diff
@@ -15,18 +18,34 @@ Components.utils.importGlobalProperties(["URL"]);
...
  */
-this.shortURL = function shortURL(link) {
-  if (!link.url && !link.hostname) {
-    return "";
+this.ShortURL = class ShortURL {
+  constructor() {
+    this.IDNService = Cc["@mozilla.org/network/idn-service;1"].getService(Ci.nsIIDNService);
+  }
+
+  getEffectiveHost(url) {
```
Mardak: A comment block for the method and renaming the parameter to `host` would make more sense than `url`. Also, what does "effective host" mean?

system-addon/common/ShortURL.jsm
```diff
@@ -15,18 +18,34 @@ Components.utils.importGlobalProperties(["URL"]);
  *         {str} link.title (optional) - The title of the link
  * @return {str}   A short url
  */
-this.shortURL = function shortURL(link) {
-  if (!link.url && !link.hostname) {
-    return "";
+this.ShortURL = class ShortURL {
+  constructor() {
+    this.IDNService = Cc["@mozilla.org/network/idn-service;1"].getService(Ci.nsIIDNService);
```
Mardak: Unless we're adding other methods, probably simpler to just leave this `jsm` as returning the single `function` instead of a `class`. You can probably have a lazy service getter, e.g., https://github.com/mozilla/activity-stream/blob/14c051f5bbfcda584532bf4ecc7c276973a15300/system-addon/lib/Screenshots.jsm#L17-L18

Then you can just use the global `IDNService` without storing it on an instance.

piatra:
@Mardak updated. please have another look.

PR #3233, opened 2017-08-24
feat(topsites): Only show top sites that have been visited more than once

### LINKED ISSUE #2739: Filter out top sites below a certain frecency threshold
This is to handle issues when user history is too minimal to determine good top sites.

### LINKED ISSUE #2993: Should we block https://www.mozilla.org/en-US/privacy/firefox/ from topsites?
For new users, it kind of sucks that https://www.mozilla.org/en-US/privacy/firefox/ ends up as their top site. I wonder if there's a better way to prevent privacy/update related-pages from getting added?

Mardak:
Fix #2739 and fix #2993. r?@sarracini


as-pine-proxy:
[Treeherder: 6073cdcf683a7a2cb44f4f34dd74ae77bd321064](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6073cdcf683a7a2cb44f4f34dd74ae77bd321064)

Issue #2979, opened 2017-07-26
Follow photon color schemes


Mardak:
http://design.firefox.com/photon/visual/color.html

Some things are already following the guide, e.g., 
```
$info-option-text-color: #0C0C0D;
$info-option-link-color: #0A84FF;
```

We might want to rename and update references to match those of photon, e.g., `$grey-90` and `$blue-50`. Might also want to alias them to friendlier names, e.g., `$link-color`

This will probably lead to some shifting around of colors in the existing interface to match photon.

aaronrbenson:
We need to update the new tab background color to match the toolbar. Should be Grey 10 (#f9f9fa) in the design system.

AmyYLee:
Hi, 

I've attached a spec for the New Tab page with the Photon color palette applied. Please use this as reference.
https://mozilla.invisionapp.com/share/8KCY4I0QU

Mardak:
@AmyYLee there's some colors with opacities, e.g., search text is Grey-90 60% opacity. Is it intended to have it slightly transparent to blend with the background? Or should the color be the closest thing to Grey-90*60% = Grey-54 -> Grey-50, i.e., #737373, to actually match Photon color scheme?

@aaronrbenson other than the question of opacity above, anything else before someone should implement this?

For whoever implements, I believe we want to rename our variables to match the photon names, so future designs would just include the photon color name instead of the hex.

AmyYLee:
Hi, 

The opacity should be set in the alpha channel. I've also updated the spec to include card hover colour, search box color, drop-shadows, etc. Thanks!

Mardak:
All text should be 100% opacity, so calculate the closest Gray-photon-acceptable number.
Search box is special and be 60% opacity.

Mardak:
@AmyYLee do any of the secondary icons need to become 80% opacity on hover:

- info icon for pocket section (maybe?)
- trending / now story icons (probably no as not separately interactable?)
- pinned icon for top sites (probably not as clicking it doesn't toggle the icon?)
- search stuff ? do these want to go from 60% to 80% on hover
-- related, when it's active, the text color should be grey-90 100%?

AmyYLee:
do any of the secondary icons need to become 80% opacity on hover:

    info icon for pocket section (maybe?) - This can match the rollover effect in the icons in the URL bar. Also it looks like the door-hanger overlaps onto the icon. It should appear just below it.
[...]

PR #3239, opened 2017-08-25
chore(styles): Fix various followup style tweaks from design

Mardak:
Followup to Amy's comment in https://github.com/mozilla/activity-stream/issues/2979#issuecomment-325012363

The inherits don't seem totally necessary as it appears that the colors are the same.. But if you set `body {color: red}` to test, you can tell they're not using the right colors.

For info-option-icon, adjusted color to be darker on hover and tweaked the position of the appearing doorhanger as requested.

### DIFF
--- a/system-addon/content-src/activity-stream.scss
+++ b/system-addon/content-src/activity-stream.scss
@@ -85,7 +85,7 @@ a {
     background: $input-secondary;
     border: $border-primary;
     border-radius: 5px;
-    color: $grey-90;
+    color: inherit;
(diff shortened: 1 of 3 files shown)


as-pine-proxy:
[Treeherder: 65aebe02675bde03ee143bfd9cc150ca3f46ef06](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=65aebe02675bde03ee143bfd9cc150ca3f46ef06)

PR #3240, opened 2017-08-25
fix(system-addon): Fix failing functional test

AdamHillier:


system-addon/test/unit/lib/SectionsManager.test.js
```diff
@@ -226,6 +229,13 @@ describe("SectionsFeed", () => {
     });
   });
   describe("#onAction", () => {
+    it("should bind this.init to SectionsManager.INIT on INIT", () => {
+      sinon.spy(SectionsManager, "once");
+      feed.onAction({type: "INIT"});
+      assert.calledOnce(SectionsManager.once);
+      assert.calledWith(SectionsManager.once, SectionsManager.INIT, feed.init);
+      SectionsManager.once.restore();
```
k88hudson: can you add this to an `afterEach()` block? This won't get get hit if one of the assertions throw

Mardak:
What was breaking? Looks like something was being initialized too early?

AdamHillier:
We don't really know, but for some reason it looks like PREFS_INITIAL_VALUES was being received before INIT in SectionsFeed, which shouldn't be possible. It could also be that because SectionsManager is a singleton, for some reason it's not being cleaned up properly from a previous test and so incorrectly reports that it's initialised. Either case might cause the the init event in SectionsFeed to fire too early, before the store is attached. The fix just guarantees that SectionsFeed.init isn't bound to SectionsManager.INIT until after it knows for sure that the store has been attached, but there's an underlying issue that we can investigate next week. (Also, confusing terminology with lots of different "init"s)

system-addon/test/unit/lib/SectionsManager.test.js
AdamHillier: I realised that we actually don't need that line there at all as SectionsManager is re-decorated with EventEmitter already in afterEach() already

Mardak:
Looks like the underlying issue is that feeds aren't getting `uninit`ed:
https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/Store.jsm#L129
`uninitFeed` expects a string/name but is getting the object. (`Map.forEach` calls `callback(value, key, map)`)

AdamHillier:
Ah that would explain it, thanks. As a result, SectionsFeed wasn't being uninit'ed, so couldn't uninit SectionsManager, which incorrectly reported that it was already initialised during the next test.

as-pine-proxy:
[Treeherder: 2cab8338a741ab6dcbc4f5b35358893f0eaabc5c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2cab8338a741ab6dcbc4f5b35358893f0eaabc5c)

PR #3242, opened 2017-08-26
fix(init): Store an init and uninit action to give to feeds

### LINKED ISSUE #3241: Feeds don't correctly uninit
Found when tracking down #3240. `Store.uninit` is trying to uninit the feeds here:
https://github.com/mozilla/activity-stream/blob/e2a58f15bc184c40a2b5f19148e625bf8166384a/system-addon/lib/Store.jsm#L129

But is failing because `Map.forEach` provides the `value` first then `key` second, so `uninitFeed` is called with an object instead of the feed name.

However, Store never calls any `feed.init` in the first place, and some Feeds `init` when they receive `INIT`. So similarly some Feeds watch for `UNINIT` to `uninit`, but this causes problems as `uninit` will be self-called on `UNINIT` then called from `Store.uninit`.

I believe we **shouldn't** have `Store` call `uninitFeed` -> `Feed.uninit` to match the behavior of `Store` *not* calling `Feed.init`…?

@k88hudson @AdamHillier

Mardak:
Fix #3241. r?@k88hudson This removes the `FEED_INIT` to stash an `INIT` and `UNINIT` event to `dispatch`/directly-call-`onAction` as appropriate.

I tested this without #3240 and it passed mochitests.


system-addon/test/unit/lib/SnippetsFeed.test.js
```diff
@@ -70,14 +70,13 @@ describe("SnippetsFeed", () => {
     feed.onAction({type: at.INIT});
     assert.calledOnce(feed.init);
   });
-  it("should call .init when a FEED_INIT happens for feeds.snippets", () => {
+  it("should call .uninit on an UNINIT aciton", () => {
```
AdamHillier: Small typo here :)

Mardak: Also fixed the line it was copied from

k88hudson:
Oops sorry, I didn't see this for some reason - looks good to me!

as-pine-proxy:
[Treeherder: 46a9ad3031b8c5dec8233dcdc124cb658816fc09](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=46a9ad3031b8c5dec8233dcdc124cb658816fc09)

PR #3247, opened 2017-08-27
fix(styles): Fix Topics and others with inline-start/end instead of left/right

### LINKED ISSUE #3238: Topics and Section info break on RTL
Notice how the first line and bottom line are still positioned in the LTR areas. (Ignore the direction of text.)
image

The title and info icon happened to be fixed when I removed the `float`s in https://github.com/mozilla/activity-stream/pull/3226/files#diff-dd48c994d9ed857586fa08dfde37fa10L7

But the Topics list is still using `float`. These were originally added in #2896 @AdamHillier and #2901 @csadilek

Mardak:
Fix #3238. r?@csadilek Topics and/or r?@AdamHillier Info Section or for both. Also removed the width on topics section (not sure why it was there and not the right width to begin with…?) and moved the logo to `::after` as `inline-block` so it can be `transform`ed.

image

RTL:
image

and RTL info dropdown:
image

### DIFF
--- a/system-addon/content-src/components/Topics/_Topics.scss
+++ b/system-addon/content-src/components/Topics/_Topics.scss
@@ -6,7 +6,6 @@
 
   @media (min-width: $break-point-large) {
     line-height: 16px;
-    min-width: 780px;
   }
 
   ul {
@@ -14,21 +13,19 @@
     padding: 0;
     @media (min-width: $break-point-large) {
       display: inline;
-      padding-left: 12px;
+      padding-inline-start: 12px;
     }
   }
 
 
   ul li {
     display: inline-block;
-    @media (min-width: $break-point-large) {
-      display: inline;
-    }
+
     &::after {
       content: '•';
-      padding-left: 8px;
-      padding-right: 8px;
+      padding: 8px;
     }
+
     &:last-child::after {
       content: none;
     }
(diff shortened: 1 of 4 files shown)


system-addon/content-src/components/Topics/_Topics.scss
```diff
@@ -39,21 +36,32 @@
...
+    &::after {
+      background-image: url('#{$image-path}topic-show-more-12.svg');
+      background-repeat: no-repeat;
+      content: '';
+      -moz-context-properties: fill;
+      display: inline-block;
+      fill: $link-primary;
+      height: 16px;
+      margin-inline-start: 5px;
+      opacity: 0.8;
```
csadilek: This makes the > look slightly brighter than the topics on my screen. I'd suggest 0.9 or remove, but fine either way :).

Mardak: I'll just remove it. Part of the photon guidelines we were matching had 0.8 opacity for icons, but I think we can just treat this as if it was text.

as-pine-proxy:
[Treeherder: 4c882e3d61eaff5963a8e3cfe089ef63c2ddaf11](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4c882e3d61eaff5963a8e3cfe089ef63c2ddaf11)

PR #3258, opened 2017-08-28
feat(topstories): Personalized recs based on user domain affinities

### LINKED ISSUE #3259: Remove client-id from impression telemetry ping for Pocket
Because of personalization experiment might leak private data, let's make sure the client-id field is not transmitted on Pocket impression ping

csadilek:
closes #3259 .

This PR adds support for personalized recs based on domain affinities. 

*This functionality is turned **off** by default.* We want to bring this in so we can run experiments in 57.

The algorithm works as follows:

- Pocket's endpoint (v2) returns a settings object containing *timeSegments* [1] and *parameterSets* [2]
- For every time segment we calculate the corresponding domain visit counts, yielding result objects of the following structure: ```{"mozilla.org": 12, "mozilla.com": 34}```
- These visit counts are transformed to domain affinity scores for all provided parameter sets: ```{"mozilla.org": {"paramSet1": 0.8, "paramSet2": 0.9}, "mozilla.org": {"paramSet1": 1, "paramSet2": 0.9}}```
- The parameter sets provide factors for weighting which allows for flexible targeting. The functionality to calculate final scores can be seen in [3]
- The user domain affinity scores are summed up across all time segments, cached and only updated once a day

---

- Every item in Pocket's feed contains a list of domain affinities which are matched to the user's domain affinity scores by calculating an item relevance score [4]
- The items are sorted using these relevance scores and the top (3) items are displayed [5]
- Client-side rotation (only active if personalization is on) makes sure we don't display the same top 3 items all the time. This will be improved in future iterations to factor in user signals

---

Notes on implementation:
- The data structure was chosen to allow for fast cache lookups during relevance score calculation. While user domain affinities are calculated only once a day, the item relevance score (potentially) needs to be calculated every time we fetch the feed. Therefore allowing cache lookups of ```scores[domain][parameterSet]``` is beneficial
- I decided to use map/reduce where possible so the code very closely matches the description of the algorithm, to avoid mutable state, and to limit the lines of code required to implement this functionality
- Test coverage for all of the above included ;)

[1] [Time segments](https://github.com/csadilek/activity-stream/blob/personalization-poc/system-addon/lib/UserDomainAffinityProvider.jsm#L11)
[2] [Parameter sets](https://github.com/csadilek/activity-stream/blob/personalization-poc/system-addon/lib/UserDomainAffinityProvider.jsm#L19)
[3] [Scoring with weights](https://github.com/csadilek/activity-stream/blob/personalization-poc/system-addon/lib/UserDomainAffinityProvider.jsm#L160)
[4] [Relevance scoring for items](https://github.com/csadilek/activity-stream/blob/personalization-poc/system-addon/lib/UserDomainAffinityProvider.jsm#L226)
[5] [Item sorting](https://github.com/csadilek/activity-stream/blob/personalization-poc/system-addon/lib/TopStoriesFeed.jsm#L86)

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -3,19 +3,23 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-const {utils: Cu} = Components;
+const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/NewTabUtils.jsm");
 Cu.importGlobalProperties(["fetch"]);
 
 const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
+
 const {Prefs} = Cu.import("resource://activity-stream/lib/ActivityStreamPrefs.jsm", {});
 const {shortURL} = Cu.import("resource://activity-stream/lib/ShortURL.jsm", {});
 const {SectionsManager} = Cu.import("resource://activity-stream/lib/SectionsManager.jsm", {});
 
+const {UserDomainAffinityProvider} = Cu.import("resource://activity-stream/lib/UserDomainAffinityProvider.jsm", {});
+
 const STORIES_UPDATE_TIME = 30 * 60 * 1000; // 30 minutes
 const TOPICS_UPDATE_TIME = 3 * 60 * 60 * 1000; // 3 hours
+const DOMAIN_AFFINITY_UPDATE_TIME = 24 * 60 * 60 * 1000; // 24 hours
 const STORIES_NOW_THRESHOLD = 24 * 60 * 60 * 1000; // 24 hours
 const SECTION_ID = "topstories";
 const FEED_PREF = "feeds.section.topstories";
@@ -25,6 +29,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
   init() {
     this.storiesLastUpdated = 0;
     this.topicsLastUpdated = 0;
+    this.affinityLastUpdated = 0;
 
     SectionsManager.onceInitialized(this.parseOptions.bind(this));
   }
@@ -36,9 +41,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
       const apiKey = this._getApiKeyFromPref(options.api_key_pref);
       this.stories_endpoint = this._produceFinalEndpointUrl(options.stories_endpoint, apiKey);
       this.topics_endpoint = this._produceFinalEndpointUrl(options.topics_endpoint, apiKey);
-
       this.read_more_endpoint = options.read_more_endpoint;
       this.stories_referrer = options.stories_referrer;
+      this.personalized = options.personalized;
+      this.maxHistoryQueryResults = options.maxHistoryQueryResults;
 
       this.fetchStories();
       this.fetchTopics();
@@ -61,8 +67,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
           throw new Error(`Stories endpoint returned unexpected status: ${response.status}`);
         })
         .then(body => {
-          let items = JSON.parse(body).recommendations;
-          items = items
+          const response = JSON.parse(body);
+          this.updateDomainAffinities(response.settings);
+
+          const items = response.recommendations
             .filter(s => !NewTabUtils.blockedLinks.isBlocked({"url": s.url}))
             .map(s => ({
               "guid": s.id,
@@ -73,9 +81,12 @@ this.TopStoriesFeed = class TopStoriesFeed {
               "image": this._normalizeUrl(s.image_src),
               "referrer": this.stories_referrer,
               "url": s.url,
-              "eTLD": this._addETLD(s.url)
-            }));
-          return items;
+              "eTLD": this._addETLD(s.url),
+              "score": this.personalized ? this.affinityProvider.calculateItemRelevanceScore(s) : 1
+            }))
+            .sort(this.personalized ? this.compareScore : (a, b) => 0);
+
+          return this.rotate(items);
         })
         .catch(error => Cu.reportError(`Failed to fetch content: ${error.message}`));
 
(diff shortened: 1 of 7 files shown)


tspurway:
@csadilek can we link this telemetry bug to this patch: #3259 ?

This would allow us to collect unique impression data from the 'normal' case (by leaving the client-id in), but would block transmitting the client-id while running the personalization.

csadilek:
@tspurway OK, makes sense. I will fix this as part of this PR.

csadilek:
Updated. Now also closes #3259 .

csadilek:
Alright, who wants the hot potato? :) 

@ncloudioj @Mardak @k88hudson would you mind taking a look? This is an experiment we want to get in and we will be iterating on this. So, while it doesn't have to be perfect now I am hoping it's at least clean and maintainable.

system-addon/lib/UserDomainAffinityProvider.jsm
```diff
@@ -0,0 +1,262 @@
...
+      // Summing up the parameter set specific scores of each domain
+      .map(ps => ({[d]: {[ps]: Math.min(1, ((a[d] && a[d][ps]) || 0) + ((b[d] && b[d][ps]) || 0))}}))
+      // Reducing from an array of objects with a single parameter set to a single object
+      // [{"a.com":{"ps1":11}}, {"a.com: {"ps2":12}}] => {"a.com":{"ps1":11,"ps2":12}}
+      .reduce((acc, cur) => ({[d]: merge(acc[d], cur[d])}));
+  }
+
+  /**
+   * Calculates a value on the curve described by the provided parameters. The curve we're using is
+   * (a^(bx) - 1) / (a^b -1): https://www.desmos.com/calculator/maqhpttupp
```
ncloudioj: nit: I had some difficulty to associate the variables in this formula with the following function arguments. For example, I couldn't find `b` in the function, and this is no `y` in this formula. Could you clarify or reorganize it please?

csadilek:
@ncloudioj thanks, all good suggestions, fixed them all now. Can you take another look please!

system-addon/lib/UserDomainAffinityProvider.jsm
```diff
@@ -0,0 +1,262 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
```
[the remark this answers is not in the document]
csadilek: ok, detailed description added on class level.

system-addon/lib/UserDomainAffinityProvider.jsm
csadilek: Thanks, clarified and improved jsdocs.

ncloudioj:
@csadilek Cool, I am doing another run now. Looks like Travis is unhappy, you can take a look what's going on there.

csadilek:
@ncloudioj thanks, yes, had to update a test after the refactoring which got rid of side-effects.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -73,9 +81,12 @@ this.TopStoriesFeed = class TopStoriesFeed {
               "image": this._normalizeUrl(s.image_src),
               "referrer": this.stories_referrer,
               "url": s.url,
-              "eTLD": this._addETLD(s.url)
-            }));
-          return items;
+              "eTLD": this._addETLD(s.url),
+              "score": (this.personalized) ? this.affinityProvider.calculateItemRelevanceScore(s) : 1
```
ncloudioj: nit: `()` is redundant, ditto for for the following statement.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -109,6 +120,42 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+        this.maxHistoryQueryResults);
+      this.affinityLastUpdated = Date.now();
+    }
+  }
+
+  // If personalization is turned on we have to rotate stories on the client.
+  // An item can only be on top for two iterations (1hr) before it gets moved
+  // to the end. This will later be improved based on interactions/impressions.
+  rotate(items) {
+    if (!this.personalized || items.length === 0 || items.length <= 3) {
```
ncloudioj: I mean we could just have `items.length <= 3` :)

system-addon/lib/UserDomainAffinityProvider.jsm
```diff
@@ -0,0 +1,307 @@
...
+      // Summing up the parameter set specific scores of each domain
+      .map(ps => ({[d]: {[ps]: Math.min(1, ((a[d] && a[d][ps]) || 0) + ((b[d] && b[d][ps]) || 0))}}))
+      // Reducing from an array of objects with a single parameter set to a single object
+      // [{"a.com":{"ps1":11}}, {"a.com: {"ps2":12}}] => {"a.com":{"ps1":11,"ps2":12}}
+      .reduce((acc, cur) => ({[d]: merge(acc[d], cur[d])}));
+  }
+
+  /**
+   * Calculates a value on the curve described by the provided parameters. The curve we're using is
+   * (a^(bx) - 1) / (a^b -1): https://www.desmos.com/calculator/maqhpttupp
```
ncloudioj: Let's use `(a^(b*x) - 1) / (a^b - 1)` :)

csadilek:
@ncloudioj thanks, nits fixed. :)

as-pine-proxy:
[Treeherder: 6d101c99e2f64ab5acecb731b2319c82cf9ed683](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6d101c99e2f64ab5acecb731b2319c82cf9ed683)

jamskahler:
Hey I've got a serious question here. I just came across this after looking through some of my cookies in Firefox. I can't help but notice that there's no real clear linkage between the "domain affinities" that Pocket tags along with their stories and the stories that show up in my feed.

So what exactly is the point of this feature again? 

Given that Firefox is "open-source", it'd be nice to have something to show that they're not on the bandwagon for making our internet experiences a PITA... and I think most people have found themselves frustrated at the quality of content most "news feeds" are providing these days...

...

jamskahler:
Great example.

Story A:

title: "The heart-warming tale of the 8-year-old chess champion is quintessentially American"
excerpt: "The victory of 8-year-old Tanitoluwa Adewumi in the New York State K-3 championship this month has received more attention than any chess story in a long time."

image

What exactly does "chess" and "8-year-old" and "heartwarming" have anything to do with the typical demographic prescribed by these sites? I can guarantee your average football, hockey, and NBA fan couldn't give a single _____ about this story.

Issue #3265, opened 2017-08-28
Replace "Send Feedback" link in TopStories info section with link to privacy notice


csadilek:
If we can't get an alternative link and text by the end of this week, let's just remove the link altogether.

<img width="431" alt="screen shot 2017-08-28 at 2 51 00 pm" src="https://user-images.githubusercontent.com/472523/29788414-7d6000a2-8c00-11e7-9622-360b8f1285ac.png">

Mardak:
Theres #3106 for removing the link completely.

mariappopova:
Maria to check with Elvin and update the string

Mardak:
For now, I've put "Privacy Policy" in the in-progress PR #3274.

PR #3216, opened 2017-08-23
feat(sections): Add section ordering, and custom sections to the pref pane

AdamHillier:
Move TopStories config out of TopStoriesFeed and PreferencesPane and
into SectionsManager.jsm. Underlying work for #3194.

Issue #3155, opened 2017-08-11
Add section in prefs to turn off parts of Highlights


Mardak:
At minimum the whole section can be turned off but the full feature is to be able to individually turn off bookmarks and visited sites.

https://mozilla.invisionapp.com/share/4CCZSU2BU#/screens/248174509_New_Tab_Prefs

<img width="357" alt="image" src="https://user-images.githubusercontent.com/438537/29232281-b1370696-7e9f-11e7-90f4-b4c30b8c38eb.png">

Pref pane behavior:
- If the user unchecks both Bookmarks and Visited Sites, the Highlights section is also unchecked.
- If Highlights section is unchecked, the sub option box is disabled at 50% opacity not allowing clicks but maintains whatever the user has checked for Bookmarks and Visited Sites.
- If all sub options are unchecked and the user is turning on Highlights section, all options are now checked.

Highlight section behavior (from https://github.com/mozilla/activity-stream/issues/3378#issuecomment-327960680):
- If the user is excluding Bookmarks, all items should appear as Visited but context menu allows removing a bookmarked page. They keep the Visited type even when bookmarked and without shifting around cards, but the context menu will update to allow Remove Bookmark.
- If the user is excluding Visited Sites, only bookmarks should appear, and these show Bookmarked. If the user then Remove Bookmark from context menu, that card is dismissed immediately.

Strings:
- settings_pane_highlights_header=Highlights
- settings_pane_highlights_body2=Find your way back to interesting things you’ve recently visited or bookmarked.
- settings_pane_highlights_options_bookmarks=Bookmarks
- settings_pane_highlights_options_visited=Visited Sites

Mardak:
The strings for the pref pane were removed when the Highlights header was removed.

sarracini:
Blocked on actually implementing highlights

Mardak:
Added a note that there's more complex behavior of excluding bookmarks or excluding history https://github.com/mozilla/activity-stream/issues/3378#issuecomment-327960680

Mardak:
Moved to https://bugzilla.mozilla.org/show_bug.cgi?id=1432595

Issue #3147, opened 2017-08-10
Add section called "Highlights"


Mardak:
Fill as many cards on Highlights with visited bookmarks up to 5 days old: "Bookmarked"
Fill the rest with most recent history item (can be bookmarked): "Visited"
Looks like the total is 9 cards with no timestamp but "Bookmarked" or "Visited" with matching icon. If a card's url is actually bookmarked, its context menu should have "Remove Bookmark".
None of the highlights should exactly match the url of a recently bookmarked or recently visited top site. But pages/articles visited from the top site should appear as Highlights.
The prefs from #3155 section allows for turning off of bookmarks and/or history.

![image](https://user-images.githubusercontent.com/438537/29196173-98a97888-7de7-11e7-81bd-4d7695c8607a.png)

https://trello.com/c/kijapFZ2/156-i-want-to-see-a-highlights-section-on-new-tab

Strings:
- header_highlights=Highlights
- highlights_empty_state=Start browsing, and we’ll show some of the great articles, videos, and other pages you’ve recently visited or bookmarked here.

https://mozilla.invisionapp.com/share/FUD0ADOYE#/screens/248319677_NewTab_-_Empty_States

sarracini:
Blocked on metadata

Mardak:
"This should work the same way as Pocket" so use section manager and reuse telemetry provided by section manager.

Mardak:
Bookmarks should be always be shown even without metadata "because it’s a bookmark it’s more likely to feel familiar even if it’s missing thumbnails." Whereas a history page should only be shown if there's an image

Mardak:
Updated initial comment to note some additional behaviors from UX:
- old bookmarks can appear as "Visited" but context menu will show "Remove Bookmark"
- top site urls should not appear in Highlights (but their pages/articles should appear)

PR #3274, opened 2017-08-29
chore(l10n): Add remaining strings for Firefox 57 (highlights, Pocket, prefs)

### LINKED ISSUE #3267: Land strings for string freeze
This will be strings for #3146, #3147 and #3155.

### LINKED ISSUE #3271: Update Pocket prefs description
Discover high-quality content you might otherwise miss, with help from Pocket, now part of Mozilla.

https://mozilla.invisionapp.com/share/4CCZSU2BU#/screens/248174509_New_Tab_Prefs

Mardak:
Fix #3267 by adding strings for #3146, #3147 and #3155. Also fix #3271. r?@sarracini 

screen shot 2017-08-31 at 4 33 13 pm

---

old stuff

@flodolo this isn't the final set of strings yet as we have some "Design feedback needed" issues (`settings_pane_snippets_header=Snippets` -> "Tips from Mozilla" ? and #3265), but we're trying to be string frozen by end of this week, so I'll update this PR by then. A couple questions for you:

1. We previously removed Highlights in #2740 and split it to a Recently Visited section and a Bookmarks section, but it was decided to combine them for the initial 57 release and potentially split again later. `header_highlights=Highlights` is actually exactly the same from before but `settings_pane_highlights_body` is a previously used and currently removed identifier with a new string. Is it okay to reuse that string identifier as Pontoon has removed it automatically?

2. In #3216 @AdamHillier made it so that (the section header + description) are both used in the (section top line + info-option/feedback area) as well as in the preferences pane (line with checkbox + body). Is that okay to reuse the strings or should they stay duplicated? E.g., I've renamed to `pocket_description` for the double-use of the string.

3. There's a number of currently unused strings (that were used in Test Pilot) but haven't been ported to the system-addon version as well as a few strings that seem like they won't be used at all. Should we try to clean some of that up potentially running into the highlights-removed-and-now-back issue? -> #3314

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -3,6 +3,7 @@ default_label_loading=Loading…
 
 header_top_sites=Top Sites
 header_stories=Top Stories
+header_highlights=Highlights
 header_visit_again=Visit Again
 header_bookmarks=Recent Bookmarks
 # LOCALIZATION NOTE(header_recommended_by): This is followed by the name
@@ -72,6 +73,7 @@ search_settings=Change Search Settings
 # LOCALIZATION NOTE (section_info_option): This is the screenreader text for the
 # (?) icon that would show a section's description with optional feedback link.
 section_info_option=Info
+section_info_send_feedback=Send Feedback
 section_info_privacy_notice=Privacy Notice
 
 # LOCALIZATION NOTE (welcome_*): This is shown as a modal dialog, typically on a
@@ -90,7 +92,7 @@ time_label_day={number}d
 # LOCALIZATION NOTE (settings_pane_*): This is shown in the Settings Pane sidebar.
 settings_pane_button_label=Customize your New Tab page
 settings_pane_header=New Tab Preferences
-settings_pane_body=Choose what you see when you open a new tab.
+settings_pane_body2=Choose what you see on this page.
 settings_pane_search_header=Search
 settings_pane_search_body=Search the Web from your new tab.
 settings_pane_topsites_header=Top Sites
@@ -100,8 +102,16 @@ settings_pane_bookmarks_header=Recent Bookmarks
 settings_pane_bookmarks_body=Your newly created bookmarks in one handy location.
 settings_pane_visit_again_header=Visit Again
 settings_pane_visit_again_body=Firefox will show you parts of your browsing history that you might want to remember or get back to.
-settings_pane_pocketstories_header=Top Stories
-settings_pane_pocketstories_body=Pocket, a part of the Mozilla family, will help connect you to high-quality content that you may not have found otherwise.
+settings_pane_highlights_header=Highlights
+settings_pane_highlights_body2=Find your way back to interesting things you’ve recently visited or bookmarked.
+settings_pane_highlights_options_bookmarks=Bookmarks
+settings_pane_highlights_options_visited=Visited Sites
+# LOCALIZATION NOTE(settings_pane_snippets_header): For the "Snippets" feature
+# traditionally on about:home. Alternative translation options: "Small Note" or
+# something that expresses the idea of "a small message, shortened from
+# something else, and non-essential but also not entirely trivial and useless."
+settings_pane_snippets_header=Snippets
+settings_pane_snippets_body=Read short and sweet updates from Mozilla about Firefox, internet culture, and the occasional random meme.
 settings_pane_done_button=Done
 
 # LOCALIZATION NOTE (edit_topsites_*): This is shown in the Edit Top Sites modal
@@ -136,19 +146,19 @@ pocket_read_even_more=View More Stories
 # LOCALIZATION NOTE (pocket_feedback_header): This is shown as an introduction
 # to Pocket as part of the feedback form.
 pocket_feedback_header=The best of the web, curated by over 25 million people.
-# LOCALIZATION NOTE (pocket_feedback_body): This is shown below
-# (pocket_feedback_header) to provide more information about Pocket.
-pocket_feedback_body=Pocket, a part of the Mozilla family, will help connect you to high-quality content that you may not have found otherwise.
-pocket_send_feedback=Send Feedback
+# LOCALIZATION NOTE (pocket_description): This is shown in the settings pane and
+# below (pocket_feedback_header) to provide more information about Pocket.
+pocket_description=Discover high-quality content you might otherwise miss, with help from Pocket, now part of Mozilla.
 
+highlights_empty_state=Start browsing, and we’ll show some of the great articles, videos, and other pages you’ve recently visited or bookmarked here.
 # LOCALIZATION NOTE (topstories_empty_state): When there are no recommendations,
 # in the space that would have shown a few stories, this is shown instead.
 # {provider} is replaced by the name of the content provider for this section.
 topstories_empty_state=You’ve caught up. Check back later for more top stories from {provider}. Can’t wait? Select a popular topic to find more great stories from around the web.
 
-# LOCALIZATION NOTE (manual_migration_explanation): This message is shown to encourage users to
+# LOCALIZATION NOTE (manual_migration_explanation2): This message is shown to encourage users to
 # import their browser profile from another browser they might be using.
-manual_migration_explanation=Try Firefox with your favorite sites and bookmarks from another browser.
+manual_migration_explanation2=Try Firefox with the bookmarks, history and passwords from another browser.
 # LOCALIZATION NOTE (manual_migration_cancel_button): This message is shown on a button that cancels the
 # process of importing another browser’s profile into Firefox.
 manual_migration_cancel_button=No Thanks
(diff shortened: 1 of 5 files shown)


Mardak:
@aaronrbenson @bryanbell can you confirm that these are the final strings we want for 57? In particular, looks like there's still some discussion about "Snippets" #3146 and pocket feedback #3265.

Here's some screenshots of the strings that will be in 57 if we don't change anything:

image

image

image

image

image

image (from design)

image (from design)

image

image

image

image

image

image

image

image

image

image

image (edit modal)

image

image

image

image

image

image

image

image

image

image

image (from design)

image

image

image (from design)

image (from design)

flodolo:
> header_highlights=Highlights is actually exactly the same from before but settings_pane_highlights_body is a previously used and currently removed identifier with a new string. Is it okay to reuse that string identifier as Pontoon has removed it automatically?

Pontoon doesn't remove obsolete strings, unless there's human activity in the file (i.e. you translate a string). So we have plenty of `settings_pane_highlights_body` [still around](https://github.com/mozilla/activity-stream/search?utf8=%E2%9C%93&q=settings_pane_highlights_body&type=).

> made it so that (the section header + description) are both used in the (section top line + info-option/feedback area) as well as in the preferences pane (line with checkbox + body). Is that okay to reuse the strings or should they stay duplicated

Let's keep them separated, with a clear comment on what is used where. I'm mostly worried about space.

> Should we try to clean some of that up potentially running into the highlights-removed-and-now-back issue?

Yes please. There's one major pro: some locales might have not translated those strings, and will if they decide to start or finish translating AS.

Mardak:
> So we have plenty of settings_pane_highlights_body still around.

Okay, I'll name it `settings_pane_highlights_body2`.

> I'm mostly worried about space.

For the PR-named `pocket_description`, it's being used in two paragraph contexts that should be fine to expand to multiple lines. What's the concern of space? And if there shouldn't be a concern, is it okay to reuse the same string?
image
image


> Yes please. 
Okay, I'll double check to see which things are not going to be used, and at least remove those.

flodolo:
> For the PR-named pocket_description, it's being used in two paragraph contexts that should be fine to expand to multiple lines. What's the concern of space? And if there shouldn't be a concern, is it okay to reuse the same string?

I should have looked at the image. Yes, since the context is basically the same, and there are no issues with space, it makes definitely sense to use the same string.

Mardak:
@flodolo can you take a look at these string changes? I wasn't sure if "Snippets" needed a special localization note, but a quick look through m-c shows it's not a string that's translated for users to see until now:
https://github.com/mozilla/activity-stream/pull/3274/files#diff-906742e1af7418eb557312dc9217f2f1R109

flodolo:
Are we sure we want to call them out as "Snippets"? That's really internal jargon at Mozilla, nobody will understand it. In fact, there's no UI reference to them anywhere.

Either we explain somewhere what a snippet is, or IMO we're throwing a very confusing term to users.

Mardak:
Design said they did indeed do their naming pow-wow and still ended up with "Snippets".

flodolo:
> Design said they did indeed do their naming pow-wow and still ended up with "Snippets".

/me shakes his tiny fist at designers 👊 

Localizers know what snippets are, but that's going to be a really though one to translate. It would be great if we could have an alternative text in the comment to get them unstuck.

bryanbell:
Snippets are the best option of all the bad or potentially biasing names we could call these types of things.  FIrefox Messages, Top Tips, News Blobs, are all more descriptive, but also so narrowly cast that they don't correctly describe the types of content that snippets cover.  Snippets are meaningful enough to imply a short message and meaningless enough not to suggest the nature of the content they will show.

Mardak:
The thing is, the long description is the very next line after "Snippets" and hopefully users will understand after that. I could definitely see how if "Snippets" was by itself under the "View" menu, users would have no idea what it would do.

@bryanbell Looking at some translations/synonyms, the connotation we want with Snippet is that it's a small piece "cut" from the larger, like a patch from fabric. It should **not** be like a "splinter" / "shred" breaking from a larger piece or a "crumb" / "scrap" from food or the "remains" of something left over.

flodolo:
> Snippets are the best option of all the bad or potentially biasing names we could call these types of things. FIrefox Messages, Top Tips, News Blobs, are all more descriptive, but also so narrowly cast that they don't correctly describe the types of content that snippets cover. Snippets are meaningful enough to imply a short message and meaningless enough not to suggest the nature of the content they will show.

If you had to pick a different name than "Snippets", something that explain what they do, what would it be? Again, I think we owe localizers an alternative text if they can't find a solution for "Snippets".

bryanbell:
Open to suggestions here. "Notes" might be a decent analog for non-english languages.

flodolo:
I think "Messages" is closer to the actual functionality than notes (and I kind of like "Firefox Messages"…).

bryanbell:
I agree with you, but we want to reserve the word Messages because there's talk of creating a Firefox Message Center that would list Push Notifications, Release Notes, Longer form Snippets, Promotions Info, Tips on how to use Firefox, Recommended Add-ons, Etc.

Mardak:
Mozilla Feed? Like a news feed with updates but only for Mozilla stuff..

Mardak:
@aaronrbenson I see there was some activity around Link vs Page vs.. location bar says "Search or enter address"
image

aaronrbenson:
@flodolo @Mardak Just to be clear, this wasn't a name that we put together last minute. It went through the channels of marketing, content strategy and the snippets team themselves. 

Firefox Messages was considered but ultimately dropped because much of the messaging will come from the Mozilla organization, so it isn't appropriate to associate them to Firefox. Mozilla Messages sounds too stuffy and official.

We've not referenced snippets in the past because there was no mechanism of managing them in the UI; they existed only as dismissable boxes with no way of opting out of them completely. Now, the section in the preferences pane has a description that should make it relatively clear what the intention is (even if we didn't translate the headline 'Snippet', however I'm not sure of the precedent for that sort of thing). 

I understand it's a bit of a colloquialism and challenging to translate. Honestly it's one of the reasons we came back to it. Other words (messages, notifications) are loaded and don't suggest the proper tone. 

As for alternative text, something along the lines of "small note" or "simple note" would probably work.

aaronrbenson:
@Mardak re: the menu there I think those terms are fine. The conversation about the Page Action Menu was to put some consistency about how we referred to things in that pane. Here, the term URL is more specific (and I think accurate) than referring to it as a link.

Mardak:
Updated with final final strings for 57. r?@sarracini 

screen shot 2017-08-31 at 1 25 07 pm

Mardak:
Got a string update from @aaronrbenson to have the "Choose what you see when you open a new tab." not wrap for en-US with "Choose what you see on this page." given the new styles with #3330:
image

I don't think it needs a string id change as it's still basically the same string. ? @flodolo 

I also added more to the "Snippets" localization note:

For the "Snippets" feature traditionally on about:home. Alternative translation options: "Small Note" or something that expresses the idea of "a small message, shortened from something else, and non-essential but also not entirely trivial and useless."

flodolo:
> I don't think it needs a string id change as it's still basically the same string. ? @flodolo

I would still change the string ID to make them re-evaluate their translation and don't mess up translation memories.

locales/en-US/strings.properties
```diff
@@ -136,19 +146,19 @@ pocket_read_even_more=View More Stories
...
 pocket_feedback_header=The best of the web, curated by over 25 million people.
-# LOCALIZATION NOTE (pocket_feedback_body): This is shown below
-# (pocket_feedback_header) to provide more information about Pocket.
-pocket_feedback_body=Pocket, a part of the Mozilla family, will help connect you to high-quality content that you may not have found otherwise.
-pocket_send_feedback=Send Feedback
+# LOCALIZATION NOTE (pocket_description): This is shown in the settings pane and
+# below (pocket_feedback_header) to provide more information about Pocket.
+pocket_description=Discover high-quality content you might otherwise miss, with help from Pocket, now part of Mozilla.
 
+highlights_empty_state=Start browsing, and we’ll show some of the great articles, videos, and other pages you’ve recently visited or bookmarked here.
```
flodolo: Unrelated to localization: isn’t *here* way too far from show? It sounds like it’s associated to bookmark (but that might be my non English brain).

locales/en-US/strings.properties
```diff
@@ -90,7 +92,7 @@ time_label_day={number}d
 # LOCALIZATION NOTE (settings_pane_*): This is shown in the Settings Pane sidebar.
 settings_pane_button_label=Customize your New Tab page
 settings_pane_header=New Tab Preferences
-settings_pane_body=Choose what you see when you open a new tab.
+settings_pane_body=Choose what you see on this page.
```
flodolo: As said, I would rename this string ID at this point.

locales/en-US/strings.properties
Mardak: I believe that "place" adverbs (e.g., here) typically go at the end of sentences even when split by a (long) preposition.

"We'll show some of … here." -> "We'll show some here." vs "We'll show here some."

Or a difference sentence:
"Play with the ball there" vs "Play there with the ball" where the former with "place" adverb at end sounds better (at least to me…)

locales/en-US/strings.properties
Mardak: Renamed to settings_pane_body2.

PR #3279, opened 2017-08-29
fix(mc): Disable private browsing about:home test. Closes #3278.

### LINKED ISSUE #3278: PERMA-FAIL | browser/components/privatebrowsing/test/browser/browser_privatebrowsing_aboutHomeButtonAfterWindowClose.js | Uncaught exception
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1395319

```
FAIL | browser/components/privatebrowsing/test/browser/browser_privatebrowsing_aboutHomeButtonAfterWindowClose.js | Uncaught exception
TypeError: Argument 1 of Window.getComputedStyle is not an object.
```
https://treeherder.mozilla.org/#/jobs?repo=try&revision=4e323761c128b40621ffdff46c0e92dc6463df34&group_state=expanded&selectedJob=127119664

```
let button = content.document.getElementById("restorePreviousSession");
```
Is tested to be disabled.

piatra:


Mardak:
Why are we skipping [browser_privatebrowsing_about.js] ? It looks like it's testing that about:privatebrowsing is loaded on a new private window…?

piatra:
@Mardak I thought the `skip-if = true` line skips the following line. Is that not how it works?

Mardak:
These are `.ini` files, so there's a section `[header]` and any number of key/value properties `skip-if=true` for that section.

https://en.wikipedia.org/wiki/INI_file

You can test it out by putting the `skip-if` and seeing if the test runs (or fails in this situation).

piatra:
https://bugzilla.mozilla.org/show_bug.cgi?id=1395319

PR #3282, opened 2017-08-29
fix(systemaddon): Use tippytop title as hostname for default sites.

### LINKED ISSUE #3229: Topsite deduping issue
image
Amazon shows up twice. 1 is from default sites, 1 is from an actual visit. The default site has no eTLD information while the visit will have `eTLD: "co.uk"` this results in different hostname (which is the key after which we dedupe).
My proposed fix is to just include eTLD with the default sites.
This also overcomes the issue on first run (you have a cold cache, no eTLD for any website).

cc @Mardak @k88hudson

### LINKED ISSUE #3231: Default top sites only remove .com eTLD
@piatra noticed in #3221 that the default sites don't remove eTLD unless .com:

image

This is because there's no eTLD info and we fall back to "com".

This causes problems like #3229

piatra:
Closes #3231 
Closes #3229

### DIFF
--- a/system-addon/lib/ShortURL.jsm
+++ b/system-addon/lib/ShortURL.jsm
@@ -1,5 +1,6 @@
 const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
 
 XPCOMUtils.defineLazyServiceGetter(this, "IDNService", "@mozilla.org/network/idn-service;1", "nsIIDNService");
 
@@ -21,6 +22,20 @@ function handleIDNHost(hostname) {
 }
 
 /**
+ * Returns the public suffix of a URL or empty string in case of error.
+ * @param {string} url The url to be analyzed.
+ */
+function getETLD(url) {
+  try {
+    return Services.eTLD.getPublicSuffix(Services.io.newURI(url));
+  } catch (err) {
+    return "";
+  }
+}
+
+this.getETLD = getETLD;
+
+  /**
  * shortURL - Creates a short version of a link's url, used for display purposes
  *            e.g. {url: http://www.foosite.com, eTLD: "com"}  =>  "foosite"
  *
@@ -39,7 +54,7 @@ this.shortURL = function shortURL(link) {
   if (!link.url && !link.hostname) {
     return "";
   }
-  const {eTLD} = link;
+  const eTLD = link.eTLD || getETLD(link.url);
   const hostname = (link.hostname || new URL(link.url).hostname).replace(/^www\./i, "");
 
   // Remove the eTLD (e.g., com, net) and the preceding period from the hostname
@@ -49,4 +64,4 @@ this.shortURL = function shortURL(link) {
   return handleIDNHost(hostname.slice(0, eTLDExtra).toLowerCase() || hostname) || link.title || link.url;
 };
 
-this.EXPORTED_SYMBOLS = ["shortURL"];
+this.EXPORTED_SYMBOLS = ["shortURL", "getETLD"];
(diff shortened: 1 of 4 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -52,7 +52,12 @@ this.TopSitesFeed = class TopSitesFeed {
   }
   async getLinksWithDefaults(action) {
     let frecent = await NewTabUtils.activityStreamLinks.getTopSites();
-    const notBlockedDefaultSites = DEFAULT_TOP_SITES.filter(site => !NewTabUtils.blockedLinks.isBlocked({url: site.url}));
+    const notBlockedDefaultSites = DEFAULT_TOP_SITES
+      .filter(site => !NewTabUtils.blockedLinks.isBlocked({url: site.url}))
+      .map(site => {
+        const tippySite = this._tippyTopProvider.processSite(site);
```
Mardak: @rlr should tippy top only be applying to the default sites and not other links? If so, should this `processSite` in `getLinksWithDefaults` remove the need to call `processSite` from `refresh` ?

rlr: @Mardak We want to run tippy top on all sites. Although, for now (in 56), we only have assets for default sites. I still think it's a better experience if you get the nicer screenshot for facebook, for example, whether you already visited it or it's displaying because of it's default status.

Mardak: Adding this dependency on tippy top doesn't seem right when metadata might be replacing it. @piatra any reason you can't just use `eTLD` with `shortURL`?

piatra: DEFAULT_URLS and frecent website information is never merged. You will have two `amazon` entities (the default and the actual visit) and the default_url entity will never receive eTLD information. It also won't be discarded by dedupe because we won't be able to extract the correct hostname key.

Mardak: What do you mean default_urls and frecent never get merged? Isn't the whole point of this method to combine the frecent links with defaults?

piatra: https://github.com/mozilla/activity-stream/blob/560e9b62a2bde24feb795f16418ffd49df54ee95/system-addon/lib/TopSitesFeed.jsm#L77 We dedupe pinned and default topsites. If an item is found in both groups, both will show up in your topsites. It will not be deduped because eTLD field will never be present for default topsites.

Mardak: Right... so why can't we add in eTLD as I suggested?

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -81,7 +81,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
               "image": this._normalizeUrl(s.image_src),
               "referrer": this.stories_referrer,
               "url": s.url,
-              "eTLD": this._addETLD(s.url),
+              "eTLD": getETLD(s.url),
```
Mardak: Your previous PR #2987 added the hostname field in `TopStoriesFeed` but actually regressed #3031 because the eTLD no longer existed by the time shortURL was called.

https://github.com/piatra/activity-stream/commit/d782ac9e6ea4e720b689b440300a2c38d9f8a4a1#diff-506774124923d486589dbeda753ebe38L51

This line is no longer necessary as `shortURL` will get the `eTLD` as necessary.

system-addon/lib/ShortURL.jsm
```diff
@@ -39,7 +54,7 @@ this.shortURL = function shortURL(link) {
   if (!link.url && !link.hostname) {
     return "";
   }
-  const {eTLD} = link;
+  const eTLD = link.eTLD || getETLD(link.url);
   const hostname = (link.hostname || new URL(link.url).hostname).replace(/^www\./i, "");
 
   // Remove the eTLD (e.g., com, net) and the preceding period from the hostname
```
Mardak: The following line for defaulting to .com shouldn't be necessary now that we're getting the actual eTLD always. `|| (hostname.match(/\.com$/) && 3);`

as-pine-proxy:
[Treeherder: a990baa58530577566f5e0ad4172d7bd36f0d80a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a990baa58530577566f5e0ad4172d7bd36f0d80a)

Issue #3284, opened 2017-08-29
Backport Bug 863246


k88hudson:
A slight change was made to jar.mn that needs to be backported. Bug here https://bugzilla.mozilla.org/show_bug.cgi?id=863246

k88hudson:
Fixed by 720e1638522cc348e4105f2b71b409aa5160648f

PR #3285, opened 2017-08-29
fix(mc): #3284 Backport Bug 863246

k88hudson:
This bug added contentacessible=yes to jar.mn. Here is the diff: https://hg.mozilla.org/mozilla-central/diff/a676a1188572/browser/extensions/activity-stream/jar.mn

Not including this causes major failures (no scripts can load)


k88hudson:
Closing as a dupe

PR #3286, opened 2017-08-29
fix(build): Fix bustage by semi-pinning node 7 (closes #3283)

### LINKED ISSUE #3283: buildmc fails in node 7
Right now some changes that landed in https://github.com/mozilla/activity-stream/commit/828acca53137caaf1cb5cce840f441b0b5e8917c use syntax that is only available in node 8.

I think it's a bit early to require node 8 for the project just for this task? What do you think @dmose?

dmose:
This semi-pins (actually pinning is more work) everything to 7.10.1, which fixes build bustage.

To test:
```bashrc
# nvm use
# npm install
# npm run packagemc
```


dmose:
Assuming this makes travis happy, I'll send out an email to the list suggesting that people reinstall stuff as I land.

dmose:
Alternately, I could do something less disruptive right now and just specify "node 7" without any more details.

bin/update-version.js
```diff
@@ -23,6 +23,21 @@ const simpleGit = require("simple-git")(process.cwd());
 const time = new Date();
 const minuteOfDay = time.getUTCHours() * 60 + time.getUTCMinutes();
 
+/**
+ * Return the given string padded with 0s out to the given width.
```
k88hudson: Maybe note that this can be replaced with `.padStart`when we upgrade to node 8?

dmose: Sure

as-pine-proxy:
[Treeherder: 1dd34d81b3e47a718712970888f62293143cf40c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1dd34d81b3e47a718712970888f62293143cf40c)

PR #3289, opened 2017-08-29
chore(build): Restore version zero pad of minutes to 4 instead of 5

Mardak:
Versions slightly changed with #3286. r?@dmose


dmose:
That change was intentional; when writing that patch, I realized the number of minutes in a day is 86400, so we actually want it to be padded out to 5 characters.

dmose:
And that was, of course, a thinko on my part.  Thanks for catching it.

Mardak:
In a day, there's 24 hours, and in an hour there's 60 minutes, so 24*60=
1440. It's a little over an hour from UTC midnight, so if you run the
script now, it should be close to the maximum number of minutes in a day.

as-pine-proxy:
[Treeherder: c27c91272c0644217fd4cc3c1be1920a570129ac](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c27c91272c0644217fd4cc3c1be1920a570129ac)

PR #3290, opened 2017-08-29
fix (systemaddon): use `new URL()` instead of regex for validation

rlr:
@Mardak thoughts on this approach?

One difference is that `new URL("http://whatever")` is valid. That's weird, but I guess it is technically correct?

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -467,7 +467,7 @@ class TopSiteForm extends React.Component {
   cleanUrl() {
     let url = this.state.url;
     // If we are missing a protocol, prepend http://
-    if (!url.startsWith("http")) {
+    if (!url.startsWith("http:") && !url.startsWith("https:")) {
       url = `http://${url}`;
     }
     return url;
@@ -478,16 +478,11 @@ class TopSiteForm extends React.Component {
     }
   }
   validateUrl() {
-    let url = this.state.url;
-    if (url && !url.startsWith("http")) {
-      url = `http://${url}`;
-    }
     try {
-      url = new URL(url);
+      return !!new URL(this.cleanUrl());
     } catch (e) {
       return false;
     }
-    return true;
   }
   validateForm() {
     this.resetValidation();
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -478,13 +478,16 @@ class TopSiteForm extends React.Component {
     }
   }
   validateUrl() {
-    const pattern = new RegExp("^(https?:\\/\\/)?" + // protocol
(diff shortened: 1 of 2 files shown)


Mardak:
"whatever" is similar to "localhost" and can be common for business networks

system-addon/test/unit/content-src/components/TopSites.test.jsx
```diff
@@ -651,8 +651,6 @@ describe("<TopSiteForm>", () => {
     assert.ok(wrapper.instance().validateUrl());
     wrapper.setState({"url": "http://mozilla.org"});
     assert.ok(wrapper.instance().validateUrl());
-    wrapper.setState({"url": "mozillaorg"});
-    assert.isFalse(wrapper.instance().validateUrl());
```
Mardak: Probably still still have an `isFalse` test probably with "mozilla org"

system-addon/test/unit/content-src/components/TopSites.test.jsx
```diff
@@ -601,7 +601,7 @@ describe("<TopSiteForm>", () => {
       assert.notCalled(wrapper.instance().props.dispatch);
     });
     it("should show error and not call onClose or dispatch if URL is invalid", () => {
-      wrapper.setState({"url": "invalid"});
+      wrapper.setState({"url": ""});
```
Mardak: if you want a non-empty test, "not valid" is good enough

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -478,13 +478,16 @@ class TopSiteForm extends React.Component {
...
   validateUrl() {
-    const pattern = new RegExp("^(https?:\\/\\/)?" + // protocol
-      "((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|" + // domain name
-      "((\\d{1,3}\\.){3}\\d{1,3}))" + // OR ip (v4) address
-      "(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*" + // port and path
-      "(\\?[;&a-z\\d%_.~+=-]*)?" + // query string
-      "(\\#[\\/-a-z\\d_]*)?$", "i"); // fragment locater
-    return pattern.test(this.state.url);
+    let url = this.state.url;
+    if (url && !url.startsWith("http")) {
```
Mardak: Some hostname could start with "http". I do see this same logic in `cleanUrl` too.. so can you fix that to just check for "http:" or "https:" and then do something like:

```js
validateUrl() {
  try {
    return !!new URL(this.cleanUrl());
  } catch (e) {
    return false;
  }
}
```

rlr:
@Mardak that cleaned it up nicely ^

as-pine-proxy:
[Treeherder: a94b47cad53f97790dce48dce9b8c984f9e2360c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a94b47cad53f97790dce48dce9b8c984f9e2360c)

PR #3297, opened 2017-08-30
fix(topsites): Closes #3296 Deleting from history should update the top sites

### LINKED ISSUE #3296: Deleting from history should update the top sites
Similar to how we update top sites when we block a top site, we should do the same when we delete from history a top site

sarracini:
Fix #3296. Instead of the Reducer listening for a ```BLOCK_URL``` and also a ```TOP_SITES_UPDATED``` event that comes from TopSitesFeed.jsm when you block -> refresh, we should just send a ```TOP_SITES_UPDATED``` event when we received the OK from PlacesFeed that the link has been blocked/deleted/history cleared. That way we save ourselves a bounce of a message.

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -135,8 +135,11 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
         return site;
       });
       return Object.assign({}, prevState, {rows: newRows});
-    case at.PLACES_LINK_DELETED:
-    case at.PLACES_LINK_BLOCKED:
+    case at.BLOCK_URL:
+    case at.DELETE_HISTORY_URL:
+      // Optimistically update the UI by responding to the context menu action
+      // events and removing the site that was blocked/deleted with an empty slot.
+      // Once refresh() finishes, we update the UI again with a new site
       newRows = prevState.rows.filter(val => val && val.url !== action.data.url);
       return Object.assign({}, prevState, {rows: newRows});
     case at.PINNED_SITES_UPDATED:
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -177,10 +177,9 @@ this.TopSitesFeed = class TopSitesFeed {
           this.refresh(action.meta.fromTarget);
         }
         break;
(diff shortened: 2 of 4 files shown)


system-addon/common/Reducers.jsm
```diff
@@ -135,10 +135,6 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
         return site;
       });
       return Object.assign({}, prevState, {rows: newRows});
-    case at.PLACES_LINK_DELETED:
-    case at.PLACES_LINK_BLOCKED:
-      newRows = prevState.rows.filter(val => val && val.url !== action.data.url);
-      return Object.assign({}, prevState, {rows: newRows});
```
Mardak: So @k88hudson has mentioned how it's nice to have the UI update right away without waiting for Places roundtrip especially on slow machines. Although while waiting for the refresh, there will be fewer than 12 items. until the `refresh` completes... @k88hudson ?

k88hudson: Yeah, this has the potential to be pretty slow on some machines, as well as causing some weird effects if the query returns new stuff

Mardak: So, these aren't the user events triggering the delete/block in the first place. These happen when PlacesFeed notices newtabutils has changed something. So keeping this reducer maybe saves the UI time between this action and `TOP_SITES_UPDATED`

Mardak: The user events seem to be `BLOCK_URL and `DELETE_HISTORY_URL` (`DELETE_HISTORY_URL_CONFIRM` and `DELETE_CONFIRM` were added in #2749 but are unused)

k88hudson: Ok ,since the broadcasting of updates to other pages is handled by a refresh now, let's just change these to use `BLOCK_URL` and `DELETE_HISTORY_URL` for the local instance only to help with the UI lag.

as-pine-proxy:
[Treeherder: 82397ac23e56d4d525072af4967c49d053e4037f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=82397ac23e56d4d525072af4967c49d053e4037f)

PR #3311, opened 2017-08-30
fix(topsites): Compute hostname once for defaults and only require url for shortURL

### LINKED ISSUE #3310: Default sites don't get deduped against pinned sites
@piatra 

image

Frecent sites do correctly dedupe against default sites.

This is because `shortURL` is unnecessarily being called multiple times within `TopSitesFeed`.

Pinned links first get a `hostname` field added:
https://github.com/mozilla/activity-stream/blob/c3bc0a1709c0545f45afec3bcda6f8775d489910/system-addon/lib/TopSitesFeed.jsm#L58-L61

Then `shortURL` is called a second time when there's already a `hostname`:
https://github.com/mozilla/activity-stream/blob/c3bc0a1709c0545f45afec3bcda6f8775d489910/system-addon/lib/TopSitesFeed.jsm#L74

This causes `shortURL` to slice off ".com" length of "amazon" (i.e., "azon") and say the shorturl is "am":
https://github.com/mozilla/activity-stream/blob/c3bc0a1709c0545f45afec3bcda6f8775d489910/system-addon/lib/ShortURL.jsm#L58-L64

Mardak:
r?@piatra This has a few behavior changes. `Dedupe.group` now takes each array as individual parameters instead of an array of arrays. `shortURL` no longer takes a `link` with `eTLD` nor `hostname` as those are unnecessarily provided (or worse -- could be wrong) when `url` includes them. `TopSitesFeed` no longer unnecessarily calls `shortURL`.

Either of the latter 2 changes would fix #3310. Either by not incorrectly slicing hostname or not incorrectly passing a pinned link with a `hostname` to `shortURL`. This fixes both as doing things unnecessarily has led to this issue.

Screenshot of fix where pinned amazon.com and visited bbc.com dedupe GB's default amazon.co.uk and bbc.co.uk:
screen shot 2017-08-30 at 4 23 19 pm

Also, might be useful to note that `TopSitesFeed.test.js` did **not** catch this because there's a fake shortURL being used.

### DIFF
--- a/system-addon/common/Dedupe.jsm
+++ b/system-addon/common/Dedupe.jsm
@@ -13,13 +13,12 @@ this.Dedupe = class Dedupe {
   }
 
   /**
-   * Dedupe an array containing groups of elements.
-   * Duplicate removal favors earlier groups.
+   * Dedupe any number of grouped elements favoring those from earlier groups.
    *
    * @param {Array} groups Contains an arbitrary number of arrays of elements.
-   * @returns {Array}
+   * @returns {Array} A matching array of each provided group deduped.
    */
-  group(groups) {
+  group(...groups) {
     const globalKeys = new Set();
     const result = [];
     for (const values of groups) {
--- a/system-addon/lib/ShortURL.jsm
+++ b/system-addon/lib/ShortURL.jsm
@@ -37,31 +37,26 @@ this.getETLD = getETLD;
 
   /**
  * shortURL - Creates a short version of a link's url, used for display purposes
- *            e.g. {url: http://www.foosite.com, eTLD: "com"}  =>  "foosite"
+ *            e.g. {url: http://www.foosite.com}  =>  "foosite"
  *
  * @param  {obj} link A link object
  *         {str} link.url (required)- The url of the link
- *         {str} link.eTLD (required) - The tld of the link
- *               e.g. for https://foo.org, the tld would be "org"
- *               Note that this property is added in various queries for ActivityStream
- *               via Services.eTLD.getPublicSuffix
- *         {str} link.hostname (optional) - The hostname of the url
- *               e.g. for http://www.hello.com/foo/bar, the hostname would be "www.hello.com"
  *         {str} link.title (optional) - The title of the link
  * @return {str}   A short url
  */
 this.shortURL = function shortURL(link) {
-  if (!link.url && !link.hostname) {
+  if (!link.url) {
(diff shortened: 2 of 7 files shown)


piatra:
> Also, might be useful to note that TopSitesFeed.test.js did not catch this because there's a fake shortURL being used.
Most of our test cases related to URLs are "made up" (`http://www.site${i}.com`) and do not reflect a real browsing history.
Could you add a test case that reproduces https://github.com/mozilla/activity-stream/issues/3310

Mardak:
Finally updated the test to fail without the fix. There was a surprising number of pieces that needed to be changed to have it fail:

1. `eTLD.getPublicSuffix` returned nothing, so the hostname wasn't sliced https://github.com/mozilla/activity-stream/pull/3311/files#diff-d727ceb3769ae1d9e0a2f2cfe333ce74R196
2. Even when `eTLD` returns something, it wasn't actually getting called because the dummy global `Services.io.newURI` was misspelled resulting in `getETLD` ending in the `catch` block https://github.com/mozilla/activity-stream/pull/3311/files#diff-d93b737609a1354bd3104347a979f859L66
3. The test for preferring pinned over links was wrongly named copy/pasted from the previous "should not dedupe pinned sites" https://github.com/mozilla/activity-stream/pull/3311/files#diff-d727ceb3769ae1d9e0a2f2cfe333ce74L205
4. The frecent test links incorrectly provided multiple links from the same host (where `getTopSites` would just return one per host) https://github.com/mozilla/activity-stream/pull/3311/files#diff-d727ceb3769ae1d9e0a2f2cfe333ce74L212
5. The frecent test links didn't have frecency set so they were actually being filtered out for low frecency instead of being duped against the pinned https://github.com/mozilla/activity-stream/pull/3311/files#diff-d727ceb3769ae1d9e0a2f2cfe333ce74R219
6. Related to the previous two, there were no other host frecent link to test that it wasn't being removed so just checking a length of 2 happened to match even if the links weren't actually the expected ones https://github.com/mozilla/activity-stream/pull/3311/files#diff-d727ceb3769ae1d9e0a2f2cfe333ce74L218

as-pine-proxy:
[Treeherder: 627e033aa3d35f9288b9c508d929c0fe7e4d5614](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=627e033aa3d35f9288b9c508d929c0fe7e4d5614)

PR #3317, opened 2017-08-31
fix(systemaddon): Fix #3203 Avoid impacting performance when listening to onItemChanged for bookmarks - avoid the fetch by disabling the unused code for now.

### LINKED ISSUE #3203: Avoid impacting performance with fetching from onItemChanged
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1392267

This code would probably be used by #3147 but is currently unused. So if it's not making it in for 57, we should probably clean this up to avoid impacting performance.

Standard8:
As this isn't used yet, I'm suggesting that we disable the code for now - we're starting to look at bookmarks performance with respect to recent changes, so I'd like to get this cleaned up.

I've also added a comment that we're happy to talk about adding more fields if we need to in future.

### DIFF
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -127,21 +127,29 @@ class BookmarksObserver extends Observer {
    * @param  {str} guid         The unique id of the bookmark
    */
   async onItemChanged(...args) {
-    const property = args[1];
-    const type = args[5];
-    const guid = args[7];
+    // Disabled due to performance cost, see Issue 3203 /
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1392267.
+    //
+    // If this is used, please consider avoiding the call to
+    // NewTabUtils.activityStreamProvider.getBookmark which performs an additional
+    // fetch to the database.
+    // If you need more fields, please talk to the places team.
 
-    // Only process this event if it is a TYPE_BOOKMARK, and uri or title was the property changed.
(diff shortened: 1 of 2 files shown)


system-addon/lib/PlacesFeed.jsm
```diff
@@ -127,21 +127,29 @@ class BookmarksObserver extends Observer {
...
-      return;
-    }
-    try {
-      // bookmark: {bookmarkGuid, bookmarkTitle, lastModified, url}
-      const bookmark = await NewTabUtils.activityStreamProvider.getBookmark(guid);
-      this.dispatch({type: at.PLACES_BOOKMARK_CHANGED, data: bookmark});
-    } catch (e) {
-      Cu.reportError(e);
-    }
+    // Disabled due to performance cost, see Issue 3203 /
```
Mardak: nit: add a empty line (eslint) and comment block opening before these `//` comments

system-addon/lib/PlacesFeed.jsm
```diff
@@ -127,21 +127,29 @@ class BookmarksObserver extends Observer {
...
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1392267.
+    //
+    // If this is used, please consider avoiding the call to
+    // NewTabUtils.activityStreamProvider.getBookmark which performs an additional
+    // fetch to the database.
+    // If you need more fields, please talk to the places team.
+
+    // const property = args[1];
```
Mardak: then un-comment this original code adding the comment block closing at the end of the method to preserve blame.

PR #3322, opened 2017-08-31
fix(places): Avoid impacting performance when listening to onItemChanged for bookmarks - avoid the fetch by disabling the unused code for now

### LINKED ISSUE #3203: Avoid impacting performance with fetching from onItemChanged

Mardak:
Fix #3203 and closes #3317 by superseding.

### DIFF
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -127,6 +127,16 @@ class BookmarksObserver extends Observer {
    * @param  {str} guid         The unique id of the bookmark
    */
   async onItemChanged(...args) {
+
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 206b8722011e48be3def0132c82b6a3448880e10](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=206b8722011e48be3def0132c82b6a3448880e10)

Issue #3262, opened 2017-08-28
Port the placeholders for Top Sites, Sections


rlr:
In the test pilot addon, we would pad the Top Sites with placeholders when the user didn't have enough to fill the 1 or 2 rows. This is noticeable on fresh profiles with the top sites pref set to 2 rows:

<img width="1026" alt="screen shot 2017-08-28 at 1 48 36 pm" src="https://user-images.githubusercontent.com/36629/29786028-9f7e6c4a-8bf7-11e7-9a82-f7d56b93cd7c.png">

Mardak:
This might be handled as part of prerendering to avoid jumping around.

Mardak:
Fixed top sites #3323 and sections #3337.

Issue #12, opened 2016-01-30
Get some sample data for testing


k88hudson:
We should grab some images, text, etc. for a few sites so we can test our UI with it, until the enhanced data api is ready. We could use something like https://www.npmjs.com/package/open-graph-scraper

tspurway:
@nchapman to donate embedly account

PR #3323, opened 2017-08-31
feat(mc): #3262 Add placeholders for TopSites

k88hudson:
This adds placeholders for TopSites. Part of the prerendering work.

This is alternative approach that requires fewer changes, which one do you think is better?

https://github.com/mozilla/activity-stream/compare/master...k88hudson:gh3262b?expand=1

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -5,6 +5,41 @@ const LinkMenu = require("content-src/components/LinkMenu/LinkMenu");
 
 const {TOP_SITES_SOURCE, TOP_SITES_CONTEXT_MENU_OPTIONS} = require("./TopSitesConstants");
 
+const TopSiteLink = props => {
+  const {link} = props;
+  const topSiteOuterClassName = `top-site-outer${props.className ? ` ${props.className}` : ""}`;
+  const {tippyTopIcon} = link;
+  let imageClassName;
+  let imageStyle;
+  if (tippyTopIcon) {
+    imageClassName = "tippy-top-icon";
+    imageStyle = {
+      backgroundColor: link.backgroundColor,
+      backgroundImage: `url(${tippyTopIcon})`
+    };
+  } else {
+    imageClassName = `screenshot${link.screenshot ? " active" : ""}`;
+    imageStyle = {backgroundImage: link.screenshot ? `url(${link.screenshot})` : "none"};
+  }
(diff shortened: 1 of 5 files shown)


system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -5,6 +5,41 @@ const LinkMenu = require("content-src/components/LinkMenu/LinkMenu");
 
 const {TOP_SITES_SOURCE, TOP_SITES_CONTEXT_MENU_OPTIONS} = require("./TopSitesConstants");
 
+const TopSiteLink = props => {
+  const {link} = props;
+  const topSiteOuterClassName = `top-site-outer${props.className ? ` ${props.className}` : ""}`;
```
rlr: crazy that you can have template literal inside a template literal.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -4,22 +4,27 @@ const {FormattedMessage} = require("react-intl");
...
-        link={link}
-        index={index}
-        intl={props.intl} />)}
-    </ul>
-    <TopSitesEdit {...props} />
-  </section>
-</TopSitesPerfTimer>);
+const TopSites = props => {
+  const realTopSites = props.TopSites.rows.slice(0, props.TopSitesCount);
+  const placeholderCount = props.TopSitesCount - realTopSites.length;
```
rlr: Do we no longer allow "holes" in the top sites? Say the user pinned something in the very last slot (#12) but only has 6 other top sites. What does the UI show for that? Does it collapse it? I guess I can test it out.

as-pine-proxy:
[Treeherder: a637b4d8a29f85b2e0dadfccae2843fed096640b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a637b4d8a29f85b2e0dadfccae2843fed096640b)

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -135,7 +144,7 @@
 
     .title {
       font: message-box;
-      height: 20px;
+      height: $top-sites-title-height;
```
Mardak: Was this change intentional? The spec shows about 67px from the bottom of the top site square to the top of the Pocket logo, but with this change, there's now about 72px. (I.e., the height was 10px shorter than the line height resulting in 5px closer square.)

before/expected: image vs image :now

Issue #2007, opened 2017-01-18
Mozilla Central about:home search is autofocused


sarracini:
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1382891
Split from #1974. When opening about:home it auto focuses to the search bar

Mardak:
Blocked on #2598

Mardak:
@bryanbell says the location bar should be focused when opening about:home. Or effectively there should be no difference between opening about:newtab and opening about:home.

Mardak:
So we will be doing this to avoid impacting any existing search metrics.

Mardak:
Looks like we can't use `autofocus` https://github.com/facebook/react/issues/3066

So we'll want to detect visibility to focus but only the first visibility after the latest mount to avoid stealing focus. There's existing `lib` and `components` that individually check visibility, so potentially that can be shared and exported from https://github.com/mozilla/activity-stream/blob/master/system-addon/content-src/lib/detect-user-session-start.js

Mardak:
We will be conservative and keep the existing focus-search-in-content behavior. But we will put this behind a pref so that we can experiment via Shield study during Beta 57 to keep focus in location bar.

Mardak:
We will be conservative and keep the existing focus-search-in-content behavior. But we will put this behind a pref so that we can experiment via Shield study during Beta 57 to keep focus in location bar.

The experiment will want to measure total search volume (from location bar, content, etc.) and probably go with the one that has most search.

Mardak:
I suppose we can use this pref name: `browser.newtabpage.activity-stream.aboutHome.autoFocus`

rlr:
0c188ea2a98aeca6e1abe113daa707c066940915

PR #3325, opened 2017-08-31
chore(search): Closes #3321 Change contentSearchUIController to better match existing tests

### LINKED ISSUE #3320: Create an Activity Stream search source for about:home
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1396272
Similar to https://github.com/mozilla/activity-stream/issues/2729
We will (probably) end up with two `ContentSearchUIController` instances that will be used to initiate searches depending if you are on about:home or about:newtab.
Any thoughts for how it would be best to implement this?

### LINKED ISSUE #3321: Change contentSearchUIController to better match existing tests
Basically a lot the tests need a gContentSearchController to be exposed on window, and certain ids put in so that the tests can pick them up in browser_aboutHome.js
@k88hudson i know it's shitty to add it to window, but how would you feel about something like this: https://gist.github.com/sarracini/047ad959a1e896d11e308e5522a2ee3b

sarracini:
Fix #3321.
1. Exposes  gContentSearchcontroller to window so that various tests in browser_aboutHome.js can find it (and removes it on unmount)
2. Adds a healthReportKey and a source depending on the document uri. see -> http://searchfox.org/mozilla-central/source/browser/base/content/abouthome/aboutHome.js#235-236

### DIFF
--- a/system-addon/content-src/components/Search/Search.jsx
+++ b/system-addon/content-src/components/Search/Search.jsx
@@ -4,6 +4,7 @@ const React = require("react");
 const {connect} = require("react-redux");
 const {FormattedMessage, injectIntl} = require("react-intl");
 const {actionCreators: ac} = require("common/Actions.jsm");
+const {IS_NEWTAB} = require("content-src/lib/constants");
 
 class Search extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


sarracini:
@k88hudson can i get your eyes on this too to make sure it looks ok

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -48,13 +54,15 @@ class Search extends React.Component {
         <span className="sr-only"><FormattedMessage id="search_web_placeholder" /></span>
       </label>
       <input
+        autoFocus="autofocus"
```
Mardak: we want this only for about:home

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -28,11 +31,14 @@ class Search extends React.Component {
       // this name, and can add the appropriate telemetry probes for search. Without the
       // correct name, certain tests like browser_UsageTelemetry_content.js will fail (See
       // github ticket #2348 for more details)
-      this.controller = new ContentSearchUIController(input, input.parentNode,
-        "newtab", "newtab");
+      const isDocumentNewtabOrHome = document.documentURI === "about:newtab";
```
Mardak: from a programmer "or", this should always be `true` ;) maybe just `isNewtab` but we'll want it available in `render` too

system-addon/content-src/components/Search/Search.jsx
Mardak: and only on first visible ?

sarracini: Ah yes you're right, I'll just remove this because it's broken anyways and there's a ticket to so autofocus anyways #2007

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -19,20 +20,25 @@ class Search extends React.Component {
...
-      // github ticket #2348 for more details)
-      this.controller = new ContentSearchUIController(input, input.parentNode,
-        "newtab", "newtab");
+      // The first parameter here is called the "healthReportKey" and needs
+      // to be "newtab" or "abouthome" so that BrowserUsageTelemetry.jsm knows to
+      // handle events with this name, and can add the appropriate telemetry
+      // probes for search. Without the correct name, certain tests like
+      // browser_UsageTelemetry_content.js will fail (See github ticket #2348 for more details)
+      const healthReportKey = IS_NEWTAB ? "newtab" : "abouthome";
+      const source = IS_NEWTAB ? "newtab" : "homepage";
```
Mardak: Let's name this `searchSource` and also have a comment describing that it's so search engine plugins can correctly attribute referrals especially.

system-addon/test/unit/content-src/components/Search.test.jsx
```diff
@@ -50,7 +50,7 @@ describe("<Search>", () => {
 
     wrapper.find(".search-button").simulate("click");
 
-    const {search} = wrapper.node.controller;
+    const {search} = window.gContentSearchController;
```
Mardak: I suppose this effectively tests that we place it on the global. Might make sense to have a test dedicated to just making sure `window.gContentSearchController` exists after mounting with a description of it being an expected global.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -19,20 +20,25 @@ class Search extends React.Component {
...
+      window.gContentSearchController = new ContentSearchUIController(input, input.parentNode,
```
Mardak: Just to be extra safe, an extra comment here saying this `gContentSearchController` global is to allow existing about:home tests to pass, but potentially could be renamed / refactored when only activity stream is the only about:home.

as-pine-proxy:
[Treeherder: b31e2ae99fbefe7bda8f36dd982f003707379cbe](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b31e2ae99fbefe7bda8f36dd982f003707379cbe)

PR #3326, opened 2017-08-31
fix(topstories): Replace Send Feedback link with privacy notice. Clos…

### LINKED ISSUE #3265: Replace "Send Feedback" link in TopStories info section with link to privacy notice

csadilek:
Fixes #3265.

@Mardak not sure you want to merge this before or after https://github.com/mozilla/activity-stream/pull/3274/ ... so included the "Privacy Policy" string here as well.

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -72,6 +72,7 @@ search_settings=Change Search Settings
 # LOCALIZATION NOTE (section_info_option): This is the screenreader text for the
 # (?) icon that would show a section's description with optional feedback link.
 section_info_option=Info
+section_info_privacy_notice=Privacy Notice
 
 # LOCALIZATION NOTE (welcome_*): This is shown as a modal dialog, typically on a
 # first-run experience when there's no data to display yet
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -52,13 +52,14 @@ const PREFS_CONFIG = new Map([
       // Use the opposite value as what default value the feed would have used
       hidden: !PREFS_CONFIG.get("feeds.section.topstories").getValue(args),
       learn_more_endpoint: "https://getpocket.cdn.mozilla.net/firefox_learnmore?src=ff_newtab",
+      provider_header: "pocket_feedback_header",
       provider_description: "pocket_feedback_body",
       provider_icon: "pocket",
       provider_name: "Pocket",
       read_more_endpoint: "https://getpocket.cdn.mozilla.net/explore/trending?src=ff_new_tab",
       stories_endpoint: `https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=2&consumer_key=$apiKey&locale_lang=${args.locale}`,
       stories_referrer: "https://getpocket.com/recommendations",
-      survey_link: "https://www.surveymonkey.com/r/newtabffx",
+      info_link: "https://www.mozilla.org/privacy/firefox/#pocketstories",
       topics_endpoint: `https://getpocket.cdn.mozilla.net/v3/firefox/trending-topics?version=2&consumer_key=$apiKey&locale_lang=${args.locale}`,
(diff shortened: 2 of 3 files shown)


locales/en-US/strings.properties
```diff
@@ -72,6 +72,7 @@ search_settings=Change Search Settings
 # LOCALIZATION NOTE (section_info_option): This is the screenreader text for the
 # (?) icon that would show a section's description with optional feedback link.
 section_info_option=Info
+section_info_privacy_policy=Privacy Policy
```
Mardak: Looks like it'll be "Privacy Notice"

system-addon/lib/ActivityStream.jsm
```diff
@@ -52,13 +52,14 @@ const PREFS_CONFIG = new Map([
       // Use the opposite value as what default value the feed would have used
       hidden: !PREFS_CONFIG.get("feeds.section.topstories").getValue(args),
       learn_more_endpoint: "https://getpocket.cdn.mozilla.net/firefox_learnmore?src=ff_newtab",
+      provider_header: "pocket_feedback_header",
       provider_description: "pocket_feedback_body",
```
Mardak: Do these options need to be here if there's a fallback in `SectionsManager`?

system-addon/lib/SectionsManager.jsm
```diff
@@ -27,9 +27,9 @@ const BUILT_IN_SECTIONS = {
     maxRows: 1,
     availableContextMenuOptions: ["CheckBookmark", "SaveToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl"],
     infoOption: {
-      header: {id: "pocket_feedback_header"},
-      body: {id: options.provider_description},
-      link: {href: options.survey_link, id: "pocket_send_feedback"}
+      header: {id: options.provider_header || "pocket_feedback_header"},
+      body: {id: options.provider_description || "pocket_feedback_body"},
```
Mardak: Similarly, should this just be the the thing we want and not check options?

Mardak:
I'll do the other strings after this PR merges

locales/en-US/strings.properties
csadilek: yeah, changing this to section_info_privacy_notice = Privacy Notice

system-addon/lib/ActivityStream.jsm
csadilek: I'd prefer that, yes. it will make it easier to customize. Just added the fallback to avoid rendering errors in case the id / config is missing.

Mardak: customization by whom?

csadilek: The requirement was to have the topstories component be customizable => to make it support other content providers. So, we need to have the description be customizable too as then it wouldn't be Pocket. Now, how could this be customized? That's a good question, but shield studies for certain locales comes to mind. Separate release channels. Not saying there's a plan for that, but it was asked of me.

csadilek:
@Mardak thanks, string updated.

system-addon/lib/ActivityStream.jsm
Mardak: That seems reasonable as long as the understanding is that it wouldn't be customization in addition to Pocket as it would replace Pocket. I suppose a related issue is that if it's a shield study that can only change the value of one pref, that would point to `sections.topstories` should always being `true`/enabled instead of the conditionally enabled based on geo/locale of Pocket. And the actual enabled would be set as part of the `option`.

as-pine-proxy:
[Treeherder: cd71c7b443595c2d558c0b486b62cb9d6270ac59](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=cd71c7b443595c2d558c0b486b62cb9d6270ac59)

PR #3332, opened 2017-09-01
feat(migration): Update top sites immediately when migration completed

### LINKED ISSUE #3266: Only default sites shown after importing from another browser
@Softvision-VladBacia says imported sites only appear after 15 minutes. If so, that makes it sound like the initial top sites query happens before import and saves that until the 15 minute timer.

Mardak:
Fix #3266. r?@piatra Dispatch new action on the same observed migration ended event where we set prefs to hide the message.


as-pine-proxy:
[Treeherder: b923869c8ef82318f656d3fc09e942a9f5a4ede4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b923869c8ef82318f656d3fc09e942a9f5a4ede4)

PR #3337, opened 2017-09-01
Placeholders for sections

k88hudson:
This also changes the behaviour of sections relative to `initialized` somewhat.

- When sections are first registered, `initialized` is  `false` unless the register action contains non-empty `rows` (in that case, `initialized` will be `true`). 
- When a section is *updated* with an action that includes the `rows` property, `initialized` is set to `true` whether or not rows are empty.

The result of this is that when the section first starts, it will show placeholders; it will only show the empty state if it receives an empty response and there is no data to show.

### DIFF
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -2,11 +2,13 @@ const React = require("react");
 const {connect} = require("react-redux");
 const {injectIntl, FormattedMessage} = require("react-intl");
 const Card = require("content-src/components/Card/Card");
+const {PlaceholderCard} = Card;
 const Topics = require("content-src/components/Topics/Topics");
 const {actionCreators: ac} = require("common/Actions.jsm");
 
 const VISIBLE = "visible";
 const VISIBILITY_CHANGE_EVENT = "visibilitychange";
+const CARDS_PER_ROW = 3;
 
 class Section extends React.Component {
   constructor(props) {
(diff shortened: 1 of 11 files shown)


system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -2,11 +2,13 @@ const React = require("react");
 const {connect} = require("react-redux");
 const {injectIntl, FormattedMessage} = require("react-intl");
 const Card = require("content-src/components/Card/Card");
+const {PlaceholderCard} = Card;
 const Topics = require("content-src/components/Topics/Topics");
 const {actionCreators: ac} = require("common/Actions.jsm");
 
 const VISIBLE = "visible";
 const VISIBILITY_CHANGE_EVENT = "visibilitychange";
+const CARDS_PER_ROW = 3;
```
AdamHillier: Is there any way we can detect screen width and base this value on that? Otherwise this can happen. It's quite an edge case though...

screen shot 2017-09-01 at 3 48 46 pm

k88hudson: Interesting. Yes, this could be done, I think I'd like to file a follow-up for it though

as-pine-proxy:
[Treeherder: 9827ad31f04295207d96595984f788377ff9f33a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=9827ad31f04295207d96595984f788377ff9f33a)

PR #3341, opened 2017-09-04
fix(places): Avoid impacting performance when listening to onItemAdded - switch to dateAdded to avoid an extra fetch (#3204)

### LINKED ISSUE #3204: Use dateAdded instead of lastModified to avoid onItemAdded bookmark fetches
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1392248

Potentially we could use `dateAdded` that is provided to the observer instead of doing a full fetch to get `lastModified`

Looks like the original usage was from the original bookmark code from #110, but it's unclear if using modified was actually important or just convenient.

Standard8:
Fix #3204. I switched to dateAdded as discussed, and hence we can now avoid the extra fetch.

### DIFF
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -88,19 +88,24 @@ class BookmarksObserver extends Observer {
    * @param  {int} dateAdded
    * @param  {str} guid      The unique id of the bookmark
    */
-  async onItemAdded(...args) {
+  onItemAdded(...args) {
     const type = args[3];
-    const guid = args[7];
     if (type !== PlacesUtils.bookmarks.TYPE_BOOKMARK) {
       return;
     }
-    try {
-      // bookmark: {bookmarkGuid, bookmarkTitle, lastModified, url}
-      const bookmark = await NewTabUtils.activityStreamProvider.getBookmark(guid);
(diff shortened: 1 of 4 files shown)


system-addon/lib/PlacesFeed.jsm
```diff
@@ -88,19 +88,24 @@ class BookmarksObserver extends Observer {
...
-      // bookmark: {bookmarkGuid, bookmarkTitle, lastModified, url}
-      const bookmark = await NewTabUtils.activityStreamProvider.getBookmark(guid);
-      this.dispatch({type: at.PLACES_BOOKMARK_ADDED, data: bookmark});
-    } catch (e) {
-      Cu.reportError(e);
-    }
+    const uri = args[4];
+    const bookmarkTitle = args[5];
+    const dateAdded = args[6];
+    const bookmarkGuid = args[7];
```
Mardak: I might normally suggest something like `[,,,type,uri…]` in some destructured assignment or in the parameter list, but silly webpack inserting unused lines causing istanbul test coverage to drop below 100% :(

as-pine-proxy:
[Treeherder: 5f51ff27af58f11ee346c42715d795fc00cac531](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5f51ff27af58f11ee346c42715d795fc00cac531)

PR #3359, opened 2017-09-06
Prerendering

k88hudson:
This patch adds prerendering in a few commits:

- a5356f7: This is just some basic refactoring to not crash in node (using `global` for globals, etc.)
- 2193572: This changes the state rehydration to work with a new action, `NEW_TAB_STATE_REQUEST`, instead of `NEW_TAB_LOAD`. This is needed for prerendering in order to ensure the first render matches the prerendered state.
- 45357fb: This adds the `buildmc:html` task, and all the basic prerendering tasks.
- 686518a: This adds a pref to determine whether or not the pre-rendered version should be used. I need to actually add a patch to `aboutNewTabService` to make use of this.
- 209f43dd: This is a small change to `<Topics />` to allow it to render but be empty. This is to ensure things don't move around when the pre-rendered version switches to React.


system-addon/common/PerfService.jsm
```diff
@@ -1,26 +1,26 @@
...
   usablePerfObj = Services.appShell.hiddenDOMWindow.performance;
-} else { // we must be running in content space
+} else if (typeof performance !== "undefined") {
+  // we must be running in content space
   usablePerfObj = performance;
+} else {
+  usablePerfObj = {
+    now() {},
+    mark() {}
+  };
```
dmose: I'm assuming this is needed so that it runs under node.  A comment indicating that would nice.

system-addon/content-src/lib/constants.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   // constant to know if the page is about:newtab or about:home
-  IS_NEWTAB: document.documentURI === "about:newtab"
+  IS_NEWTAB: global.document && global.document.documentURI === "about:newtab"
```
dmose: What is it that keeps this code running in the browser?  Webpack magic?  Do we assign global=Window somewhere?

system-addon/content-src/components/Base/Base.jsx
```diff
@@ -22,6 +23,9 @@ function addLocaleDataForReactIntl({locale}) {
 
 class Base extends React.Component {
   componentDidMount() {
+    // Request state after the first render
+    this.props.dispatch(ac.SendToMain({type: at.NEW_TAB_STATE_REQUEST}));
+
```
dmose: A comment explaining why we do this extra round trip (e.g. for pre-rendering), perhaps even in a bit of details, would likely be useful to future code readers.

system-addon/lib/NewTabInit.jsm
```diff
@@ -15,7 +15,7 @@ this.NewTabInit = class NewTabInit {
   onAction(action) {
     let newAction;
     switch (action.type) {
-      case at.NEW_TAB_LOAD:
+      case at.NEW_TAB_STATE_REQUEST:
```
dmose: This wants a unit test, I think.  Maybe a look at the code coverage numbers would show other changes that want one too?

system-addon/content-src/components/Base/Base.jsx
dmose: A unit test for this would be good too; https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md has an example using sinon that should be similar.

bin/render-activity-stream-html.js
```diff
@@ -0,0 +1,59 @@
...
+    <script src="${options.baseUrl}vendor/react-redux.js"></script>
+    <script src="${options.baseUrl}data/content/activity-stream.bundle.js"></script>
+  </body>
+</html>
+`;
+}
+
+function templateJs(state) {
+  return `// Note - this is a generated file.
+  window.gActivityStreamPrerenderedState = ${JSON.stringify(state, null, 2)};
```
dmose: This wants a comment describing what it's doing and why.  Probably in the rendered file so that people who find that artifact quickly see what is.

bin/render-activity-stream-html.js
```diff
@@ -0,0 +1,59 @@
...
+`;
+}
+
+const args = require("minimist")(process.argv.slice(2), {alias: {baseUrl: "b", title: "t", locale: "l", direction: "d"}});
```
dmose: This is a bit hard to skim.  Maybe break it up into two lines, and explain the magic constant of 2 in a comment?

bin/render-activity-stream-html.js
```diff
@@ -0,0 +1,59 @@
...
+const options = Object.assign({}, DEFAULT_OPTIONS, args || {});
+const {html, state} = prerender(options.locale);
+
+fs.writeFileSync(HTML_FILE_PATH, templateHTML(options)); // eslint-disable-line no-sync
+fs.writeFileSync(PRERENDERED_HTML_FILE_PATH, templateHTML(options, html)); // eslint-disable-line no-sync
+fs.writeFileSync(INITIAL_STATE_JS_FILE_PATH, templateJs(state)); // eslint-disable-line no-sync
+
```
dmose: It'd improve readaibility to just put /* eslint-disable no-sync */ at the top of the file.

system-addon/content-src/components/Base/Base.jsx
```diff
@@ -46,11 +54,18 @@ class Base extends React.Component {
...
+    if (!props.isPrerendered && !initialized) {
       return null;
     }
 
-    return (<IntlProvider key={locale} locale={locale} messages={strings}>
+    // Note: the key on IntlProvider must be static in order to not blow away
+    // all elements on a locale change (such as after preloading).
+    // See https://github.com/yahoo/react-intl/issues/695 for more info.
+    return (<IntlProvider key="STATIC" locale={locale} messages={strings}>
```
Mardak: This change seems to cause the search placeholder and migration texts to not appear on the first about:home. (You can also test with `./mach run about:newtab`) Changing back to key={locale} avoids this:
image

My guess is that `LOCALE_UPDATED` happens at an unexpected time?

Mardak: Looks like passing in a dummy `locale` prop, e.g., `<Search locale={locale} />`, makes it `render` again

system-addon/content-src/lib/constants.js
k88hudson: `global` is `window` in webpack, yes

Mardak:
I pushed to talos moving the activity-stream-prerendered.html to activity-stream.html:
https://treeherder.mozilla.org/perf.html#/compare?originalProject=mozilla-central&originalRevision=c959327c6b75cd4930a6ea087583c38b805e7524&newProject=try&newRevision=099db27cebe866024186aac8f4f0274c60839d09&framework=1&filter=7%20opt

Nothing significant so far although tabpaint shows 6% improvement but low confidence after 10 before and 10 after runs.

system-addon/data/content/activity-stream.html
```diff
@@ -1,24 +0,0 @@
...
-    <meta http-equiv="Content-Security-Policy-Report-Only" content="script-src 'unsafe-inline'; img-src http: https: data: blob:; style-src 'unsafe-inline'; child-src 'none'; object-src 'none'; report-uri https://tiles.services.mozilla.com/v4/links/activity-stream/csp">
-    <title></title>
-    <link rel="icon" type="image/png" id="favicon" href="chrome://branding/content/icon32.png"/>
-    <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
-    <link rel="stylesheet" href="resource://activity-stream/data/content/activity-stream.css" />
-  </head>
-  <body class="activity-stream">
-    <div id="root"></div>
-    <div id="snippets-container">
-      <div id="snippets"></div>
```
Mardak: We'll still need these `snippets` related div#ids, yeah?

Mardak:
Still nothing significant (and maybe even a slight tart regression?)

I also pushed to try with aboutHome enabled:
https://treeherder.mozilla.org/perf.html#/compare?originalProject=mozilla-central&originalRevision=c959327c6b75cd4930a6ea087583c38b805e7524&newProject=try&newRevision=0c03e6c544e7a8a222b775b009110932d2a16cf9&framework=1

system-addon/lib/NewTabInit.jsm
k88hudson: Currently this file isn't being tested (which is an oversight), do you mind if I file a follow-up for this?

dmose: Because of the high-value in getting this into the tree sooner rather than later, yes.

bin/render-activity-stream-html.js
```diff
@@ -0,0 +1,59 @@
...
+    <meta charset="utf-8">
+    <meta http-equiv="Content-Security-Policy-Report-Only" content="script-src 'unsafe-inline'; img-src http: https: data: blob:; style-src 'unsafe-inline'; child-src 'none'; object-src 'none'; report-uri https://tiles.services.mozilla.com/v4/links/activity-stream/csp">
+    <title>${options.title}</title>
+    <link rel="icon" type="image/png" id="favicon" href="chrome://branding/content/icon32.png"/>
+    <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
+    <link rel="stylesheet" href="${options.baseUrl}data/content/activity-stream.css" />
+    <link rel="icon" type="image/svg+xml" href="${options.baseUrl}img/newtab-icon.svg">
+  </head>
+  <body>
+    <div id="root">${isPrerendered ? html : ""}</div>${isPrerendered ? `
```
dmose: Having so much of the semantics come from punctuation, and specifically nested template strings, makes this a bit hard to follow.  Pull out the <script> for stream-initial-state into a const higher up in the script?

bin/render-activity-stream-html.js
```diff
@@ -0,0 +1,59 @@
+const fs = require("fs");
+const path = require("path");
+const prerender = require("./prerender");
```
dmose: Add a comment here about what this is and how it's built?

bin/render-activity-stream-html.js
```diff
@@ -0,0 +1,59 @@
...
+    <script src="${options.baseUrl}vendor/react-intl.js"></script>
+    <script src="${options.baseUrl}vendor/redux.js"></script>
+    <script src="${options.baseUrl}vendor/react-redux.js"></script>
+    <script src="${options.baseUrl}data/content/activity-stream.bundle.js"></script>
+  </body>
+</html>
+`;
+}
+
+function templateJs(state) {
```
dmose: JSDoc here for first-time readers would be helpful, mostly for the info about "state"

system-addon/content-src/activity-stream-prerender.jsx
```diff
@@ -0,0 +1,63 @@
...
+  if (locale) {
+    if (!allStrings[locale]) {
+      throw new Error(`Tried to get strings for ${locale} but none were found.`);
+    }
+    strings = allStrings[locale];
+  } else {
+    Object.keys(allStrings["en-US"]).forEach(key => {
+      strings[key] = " ";
+    });
+  }
```
dmose: @Mardak, does this need a spin-off bug to properly do individual string fallback correctly?

k88hudson: I think string fallbacks work decently well at the moment (they fall back to English for any missing strings), although there are some cases where dynamic string ids get injected (e.g. TopStories) that could be handled better on a case-by-case basis

system-addon/content-src/activity-stream-prerender.jsx
```diff
@@ -0,0 +1,63 @@
...
+      enabled: true,
+      icon: "pocket",
+      id: "topstories",
+      order: 1,
+      title: {id: "header_recommended_by", values: {provider: "Pocket"}},
+      read_more_endpoint: "https://getpocket.cdn.mozilla.net/explore/trending?src=ff_new_tab",
+      topics: [{name: "Must Reads", category: "World", url: "https://getpocket.com/explore/must-reads?src=ff_new_tab"}]
+    }
+  });
+
```
dmose: Since each of the above dispatches effectively replaces ones that currently come from chrome, but have a comment above each one talking about how they differ or any assumptions in the interest of future code maintainability.

dmose: Particularly keeping in mind that effectively this code may also need to be updated as the chrome code changes over time.

k88hudson: It's not really so much that it comes from chrome as that specific actions are needed to set properties in certain reducers

k88hudson: I can actually add tests to help with this

system-addon/content-src/activity-stream-prerender.jsx
```diff
@@ -0,0 +1,63 @@
+const React = require("react");
```
dmose: This file definitely wants unit tests.

system-addon/content-src/activity-stream.jsx
```diff
@@ -6,11 +6,19 @@ const initStore = require("content-src/lib/init-store");
 const {reducers} = require("common/Reducers.jsm");
 const DetectUserSessionStart = require("content-src/lib/detect-user-session-start");
 const {addSnippetsSubscriber} = require("content-src/lib/snippets");
+const {actionTypes: at, actionCreators: ac} = require("common/Actions.jsm");
 
 new DetectUserSessionStart().sendEventOrAddListener();
 
-const store = initStore(reducers);
+const store = initStore(reducers, global.gActivityStreamPrerenderedState);
```
dmose: The changes to this file want unit tests too.

system-addon/content-src/lib/init-store.js
```diff
@@ -49,20 +49,23 @@ const messageMiddleware = store => next => action => {
  * @param  {object} reducers An object containing Redux reducers
```
dmose: There should probably be one or two unit tests added here.

system-addon/common/PrerenderData.jsm
```diff
@@ -0,0 +1,42 @@
+const prefConfig = {
+  // Prefs listed with "invalidates: true" will prevent the prerendered version
+  // of AS from being used if their value is something other than what is listed
+  // here.
```
dmose: As we were discussing yesterday, some of these have minor consequences, which we currently believe are acceptable.  Documenting those consequences and tradeoffs here would help maintainability.

system-addon/content-src/activity-stream-prerender.jsx
dmose: That sounds good.

system-addon/test/unit/lib/PrefsFeed.test.js
```diff
@@ -41,4 +45,38 @@ describe("PrefsFeed", () => {
...
+      feed.onAction({type: at.INIT});
+      assert.calledWith(feed._prefs.set, PRERENDER_PREF_NAME, false);
+    });
+  });
+  describe("onPrefChanged prerendering", () => {
+    it("should not change the prerender pref if the pref is not included in PrerenderData.invalidatingPrefs", () => {
+      feed.onPrefChanged("foo123", true);
+      assert.notCalled(feed._prefs.set);
+    });
+    it("should set the prerender pref if one of the prefs in PrerenderData.invalidatingPrefs is changed", () => {
```
dmose: This wants to be split into two tests for the true -> false change and the false -> true change.

bin/render-activity-stream-html.js
dmose: Very nice; thanks!

bin/render-activity-stream-html.js
```diff
@@ -1,3 +1,4 @@
+ /* eslint-disable no-sync */  /* eslint-disable no-sync */  /* eslint-disable no-sync */
```
dmose: Three times?  :-)

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -82,5 +82,5 @@ class Search extends React.Component {
 }
 
 // initialized is passed to props so that Search will rerender when it receives strings
```
dmose: This comment needs to be updated too.

bin/render-activity-stream-html.js
```diff
@@ -29,6 +29,7 @@ const DEFAULT_OPTIONS = {
  */
 function templateHTML(options, html) {
   const isPrerendered = !!html;
+  const scriptForPrerendering = `\n<script src="${options.baseUrl}data/content/activity-stream-initial-state.js"></script>`;
```
dmose: nit: It'd be even more readable if you just embedded a new line in the string rather than using /n.

bin/render-activity-stream-html.js
```diff
@@ -44,8 +45,7 @@ function templateHTML(options, html) {
     <div id="root">${isPrerendered ? html : ""}</div>
     <div id="snippets-container">
       <div id="snippets"></div>
-    </div>${isPrerendered ? `
-    <script src="${options.baseUrl}data/content/activity-stream-initial-state.js"></script>` : ""}
+    </div>${isPrerendered ? `${scriptForPrerendering}` : ""}
```
dmose: Much nicer; thanks!

bin/render-activity-stream-html.js
dmose: Great!

k88hudson:
Ok, I filed a bunch of follow-ups and added them to this meta-bug: https://github.com/mozilla/activity-stream/issues/3372

system-addon/content-src/activity-stream-prerender.jsx
k88hudson: Done!

k88hudson:
Filed a bugzilla ticket for the aboutNewTabService changes that need to land in mc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1397875

system-addon/test/unit/activity-stream-prerender.test.jsx
```diff
@@ -0,0 +1,60 @@
...
+const {PrerenderData} = require("common/PrerenderData.jsm");
+
+describe("prerenderStore", () => {
+  it("should create a store", () => {
+    const store = prerenderStore();
+    assert.isFunction(store.getState);
+  });
+  it("should start uninitialized", () => {
+    const store = prerenderStore();
+    const state = store.getState();
```
dmose: In all of these tests, calling getState wants to be part of the verification block, not the execution block.

system-addon/test/unit/activity-stream-prerender.test.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+    assert.equal(state.App.initialized, false);
+  });
+  it("should set the right locale and strings", () => {
+    const strings = {foo: "foo"};
```
dmose: The above line wants to be in its own setup block.

dmose:
This looks great; r+, and feel free to fix any remaining nits without it going by me again.  Also, of the list of meta bugs, I assume that at least the rtl bug should be marked fx57 if things look visually wrong or jumpy on an RTL build.

system-addon/content-src/lib/init-store.js
```diff
@@ -58,7 +58,17 @@ module.exports = function initStore(reducers, initialState) {
   );
 
   if (global.addMessageListener) {
+    let didInitialize = false;
```
k88hudson: I used this instead of `App.initialized` to decouple this from the specifics of the reducer's logic, in case it ever gets changed.

system-addon/content-src/lib/init-store.js
```diff
@@ -47,22 +47,36 @@ const messageMiddleware = store => next => action => {
...
-    } catch (ex) {
-      console.error("Content msg:", msg, "Dispatch error: ", ex); // eslint-disable-line no-console
-      dump(`Content msg: ${JSON.stringify(msg)}\nDispatch error: ${ex}\n${ex.stack}`);
-    }
-  });
+  if (global.addMessageListener) {
+    let didInitialize = false;
+    global.addMessageListener(INCOMING_MESSAGE_NAME, msg => {
+      // We should discard any messages received before the initial rehydration
+      // message from Chrome, since that replaces the whole state anyway.
```
Mardak: nit: Firefox chrome is "chrome" ;)

Also, to get early lost `store.dispatch(ac.SendToMain({type: at.NEW_TAB_STATE_REQUEST})); `, on `INIT` here (i.e., activity stream just finished loading), re-dispatch STATE_REQUEST

system-addon/data/content/activity-stream.html
```diff
@@ -1,24 +0,0 @@
...
-<html lang="en-us" dir="ltr">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="Content-Security-Policy-Report-Only" content="script-src 'unsafe-inline'; img-src http: https: data: blob:; style-src 'unsafe-inline'; child-src 'none'; object-src 'none'; report-uri https://tiles.services.mozilla.com/v4/links/activity-stream/csp">
-    <title></title>
-    <link rel="icon" type="image/png" id="favicon" href="chrome://branding/content/icon32.png"/>
-    <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
-    <link rel="stylesheet" href="resource://activity-stream/data/content/activity-stream.css" />
-  </head>
-  <body class="activity-stream">
```
Mardak: as_load_location is looking for `body.activity-stream`

k88hudson:
@Mardak @dmose 

Ok, I added three more commits;

- 56df179 resolves the issue where a request can happen before activity stream has initialized; 
- c891f96 restores the class on the body element;
- 5614734 resolves the issue that @dmose and I talked that causes an empty locale to be rendered in the UI

dmose:
Do we need/want to run this through Talos again, given that the performance characteristics have probably been changed a bit by the race fixes?

Mardak:
This will need a patch to keep pine / m-c happy:
```diff
diff --git a/browser/base/content/test/static/browser_all_files_referenced.js b/browser/base/content/test/static/browser_all_files_referenced.js
--- a/browser/base/content/test/static/browser_all_files_referenced.js
+++ b/browser/base/content/test/static/browser_all_files_referenced.js
@@ -88,4 +88,8 @@ var whitelist = [
   {file: "resource://app/modules/NewTabWebChannel.jsm"},
 
+  // browser/extensions/activity-stream/data/content/activity-stream-prerendered.html
+  // will be used very soon by bug 1397875
+  {file: "resource://activity-stream/data/content/activity-stream-prerendered.html"},
+
   // layout/mathml/nsMathMLChar.cpp
   {file: "resource://gre/res/fonts/mathfontSTIXGeneral.properties"},
```

Mardak:
Looks like this test is failing with this PR:
browser/components/search/test/browser_searchEngine_behaviors.js

Some brief debugging points to it waiting for and never getting content search events:
https://searchfox.org/mozilla-central/source/browser/components/search/test/browser_searchEngine_behaviors.js#96-107

k88hudson:
Weird ok, I'll take a look

k88hudson:
@Mardak @dmose ok, 5d8448 fixes the race condition where Search was getting loaded to early for the event to fire, as well as the unreferenced file.

mozilla-central-patches/fix-search-test-race.diff
```diff
@@ -0,0 +1,21 @@
...
+   return ContentTask.spawn(browser, {}, async function(args) {
+-    return new Promise(resolve => {
+-      content.addEventListener("ContentSearchService", function listener(aEvent) {
+-        if (aEvent.detail.type == "State") {
+-          content.removeEventListener("ContentSearchService", listener);
+-          resolve();
+-        }
+-      });
+-    });
++    await ContentTaskUtils.waitForCondition(() => content.wrappedJSObject.gContentSearchController &&
```
dmose: It'd be more readable if each of the "content.wrappedJSObject" clauses had its own line, eslint-permitting.

dmose: The original style of test seems more robust, though I get that it wasn't quite right.  This fix seems ok for now, but I'd suggest also adding an XXX comment suggesting adding an event to the code or some such so that the test depends less on implementation details and is more useful for refactoring.

dmose:
The aboutNewTabService patch to firefox.js seems to have bit-rotted.  Re-create it, using -U2 so that it will be more robust and likely to last a bit longer?

mozilla-central-patches/fix-search-test-race.diff
k88hudson: I don't really see how this relies more on implementation details, given that that particular event has to be fired (and also in a particular order, i.e. after the listener is added in the test). Additionally, I kind of feel like using the event makes this test way harder to debug/step through 😜

dmose: OK, that's a reasonable enough rationale.

Mardak:
There's a similar browser_google_behavior that has a "promiseContentSearchReady" but isn't failing ?
https://searchfox.org/mozilla-central/search?q=promiseContentSearchReady

as-pine-proxy:
[Treeherder: 524b2f921428edc800ec2e33de4d85caf8719911](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=524b2f921428edc800ec2e33de4d85caf8719911)

Mardak:
I reverted with https://github.com/mozilla/activity-stream/commit/ade80335b78346a740dd43922ea8f1edd00ba3f2 part of https://github.com/mozilla/activity-stream/commit/84534e4555deb97f6049c16ed496d3da5600f5e5 that added  mozilla-central-patches/fix-search-test-race.diff now that the export is on m-c with https://hg.mozilla.org/mozilla-central/rev/57f1221bf914

PR #3361, opened 2017-09-06
Enhancements for content events

csadilek:
@k88hudson Can you take a look at this, please. As discussed, these are the two enhancements to eventing:

- Introducing a NEW_TAB_RENDERED event
- Introducing skipMain meta field to send events to content without updating global state


dmose:
What's the point of the NEW_TAB_RENDERED event?

k88hudson:
Still have a few things to look at, but I will get to this soon yep 👍

dmose:
@csadilek what will the NEW_TAB_RENDERED event be used for?  I'm curious to understand if this might want to be coalesced with existing code...

csadilek:
@dmose it was introduced so feeds have a reliable way of knowing / calculating how many new tabs have been rendered since the content was last updated. This is to implement both frequency caps and ensuring content is displayed at a certain frequency. I found the other NEW_TAB* events to not work reliably for these use cases (got too many events). I also require timing so that content of a tab can still be updated (in response to this event), if needed. All for consolidating, if there's overlap.

system-addon/common/Actions.jsm
```diff
@@ -39,6 +39,7 @@ for (const type of [
   "NEW_TAB_INIT",
   "NEW_TAB_INITIAL_STATE",
   "NEW_TAB_LOAD",
+  "NEW_TAB_RENDERED",
```
k88hudson: Can you rename this to `"NEW_TAB_REHYDRATED"` since that's closer to what it is?

system-addon/content-src/components/Base/Base.jsx
```diff
@@ -22,13 +23,17 @@ function addLocaleDataForReactIntl({locale}) {
 
 class Base extends React.Component {
   componentDidMount() {
+    this.sendNewTabRenderedEvent(this.props.App);
```
k88hudson: Maybe rename this function to `sendNewTabRehydrated`.

Also, you should probably use either `componentWillMount` + `componentWillUpdate` OR `componentDidMount` + `compontentDidUpdate` to be consistent with the timing

k88hudson:
Maybe add a comment or two to your helper function as well to describe why it's needed?

csadilek:
@k88hudson Thanks, changes applied. Can you take another look, please?

as-pine-proxy:
[Treeherder: 9fa8dd81271f31e6cdcff7376bdbce777fb64dad](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=9fa8dd81271f31e6cdcff7376bdbce777fb64dad)

PR #3375, opened 2017-09-07
fix(systemaddon): Fix issue related to topsites url editing. 

### LINKED ISSUE #3347: Editing a Top Site URL wrongly creates a new Top Site
https://bugzilla.mozilla.org/show_bug.cgi?id=1396835

[Affected versions]:
- Firefox 57.0a1 Build ID 20170904220027

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:

[Steps to reproduce]:
2. Click the "Edit" button from the Top Sites section.
3. Click the "Pencil" for one of the Top Sites.
4. Edit the URL and save.
5. Hover the previously edited Top site and observe the URL.

[Expected results]:
- The Top Site URL was edited.

[Actual results]:
- The Top Site URL was not edited and a new Top Site with the edited URL is created.

[Notes]:
- If you create two Top Sites in this way, when you dismiss one of them, both will be removed,
editing a top site url wrongly creates a new top site

piatra:
Closes #3347

Now when editing a topsite url the form also gets the link information to check for `isPinned` and to be able to call `unpin`.
One behavior that I noticed is that we need to call `refresh` on the TopSites feed after these changes because adding or removing pins can influence deduping results. If you don't call refresh topsites will "randomly" disappear after 15 minutes or on browser restart.

f? @Mardak . I'll add tests after you 👍 / 👎  this solution.

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -110,7 +110,10 @@ class TopSite extends React.Component {
     }
   }
   onEditButtonClick() {
-    this.props.onEdit(this.props.index);
+    this.props.onEdit({
+      index: this.props.index,
+      link: this.props.link
+    });
   }
   render() {
     const {props} = this;
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -110,10 +110,7 @@ class TopSite extends React.Component {
     }
   }
   onEditButtonClick() {
-    this.props.onEdit({
-      index: this.props.index,
-      link: this.props.link
-    });
+    this.props.onEdit(this.props.index);
   }
   render() {
     const {props} = this;
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -146,12 +146,12 @@ this.TopSitesFeed = class TopSitesFeed {
   pin(action) {
     const {site, index} = action.data;
     NewTabUtils.pinnedLinks.pin(site, index);
-    this.refresh();
+    this._broadcastPinnedSitesUpdated();
   }
   unpin(action) {
     const {site} = action.data;
     NewTabUtils.pinnedLinks.unpin(site);
-    this.refresh();
+    this._broadcastPinnedSitesUpdated();
   }
   _insertPin(site, index) {
     // Insert a pin at the given index. If that slot is already taken, we need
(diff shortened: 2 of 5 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -146,12 +146,12 @@ this.TopSitesFeed = class TopSitesFeed {
...
     const {site, index} = action.data;
     NewTabUtils.pinnedLinks.pin(site, index);
-    this._broadcastPinnedSitesUpdated();
+    this.refresh();
   }
   unpin(action) {
     const {site} = action.data;
     NewTabUtils.pinnedLinks.unpin(site);
-    this._broadcastPinnedSitesUpdated();
+    this.refresh();
```
Mardak: @rlr has said that we purposely avoid refresh on pins per @k88hudson's suggestion. Now with deduping, the new pinned site might cause some frecent or default sites to disappear when correctly refreshed.

I'm not sure if this means we should just call `refresh` or update `refresh -> getLinksWithDefaults` to conditionally fetch new links from `NewTabUtils`?

Mardak: What's the potential deduping? The user adds/edits in a frecent or default site? The top sites list could change after 15 minutes anyway from just regular browsing, so it's not wrong for things to change.

piatra: Yes, it's a matter of do we want to do it ourselves and update it instantly or let the 15 minutes pass.
This is more obvious when you change the URL but the dedupe key (hostname) is the same. If we call refresh ourselves we won't get 2 websites after closing the edit form just 1.
I'll remove the `refresh` calls and go forward with the solution. I can easily add them back in later.

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -110,7 +110,10 @@ class TopSite extends React.Component {
     }
   }
   onEditButtonClick() {
-    this.props.onEdit(this.props.index);
+    this.props.onEdit({
+      index: this.props.index,
+      link: this.props.link
```
Mardak: There's potential for wrong behavior if link is not the link at the index. I believe we should be able to look up the link via index from `TopSiteForm` with some `getState()`. Thoughts?

piatra: 👍

piatra:
Updated.

as-pine-proxy:
[Treeherder: f27b735e4feb76821bc6fded718ef02813125c31](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f27b735e4feb76821bc6fded718ef02813125c31)

Mardak:
Reverted as requested https://github.com/mozilla/activity-stream/issues/3347#issuecomment-329996134

PR #3380, opened 2017-09-08
fix (systemaddon): #3379 use dedupe and update highlights on TOP_SITES_UPDATED

### LINKED ISSUE #3379: Refactor topsites vs highlights deduping to use Dedupe and watch for TOP_SITES_UPDATED
Noted https://github.com/mozilla/activity-stream/pull/3369#discussion_r137668744

There is also a race condition where Highlights might check TopSites state and see there's no top sites, but it's just that top sites is slightly slower. Perhaps Highlights should just watch for `TOP_SITES_UPDATED`

rlr:
r? @Mardak 

fixes #3379

### DIFF
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -83,6 +83,7 @@ this.HighlightsFeed = class HighlightsFeed {
         break;
       case at.PLACES_BOOKMARK_ADDED:
       case at.PLACES_BOOKMARK_REMOVED:
+      case at.TOP_SITES_UPDATED:
         this.fetchHighlights(false);
         break;
       case at.UNINIT:
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -13,6 +13,7 @@ const {actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.j
 
 const {shortURL} = Cu.import("resource://activity-stream/lib/ShortURL.jsm", {});
 const {SectionsManager} = Cu.import("resource://activity-stream/lib/SectionsManager.jsm", {});
+const {Dedupe} = Cu.import("resource://activity-stream/common/Dedupe.jsm", {});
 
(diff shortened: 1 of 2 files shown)


system-addon/lib/HighlightsFeed.jsm
```diff
@@ -77,6 +83,7 @@ this.HighlightsFeed = class HighlightsFeed {
         break;
       case at.PLACES_BOOKMARK_ADDED:
       case at.PLACES_BOOKMARK_REMOVED:
+      case at.TOP_SITES_UPDATED:
         this.fetchHighlights(false);
```
Mardak: Hrmm... if we're seeing a TOP_SITES_UPDATED, it will probably cause content to show the new top sites, so then we should maybe broadcast too? Otherwise without broadcast, the highlights will still show some things that should have been deduped.. maybe?

rlr: Yeah, I went back and forth on it. I guess we can broadcast because highlights are below the fold anyway so it wont move things around for the user?

Mardak: Oh. And also, TopSitesFeed is watching for a lot of these same actions to then refresh. Seems like we might double fetchHighlights

Mardak: So if we don't broadcast, we're okay.. ?

Mardak: The way I'm testing right now is bookmarking a default top site. On first run, it shows the bookmark and next new tab has it removed.

rlr: well hmm. So if the user starts editing their top sites, we will potentially move the highlights around for them while they do that. That seems annoying, but it's probably not visible to them unless they dont have the Top Stories section. So I guess I could go either way.

as-pine-proxy:
[Treeherder: 700d1cff42b395e1666b44f1055b265f454aaf7b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=700d1cff42b395e1666b44f1055b265f454aaf7b)

Issue #3131, opened 2017-08-10
Pocket story bookmark status is not persisted in context menu


sarracini:
STR:
1 .Open nightly (ensure Pocket is showing some recommended stories)
2. Bookmark a card via the context menu
3. Open the context menu of that same card and observe that it now says "Remove Bookmark". Open Bookmarks panel and confirm bookmark has been added
4. Quit nightly
5. Open nightly and open the context menu of the card you previously bookmarked

Expected results:
The context menu says "Remove Bookmark" and it should not allow you to bookmark the same card again

Actual results:
The context menu says "Bookmark" and gives you the option to bookmark the same site again. Actually clicking the context menu option to bookmark the card actually puts a new entry in the bookmarks library, so now you have duplicate bookmarks.

piatra:
I've looked into this and what happens is that the story feed is subscribed to bookmark events through the `Sections` reducer `case at.PLACES_BOOKMARK_ADDED:`. When you bookmark a story it will update the state of the story (isBookmarked = true), but the information is lost on restart.
I don't think we have any mechanism to persist that. My proposed solution is to use the urls and `executePlacesQuery` to fetch bookmarkGuids from PlacesProvider, if they exist. Whatever we use as `guid` for the top stories now is not actually a valid guid (it's just the story id) for querying using `getBookmark` from `NewTabUtils`.

Mardak:
As noted in https://github.com/mozilla/activity-stream/pull/3381#issuecomment-328110448
> Design says reflecting "bookmarked" for pocket stories makes sense. And the behavior is that top stories show "Bookmarked" immediately without moving. On refresh, that card no longer shows up in Top Stories and appears in Highlights (with image and description and Bookmarked).

piatra:
@Mardak right now the query for fetching recent bookmarks is [filtering out](https://searchfox.org/mozilla-central/rev/00fa5dacedb925022f53d025121f1a919508e7ce/toolkit/modules/NewTabUtils.jsm#866) bookmarks with no visits. Do we want to remove this constrain?

Mardak:
Ahhh.. doh. Well the query explicitly looks for visited pages. I think we'll need to do what you ended up doing for the test pilot version of filtering out `min(profile creation date + buffer, 5 day recent)`

Not sure if it'll be any better to use `await new (resource://gre/modules/ProfileAge.jsm).created` vs doing a places query…

Mardak:
[...]

PR #3381, opened 2017-09-08
fix (systemaddon): #3378 update contextmenu on unbookmark/bookmark

### LINKED ISSUE #3378: Highlight context menu bookmark/unbookmark doesn't update context menu or card type
We purposely don't refresh highlights on bookmark add/remove, but it looks like the context menu should be updating. Perhaps the Reducer is failing to update (at least on-bookmark -- screenshot below seems to indicate on-remove-bookmark it updated).

Design has previously said to treat a visited old bookmark as "Visited" but allow removing the bookmark.

It's unclear how the context menu and card type should behave when bookmarking/removing bookmark.

Here's a screenshot of what was a bookmark but just removed it from the context menu:
image

rlr:
@Mardak this mostly fixes it for me. But I am still seeing weird behavior sometimes, or it's just getting late.

Fixes #3378

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -237,6 +237,9 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           if (item.url === action.data.url) {
             const {bookmarkGuid, bookmarkTitle, dateAdded} = action.data;
             Object.assign(item, {bookmarkGuid, bookmarkTitle, bookmarkDateCreated: dateAdded});
+            if (!item.type || item.type === "history") {
+              item.type = "bookmark";
+            }
           }
           return item;
         })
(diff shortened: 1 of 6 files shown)


system-addon/common/Reducers.jsm
```diff
@@ -253,6 +253,9 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
             delete newSite.bookmarkGuid;
             delete newSite.bookmarkTitle;
             delete newSite.bookmarkDateCreated;
+            if (newSite.type === "bookmark") {
+              newSite.type = "history";
```
Mardak: We'll need to do the same for adding a bookmark.

rlr:
@Mardak ^^ updated.

I included the `if (!newSite.type || newSite.type === ...` guard because I didn't want trending cards in Top Stories to change, for example. Does that seem reasonable?

Mardak:
Design says reflecting "bookmarked" for pocket stories makes sense. And the behavior is that top stories show "Bookmarked" immediately without moving. On refresh, that card no longer shows up in Top Stories and appears in Highlights (with image and description and Bookmarked). I suppose this can be done as a followup in fixing #3131.

as-pine-proxy:
[Treeherder: a85602b2597643fc17aa8f1983e5ac28780921e0](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a85602b2597643fc17aa8f1983e5ac28780921e0)

PR #3386, opened 2017-09-08
fix (systemaddon): #3377 dont skip refresh if already refreshing

### LINKED ISSUE #3377: Early new tab pages might not have top sites
I believe this regressed by #3353. I think it's because a specific page loaded (`NEW_TAB_LOAD`) and another page was refreshing. This causes `refresh` to return early and never dispatch back to that desired target.

@AdamHillier @rlr 

screen shot 2017-09-07 at 3 29 17 pm

rlr:
This is one (simple) way to fix #3377 the problem.

The downside is that TippyTop provider can be initialized twice in some cases, but that does no harm other than wasting a few cycles.

Another option is to not call `refresh()` on `INIT` and wait until `NEW_TAB_LOAD`. Downside of that is the first tab load might take more time.

There are more complex solutions. Like we could have some sort of list of targets that need to be refreshed when the current refresh is complete. 

@AdamHillier @Mardak thoughts?

### DIFF
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -81,9 +81,6 @@ this.TopSitesFeed = class TopSitesFeed {
     return pinned.slice(0, TOP_SITES_SHOWMORE_LENGTH);
   }
   async refresh(target = null) {
-    if (this.isRefreshing) { return; }
-    this.isRefreshing = true;
-
     if (!this._tippyTopProvider.initialized) {
       await this._tippyTopProvider.init();
     }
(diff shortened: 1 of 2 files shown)


AdamHillier:
This seems reasonable as long as there isn't a problem with initialising twice. I mean the actual fix in my patch was ensuring initialisation happens before refresh, which this doesn't affect.

Mardak:
Well, the fix here in addition to having TippyTop initialize twice, it also causes refresh to happen twice (or each additional refresh+init). We can relatively easily fix the multi-init by having init know it's initializing and return a promise for when the initial init finishes, but this still has all `refresh` callers fully refresh by calling `getLinksWithDefaults`.

I'll think about this a bit more…

as-pine-proxy:
[Treeherder: ee98b2e502e0d4da5cdcd304f956460d098d44b8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=ee98b2e502e0d4da5cdcd304f956460d098d44b8)

PR #3389, opened 2017-09-08
fix: Remove testpilot code

AdamHillier:
I have been reasonably aggressive with removing stuff so thorough review is quite important. All tests and scripts are working which is a positive sign.

Before merging we should create a "legacy" fork that has the TP code on it.

Follow-ups to be filed:
- Sass linting was pointing to TP, so wasn't working. Pointing it to /system-addon throws up a fair few errors that need to be fixed.
- Documentation (especially the `README`) needs to be updated to reflect the fact that AS is now a system addon.


k88hudson:
Alright, this looks good, before you land this just:
- move the v2 docs back to their original location for now (since there are links pointing to them)
- file a follow-up for re-enabling sass linting

as-pine-proxy:
[Treeherder: 2fb432ffa745b1a80c081ac2c7fc131bd82ee959](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2fb432ffa745b1a80c081ac2c7fc131bd82ee959)

PR #3395, opened 2017-09-08
feat(topsites): Set Top Site card image source to high-res icon when available

### LINKED ISSUE #3327: Set Top Site card image source to high-res icon when available
Blocked on #3113 

This is related to #3150 , and these two tickets will finish the 'parity with Test Pilot add-on' story for Top Sites.  

If a high-res icon is available for this site in Places, then make sure it is set as the image source for the Top Site card.

sarracini:
Fix #3327. Generalize the icons so that it's just a rich icon instead of a screenshot

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -3,19 +3,19 @@ const {actionCreators: ac, actionTypes: at} = require("common/Actions.jsm");
 
 const LinkMenu = require("content-src/components/LinkMenu/LinkMenu");
 
-const {TOP_SITES_SOURCE, TOP_SITES_CONTEXT_MENU_OPTIONS} = require("./TopSitesConstants");
+const {TOP_SITES_SOURCE, TOP_SITES_CONTEXT_MENU_OPTIONS, MIN_FAVICON_SIZE} = require("./TopSitesConstants");
 
 const TopSiteLink = props => {
   const {link} = props;
   const topSiteOuterClassName = `top-site-outer${props.className ? ` ${props.className}` : ""}`;
-  const {tippyTopIcon} = link;
+  const {tippyTopIcon, faviconSize} = link;
   let imageClassName;
   let imageStyle;
-  if (tippyTopIcon) {
-    imageClassName = "tippy-top-icon";
+  if (tippyTopIcon || faviconSize >= MIN_FAVICON_SIZE) {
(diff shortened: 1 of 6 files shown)


Mardak:
The m-c change to provide favicon size has landed, and even without nan's fetch high resolution icons, some sites already put a large icon as their usual favicon (`<link rel="icon"/>`):

Before:
screen shot 2017-09-10 at 9 07 34 am

After:
screen shot 2017-09-10 at 9 07 46 am

So this PR seems like it can be reviewed now. (Looks like we want to set a background when using icons in case they're transparent. Or maybe hide the letter fallback.)

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -129,7 +129,7 @@
       }
     }
 
-    .tippy-top-icon {
+    .rich-icon {
       position: absolute;
```
Mardak: Might be okay to just put a `background-color: white` here? Checking with design…

image

Mardak: Might be okay to just put a `background-color: white` here?

image

Or use `$background-primary` (grey-10 #f9f9fa):
image

As long as it hides the fallback ;)
screen shot 2017-09-10 at 9 07 46 am

@aaronrbenson ?

aaronrbenson: I lean towards the grey-10 option here just in case there's a white icon. It's not perfect but at least you could make out that something is there.

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -8,14 +8,14 @@ const {TOP_SITES_SOURCE, TOP_SITES_CONTEXT_MENU_OPTIONS} = require("./TopSitesCo
 const TopSiteLink = props => {
   const {link} = props;
   const topSiteOuterClassName = `top-site-outer${props.className ? ` ${props.className}` : ""}`;
-  const {tippyTopIcon} = link;
+  const {tippyTopIcon, faviconSize} = link;
   let imageClassName;
   let imageStyle;
-  if (tippyTopIcon) {
-    imageClassName = "tippy-top-icon";
+  if (tippyTopIcon || faviconSize >= 96) {
```
k88hudson: Maybe make this a constant like `MIN_FAVICON_SIZE`

Issue #3394, opened 2017-09-08
Update README


AdamHillier:
It still refers to the testpilot version of AS

Mardak:
Fixed by #3637

PR #3401, opened 2017-09-09
add separators to README.md file

tmm88:
**this PR does basically aim at:**
- *adding separators to the README.md file*

---

**main motivation for that does basically lie within:**

- *making file easier to be read*;
- *making file simpler to be read*;
- *making file more direct to be read*;
- *making file easier to be understood*;
- *making file simpler to be understood*;
- *making file more direct to be understood*;

---

**it does also aim at beautifying the file**

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,15 +5,21 @@
 
 This is the source code for the Activity Stream project, which replaces the "New Tab" of Firefox with a new design based on rich metadata and browsing behavior.
 
+---
+
 ## Installing
 
 To install the release version of the add-on, check out [Test Pilot](https://testpilot.firefox.com/experiments/activity-stream).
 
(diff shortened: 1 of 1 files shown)


Mardak:
FYI, the `## Header`

## Header

already has a line underneath.

But not so for `### Subheader`

### Subheader

Not sure if we want to additionally split up with separators. @k88hudson ?

k88hudson:
In this case I don't think the separators are needed since, as @Mardak pointed out, headings already have underlines.

Also, we are replacing this README file very soon as it refers to the old testpilot experiment (see https://github.com/mozilla/activity-stream/issues/3394). Perhaps once that lands you could take a look and see if any readability improvements could be made to the new one?

PR #3404, opened 2017-09-10
feat(prerender): Compute rtl in Reducer for use in Base and prerendered

### LINKED ISSUE #3371: Support rtl for prerendered html
Right now the prerendered version is `ltr` only, we should support a `rtl` version.

Mardak:
Fix #3371. r?@k88hudson Adds support but doesn't actually generate the prerendered rtl file. (Currently working on doing it for each "mozilla-central locale".)


bin/render-activity-stream-html.js
```diff
@@ -97,6 +97,7 @@ function main() {
 
   const options = Object.assign({}, DEFAULT_OPTIONS, args || {});
   const {html, state} = prerender(options.locale);
+  options.direction = state.App.textDirection;
```
k88hudson: maybe remove the default / command line option if this is generated from the locale to prevent confusion?

Mardak: Removed `direction` from `minimist` and `DEFAULT_OPTIONS`

PR #3405, opened 2017-09-11
feature(telemetry): send correct session.page property (closes #2510) (WIP)

### LINKED ISSUE #2510: Set the right page property for telemetry from about:home
Right now the `page` property for all events in `TelemetryFeed.jsm` is hardcoded to `NEW_TAB`. When we support taking over `about:home`, we should implement logic to detect the right page. This should also update the search box telemetry.

dmose:
There is still a race condition left in here to solve.

### DIFF
--- a/system-addon/lib/ActivityStreamMessageChannel.jsm
+++ b/system-addon/lib/ActivityStreamMessageChannel.jsm
@@ -136,8 +136,8 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
     this.channel.addMessageListener(this.incomingMessageName, this.onMessage);
 
     // Some pages might have already loaded, so we won't get the usual message
-    for (const {loaded, portID} of this.channel.messagePorts) {
-      const simulatedMsg = {target: {portID}};
+    for (const {browser, loaded, portID} of this.channel.messagePorts) {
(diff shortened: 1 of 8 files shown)


system-addon/lib/ActivityStreamMessageChannel.jsm
```diff
@@ -136,8 +136,8 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
     this.channel.addMessageListener(this.incomingMessageName, this.onMessage);
 
     // Some pages might have already loaded, so we won't get the usual message
-    for (const {loaded, portID} of this.channel.messagePorts) {
-      const simulatedMsg = {target: {portID}};
+    for (const {browser, loaded, portID} of this.channel.messagePorts) {
+      const simulatedMsg = {target: {browser, portID}};
       this.onNewTabInit(simulatedMsg);
       if (loaded) {
```
Mardak: When https://bugzilla.mozilla.org/show_bug.cgi?id=1398955 lands, you can access `.url` from the message port. Probably clean this up to just do
```js
for (const target of this.channel.messagePorts) {
  const simulatedMsg = {target};
  …
  if (target.loaded) {
    …
```

dmose:
Taking a different approach

PR #3421, opened 2017-09-12
feature(telemetry): make session.page optional & support about:home (closes #2510)

### LINKED ISSUE #2510: Set the right page property for telemetry from about:home

dmose:
Fix #2510. This shouldn't be landed/tested until [bug 1398955](https://bugzilla.mozilla.org/show_bug.cgi?id=1398955) is merged to mozilla-central.

Once that happens, it's only easy to see that it's working with the first PingCentre commit in this PR (which should be removed before this patch is merged to master).

r? @ncloudioj and r? @emtwo, since this makes session.page optional in some cases.

### DIFF
--- a/docs/v2-system-addon/data_dictionary.md
+++ b/docs/v2-system-addon/data_dictionary.md
@@ -18,7 +18,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "addon_version": "1.0.0",
   "client_id": "374dc4d8-0cb2-4ac5-a3cf-c5a9bc3c602e",
   "locale": "en-US",
-  "page": "about:newtab or about:home",
+  "page": ["about:newtab" | "about:home" | "unknown"],
   "session_duration": 1635,
   "session_id": "{12dasd-213asda-213dkakj}",
   "user_prefs": 7
@@ -41,7 +41,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "client_id": "374dc4d8-0cb2-4ac5-a3cf-c5a9bc3c602e",
   "event": "click or scroll or search or delete",
   "locale": "en-US",
-  "page": "about:newtab or about:home",
+  "page": ["about:newtab" | "about:home" | "unknown"],
   "source": "top sites, or bookmarks, or...",
   "session_id": "{12dasd-213asda-213dkakj}",
   "recommender_type": "pocket-trending",
@@ -66,6 +66,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "event": "previewCacheHit",
   "event_id": "45f1912165ca4dfdb5c1c2337dbdc58f",
   "locale": "en-US",
+  "page": "unknown", // all session-specific perf events should be part of the session perf object
   "receive_at": 1457396660000,
   "source": "TOP_FRECENT_SITES",
   "value": 1,
@@ -88,7 +89,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "client_id": "26288a14-5cc4-d14f-ae0a-bb01ef45be9c",
   "event": "MISSING_IMAGE",
   "locale": "en-US",
-  "page": "about:newtab or about:home",
+  "page": ["about:newtab" | "about:home" | "unknown"]
   "source": "HIGHLIGHTS",
   "value": 0,
   "user_prefs": 7,
@@ -110,7 +111,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "addon_version": "1.0.12",
   "locale": "en-US",
   "source": "pocket",
-  "page": "about:newtab",
+  "page": ["about:newtab" | "about:home" | "unknown"]
   "tiles": [{"id": 10000}, {"id": 10001}, {"id": 10002}]
   "user_prefs": 7
 }
@@ -124,7 +125,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "addon_version": "1.0.12",
   "locale": "en-US",
   "source": "pocket",
-  "page": "about:newtab",
+  "page": "unknown",
   "user_prefs": 7,
 
   // "pos" is the 0-based index to record the tile's position in the Pocket section.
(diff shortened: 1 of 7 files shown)


Mardak:
I suppose to be clear, we can probably merge this PR and be no different than we are now where instead of always getting page="about:newtab", page will be unset until bug 1398955 makes it to m-c.

Mardak:
Bug 1398955 is in m-c: https://hg.mozilla.org/mozilla-central/rev/dad72fdc1cbf

dmose:
Awesome; thanks, Ed!

docs/v2-system-addon/data_dictionary.md
```diff
@@ -152,7 +152,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
 | `load_trigger_ts` | [Optional][Server Counter][Server Alert for too many omissions]  DOMHighResTimeStamp of the action perceived by the user to trigger the load of this page. | :one:
 | `load_trigger_type` | [Server Counter][Server Alert for too many omissions] Either ["menu_plus_or_keyboard", "unexpected"]. | :one:
 | `metadata_source` | [Optional] The source of which we computed metadata. Either (`MetadataService` or `Local` or `TippyTopProvider`). | :one:
-| `page` | [Required] Either ["NEW_TAB", "HOME"]. | :one:
+| `page` | [Optional] Either ["about:newtab", "about:home"]. | :one:
```
ncloudioj: Any reason for this change? Is it because it may fail to get the `url` in some cases?

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -387,6 +390,20 @@ this.TelemetryFeed = class TelemetryFeed {
   }
 
   /**
+   * Assign the given data to the session page property for the given port.
+   *
+   * @param {String} port  The session with which this is associated
+   * @param {String} data  A valid session.page value from pings.js
+   */
+  saveSessionPage(port, data) {
```
emtwo: I guess you plan to use this function in a future PR?

docs/v2-system-addon/data_dictionary.md
ncloudioj: If we really can't get the `url` in certain cases (whether expected or not), I'd recommend to keep it as it is, even simply set it as "unknown" or "n/a" or any thing more specific, and optionally send an undesired event ping for later issue diagnose.

Our database doesn't support `ALTER COLUMN`, unfortunately, switching it from `not null` to `nullable` needs a lot of effort for the data migration. Usually we would have to populate it with the default zero value before database inserts as a workaround.

dmose: @ncloudioj There are a couple of cases:

For some pings, it won't necessarily make sense to have one (e.g. PerfPings these days are likely to be mostly (entirely?) used to report on chrome performance where the cost is typically amortized across all the various pages and/or startup).  For that case, I think it'll be fairly straightforward to switch to "n/a" or something.

The bigger driver, though, is that the existing code sometimes creates pings without specifying what port they are attached to, meaning that we don't actually know, and up until now always just sent "about:newtab" because that was the only possibility..  This code is going to require some auditing/tweaking to handle these cases, but it shouldn't be too bad.  I'll give it a shot.

Out of curiosity, what database is it that doesn't support ALTER COLUMN?

ncloudioj: @dmose Oh, I see. BTW, the ETL jobs already treat `page` as an optional field for "perf", "undesired event", and "impression" pings.

> existing code sometimes creates pings without specifying what port they are attached to

Does this mean that we might not have `page` for the "session" and "event" pings *sometimes*?  We do check the existence of "page" for those two pings in the ETL. If so, we need to update the ETL rules for this change, i.e. use the default value "n/a" if it's missing.

> Out of curiosity, what database is it that doesn't support ALTER COLUMN?

We are using AWS Redshift, a column-oriented data warehouse. Unlike other DBMSs, it sacrifices some traditional features (e.g. ALTER COLUMN) to reach high capacity and query performance. Altering a column is usually done by "Rename the old col -> Create a new one -> Copy data over -> Drop the old one" :/

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -387,6 +390,20 @@ this.TelemetryFeed = class TelemetryFeed {
...
+    // XXX should use try/catch and send a bad state indicator if this
+    // get blows up.
```
ncloudioj: typo: s/get blows/blows/

ncloudioj:
This looks good to me. @dmose 

Looks like it's reasonable to have `page` as an optional field at least for "PerfPing" and "BadStates". For "Session" and "Event" pings, if we really don't know it, we can either populate the default value here, or do that in the ETL jobs, that's up to you ;). Although I do recommend sending a "bad state" ping with more context, since missing "page" could make the session/event analysis trickier, and ideally we'd like to make `page` available for all the session/event pings.

dmose:
Updated version of the patch which introduces an "unknown" value.  I'll file an issue to split out "n/a" at a later date.

system-addon/test/unit/lib/TelemetryFeed.test.js
```diff
@@ -104,10 +104,10 @@ describe("TelemetryFeed", () => {
 
       assert.propertyVal(session, "page", "about:monkeys");
     });
-    it("should not set the page property of no URL parameter is given", () => {
+    it("should set the page propto 'unknown' if no URL parameter given", () => {
```
ncloudioj: Is `propto` a real word? It's definitely beyond my vocabulary :)

dmose: Heh; will fix.

as-pine-proxy:
[Treeherder: 377bb9722bf1203ffd628654c1a5f5bdbe7dba63](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=377bb9722bf1203ffd628654c1a5f5bdbe7dba63)

dmose:
The unknown / n/a split has been spun off to #3442.

dmose:
@fmarier I forgot to get the data review before landing this.  It's just extending an existing piece of telemetry; there's nothing more personal to the new data bits.  If I could get your review on the data_*.md changes, that would be wonderful.  Thanks!

fmarier:
datareview+

dmose:
Wonderful; thanks!

PR #3422, opened 2017-09-12
feat(highlights): Filter out adult sites/pages but allow turning off via about:config

### LINKED ISSUE #1195: Filter NSFW cards for (unpinned) top sites and highlights
Eng bug for: https://trello.com/c/iDJ1IY5e

Need Engineering design and architecture. We should consider consuming a NSFW blacklist on the server.

Mardak:
Fix #1195. r?@sarracini Tested with some fake pinned links and fake metadata. In particular, even pinned links get filtered out, but if users really want it, they can flip the pref via about:config.

Before:
screen shot 2017-09-11 at 5 13 23 pm

After:
screen shot 2017-09-11 at 5 49 19 pm


system-addon/test/unit/lib/FilterAdult.test.js
```diff
@@ -0,0 +1,56 @@
...
+describe("filterAdult", () => {
+  let globals;
+  let filterAdult;
+  let hashStub;
+  let hashValue;
+
+  beforeEach(() => {
+    globals = new GlobalOverrider();
+    globals.set("Services", {
+      eTLD: {getBaseDomain({spec}) { return spec.match(/\/([^/]+)/)[1]; }},
```
sarracini: You should be able to add ```getBaseDomain``` to ```unit-entry.js```'s overrider of Services. You can also alter the existing one for ```newURI``` to return ```{spec: url}``` like you have here and it won't fail any of the other tests

Mardak: Aha! indeed. No need for a separate `GlobalOverrider` here now.

system-addon/lib/FilterAdult.jsm
```diff
@@ -0,0 +1,2974 @@
...
+      return true;
+    }
+  });
+};
+
+this.EXPORTED_SYMBOLS = ["filterAdult"];
+
+// These are md5 hashes of base domains to be filtered out. Originally from:
+// https://hg.mozilla.org/mozilla-central/log/default/browser/base/content/newtab/newTab.inadjacent.json
+gAdultSet = new Set([
```
pdehaan: These are a lot of strings. Is there any way to get the list from Firefox itself rather than copy-pasta the entire array of 2900 sites?

Mardak: Those strings dont exist in mozilla-central anymore.

Mardak:
Just got this:
> If a user Pins a site (even if it’s in adult category) we don’t apply NSFW filtering.

and

> my opinion here is that if they pinned it, it stays. The objective of this feature is to protect accidental/unwanted items that could be embarrassing from showing up on New Tab. If they add something to their Top Sites or pinned the thing, they can have it.

as-pine-proxy:
[Treeherder: 62be12d2c780592758977a6514612bf32bd3f11d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=62be12d2c780592758977a6514612bf32bd3f11d)

Issue #2841, opened 2017-07-10
visualize telemetry data for bad states


dmose:
This is the dashboard piece of #2672.

User Story:
> As a developer or product/eng manager, I should be able to see how often activity stream gets into a bad state in the field, as well as how those ratios (bad states / client, I suspect) change, so that regressions and improvements can be spotted.

@emtwo, your thoughts on dashboard design here would be most welcome

emtwo:
Tracked here now: https://github.com/mozilla/ping-centre/issues/88

PR #3424, opened 2017-09-12
Record bad states for Topsites and Highlights

### LINKED ISSUE #2672: telemetry data for bad states
We need to notice if topsites or highlights data is ready too late (i.e. later than the (first?) call to that component's render method).

Schema work is still in progress, and is happening in #2653 and https://github.com/mozilla/activity-stream/pull/2660/files

My guess is that we'll want to only report top_sites_data_ready_ts and highlight_data_ready_ts if they happen too late.

piatra:
Closes #2672


piatra:
@ncloudioj is there anything else that needs to go into the telemetry md documentation? 

I'll request data review once the code is reviewed, to make sure nothing will change.

docs/v2-system-addon/data_dictionary.md
```diff
@@ -134,6 +134,40 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "click|block|pocket": 0
 }
 ```
+# Example Activity Stream `undesired_event` Logs
+
+```js
+{
+  "topic": "activity-stream",
```
ncloudioj: `topic` is just a internal field for Ping-Centre, you can omit here.

docs/v2-system-addon/data_dictionary.md
```diff
@@ -134,6 +134,40 @@ Schema definitions/validations that can be used for tests can be found in `syste
...
+  "client_id": "8095a611-4ee4-b94d-b7b6-37923a0faa69",
+  "shield_id": "",
```
ncloudioj: ditto

docs/v2-system-addon/data_dictionary.md
```diff
@@ -167,10 +201,12 @@ and losing focus. | :one:
 | `value` | [Required] An integer that represents the measured performance value. Can store counts, times in milliseconds, and should always be a positive integer.| :one:
 | `ver` | [Auto populated by Onyx] The version of the Onyx API the ping was sent to. | :one:
 | `highlights_size` | [Optional] The size of the Highlights set. | :one:
+| `highlights_missing_data`| [Optional] Difference between data availability timestamp and first call to component render function. | :one:
```
ncloudioj: You don't need to document the values for the `event` key here, this is only for the `keys` in the ping.

docs/v2-system-addon/data_dictionary.md
```diff
@@ -167,10 +201,12 @@ and losing focus. | :one:
...
 | `topsites_first_painted_ts` | [Optional][Service Counter][Server Alert for too many omissions] Timestamp of when the Top Sites element finished painting (possibly with only placeholder screenshots) | :one:
 | `topsites_size` | [Optional] The size of the Topsites set. | :one:
 | `topsites_screenshot` | [Optional] The size of the Topsites set with screenshot metadata. | :one:
 | `topsites_tippytop` | [Optional] The size of the Topsites set with TippyTop metadata. | :one:
+| `topsites_missing_data`| [Optional] Difference between data availability timestamp and first call to component render function. | :one:
```
ncloudioj: ditto

ncloudioj:
The telemetry party looks good after fixing those nits. r=nanj

I'll leave the rest and data review to @dmose and @fmarier

docs/v2-system-addon/data_dictionary.md
dmose: @ncloudioj so where does that stuff get documented?

ncloudioj: You can document this under the corresponding sections in the `data_event.md`.

dmose:
Building wasn't working, because the mozilla-central-patches had bitrotted in a way that won't apply against current mozilla-center.  So I've rebased all the commits against current master and force-pushed.

dmose:
So I suspect this probably also occurs on master, and is unrelated to your patch, but it'd be good to verify:  when I start the browser and JS console, and then open a new tab, close it, and quite, I reproducibly see the following on stderr:

```
###!!! [Parent][MessageChannel] Error: (msgtype=0x150081,name=PBrowser::Msg_Destroy) Closed channel: cannot send/recv


###!!! [Child][MessageChannel] Error: (msgtype=0x150001,name=PBrowser::Msg_AsyncMessage) Closed channel: cannot send/recv

JavaScript error: chrome://global/content/browser-child.js, line 125: NS_ERROR_UNEXPECTED: Component returned failure code: 0x8000ffff (NS_ERROR_UNEXPECTED) [nsIMessageSender.sendAsyncMessage]
*** UTM:SVC TimerManager:registerTimer called after profile-before-change notification. Ignoring timer registration for id: telemetry_modules_ping
mozilla-29181:mozilla-central dmose$
```

I have no sense of whether this actually indicates a real problem.

dmose:
I suspect that you've already done some manual functional testing to verify that this works from end-to-end (eg with Console.jsm and console.log).  If not, please do so before landing.

docs/v2-system-addon/data_dictionary.md
```diff
@@ -135,6 +135,40 @@ Schema definitions/validations that can be used for tests can be found in `syste
...
+  "addon_version": "0.0.0",
+  "locale": "en-US",
+  "user_prefs": 3,
+  "session_id": "{ca150256-a6ca-4e43-a375-28ce5fa3bffc}",
+  "page": "about:newtab",
+  "value": 10,
+  "event": "highlights_missing_data",
+  "action": "activity_stream_undesired_event"
+}
+```
```
dmose: I'm thinking there's a bunch of value in, when possible, having the event names describe what the data is exactly without having to refer to the documentation.  For example, I think highlights_missing_data probably wants to be called something like "highlights_data_late_by_ms" or something similar.

system-addon/content-src/components/Sections/SectionsPerfTimer.jsx
```diff
@@ -0,0 +1,104 @@
...
+const RECORDED_SECTIONS = ["highlights"];
+
+class SectionsPerfTimer extends React.Component {
+  constructor(props) {
+    super(props);
+    // Just for test dependency injection:
+    this.perfSvc = this.props.perfSvc || perfSvc;
+
+    this._sendBadStateEvent = this._sendBadStateEvent.bind(this);
+    this._reportMissingData = false;
```
dmose: Maybe _shouldReportMissingData for clarity?

system-addon/content-src/components/Sections/SectionsPerfTimer.jsx
```diff
@@ -0,0 +1,104 @@
...
+   * @param {Function} callback
+   *
+   * @returns void
+   */
+  _afterFramePaint(callback) {
+    requestAnimationFrame(() => setTimeout(callback, 0));
+  }
+
+  _maybeSendPaintedEvent() {
+    if (RECORDED_SECTIONS.indexOf(this.props.id) === -1) {
```
dmose: Perhaps use .includes for improved readability?

system-addon/content-src/components/Sections/SectionsPerfTimer.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    // Only record first call to render.
+    if (this._timestampHandled) {
+      return;
+    }
+
+    this._timestampHandled = true;
+
+    const firstPaintKey = `${this.props.id}_first_painted_ts`;
+    this._afterFramePaint(() => this.perfSvc.mark(firstPaintKey));
```
dmose: I'm confused.  Doesn't this mean that topsites_first_painted_ts is going to be marked both in the Sections timer as well as in the Topsites timer?  Or is topsites not a Section?

dmose: I see now that topsites is not in the RECORDED_SECTIONS array, but if it were, would it simplify the patch substantially and allow some code removal?  Or is that too much code churn for this moment?

dmose:
You're totally right after the lameness of having to do the extra testing.  For the future, it might work to use mocha [shared behaviors](https://github.com/mochajs/mocha/wiki/Shared-Behaviours).  But I'm also open to the idea that we've hit the point of diminishing returns around copy-pasta, and that the next thing we need to time should just bite the bullet and implement the Higher-Order Component.

dmose:
Be sure to also file a [user story issue](https://github.com/mozilla/activity-stream/blob/master/docs/v2-system-addon/telemetry.md), although note that those now need to be filed in the mozilla/ping-centre repo (that doc needs to be updated).

dmose:
Overall, though, excellent work!  Thanks for pulling this together.

piatra:
I've updated the pull request
- [ ] now uses just one perf component that wraps both topsites and sections https://github.com/piatra/activity-stream/blob/bug-2672-telemetry-bad-state/system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
- [ ] Changed the event names to `*_data_late_by_ms`
- [ ] Moved the documentation in it's right file (data_events) and updated the checklist to reflect that the user story should be submitted to the ping-centre repo and submitted the story for this https://github.com/mozilla/ping-centre/issues/98

system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
```diff
@@ -64,40 +66,66 @@ class TopSitesPerfTimer extends React.PureComponent {
...
-    // sending back the wrong data, and we have to do it here, so that other
-    // calls to this method while waiting for the next frame won't also try to
-    // handle handle it.
     this._timestampHandled = true;
-
     this._afterFramePaint(this._sendPaintedEvent);
   }
 
+  _sendBadStateEvent() {
+    const dataReadyKey = `${this.props.id}_data_ready_ts`;
```
Mardak: nit: for each of these dynamically generated keys, could you add a comment above with the expected names:
`// topsites_data_ready_ts, highlights_data_ready_ts`

I know it seems a bit silly, but it's so that it's easier to find/`grep` for where these id/keys are made. Where someone not familiar with the code wouldn't know to search for `_data_ready_ts`.

dmose:
This is looking good.  I'd get the details after you rejigger things.  

Looking at the code, I wish the tests had been written more using the outward facing API to support future refactoring (i.e. just permuting props and using fakes), but that's totally on me, as I didn't write it that way originally.  Furthermore, now is clearly not the moment to spend rewriting them for that.

So I look forward to your new patch which preserves the existing semantics as discussed.

piatra:
@dmose updated the patch. Added a new event `*_first_render_ts` called on the very first `render()` call. This is used to compute how long it takes for data to be available for the component. 
Previous metric of `*_first_painted_ts` is now working as before.

piatra:
@dmose Updated again. Moved recording into componentDidUpdate, and now handled by a separate method.

piatra:
@dmose Updated: I split the maybeSendBadState method into two, one does the first render mark, the other actually sends the bad state if necessary. The two actions didn't really need to be together and that simplifies the logic. There wasn't much else I could think about changing.

piatra:
@dmose 
* not sure about the changes in `data_dictionary.md`. [I added them before and was asked to remove them](https://github.com/mozilla/activity-stream/pull/3424#discussion_r138919978).
* [ping center already fixed the issue/story for this PR](https://github.com/mozilla/ping-centre/issues/98)

Working on the rest.

docs/v2-system-addon/data_events.md
```diff
@@ -379,3 +379,38 @@ This reports when the addon fails to initialize
   "value": -1
 }
 ```
+
+### Late data initialization
+
+This reports how long it took to get data ready to be displayed.
+From the moment the component first receives default state of `initialized: false`
+until this becomes `true`.
```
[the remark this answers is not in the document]
piatra: @dmose The value can only be positive so I took that part out.

dmose:
OK, you're right about data dictionary; leaving those changes out is fine.

system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
```diff
@@ -103,6 +103,7 @@ class ComponentPerfTimer extends React.Component {
    * Creates `TELEMETRY_UNDESIRED_EVENT` with timestamp in ms
    * of how much longer the data took to be ready for display than it would
    * have been the ideal case.
+   * https://github.com/mozilla/ping-centre/issues/98
```
dmose: So that issue appears to be where they ensured that the tables are available to receive the data.  The actual user story was originally at mozilla/activity-stream#2841 and ultimately moved to mozilla/ping-centre#88.  If it's worth linking to anything from the code, (I have mixed feelings), it's probably that last one.

docs/v2-system-addon/telemetry.md
```diff
@@ -2,7 +2,7 @@
 
 Adding telemetry generally involves a few steps:
 
-1. - [ ] File a "user story issue" about who wants what question answered.  This will be used to track server-side data handling implementation (ETL, dashboard implementation...).
+1. - [ ] File a "user story issue" [in ping-centre](https://github.com/mozilla/ping-centre) about who wants what question answered.  This will be used to track server-side data handling implementation (ETL, dashboard implementation...).
```
dmose: Thanks for fixing this!

piatra:
r? @fmarier 
This PR adds 2 pings for TopSites and Highlights (only if the user has those sections enabled) that send the amount of time in ms until data is being displayed. [A more in-depth explanation](https://github.com/mozilla/activity-stream/pull/3424/files#diff-f9d20d6688eef2754298456e5d5b836bR383) in the updated readme with payload example.
The pings are sent for every new session. As long as `browser.newtabpage.activity-stream.telemetry` is `true`.

fmarier:
The new pings are Category 1 data.

datareview+

PR #3427, opened 2017-09-12
fix(highlights): Add image caching and image placeholders.

### LINKED ISSUE #2322: Add image caching for preview image metadata
This is to avoid doing network requests when rendering the page. Maybe use screenshot backend to "screenshot" the image and request the screenshotted image.

### LINKED ISSUE #3406: Show placeholder image for Highlight bookmarks without images
https://bugzilla.mozilla.org/show_bug.cgi?id=1398715

[Notes]:
- The bookmarks have to have been created in a Nightly version that did not have Highlights.
- The issue is also reproducible if you import History from another browser in a new profile.

[Affected versions]:
- Firefox 57.0a1 Build ID 20170910220126

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- browser.newtabpage.activity-stream.enabled is set to true in about:config on a clean new profile of Nightly 2017-09-07.
- Have enough history so that all 12 Top Sites are populated.

[Steps to reproduce]:
1. Start the browser with the profile from prerequisites.
2. Navigate to three pages and bookmark them.
3. Update Nightly.
4. Open a New Tab and observe the first three cards from the Highlights section.

[Expected results]:
- The cards have an image, a title and a short description of the bookmarked page.

[Actual results]:
- The cards have titles, but lack descriptions and images.

[Additional Notes]:
bookmarks with no images after update

AdamHillier:
Closes #2322. Closes #3406.


Mardak:
Design says to have a blended border that shares the gradient. I mocked it quickly with a box-shadow inset as borders go outside of the image:

`box-shadow: inset 0 -1px 1px rgba(0, 0, 0, 0.05)`

pasted image at 2017_09_12 01_12 pm
pasted image at 2017_09_12 01_15 pm

Design says ideally no extra shadow creeping in from left/right edges but if that can't be fixed, then it's still better than the solid grey-30 border along the bottom.

Mardak:
Oh, here we go with no side shadow:
```css
.card-outer .card-preview-image-outer::after {
  border-bottom: 1px solid rgba(0, 0, 0, 0.05);
  bottom: 0;
  content: " ";
  position: absolute;
  width: 100%;
}
```
With the `.card-preview-image-outer` parent being `position: relative;`

Mardak:
Oh, here we go with no side shadow:
```css
.card-outer .card-preview-image-outer::after {
  border-bottom: 1px solid rgba(0, 0, 0, 0.05);
  bottom: 0;
  content: " ";
  position: absolute;
  width: 100%;
}
```
With the `.card-preview-image-outer` parent being `position: relative;`

pasted image at 2017_09_12 01_56 pm

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -62,10 +65,16 @@ this.HighlightsFeed = class HighlightsFeed {
...
 
+      // If we already have the image for the card in the cache, use that
+      // immediately. Then asynchronously fetch the image (refreshes the cache).
+      const image = this.imageCache.get(page.url);
+      this.fetchImage(page.url, page.preview_image_url);
+
       // We want the page, so update various fields for UI
       Object.assign(page, {
+        image,
+        imageLoading: true, // We always have an image - fall back to a screenshot
```
Mardak: I would assume this `imageLoading` is for other sections to have just text? Let's just rename this to `hasImage` as it would seem to imply it should be `false` if it's already loaded.

system-addon/content-src/components/Card/Card.jsx
```diff
@@ -56,14 +56,18 @@ class Card extends React.Component {
...
+    const hasImage = link.image || link.imageLoading;
+    const imageStyle = {backgroundImage: link.image ? `url(${link.image})` : "none"};
 
     return (<li className={`card-outer${isContextMenuOpen ? " active" : ""}${props.placeholder ? " placeholder" : ""}`}>
       <a href={link.url} onClick={!props.placeholder && this.onLinkClick}>
         <div className="card">
-          {link.image && <div className="card-preview-image" style={{backgroundImage: `url(${link.image})`}} />}
-          <div className={`card-details${link.image ? "" : " no-image"}`}>
+          {hasImage && <div className="card-preview-image-outer">
+            <div className={`card-preview-image${link.image ? " active" : ""}`} style={imageStyle} />
```
Mardak: The use of "active" here is confusing as at least with css-related stuff, it's when the user is interacting with it. Probably name this "loaded".

system-addon/content-src/components/Card/Card.jsx
```diff
@@ -56,14 +56,18 @@ class Card extends React.Component {
     const isContextMenuOpen = this.state.showContextMenu && this.state.activeCard === index;
     // Display "now" as "trending" until we have new strings #3402
     const {icon, intlID} = cardContextTypes[link.type === "now" ? "trending" : link.type] || {};
+    const hasImage = link.image || link.imageLoading;
```
Mardak: With `imageLoading` renamed to `hasImage`, I think we should just rely on that value instead of additionally checking `image`, so `const {hasImage} = link;`

Mardak: Oh. Although I see you probably did this for backwards compatibility for other things passing in just `image`…

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -81,6 +90,20 @@ this.HighlightsFeed = class HighlightsFeed {
 
     SectionsManager.updateSection(SECTION_ID, {rows: this.highlights}, this.highlightsLastUpdated === 0 || broadcast);
     this.highlightsLastUpdated = Date.now();
+    this.imageCache.clear();
```
Mardak: What's the purpose of the cache if we clear it on every fetch? It seems like it would only be beneficial if multiple highlights ended up using the same image?

system-addon/lib/SectionsManager.jsm
```diff
@@ -139,6 +139,24 @@ const SectionsManager = {
...
+   *
+   * @param id              The id of the section with the card to be updated
+   * @param url             The url of the card to update
+   * @param options         The options to update for the card
+   * @param shouldBroadcast Whether or not to broadcast the update
+   */
+  updateSectionCard(id, url, options, shouldBroadcast) {
+    if (this.sections.has(id)) {
+      const card = this.sections.get(id).rows.find(elem => elem.url === url);
+      if (card) { Object.assign(card, options); }
```
Mardak: nit: put the `if` code block body on its own line

system-addon/lib/HighlightsFeed.jsm
AdamHillier: The image cache cleared at the end of every fetchHighlights once it's been used (in that method). The async fetchImage calls then fill up the cache again for the next fetchHighlights call. I think, though I could be wrong, that none of the fetchImage calls are executed before fetchHighlights has finished executing, so they fill up the cache after it has been cleared. Without the image cache I found that subsequent refreshes caused flickering of images for existing cards as the src was unset then reset again asynchronously.

Mardak:
Oh and we'll want to pass in backgroundColor of grey-10 #f9f9fa

system-addon/lib/HighlightsFeed.jsm
Mardak: Hrm.. That makes sense. Although it's also quite dangerous how it is right now. If something were to delay this `clear()` until after some images were fetched and set, then it's just throwing away that work.

The cache seems to be acting more as a prefetcher, but it could be a better actual cache if it only removed items that weren't used if the concern is holding on to fetched images too long. But this can be done in a followup bug.

At minimum better comment on why it's `clear`ing here and how the cache will still be useful.

system-addon/content-src/components/Card/Card.jsx
AdamHillier: Yeah because pocket for example doesn't set hasImage

as-pine-proxy:
[Treeherder: 11ca06b247d32862196a50d0051ac94a343938e3](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=11ca06b247d32862196a50d0051ac94a343938e3)

PR #3430, opened 2017-09-12
fix(prerender): Use prerendered if either pocket or highlights are on

k88hudson:
This patch also:
* Refactors PrerenderData to compute whether prefs are valid, allowing
  it to be used in content.
* Adds highlights to the prerendered initial sections
* Adds some styles that reduce the extent to which elements move around
  on the page: a new fixed-to-top class that gets added to outer-wrapper

### DIFF
--- a/system-addon/common/PrerenderData.jsm
+++ b/system-addon/common/PrerenderData.jsm
@@ -1,32 +1,73 @@
-const prefConfig = {
-  // Prefs listed with "invalidates: true" will prevent the prerendered version
+class _PrerenderData {
+  constructor(options) {
+    this.initialPrefs = options.initialPrefs;
+    this.initialSections = options.initialSections;
+    this._setValidation(options.validation);
+  }
+
+  get validation() {
+    return this._validation;
+  }
+
+  set validation(value) {
+    this._setValidation(value);
+  }
+
+  get invalidatingPrefs() {
+    return this._invalidatingPrefs;
+  }
+
+    // This is needed so we can use it in the constructor
+  _setValidation(value = []) {
+    this._validation = value;
+    this._invalidatingPrefs = value.reduce((result, next) => {
+      if (typeof next === "string") {
+        result.push(next);
+        return result;
+      } else if (next && next.oneOf) {
+        return result.concat(next.oneOf);
+      }
+      throw new Error("Your validation configuration is not properly configured");
+    }, []);
+  }
+
+  arePrefsValid(getPref) {
+    for (const prefs of this.validation) {
+      // {oneOf: ["foo", "bar"]}
+      if (prefs && prefs.oneOf && !prefs.oneOf.some(name => getPref(name) === this.initialPrefs[name])) {
+        return false;
+
+      // "foo"
+      } else if (getPref(prefs) !== this.initialPrefs[prefs]) {
+        return false;
+      }
+    }
+    return true;
+  }
+}
+
+this.PrerenderData = new _PrerenderData({
+  initialPrefs: {
+    "migrationExpired": true,
+    "showTopSites": true,
+    "showSearch": true,
+    "topSitesCount": 6,
+    "feeds.section.topstories": true,
+    "feeds.section.highlights": true
+  },
+  // Prefs listed as invalidating will prevent the prerendered version
   // of AS from being used if their value is something other than what is listed
   // here. This is required because some preferences cause the page layout to be
   // too different for the prerendered version to be used. Unfortunately, this
   // will result in users who have modified some of their preferences not being
   // able to get the benefits of prerendering.
-  "migrationExpired": {value: true},
-  "showTopSites": {
-    value: true,
-    invalidates: true
-  },
-  "showSearch": {
-    value: true,
-    invalidates: true
-  },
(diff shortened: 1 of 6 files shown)


system-addon/common/PrerenderData.jsm
```diff
@@ -1,32 +1,80 @@
...
+    }, []);
+  }
+
+  arePrefsValid(getPref) {
+    for (const name of this.validation) {
+      if (typeof name === "string") {
+        if (getPref(name) !== this.initialPrefs[name]) {
+          return false;
+        }
+      } else if (name && name.oneOf) {
```
Mardak: I suppose you could just make each item an array. A 1 item array means it must match. A multi-item array means any of them can match.

system-addon/content-src/components/Base/_Base.scss
```diff
@@ -1,8 +1,15 @@
...
   display: flex;
-  flex-grow: 1;
   padding: $section-spacing $base-gutter $base-gutter;
   height: 100%;
+  flex-grow: 1;
+
+  &.fixed-to-top {
+    height: auto;
+    @media (min-width: $break-point-large) and (min-height: 800px) {
+      padding-top: 10vh;
```
Mardak: Where did these come from? The existing outer-wrapper seems to have a top padding of 40px, which is `$section-spacing`

system-addon/common/PrerenderData.jsm
```diff
@@ -1,32 +1,80 @@
...
+    this.initialSections = options.initialSections;
+    this._setValidation(options.validation);
+  }
+
+  get validation() {
+    return this._validation;
+  }
+
+  set validation(value) {
+    this._setValidation(value);
```
Mardak: Not sure if we really need these getter/setters. But I suppose they're fine.

system-addon/common/PrerenderData.jsm
```diff
@@ -1,32 +1,80 @@
...
+  arePrefsValid(getPref) {
+    for (const name of this.validation) {
+      if (typeof name === "string") {
+        if (getPref(name) !== this.initialPrefs[name]) {
+          return false;
+        }
+      } else if (name && name.oneOf) {
+        let hasValid = false;
+        for (const innerName of name.oneOf) {
+          if (getPref(innerName) === this.initialPrefs[innerName]) {
```
Mardak: With every item being an array of either 1 item or multiple items, we could then just do
```js
for (const prefs of this.validation) {
  if (!prefs.some(name => getPref(name) === this.initialPrefs[name])) {
    return false;
  }
}
return true;
```

system-addon/common/PrerenderData.jsm
k88hudson: I feel like have the `oneOf` makes it more immediately obvious what the behaviour is though when you're just reading the file

system-addon/test/unit/common/PrerenderData.test.js
```diff
@@ -0,0 +1,98 @@
...
+      instance.validation = newValidation;
+      assert.equal(instance.validation, newValidation);
+      assert.deepEqual(instance.invalidatingPrefs, ["foo", "blah", "gloo"]);
+    });
+    it("should throw if an invalid validation config is set", () => {
+      // {stuff: []} is not a valid configuration type
+      assert.throws(() => new _PrerenderData({validation: ["foo", {stuff: ["bar"]}]}));
+    });
+  });
+  describe("#ca", () => {
```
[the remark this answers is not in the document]
k88hudson: oops

system-addon/common/PrerenderData.jsm
k88hudson: I figured this would be easier than having cacheing for `.invalidatablePrefs` since it's computed from `.validation`, but it's not strictly necessary if we don't care about modifying `.validation` after the constructing the instance

system-addon/content-src/components/Base/_Base.scss
k88hudson: I was playing around with some stuff to make it look better for larger screens, but I talked to aaron and he figures `height: auto` is fine so i'll remove it

as-pine-proxy:
[Treeherder: 69599cfe9df95258e75ed7c62eb5ae13fbcf73eb](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=69599cfe9df95258e75ed7c62eb5ae13fbcf73eb)

system-addon/common/PrerenderData.jsm
```diff
@@ -1,32 +1,73 @@
...
-  initialPrefs: Object.keys(prefConfig).reduce((obj, key) => {
-    obj[key] = prefConfig[key].value;
-    return obj;
-  }, {}), // This creates an object of the form {prefName: value}
+  validation: [
+    "showTopSites",
+    "showSearch",
+    // This means if either of these are set to their default values,
+    // prerendering can be used.
+    {oneOf: ["feeds.section.topstories", "feeds.section.highlights"]}
```
Mardak: Oh, I forgot to mention that this list doesn't include `topSitesCount` and `migrationExpired` where it did before. I would assume that was intentional.

PR #3434, opened 2017-09-12
fix(topsites): Fixes #3428 When pinning a site the corner icon disappears

### LINKED ISSUE #3428: When pinning a top site, the corner icon goes away
STR:
1. visit nytimes.com
2. wait for it to show up in top sites (notice it has a corner icon + screenshot)
3. pin the nytimes top site
4. restart nightly

Expected results:
The nytimes site still has a corner icon + screenshot

Actual results:
The nytimes site only has a screenshot - the corner icon is gone

sarracini:
This generalizes the ```_getPinnedWithData``` function so that we can augment with the data that comes from either some links that we give, or from the store if we don't give links. It wasn't enough to just change the Object.assign to smoosh onto the pinnedLink, because when we restart firefox and it gets it from the store, the favicons don't exist yet - we have to get them out of places. So to fix that we need to also augment the pinnedLinks on the refresh with the links that come out of places.


as-pine-proxy:
[Treeherder: f61fa1fca6cc9cf1ba68fac71927b68e3bd408c9](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f61fa1fca6cc9cf1ba68fac71927b68e3bd408c9)

PR #3435, opened 2017-09-12
fix (systemaddon) : #3149 icons for default sites

### LINKED ISSUE #3149: Update Tippy Top with scraped icons
When tippy top has allowed icons, we'll want to use those

rlr:
ICONS 🔥 !!!! Fixes #3149


rlr:
forgot the screenshot

screen shot 2017-09-12 at 4 00 20 pm

system-addon/data/content/tippytop/top_sites.json
```diff
@@ -46,12 +26,7 @@
...
   {
     "title": "ebay",
-    "urls": ["https://www.ebay.com", "https://www.ebay.co.uk/"],
-    "image_url": "ebay@2x.png"
-  },
-  {
-    "title": "ebay",
-    "url": "https://ebay.de",
+    "urls": ["https://www.ebay.com", "https://www.ebay.co.uk/", "https://ebay.de"],
     "image_url": "ebay-de@2x.png"
```
k88hudson: Nit, but should the image not be `"ebay@2x.png"` here instead of `"ebay-de@2x.png"`?

k88hudson: Note that you'll have to rename the image itself too

rlr: yes. i blame @bryanbell !! lol

system-addon/data/content/tippytop/top_sites.json
```diff
@@ -11,33 +11,13 @@
...
-  {
-    "title": "amazon",
-    "url": "https://www.amazon.de/",
-    "image_url": "amazon-de@2x.png"
-  },
-  {
-    "title": "amazon",
-    "url": "https://www.amazon.fr/",
-    "image_url": "amazon-fr@2x.png"
+    "urls": ["https://www.amazon.ca/", "https://www.amazon.co.uk/", "https://www.amazon.com/", "https://www.amazon.de/", "https://www.amazon.fr/"],
```
pdehaan: Sorry, I haven't been watching this repo as much recently, but why are we duplicating some tippy-top functionality here? It's already out of sync w/ tippy top here, re: smile.amazon.com.

Wouldn't it be easier to just add another flag to tippy-top JSON for `"default_site": true` or something and then do a quick `.filter()` if we just want the subset?

system-addon/data/content/tippytop/top_sites.json
```diff
@@ -46,28 +26,23 @@
   },
   {
     "title": "ebay",
-    "urls": ["https://www.ebay.com", "https://www.ebay.co.uk/"],
+    "urls": ["https://www.ebay.com", "https://www.ebay.co.uk/", "https://ebay.de"],
```
pdehaan: Same here. Should we include "https://ebay.ca/"?

https://github.com/mozilla/tippy-top-sites/blob/df4de4b322d3cf0742b6b1527a15bf9f291411d5/top_sites.json#L389

as-pine-proxy:
[Treeherder: a5d355937075aaab9261f03851bc5299cb08962f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a5d355937075aaab9261f03851bc5299cb08962f)

system-addon/data/content/tippytop/top_sites.json
rlr: ebay isn't a default site for Canada, so that's why it wasn't included.

system-addon/data/content/tippytop/top_sites.json
rlr: yeah, we sort of branched off from tippy top master to include only default sites. But then we changed to screenshots and just brought everything over here for that because it was easier. I think we can go back to tippy top after things stabilize. Although we will never ship all the icons so I'm not sure if it even makes sense because the new icons are way different than the old composite ones.

PR #3452, opened 2017-09-13
fix (systemaddon): #2007 autofocus search box for about:home

rlr:
@Mardak f? on this approach?

I kind of cheated and listened for telemetry actions. I still need to put it behind a pref and add tests.

Hope your house isn't on 🔥 😱


system-addon/lib/NewTabInit.jsm
```diff
@@ -37,6 +37,11 @@ this.NewTabInit = class NewTabInit {
           this._queue.clear();
         }
         break;
+      case at.SAVE_SESSION_PERF_DATA:
+        if (action.data.topsites_first_painted_ts && action._target.url === "about:home") {
```
rlr: I wanted to use `visibility_event_rcvd_ts` but that one isn't coming through for me for the first `about:home` on browser startup.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -22,6 +22,13 @@ class Search extends React.Component {
   onClick(event) {
     window.gContentSearchController.search(event);
   }
+  componentDidMount() {
+    // Focus the search box if we are on about:home
+    if (!IS_NEWTAB) {
+      const searchBox = document.getElementById("newtab-search-text");
+      if (searchBox) { searchBox.focus(); }
```
Mardak: The tricky part here is if we mount multiple times for unknown reasons.. But this might be fine for now.

rlr: I think I can request browser focus from here by dispatching an action?

system-addon/lib/NewTabInit.jsm
Mardak: So... the things to consider is that first_painted might happen before visibility if preloaded, but we only have preloaded `about:newtab`s and not `about:home` (at least for now). Also this assumes search gets rendered before topsites, which might be okay to assume for now?

I'm guessing you checked what possible actions were already being sent on early loads (running into #3455).

Otherwise we would need to detect if visible on mount to dispatch an action. And if not visible yet wait until then..

rlr: Ideally, we would have content announce when things are happening and different feeds can take appropriate action. But right now it's written a bit more imperatively.

system-addon/content-src/components/Search/Search.jsx
```diff
@@ -70,13 +77,13 @@ class Search extends React.Component {
...
-          title={this.props.intl.formatMessage({id: "search_button"})}>
-          <span className="sr-only"><FormattedMessage id="search_button" /></span>
-        </button>
+      <button
+        id="searchSubmit"
+        className="search-button"
+        onClick={this.onClick}
+        title={this.props.intl.formatMessage({id: "search_button"})}>
+        <span className="sr-only"><FormattedMessage id="search_button" /></span>
+      </button>
```
Mardak: Maybe leave out this change in case we just want to revert this commit down the line.

system-addon/content-src/components/Search/Search.jsx
Mardak: Ah actually that might work where we assume about:home is not preloaded. Also, any reason not to do this from `onInputMount`?

dmose: about:home is currently never preloaded, though it's likely to be pre-rendered.

rlr:
@Mardak ok, ready for r? for realz. i broke out the indentation fix into a separate commit

as-pine-proxy:
[Treeherder: 5e28fd590bb922a6e468c1511f4c8d466bd0885c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5e28fd590bb922a6e468c1511f4c8d466bd0885c)

system-addon/lib/NewTabInit.jsm
```diff
@@ -37,6 +37,11 @@ this.NewTabInit = class NewTabInit {
           this._queue.clear();
         }
         break;
+      case at.SEARCH_BOX_FOCUSED:
+        if (action._target.url === "about:home" && this.store.getState().Prefs.values["aboutHome.autoFocus"]) {
+          action._target.browser.focus();
```
Mardak: Oh duhh. I suppose we don't actually need to wait for the page to want focus we if know the page will want focus anyway. I.e., it's about:home. We didn't need this separate action, and instead we could just focus the browser immediately on NEW_TAB_INIT or some earlier action. Otherwise we have https://bugzilla.mozilla.org/show_bug.cgi?id=1400585

rlr: oooh. right.

Mardak: I just did some initial testing, and NEW_TAB_INIT is still not early enough. Here are the frames that the location bar gets focus then loses focus (if we call `focus()` from NEW_TAB_INIT) then when the search box appears focused.
23 -> 31 -> 33 = 8 frames of wrong focus, 2 frames of no focus
28 -> 35 -> 39 = 7 frames of wrong focus, 4 frames of no focus
7 -> 14 -> 17 = 7 frames of wrong focus, 3 frames of no focus

Mardak: Although it is faster than waiting for SEARCH_BOX_FOCUSED:
2 -> 12 -> 21 = 10 frames of wrong focus, 9 frames of no focus
19 -> 30 -> 37 = 11 frames of wrong focus, 7 frames of no focus
4 -> 20 -> 26 = 16 frames of wrong focus, 6 frames of no focus

Seems like the fix needs to be on the Firefox side to have the focus not go to the location bar in the first place.

marco-c:
> I'm sure someone will file a bug about focus moving from location bar to content search box now…

I was thinking about this. The location bar is "awesome"r than the content search box, as you also get suggestions from history.

PR #3471, opened 2017-09-14
fix(topstories): Fix story rotation in personalization experiment

csadilek:
Previous solution didn't take into account that:

- after re-fetching the feed, a top item that was previously
rotated could be top scored again and therefore needs to be
rotated again to not appear on top.

- a single rotation might not be good enough as the subsequent
item could have also been on top for too long.


as-pine-proxy:
[Treeherder: 685409c22846983f4361beb65be9498ae17a9cdd](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=685409c22846983f4361beb65be9498ae17a9cdd)

PR #3474, opened 2017-09-14
Persist Topstory bookmark status

### LINKED ISSUE #3131: Pocket story bookmark status is not persisted in context menu

piatra:
Closes #3131

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -261,9 +261,7 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           if (item.url === action.data.url) {
             const {bookmarkGuid, bookmarkTitle, dateAdded} = action.data;
             Object.assign(item, {bookmarkGuid, bookmarkTitle, bookmarkDateCreated: dateAdded});
-            if (!item.type || item.type === "history") {
-              item.type = "bookmark";
-            }
+            item.type = "bookmark";
           }
           return item;
         })
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -260,8 +260,12 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           // find the item within the rows that is attempted to be bookmarked
           if (item.url === action.data.url) {
             const {bookmarkGuid, bookmarkTitle, dateAdded} = action.data;
-            Object.assign(item, {bookmarkGuid, bookmarkTitle, bookmarkDateCreated: dateAdded});
-            item.type = "bookmark";
+            return Object.assign({}, item, {
+              bookmarkGuid,
+              bookmarkTitle,
+              bookmarkDateCreated: dateAdded,
+              type: "bookmark"
+            });
           }
           return item;
         })
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -154,10 +154,6 @@ class Section extends React.PureComponent {
       "tabIndex": 0
     };
 
-    if (id === "topstories") {
-      console.log(rows);
-    }
-
     const sectionInfoTitle = intl.formatMessage({id: "section_info_option"});
 
     const realRows = rows.slice(0, maxCards);
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -154,6 +154,10 @@ class Section extends React.PureComponent {
       "tabIndex": 0
     };
 
+    if (id === "topstories") {
+      console.log(rows);
+    }
+
     const sectionInfoTitle = intl.formatMessage({id: "section_info_option"});
 
     const realRows = rows.slice(0, maxCards);
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -101,7 +101,7 @@ this.HighlightsFeed = class HighlightsFeed {
 
     // Request more than the expected length to allow for items being removed by
     // deduping against Top Sites or multiple history from the same domain, etc.
-    const manyPages = await NewTabUtils.activityStreamLinks.getHighlights({
+    const manyPages = await this.linksCache.request({
       numItems: MANY_EXTRA_LENGTH,
       bookmarkSecondsAgo: await this._getBookmarksThreshold()
     });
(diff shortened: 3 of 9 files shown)


system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -238,6 +240,20 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.newTabsSinceSpoc++;
   }
 
+  updateBookmarkMetadata(url) {
+    const story = this.stories.find(s => s.url === url);
```
csadilek: it's unlikely, but this.stories could be undefined here if PLACES_BOOKMARK_ADDED fires before we fetch the feed the first time. so, we should check for that.

piatra:
f? @csadilek What do you think about this method of updating Topstories? The issue I ran into was that stories are initialized before highlights, checking for highlights bookmarks when fetching the stories returns `[]`.
There is still an issue where on a fresh browser start deduping does not happen and I need to add tests.

csadilek:
@piatra The deduping of top stories against highlights looks OK to me.

However, I don't think the ```updateBookmarkMetadata``` will work reliably. As the feeds are independent of the UI, you could have a tab rendered and kept in the background for hours. Then the data displayed will have diverged from the feed state which updates frequently. If you then select this 'old' tab and bookmark a story it might no longer be part of this.stories inside the feed.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -253,6 +266,28 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+  /**
+   * Update bookmark title, description and image with metadata from Pocket.
+   * @param url Used to identify if the bookmark was a topstory.
+   */
+  updateBookmarkMetadata(url) {
+    if (!this.stories) {
+      return;
+    }
+
+    const story = this.stories.find(s => s.url === url);
```
k88hudson: Why not just include the data you need here in the event from content? Isn't the data the most reliable / guaranteed to be there at that point?

piatra:
@csadilek I don't think we have to guard against that usecase. The story is:

> Design says reflecting "bookmarked" for pocket stories makes sense. And the behavior is that top stories show "Bookmarked" immediately without moving. On refresh, that card no longer shows up in Top Stories and appears in Highlights (with image and description and Bookmarked).

system-addon/lib/TopStoriesFeed.jsm
piatra: @k88hudson the bookmark moves into highlights after refresh, not when you bookmark. So there still needs to be some exchange from Highlights to TopStories to get the bookmarks.

piatra: I think the tricky case is on fresh browser start. When the App has no state.

csadilek:
@piatra ah ok, true. I guess I am missing what makes the top story cards show "Bookmarked" immediately?

piatra:
@k88hudson @csadilek I misunderstood the comment and it seems the card type also has to change to `bookmarked`. The issue with this is having an older tab open where the story is no longer present.

csadilek:
@piatra @k88hudson How about observing the ```BOOKMARK_ADDED``` event in Section(s) and updating all sites to set the type to "Bookmarked" if the URL is equal to ```action.data.url```? 

It seems that would fix the "problem" as then we wouldn't need to do anything in the feed.

csadilek:
New deduping mechanism is implemeted as part of this PR now: https://github.com/mozilla/activity-stream/pull/3538

system-addon/common/Reducers.jsm
```diff
@@ -260,10 +260,12 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           // find the item within the rows that is attempted to be bookmarked
           if (item.url === action.data.url) {
             const {bookmarkGuid, bookmarkTitle, dateAdded} = action.data;
-            Object.assign(item, {bookmarkGuid, bookmarkTitle, bookmarkDateCreated: dateAdded});
-            if (!item.type || item.type === "history") {
-              item.type = "bookmark";
```
piatra: Changing this didn't cause any tests to fail. Any particular scenario we were guarding against?

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -180,7 +180,7 @@ class Section extends React.PureComponent {
         </div>
         {!shouldShowEmptyState && (<ul className="section-list" style={{padding: 0}}>
           {realRows.map((link, index) => link &&
-            <Card key={index} index={index} dispatch={dispatch} link={link} contextMenuOptions={contextMenuOptions} eventSource={eventSource} />)}
```
piatra: React PureComponent prevent component update from happening.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -21,11 +21,16 @@ XPCOMUtils.defineLazyModuleGetter(this, "NewTabUtils",
...
 XPCOMUtils.defineLazyModuleGetter(this, "Screenshots",
   "resource://activity-stream/lib/Screenshots.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge",
+  "resource://gre/modules/ProfileAge.jsm");
 
 const HIGHLIGHTS_MAX_LENGTH = 9;
 const HIGHLIGHTS_UPDATE_TIME = 15 * 60 * 1000; // 15 minutes
 const MANY_EXTRA_LENGTH = HIGHLIGHTS_MAX_LENGTH * 5 + TOP_SITES_SHOWMORE_LENGTH;
 const SECTION_ID = "highlights";
+const BOOKMARKS_THRESHOLD = 3000; // 3 seconds to milliseconds.
```
piatra: This might want to be increased I'll also test on the reference hardware.

PR #3491, opened 2017-09-15
fix(highlights): Reduce card renders by only fetching new images.

### LINKED ISSUE #3481: Highlights should not re-check images from the cache
Especially since we're refreshing on every new tab load if we don't have enough highlights, we shouldn't be refreshing/re-checking screenshots cache every time. Let's reuse images once they've been added once

AdamHillier:
Closes #3481.


as-pine-proxy:
[Treeherder: e4eeca0401ee826175b4336ae1fa065c335499b5](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=e4eeca0401ee826175b4336ae1fa065c335499b5)

PR #3500, opened 2017-09-15
fix(pocket): Change 'ff' prefix to 'fx'

### LINKED ISSUE #3476: Pocket links use "ff" instead of "fx"
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1399995

> The abbreviation of Firefox is fx, not ff [1]. Since this is part of an official product, the preferred abbreviation should be used instead, like src=fx_new_tab.

csadilek:
Closes #3476 .

@Mardak Pocket endpoint has been updated as well so that the topic links returned use src=fx_new_tab (instead of ff_new_tab). Also links go through the CDN now e.g. https://getpocket.cdn.mozilla.net/explore/must-reads?src=fx_new_tab

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -52,12 +52,11 @@ const PREFS_CONFIG = new Map([
       api_key_pref: "extensions.pocket.oAuthConsumerKey",
       // Use the opposite value as what default value the feed would have used
       hidden: !PREFS_CONFIG.get("feeds.section.topstories").getValue(args),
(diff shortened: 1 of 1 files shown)


csadilek:
The learn_more_endpoint is actually no longer being used. So, I removed that and just updated the read_more_endpoint.

as-pine-proxy:
[Treeherder: d66fab3dab91d6c350dec324c73fb4dac4b8d43a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=d66fab3dab91d6c350dec324c73fb4dac4b8d43a)

PR #3508, opened 2017-09-16
feat(topsites): Migrate browser.newtabpage.enabled by disabling all sections

### LINKED ISSUE #3020: Tiles pref parity: browser.newtabpage.enabled
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1383599

This pref was used on Tiles to have a "blank" page (gear still appeared). Issue #2683 was decided against having UI for having a "show blank" option.

Potentially we could honor the existing pref to just show nothing (except the gear?) in activity stream even if there's no UI to toggle blank in the preference menu. Perhaps if the user does interact with the preference menu, we just set enabled to `true`, so this allows users to "transition" to activity stream.

Mardak:
Fix #3020. r?@k88hudson I see that turning off various sections have different "types" of prefs. In particular, Snippets seems to rely on the feed not being on for the UI to not show? Just making sure this will be the right pref, although I suppose if we do end up with a `.showSnippets` we would just update this.

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -141,6 +141,22 @@ function onBrowserReady() {
   }
 
   // Do a one time migration of Tiles about:newtab prefs that have been modified
+  migratePref("browser.newtabpage.enabled", enabled => {
+    // The only modified value should be false, but check to be safe
+    if (!enabled) {
+      // Pretend the user turned off all sections for now until there's a sticky
+      // pref (maybe "browser.newtabpage.enabled") that UI can control #3020
+      for (const pref of [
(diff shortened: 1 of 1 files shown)


k88hudson:
Is this really want we want? I would assume that we want people to see the new design we have, and if they want to turn it off again, they can just do that in the same place

Mardak:
Closing. @aaronrbenson says
> no, this is a new enough thing that they should get to see the _new_ new tab

and even for the special case of turning off for students, showing the new experience for not-special-case probably outweights the special case, and the special case would need to either update their deploy process to flip the new additional prefs or manually toggle things

Issue #3490, opened 2017-09-15
Add Edit to context menu for Top Sites


aaronrbenson:
Edit should be grouped with Pin (see https://github.com/mozilla/activity-stream/issues/2563). Clicking Edit should take you straight to the edit form for that top site and behave as it would if you entered Edit mode and clicked the edit icon. Clicking Save should take you straight back to New Tab.

Mardak:
If for 57, we'll just reuse `edit_topsites_button_text=Edit` But if it's after 57, we should add a new string.

aaronrbenson:
This would be great for 57 if we can reuse the string.

Issue #3447, opened 2017-09-13
Add caching to TopSitesFeed and HighlightsFeed


Mardak:
There's been some desire to refresh/update the results for UI but maybe not actually need truly fresh (and potentially expensive) Places/NewTabUtils data. This could be to adjust from deduping, pinning, or other user actions.

E.g.,
#2788 purposefully avoided refreshing from DB and Reduced instead
#3375 editing maybe should cause things to dedupe right away
#3410 could cause multiple early DB requests from INIT, TOP_SITES_UPDATED, NEW_TAB_LOAD

For both Feeds, we should be able to just save the result of `NewTabUtils.activityStreamLinks.getTopSites` or `NewTabUtils.activityStreamLinks.getHighlights` and use that data if some last updated time has not been expired. If we do need to force a refresh, we would just clear/reset that last updated time. Note the updated time should be of when we got the data from places not from when we last computed the results for UI.

For `HighlightsFeed`, it already has an `imageCache` from #3427, but that gets cleared on every fetch and acts more like a prefetching for the next time `fetchHighlights` is called. With a results cache, we could add in the image data. Also, there's a `this.highlights` that caches the last returned result for UI, but it's mostly used to determine if we've reached the max number of highlights and for testing.

PR #3515, opened 2017-09-18
fix(topsites): Use title instead of label to reuse existing titles of pinned links

### LINKED ISSUE #3514: Pinned sites don't use original titles
Not sure if it was intended to drop those or not. Old pinned links from tiles used `title` whereas adding/editing a top site uses `label`.

Mardak:
Fix #3514. r?@rlr Switches from `label` to `title`. Somewhat amusingly, the input box field that the user sees asks for "Enter a title". One behavior change is that when a link gets pinned, it shows its title as opposed to the hostname, but the user can edit it. Maybe with #3490.

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -135,7 +135,8 @@ class TopSite extends React.PureComponent {
     const {props} = this;
     const {link} = props;
     const isContextMenuOpen = this.state.showContextMenu && this.state.activeTile === props.index;
-    const title = link.label || link.hostname;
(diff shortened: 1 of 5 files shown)


rlr:
Wait, so if I pin, will the text change on the user? That would be annoying /me tries

Mardak:
Well, I have a fix as part of #3447 but I could do that here. Basically just copy over the desired properties instead of all properties when pinning.

rlr:
Maybe when we pin, we should send the `hostname` so that it get stores in the pref as the default title?

Mardak:
Looks like #3514 got WONTFIXed.

Issue #3513, opened 2017-09-17
Pinned top sites can expose tippyTop


Mardak:
I noticed this when tracking down #3512. Looks like we're modifying the pinned links object that then get saved when creating a new pinned link:

`browser.newtabpage.pinned;[{"url":"https://www.youtube.com/","tippyTopIcon":"resource://activity-stream/data/content/tippytop/images/youtube-com@2x.png"}]`

Not sure if this will cause problems if we change tippyTop data down the line.

Note: This doesn't happen when pinning just the single default top site. Looks like the default is pinned first then page refreshed then a non-default pinned? Not quite sure of the exact steps.

Mardak:
Fixed by #3530 c7ad97c

PR #3516, opened 2017-09-18
fix(topsites): Have pin actions trigger refresh to avoid bugs from _getPinnedWithData diverging

### LINKED ISSUE #3347: Editing a Top Site URL wrongly creates a new Top Site

Mardak:
Fix #3347. r? @rlr This should land after #3515 for using title instead of label. But this does not actually solve #3513 as it looks like we edit a live object and that gets persisted when another link is pinned.

This fixes #3347 by avoiding the not quite logic of `_getPinnedWithData`. I plan on getting rid of the rest of it with #3447.

In terms of performance, the speed of the query has been vastly improved by using a frecency index. But #3447 will avoid querying in the first place to keep `refresh` instead of `getPinned` fast.


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -154,39 +154,66 @@ this.TopSitesFeed = class TopSitesFeed {
...
+  /**
+   * Inform others that top sites data has been updated due to pinned changes.
+   */
   _broadcastPinnedSitesUpdated() {
-    this.store.dispatch(ac.BroadcastToContent({
-      type: at.PINNED_SITES_UPDATED,
-      data: this._getPinnedWithData()
-    }));
+    // Refresh to update pinned sites with screenshots, trigger deduping, etc.
+    this.refresh();
```
rlr: The first pinning implementation called `refresh()` on every pinning change. Everything seems to come back around, heh. When we get drag and drop that might cause a lot of calls in a short period of time. But I'm assuming refresh will now used cached places db results?

Mardak:
Updated to save `label` instead of `title`.

system-addon/lib/TopSitesFeed.jsm
Mardak: Hmm.. For drag-n-drop I would think the page is locally updating state during the drag until a drop to actually update the pinned states for things, and that should only trigger one `refresh`?

as-pine-proxy:
[Treeherder: 1468b3a070f8254a8a398436679338e8c1bacbd8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1468b3a070f8254a8a398436679338e8c1bacbd8)

PR #3519, opened 2017-09-18
fix (stories): #3423 dedupe highlights from top stories

### LINKED ISSUE #3423: Dismissing a Pocket recommendation wrongly dismisses the Highlights card of the same page
https://bugzilla.mozilla.org/show_bug.cgi?id=1399052

[Affected versions]:
- Firefox 57.0a1 Build ID 20170912013600

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- "browser.newtabpage.activity-stream.enabled" and "browser.newtabpage.activity-stream.feeds.section.topstories" are set to true in about:config on a clean new profile.

[Steps to reproduce]:
2. Navigate to the three Top Stories from Pocket.
3. Open a new tab and dismiss one of the Top Stories.
4. Observe the Top Stories and Highlights sections. 

[Expected results]:
- The Top Story is dismissed from its section, another one takes it place and there is no change in the Highlights section.

[Actual results]:
- The Top Story is dismissed from its section, but the Highlight card of the same page is also dismissed.

[Notes]:
- Dismissing a Highlight will not dismiss a Top Story of the same page.
dismissing a top story will dismiss a highlight of the same page

rlr:
Fixes #3423

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -7,7 +7,6 @@ const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/NewTabUtils.jsm");
-Cu.import("resource://gre/modules/Console.jsm");
 Cu.importGlobalProperties(["fetch"]);
 
 const {actionTypes: at, actionCreators: ac} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
@@ -33,11 +32,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.spocsPerNewTabs = 0;
     this.newTabsSinceSpoc = 0;
     this.contentUpdateQueue = [];
-    this.dedupe = new Dedupe(this._dedupeKey);
-  }
-
-  _dedupeKey(site) {
-    return site && site.url;
+    this.dedupe = new Dedupe(site => site && site.url);
   }
 
   init() {
@@ -80,7 +75,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
       const body = await response.json();
       this.updateSettings(body.settings);
       this.stories = this.rotate(this.transform(body.recommendations));
-      this.dedupeFromHighlights();
+      this.dedupeFromHighlights(false);
       this.spocs = this.show_spocs && this.transform(body.spocs).filter(s => s.score >= s.min_score);
 
       this.dispatchUpdateEvent(this.storiesLastUpdated, {rows: this.stories});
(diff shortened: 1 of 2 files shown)


system-addon/lib/HighlightsFeed.jsm
```diff
@@ -118,6 +118,8 @@ this.HighlightsFeed = class HighlightsFeed {
 
     SectionsManager.updateSection(SECTION_ID, {rows: this.highlights}, this.highlightsLastUpdated === 0 || broadcast);
     this.highlightsLastUpdated = Date.now();
+    const updatedAction = {type: at.HIGHLIGHTS_UPDATED};
+    this.store.dispatch(ac.SendToMain(updatedAction));
```
piatra: @rlr I'm not sure about this approach. The SectionManager already has a system in place for notifying when a section is updated. Doing this would bypass it. I would get a second opinion on this.

rlr:
@k88hudson and/or @csadilek: mind taking a look? Thanks!

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -7,13 +7,15 @@ const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/NewTabUtils.jsm");
+Cu.import("resource://gre/modules/Console.jsm");
```
csadilek: Probably a left-over and can be removed?

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -31,6 +33,11 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.spocsPerNewTabs = 0;
     this.newTabsSinceSpoc = 0;
     this.contentUpdateQueue = [];
+    this.dedupe = new Dedupe(this._dedupeKey);
```
csadilek: Can we pass site => site && site.url here instance of creating a pseudo-private member function?

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -272,6 +289,11 @@ this.TopStoriesFeed = class TopStoriesFeed {
       case at.NEW_TAB_REHYDRATED:
         this.maybeAddSpoc(action.meta.fromTarget);
         break;
+      case at.SECTION_UPDATE:
+        if (action.data.id === "highlights") {
+          this.dedupeFromHighlights();
```
csadilek: Updating ```this.stories``` in response to SECTION_UPDATE doesn't guarantee the UI is updated and the modified ```this.stories``` are actually used. We would have to fire a SECTION_UPDATE event for the topstories section to make sure, see https://github.com/mozilla/activity-stream/pull/3474/files#diff-13ca5ab376313a4fb771454ba25daa01R72.

We'd only need to fire this event if ```dedupefromHighlights``` is invoked as a response to SECTION_UPDATE. Not needed when invoked from fetchStories (there the even is fired anyway).



In general, I am not qui

system-addon/lib/TopStoriesFeed.jsm
rlr: oopsies!

system-addon/lib/TopStoriesFeed.jsm
rlr: yep. makes sense!

csadilek: Ah, and in the former case we probably should only fire an SECTION_UPDATE if the stories actually changed after deduping from highlights.

rlr:
@csadilek does that work ^. I called `SectionManager.updateSection()` directly

csadilek:
@rlr ```SectionManager.updateSection``` would work. You'd also have to decide if you want that change broadcasted to all open new tabs. 

Another disadvantage I see by solving it in the feed and not in SectionsManager is that users might see stories change as they are deduped against highlights.

rlr:
I guess you would broadcast it only if the highlights change is being
updated?

I'll think a little bit about how to do it in the SectionManager.

rlr:
closing in favor of pull #3538

PR #3530, opened 2017-09-19
feat(topsites): Add caching of links with image migration for top sites and highlights

### LINKED ISSUE #3447: Add caching to TopSitesFeed and HighlightsFeed

### LINKED ISSUE #3513: Pinned top sites can expose tippyTop

Mardak:
Fix #3447 and fix #3513 . r?@k88hudson Various changes:
- self contained link cache module (no imports!)
  - topsites caches places frecent query and pinned links
  - highlights caches places highlights query
- link objects should be directly modified to save to cache
- custom image caching is removed and instead images are updated on the link
  - screenshot fetching promise is stashed on link to avoid repeat requests
  - also make sure there's only one "update" notification sent out on getting screenshot
- cached links are copies of source data (in particular avoid pinned links #3513)
- use `topsiteFrecency` (`>=` check) to filter out frecency instead of locally (was `>` check)
- get rid of `getPinnedWithData` and have pinned links inherit frecent data
- changes HighlightsFeed "api" to not expose result cache `.highlights` that was used for testing

One thing in particular that did **not** change is the update timing, `onAction` handling of the feeds. Originally I took out the timing checks, but after some more thought, we might still want to be smarter than just "refreshing" on each `NEW_TAB_LOAD` even if it's really fast because we're already sending over that data on rehydrating. I briefly had it set to just update in the background on `SYSTEM_TICK` but figured this change was already big enough.

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -84,7 +84,7 @@ function insertPinned(links, pinned) {
   // Then insert them in their specified location
   pinned.forEach((val, index) => {
     if (!val) { return; }
-    let link = Object.assign(val, {isPinned: true, pinIndex: index});
+    let link = Object.assign({}, val, {isPinned: true, pinIndex: index});
     if (index > newLinks.length) {
       newLinks[index] = link;
     } else {
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -84,7 +84,7 @@ function insertPinned(links, pinned) {
   // Then insert them in their specified location
   pinned.forEach((val, index) => {
     if (!val) { return; }
-    let link = Object.assign({}, val, {isPinned: true, pinIndex: index});
+    let link = Object.assign(val, {isPinned: true, pinIndex: index});
     if (index > newLinks.length) {
       newLinks[index] = link;
     } else {
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -15,6 +15,8 @@ const {Dedupe} = Cu.import("resource://activity-stream/common/Dedupe.jsm", {});
 
 XPCOMUtils.defineLazyModuleGetter(this, "filterAdult",
   "resource://activity-stream/lib/FilterAdult.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "LinksCache",
+  "resource://activity-stream/lib/LinksCache.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "NewTabUtils",
   "resource://gre/modules/NewTabUtils.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "Screenshots",
@@ -28,8 +30,18 @@ const SECTION_ID = "highlights";
 this.HighlightsFeed = class HighlightsFeed {
   constructor() {
     this.highlightsLastUpdated = 0;
-    this.highlights = [];
+    this.highlightsLength = 0;
     this.dedupe = new Dedupe(this._dedupeKey);
+    this.linksCache = new LinksCache(NewTabUtils.activityStreamLinks,
+      "getHighlights", (oldLink, newLink) => {
+        // Migrate any pending images or images to the new link
+        for (const property of ["fetchingImage", "image"]) {
+          const oldValue = oldLink[property];
+          if (oldValue) {
+            newLink[property] = oldValue;
+          }
+        }
+      });
   }
 
   _dedupeKey(site) {
@@ -50,6 +62,11 @@ this.HighlightsFeed = class HighlightsFeed {
   }
 
   async fetchHighlights(broadcast = false) {
+    // We broadcast when we want to force an update, so get fresh links
+    if (broadcast) {
+      this.linksCache.expire();
+    }
+
     // We need TopSites to have been initialised for deduping
     if (!this.store.getState().TopSites.initialized) {
       await new Promise(resolve => {
@@ -64,7 +81,7 @@ this.HighlightsFeed = class HighlightsFeed {
 
     // Request more than the expected length to allow for items being removed by
     // deduping against Top Sites or multiple history from the same domain, etc.
-    const manyPages = await NewTabUtils.activityStreamLinks.getHighlights({numItems: MANY_EXTRA_LENGTH});
+    const manyPages = await this.linksCache.request({numItems: MANY_EXTRA_LENGTH});
 
     // Remove adult highlights if we need to
     const checkedAdult = this.store.getState().Prefs.values.filterAdult ?
@@ -73,16 +90,8 @@ this.HighlightsFeed = class HighlightsFeed {
     // Remove any Highlights that are in Top Sites already
     const [, deduped] = this.dedupe.group(this.store.getState().TopSites.rows, checkedAdult);
 
-    // Store existing images in case we need to reuse them
-    const currentImages = {};
-    for (const site of this.highlights) {
-      if (site && site.image) {
-        currentImages[site.url] = site.image;
-      }
-    }
-
     // Keep all "bookmark"s and at most one (most recent) "history" per host
-    this.highlights = [];
+    const highlights = [];
     const hosts = new Set();
     for (const page of deduped) {
       const hostname = shortURL(page);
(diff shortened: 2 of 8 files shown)


system-addon/common/Reducers.jsm
```diff
@@ -84,7 +84,7 @@ function insertPinned(links, pinned) {
   // Then insert them in their specified location
   pinned.forEach((val, index) => {
     if (!val) { return; }
-    let link = Object.assign({}, val, {isPinned: true, pinIndex: index});
+    let link = Object.assign(val, {isPinned: true, pinIndex: index});
```
k88hudson: What's your reasoning behind this change? Mutating stuff that comes from function arguments seems like a bad idea to me

system-addon/lib/LinksCache.jsm
```diff
@@ -0,0 +1,101 @@
...
+ * links to the new links with the same url.
+ */
+this.LinksCache = class LinksCache {
+
+  /**
+   * Create a links cache for a given object property.
+   *
+   * @param {object} linkObject Object containing the link property
+   * @param {string} linkProperty Name on object to access
+   * @param {function} migrator Optional callback receiving the old and new link
```
k88hudson: Can you describe what this does in a little more detail in this comment

system-addon/lib/LinksCache.jsm
```diff
@@ -0,0 +1,101 @@
...
+ * amount of time has passed. Allows for migrating data from previously cached
+ * links to the new links with the same url.
+ */
+this.LinksCache = class LinksCache {
+
+  /**
+   * Create a links cache for a given object property.
+   *
+   * @param {object} linkObject Object containing the link property
+   * @param {string} linkProperty Name on object to access
```
k88hudson: `Property on object` or `name of property on object` maybe?

system-addon/lib/LinksCache.jsm
```diff
@@ -0,0 +1,101 @@
...
+    this.cache = Promise.resolve([]);
+    this.lastOptions = {};
+  }
+
+  /**
+   * Force the next request to update the cache.
+   */
+  expire() {
+    // Pretend the we expired at some negative time (tests start at T=0)
+    this.lastUpdate = 0 - EXPIRATION_TIME - 1;
```
k88hudson: Personally I would change `request` to check `if (!this.lastUpdate || now > this.lastUpdate + EXPIRATION_TIME`) and have this method just delete `this.lastUpdate` to avoid the weird math there

k88hudson: This would also prevent having to explicitly call `expire()` in the constructor, since `lastUpdate` is only set by `request`.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -54,30 +62,45 @@ this.TopSitesFeed = class TopSitesFeed {
...
-    const defaultUrls = notBlockedDefaultSites.map(site => site.url);
-    let pinned = this._getPinnedWithData(frecent);
-    pinned = pinned.map(site => site && Object.assign({}, site, {
-      isDefault: defaultUrls.indexOf(site.url) !== -1,
-      hostname: shortURL(site)
-    }));
+    const frecent = (await this.frecentCache.request({
+      numItems,
+      topsiteFrecency: FRECENCY_THRESHOLD
+    })).map(link => Object.assign(link, {hostname: shortURL(link)}));
```
k88hudson: Again, I generally don't like the idea of assuming no side effects will occur if you mutate stuff that comes from other functions. It relies on the particular implementation of the cache to return copies, and if that were changed, we'd have to remember to check/update this code

system-addon/common/Reducers.jsm
Mardak: This is so that we keep the original pinned link object. Otherwise screenshots are saved to this copy instead of to the one in the cache.

system-addon/lib/TopSitesFeed.jsm
Mardak: How would you update the cache? We could expose a method on `LinksCache` to update an entry, but either we need to search for the entry on each update or keep around a mapping for lookup?

Mardak: I suppose it would be similar to migrating from old link to new link.

system-addon/lib/LinksCache.jsm
Mardak: Okay switching to `delete` although I just realized we actually want to call `expire` from `clear` to correctly forget the last update time.

system-addon/lib/LinksCache.jsm
Mardak: used "Name of property on object to access"

system-addon/lib/LinksCache.jsm
Mardak: Expanded this and for `shouldRefresh`

system-addon/common/Reducers.jsm
Mardak: Restored the `.assign({}`

system-addon/lib/TopSitesFeed.jsm
Mardak: Added a `{}` as the first agument to the `assign` here and for pinned too.

Mardak:
@k88hudson added an `updateCache` method on all cached links that then should be called when it wants to update the cache.

https://github.com/mozilla/activity-stream/pull/3530/commits/9b4d6f566f941b52d913761b382b10c41b76b9a8#diff-da3bc0f65a51b010768a775b2cbfdffeR94

Not sure if we need to be remove that method before `dispatch`. The added method happens to disappear on message passing, but we can be cleaner. Although the method needs to survive through `refresh`. Although I suppose then again we could move the icon fetching into `getLinksWithDefault` and get rid of `updateCache` by the time links are returned.

Mardak:
@k88hudson I overloaded the `updateCache` method to have it remove itself so that it doesn't get exposed outside of `getLinksWithDefault`:
https://github.com/mozilla/activity-stream/pull/3530/commits/a0b39be800808bf9c408b06269fb5796aeb82f9c#diff-da3bc0f65a51b010768a775b2cbfdffeR93

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -93,46 +102,51 @@ this.HighlightsFeed = class HighlightsFeed {
...
+   * image is available then fallback to fetching a screenshot.
    */
-  async fetchImage(url, imageUrl) {
-    const image = await Screenshots.getScreenshotForURL(imageUrl || url);
-    SectionsManager.updateSectionCard(SECTION_ID, url, {image}, true);
+  async fetchImage(page) {
+    // Request a screenshot if we don't already have one pending
+    const {preview_image_url: imageUrl, url} = page;
+    if (!page.fetchingImage) {
+      page.fetchingImage = new Promise(async resolve => {
```
k88hudson: Does  `.fetchingImage` need to be removed on completion? I see it sticking around in the content reducers

k88hudson:
Ok, so I definitely like this better, but adding/removing/calling methods and promises (i.e. `.updateCache` /`.fetchingImage`) is still breaking the redux model that makes state immutable outside of reducers, and that reducers should be pure functions...

... however, in the interest of getting this landed since it is really helpful, and because `updateCache`/`fetchingImage` don't actually manipulate any data that react depends on, I would be ok landing this with some comments and maybe marking them in some special way (maybe underscores? maybe nesting them under a `__HELPERS__` property or something?)

It's also a bit weird that `.fetchingImage` is sent down to content and ends up being an empty object since it can't be sent across the process boundary. If you do end up nesting stuff under `__HELPERS__` you could remove it in some middleware maybe.

system-addon/lib/HighlightsFeed.jsm
Mardak: Shared this adding and removing logic for `__fetchingScreenshot` as part of `Screenshots.maybeGetAndSetScreenshot` and updated `__updateCache` to delete properties.

Mardak:
Ended up renaming to `__updateCache` and `__fetchingScreenshot` with tests to check for `__` properties. The `__HELPERS__` sub object doesn't quite work at least not directly as implemented as `__updateCache` relies on being able to refer to the current object.

Mardak:
Rebased, squashed and merged c7ad97c0bd205e15effdb77683f7d8fc0ca5ea7f

Mardak:
>  still breaking the redux model that makes state immutable outside of reducers

After thinking about this a bit more, the intent of stashing the additional properties isn't so much to update content (that happens via the action) but more for updating the cache. Filed #3561

PR #3536, opened 2017-09-19
feat(style): Hide top sites and sections before we've initialized

k88hudson:
This hides top sites and sections until we've initialized, but leaves the search box visible from the start. Let's see if this looks better?

### DIFF
--- a/system-addon/content-src/components/Base/_Base.scss
+++ b/system-addon/content-src/components/Base/_Base.scss
@@ -46,3 +46,11 @@ main {
     vertical-align: middle;
   }
 }
+
+.body-wrapper {
+  opacity: 0;
+  transition: opacity 0.1s ease-in-out;
+  &.on {
+    opacity: 1;
+  }
+}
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/Base/_Base.scss
```diff
@@ -46,3 +46,11 @@ main {
     vertical-align: middle;
   }
 }
+
+.body-wrapper {
+  opacity: 0;
```
Mardak: Oh. ha ha, so the prerendered content can still trigger scrollbars without shifting ;)

Mardak:
Hrmm. At least for me on a clean profile, the very first launch has  some slight shifting but launching firefox a second time doesn't shift?
https://ed.agadak.net/as/activity-stream.fade-on-init.webm

Mardak:
What's the status here after chatting with @aaronrbenson? Sounds like making big sections appear is not great for perceived performance?

aaronrbenson:
@k88hudson was going to land it as a pref so we could evaluate further. It's difficult to judge without seeing it first hand. However, from what I saw in her demo to me, the effect was super nice on new windows.

as-pine-proxy:
[Treeherder: 0437c6bc846b8ef90983e89eb724e640527a2a51](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0437c6bc846b8ef90983e89eb724e640527a2a51)

PR #3538, opened 2017-09-19
feat (sections): #3423 implement deduping for sections

rlr:
@csadilek is this what you were thinking? f? Still need to write tests.

### DIFF
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -140,17 +140,18 @@ const SectionsManager = {
     }
   },
   dedupeRows(id, options) {
+    const newOptions = Object.assign({}, options);
     const dedupeFrom = this.sections.get(id).dedupeFrom;
     if (dedupeFrom && dedupeFrom.length > 0 && options.rows) {
       for (const sectionId of dedupeFrom) {
         const section = this.sections.get(sectionId);
         if (section && section.rows) {
           const [, newRows] = this.dedupe.group(section.rows, options.rows);
-          options.rows = newRows;
+          newOptions.rows = newRows;
         }
       }
     }
-    return options;
+    return newOptions;
   },
 
   /**
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -4,9 +4,10 @@
 "use strict";
 
 const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/EventEmitter.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 const {actionCreators: ac, actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
-Cu.import("resource://gre/modules/EventEmitter.jsm");
+const {Dedupe} = Cu.import("resource://activity-stream/common/Dedupe.jsm", {});
 
 /*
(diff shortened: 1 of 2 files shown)


csadilek:
@rlr Nice! Yes, I think that is a much better solution as it will work for any other feed/section brought in later. Feeds can simply declare what feeds to dedupe from. In most cases this will also prevent rendering cards and immediately updating them as this happens in updateSection now.

So, 👍 :)

rlr:
@csadilek ok, got a couple of tests added and passing. It was another test with a side effect that had me pulling my hair out. r?

system-addon/lib/SectionsManager.jsm
```diff
@@ -123,9 +127,30 @@ const SectionsManager = {
...
+    }
+  },
+  dedupeRows(id, options) {
+    const dedupeFrom = this.sections.get(id).dedupeFrom;
+    if (dedupeFrom && dedupeFrom.length > 0 && options.rows) {
+      for (const sectionId of dedupeFrom) {
+        const section = this.sections.get(sectionId);
+        if (section && section.rows) {
+          const [, newRows] = this.dedupe.group(section.rows, options.rows);
+          options.rows = newRows;
```
piatra: Could we avoid mutating the arguments?

piatra:
There is an issue I ran into when rebasing over this patch. I will try to describe the flow:
1. You bookmark a topstory
2. HighlightsFeed listens for bookmark events and calls fetchHighlights **but** does not broadcast the update.
3. SectionsManager will try to dedupe and looks at `Highlights.rows` but that is not actually the state reflected in the page.
Let me know if this makes sense. My suggestion would be to query the store instead.

rlr:
@piatra That makes sense. But even if fetchHighlights doesn't broadcast, it still calls `SectionManager.updateSection()`: https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/HighlightsFeed.jsm#L119

So that should keep the `SectionManager` data in sync with the main store. But yeah, there might be existing Activity Stream instances not getting updated but they won't get the update from the dedupe either. They just have a stale store but I think that's by design?

csadilek:
@rlr looks good to me! I will do some manual testing as well. @piatra What you describe basically means that existing tabs wouldn't get deduped/updated, right? I think that's fine as in https://github.com/mozilla/activity-stream/pull/3474 we are working on marking the card as "bookmarked" immediately and new tabs from then onwards will get deduped. That matches the requirement if I understand correctly?

piatra:
@csadilek yes that is the requirement. marked as "bookmarked" immediately.

This is kind of extreme but worth mentioning. If you bookmark 10 top stories the first one you bookmarked will surface again in Topstories.

csadilek:
@piatra yes, this makes sense. We only have 9 highlight cards, so after the 10th bookmark, the first one will be back in the feed at its default position (either sorting or scoring in personalization).

csadilek:
@rlr @piatra so this works well in my tests. Happy to approve. You can to decide when to merge it. Before or after beta ;).

as-pine-proxy:
[Treeherder: 2ccedbd9dcba0fdc8e0edde413354f8f55a06d04](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2ccedbd9dcba0fdc8e0edde413354f8f55a06d04)

PR #3542, opened 2017-09-19
fix(snippets): Don't block snippets on onboardingFinished

k88hudson:
Manual test:
- onboarding should show if snippets are enabled in the prefs panel, and onboarding has notifications to show
- onboarding should be visually hidden if snippets are turned off in the prefs panel
- onboarding should show up on the next new tab if snippets are turned back on in the prefs panel

- while onboarding is showing, call `gSnippetsMap.disableOnboarding()` in your console. Nothing should visually happen. on the next new tab, onboarding should be hidden and the default pref for onboarding.notifications.finished should be true.
- restart your browser. onboarding should be shown again


system-addon/content-src/lib/snippets.js
```diff
@@ -316,17 +327,19 @@ function addSnippetsSubscriber(store) {
...
+    if (state.Prefs.values["feeds.snippets"] &&
+      state.Snippets.initialized &&
+     !snippets.initialized &&
+     // Don't call init multiple times
+     !initializing
+    ) {
+      initializing = true;
+      await snippets.init({appData: state.Snippets});
+      initializing = false;
+    } else if (state.Prefs.values["feeds.snippets"] === false && snippets.initialized) {
```
Mardak: Just noting that we could be `initializing` and forget to `uninit`.

Mardak: Also, this has a behavior change from before where it checked `state.Snippets.initialized` and that could be false when activity stream is shutting down with `UNINIT`. We no longer `SnippetsProvider.uninit` when shutting down, so stuff could be differently broken, but other things are broken with an existing page when the pref is turned off. And soon we wouldn't shut down unless firefox itself is shutting down.

as-pine-proxy:
[Treeherder: 3c720fcacfcfdf0743ac74d33e5db0d8a875c2ee](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=3c720fcacfcfdf0743ac74d33e5db0d8a875c2ee)

PR #3543, opened 2017-09-19
fix(telemetry): Send an undesired event ping if addon fails to init

### LINKED ISSUE #3522: Add a ping to check the health of initialization
We should add a ping that looks roughly like

let initializedProperly = false;
let errorMessage;

try {
    initailizeActivityStream();
    initalizedProperly = true;
} catch(err) {
   errorMessage = `${err}`;
}

pingCetnre.sendPing({initailizedProperly, errorMessage});

This will help us monitor the health of whether activity stream is initializing properly and debug issues surrounding activity stream clients breaking due to internal errors and then never receiving any telemetry from them, which is happening already in our Shield studies.

ncloudioj:
This closes #3522 

Patch notes:

* The init failure ping was sent from `ActivityStream.init` other than`bootstrap.js` for better test-ability.
* Not sure if we can re-arrange the `ActivityStream.init` function so that it could init the TelemetryFeed as early as possible to be able to report other potential init errors. @k88hudson What do you think about this?

### DIFF
--- a/system-addon/test/unit/lib/Store.test.js
+++ b/system-addon/test/unit/lib/Store.test.js
@@ -163,6 +163,17 @@ describe("Store", () => {
       assert.calledOnce(store.dispatch);
       assert.calledWith(store.dispatch, action);
     });
+    it("should initialize the telemtry feed first", () => {
+      store._prefs.set("feeds.foo", true);
+      store._prefs.set("feeds.telemetry", true);
+      const telemetrySpy = sandbox.stub().returns({});
+      const fooSpy = sandbox.stub().returns({});
+      // Intentionally put the telemetry feed as the second item.
+      const feedFactories = new Map([["feeds.foo", fooSpy],
+                                     ["feeds.telemetry", telemetrySpy]]);
+      store.init(feedFactories);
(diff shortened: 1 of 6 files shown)


ncloudioj:
@fmarier Could you r? this as well, please?

system-addon/test/unit/lib/Store.test.js
```diff
@@ -163,6 +163,17 @@ describe("Store", () => {
       assert.calledOnce(store.dispatch);
       assert.calledWith(store.dispatch, action);
     });
+    it("should initialize the telemtry feed first", () => {
+      store._prefs.set("feeds.foo", true);
+      store._prefs.set("feeds.telemetry", true);
+      const telemetrySpy = sandbox.spy(() => ({}));
```
k88hudson: I think you want `sandbox.stub().returns({})`?

as-pine-proxy:
[Treeherder: 5a707f687a736b3f99603f47f49bc6936c795b45](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5a707f687a736b3f99603f47f49bc6936c795b45)

PR #3550, opened 2017-09-19
fix(styles): Use right aligned ltr for url input

### LINKED ISSUE #3549: The URL field in the Activity Streams edit UI needs to be LTR always
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1401399

> See the screenshot from an RTL build showing the edit UI for the new tab page.  The URL field shows the slash at the end of the URL at the left hand of the URL, because the text box is inheriting the direction of the page (LTR).  The text box's direction needs to be set to ltr explicitly to avoid this.  We do this everywhere in our UI where we display URLs.

image

Mardak:
Fix #3549. r?@rlr 

Before top, after bottom:
screen shot 2017-09-19 at 4 51 20 pm

title I put as "youtube..."

### DIFF
--- a/system-addon/content-src/components/TopSites/_TopSites.scss
+++ b/system-addon/content-src/components/TopSites/_TopSites.scss
@@ -323,6 +323,11 @@
       position: relative;
     }
 
+    .url input:not(:placeholder-shown):dir(rtl) {
+      direction: ltr;
+      text-align: right;
+    }
+
     .section-title {
       margin-bottom: 5px;
       margin-inline-start: 205px;


as-pine-proxy:
[Treeherder: dee5f242252601403ebb9fb1e6ed3d0a8e109e88](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=dee5f242252601403ebb9fb1e6ed3d0a8e109e88)

Mardak:
Just to log it, the second commit was to fix placeholder to work just in case a translation ended a rtl string with punctuation:
screen shot 2017-09-19 at 5 24 13 pm

PR #3551, opened 2017-09-20
fix(search): Only attempt to autofocus if there would be a search box

### LINKED ISSUE #3532: When Firefox is opened and about:home is opened with its search bar disabled, the address bar is not focused
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1400409

Probably could leave focus in the location bar if the search box is hidden. Might depend on #3505 if it doesn't focus the location bar in the first place though.

Mardak:
Fix #3532. r?@rlr

### DIFF
--- a/system-addon/lib/NewTabInit.jsm
+++ b/system-addon/lib/NewTabInit.jsm
@@ -38,8 +38,11 @@ this.NewTabInit = class NewTabInit {
         }
         break;
       case at.NEW_TAB_INIT:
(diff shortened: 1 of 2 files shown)


as-pine-proxy:
[Treeherder: 4839242839d398db8aea6d51ad96b6e73c944c90](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4839242839d398db8aea6d51ad96b6e73c944c90)

PR #3572, opened 2017-09-21
fix(sections): Section option pref change should re-initialize feeds

### LINKED ISSUE #3569: Feed option pref change should re-initialize feed
If a section's option pref is changed we should send an event that feeds can observe to re-initialize.

csadilek:
Closes #3569 .


system-addon/lib/SectionsManager.jsm
```diff
@@ -278,11 +278,16 @@ class SectionsFeed {
...
         SectionsManager.init(action.data);
         break;
-      case at.PREF_CHANGED:
-        if (action.data && action.data.name.match(/^feeds.section.(\S+).options$/i)) {
-          SectionsManager.addBuiltInSection(action.data.name.slice(0, -8), action.data.value);
+      case at.PREF_CHANGED: {
+        if (action.data) {
+          const matched = action.data.name.match(/^feeds.section.(\S+).options$/i);
+          if (matched) {
+            SectionsManager.addBuiltInSection(action.data.name.slice(0, -8), action.data.value);
```
Mardak: Seems like if we're touching this, might as well grab the feed pref name from the regex instead of slicing some number.
`/^(feeds\.section\.(.+))\.options$/i`

system-addon/lib/SectionsManager.jsm
```diff
@@ -278,11 +278,16 @@ class SectionsFeed {
...
+            this.store.dispatch({type: at.SECTION_OPTIONS_CHANGED, data: matched[1]});
```
Mardak: I'm not super familiar with this code, but is this the way to talk to sections? E.g., not a direct method or `emit` or something?

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -272,6 +272,15 @@ this.TopStoriesFeed = class TopStoriesFeed {
       case at.NEW_TAB_REHYDRATED:
         this.maybeAddSpoc(action.meta.fromTarget);
         break;
+      case at.SECTION_OPTIONS_CHANGED:
+        if (action.data === SECTION_ID) {
+          this.uninit();
+          this.storiesLastUpdated = 0;
+          this.topicsLastUpdated = 0;
+          this.affinityLastUpdated = 0;
```
Mardak: Seems like these initial values might just want to be in `init` instead of `constructor` ?

system-addon/lib/SectionsManager.jsm
csadilek: True, I'll take care of that.

system-addon/lib/TopStoriesFeed.jsm
csadilek: Yes, fixing.

system-addon/lib/SectionsManager.jsm
csadilek: Yes, I think so. ```emit``` it uses to relay events to itself.

csadilek:
@Mardak OK, updated with requested changes.

as-pine-proxy:
[Treeherder: 81ab0161d96baa50fda2ecb403b896dfd8b697e8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=81ab0161d96baa50fda2ecb403b896dfd8b697e8)

PR #3610, opened 2017-09-26
fix(topics): Allow showing uninitialized topics to avoid height changes on receiving topics

Mardak:
Fix #3604. r?@k88hudson Make topics initial data be like the actual initial data of undefined and render the empty section until actual topics come in.

PR #3611, opened 2017-09-27
fix(card): Fade from gradient to loaded image only after loading

### LINKED ISSUE #3605: Fade in pocket images from gradient placeholder image
Right now they jump right in when they finally load from network

Mardak:
Fix #3605. r?@sarracini Adds an image loader that sets the "loaded" class only after loading instead of when the image is set.

See video with existing on left and with PR on right:
https://ed.agadak.net/as/activity-stream.fade-pocket.webm
This video includes the changes from #3610.


as-pine-proxy:
[Treeherder: 5db4841aec9a9bcf146a47b0580aa7b446d7451a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5db4841aec9a9bcf146a47b0580aa7b446d7451a)

PR #3618, opened 2017-09-27
chore:(ping-centre) Ping Centre shield ID filtering should happen per ping sent.

emtwo:


emtwo:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1403695 we will be adding a "global" PingCentre instance that can be used to send general health telemetry. The first use case for it will be sending an Activity Stream health ping. 

This instance of PingCentre shouldn't always filter for the same shield IDs each time its run. But in our case when we send our A-S ping, we do want to filter for the a-s shield ID's

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -363,7 +362,8 @@ this.TelemetryFeed = class TelemetryFeed {
 
   async sendEvent(event_object) {
     if (this.telemetryEnabled) {
-      this.pingCentre.sendPing(event_object);
+      this.pingCentre.sendPing(event_object,
+      {filter: ACTIVITY_STREAM_ID});
```
Mardak: nit: just keep this on the previous line

as-pine-proxy:
[Treeherder: f4d0afbb1b81ae5f0e3fbc0fbe531ef06f589af7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f4d0afbb1b81ae5f0e3fbc0fbe531ef06f589af7)

Mardak:
https://hg.mozilla.org/releases/mozilla-beta/rev/79375774e4fe

PR #3622, opened 2017-09-28
fix(styles): Always keep vertical scroll visible to avoid horizontal jank

### LINKED ISSUE #3621: Avoid horizontal visual jank when showing prerendered with Pocket then render without
The vertical scrollbar is initially rendered because we prerender both pocket and highlights. Then when it decides pocket shouldn't be shown (and only 1 row of highlights), the scrollbar goes away.

Mardak:
Fix #3621. r?@sarracini


as-pine-proxy:
[Treeherder: 081d60b0b0f35fc41093ab5785fe8109f7331bbc](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=081d60b0b0f35fc41093ab5785fe8109f7331bbc)

PR #3641, opened 2017-10-03
chore(build): Use dev builds of react in nightly for unofficial builds if debug pref is true

k88hudson:
How do we feel about this approach (i.e. adding configuration to `.mozconfig`) v.s. adding a `--debug` option to `npm run buildmc` v.s. adding a pref:

Advantages of the .mozconfig approach:
- it's consistent with how devtools turns on dev react builds
- it doesn't require the github repo to enable debugging, anyone who can build firefox can turn it on

Disadvantages
- it could be unexpectedly turned on if someone is turning on DEBUG_JS_MODULES for some other reason
- requires a full`./mach build` when .mozconfig is changed


Mardak:
Just to summarize the discussion on irc:

We can conditionally `jar.mn` package and check a hidden default-`false` `firefox.js` pref from `aboutNewTabService` when `!RELEASE_OR_BETA` for nightly/local builds. We would additionally package the `-dev.js` react files as well as a `-dev.html` activity stream files that load the appropriate js files. Similar to how prerendered html file is selected by `aboutNewTabService` it can select the `-dev.html` version if the hidden pref has a user set `true` value.

These `-dev.html` files will be english only and matching the two regular versions to make sure we can get dev output for both prerendered or not.

mozilla-central-patches/aboutnewtabservice.patch
```diff
@@ -0,0 +1,226 @@
...
+ Cu.import("resource://gre/modules/Services.jsm");
++Cu.import("resource://gre/modules/AppConstants.jsm");
++
+ XPCOMUtils.defineLazyModuleGetter(this, "AboutNewTab",
+                                   "resource:///modules/AboutNewTab.jsm");
+@@ -16,4 +18,5 @@ XPCOMUtils.defineLazyModuleGetter(this, 
+ const LOCAL_NEWTAB_URL = "chrome://browser/content/newtab/newTab.xhtml";
+ 
++const BASE_AS_PATH = "resource://activity-stream/data/content/activity-stream";
+ const ACTIVITY_STREAM_URL = "resource://activity-stream/data/content/activity-stream.html";
```
Mardak: Looks like this (and `ACTIVITY_STREAM_PRERENDER_URL` are unused now)

bin/render-activity-stream-html.js
```diff
@@ -111,11 +109,14 @@ function main() {
   const {html, state} = prerender(options.locale);
   options.direction = state.App.textDirection;
 
-  fs.writeFileSync(HTML_FILE_PATH, templateHTML(options));
-  fs.writeFileSync(PRERENDERED_HTML_FILE_PATH, templateHTML(options, html));
-  fs.writeFileSync(INITIAL_STATE_JS_FILE_PATH, templateJs(state));
+  fs.writeFileSync(path.join(BASE_FILE_PATH, "activity-stream.html"), templateHTML(options));
+  fs.writeFileSync(path.join(BASE_FILE_PATH, "activity-stream-debug.html"), templateHTML(options, null, true));
```
Mardak: Looks like you want options on your options! ;) And that I believe passing a bool argument is generally bad API design. But ah well, it's a `bin` script.

mozilla-central-patches/aboutnewtabservice.patch
```diff
@@ -0,0 +1,226 @@
...
++    if (!IS_RELEASE_OR_BETA) {
++      this._activityStreamDebug = Services.prefs.getBoolPref(PREF_ACTIVITY_STREAM_DEBUG, false);
++    }
+     this._newtabURL = ABOUT_URL;
+     return true;
+@@ -152,5 +168,5 @@ AboutNewTabService.prototype = {
+   get defaultURL() {
+     if (this.activityStreamEnabled) {
+-      return this.activityStreamPrerender ? this.activityStreamPrerenderURL : this.activityStreamURL;
++      return `${BASE_AS_PATH}${this.activityStreamPrerender ? "-prerendered" : ""}${this.activityStreamDebug ? "-debug" : ""}.html`
```
Mardak: A comment for the expanded strings (maybe just the file part) would be good for anyone searching for how these files get referenced. E.g., `// activity-stream.html, activity-stream-prerendered.html, activity-stream-debug.html`

mozilla-central-patches/aboutnewtabservice.patch
```diff
@@ -0,0 +1,226 @@
...
+-   * Returns the activity stream resource URL for the newtab page
++   * Returns true if the the debug pref for activity stream is true
+    */
+-  readonly attribute ACString activityStreamURL;
+-
+-  /**
+-   * Returns the prerendered activity stream resource URL for the newtab page
+-   */
+-  readonly attribute ACString activityStreamPrerenderURL;
++  readonly attribute bool activityStreamDebug;
```
Mardak: I suppose if these were only used for testing, it should be safe to remove…

mozilla-central-patches/aboutnewtabservice.patch
```diff
@@ -0,0 +1,226 @@
...
+ function AboutNewTabService() {
+@@ -32,4 +38,9 @@ function AboutNewTabService() {
+   Services.prefs.addObserver(PREF_ACTIVITY_STREAM_ENABLED, this);
+   Services.prefs.addObserver(PREF_ACTIVITY_STREAM_PRERENDER_ENABLED, this);
++  this.initialized = true;
++
++  if (!IS_RELEASE_OR_BETA) {
++    Services.prefs.addObserver(PREF_ACTIVITY_STREAM_DEBUG, this);
++  }
+   this.toggleActivityStream();
```
Mardak: Probably move the `this.initialized = true` line (from between `addObserver`s) to next to `toggleActivityStream` with a comment that more initialization happens there

mozilla-central-patches/aboutnewtabservice.patch
k88hudson: I still need `activityStreamDebug` in here for the purpose of testing though

bin/render-activity-stream-html.js
k88hudson: Generally I definitely agree... I could object assign again I suppose

as-pine-proxy:
[Treeherder: 04b80dec0e62d727919c117a83ccb588f6c74392](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=04b80dec0e62d727919c117a83ccb588f6c74392)

PR #3330, opened 2017-08-31
fix(styles): Update preferences to be scrollable with new styling

Mardak:
Fix #3256 and fix #3273. r?@rlr 

short:
image

tall:
image

PR #3642, opened 2017-10-03
feat(sections): Add support for nested prefs

csadilek:
Enhancement to support nested prefs in Sections.

This will be needed when
- refactoring TopSites to a Section as it has a nested pref ("Show two rows")
- enhancing the spoc experiment to support opt-out functionality via a nested pref

To declare nested prefs in SectionManager:

```javascript
const BUILT_IN_SECTIONS = {
  "feeds.section.topstories": options => ({
  id: "topstories",
  pref: {
    titleString: {id: "header_recommended_by", values: {provider: options.provider_name}},
    descString: {id: options.provider_description || "pocket_feedback_body"},
    nestedPrefs: [{
      name: "foo",
        titleString: {id: "Some nested pref", values: {}},
        icon: "icon-info"
      }, {
        name: "bar",
        titleString: {id: "Some other nested pref", values: {}},
        icon: "icon-info"
    }]
  ...
```

UI:
sectionmanager-screenshot2

### DIFF
--- a/system-addon/content-src/components/PreferencesPane/PreferencesPane.jsx
+++ b/system-addon/content-src/components/PreferencesPane/PreferencesPane.jsx
@@ -110,10 +110,20 @@ class PreferencesPane extends React.PureComponent {
               {sections
                 .filter(section => !section.shouldHidePref)
                 .map(({id, title, enabled, pref}) =>
-                  <PreferencesInput key={id} className="showSection" prefName={(pref && pref.feed) || id}
-                    value={enabled} onChange={(pref && pref.feed) ? this.handlePrefChange : this.handleSectionChange}
-                    titleString={(pref && pref.titleString) || title} descString={pref && pref.descString} />)}
+                  <div key={id}>
+                    <PreferencesInput className="showSection" prefName={(pref && pref.feed) || id}
+                      value={enabled} onChange={(pref && pref.feed) ? this.handlePrefChange : this.handleSectionChange}
+                      titleString={(pref && pref.titleString) || title} descString={pref && pref.descString} />
 
+                        {pref.nestedPrefs && pref.nestedPrefs.map(nestedPref =>
+                          <div key={pref.nestedPrefs.name} className={`options${enabled ? "" : " disabled"}`}>
+                            <PreferencesInput prefName={nestedPref.name} disabled={!enabled} value={prefs[nestedPref.name]}
+                              onChange={this.handlePrefChange} titleString={nestedPref.titleString}
+                              labelClassName={`icon ${nestedPref.icon}`} />
+                          </div>
+                        )}
+                  </div>
+                )}
               <hr />
 
               <PreferencesInput className="showSnippets" prefName="feeds.snippets"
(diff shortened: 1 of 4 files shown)


system-addon/content-src/components/PreferencesPane/PreferencesPane.jsx
```diff
@@ -110,10 +110,20 @@ class PreferencesPane extends React.PureComponent {
...
                 .map(({id, title, enabled, pref}) =>
-                  <PreferencesInput key={id} className="showSection" prefName={(pref && pref.feed) || id}
-                    value={enabled} onChange={(pref && pref.feed) ? this.handlePrefChange : this.handleSectionChange}
-                    titleString={(pref && pref.titleString) || title} descString={pref && pref.descString} />)}
+                  <div key={id}>
+                    <PreferencesInput className="showSection" prefName={(pref && pref.feed) || id}
+                      value={enabled} onChange={(pref && pref.feed) ? this.handlePrefChange : this.handleSectionChange}
+                      titleString={(pref && pref.titleString) || title} descString={pref && pref.descString} />
 
+                        {pref.nestedPrefs && pref.nestedPrefs.map(nestedPref =>
```
Mardak: This indentation makes it look like these nested prefs are children of `PreferencesInput` and indeed the `options` for TopSites does this same sibling behavior. @rlr noted there was some oddness of not having the nested things as actually nested… I think this was in the context of when we were making the pane scrollable. Not sure if we should just insert `props.children` for `PreferencesInput` ?

csadilek: Good question. We'd need to pass activity stream prefs (or at least the required values) along as well and should probably also move the handler functions. Might be easier to refactor when moving TopSites to a Section to get rid of the special case, but not sure either. I guess it depends on what problem the current nesting is causing?

Mardak: I don't think there's a current bug, but we did run into one during https://github.com/mozilla/activity-stream/pull/3330#discussion_r136465642 where I had set a `section:last-child {margin-bottom}` to make sure the last item is scrollable.. but turns out the `div.options` was actually the last item! Also, there's other pref sections now and we added the padding to the parent instead of margin on the last item, so again, no bug right now…

csadilek: OK, I am up for trying the refactoring. I am thinking in a separate commit/pr though...

k88hudson: I think it's fine to have the nestedPrefs be on the same level as the main one for now, but I agree the indentation is a little hard to read/misleading; perhaps you could change it to something like [this](https://gist.github.com/k88hudson/b4684bbce7d2fb181ae1c51b3672bed1)?

csadilek:
Thanks @k88hudson! I will fix the nit, also attempting the refactoring @Mardak suggested. Let's wait a  bit before merging.

csadilek:
@Mardak @k88hudson OK, this cleaned up nicely. Added a separate commit for the refactoring. Can you take another look?

as-pine-proxy:
[Treeherder: 7fc28889684aac89338ccbf7707334ab9a619a16](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7fc28889684aac89338ccbf7707334ab9a619a16)

PR #3652, opened 2017-10-04
Log TopSites icons stats in session object.

### LINKED ISSUE #3449: Add telemetry for types of images shown for top sites
Could be useful to know how many users are seeing just screenshots (favicon too low quality) vs screenshot + icon in corner (#3150 vs high res icons (#3327).

I went through Dismiss/blocking top sites until I got all rich icons and my ~70th most frecent page got the 12th rich icon top site.

piatra:
Closes #3449.
I've moved the decision logic for topsite image and favicon from `TopSiteLink.render` to TopSites into it's own fn `selectTopSiteIcon`. This way on render `<TopSites>` knows the image quality of all its children. I also think it makes it easier to test.
The information about the 3 types of TopSite images: `tippytop`, `screenshot + favicon`, `screenshot + fallback letter` are added to the user session and updated every time TopSites updates. They are only sent once when the session ends.

@emtwo we currently handle tippytop and rich icons as one (because visually it's the same thing when we render). Are we interested in splitting the telemetry information?

### DIFF
--- a/docs/v2-system-addon/data_dictionary.md
+++ b/docs/v2-system-addon/data_dictionary.md
@@ -171,9 +171,11 @@ and losing focus. | :one:
 | `highlights_data_late_by_ms` | [Optional] Time in ms it took for Highlights to become initialized | :one:
 | `topsites_data_late_by_ms` | [Optional] Time in ms it took for TopSites to become initialized | :one:
 | `topsites_first_painted_ts` | [Optional][Service Counter][Server Alert for too many omissions] Timestamp of when the Top Sites element finished painting (possibly with only placeholder screenshots) | :one:
-| `topsites_size` | [Optional] The size of the Topsites set. | :one:
-| `topsites_screenshot` | [Optional] The size of the Topsites set with screenshot metadata. | :one:
-| `topsites_tippytop` | [Optional] The size of the Topsites set with TippyTop metadata. | :one:
+| `screenshot_with_icon` | [Optional] Number of topsites that display a screenshot and a favicon. | :one:
+| `screenshot` | [Optional] Number of topsites that display only a screenshot. | :one:
+| `tippytop` | [Optional] Number of topsites that display a tippytop icon. | :one:
+| `rich_icon` | [Optional] Number of topsites that display a high quality favicon. | :one:
+| `no_image` | [Optional] Number of topsites that have no screenshot. | :one:
 | `visibility_event_rcvd_ts` | [Optional][Server Counter][Server Alert for too many omissions] DOMHighResTimeStamp of when the page itself receives an event that document.visibilityState == visible. | :one:
 | `tiles` | [Required] A list of tile objects for the Pocket articles. Each tile object mush have a ID, and optionally a "pos" property to indicate the tile position | :one:
 | `click` | [Optional] An integer to record the 0-based index when user clicks on a Pocket tile. | :one:
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -6,29 +6,88 @@ const TopSitesEdit = require("./TopSitesEdit");
 const {TopSite, TopSitePlaceholder} = require("./TopSite");
 const CollapsibleSection = require("content-src/components/CollapsibleSection/CollapsibleSection");
 const ComponentPerfTimer = require("content-src/components/ComponentPerfTimer/ComponentPerfTimer");
+const {actionCreators: ac, actionTypes: at} = require("common/Actions.jsm");
+const {MIN_RICH_FAVICON_SIZE, MIN_CORNER_FAVICON_SIZE} = require("./TopSitesConstants");
 
-const TopSites = props => {
-  const realTopSites = props.TopSites.rows.slice(0, props.TopSitesCount);
-  const placeholderCount = props.TopSitesCount - realTopSites.length;
-  const infoOption = {
-    header: {id: "settings_pane_topsites_header"},
-    body: {id: "settings_pane_topsites_body"}
+/**
+ * Iterates through TopSites and counts types of images.
+ * @param acc Accumulator for reducer.
+ * @param topsite Entry in TopSites.
+ */
+function countTopSitesIconsTypes(topSites) {
+  const countTopSitesTypes = (acc, link) => {
+    if (link.tippyTopIcon) {
+      acc.tippytop++;
+    } else if (link.faviconSize >= MIN_RICH_FAVICON_SIZE) {
+      acc.rich_icon++;
+    } else if (link.screenshot && link.faviconSize >= MIN_CORNER_FAVICON_SIZE) {
+      acc.screenshot_with_icon++;
+    } else if (link.screenshot) {
+      acc.screenshot++;
+    } else {
+      acc.no_image++;
+    }
+
+    return acc;
   };
-  return (<ComponentPerfTimer id="topsites" initialized={props.TopSites.initialized} dispatch={props.dispatch}>
-    <CollapsibleSection className="top-sites" icon="topsites" title={<FormattedMessage id="header_top_sites" />} infoOption={infoOption} prefName="collapseTopSites" Prefs={props.Prefs} dispatch={props.dispatch}>
-      <ul className="top-sites-list">
-        {realTopSites.map((link, index) => link && <TopSite
-          key={link.guid || link.url}
-          dispatch={props.dispatch}
-          link={link}
-          index={index}
-          intl={props.intl} />)}
-        {placeholderCount > 0 && [...Array(placeholderCount)].map((_, i) => <TopSitePlaceholder key={i} />)}
-      </ul>
-      <TopSitesEdit {...props} />
-    </CollapsibleSection>
-  </ComponentPerfTimer>);
-};
+
+  return topSites.reduce(countTopSitesTypes, {
+    "screenshot_with_icon": 0,
+    "screenshot": 0,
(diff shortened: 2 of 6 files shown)


emtwo:
@piatra Thanks for this PR! I think if tippy top and rich icons are visually the same then it makes sense to report them as the same. I just had a discussion with @ncloudioj that the server side will compress all the topsite icon counts into one bigint column. This means if we change our minds at any point and want to add another icon type we can do so quite easily. But for now we can keep those two together.

piatra:
After some discussion the final form of the session ping will include the TopSite icons stats in a separate field of the session. It will count rich icons and tippytop separately. PR to follow.
```
topsites_icon_stats: {rich_icon: 6, screenshot: 3, tippytop: 1, screenshot_with_icon: 2}
```

piatra:
@ncloudioj while trying to document the new fields I found that our [documentation is not quite to date](https://github.com/mozilla/activity-stream/blame/master/docs/v2-system-addon/data_dictionary.md#L173-L174). I couldn't find those fields anywhere else. Should those be removed? Also is that the right place to document `topsites_icon_stats` entries?

piatra:
@k88hudson There's a weird situation where `<TopSites>` & `<TopSitesEdit>` render different `<TopSite>` components. The `Edit` component was couting on `<TopSite>` to decide what image/favicon combination to use but now that logic has been extracted to `<TopSites>`. 

[My fix](https://github.com/mozilla/activity-stream/pull/3652/commits/0a505fea6477ca9b9abf9384e0ee6287f190a0b7#diff-471c641a260f185a42371906d3b094f0R73) was to send down `realTopSites` as a prop from `<TopSites>`, the `||` is in place only to get the tests to pass (`<TopSitesEdit>` always receives `realTopSites` except in the tests where that setup was not updated). 

I wanted to check first that sending down `realTopSites` as a prop is enough to get this landed & create a follow up bug to remove two different `<TopSites>` views.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -6,9 +6,92 @@ const TopSitesEdit = require("./TopSitesEdit");
...
+
+  const topSitesIconsStats = realTopSites.reduce(countTopSitesIconsTypes, {
+    "screenshot_with_icon": 0,
+    "screenshot": 0,
+    "tippytop": 0,
+    "rich_icon": 0
+  });
+
+  // Dispatch telemetry event with the count of TopSites images types.
+  props.dispatch(ac.TopSitesStats(topSitesIconsStats));
```
k88hudson: I don't think dispatching/checking directly in render is the right place for this (it will result in a lot of duplicate stats being sent), you probably want to do something similar to how top stories counts impressions

k88hudson: Could you maybe do a check in `componentDidUpdate/componentDidMount`? How often do you want to check stats? Every time a change happens? When the user first sees it?

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -6,9 +6,92 @@ const TopSitesEdit = require("./TopSitesEdit");
...
+
+  return acc;
+}
 
 const TopSites = props => {
-  const realTopSites = props.TopSites.rows.slice(0, props.TopSitesCount);
+  const realTopSites = props.TopSites.rows.slice(0, props.TopSitesCount)
+                        .map(selectTopSiteIcon);
+
+  const topSitesIconsStats = realTopSites.reduce(countTopSitesIconsTypes, {
```
k88hudson: Personally I would go with `countTopSitesIconsTypes` just taking `realTopSites` as an argument and add this `.reduce`/initial state for the accumulator in that function

system-addon/content-src/components/TopSites/TopSites.jsx
piatra: @k88hudson Talking to @emtwo the only requirement is that we send them for every session. 
I'm thinking it would be best to add it to componentDidUpdate. It would cover cases such as "a user pinned a topsite, we need to fetch a screenshot".

piatra: But Topsites is not a component/doesn't have a componentDidUpdate :(

Mardak: `TopSites` is just a shorthand component where instead of a `class`, it's just a function that implements what would be in a `render` method.

piatra:
@k88hudson Topsites is not a component, there are no life cycle methods. Could moving the reporting to a life cycle method be a follow up?

piatra:
@ncloudioj can you please have a look at the payload and documentation? Thanks.

Mardak:
I believe what @k88hudson is indirectly referring to when asking "when to check stats" is that by doing it from `render`, it will happen every time React wants to render something, and this could possibly be many times without actually mounting. I.e., do we want to report stats every time React internally needs the result of `render`?

piatra:
@Mardak the TopSiteStats are stored in `session.perf` so nothing gets reported until `NEW_TAB_UNLOAD`. Inefficient because a lot of messages are sent internally but maybe good enough to land as is and actually make a TopSites component in a separate PR?

Mardak:
@k88hudson I wonder if we'll want some rule to disallow `function` components if there's undesired overhead of converting to `class` components?

piatra:
@k88hudson Updated to a class component. Not sure what to do about the [two](https://github.com/mozilla/activity-stream/pull/3652/files#diff-e37494e6f7005fab440ea335b2fe1f90R93) [calls](https://github.com/mozilla/activity-stream/pull/3652/files#diff-e37494e6f7005fab440ea335b2fe1f90R103) to `selectTopSitesIcon`. Could I use state + `componentWillReceiveProps` to only compute once and reuse? We don't use state that much in our code so maybe there's a different approach?

k88hudson:
@piatra the easiest way to do it is probably just to check `prevProps` in `componentDidUpdate` to see if the data you care about (i.e. rows) actually changed. See this code here:
- https://github.com/mozilla/activity-stream/blob/master/system-addon/content-src/components/Sections/Sections.jsx#L89

docs/v2-system-addon/data_dictionary.md
```diff
@@ -171,6 +171,7 @@ and losing focus. | :one:
 | `highlights_data_late_by_ms` | [Optional] Time in ms it took for Highlights to become initialized | :one:
 | `topsites_data_late_by_ms` | [Optional] Time in ms it took for TopSites to become initialized | :one:
 | `topsites_first_painted_ts` | [Optional][Service Counter][Server Alert for too many omissions] Timestamp of when the Top Sites element finished painting (possibly with only placeholder screenshots) | :one:
+| `topsites_icon_stats` | [Optional] Object with keys: "screenshot_with_icon", "screenshot", "tippytop", "rich_icon" and the value will be the number of TopSites with icons in the respective category | :one:
```
ncloudioj: nit: Each key in the `topsites_icon_stats` should be described separately here.

Also, please delete `topsites_size`, `topsites_screenshot`, and `topsites_tippytop`, as those are only for the test pilot.

ncloudioj:
@piatra payload schema and document r+, just dropped one comment in `data_dictionary.md`.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -6,29 +6,75 @@ const TopSitesEdit = require("./TopSitesEdit");
...
+  return topSites.reduce(countTopSitesTypes, {
+    "screenshot_with_icon": 0,
+    "screenshot": 0,
+    "tippytop": 0,
+    "rich_icon": 0
+  });
+}
+
+class TopSites extends React.PureComponent {
+  componentDidUpdate(props) {
```
k88hudson: Can you also run this on `componentDidMount`? This won't cover the first load case

piatra:
@fmarier this patch adds a new telemetry entry containing information about the quality of the images from your topsites. Details are here https://github.com/mozilla/activity-stream/pull/3652/files#diff-f92d07e23fd9de1eb2f4ad96bd041287R174 it does not send the images, or urls only information regarding the number of topsites that fall into the different categories:
 * high quality image
 * tippytop
 * screenshot fallback
 * favicon fallback

fmarier:
> it does not send the images, or urls only information regarding the number of topsites that fall into the different categories

Based on that, it seems like Category 1 data.

datareview+

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -6,29 +6,88 @@ const TopSitesEdit = require("./TopSitesEdit");
...
+
+  componentDidMount() {
+    const realTopSites = this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
+
+    const topSitesIconsStats = countTopSitesIconsTypes(realTopSites);
+    // Dispatch telemetry event with the count of TopSites images types.
+    this.props.dispatch(ac.SendToMain({
+      type: at.SAVE_SESSION_PERF_DATA,
+      data: {topsites_icon_stats: topSitesIconsStats}
+    }));
```
Mardak: Is there supposed to be any differences between counting and dispatching stats in `componentDidUpdate` vs `componentDidMount`?

piatra: No, we just do it twice to capture the preloaded case.

Mardak: Why was this written in a way to allow for differences between the two leading to hard to main code? `componentDidUpdate` and `componentDidMount` can call other methods.

Issue #3638, opened 2017-10-03
Reduce duplicates places queries near startup


Mardak:
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1401817
Right now TopSitesFeed and HighlightsFeed both call and expire places results at least 2 times on INIT, NEW_TAB_LOAD.

HighlightsFeed additionally does work on TOP_SITES_UPDATED.

piatra:
Both TopSitesFeed and HighlightsFeed refresh twice from INIT and NEW_TAB_LOAD and it looks like NEW_TAB_LOAD is always second.  The event is triggered by `ActivityStreamMessageChannel` [that is created after the feeds are initialized](https://github.com/mozilla/activity-stream/blob/444feabd2e98ed585d7ba97119e6005c34e00b70/system-addon/lib/Store.jsm#L135).

* TopSites are not affected by NEW_TAB_LOAD because they update only once every 30 minutes.
* Highlights will update on every NEW_TAB_LOAD if they have less than 9 entries.
  * I think we could easily add an extra `&& Date.now() - this.lastUpdated > 30 seconds` to counter this.
  * Highlights already clear the cache on changes to bookmarks and places so actual "stale data" could not occur as a result of a 30s delay.

The extra work done by the HighlightsFeed on TOP_SITES_UPDATED isn't that important. It will hit the cache.

Thoughts? @Mardak @k88hudson

piatra:
I'm also looking into writing a mochitest that loads AS and tests how many times we call into NewTabUtils like `getHighlights`.

Mardak:
I believe if you look at dispatch calls during startup, you should be able to see two from TopSites because the `refresh` call from INIT and NEW_TAB_LOAD happen concurrently (blocked on `this._tippyTopProvider.init`).

Also, even if Highlights hits the cache, it's most likely unnecessarily dispatching. Although that could be treated as its own issue as one was filed in terms of places queries.

I believe @k88hudson has mentioned just removing the "fewer than 9" Highlights check similar to how it was removed from TopSites.

PR #3653, opened 2017-10-05
fix(cache): Allow concurrent cache callers to get links with shared state

### LINKED ISSUE #3561: Don't update link objects that have already been dispatched
Noted from https://github.com/mozilla/activity-stream/pull/3530#issuecomment-330893271

This is because doing the updates would be 
> breaking the redux model that makes state immutable outside of reducers

### LINKED ISSUE #3564: Early top sites refreshes unnecessarily process duplicate screenshot requests
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1401817

https://perfht.ml/2yewfB1

We might be able to just change the screenshot size:
https://searchfox.org/mozilla-central/source/browser/app/profile/firefox.js#1620-1624

From 560x380 to 192x192. That's almost 6x smaller.

Mardak:
Fix #3564. r?@k88hudson The main change is adding a `__sharedCache` that copies of a given link will share via shallow copying. I also stashed the `updateLink` helper on that object so only one thing needs to be `delete`d before dispatch. I also made it so that `Screenshots` doesn't modify the link as generally it'll be too late anyway and would need a `dispatch`, so this fixes #3561.

Added concurrency tests where the first one (backing data called once) was already passing, but the two additional ones are specifically testing for #3564's issues. Other tests were updated for the various naming changes.

### DIFF
--- a/system-addon/lib/LinksCache.jsm
+++ b/system-addon/lib/LinksCache.jsm
@@ -19,19 +19,21 @@ this.LinksCache = class LinksCache {
    *
    * @param {object} linkObject Object containing the link property
    * @param {string} linkProperty Name of property on object to access
-   * @param {function} migrator Optional callback receiving the old and new link
-   *                            to allow custom migrating data from old to new.
+   * @param {array} properties Optional properties list to migrate to new links.
    * @param {function} shouldRefresh Optional callback receiving the old and new
    *                                 options to refresh even when not expired.
    */
-  constructor(linkObject, linkProperty, migrator = () => {}, shouldRefresh = () => {}) {
+  constructor(linkObject, linkProperty, properties = [], shouldRefresh = () => {}) {
     this.clear();
+
     // Allow getting links from both methods and array properties
     this.linkGetter = options => {
       const ret = linkObject[linkProperty];
       return typeof ret === "function" ? ret.call(linkObject, options) : ret;
     };
-    this.migrator = migrator;
+
+    // Always migrate the shared cache data in addition to any custom properties
+    this.migrateProperties = ["__sharedCache", ...properties];
     this.shouldRefresh = shouldRefresh;
   }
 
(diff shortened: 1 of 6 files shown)


system-addon/lib/LinksCache.jsm
```diff
@@ -78,26 +80,31 @@ this.LinksCache = class LinksCache {
...
 
-        // Migrate data to the new link if we have an old link
+        // Modify each of the link objects by migrating or adding data
         for (const newLink of copied) {
-          if (newLink) {
-            const oldLink = toMigrate.get(newLink.url);
-            if (oldLink) {
-              this.migrator(oldLink, newLink);
-            }
+          if (!newLink) {
```
piatra: Is there a reason we can't filter these out before the map on line 84?

Mardak: We need to keep holes for pinned links

Mardak: I suppose I could clean this up a little to avoid the `map` then `for of` both checking for object existing.

Mardak: Saved a few lines with better clarity (added comment about holes) `1 file changed, 9 insertions(+), 13 deletions(-)`

k88hudson:
Ok the code (and tests) look good to me, although I'm having a hard time understanding the issue described in #3564 and why this fixes it; feel free to land, although maybe you can help walk me through this tmr at some point?

Mardak:
I don't quite understand #3564 either… ;) But one of the problems noted in the bug is that we process screenshots "many times for the same URL." https://bugzilla.mozilla.org/show_bug.cgi?id=1401817#c13

I would think at best this would cut down processing time in half as we would only process one set of top sites screenshots instead unnecessarily doing it twice (`INIT` and `NEW_TAB_LOAD` -- not sure what's causing "many times" from the bug report). So for jryan, instead of seeing 60x worse performance than what I see, it would only be 30x worse…

Mardak:
I split out the other aspects of the original bug to #3661 and #3638.

as-pine-proxy:
[Treeherder: 0f95b59703cb865558b2ac1e0d03969abb9ad1b5](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0f95b59703cb865558b2ac1e0d03969abb9ad1b5)

PR #3658, opened 2017-10-05
fix(styles): Only set a fade transition delay for about:newtab

### LINKED ISSUE #3657: Remove fade delay for about:home
#3640 will be to remove the delay for everywhere, but in the mean time we can just keep the delay for about:newtab to avoid the tart regression.

Mardak:
Fix #3657. r?@k88hudson


as-pine-proxy:
[Treeherder: e94373f4b4e2528b7e32604033dc5be73aa6fa04](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=e94373f4b4e2528b7e32604033dc5be73aa6fa04)

PR #3660, opened 2017-10-05
feat(reporting): Add prerendering and preloading to the session ping

### LINKED ISSUE #3144: Record in perf object if the browser started preloaded
Split from #3118. We can provide some additional insight by remembering whether a browser started as a preloaded browser. This can be detected on handling `RemotePage:Init`:

This piece of code for `ActivityStreamMessageChannel` seems to detect preloaded-ness:
```diff
   onNewTabInit(msg) {
-    this.onActionFromContent({type: at.NEW_TAB_INIT}, msg.target.portID);
+    this.onActionFromContent({
+      type: at.NEW_TAB_INIT,
+      preloaded: msg.target.browser === msg.target.browser.ownerGlobal.gBrowser._preloadedBrowser
+    }, msg.target.portID);
   }
```

Without this, we can mostly infer when a page is preloaded by checking the `topsites_first_painted_ts` being before `visibility_event_rcvd_ts` but in the slight case of a preloaded browser started and soon became visible before painting, that could be confused with the case where a non-preloaded about:newtab is opened.

### LINKED ISSUE #3498: Add telemetry for prerendered or not
We can already detect preloading based on some timing numbers but that's covered in #3144. Prerendering probably will have impact on various metrics, so it would be good to be able to differentiate

ncloudioj:
This closes #3498 fix #3144


ncloudioj:
@fmarier r? for data review :)

sarracini:
Looking at this patch now, I actually think @dmose would be a better person to review. He's got a better understanding of the telemetry this would affect, and he's got the right background on both the tickets that this would be fixing. I'm going to re-assign to Dan

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -274,6 +280,37 @@ this.TelemetryFeed = class TelemetryFeed {
...
+   *      does not take that into account provided that it should not skew the
+   *      global measurement.
+   *
+   * @param {obj} session A session object
+   */
+  setPageModeForPerf(session) {
+    let pageMode = 0;
+
+    if (this._prefs.get("prerender")) {
+      pageMode |= PAGE_MODES_ENCODING.prerendered;
```
Mardak: This gets called at the session end where the user might have changed something to have `prerender` pref change. The pref check should be fine when the page is loading though.

ncloudioj: Good catch. Since this function relies on the other two fields from `session.perf`, perhaps the best place to call it is inside `saveSessionPerfData`.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -274,6 +280,37 @@ this.TelemetryFeed = class TelemetryFeed {
...
+    }
+
+    if (session.page !== "about:home" &&
+        Services.prefs.getBoolPref("browser.newtab.preload") &&
+        session.perf.topsites_first_painted_ts < session.perf.load_trigger_ts) {
+      pageMode |= PAGE_MODES_ENCODING.preloaded;
```
dmose: This also feels susceptible to races.  One that jumps out at me is when preloading starts to happen, then the user triggers a load, then the first paint from the preload finishes.  I'm wondering if there's a way to fix this by using or adapting one of the [`tabbrowser.xml` attributes](http://searchfox.org/mozilla-central/source/browser/base/content/tabbrowser.xml#2194), but it's not behaving quite how I expect, so I'm not yet sure.

system-addon/lib/TelemetryFeed.jsm
ncloudioj: Yep, I like this idea!

docs/v2-system-addon/data_events.md
```diff
@@ -290,6 +290,20 @@ perf: {
...
+  //   "regular": 0,
+  //   "preloaded": 1 << 0,
+  //   "prerendered": 1 << 1
+  // }
+  "page_mode": [
+    0 | // regular
+    1 | // preloaded
+    2 | // prerendered
+    3 | // prerendered&preloadedi
+  ],
```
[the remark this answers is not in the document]
ncloudioj: Doing integer encoding here is not for saving space, instead, it's intended to make the table schema slim and compact. This trick enables us to add new options to it without adding new columns to the table schema. Unless we're certain that it's unlikely to have new loading optimizations in the future, I'd recommend to do it in this way. We have done the same encoding for recording the user's preferences, turns out it's been working quite well both on the client and the server side.

Indeed, using encoded integer hurts the readability, as a workaround, we can also include those original booleans in the ping for the better readability, only the encoded integer will be stored in the database on the server side.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -274,6 +280,37 @@ this.TelemetryFeed = class TelemetryFeed {
...
+   *   "regular": 0,
+   *   "preloaded": 1 << 0,
+   *   "prerendered": 1 << 1
+   * }
+   *
+   * Note:
+   *   1. Preloading doesn't apply to about:home.
+   *   2. The first newtab page is not preloaded, nor prerendered. This function
+   *      does not take that into account provided that it should not skew the
+   *      global measurement.
```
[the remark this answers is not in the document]
ncloudioj: Sure, will do.

system-addon/lib/TelemetryFeed.jsm
ncloudioj: Hmm, not sure, depends on how often that happens in practice. I believe that `session.perf.topsites_first_painted_ts < session.perf.load_trigger_ts` should cover the vast majority of preloading loads.

dmose: Agreed, this isn't likely to be a huge problem, but if we can keep the data clean without too much work, that can save us pain and context-switching in the future.

dmose: http://searchfox.org/mozilla-central/source/browser/base/content/tabbrowser.xml#2223 is actually the link I should have given above.

dmose: https://bugzilla.mozilla.org/show_bug.cgi?id=1376895#c5 is where that isPreloaded attribute came from.  I suspect the easiest thing to do would be to talk to mconley about whether making that attribute persist (or change names?) rather than being removed.

dmose: Though of course we'd still want to remove the attribute when the client navigates away from the page (there is already code in the tree to do that).

ncloudioj:
Have tweaked the patch a bit:
* Drop the encoded integer to favor the better readability of two booleans
* Add a "PAGE_PRERENDERED" action dispatched in `base#componentDidUpdate`
* Capture "is_preloaded" via the trick @Mardak suggested in [#3144](https://github.com/mozilla/activity-stream/issues/3144#issue-249470787), this actually solves a bunch of race conditions in the original implementation

@dmose Another r?

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -274,6 +278,32 @@ this.TelemetryFeed = class TelemetryFeed {
...
+  }
+
+  /**
+   * handleNewTabInit - Handle NEW_TAB_INIT
+   *
+   * @param  {obj} action the Action object
+   */
+  handleNewTabInit(action) {
+    const session = this.addSession(au.getPortIdOfSender(action), action.data.url);
+    session.perf.is_preloaded = action.data.browser === action.data.browser.ownerGlobal.gBrowser._preloadedBrowser;
```
Mardak: Oh, this is probably an easier check ;)
https://searchfox.org/mozilla-central/search?q=%22isPreloadBrowser%22&path=browser

`browser.getAttribute("isPreloadBrowser") == "true"`

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -274,6 +278,32 @@ this.TelemetryFeed = class TelemetryFeed {
...
+  }
+
+  /**
+   * handleNewTabInit - Handle NEW_TAB_INIT
+   *
+   * @param  {obj} action the Action object
+   */
+  handleNewTabInit(action) {
+    const session = this.addSession(au.getPortIdOfSender(action), action.data.url);
+    session.perf.is_preloaded = action.data.browser.getAttribute("isPreloadBrowser") === "true";
```
dmose: @krizsa This is getting called, indirectly, via a RemotePage:Init message handler in the main process.  Since I know this is removed by the redirect code from [bug 1376895](https://bugzilla.mozilla.org/show_bug.cgi?id=1376895#c5), I'm concerned that this could race, possibly, but not necessarily just, in the case where someone opens multiple tabs in a row.  Can you offer any insight into that whether that's true/possible?  If so, how would you feel about somehow persisting the isPreloadBrowser attribute on all tabs that were preloaded and haven't yet navigated away from about:newtab?  This would be super helpful for our telemetry...

Mardak: @dmose `isPreloadBrowser` attribute does stay until a different URI is loaded. Here's after opening some tabs very quickly:
```js
gBrowser.browsers.forEach(b => console.log(b.currentURI.spec, b.getAttribute("isPreloadBrowser"), b));

about:newtab true <browser type="content" message="true" messagemanagergroup="browsers" contextmenu="contentAreaContextMenu" tooltip="aHTMLTooltip" remoteType="web" remote="true" isPreloadBrowser="true" selectmenulist="ContentSelectDropdown" datetimepicker="DateTimePickerPanel" autoscrollpopup="autoscroller" nodefaultsrc="true" clickthrough="never" autocompletepopup="PopupAutoComplete"> 
about:newtab  <browser type="content" message="true" messagemanagergroup="browsers" contextmenu="contentAreaContextMenu" tooltip="aHTMLTooltip" remoteType="web" remote="true" autocompletepopup="PopupAutoComplete" selectmenulist="ContentSelectDropdown" datetimepicker="DateTimePickerPanel" autoscrollpopup="autoscroller" nodefaultsrc="true" clickthrough="never"> 
about:newtab true <browser type="content" message="true" messagemanagergroup="browsers" contextmenu="contentAreaContextMenu" tooltip="aHTMLTooltip" remoteType="web" remote="true" isPreloadBrowser="true" selectmenulist="ContentSelectDropdown" datetimepicker="DateTimePickerPanel" autoscrollpopup="autoscroller" nodefaultsrc="true" clickthrough="never" autocompletepopup="PopupAutoComplete"> 
about:newtab true <browser type="content" message="true" messagemanagergroup="browsers" contextmenu="contentAreaContextMenu" tooltip="aHTMLTooltip" remoteType="web" remote="true" isPreloadBrowser="true" selectmenulist="ContentSelectDropdown" datetimepicker="DateTimePickerPanel" autoscrollpopup="autoscroller" nodefaultsrc="true" clickthrough="never" autocompletepopup="PopupAutoComplete"> 
about:newtab true <browser type="content" message="true" messagemanagergroup="browsers" contextmenu="contentAreaContextMenu" tooltip="aHTMLTooltip" remoteType="web" remote="true" isPreloadBrowser="true" selectmenulist="ContentSelectDropdown" datetimepicker="DateTimePickerPanel" autoscrollpopup="autoscroller" nodefaultsrc="true" clickthrough="never" autocompletepopup="PopupAutoComplete"> 
about:newtab  <browser type="content" message="true" messagemanagergroup="browsers" contextmenu="contentAreaContextMenu" tooltip="aHTMLTooltip" remoteType="web" remote="true" autocompletepopup="PopupAutoComplete" selectmenulist="ContentSelectDropdown" datetimepicker="DateTimePickerPanel" autoscrollpopup="autoscroller" nodefaultsrc="true" clickthrough="never" primary="true">
```

dmose: @Mardak: interesting!  When I was poking around with DOMi, I wasn't able to actually see that attribute, so perhaps I was inadvertantly doing it wrong.  That said, the code I'm specifically concerned about is http://searchfox.org/mozilla-central/source/browser/base/content/browser.js#1170

dmose: @Mardak Which, it appears, might be the two cases in your example that don't have the "true" in them.

Mardak: The ones without "true" are probably because it wasn't preloaded. I opened the tabs quickly.

dmose: After a bunch of discussion wit @krizsa it sounds like that's right, and the code should be usable as-is. The redirect code I was concerned about just removes the attribute and causes it to be re-evaluated (and potentially re-added if necessary) in the parent process, because the child process doesn't have the information to know whether it's a preloaded browser or not.

@krizsa has said he would like to have one more look...

krizsa: Yes the code should be usable as-is. There are several ways to navigate away from about:newtab with different code paths, and to catch all the cases I had to add browser.removeAttribute("isPreloadBrowser"); at two places: http://searchfox.org/mozilla-central/search?q=removeAttribute%28%22isPreloadBrowser%22%29%3B&path=
The above mentioned RedirectLoad is handling the case where the navigation was requested from the child process (click on a link on about:newtab / search on about:newtab). Once the attribute is removed it won't ever get readded (since from that point on we won't care any more if it was a preloaded browser or not).

dmose: Ok, now I understand.  Thanks @krizsa and @Mardak for help clarifying.

dmose:
Looking at the checklist, this still wants a user story in https://github.com/mozilla/ping-centre/issues/, but I think you could probably just update mozilla/ping-centre#90 to include this.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -274,6 +278,32 @@ this.TelemetryFeed = class TelemetryFeed {
...
+    if (!session) {
+      // It's possible the tab was never visible – in which case, there was no user session.
+      return;
+    }
+
+    session.perf.is_prerendered = true;
+  }
+
+  /**
+   * handleNewTabInit - Handle NEW_TAB_INIT
```
dmose: This wants a bit more commentary describing what it does.  It also wants a test to check that it calls addSession.

system-addon/test/unit/lib/TelemetryFeed.test.js
```diff
@@ -531,13 +537,20 @@ describe("TelemetryFeed", () => {
...
+        type: at.NEW_TAB_INIT,
+        data: {url: "about:newtab", browser}
+      }));
+      assert.calledOnce(instance.handleNewTabInit);
+    });
     it("should call .addSession() on a NEW_TAB_INIT action", () => {
-      const stub = sandbox.stub(instance, "addSession");
+      const stub = sandbox.stub(instance, "addSession").returns({perf: {}});
       sandbox.stub(instance, "setLoadTriggerInfo");
-
```
dmose: I'd prefer leaving the whitespace for the readability reason suggested elsewhere.

system-addon/test/unit/lib/TelemetryFeed.test.js
```diff
@@ -629,6 +642,36 @@ describe("TelemetryFeed", () => {
       instance.onAction(action);
       assert.calledOnce(sendEvent);
     });
+    it("should call .handlePagePrerendered on a PAGE_PRERENDERED action", () => {
+      const session = {perf: {}};
+      sandbox.stub(instance.sessions, "get").returns(session);
+      sandbox.spy(instance, "handlePagePrerendered");
+      instance.onAction(ac.SendToMain({type: at.PAGE_PRERENDERED}));
+      assert.calledOnce(instance.handlePagePrerendered);
+      assert.ok(session.perf.is_prerendered);
```
dmose: You can remove the above line; you're already testing for that in the #handlePagePrerendered block.

ncloudioj:
Thanks @dmose !

Adding the user story and review fixes now.

system-addon/content-src/components/Base/Base.jsx
```diff
@@ -30,6 +30,7 @@ class Base extends React.PureComponent {
     // dispatched right after the store is ready.
     if (this.props.isPrerendered) {
       this.props.dispatch(ac.SendToMain({type: at.NEW_TAB_STATE_REQUEST}));
+      this.props.dispatch(ac.SendToMain({type: at.PAGE_PRERENDERED}));
```
Mardak: Any reason why we added a new action instead of adding some data to NEW_TAB_STATE_REQUEST?

ncloudioj: I believe NEW_TAB_STATE_REQUEST has its own semantics, correct? @k88hudson

PR #3662, opened 2017-10-06
fix(topsites): Refresh pinned links when one is blocked

### LINKED ISSUE #3609: Trying to dismiss a pinned site keeps it pinned
Probably should unpin when dismissing/blocking. Similar to #3294 for deleting from history.

Mardak:
Fix #3609. r?@sarracini In the first commit, I just cleaned up a top level `action` variable to avoid shadowing it and test the `target` behavior. The second commit adds an integration test for pinning then unpinning.


as-pine-proxy:
[Treeherder: 90e09519a2906cee656b17b18b92abb80c740ffb](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=90e09519a2906cee656b17b18b92abb80c740ffb)

PR #3670, opened 2017-10-09
fix (systemaddon): #3665 fix alignment of collapsible header

### LINKED ISSUE #3665: Collapsible section titles are higher by 1px causing a shift when editing
image

red: edit header
green: normal header

I'm guessing it's the arrowhead being placed 1px too high causing everything to be shifted down.

Maybe the same issue -- looks like the arrowhead is higher than center:

image

@rlr

rlr:
don't ask. because i don't know really know why. lolcss

Fixes #3665

### DIFF
--- a/system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
+++ b/system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
@@ -3,16 +3,14 @@
 
     .click-target {
       cursor: pointer;
+      vertical-align: top;
     }
 
-    .icon {
-      &.icon-arrowhead-down {
-        margin: -3px 0 0 8px;
-      }
-
-      &.icon-arrowhead-forward {
-        margin: -2px 0 0 8px;
-      }
+    .icon-arrowhead-down,
+    .icon-arrowhead-forward {
+      margin-top: -1px;
+      margin-inline-start: 8px;
+      transition: transform 0.5s $photon-easing;
     }
   }
 
--- a/system-addon/content-src/components/Topics/_Topics.scss
+++ b/system-addon/content-src/components/Topics/_Topics.scss
@@ -39,6 +39,8 @@
     color: $link-secondary;
 
     @media (min-width: $break-point-large) {
+      // This is floating to accomodate a very large number of topics and/or
+      // very long topic names due to l10n.
       float: right;
 
       &:dir(rtl) {
@@ -62,5 +64,13 @@
     &:dir(rtl)::after  {
       transform: scaleX(-1);
     }
+  }
+
+  // This is a clearfix to for the topics-read-more link which is floating and causes
+  // some jank when we set overflow:hidden for the animation.
+  &::after {
+    content: "";
+    display: table;
+    clear: both;
   }
 }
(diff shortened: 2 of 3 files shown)


rlr:
^ that fixes the icon positioning, the RTL issue, the highlights moving up at the end of the animation, ...

system-addon/content-src/components/Topics/_Topics.scss
```diff
@@ -39,11 +40,9 @@
...
 
     @media (min-width: $break-point-large) {
-      float: right;
-
-      &:dir(rtl) {
-        float: left;
-      }
+      position: absolute;
+      top: 0;
+      offset-inline-end: 0px;
```
Mardak: The `float` here was to allow for long topics to wrap correctly without overlapping the link (just in case some locales have longg topics). Here's with this change:

image

system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
```diff
@@ -3,15 +3,15 @@
...
-        margin: -3px 0 0 8px;
-      }
+    .icon-arrowhead-down,
+    .icon-arrowhead-forward {
+      margin: -1px 0 0 8px;
 
-      &.icon-arrowhead-forward {
-        margin: -2px 0 0 8px;
+      &:dir(rtl) {
+        margin: -1px 8px 0 0;
```
Mardak: Probably just do `margin-top: -1px` and `margin-inline-start: 8px` to not need the separate rtl check.

system-addon/content-src/components/Topics/_Topics.scss
rlr: gah. hate floats. they change size depending on if the parent is overflow visible or not. ok. maybe flex? or maybe force the height somehow.

rlr:
ok, i think we're good now. unless we want to add some spinning to the arrow 😄

system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
```diff
@@ -3,16 +3,13 @@
...
-        margin: -3px 0 0 8px;
-      }
-
-      &.icon-arrowhead-forward {
-        margin: -2px 0 0 8px;
-      }
+    .icon-arrowhead-down,
+    .icon-arrowhead-forward {
+      margin-top: -1px;
+      margin-inline-start: 8px;
```
Mardak: from slack re transition:
> I think it’s a nice touch

let's do it! same animation timing as the section content hiding/showing. Probably add it to this style block. Need to fix up the rotate degs to be the appropriate shortest path. Probably 90deg and -90deg for the appropriate rtl/ltr.

system-addon/content-src/components/Topics/_Topics.scss
```diff
@@ -63,4 +63,10 @@
       transform: scaleX(-1);
     }
   }
+
+  &::after {
+    content: "";
+    display: table;
+    clear: both;
```
Mardak: Let's have a comment describing why this is needed. For both this clearfix and the original float.

as-pine-proxy:
[Treeherder: 5fccf8b517763992a655b8378e1771967252f0d7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5fccf8b517763992a655b8378e1771967252f0d7)

PR #3623, opened 2017-09-28
Bookmarking topstories

piatra:
closes Bug #3131 and fix #3573 and fix #3629

PR #3675, opened 2017-10-10
Dedupe sections in the reducer

### LINKED ISSUE #3573: Inconsistent state between sections/SectionManager
When working on #3131 I came across an issue where the SectionManager has state that is inconsistent with the section it manages. Filling this issue mostly for visibility otherwise it gets easily lost in irc/slack.

Highlights do a refresh [on every page load](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/HighlightsFeed.jsm#L152)
Topstories [only update on system tick](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/TopStoriesFeed.jsm#L261)
Combined with also having deduping between the sections this leads to an issue where a card "*disappears*", after deduping, from the state that [SectionManager holds in this map](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/SectionsManager.jsm#L68). 
For the purpose of #3131 this is fixed by doing a `SectionManager.updateSection` call, on page load, from `TopstoriesFeed` and update the state in `SectionManager`.

I was not involved in this part of the code so I'm interested to hear why we duplicate state in the `SectionManager`. Given the large number of events happening in the app it's easy to overlook how every piece interacts and it can lead to inconsistencies (obvious now but it took me a long time to debug). 

(I might not understand all the implications so this might not be valid) My proposal would be to have a standard field in every section that holds state. For example instead of `TopStoriesFeed.stories` -> `TopStoriesFeed.data` and the SectionManager could access that field instead of keeping a local copy (this would make it so that the bug I encountered would not happen).

cc @csadilek @k88hudson

### LINKED ISSUE #3587: Removing a bookmarked Highlight comes back as Bookmarked but it's not actually
Looks like the links cache is too aggressive in its caching. Probably need to expire the cache on `case at.PLACES_BOOKMARK_REMOVED:` although the current logic to expire is when `this.fetchHighlights(true);` So probably just need to be smarter all around on when to expire the cache or not. Or when to broadcast or not.

### LINKED ISSUE #3629: Bookmarked status for Highlights is only shown after you click the context menu a second time
https://bugzilla.mozilla.org/show_bug.cgi?id=1404261

[Affected versions]:
- Firefox 58.0a1 Build ID 20170928220658

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have enough history so that Highlights are displayed.

[Steps to reproduce]:
2. Click the "..." for one of the Highlights.
3. Click the "Bookmark" option and observe the bottom left corner of the card.

[Expected results]:
- The status of the card is changed to "Bookmarked"

[Actual results]:
- The status of the card remains unchanged.

[Regression window]:
Last good revision: d7ccda988869be632fb6b3345d3efba111db7f7d
First bad revision: 3b514f20525296a9b6a623ba680fb027bc9e6cae
Pushlog:
https://hg.mozilla.org/integration/autoland/pushloghtml?fromchange=d7ccda988869be632fb6b3345d3efba111db7f7d&tochange=3b514f20525296a9b6a623ba680fb027bc9e6cae

[Notes]:
- Dismissing a different Highlight will refresh the affected card and it will correctly show its status.
- The status is correctly displayed after the page is refreshed.
bookmark state for highlights not displayed instantly

piatra:
fix #3573 and fix #3629 and fix #3587. Re submitting https://github.com/mozilla/activity-stream/pull/3623 without highlights date filtering that was based on `ProfileAge` module.

I did a bigger try push https://treeherder.mozilla.org/#/jobs?repo=try&revision=ceb8c64e3ee8035a463a778bb3561ebfa5cb101a 
Some recurrent failures:
 * devtools/shared/webconsole/test/test_console_serviceworker.html -- consistently failing but also on latest master so disregarding.
 * Unused file whitelist item. file: chrome://global/skin/arrow/panelarrow-vertical@2x.png -- referenced files?
* `SpecialPowers.getDOMWindowUtils(...).addToStyloBlocklist is not a function` -- but I don't see how we could cause this

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -4,12 +4,15 @@
 "use strict";
 
 const {actionTypes: at} = Components.utils.import("resource://activity-stream/common/Actions.jsm", {});
+const {Dedupe} = Components.utils.import("resource://activity-stream/common/Dedupe.jsm", {});
 
 // Locales that should be displayed RTL
 const RTL_LIST = ["ar", "he", "fa", "ur"];
 
 const TOP_SITES_DEFAULT_LENGTH = 6;
 const TOP_SITES_SHOWMORE_LENGTH = 12;
+
+const dedupe = new Dedupe(site => site && site.url);
 
 const INITIAL_STATE = {
   App: {
@@ -230,7 +233,7 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
       }
       return newState;
     case at.SECTION_UPDATE:
-      return prevState.map(section => {
+      newState = prevState.map(section => {
         if (section && section.id === action.data.id) {
           // If the action is updating rows, we should consider initialized to be true.
           // This can be overridden if initialized is defined in the action.data
@@ -239,6 +242,28 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
         }
         return section;
       });
+
+      if (!action.data.dedupeConfigurations) {
+        return newState;
+      }
+
+      action.data.dedupeConfigurations.forEach(dedupeConf => {
+        newState = newState.map(section => {
+          if (section.id === dedupeConf.id) {
+            const dedupedRows = dedupeConf.dedupeFrom.reduce((rows, dedupeSectionId) => {
+              const dedupeSection = newState.find(s => s.id === dedupeSectionId);
+              const [, newRows] = dedupe.group(dedupeSection.rows, rows);
+              return newRows;
+            }, section.rows);
+
+            return Object.assign({}, section, {rows: dedupedRows});
+          }
+
+          return section;
+        });
+      });
+
+      return newState;
     case at.SECTION_UPDATE_CARD:
       return prevState.map(section => {
         if (section && section.id === action.data.id && section.rows) {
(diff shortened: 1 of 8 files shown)


piatra:
I think most of the failures were caused by the highlights query being slowed down by the call to `ProfileAge`, because these messages showed up in the try run:
```
 browser/base/content/test/urlbar/browser_urlbarAboutHomeLoading.js | This test exceeded the timeout threshold. It should be rewritten or split up. If that's not possible, use requestLongerTimeout(N), but only as a last resort. 
```
[bug 1406522](https://bugzilla.mozilla.org/show_bug.cgi?id=1406522) seems to be the only one created directly due to this PR.

As an alternative we could filter the links returned by the highlights query only when it actually returns something. On a fresh profile start, like most of the tests do, the call isn't necessary. Could this be an alternative?

piatra:
@Mardak can you have a look at this so we can land it? I've decided to [keep the call that updates the metadata for bookmarks](https://github.com/mozilla/activity-stream/pull/3675/files#diff-46c4b10ce28f1463cea556e4020e95f4R154) (with Pocket info). It won't work unless you actually visit the link but it ensures we have identical metadata.

All of the previous try failures don't seem to reproduce once the `ProfileAge` call has been removed. 

I've made [a new try push](https://treeherder.mozilla.org/#/jobs?repo=try&revision=58661421fd87654a8aa800fd6f7b57fa3b2607ce) to ensure the there is no issue with the referenced file.

I also went ahead and tested again all the bugs referenced and they are indeed fixed by this patch.

Mardak:
> It won't work unless you actually visit the link but it ensures we have identical metadata.

Visiting the page will cause the metadata to change, but that should be fine.

as-pine-proxy:
[Treeherder: f1df68438013d3e336a8042f8fb32023a605213d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f1df68438013d3e336a8042f8fb32023a605213d)

PR #3693, opened 2017-10-11
feat(styles): Show placeholders for better perceived performance

### LINKED ISSUE #3640: Fade in content sooner when webrender is on by default
As part of the fix for #3568, we ended up delaying the fade to avoid competing with tab animation, which would have resulted in a tart regression. This directly impacts our total time to done painting as we show content later exactly by the delayed amount.

https://bugzilla.mozilla.org/show_bug.cgi?id=1401682#c4

It was noted in the bug comment above that with QuantumRender / webrender, there was no performance impact with the earlier fades, so we can avoid the delay and regression when webrender is turned on.

https://bugzilla.mozilla.org/show_bug.cgi?id=webrender

### LINKED ISSUE #3692: Hide just the section title and a section and show without fade
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1383593

Instead of hiding everything under search, show placeholders without some stuff

Mardak:
Fix #3692 and fix #3640. r?@k88hudson Just backporting https://hg.mozilla.org/integration/autoland/rev/310eff6bb091 and updating comments.

### DIFF
--- a/system-addon/content-src/components/Base/_Base.scss
+++ b/system-addon/content-src/components/Base/_Base.scss
@@ -53,18 +53,20 @@ main {
 }
 
 .body-wrapper {
-  opacity: 0;
-  // The fade duration and delay below were specifically picked to avoid talos
-  // regressing tart, which measures tab opening animation performance. Tabs
-  // currently have a 100ms width transition that competes with this transition.
-  // When webrender/Quantum turns on, overlapping transitions are no problem.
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: b8c5f0f432609b7c1ff198eb1153834046e9f2f6](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b8c5f0f432609b7c1ff198eb1153834046e9f2f6)

Issue #3676, opened 2017-10-10
"Top Sites" disappear or moved to different position after they are unpinned


Softvision-MariusComan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1407172

**[Notes]:**
- The issue is not reproducible if you import history from another browser.

**[Affected versions]:**
- Firefox 57 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- On a new profile sign in to sync with an account that has enough history and bookmarks to populate the "Top Sites" area.

**[Steps to reproduce]:**
1. Open the browser with the profiles from prerequisites.
2. Pin a Top Site and open it in multiple new tabs.
3. Restart the browser.
4. Unpin the Top Site pinned at step 2 and observe its behavior.

**[Expected result]:**
- The Top Site stays in the same place.

**[Actual result]:**
- The Top Site disappears or it is moved towards the back in the "Top Sites" area.

**[Additional Notes]:**
- Screen recording of the issue: https://goo.gl/AV7fsQ.

piatra:
When we pin/unpin a Topsite we query for Topsites and they might have changed their frecency score (doesn't explain why in this example the topsite disappeared).

aaronrbenson:
I honestly think this looks okay .. if I unpin a top site it's a pretty strong signal that I don't want it in my list anymore. Besides, there's an edit function to add it back if I want to.

PR #3695, opened 2017-10-12
Reduce duplicates places queries

### LINKED ISSUE #3638: Reduce duplicates places queries near startup

piatra:
Closes #3638

* Removed `fetchHiglights` call when there are less than 9 highlights cards
* added an flag in Topsites and Highlights to signal that data is currently being fetched 
  * this will prevent `INIT` and `NEW_TAB_LOAD` actions from triggering two update calls
    * covers the case when the events are fast enough that they overlap
    * otherwise `lastUpdated` field will prevent requests from happening
* on `TOP_SITES_UPDATED` check that Topsites and Highlights have items in common before triggering `fetchHighlights`
  * `fetchHiglights` ensures deduping but we can skip it if there is nothing to dedupe


Mardak:
This seems to revert the fix for #3377 from #3386

piatra:
And probably something similar can happen for Highlights as well? 
I'll do some more testing.

piatra:
@Mardak I’ve simulated a delayed TopSitesFeed startup by doing `await new Promise(resolve => setTimeout(resolve, 3000));` in the refresh method. On startup I also open several other new tabs. They are all missing TopSites initially but as soon as the first one is initialized they all get the data.
Does this flow match the one in the issue? I'm trying to reproduce it so that maybe we can fix it some other way that doesn't involve multiple calls to `refresh`.

Mardak:
That's how things behaved before #3386. I just tracked the actual cause: NEW_TAB_LOAD happens before INIT, so the broadcast doesn't happen because that one gets skipped for a pending refresh from LOAD. Filed #3697

Mardak:
@k88hudson I wonder if this is the right place/time to just get rid of refreshing sites/highlights on NEW_TAB_LOAD especially if both are refreshing based on last updated time. Probably do it on  SYSTEM_TICK?

In terms of avoiding double Highlights places call on INIT and TOP_SITES_UPDATED, instead of awaiting for top sites from INIT, maybe just return early and remember that the next TOP_SITES_UPDATED should broadcast. Although if TopSitesFeed always broadcasts… should Highlights getting TOP_SITES_UPDATED also broadcast?

Mardak:
Actually nevermind the last question. I would think SYSTEM_TICK does /not/ broadcast and just updates the store for the next rehydration. So similarly highlights shouldn't broadcast on TOP_SITES_UPDATED unless a previous broadcast was skipped because there was no top sites data.

piatra:
> I wonder if this is the right place/time to just get rid of refreshing sites/highlights on NEW_TAB_LOAD especially if both are refreshing based on last updated time. Probably do it on SYSTEM_TICK?

This would fix the extra calls on startup for both `TopSites` and `Highlights`.

> In terms of avoiding double Highlights places call on INIT and TOP_SITES_UPDATED, instead of awaiting for top sites from INIT, maybe just return early and remember that the next TOP_SITES_UPDATED should broadcast.

Highlights should return early if Topsites are not ready and broadcast if its store state is not initialized (this should only happen for initialization).

Mardak:
Oh, checking for Highlights's section initialized value should work. I just realized Highlights probably shouldn't refresh on any actions that TopSitesFeed is also watching; otherwise, we end up with the double updates as we do now when TOP_SITES_UPDATED comes in. Although I can't think of a good way to maintain that dependency right now…

Or maybe something like:
```js
constructor() {
  this.broadcastNextUpdate = true;
}
onAction() {
  case MIGRATION:
  case PLACES_*:
    this.broadcastNextUpdate = true;
    break;
  case TOP_SITES_UPDATED:
    this.fetchHighlights();
}
```

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -54,6 +55,12 @@ this.HighlightsFeed = class HighlightsFeed {
   }
 
   async fetchHighlights(broadcast = false) {
+    if (this._updateInProgress) {
```
k88hudson: The only issue I've seen with this approach is if an update is in progress where an actual change happened (e.g. something was edited, blocked) the old data is returned instead of the more recent data

piatra: @k88hudson  We could add `&& !broadcast` if that's the only issue. What about removing refresh entirely from new tab load and defer to SYSTEM_TICK?

piatra:
```

```

piatra:
> I just realized Highlights probably shouldn't refresh on any actions that TopSitesFeed is also watching; otherwise, we end up with the double updates as we do now when TOP_SITES_UPDATED comes in. 

That's what I was trying to do [with this check, before calling `fetchHiglights` on TOP_SITES_UPDATED](https://github.com/mozilla/activity-stream/pull/3695/files#diff-63be189f2ac8e49f56542504cffa9829R139)

on TOP_SITES_UPDATED, HighlightsFeed should only call fetchHighlights if we know some items will get deduped (removed).
The only time we need to bypass that is when `Highlights.rows.length === 0` because we returned early when Topsites weren't initialized.

There is the case where you unpin something and it falls out of your Topsites (which shouldn't even happen https://github.com/mozilla/activity-stream/issues/3676) case in which maybe a Highlight should show up. But having that show up at the next refresh seems totally resonable (to simplify the logic).

Mardak:
One drawback of `highlightsWillDedupe` approach is that it conditionally fetches twice. In a state where that method returns `true`, e.g., a top site was bookmarked (now a bug with the implementation in the PR because bookmarks don't dedupe), every TOP_SITES_UPDATED will do a 2nd fetch for the next 5 days. So in the best case for a given action, there's one fetch and worst case there's two.

So a different approach that could make it in the worst case 1 fetch would perhaps be better.

piatra:
Instead of checking TopSites.rows and Highlights.rows for matching URLs I could directly call dedupe and see if the length of the result is different.
It should definitely be better than invalidating the cache & doing a new query.

Mardak:
One thing I just noticed from my own browsing is that because we currently broadcast fetchHighlights on a NEW_TAB_LOAD, often times I end up seeing highlights without images because it only just realized there are new highlights and should fetch an image. But by doing it some time other than TAB_LOAD, there's a higher chance that the image will already be fetched.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -141,13 +152,10 @@ this.HighlightsFeed = class HighlightsFeed {
       case at.INIT:
         this.init();
         break;
-      case at.NEW_TAB_LOAD:
-        if (this.highlightsLength < HIGHLIGHTS_MAX_LENGTH) {
-          // If we haven't filled the highlights grid yet, fetch again.
-          this.fetchHighlights(true);
-        } else if (Date.now() - this.highlightsLastUpdated >= HIGHLIGHTS_UPDATE_TIME) {
+      case at.SYSTEM_TICK:
+        if (Date.now() - this.highlightsLastUpdated >= HIGHLIGHTS_UPDATE_TIME) {
```
Mardak: There's 3 different "timers" that are being considered here (and for top sites):
1) system tick interval (5 minutes)
2) feed last updated (15 minutes)
3) links cache expiration (5 minutes)

Does it make sense to still have all 3? Or maybe a related question is what's each of their purpose now?

Another issue with similar but slightly different timers is that if they're a little bit off, an expected 5 minute refresh gets skipped and is only actually refreshed after 10 minutes. E.g., due to async stuff, last update was at T+5.01 and the next system tick at T+10.0 is only 4.99 minutes after the last update.

system-addon/lib/ActivityStreamMessageChannel.jsm
```diff
@@ -138,7 +138,9 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
     this.channel.addMessageListener("RemotePage:Load", this.onNewTabLoad);
     this.channel.addMessageListener("RemotePage:Unload", this.onNewTabUnload);
     this.channel.addMessageListener(this.incomingMessageName, this.onMessage);
+  }
 
+  dispatchInitLoadEvent() {
```
Mardak: Nice that this separated so easily. I would name this something else as there's already too many "dispatch" and "init"s. Probably something along the lines of `simulateNewTabEvents`.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -125,6 +120,24 @@ this.HighlightsFeed = class HighlightsFeed {
...
   /**
+   * Compare TopSites and Highlights for urls that might exist in both sections.
+   * If true we need to dedupe Highlights.
+   * @param links new TopSites state
+   */
+  _highlightsWillDedupe(links) {
+    const highlightsIndex = SectionsManager.sections.get(SECTION_ID).order;
+    const rows = this.store.getState().Sections[highlightsIndex].rows;
+    // No data available, we need to fetch and dedupe results.
+    if (!rows.length) {
```
Mardak: This special case seems to do the exact opposite of what the method says it should be doing. If there's no highlights, there's actually nothing to be deduped, i.e., highlights will *not* be deduped.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -234,13 +234,9 @@ this.TopSitesFeed = class TopSitesFeed {
...
-      case at.NEW_TAB_LOAD:
-        if (
-          // When a new tab is opened, if the last time we refreshed the data
-          // is greater than 15 minutes, refresh the data.
-          (Date.now() - this.lastUpdated >= UPDATE_TIME)
-        ) {
-          this.refresh(action.meta.fromTarget);
+      case at.SYSTEM_TICK:
+        if (Date.now() - this.lastUpdated >= UPDATE_TIME) {
+          this.refresh();
```
Mardak: I believe @k88hudson has mentioned that we want to avoid "random" updates to already opened pages. Looks like we probably want some way to refresh without dispatching to content now and only update chrome… maybe just `SendToMain`?

system-addon/test/unit/lib/ActivityStreamMessageChannel.test.js
```diff
@@ -80,32 +80,35 @@ describe("ActivityStreamMessageChannel", () => {
...
-        RPmessagePorts.push({
-          url: "about:monkeys",
-          loaded: false,
-          portID: "inited"
-        });
-        RPmessagePorts.push({
-          url: "about:sheep",
-          loaded: true,
-          portID: "loaded"
+      describe("#dispatchInitLoadEvent", () => {
```
Mardak: This should be at the same level as `#createChannel` and would avoid indenting everything.

system-addon/lib/HighlightsFeed.jsm
piatra: I think that HIGHLIGHTS_UPDATE_TIME is the only one not needed anymore. Looks like a work around to avoid frequent updates on every page load.

1. system tick interval is needed because this is the new event we react to, independent of page loads. This should also fix the problem of refreshing highlights before having the image ready [as described in comment](https://github.com/mozilla/activity-stream/pull/3695#issuecomment-336592534)

2. links cache expiration ensures we get new highlights regardless of special events (like BOOKMARKS_REMOVED/ADDED) that invalidate the cache or our different invalidation tactics (every LAST_FEED_UPDATED minutes).

I'm not sure the timing issue is very important (maybe I don't get it) but we force update on special events for blocking and bookmarking where we need to ensure consistent information. If we don't update the highlights with visits once in a while it should be ok.

system-addon/lib/HighlightsFeed.jsm
piatra: Problem is I kept adding logic to it as new edge cases emerged. Logic should be fine, maybe rename it to something like `shouldFetchHighlights` similar to `shouldUpdate` in React when it receives new props.

system-addon/lib/TopSitesFeed.jsm
piatra: @Mardak This goes for Highlights as well right?

system-addon/lib/HighlightsFeed.jsm
Mardak: The "don't update the highlights with visits once in a while" is actually "every other time" if the timing is just off. It's just being clear that the code might say "refresh every 5 minutes" but turns out to be "refresh every 10 minutes." That may be acceptable, but if 5 minutes was picked for a particular reason, it shouldn't just accidentally be twice as long.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -132,7 +132,7 @@ this.TopSitesFeed = class TopSitesFeed {
    * @param target Optional port/channel to receive the update. If not provided,
    *               the update will be broadcasted.
    */
-  async refresh(target = null) {
+  async refresh(target = null, dispatch = true) {
```
Mardak: Having multiple options here makes calling things clunky. Especially when passing in boolean args. It looks like we don't need to support `target` anymore. And the `dispatch` name is pretty bad name especially because `dispatch` is an existing method vs here it's a condition controlling logic to either `dispatch` when `true` or `dispatch` when `false`.

system-addon/lib/HighlightsFeed.jsm
piatra: Could we move 
```
```
At the top of the `fetchHighlights` method? I've tried it and it seems to consistently trigger updates at the right moment.

Mardak: I thought you were suggesting HIGHLIGHTS_UPDATE_TIME wouldn't be necessary, so there shouldn't be a need to track `highlightsLastUpdated` ? We could probably just set LinksCache update time to be 10% less than SYSTEM_TICK, i.e., 4.5 minutes, if we're sure we just want to update at least once every 5 minutes.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -55,23 +55,18 @@ this.HighlightsFeed = class HighlightsFeed {
   }
 
   async fetchHighlights(broadcast = false) {
+    // We need TopSites to have been initialised for deduping.
+    // We can return early because we will get another chance to initialize
+    // on TOP_SITES_UPDATED.
```
Mardak: nit: odd wrapping and probably can just be shorter while still clear with the code below it:
`// We need TopSites for deduping, so just wait for TOP_SITES_UPDATED`

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -141,13 +136,10 @@ this.HighlightsFeed = class HighlightsFeed {
...
-      case at.NEW_TAB_LOAD:
-        if (this.highlightsLength < HIGHLIGHTS_MAX_LENGTH) {
-          // If we haven't filled the highlights grid yet, fetch again.
-          this.fetchHighlights(true);
-        } else if (Date.now() - this.highlightsLastUpdated >= HIGHLIGHTS_UPDATE_TIME) {
+      case at.SYSTEM_TICK:
+        if (Date.now() - this.highlightsLastUpdated >= HIGHLIGHTS_UPDATE_TIME) {
           // If the last time we refreshed the data is greater than 15 minutes, fetch again.
-          this.fetchHighlights(false);
+          this.fetchHighlights(true);
```
Mardak: We don't want to broadcast on `SYSTEM_TICK` and don't want feed update time logic.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -162,7 +154,8 @@ this.HighlightsFeed = class HighlightsFeed {
         this.fetchHighlights(false);
         break;
       case at.TOP_SITES_UPDATED:
-        this.fetchHighlights(false);
+        // Only broadcast on the first TOP_SITES_UPDATED call.
+        this.fetchHighlights(false || this.highlightsLastUpdated === 0);
```
Mardak: The `false ||` is useless with the other condition. But looking at this closer, wouldn't we want to broadcast the first fetch independent of which action triggered as opposed to only for TOP_SITES_UPDATED?

system-addon/lib/Store.jsm
```diff
@@ -138,6 +138,8 @@ this.Store = class Store {
     if (initAction) {
       this.dispatch(initAction);
     }
+
+    this._messageChannel.simulateMessagesForExistingTabs();
```
Mardak: A comment for why this is here relative to the other code

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -128,23 +128,20 @@ this.TopSitesFeed = class TopSitesFeed {
 
   /**
    * Refresh the top sites data for content
-   *
-   * @param target Optional port/channel to receive the update. If not provided,
-   *               the update will be broadcasted.
    */
-  async refresh(target = null) {
+  async refresh(broadcast = true) {
```
Mardak: It's a bit confusing that this is default `true`, but `HighlightsFeed` has it default `false` although it looks like all `fetchHighlights` calls are explicitly passing in a value instead of relying on defaults. Maybe we should just remove the defaults from both places.

Mardak: I suppose if we're touching the callers, might as well clean up the API to not pass in a plain boolean arg, so something more like:
`refresh(false)` -> `refresh({broadcast: false})`
("refresh... not?" vs "refresh without broadcast")

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -234,13 +231,10 @@ this.TopSitesFeed = class TopSitesFeed {
...
-      case at.NEW_TAB_LOAD:
-        if (
-          // When a new tab is opened, if the last time we refreshed the data
-          // is greater than 15 minutes, refresh the data.
-          (Date.now() - this.lastUpdated >= UPDATE_TIME)
-        ) {
-          this.refresh(action.meta.fromTarget);
+      case at.SYSTEM_TICK:
+        if (Date.now() - this.lastUpdated >= UPDATE_TIME) {
+          // No update target, no broadcast.
```
Mardak: Like Highlights, remove the update time stuff. Also, comment is outdated and probably won't be necessary if we have `refresh({broadcast: false})`

Mardak:
One thing I noticed when testing the PR, but this probably should be its own issue/PR. #3675 added expiring the highlights links cache on BOOKMARK_ADDED. A new profile causes adds of several bookmarks causing the cache to be expired and fetched from places multiple times on startup.

system-addon/lib/HighlightsFeed.jsm
piatra: How about checking `this.store.getState().Sections[Highlights].initialized` [here](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/HighlightsFeed.jsm#L122) and dispatch when `false`.
Maybe wrap everything in a `if highlights.length > 0` as highlights can be empty for fresh profiles.

Mardak: That seems reasonable to switch to broadcasting if not previously `initialized`. Also noticed that `highlightsLength` should be unused now that we don't force fetch if fewer than 9 highlights, so that can be removed.

Not entirely sure what you're suggesting for checking `length > 0`? Don't call `updateSection` if there's nothing?

piatra: Yes, don't broadcast if there's nothing. But then it will never show the empty state as that is `initialized && !rows.length` so nvm.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -47,29 +44,22 @@ this.HighlightsFeed = class HighlightsFeed {
...
 
   uninit() {
     SectionsManager.disableSection(SECTION_ID);
   }
 
-  async fetchHighlights(broadcast = false) {
-    // We broadcast when we want to force an update, so get fresh links
-    if (broadcast) {
-      this.linksCache.expire();
+  async fetchHighlights(options = {}) {
```
Mardak: Let's describe the option we expect and for TopSites too.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -119,9 +109,12 @@ this.HighlightsFeed = class HighlightsFeed {
       }
     }
 
-    SectionsManager.updateSection(SECTION_ID, {rows: highlights}, broadcast);
-    this.highlightsLastUpdated = Date.now();
-    this.highlightsLength = highlights.length;
+    const sectionIndex = SectionsManager.sections.get(SECTION_ID).order;
+    const initialized = this.store.getState().Sections[sectionIndex].initialized;
```
Mardak: nit: `const {initialized} = …`

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -119,9 +109,12 @@ this.HighlightsFeed = class HighlightsFeed {
...
+    // Broadcast when required or if it is the first fetch.
```
Mardak: nit: "first fetch" -> "first update" as INIT is most likely the first fetch.

system-addon/lib/LinksCache.jsm
```diff
@@ -5,7 +5,10 @@
 
 this.EXPORTED_SYMBOLS = ["LinksCache"];
 
-const EXPIRATION_TIME = 5 * 60 * 1000; // 5 minutes
+// This should be slightly less than SYSTEM_TICK_INTERVAL as timers are
+// not precise enough causing skips in the refresh interval.
```
Mardak: nit: It's because timer comparisons are *too* exact ;) and the underlying issue is that async/await functionality will make the last recorded times a little bit later. Might be useful to include the 10%, e.g., "slightly less (10%)"

system-addon/test/unit/lib/HighlightsFeed.test.js
```diff
@@ -274,78 +290,69 @@ describe("Highlights Feed", () => {
...
-      await feed.fetchHighlights();
-      feed.fetchHighlights = sinon.spy();
+    it("should expire the cache on PLACES_BOOKMARK_REMOVED", async () => {
+      sandbox.stub(feed.linksCache, "expire");
+
+      feed.onAction({type: at.PLACES_BOOKMARK_REMOVED});
+
+      assert.calledOnce(feed.linksCache.expire);
+    });
+    it("should not call fetchHighlights with broadcast on initialization", () => {
```
Mardak: nit: this doesn't seem to describe the test

system-addon/test/unit/lib/TopSitesFeed.test.js
```diff
@@ -470,26 +471,20 @@ describe("Top Sites Feed", () => {
...
+      assert.calledOnce(feed.refresh);
+      assert.calledWithExactly(feed.refresh, {broadcast: false});
     });
-    it("should not call refresh if .lastUpdated is less than update time on NEW_TAB_LOAD", () => {
-      feed.lastUpdated = 0;
-      clock.tick(UPDATE_TIME - 1);
-      sinon.stub(feed, "refresh");
-      feed.onAction(newTabAction);
-      assert.notCalled(feed.refresh);
+    it("should call refresh on SYSTEM_TICK with correct params", () => {
```
Mardak: This seems to do the exact same thing as the previous test.

as-pine-proxy:
[Treeherder: dfab603038f706e8f8945d1a3c9c8b0a9fc2f6a8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=dfab603038f706e8f8945d1a3c9c8b0a9fc2f6a8)

PR #3700, opened 2017-10-12
fix(styles): Use css to calculate rtl offset instead of hardcoded of expected width

### LINKED ISSUE #3699: [RTL] The "Show two rows" icon on the New Tab Preferences overlaps the checkbox if a scrollbar is visible
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1408165

image

Mardak:
Fix #3699. r?@rlr Couldn't reproduce on mac, so below are screenshots from linux.

Before:
screen shot 2017-10-12 at 2 09 40 pm

After:
screen shot 2017-10-12 at 2 09 25 pm

### DIFF
--- a/system-addon/content-src/components/PreferencesPane/_PreferencesPane.scss
+++ b/system-addon/content-src/components/PreferencesPane/_PreferencesPane.scss
@@ -85,7 +85,8 @@
       }
 
       label {
-        background-position-x: 35px;
+        $icon-offset-start: 35px;
+        background-position-x: $icon-offset-start;
         background-position-y: 2.5px;
         background-repeat: no-repeat;
         display: inline-block;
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: b3db2ad50dbc18bfd081f9bbf80533d0aa9a989f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b3db2ad50dbc18bfd081f9bbf80533d0aa9a989f)

Mardak:
> I guess CSS is missing a background positioning from right

Oh. Apparently "Two-value syntax (support for offsets from any edge)" added in Firefox 49. https://developer.mozilla.org/en-US/docs/Web/CSS/background-position-x

PR #3703, opened 2017-10-13
chore(styles): Switch background-position calc to 2-value syntax

Mardak:
Followup to #3700. r?@rlr I guess I looked through https://developer.mozilla.org/en-US/docs/Web/CSS/background-position too quickly. Too many possible values! ;)


as-pine-proxy:
[Treeherder: 36021c3aba9023714f957867befbe1ca1d7684ef](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=36021c3aba9023714f957867befbe1ca1d7684ef)

PR #3705, opened 2017-10-13
Update activity stream to ESLint 4

Standard8:
mozilla-central is starting to get closer to upgrading to ESLint 4. ESLint 4 is generally detecting more issues that 3.x was.

Hence, I'd like to get activity stream updated a little earlier, so that we can get required fixes merged back across before we hit that point.

These two commits upgrade eslint, eslint-plugin-mozilla, eslint-plugin-promise, eslint-plugin-react and eslint-watch.

Note that padded-blocks now conflicts with lines-around-comment in some cases, so I've turned that rule off. You're welcome to try and find a better solution.


Mardak:
I've had an eslint PR to fix https://github.com/eslint/eslint/issues/8564 but I guess just turning off the rule will work.

.eslintrc.js
```diff
@@ -202,6 +204,12 @@ module.exports = {
     "semi-spacing": [2, {"before": false, "after": true}],
     "sort-imports": 2,
     "sort-vars": 2,
+    // Override this until eslint-plugin-mozilla gets updated.
+    "space-before-function-paren": ["error", {
+      "anonymous": "never",
+      "asyncArrow": "always",
```
Mardak: What will happen when `eslint-plugin-mozilla` gets updated? Looks like it will be "never" for everything, so maybe we should just match and autofix? https://searchfox.org/mozilla-central/source/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#366-367

system-addon/content-src/components/ConfirmDialog/ConfirmDialog.jsx
```diff
@@ -22,6 +22,8 @@ const {actionTypes, actionCreators: ac} = require("common/Actions.jsm");
  *   confirm_button_string_id: "menu_action_delete"
  * },
  */
+// Not an issue until React is updated.
+// eslint-disable-next-line react/no-deprecated
 const ConfirmDialog = React.createClass({
```
Mardak: I just filed #3706 but maybe we should just fix it here? ;)

Mardak: I'll fix it in #3713.

package.json
```diff
@@ -41,12 +41,12 @@
     "cpx": "1.5.0",
     "decompress-cli": "2.0.0",
     "enzyme": "2.7.0",
-    "eslint": "3.9.1",
+    "eslint": "4.8.0",
```
Mardak: Oh hey, `4.9.0` is available with my fix. https://eslint.org/blog/2017/10/eslint-v4.9.0-released

.eslintrc.js
```diff
@@ -187,7 +188,8 @@ module.exports = {
     "one-var-declaration-per-line": [2, "initializations"],
     "operator-assignment": [2, "always"],
     "operator-linebreak": [2, "after"],
-    "padded-blocks": [2, "never"],
+    // padded-blocks conflicts with lines-around-comment.
+    "padded-blocks": "off",
```
Mardak: With `4.9.0` instead of turning this off, we can do this instead:
`"lines-around-comment": [2, {"allowClassStart": true, "allowObjectStart": true, "beforeBlockComment": true}],`

.eslintrc.js
Standard8: ESLint 4 tweaked the async arrow option - it was ignored by default in v3. The v4 changes haven't landed in m-c yet, although I guess I could start landing parts of them soon.

Standard8:
@Mardak Thanks for the comments, all updated.

as-pine-proxy:
[Treeherder: 7606b86901b9f100bc2f4f07b60fe08b21633085](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7606b86901b9f100bc2f4f07b60fe08b21633085)

PR #3710, opened 2017-10-13
Frequency caps support for spoc experiment

csadilek:
This PR brings in frequency caps for the sponsored content experiment:

Frequency caps are based on campaigns (not individual spocs). We currently support two types of frequency caps:
  - lifetime: Indicates how many times spocs from a campaign can be shown in total
  - period: Indicates how many times spocs from a campaign can be shown within a period
  
So, for example, the feed configuration below defines that for campaign 1 no more than 5 spocs can be show in total, and no more than 2 per hour.
```
"campaign_id": 1,
  "caps": {
    "lifetime": 5,
    "campaign": {
      "count": 2,
      "period": 3600
    }
  }
```

This is implemented storing a mapping of campaign IDs to timestamps (impressions) in a pref. The pref is cleaned up every time the feed is fetched. Only campaigns part of the last response are considered active, all others can be removed.

The second commit contains a minor refactoring to guard against missing domain affinities in the server response.

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -23,13 +23,17 @@ const TOPICS_UPDATE_TIME = 3 * 60 * 60 * 1000; // 3 hours
 const DOMAIN_AFFINITY_UPDATE_TIME = 24 * 60 * 60 * 1000; // 24 hours
 const STORIES_NOW_THRESHOLD = 24 * 60 * 60 * 1000; // 24 hours
 const SECTION_ID = "topstories";
+const SPOC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.spoc.impressions";
+const MAX_LIFETIME_CAP = 100; // Guard against misconfiguration on the server
 
 this.TopStoriesFeed = class TopStoriesFeed {
   constructor() {
     this.spocsPerNewTabs = 0;
     this.newTabsSinceSpoc = 0;
+    this.spocCampaignMap = new Map();
     this.contentUpdateQueue = [];
     this.cache = new PersistentCache(SECTION_ID, true);
+    this._prefs = new Prefs();
   }
 
   init() {
@@ -76,7 +80,12 @@ this.TopStoriesFeed = class TopStoriesFeed {
       const body = await response.json();
       this.updateSettings(body.settings);
       this.stories = this.rotate(this.transform(body.recommendations));
-      this.spocs = this.show_spocs && this.transform(body.spocs).filter(s => s.score >= s.min_score);
+
+      if (this.show_spocs && body.spocs) {
+        this.spocCampaignMap = new Map(body.spocs.map(s => [s.id, `${s.campaign_id}`]));
+        this.spocs = this.transform(body.spocs).filter(s => s.score >= s.min_score);
+        this.cleanUpCampaignImpressionPref();
+      }
 
       this.dispatchUpdateEvent(this.storiesLastUpdated, {rows: this.stories});
       body._timestamp = this.storiesLastUpdated = Date.now();
@@ -124,7 +133,8 @@ this.TopStoriesFeed = class TopStoriesFeed {
         "referrer": this.stories_referrer,
         "url": s.url,
         "min_score": s.min_score || 0,
-        "score": this.personalized ? this.affinityProvider.calculateItemRelevanceScore(s) : 1
+        "score": this.personalized ? this.affinityProvider.calculateItemRelevanceScore(s) : 1,
+        "spoc_meta": this.show_spocs ? {campaign_id: s.campaign_id, caps: s.caps} : {}
       }))
       .sort(this.personalized ? this.compareScore : (a, b) => 0);
   }
@@ -218,7 +228,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
       return apiKeyPref;
     }
 
-    return new Prefs().get(apiKeyPref) || Services.prefs.getCharPref(apiKeyPref);
+    return this._prefs.get(apiKeyPref) || Services.prefs.getCharPref(apiKeyPref);
   }
 
   produceFinalEndpointUrl(url, apiKey) {
@@ -240,8 +250,12 @@ this.TopStoriesFeed = class TopStoriesFeed {
     return url;
   }
 
+  shouldShowSpocs() {
+    return this.show_spocs && this.store.getState().Prefs.values.showSponsored;
+  }
+
   maybeAddSpoc(target) {
-    if (!this.show_spocs || !this.store.getState().Prefs.values.showSponsored) {
+    if (!this.shouldShowSpocs()) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


ncloudioj:
@csadilek I don't think I can finish the review today, will take a closer look at it next monday :)

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -277,6 +297,83 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+  isBelowFrequencyCap(impressions, spoc) {
+    const campaignImpressions = impressions[spoc.spoc_meta.campaign_id];
+    if (!campaignImpressions) {
+      return true;
+    }
+
+    const lifeTimeCap = spoc.spoc_meta.caps && spoc.spoc_meta.caps.lifetime;
+    const campaignCap = (spoc.spoc_meta.caps && spoc.spoc_meta.caps.campaign) || {};
+
+    const lifeTimeCapExceeded = () => campaignImpressions.length >= lifeTimeCap;
```
ncloudioj: Any reason of creating two functions here, looks like it just needs two booleans.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -277,6 +297,83 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+    const campaignCapExceeded = () =>
+      campaignImpressions.filter(i => (Date.now() - i) < (campaignCap.period * 1000)).length >= campaignCap.count;
+
+    return !lifeTimeCapExceeded() && !campaignCapExceeded();
+  }
+
+  updateActiveCampaigns(spocs) {
```
ncloudioj: nit: this function actually creates the `ActiveCampaings` instead of updating it.

system-addon/lib/TopStoriesFeed.jsm
csadilek: Yes, just to get lazy evaluation. This way we don't have to calculate campaignCapExceeded if lifeTimeCapExceeded is already true.

csadilek: Of course, I could inline it instead of storing separately but found this to be easier to read.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -277,6 +297,83 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+    const timeStamps = impressions[campaignId] || [];
+    timeStamps.push(Date.now());
+    impressions = Object.assign(impressions, {[campaignId]: timeStamps});
+
+    this.writeCampaignImpressionsPref(impressions);
+  }
+
+  readCampaignImpressionsPref() {
+    const prefVal = this._prefs.get(SPOC_IMPRESSION_TRACKING_PREF);
+    return (prefVal) ? JSON.parse(prefVal) : {};
```
ncloudioj: nit: redundant parentheses in `(prefVal)`

system-addon/lib/TopStoriesFeed.jsm
csadilek: Fixed ;)

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -277,6 +297,83 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+
+    const lifeTimeCapExceeded = () => campaignImpressions.length >= lifeTimeCap;
+    const campaignCapExceeded = () =>
+      campaignImpressions.filter(i => (Date.now() - i) < (campaignCap.period * 1000)).length >= campaignCap.count;
+
+    return !lifeTimeCapExceeded() && !campaignCapExceeded();
+  }
+
+  updateActiveCampaigns(spocs) {
+    const activeCampaigns = new Map();
```
ncloudioj: Reading through this patch again, `activeCampaigns` sounds a bit confusing to me, particularly, there is a `activeCampaignIds` below. It's actually a map from tile_id to campaign_id, perhaps rename it? e.g. `mapTileToCampaign`. Heck, naming stuff is really hard ;)

csadilek: OK, I will think about the names again and update later tonight.

system-addon/lib/TopStoriesFeed.jsm
ncloudioj: Let's just inline them for better readability.

csadilek: OK, it was an unnecessary optimization anyway. Fixing it up.

csadilek:
@ncloudioj refactored to address all issues and nits. Thanks again.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -240,8 +249,12 @@ this.TopStoriesFeed = class TopStoriesFeed {
     return url;
   }
 
+  showSpocs() {
```
ncloudioj: nit: let's rename it to `shouldShowSpocs` to avoid ambiguity

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -277,6 +298,79 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+  isBelowFrequencyCap(impressions, spoc) {
+    const campaignImpressions = impressions[spoc.spoc_meta.campaign_id];
+    if (!campaignImpressions) {
+      return true;
+    }
+
+    const lifeTimeCap = spoc.spoc_meta.caps && spoc.spoc_meta.caps.lifetime;
+    const campaignCap = (spoc.spoc_meta.caps && spoc.spoc_meta.caps.campaign) || {};
+
+    const lifeTimeCapExceeded = campaignImpressions.length >= lifeTimeCap;
```
ncloudioj: Being able to short circuit is a nice thing here, I think I've confused you in the previous comment :)

```
// Check the lifetime frequency cap and campaign frequency cap, respectively
return campaignImpressions.length < lifeTimeCap &&
       campaignImpressions.filter(i => (Date.now() - i) < (campaignCap.period * 1000)).length >= campaignCap.count;
```

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -307,11 +403,21 @@ this.TopStoriesFeed = class TopStoriesFeed {
           this.spocs = this.spocs.filter(s => s.url !== action.data.url);
         }
         break;
+      case at.TELEMETRY_IMPRESSION_STATS:
```
ncloudioj: This action not only handles the impressions, but also takes care of clicks&blocks. I guess we have to only pick up the impression ones here.

csadilek: Ah good catch! I checked for action.tiles, thinking that's only the case for views, but clicks etc. have tiles too, just an array of length 1 then.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -277,6 +298,81 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+  //  }
+  // }
+  isBelowFrequencyCap(impressions, spoc) {
+    const campaignImpressions = impressions[spoc.spoc_meta.campaign_id];
+    if (!campaignImpressions) {
+      return true;
+    }
+
+    const lifeTimeCap = spoc.spoc_meta.caps && spoc.spoc_meta.caps.lifetime;
+    const lifeTimeCapExceeded = campaignImpressions.length >= lifeTimeCap;
```
ncloudioj: Given that `lifeTimeCap` and `campaignCap` are both from 3rd party, we might want to have a builtin cap defined by us, which prevents the timestamp array from growing without bound. Just be safe :)

system-addon/lib/TopStoriesFeed.jsm
ncloudioj: Yep, the easy fix would be check if the length of `action.data.tiles` === TOPSTORIES_SIZE

csadilek: Had to actually filter the event "types" because users could dismiss all but two stories, then tiles would be an array of length 2...

ncloudioj:
:shipit:

as-pine-proxy:
[Treeherder: 97f8827e842459478d385c64500853c69f0c91d1](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=97f8827e842459478d385c64500853c69f0c91d1)

PR #3717, opened 2017-10-16
chore(prefs): Add showSnippets pref separate from feeds.snippets

k88hudson:
Backport of Bug 1409058 (waiting for that to land)

### DIFF
--- a/system-addon/content-src/lib/snippets.js
+++ b/system-addon/content-src/lib/snippets.js
@@ -334,6 +334,7 @@ function addSnippetsSubscriber(store) {
     // state.Snippets.initialized             Is the snippets data initialized?
     // snippets.initialized:                  Is SnippetsProvider currently initialised?
     if (state.Prefs.values["feeds.snippets"] &&
+      state.Prefs.values.showSnippets &&
       state.Snippets.initialized &&
      !snippets.initialized &&
      // Don't call init multiple times
@@ -342,7 +343,8 @@ function addSnippetsSubscriber(store) {
       initializing = true;
       await snippets.init({appData: state.Snippets});
       initializing = false;
-    } else if (state.Prefs.values["feeds.snippets"] === false && snippets.initialized) {
+    } else if ((state.Prefs.values["feeds.snippets"] === false ||
+      state.Prefs.values.showSnippets === false) && snippets.initialized) {
       snippets.uninit();
     }
   });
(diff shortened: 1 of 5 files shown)


system-addon/content-src/lib/snippets.js
```diff
@@ -342,7 +343,8 @@ function addSnippetsSubscriber(store) {
       initializing = true;
       await snippets.init({appData: state.Snippets});
       initializing = false;
-    } else if (state.Prefs.values["feeds.snippets"] === false && snippets.initialized) {
+    } else if ((state.Prefs.values["feeds.snippets"] === false ||
+      state.Prefs.values.showSnippets === false) && snippets.initialized) {
```
Mardak: I'm assuming the `=== false` is to skip the case where `state.Prefs.initialized` is `false`, so we can make that explicit instead of implied with the first `if` condition. Also, we actually are trying to do the opposite / `else` but it's hidden in the extra conditionals. So I think we can clean it up and be more explicit and hopefully avoid bugs if someone touched one condition but not the other:
```js
// Can't decide if we don't have prefs
if (!state.Prefs.initialized) {
  return;
}
// We should show snippets
if (state.Prefs.values["feeds.snippets"] && state.Prefs.values.showSnippets) {
  if (state.Snippets.initialized && !snippets.initialized && !initializing) {
    initializing = true; …
  }
} else {
  // Shouldn't show snippets
  if (snippets.initialized) {
    snippets.uninit();
  }
}
```

I suppose alternatively, it could be just as clear with a variable, so we can keep it more like what it's now:
```js
// Can't decide if we don't have prefs
if (!state.Prefs.initialized) {
  return;
}
const shouldShow = state.Prefs.values["feeds.snippets"] && state.Prefs.values.showSnippets;
if (shouldShow && …) {
} else if (!shouldShow && …) {
}
```

system-addon/lib/ActivityStream.jsm
```diff
@@ -97,6 +97,10 @@ const PREFS_CONFIG = new Map([
     title: "Show the Search bar",
     value: true
   }],
+  ["showSnippets", {
```
ncloudioj: @k88hudson We'll need to update the pref string [here](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/TelemetryFeed.jsm#L33) in the TelemetryFeed.jsm, so that it can set the correct value for `user_prefs` in the telemetry.

Issue #3698, opened 2017-10-12
Setting `toolkit.cosmeticAnimations.enabled;false` should disable the slide-in slide-out animation of preferences


Mardak:
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1408133

If just limited to the pref pane, that might just go away when preferences move to about:preferences.

gwarser:
Thank you ;)

I think disabling/enabling parts of page, should be also covered by this pref.

Mardak:
Which parts?

gwarser:
Search, Top Sites, Highlights and Snippets - all we can enable/disable from New Tab Preferences.

Mardak:
Those don't animate when enable/disabled from the preferences?

gwarser:
Now if I have only Top Sites enabled, and then enable Highlights, all content is move up, "animating" in 2-3 frames.

gwarser:
I don't see this on clean profile :/ But still, thumbnails are fading in slowly.

gwarser:
Maybe I don't see this on clean profile because there is only one row of Highlights? On my _dirty_ Nightly, I have three rows here, and Top Sites are moved all way to the top of the page.

Mardak:
What are you suggesting happen when 3 rows of Highlights should be shown when turning on the section?

gwarser:
I don't know how to do about movement. If there is no animation, only rendering artifact, then should be left as is.

About thumbnails - they should not fade-in - they should be shown in one step.

iinc:
Are you talking about the hamburger icon that opens a popup/context menu? I haven't found a way to disable the animation. Very annoying indeed!

PR #3723, opened 2017-10-17
fix (animations): #3707 only animate when page is visible

### LINKED ISSUE #3707: Collapsed sections are animated again when switching to a previous opened New Tab
https://bugzilla.mozilla.org/show_bug.cgi?id=1408407

[Affected versions]:
- Firefox 58.0a1 Build ID 20171012220111

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Start the browser with the profile from prerequisites and open two New Tabs.
2. Collapse the Top Sites section.
3. Switch to the other New Tab and observed the collapsed section.

[Expected results]:
- The section is collapsed.

[Actual results]:
- The section is collapses when the tab is focused.

[Notes]:
- The issue is not reproducible if the section is opened in one tab and then you switch to another new tab.
- The issue is also reproducible in a new opened tab, but you have to open a new tab right after collapsing a section. 
- The issue is also reproducible with the Highlights and Recommended by Pocket sections.
collapsed section is animated again when switching to a different new tab

rlr:
The issue was that animations/transitions in background tabs are aggressively throttled or paused. So once the tab is brought into view, the rest of the animation happens. To fix, I just disabled the transitions for non visible tabs.

Fixes #3707

### DIFF
--- a/system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
+++ b/system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
@@ -2,6 +2,9 @@ const React = require("react");
 const {actionCreators: ac, actionTypes: at} = require("common/Actions.jsm");
 const {injectIntl, FormattedMessage} = require("react-intl");
 
+const VISIBLE = "visible";
+const VISIBILITY_CHANGE_EVENT = "visibilitychange";
+
 function getFormattedMessage(message) {
   return typeof message === "string" ? <span>{message}</span> : <FormattedMessage {...message} />;
 }
@@ -93,9 +96,23 @@ class CollapsibleSection extends React.PureComponent {
     this.onInfoLeave = this.onInfoLeave.bind(this);
     this.onHeaderClick = this.onHeaderClick.bind(this);
     this.onTransitionEnd = this.onTransitionEnd.bind(this);
-    this.state = {isAnimating: false, infoActive: false};
+    this.enableOrDisableAnimation = this.enableOrDisableAnimation.bind(this);
+    this.state = {enableAnimation: true, isAnimating: false, infoActive: false};
   }
 
+  componentWillMount() {
+    this.props.document.addEventListener(VISIBILITY_CHANGE_EVENT, this.enableOrDisableAnimation);
+  }
+  componentWillUnmount() {
+    this.props.document.removeEventListener(VISIBILITY_CHANGE_EVENT, this.enableOrDisableAnimation);
+  }
+  enableOrDisableAnimation() {
+    // Only animate the collapse/expand for visible tabs.
+    const visible = this.props.document.visibilityState === VISIBLE;
+    if (this.state.enableAnimation !== visible) {
+      this.setState({enableAnimation: visible});
+    }
+  }
   _setInfoState(nextActive) {
     // Take a truthy value to conditionally change the infoActive state.
     const infoActive = !!nextActive;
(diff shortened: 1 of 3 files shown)


system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -155,7 +174,14 @@ class CollapsibleSection extends React.PureComponent {
   }
 }
 
-CollapsibleSection.defaultProps = {Prefs: {values: {}}};
+CollapsibleSection.defaultProps = {
+  document: global.document || {
+    addEventListener: () => {},
+    removeEventListener: () => {},
+    visibilityState: "prerender"
```
rlr: @sarracini fixed it with this. Seem reasonable?

rlr: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState

Mardak:
Might be a good time to refactor `visibilitychange` stuff. There will be 4 places doing the similar visibility checks with this change. Also, at least for prerendered pages, the value is "hidden" -- I don't think Firefox actually does "prerender" state ? ?

Mardak:
I haven't looked closely at how other places are using visibility, but for this fix, we could probably set some `body.shouldAnimate` class to turn off animations.. maybe? Could be then reused for #3698.

rlr:
@Mardak the other checks for `visibilityState` are a little different than this case. Those seem to only want to listen for a change if the page isn't visible initially, where in this case we are constantly toggling animations as visibility changes. And I'm not sure what we could do to refactor other than create a facade on top of the DOM API.  I don't think we want to add it to the redux store?

I'm not sure about `body.shouldAnimate` since `<body/>` is outside of what redux controls, but having the logic in `Base.jsx` might be a good idea if we want a switch to disable all animations on the page. We would have to wrap all the transition/animation styles in an extra class. Maybe put them all in a separate _animations.scss file too.

Mardak:
> I don't think we want to add it to the redux store?

@k88hudson any suggestions on how to share watching for "visibilitychange" and getting `document.visibilityState`. I suppose the DOM APIs are quite simple enough, so anything we do could just be unnecessary wrapper… I.e., whatever we end up with might just be almost the same as what we would be doing with the multiple `addEventListener`s.

Mardak:
I suppose the thing that's bothering me is we end up defining the same `const` 3 places ;)

`const VISIBILITY_CHANGE_EVENT = "visibilitychange";`

But I suppose that's not really that different from all 4 places just using "visibilitychange" inline or `something.shared.onVisibilityChange`

system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -93,7 +96,23 @@ class CollapsibleSection extends React.PureComponent {
     this.onInfoLeave = this.onInfoLeave.bind(this);
     this.onHeaderClick = this.onHeaderClick.bind(this);
     this.onTransitionEnd = this.onTransitionEnd.bind(this);
-    this.state = {isAnimating: false, infoActive: false};
+    this.enableOrDisableAnimation = this.enableOrDisableAnimation.bind(this);
+    this.state = {enableAnimation: false, isAnimating: false, infoActive: false};
+  }
+
+  componentWillMount() {
+    this.enableOrDisableAnimation();
```
sarracini: Do you need this initial call to ```enableOrDisableAnimation()```? I removed it and it seemed to work the same since the function is called anyways when you open a new tab because of the event listener firing

sarracini: I guess to be safe that all visible tabs start with ```animation-enabled``` you could set ```enableAnimation : true``` in the constructor

system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
sarracini: Yup this seems good!

system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
rlr: I *think* we need to call `enableOrDisableAnimation()` on mount in case it is initially a non-visilble tab that is loading. Or does that still trigger a "visibilitychange" event? I'll play with it a little.

rlr: Actually, you're right. It doesn't seem to be needed 👍

as-pine-proxy:
[Treeherder: b5750842888acad5d853e49176a5fed40c3364d2](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b5750842888acad5d853e49176a5fed40c3364d2)

Issue #3725, opened 2017-10-17
Problematic code in "init-store.js"


ncloudioj:
Came across this bug while implementing #3455 , looks like [adding properties to store](https://github.com/mozilla/activity-stream/blob/master/system-addon/content-src/lib/init-store.js#L93-L94) won't work as expected for redux middlewares. See more details [here](https://github.com/reactjs/redux/blob/b30e6ae4aa217766dbc8bb0d5b9eb87f6c5c8bbd/src/applyMiddleware.js#L19).

It didn't break the code due to the coincidence that 
```js
store._didRehydrate = false;
store._didRequestInitialState = false;
```
has the same effect as
```js
store._didRehydrate = undefined;
store._didRequestInitialState = undefined;
```

We should fix this to avoid further bugs in the future.

Mardak:
Following from @k88hudson's explanation https://github.com/mozilla/activity-stream/pull/3731#issuecomment-337637292

Here's the code that's causing the `createStore` to have a different store than what middlewares see:
https://github.com/reactjs/redux/blob/fdfd13fbc83f2843bd6420c48b9b39d5fa17113d/src/applyMiddleware.js#L21-L29
(I know that links to a different version from what we're using, but it has the same concept.)

`applyMiddleware` returns a copy of the store created (which is what ends up getting returned by `createStore`) whereas the middlewares use a shared `middlewareAPI` wrapper store that has store-like properties.

Maybe instead of attaching properties to the store object, the state should be updated somehow, and the middleware should be doing `store.getState()` to change its behavior?

E.g., initialState should have somewhere:
```js
{
  didRehydrate: false,
  earlyMessageQueue: []
  …
}
```

?

Mardak:
Moved to https://bugzilla.mozilla.org/show_bug.cgi?id=1425498

PR #3731, opened 2017-10-18
fix(reporting): Send visibility_event when ActivityStream is ready

### LINKED ISSUE #3455: Early visibility_event lost when ActivityStream isn't ready
`DetectUserSessionStart` has some custom `sendAsyncMessage` to send "visibility_event_rcvd_ts" and is run as the page is loading. Potentially we could be smarter to have a general queue for any early messages to be sent on `INIT` similar to what @k88hudson did for `NEW_TAB_STATE_REQUEST`:
https://github.com/mozilla/activity-stream/blob/c214278f4c07121c92a9dc52cb51d856cb2ba850/system-addon/content-src/lib/init-store.js#L64-L67

@dmose

ncloudioj:
This fixes #3455

### DIFF
--- a/system-addon/content-src/lib/detect-user-session-start.js
+++ b/system-addon/content-src/lib/detect-user-session-start.js
@@ -1,13 +1,13 @@
-const {actionTypes: at} = require("common/Actions.jsm");
+const {actionTypes: at, actionCreators: ac} = require("common/Actions.jsm");
 const {perfService: perfSvc} = require("common/PerfService.jsm");
 
 const VISIBLE = "visible";
 const VISIBILITY_CHANGE_EVENT = "visibilitychange";
 
 module.exports = class DetectUserSessionStart {
-  constructor(options = {}) {
+  constructor(store, options = {}) {
+    this._store = store;
     // Overrides for testing
-    this.sendAsyncMessage = options.sendAsyncMessage || global.sendAsyncMessage;
     this.document = options.document || global.document;
     this._perfService = options.perfService || perfSvc;
     this._onVisibilityChange = this._onVisibilityChange.bind(this);
@@ -42,10 +42,10 @@ module.exports = class DetectUserSessionStart {
       let visibility_event_rcvd_ts = this._perfService
         .getMostRecentAbsMarkStartByName("visibility_event_rcvd_ts");
 
-      this.sendAsyncMessage("ActivityStream:ContentToMain", {
+      this._store.dispatch(ac.SendToMain({
         type: at.SAVE_SESSION_PERF_DATA,
         data: {visibility_event_rcvd_ts}
-      });
+      }));
     } catch (ex) {
       // If this failed, it's likely because the `privacy.resistFingerprinting`
       // pref is true.  We should at least not blow up.
--- a/system-addon/content-src/lib/init-store.js
+++ b/system-addon/content-src/lib/init-store.js
@@ -6,6 +6,7 @@ const {actionTypes: at, actionCreators: ac, actionUtils: au} = require("common/A
 const MERGE_STORE_ACTION = "NEW_TAB_INITIAL_STATE";
 const OUTGOING_MESSAGE_NAME = "ActivityStream:ContentToMain";
 const INCOMING_MESSAGE_NAME = "ActivityStream:MainToContent";
+const EARLY_QUEUED_ACTIONS = [at.SAVE_SESSION_PERF_DATA, at.PAGE_PRERENDERED];
 
 /**
  * A higher-order function which returns a reducer that, on MERGE_STORE action,
@@ -77,6 +78,33 @@ const rehydrationMiddleware = store => next => action => {
 };
 
 /**
+ * This middleware queues up all the EARLY_QUEUED_ACTIONS until it receives
+ * the first action from main. This is useful for those actions for main which
+ * require higher reliability, i.e. the action will not be lost in the case
+ * that it gets sent before the main is ready to receive it. Conversely, any
+ * actions allowed early are accepted to be ignorable or re-sendable.
+ */
+const queueEarlyMessageMiddleware = store => next => action => {
+  if (store._receivedFromMain) {
+    next(action);
+  } else if (au.isFromMain(action)) {
+    next(action);
+    store._receivedFromMain = true;
+    // Sending out all the early actions as main is ready now
+    if (store._earlyActionQueue) {
+      store._earlyActionQueue.forEach(next);
+      store._earlyActionQueue = [];
+    }
+  } else if (EARLY_QUEUED_ACTIONS.includes(action.type)) {
+    store._earlyActionQueue = store._earlyActionQueue || [];
+    store._earlyActionQueue.push(action);
+  } else {
+    // Let any other type of action go through
+    next(action);
+  }
+};
+
+/**
  * initStore - Create a store and listen for incoming actions
  *
  * @param  {object} reducers An object containing Redux reducers
@@ -87,7 +115,7 @@ module.exports = function initStore(reducers, initialState) {
   const store = createStore(
     mergeStateReducer(combineReducers(reducers)),
     initialState,
-    global.addMessageListener && applyMiddleware(rehydrationMiddleware, messageMiddleware)
+    global.addMessageListener && applyMiddleware(rehydrationMiddleware, queueEarlyMessageMiddleware, messageMiddleware)
   );
 
   store._didRehydrate = false;
@@ -108,6 +136,8 @@ module.exports = function initStore(reducers, initialState) {
 };
 
 module.exports.rehydrationMiddleware = rehydrationMiddleware;
+module.exports.queueEarlyMessageMiddleware = queueEarlyMessageMiddleware;
 module.exports.MERGE_STORE_ACTION = MERGE_STORE_ACTION;
 module.exports.OUTGOING_MESSAGE_NAME = OUTGOING_MESSAGE_NAME;
 module.exports.INCOMING_MESSAGE_NAME = INCOMING_MESSAGE_NAME;
+module.exports.EARLY_QUEUED_ACTIONS = EARLY_QUEUED_ACTIONS;
--- a/system-addon/test/unit/lib/init-store.test.js
+++ b/system-addon/test/unit/lib/init-store.test.js
@@ -1,5 +1,5 @@
 const initStore = require("content-src/lib/init-store");
-const {MERGE_STORE_ACTION, rehydrationMiddleware} = initStore;
+const {MERGE_STORE_ACTION, EARLY_QUEUED_ACTIONS, rehydrationMiddleware, queueEarlyMessageMiddleware} = initStore;
 const {GlobalOverrider, addNumberReducer} = require("test/unit/utils");
 const {actionCreators: ac, actionTypes: at} = require("common/Actions.jsm");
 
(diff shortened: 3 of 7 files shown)


ncloudioj:
@Mardak r?

@k88hudson I had to attach the `_earlyMessageQueue` to `store` in the `rehydrationMiddleware()`, as we found out that doing so in `initStore()` didn't work out. Does that sound OK to you?

ncloudioj:
Also, I've filed a separate bug #3725 to fix the other two `store` properties `_didRehydrate` and `_didRequestInitialState`.

Mardak:
Could you explain more of #3725? I'm assuming it affects this code, so why are the attached properties not being saved on the store?

ncloudioj:
@Mardak This is the best explanation that I could find,

> we could write applyMiddleware() that first obtains the final, fully wrapped dispatch() function, and returns a copy of the store using it

https://github.com/reactjs/redux/blob/f35e98ce29fe327096bcb7db4e65d76872fc2bf1/docs/advanced/Middleware.md#attempt-6-na%C3%AFvely-applying-the-middleware

ncloudioj:
Perhaps @k88hudson knows more details about the magic of redux middleware ;)

k88hudson:
The `store` that middleware receives as a constant is unfortunately not the instance as the one that is created with `createStore` (it's an internal copy), so any properties added to the outer store won't get carried over to the middleware store. However, if you add properties inside the middleware, they will be persisted.  

This test will pass:
```js
  it("should be able to add custom properties within middleware", () => {
    let countA;
    let countB;
    const A = s => next => action => {
      if (!s._count) {
        s._count = 0;
      }
      s._count++;
      countA = s._count;
      next(action);
    };
    const B = s => next => action => {
      countB = s._count;
      next(action);
    };
    store = createStore(addNumberReducer, applyMiddleware(A, B));
    store._count = 5;
    store.dispatch({type: "FOO"});
    store.dispatch({type: "FOO"});
    store.dispatch({type: "FOO"});
    assert.equal(countA, 3, "countA");
    assert.equal(countB, 3, "countB");
  });
```

We can probaby go with this approach for now in this patch, but we should probably find a better way of doing it in a separate ticket.

system-addon/test/unit/lib/init-store.test.js
```diff
@@ -107,5 +132,22 @@ describe("initStore", () => {
...
+
+      assert.notCalled(next);
+      assert.equal(store._earlyMessageQueue.length, 1);
+      next.reset();
+
+      // A MERGE_STORE_ACTION would flush the early message queue
+      rehydrationMiddleware(store)(next)({type: MERGE_STORE_ACTION});
+
+      assert.calledOnce(next);
+      assert.calledWith(spy, sessionPerfAction);
```
Mardak: nit: let's name `spy` as `dispatch`

system-addon/content-src/lib/detect-user-session-start.js
```diff
@@ -7,7 +7,7 @@ const VISIBILITY_CHANGE_EVENT = "visibilitychange";
 module.exports = class DetectUserSessionStart {
   constructor(options = {}) {
     // Overrides for testing
-    this.sendAsyncMessage = options.sendAsyncMessage || global.sendAsyncMessage;
+    this._store = options.store;
```
Mardak: Looks like `options.store` isn't really optional. Probably have `store` be the first parameter and document it. Also, move this line before the `// Overrides for testing`.

system-addon/test/unit/lib/init-store.test.js
```diff
@@ -61,6 +62,30 @@ describe("initStore", () => {
     store.dispatch({type: "FOO"});
     assert.notCalled(global.sendAsyncMessage);
   });
+  it("should be able to add custom properties within middleware", () => {
```
Mardak: Not really sure if this test is necessary. Would be a bit similar to having a test that javascript `+` can add two numbers -- i.e., it should just work. And if it doesn't work, other things will be broken, e.g., the other test you added for `assert.equal(store._earlyMessageQueue.length, 1);`

system-addon/content-src/lib/init-store.js
```diff
@@ -44,12 +44,25 @@ const messageMiddleware = store => next => action => {
...
+
   if (store._didRehydrate) {
     return next(action);
   }
 
   const isMergeStoreAction = action.type === MERGE_STORE_ACTION;
   const isRehydrationRequest = action.type === at.NEW_TAB_STATE_REQUEST;
+  const isSessionPerfAction = action.type === at.SAVE_SESSION_PERF_DATA;
+
+  // Treasure up the session perf actions to the early message queue if the hydration
```
Mardak: not sure what you mean by "Treasure up" ?

system-addon/content-src/lib/init-store.js
```diff
@@ -44,12 +44,25 @@ const messageMiddleware = store => next => action => {
...
+  // is not done yet. All those actions will be dispatched on the MERGE_STORE_ACTION,
+  // which is guaranteed to come from chrome.
+  if (!store._didRehydrate && isSessionPerfAction) {
+    store._earlyMessageQueue.push(action);
+    return null;
```
Mardak: I don't quite understand what's expected from the return value of middleware action handling… @k88hudson ?

system-addon/content-src/lib/init-store.js
```diff
@@ -58,6 +71,10 @@ const rehydrationMiddleware = store => next => action => {
 
   if (isMergeStoreAction) {
     store._didRehydrate = true;
+    if (store._earlyMessageQueue.length) {
+      store._earlyMessageQueue.forEach(a => store.dispatch(a));
```
Mardak: Not sure if there's any concern of potentially mixed actions? E.g., getting an action dispatched in the middle of another action (vs after -- but it's not uncommon for an action to trigger a subsequent action). I would guess maybe the new action is appended to some action queue that is dispatched after the current action… @k88hudson ?

system-addon/test/unit/lib/init-store.test.js
ncloudioj: You're right, perhaps adding some comments in init-store.jsm is better for now. Eventually, this could be fixed by #3725

system-addon/content-src/lib/init-store.js
ncloudioj: Oops, my bad, what I was thinking here. Perhaps got influenced by this https://youtu.be/vm1GJMp0QN4?t=2681

system-addon/content-src/lib/init-store.js
```diff
@@ -44,12 +44,27 @@ const messageMiddleware = store => next => action => {
 };
 
 const rehydrationMiddleware = store => next => action => {
+  // XXX Due to the internal mechanism of redux middleware, we had to add this property in here, see
+  // more details in #3725. We should refactor this workaround via fixing #3725.
+  if (!store._earlyMessageQueue) {
```
k88hudson: I would put this after the `store._didRehydrate` check, since it will save a check for most actions

system-addon/content-src/lib/init-store.js
Mardak: Just to be clear, when you say "dispatched *inside* its middleware", is this via `next` instead of `store.dispatch`?

ncloudioj:
Have refactored the `rehydrateMiddleware` and added a `queueEarlyPerfActionMiddleware`. Note that it only queues up those `SAVE_SESSION_PERF_DATA` actions, not sure if we wanted to do that for other send-to-chrome actions.

@k88hudson One more r?

system-addon/content-src/lib/init-store.js
```diff
@@ -76,6 +76,29 @@ const rehydrationMiddleware = store => next => action => {
...
+  }
+
+  if (action.type === MERGE_STORE_ACTION) {
+    next(action);
+    store._ready = true;
+    if (store._queue && store._queue.length) {
+      store._queue.forEach(next);
+    }
+    return null;
+  } else if (action.type === at.SAVE_SESSION_PERF_DATA) {
```
Mardak: This is going to miss on the newly added PAGE_PRERENDERED. I suppose even if we added a `prerendered: true` to NEW_TAB_STATE_REQUEST, we would need to stash that action in `rehydrationMiddleware` to replay the original action with extra data instead of sending a newly created REQUEST.

system-addon/content-src/lib/init-store.js
```diff
@@ -76,6 +76,29 @@ const rehydrationMiddleware = store => next => action => {
...
+    store._queue = store._queue || [];
+    store._queue.push(action);
+    return null;
+  }
+
+  return next(action);
```
Mardak: I believe part of the refactoring was to remove the `return next()` to avoid the confusion of needing to return some value.

system-addon/content-src/lib/init-store.js
```diff
@@ -76,6 +76,29 @@ const rehydrationMiddleware = store => next => action => {
   return next(action);
 };
 
+// This middleware queues up all the SAVE_SESSION_PERF_DATA actions until it
+// receives the MERGE_STORE_ACTION, which is guaranteed to come from chrome.
```
Mardak: There's plenty of messages from main (I think we should be consistent with just calling chrome "main"), that would allow the queue to be emptied. @k88hudson Would it be safe to treat any `!au.isSendToMain(action)` as coming from main? I suppose some content could dispatch to itself, so I guess not. Although we can just as easily add an `au.isFromMain(action)`.

Then I would see this early queue behaving like:
```js
… => {
  if (store._receivedFromMain || action.type === at.NEW_TAB_STATE_REQUEST) {
    next(action);
  } else if (au.isFromMain(action)) {
    // empty the queue
  } else {
    // add to queue
  }
}
```

system-addon/content-src/lib/init-store.js
ncloudioj: Indeed, would it be possible to queue up all the content-to-main actions and empty them once the main is ready (e.g. "NEW_TAB_INITIAL_STATE")? Not quite sure about that, in particular, given this [test](https://github.com/mozilla/activity-stream/blob/master/system-addon/test/unit/lib/init-store.test.js#L55). Thoughts? @k88hudson

ncloudioj: After discussing this with @k88hudson, instead of stashing all the early content-to-main actions, we really just wanted to do that for "SAVE_SESSION_PERF_DATA" and "PAGE_PRERENDERED" for now. This middleware will let all other actions (content-to-content, content-to-main, main-to-content) go through immediately.

ncloudioj:
@Mardak Refactored it a bit based on your review, r? :P

system-addon/content-src/lib/init-store.js
```diff
@@ -76,6 +77,28 @@ const rehydrationMiddleware = store => next => action => {
   return next(action);
 };
 
+// This middleware queues up all the EARLY_QUEUED_ACTIONS until it receives
+// the first action from main. Note that NEW_TAB_INITIAL_STATE is guaranteed
+// to be sent from main
```
Mardak: Switch this to a `/* */` comment block, and explain things a bit better probably without the "NEW_TAB_INITIAL_STATE" mention. Probably something including "because any messages for main before receiving from main might be lost. Any actions allowed early are accepted to be unreliable, i.e., can handle re-sending or be ignored."

system-addon/content-src/lib/init-store.js
```diff
@@ -76,6 +77,28 @@ const rehydrationMiddleware = store => next => action => {
...
+const queueEarlyMessageMiddleware = store => next => action => {
+  if (store._receivedFromMain) {
+    next(action);
+  } else if (au.isFromMain(action)) {
+    next(action);
+    store._receivedFromMain = true;
+    if (store._queue && store._queue.length) {
```
Mardak: The `length` check shouldn't be necessary. And the `_queue` name is a bit generic especially we're just attaching to some effectively shared/global `store` object.

Mardak: Also, add a comment here saying that we're sending out any early actions /after/ processing the message from main.

system-addon/test/unit/lib/init-store.test.js
```diff
@@ -108,4 +108,50 @@ describe("initStore", () => {
...
+  describe("queueEarlyMessageMiddleware", () => {
+    it("should allow all local actions to go through", () => {
+      const action = {type: "FOO"};
+      const next = sinon.spy();
+
+      queueEarlyMessageMiddleware(store)(next)(action);
+
+      assert.calledWith(next, action);
+    });
+    it("should allow none EARLY_QUEUED_ACTIONS actions to main to go through", () => {
```
Mardak: Probably looking for a word other than "none" ?

system-addon/test/unit/lib/init-store.test.js
```diff
@@ -108,4 +108,50 @@ describe("initStore", () => {
...
+        const testStore = initStore({number: addNumberReducer});
+        const next = sinon.spy();
+        const action = ac.SendToMain({type: actionType});
+        const mergeStoreAction = ac.SendToContent({type: MERGE_STORE_ACTION}, 123);
+
+        // Early action should be added to the queue
+        queueEarlyMessageMiddleware(testStore)(next)(action);
+
+        assert.notCalled(next);
+        assert.equal(testStore._queue.length, 1);
```
Mardak: I think this test would be more interesting if multiple things got queued. Either the same action or different actions.

PR #3734, opened 2017-10-18
Add visit for Top Story bookmarks

### LINKED ISSUE #3131: Pocket story bookmark status is not persisted in context menu

piatra:
Closes #3131.

Add a visit for Top Story bookmarks so they get picket up by the Highlights query.

This also includes a small fix for updateBookmarkMetadata (when bookmarking a Top Story we want to save metadata so that it shows up the same way in Highlights).
The first issue was that we were iterating through all the sections but we are only interested if it is a Top Story. 
If you bookmarked a "history" Highlight the `card.image` field would have been a base64 with length > DB_URL_LENGTH_MAX triggering an error.

### DIFF
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -247,6 +247,18 @@ class PlacesFeed {
     win.openLinkIn(action.data.url, "window", params);
   }
 
+  /**
+   * Currently used when bookmarking a TopStory to ensure it gets picked up in
+   * the Highlights section.
+   * If used in other cases, should ensure URLs actually have 0 visits.
+   *
+   * @param {string} site.url
+   * @param {string} site.title
+   */
+  addURLVisit({url, title}) {
+    PlacesUtils.history.insert({url, title, visits: [new Date()]});
+  }
+
   onAction(action) {
     switch (action.type) {
       case at.INIT:
@@ -261,6 +273,9 @@ class PlacesFeed {
         break;
       case at.BOOKMARK_URL:
         NewTabUtils.activityStreamLinks.addBookmark(action.data, action._target.browser);
+        break;
+      case at.ADD_URL_VISIT:
+        this.addURLVisit(action.data);
         break;
       case at.DELETE_BOOKMARK_BY_ID:
         NewTabUtils.activityStreamLinks.deleteBookmark(action.data);
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -247,16 +247,6 @@ class PlacesFeed {
     win.openLinkIn(action.data.url, "window", params);
   }
 
-  /**
-   * Add a visit to the provided URL with the current timestamp.
-   *
-   * @param {string} site.url
-   * @param {string} site.title
-   */
-  addURLVisit({url, title}) {
-    PlacesUtils.history.insert({url, title, visits: [new Date()]});
-  }
-
   onAction(action) {
     switch (action.type) {
       case at.INIT:
@@ -271,9 +261,6 @@ class PlacesFeed {
         break;
       case at.BOOKMARK_URL:
         NewTabUtils.activityStreamLinks.addBookmark(action.data, action._target.browser);
-        break;
-      case at.ADD_URL_VISIT:
-        this.addURLVisit(action.data);
         break;
       case at.DELETE_BOOKMARK_BY_ID:
         NewTabUtils.activityStreamLinks.deleteBookmark(action.data);
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -248,9 +248,7 @@ class PlacesFeed {
   }
 
   /**
-   * Currently used when bookmarking a TopStory to ensure it gets picked up in
-   * the Highlights section.
-   * If used in other cases, should ensure URLs actually have 0 visits.
+   * Add a visit to the provided URL with the current timestamp.
    *
    * @param {string} site.url
    * @param {string} site.title
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -7,7 +7,6 @@ const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/EventEmitter.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
-Cu.import("resource://gre/modules/Console.jsm");
 const {actionCreators: ac, actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
 
 XPCOMUtils.defineLazyModuleGetter(this, "PlacesUtils", "resource://gre/modules/PlacesUtils.jsm");
@@ -153,7 +152,7 @@ const SectionsManager = {
   },
 
   /**
-   * Save Top Story metadata to places db.
+   * Save metadata to places db and add a visit for that URL.
    */
   updateBookmarkMetadata({url}) {
     this.sections.forEach((section, id) => {
@@ -170,12 +169,10 @@ const SectionsManager = {
               description: card.description,
               previewImageURL: card.image
             });
-            this.emit(this.DISPATCH_TO_MAIN, {
-              type: at.ADD_URL_VISIT,
-              data: {
-                url: card.url,
-                title: card.title
-              }
+            PlacesUtils.history.insert({
+              url,
+              title: card.title,
+              visits: [new Date()]
             });
           }
         });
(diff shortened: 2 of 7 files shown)


system-addon/lib/SectionsManager.jsm
```diff
@@ -151,21 +151,37 @@ const SectionsManager = {
...
+  /**
+   * When we bookmark a TopStory we need to add a visit to that URL
+   * to ensure it gets picked up in Highlights.
+   *
+   * @param {string} site.url
+   * @param {string} site.title
+   * @param {string} site.type Required to determine if it is a Top Story.
+   */
+  addBookmarkVisit({url, title, type}) {
+    if (type && (type === "trending" || type === "now")) {
```
Mardak: This seems like a pretty fragile way to determine if the thing being bookmarked is a story. We happen to only use those types for stories right now, but it doesn't seem unreasonable that Highlights or other sections might reuse those types.

On the other hand, if some other section say through webextension, allows a card to be bookmarked (are all cards bookmarkable?) would we expect those items to show up in Highlights if unvisited?

So I guess either the specificity of this behavior is too coupled with types when it should be checking for top-stories-sourced-cards or that it should be more general. If it should only be for stories, this action handler should probably be in `TopStoriesFeed` or if more general, in `PlacesFeed`.

system-addon/lib/SectionsManager.jsm
```diff
@@ -151,21 +151,37 @@ const SectionsManager = {
...
-              url: card.url,
-              title: card.title,
-              description: card.description,
-              previewImageURL: card.image
-            });
-          }
-        });
-      }
-    });
+    const {rows} = this.sections.get("topstories");
```
Mardak: Similar to the another comment about being specific to top stories or general for any section… if it's story specific, this should probably be in `TopStoriesFeed`. Or if it's supposed to be more general, e.g., for webextension, then maybe it should check each section.. except highlights? @k88hudson any thoughts on if we should be saving metadata when bookmarking any section's cards?

k88hudson: It would probably be a good idea to save metadata if we have it, yeah 👍

piatra: @Mardak I think building something general enough seems out of scope for this bug since webextensions are somewhat experimental. Maybe for now the logic should be handled by the TopStoriesFeed since its very specific.

system-addon/lib/SectionsManager.jsm
piatra: @Mardak would the correct behavior be to move this to PlacesFeed and add a history visit if we can't find any? Do you think that might be to expensive?

system-addon/lib/SectionsManager.jsm
Mardak: The general approach is what the code was doing before. Except it should skip highlights to avoid the bug you noticed.

system-addon/lib/SectionsManager.jsm
Mardak: Probably not worth it to check if visited first, but probably good to note it in a comment.

system-addon/lib/PlacesFeed.jsm
```diff
@@ -239,6 +239,18 @@ class PlacesFeed {
     win.openLinkIn(action.data.url, "window", params);
   }
 
+  /**
+   * Currently used when bookmarking a TopStory to ensure it gets picked up in
+   * the Highlights section.
+   * If used in other cases, should ensure URLs actually have 0 visits.
```
Mardak: Oh. I was thinking this would just be called as part of BOOKMARK_URL, but this is good. This method description should describe what it does not how it's used. Also, the "should ensure URLs" is misleading as it seems to imply TopStory is making sure there's 0 visits. This method seems to be more of "add url with visit" unless it really does break if the url already exists?

piatra: No breakage just adds another visit.

system-addon/lib/SectionsManager.jsm
```diff
@@ -162,6 +170,13 @@ const SectionsManager = {
               description: card.description,
               previewImageURL: card.image
             });
+            this.emit(this.DISPATCH_TO_MAIN, {
```
Mardak: I'm not sure what's the expected / allowed `emit`s. Wherever we have this behavior, it will still want a comment as previously requested.

Mardak: I'm not sure what's the expected / allowed `emit`s. @k88hudson ? Wherever we have this behavior, it will still want a comment as previously requested.

k88hudson: I'm confused as to why this needs to be emitted/added to PlacesFeed, why can't you just call ` PlacesUtils.history.insert` directly? Also is this the correct behaviour if the item is already in the history?

piatra: @k88hudson my thinking was that doing the insert in PlacesFeed would allow us to have more control over the action in the future (like checking if the website has already been visited).
Regarding your second question this will now only happen for topstories, highlights are skipped over. I don't see any problem in having the extra visit.

k88hudson: Personally that seems like premature abstraction to me, but if you are going to do it that way I'd probably go with a little more encapsulation and have a specific ADD_URL event rather than a generic DISPATCH_TO_MAIN

piatra: I'll revert to directly calling `PlacesUtils.history.insert`.

system-addon/lib/SectionsManager.jsm
```diff
@@ -162,6 +169,11 @@ const SectionsManager = {
               description: card.description,
               previewImageURL: card.image
             });
+            PlacesUtils.history.insert({
```
Mardak: This still needs a comment

system-addon/test/unit/lib/SectionsManager.test.js
```diff
@@ -210,23 +210,38 @@ describe("SectionsManager", () => {
     });
   });
   describe("#updateBookmarkMetadata", () => {
-    let rows;
+    let clock;
+
     beforeEach(() => {
-      rows = [{
+      clock = sinon.useFakeTimers();
```
Mardak: What's the clock for?

piatra: Testing the call arguments on `insert`, one of the arguments is `new Date()`.

system-addon/test/unit/lib/SectionsManager.test.js
```diff
@@ -237,6 +252,16 @@ describe("SectionsManager", () => {
...
       });
     });
+    it("should call PlacesUtils.history.insert", () => {
+      SectionsManager.updateBookmarkMetadata({url: "bar"});
+
+      assert.calledOnce(fakePlacesUtils.history.insert);
+      assert.calledWithExactly(fakePlacesUtils.history.insert, {
+        url: "bar",
+        title: "title",
+        visits: [new Date()]
```
Mardak: Not sure if we need to be testing that the specific added visit date ?

piatra: Specific date isn't important but if it's missing insert will trow an error https://searchfox.org/mozilla-central/source/toolkit/components/places/PlacesUtils.jsm#1034
Could check the arguments keys instead.

system-addon/lib/SectionsManager.jsm
```diff
@@ -162,6 +169,11 @@ const SectionsManager = {
...
+              url,
+              title: card.title,
+              visits: [new Date()]
```
Mardak: Looking at `PlacesUtils.validatePageInfo` looks like it will use now as a default visit date anyway: https://searchfox.org/mozilla-central/source/toolkit/components/places/PlacesUtils.jsm#1037-1058

Mardak: And actually it doesn't expect a `Date` object to come in in the first place. Where did you see something passing in visits array of `Date`s to do it here?

as-pine-proxy:
[Treeherder: f58fcc9d071ab1c47d0fd7cb865ba18ee3dd6b6c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f58fcc9d071ab1c47d0fd7cb865ba18ee3dd6b6c)

PR #3713, opened 2017-10-14
chore(react): Switch ConfirmDialog to extend PureComponent

Mardak:
Fix #3706. r?@rlr Like #3480 but also making it a `class`.

PR #3735, opened 2017-10-19
chore(react): Remove unused and unnecessary defaultProps from ConfirmDialog

Mardak:
This was in the console when prerendering:
`Warning: getDefaultProps was defined on ConfirmDialog, a plain JavaScript class. This is only supported for classes created using React.createClass. Use a static property to define defaultProps instead.`

Followup from #3713. r?@rlr We get the props through `connect` state.
https://github.com/mozilla/activity-stream/blob/d6b67193c571c697105d1ee04f595f4f7f4477b3/system-addon/common/Reducers.jsm#L47-L49


as-pine-proxy:
[Treeherder: 8d080236cccbeb7afca11e0fa34129e7da5065c1](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8d080236cccbeb7afca11e0fa34129e7da5065c1)

PR #3740, opened 2017-10-19
feat(prerender): Use real strings and text direction for per-locale prerendered html

### LINKED ISSUE #3370: Support real strings in prerendered html
Not sure if this will be covered by https://github.com/mozilla/activity-stream/issues/3367, but we would like to have real strings in our prerendered html

Mardak:
Fix #3370. r?@k88hudson Running tests results in spewing these errors: `ERROR: '[React Intl] Missing message: "search_web_placeholder" for locale: "en-US"'`

Any good way to suppress? Or I suppose I could pass in actual strings…

Here's a screenshot of prerendered "ar":
screen shot 2017-10-19 at 9 57 41 am

As well as partially-"prerendered" (just the title/locale/dir) "ar":
screen shot 2017-10-19 at 9 58 59 am


bin/render-activity-stream-html.js
```diff
@@ -100,26 +258,101 @@ function main() {
...
+  }
 
-  console.log(`Done writing html and js to ${BASE_FILE_PATH}.`); // eslint-disable-line no-console
+  // Replace existing render-generated lines in jar.mn
+  const scriptName = path.basename(__filename);
+  const jarFile = path.join(addonPath, "jar.mn");
+  const jarLines = fs.readFileSync(jarFile, "utf8").split("\n");
+  const startLine = jarLines.findIndex(line => line.includes(scriptName));
+  const endLine = jarLines.indexOf("#endif", startLine);
+  jarLines.splice(startLine, endLine - startLine + 1, ...[
```
k88hudson: 😎

bin/render-activity-stream-html.js
```diff
@@ -100,26 +258,101 @@ function main() {
...
+      title: state.App.strings.newtab_page_title
+    });
+
+    // Put locale-specific files in their own directory
+    const basePath = path.join(addonPath, "prerendered", locale);
+    mkdir("-p", basePath);
+    for (const [file, templater] of localizedFiles) {
+      fs.writeFileSync(path.join(basePath, file), templater({html, options, state}));
+    }
+    console.log(`Done writing "${locale}" localized files to ${basePath}.`);
```
k88hudson: Personally, I would do something like this for the output:
```
Writing localized files to /whatever/system-addon/prerendered
✔ en-US
✔ debug
✔ ach
✔ ar
...
Skipped the following locales because they had the same strings as en-US: af, ak, an, as, ...
Done.
```

You could maybe try something with `process.stdout.write` to get the locales on the same line, but yeah this obviously up to you

bin/render-activity-stream-html.js
```diff
@@ -100,26 +258,101 @@ function main() {
...
+      targetLines = ifLines;
+    }
+    for (const file of localizedFiles.keys()) {
+      targetLines.push(`  content/data/content/${file} (./prerendered/${locale}/${file})`);
+    }
+  }
 
-  console.log(`Done writing html and js to ${BASE_FILE_PATH}.`); // eslint-disable-line no-console
+  // Replace existing render-generated lines in jar.mn
+  const scriptName = path.basename(__filename);
```
k88hudson: Can you actually include a complete path here if someone is trying to find this when looking at mc?

bin/render-activity-stream-html.js
```diff
@@ -1,18 +1,176 @@
...
-  baseUrl: "resource://activity-stream/",
-  locale: "", // TODO: pass in options.locale. For now, we're just doing empty strings.
-  title: ""
+  addonPath: "../system-addon",
+  baseUrl: "resource://activity-stream/"
 };
 
+// From https://hg.mozilla.org/l10n-central/
+// console.log(`const MOZILLA_CENTRAL_LOCALES = ${JSON.stringify([...document.querySelectorAll("a.list:not([href*='x-testing']")].map(a => a.textContent.trim()), null, 2)};`)
+const MOZILLA_CENTRAL_LOCALES = [
```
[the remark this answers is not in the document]
Mardak: So… "ia" was added since I originally created a local branch 1 month ago, and it has 100% strings localized for activity stream: https://pontoon.mozilla.org/ia/activity-stream-new-tab/strings.properties/

Would be unfortunate if we didn't notice and missed the prerendering for that locale. But on that note, if we have strings for a locale that we aren't building, we should be able to notice right away !

Should it break this render script? Big warning? I suppose the strings import script could be smarter, but then we'll want to share this array outside of this file.

Or alternatively, we can dynamically augment MOZILLA_CENTRAL_LOCALES with the locales that we have localized strings.

Actually, I think I'll go with the dynamic augment and big warning.

And just to be explicit, we can't rely on our own locale list from locales.json as we would have missed out on bn-IN, ja-JP-mac, ta-LK which we have similar strings to reuse from bn-BD, ja, ta respectively.

bin/render-activity-stream-html.js
Mardak: What's a "complete" path? "bin/render-activity-stream-html.js" or "https://github.com/mozilla/activity-stream/blob/master/bin/render-activity-stream-html.js"

bin/render-activity-stream-html.js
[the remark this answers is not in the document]
Mardak: Added a comment documenting the purpose of this list.

bin/render-activity-stream-html.js
Mardak: Maybe.. I.. got.. carried.. away…

screen shot 2017-10-20 at 2 04 56 pm

bin/render-activity-stream-html.js
```diff
@@ -1,18 +1,176 @@
- /* eslint-disable no-sync */
+ /* eslint-disable no-console, no-sync */
```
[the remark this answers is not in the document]
Mardak: Set `no-sync: 0`

bin/render-activity-stream-html.js
Mardak: Used full url:
`# These defines and ifs below are generated by https://github.com/mozilla/activity-stream/blob/master/bin/render-activity-stream-html.js`

as-pine-proxy:
[Treeherder: 097b2e18d5c85f2f6875c88f096567ccfadf6c68](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=097b2e18d5c85f2f6875c88f096567ccfadf6c68)

PR #3744, opened 2017-10-19
Register thumbnail expiration filters for topsites and highlights

### LINKED ISSUE #3646: Be smarter about when to expire thumbnails (especially for Highlights)
Right now thumbnails are expired based on other thumbnail user's logic:
https://searchfox.org/mozilla-central/search?q=filterForThumbnailExpiration

This means Highlights most likely are being thrown out of the cache pretty regularly whereas Top Sites might happen to work because they generally show the same thing as Tiles.

piatra:
Closes #3646. 

Needs a follow up bug in m-c to remove the current TopSites expiration https://github.com/mozilla/activity-stream/issues/3646#issuecomment-337616103. Should be fine to land this independently right?

### DIFF
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -61,7 +61,17 @@ this.HighlightsFeed = class HighlightsFeed {
       const {initialized, rows} = this.store.getState().Sections[sectionIndex];
 
       if (initialized) {
-        callback(rows.map(site => site.url));
+        const linksToKeep = rows.reduce((acc, site) => {
+          // Screenshots will search for preview_image_url or fallback to URL,
+          // so we prevent both from being expired.
+          // https://github.com/mozilla/activity-stream/blob/95b4c35393b7192d680d1291b6960200be5e7570/system-addon/lib/HighlightsFeed.jsm#L131
+          acc.push(site.url);
+          if (site.preview_image_url) {
+            acc.push(site.preview_image_url);
+          }
+          return acc;
+        }, []);
+        callback(linksToKeep);
       } else {
         callback([]);
       }
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -57,27 +57,18 @@ this.HighlightsFeed = class HighlightsFeed {
 
   filterForThumbnailExpiration(callback) {
     const sectionIndex = SectionsManager.sections.get(SECTION_ID).order;
-    if (this.store.getState().Sections[sectionIndex]) {
-      const {initialized, rows} = this.store.getState().Sections[sectionIndex];
+    const state = this.store.getState().Sections[sectionIndex];
 
-      if (initialized) {
-        const linksToKeep = rows.reduce((acc, site) => {
-          // Screenshots will search for preview_image_url or fallback to URL,
-          // so we prevent both from being expired.
-          // https://github.com/mozilla/activity-stream/blob/95b4c35393b7192d680d1291b6960200be5e7570/system-addon/lib/HighlightsFeed.jsm#L131
-          acc.push(site.url);
-          if (site.preview_image_url) {
-            acc.push(site.preview_image_url);
-          }
-          return acc;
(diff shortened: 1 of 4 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -57,6 +65,16 @@ this.TopSitesFeed = class TopSitesFeed {
     }
   }
 
+  filterForThumbnailExpiration(callback) {
+    const {initialized, rows} = this.store.getState().TopSites;
+
+    if (initialized) {
+      callback(rows.map(site => site.url));
+    } else {
+      callback([]);
```
Mardak: nit: TopSites state should always has an array for rows, so the condition is unnecessary

Mardak:
Any good way to test this?

piatra:
If you mean just manual testing I reduced https://searchfox.org/mozilla-central/source/toolkit/components/thumbnails/PageThumbs.jsm#17 and logged the urls https://searchfox.org/mozilla-central/rev/d0448c970093f94bd986a21d3a9e8c366b316eb4/toolkit/components/thumbnails/PageThumbs.jsm#845.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -49,6 +52,18 @@ this.HighlightsFeed = class HighlightsFeed {
 
   uninit() {
     SectionsManager.disableSection(SECTION_ID);
+    PageThumbs.removeExpirationFilter(this);
+  }
+
+  filterForThumbnailExpiration(callback) {
+    const sectionIndex = SectionsManager.sections.get(SECTION_ID).order;
+    const {initialized, rows} = this.store.getState().Sections[sectionIndex];
```
Mardak: Looks like it's possible for this line to throw if the section doesn't exist in the state. Seems unlikely but if it were to break, it looks like thumbnail expiration filter isn't smart enough to recover and waits forever.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -49,6 +52,22 @@ this.HighlightsFeed = class HighlightsFeed {
 
   uninit() {
     SectionsManager.disableSection(SECTION_ID);
+    PageThumbs.removeExpirationFilter(this);
+  }
+
+  filterForThumbnailExpiration(callback) {
+    const sectionIndex = SectionsManager.sections.get(SECTION_ID).order;
+    if (this.store.getState().Sections[sectionIndex]) {
+      const {initialized, rows} = this.store.getState().Sections[sectionIndex];
```
Mardak: This unnecessarily calls `getState()` twice in the common case. Probably just convert these 11 lines to:
```js
const state = this.store.getState().Sections[sectionIndex];
callback(state && state.initialized ? state.rows.map(({url}) => url) : []);
```

piatra:
It appears that for some URLs this is failing because PageThumbs doesn't have the image for it.

```
PageThumbsStorage.fileExistsForURL(url); // returns false
```

I have two bookmarks with preview image in my Highlights and the method above returns true for one and false for the other. I'm still looking into it.

piatra:
@Mardak I think I figured it out: when we make a request our [first option is to use preview_image_url](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/HighlightsFeed.jsm#L131) which seems to always exist. But when we make the cache request we use the highlight card url.
Should we change the caching to keep `preview_image_url` or what would be the correct approach?

Mardak:
Probably should make it keep the thumbnails that Highlights actually uses. I suppose we could be a bit excessive in requesting both the `url` and `preview_image_url` to not expire?

piatra:
> I suppose we could be a bit excessive

Even if we are, `const EXPIRATION_INTERVAL_SECS = 3600;` makes sure things are deleted often enough.

as-pine-proxy:
[Treeherder: 50384932db6b5a219e330d6bebbb71d5652efa8b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=50384932db6b5a219e330d6bebbb71d5652efa8b)

PR #3746, opened 2017-10-19
chore(performance): Cache affinity scores and calculate on idle-daily

csadilek:
This is a refactoring to no longer trigger domain affinity calculation on the first fetch, but on idle-daily, and cache the result so it's available on next startup.

This is a performance trade-off. So, in case idle-daily hasn't run yet, there will be no sorting, and the result is the same as the un-personalized, global feed.


as-pine-proxy:
[Treeherder: 58139d3e4ec3819c20300a53a7c79eaf58f353ac](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=58139d3e4ec3819c20300a53a7c79eaf58f353ac)

PR #3754, opened 2017-10-23
Ignore first run bookmark added events.

### LINKED ISSUE #3724: Multiple calls to linksCache.expire/fetchHighlights on startup in a new profile
> One thing I noticed when testing the PR, but this probably should be its own issue/PR. #3675 added expiring the highlights links cache on BOOKMARK_ADDED. A new profile causes adds of several bookmarks causing the cache to be expired and fetched from places multiple times on startup.

from https://github.com/mozilla/activity-stream/pull/3695#issuecomment-337165574

piatra:
Closes #3724

### DIFF
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -108,9 +108,11 @@ class BookmarksObserver extends Observer {
     const type = args[3];
     const source = args[9];
     const uri = args[4];
+    // Skips items that are not bookmarks (like folders), about:* pages or
+    // default bookmarks, added when the profile is created.
     if (type !== PlacesUtils.bookmarks.TYPE_BOOKMARK ||
         source === PlacesUtils.bookmarks.SOURCES.IMPORT_REPLACE ||
-        (uri && (uri.scheme !== "http" && uri.scheme !== "https"))) {
+        (uri.scheme !== "http" && uri.scheme !== "https")) {
       return;
     }
     const bookmarkTitle = args[5];
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -109,7 +109,7 @@ class BookmarksObserver extends Observer {
     const source = args[9];
     const uri = args[4];
     if (type !== PlacesUtils.bookmarks.TYPE_BOOKMARK ||
-        source !== PlacesUtils.bookmarks.SOURCES.DEFAULT ||
+        source === PlacesUtils.bookmarks.SOURCES.IMPORT_REPLACE ||
         (uri && (uri.scheme !== "http" && uri.scheme !== "https"))) {
       return;
     }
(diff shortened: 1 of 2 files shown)


system-addon/lib/PlacesFeed.jsm
```diff
@@ -100,13 +100,19 @@ class BookmarksObserver extends Observer {
...
+   * @param  {int} source    Used to distinguish bookmarks made by different
+   *                         actions: sync, bookmarks import, other.
    */
   onItemAdded(...args) {
     const type = args[3];
-    if (type !== PlacesUtils.bookmarks.TYPE_BOOKMARK) {
+    const source = args[9];
+    const uri = args[4];
+    if (type !== PlacesUtils.bookmarks.TYPE_BOOKMARK ||
+        source !== PlacesUtils.bookmarks.SOURCES.DEFAULT ||
```
Mardak: We should have comments on why we're doing these checks. And we probably should skip for `SOURCES.IMPORT_REPLACE` instead of skip if not `SOURCES.DEFAULT` as we haven't run into bugs for other types of sources but filtering out too much might introduce new bugs.

system-addon/lib/PlacesFeed.jsm
```diff
@@ -100,13 +100,19 @@ class BookmarksObserver extends Observer {
...
+        (uri && (uri.scheme !== "http" && uri.scheme !== "https"))) {
```
Mardak: This check for `uri` shouldn't be necessary as if we're getting a `TYPE_BOOKMARK`, it should definitely have a `uri`.

piatra: @Mardak the very first event that comes in to create the `Firefox Nightly Resources` folder does not have an uri.

system-addon/lib/PlacesFeed.jsm
piatra: 👍

system-addon/lib/PlacesFeed.jsm
Mardak: The `||` operator short circuits, so something like `true || null.foo` does not throw. The `uri.scheme` won't throw an exception for "Firefox Nightly Resources", which is both *not* a `TYPE_BOOKMARK` and *is* a `SOURCES.IMPORT_REPLACE`. That's why I said a `TYPE_BOOKMARK` should definitely have a `uri`.

Mardak: Here's more info https://en.wikipedia.org/wiki/Short-circuit_evaluation

Mardak: Or javascript specific on MDN:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-circuit_evaluation

Mardak: Or if you were confused by the precedence of things: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

as-pine-proxy:
[Treeherder: 8839c4856b76f040129f984052409a4e0fe1f262](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8839c4856b76f040129f984052409a4e0fe1f262)

PR #3755, opened 2017-10-24
fix(prerender): Save strings outside of redux state that are immutable

### LINKED ISSUE #3741: Only use prerendered strings instead of waiting for LOCALE_UPDATED
With #3740 we have the strings available in `gActivityStreamPrerenderedState` for fully prerendered html, and we could make something like that available for partially prerendered html to have strings available. Then in prerendered or not, the state will already have strings, and there wouldn't be a need to have `LocalizationFeed` separately dispatch strings.

This means we would rely on the packaged locale instead of the current `negotiateLanguages`:
https://github.com/mozilla/activity-stream/blob/9e996c8dbb308b43805ae8262084769074c6ba35/system-addon/lib/LocalizationFeed.jsm#L34-L37

But this would allow us to send NEW_TAB_INITIAL_STATE sooner without waiting for strings to be loaded from the feed:
https://github.com/mozilla/activity-stream/blob/9e996c8dbb308b43805ae8262084769074c6ba35/system-addon/lib/NewTabInit.jsm#L24-L37

@k88hudson

### LINKED ISSUE #3750: Move locale related state (strings and direction) out of redux to a global
From #3740:
> We should do a follow-up to this to remove the strings from the state entirely and just include the locale/strings in a top-level variable for the app to read

Instead of having react process redux state related to locale to decide to not update things on first load and locale update (before #3741), we can just do things once instead of each update of Base.

### LINKED ISSUE #3790: History entry is named "New Tab" when navigating directly to a video from about:home and New Tab
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1403414

Navigating to a video, e.g., https://ed.agadak.net/as/quantum%20startup%20dominik.mp4 shows "New Tab" in history.

Mardak:
Fix #3741 and fix #3750 and fix #3790. @k88hudson is this what you were thinking of? Add a global `gActivityStreamStrings` that gets passed in to `Base` and removing things that depended on `LocalizationFeed`? The `<title>` isn't necessary as Firefox will just use the packaged "tabs.emptyTabTitle" string: https://searchfox.org/mozilla-central/source/browser/base/content/tabbrowser.xml#2653

Wasn't sure where `system-addon/data/locales.json` should go. Maybe just stay there for now and removed before packaging?


system-addon/common/Reducers.jsm
```diff
@@ -18,12 +15,6 @@ const INITIAL_STATE = {
   App: {
     // Have we received real data from the app yet?
     initialized: false,
-    // The locale of the browser
-    locale: "",
```
k88hudson: This is set in `TelemetryFeed`, we probably want to have that send `Services.locale.getRequestedLocale()` instead

Mardak: Oh doh. I looked at `TelemetryFeed` too quickly. It said `appInfo` and I assumed it was `Services.appinfo`.

k88hudson:
Other than that one little nit this looks great, pretty much exactly what I was thinking 👍

Mardak:
Rebased on #3775.

system-addon/common/Reducers.jsm
Mardak: Switched to `Services.locale.getRequestedLocale()` and updated test.

Mardak:
@k88hudson I updated the PR with the `prerendered/{locales,static}` structure. And moved the initial state to static.

system-addon/lib/NewTabInit.jsm
```diff
@@ -34,21 +33,8 @@ this.NewTabInit = class NewTabInit {
   onAction(action) {
     switch (action.type) {
       case at.NEW_TAB_STATE_REQUEST:
-        // If localization hasn't been loaded yet, we should wait for it.
```
k88hudson: \o/ 🎉

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -259,7 +259,7 @@ this.TelemetryFeed = class TelemetryFeed {
     const appInfo = this.store.getState().App;
     const ping = {
       addon_version: appInfo.version,
-      locale: appInfo.locale,
+      locale: Services.locale.getRequestedLocale(),
```
k88hudson: How expensive is this check? Inexpensive enough that we can check on every ping created?

Mardak: Mmm... 
```js
s = performance.now(); for (i = 0; i < 1000000; i++) Services.locale.getRequestedLocale(); performance.now() - s`
1078.2949999999255
```
1 microsecond for me?

k88hudson: sounds good to me 🚗

as-pine-proxy:
[Treeherder: fd68c6220822ef09fd4343ec9ecd6427b9f8126d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=fd68c6220822ef09fd4343ec9ecd6427b9f8126d)

as-pine-proxy:
[Treeherder: 06bb0dd6d1028e2bb61ad4f9a656e5773ce885ad](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=06bb0dd6d1028e2bb61ad4f9a656e5773ce885ad)

PR #3759, opened 2017-10-24
fix(snippets): do not redirect to about:accounts

eoger:
This supports [Bugzilla #1395460](https://bugzilla.mozilla.org/show_bug.cgi?id=1395460).
TLDR: we are going to get rid of about:accounts, and clients should link to the FxA URLs directly.

Note: Do not merge until the Firefox changes are merged onto mozilla-central.

### DIFF
--- a/system-addon/lib/SnippetsFeed.jsm
+++ b/system-addon/lib/SnippetsFeed.jsm
@@ -13,6 +13,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "ShellService",
   "resource:///modules/ShellService.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge",
   "resource://gre/modules/ProfileAge.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "fxAccounts",
+  "resource://gre/modules/FxAccounts.jsm");
 
 // Url to fetch snippets, in the urlFormatter service format.
 const SNIPPETS_URL_PREF = "browser.aboutHomeSnippets.updateUrl";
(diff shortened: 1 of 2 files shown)


system-addon/lib/SnippetsFeed.jsm
```diff
@@ -138,7 +141,7 @@ this.SnippetsFeed = class SnippetsFeed {
         this.uninit();
         break;
       case at.SHOW_FIREFOX_ACCOUNTS:
-        this.showFirefoxAccounts(action._target.browser);
+        await this.showFirefoxAccounts(action._target.browser);
```
k88hudson: Don't `await` here,  just calling`showFirefoxAccounts`is sufficient

eoger: I actually need to await on that for testing reasons :/

piatra: Your stub doesn't have to return a `Promise`. Just do `return url;` and test without the await.

eoger: This will not work, `await` does not behave in a synchronous way with non-promises.  

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await): 
> If the value of the expression following the await operator is not a Promise, it's converted to a resolved Promise.

eoger:
@k88hudson Updated PR!

k88hudson:
Thanks!

as-pine-proxy:
[Treeherder: 96348e2bfd293610da6cf331e4de24c6adca55b6](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=96348e2bfd293610da6cf331e4de24c6adca55b6)

eoger:
Thank you! 🙂

PR #3771, opened 2017-10-26
fix(prerender): Package localized files in a locales/jar.mn and use them via chrome:

### LINKED ISSUE #3758: Actually build localized prerendered files
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1412930
Looks like the jar.mn changes from #3740 were not enough. I see there's a `browser/locales/Makefile.in`: https://searchfox.org/mozilla-central/source/browser/locales/Makefile.in

https://firefox-source-docs.mozilla.org/build/buildsystem/locales.html
https://developer.mozilla.org/en-US/docs/Mozilla/Creating_a_language_pack#L10n_binary_repack

> also, try works for l10n. you can ./mach try fuzzy --full, and search for 'repackage-l10n-it and trigger the platforms you're interested in

Mardak:
Somewhat depends on Pike's review of the locale bits: https://reviewboard.mozilla.org/r/193616/diff/1#index_header

Fix #3758. r? @k88hudson Mainly moves locale stuff to `locales/jar.mn`. I suppose it makes the generated file not so verbose. ;)

Steps to build and repack (can't be artifact): https://bugzilla.mozilla.org/show_bug.cgi?id=1411452#c5


Mardak:
Looks like we might need some platform changes to support per-locale scripts. I proposed an alternative: https://bugzilla.mozilla.org/show_bug.cgi?id=1411452#c8

Mardak:
Different approach in #3775.

PR #3773, opened 2017-10-27
Skip failed Screenshot requests.

### LINKED ISSUE #3763: Failed thumbnails get requested every 5 minutes
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1411637

#3695 made things refresh every 5 minutes, but now if a screenshot fails, we end up retrying as the promise has resolved but there's no screenshot.

piatra:
Closes #3763.

@Mardak would this be a good enough fix? Would prevent per browser session repeated requests while still allow to make the request on restart for websites that might have legitimate issues.

I've looked into the Screenshots/BackgroundPageThumbs code but it doesn't give more information in case of error to make this more precise. `page-thumbnail:error` is a generic fail.
> fails (eg, times out, remote process crashes.)

### DIFF
--- a/system-addon/lib/Screenshots.jsm
+++ b/system-addon/lib/Screenshots.jsm
@@ -73,9 +73,10 @@ this.Screenshots = {
    @ @param onScreenshot {function} Callback for when the screenshot loads
    */
   async maybeCacheScreenshot(link, url, property, onScreenshot) {
-    // Nothing to do if we already have a pending screenshot
+    // Nothing to do if we already have a pending screenshot or
+    // if a previous request failed and returned null.
     const cache = link.__sharedCache;
-    if (cache.fetchingScreenshot) {
+    if (cache.fetchingScreenshot || link[property] !== undefined) {
       return;
     }
 
@@ -87,9 +88,7 @@ this.Screenshots = {
     delete cache.fetchingScreenshot;
 
     // Update the cache for future links and call back for existing content
-    if (screenshot) {
-      cache.updateLink(property, screenshot);
-      onScreenshot(screenshot);
-    }
+    cache.updateLink(property, screenshot);
+    onScreenshot(screenshot);
   }
 };
--- a/system-addon/test/unit/lib/Screenshots.test.js
+++ b/system-addon/test/unit/lib/Screenshots.test.js
@@ -52,7 +52,7 @@ describe("Screenshots", () => {
   describe("#maybeCacheScreenshot", () => {
     let link;
     beforeEach(() => {
-      link = {__sharedCache: {updateLink: sinon.stub()}};
+      link = {__sharedCache: {updateLink: (prop, val) => { link[prop] = val; }}};
     });
     it("should call getScreenshotForURL", () => {
       sandbox.stub(Screenshots, "getScreenshotForURL");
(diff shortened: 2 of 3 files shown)


piatra:
@Mardak is that enough? Even if the `null` value is cached, won't it still go out and make a call:

https://github.com/mozilla/activity-stream/blob/657a05d5aad13d2d0da13522da9c1f9aa5ac71db/system-addon/lib/HighlightsFeed.jsm#L111-L113

Mardak:
Yup, you'll also need to short circuit https://github.com/mozilla/activity-stream/blob/c67fa77c39a96997d9e3648a75759ec8756181ad/system-addon/lib/Screenshots.jsm#L78

piatra:
@Mardak updated.

system-addon/lib/Screenshots.jsm
```diff
@@ -73,9 +73,10 @@ this.Screenshots = {
    @ @param onScreenshot {function} Callback for when the screenshot loads
    */
   async maybeCacheScreenshot(link, url, property, onScreenshot) {
-    // Nothing to do if we already have a pending screenshot
+    // Nothing to do if we already have a pending screenshot or
+    // if a previous request failed and returned null.
     const cache = link.__sharedCache;
-    if (cache.fetchingScreenshot) {
+    if (cache.fetchingScreenshot || link[property] !== undefined) {
```
Mardak: I was thinking `=== null` but this is also a good sanity check if the link already has a screenshot, so it probably shouldn't fetch either.

piatra:
@Mardak added the tests.

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -49,6 +49,34 @@ describe("Screenshots", () => {
...
+      sandbox.stub(Screenshots, "getScreenshotForURL");
+    });
+    it("should call getScreenshotForURL", () => {
+      Screenshots.maybeCacheScreenshot(link, "mozilla.com", "image", sinon.stub());
+
+      assert.calledOnce(Screenshots.getScreenshotForURL);
+      assert.calledWithExactly(Screenshots.getScreenshotForURL, "mozilla.com");
+    });
+    it("should not call getScreenshotForURL if a fetch is in progress", () => {
+      link.__sharedCache.fetchingScreenshot = sinon.stub();
```
Mardak: You should be able to test this without reaching into the details with two calls to `maybeCacheScreenshot` of the same `link`

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -49,6 +49,34 @@ describe("Screenshots", () => {
...
+
+      Screenshots.maybeCacheScreenshot(link, "mozilla.com", "image", sinon.stub());
+
+      assert.notCalled(Screenshots.getScreenshotForURL);
+    });
+    it("should not call getScreenshotsForURL if property !== undefined", () => {
+      link.image = null;
```
Mardak: Similarly, you can test the actual bug instead of testing the implementation detail by having `getScreenshotForURL` resolve `null` and calling `maybeCacheScreenshot` a second time.

piatra: Wouldn't that (testing by calling maybeCacheScreenshot 2 times) require a proper implementation of `updateLink`that actually sets the `image` property?

piatra: @Mardak the comment got collapsed but this still needs to be addressed.

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -49,6 +49,36 @@ describe("Screenshots", () => {
     });
   });
 
+  describe("#maybeCacheScreenshot", () => {
+    let link;
+    beforeEach(() => {
+      link = {__sharedCache: {updateLink: sinon.stub()}};
```
Mardak: You can get the link to update with `updateLink(prop, val) { link[prop] = val; }`

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -49,6 +49,36 @@ describe("Screenshots", () => {
...
+    });
+    it("should call getScreenshotForURL", () => {
+      sandbox.stub(Screenshots, "getScreenshotForURL");
+      Screenshots.maybeCacheScreenshot(link, "mozilla.com", "image", sinon.stub());
+
+      assert.calledOnce(Screenshots.getScreenshotForURL);
+      assert.calledWithExactly(Screenshots.getScreenshotForURL, "mozilla.com");
+    });
+    it("should not call getScreenshotForURL twice if a fetch is in progress", () => {
+      sandbox.stub(Screenshots, "getScreenshotForURL").callsFake(() => new Promise(resolve => setTimeout(resolve, 0)));
```
Mardak: nit: `.returns(new Promise(() => {}));`

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -49,6 +49,36 @@ describe("Screenshots", () => {
...
+      Screenshots.maybeCacheScreenshot(link, "mozilla.com", "image", sinon.stub());
+      Screenshots.maybeCacheScreenshot(link, "mozilla.org", "image", sinon.stub());
+
+      assert.calledOnce(Screenshots.getScreenshotForURL);
+      assert.calledWithExactly(Screenshots.getScreenshotForURL, "mozilla.com");
+    });
+    it("should not call getScreenshotsForURL if property !== undefined", () => {
+      sandbox.stub(Screenshots, "getScreenshotForURL");
+      link.image = null;
```
Mardak: This should look just like the previous test except with `.returns(Promise.resolve(null));`

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -49,6 +49,36 @@ describe("Screenshots", () => {
...
+    });
+    it("should call getScreenshotForURL", () => {
+      sandbox.stub(Screenshots, "getScreenshotForURL");
+      Screenshots.maybeCacheScreenshot(link, "mozilla.com", "image", sinon.stub());
+
+      assert.calledOnce(Screenshots.getScreenshotForURL);
+      assert.calledWithExactly(Screenshots.getScreenshotForURL, "mozilla.com");
+    });
+    it("should not call getScreenshotForURL twice if a fetch is in progress", () => {
+      sandbox.stub(Screenshots, "getScreenshotForURL").callsFake(() => new Promise(() => {}));
```
Mardak: My comment from before was `nit: .returns(new Promise(() => {}));`. Note the `.returns` instead of `.callsFake`. But this is fine…

piatra: I missed that. Fixed now.

as-pine-proxy:
[Treeherder: 47dcccf4568b63c07a4dc7bc713399aaf548d7fa](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=47dcccf4568b63c07a4dc7bc713399aaf548d7fa)

PR #3778, opened 2017-10-31
fix(init): Remember when rehydrating early tabs to only do it once

### LINKED ISSUE #3747: First about:home gets rehydrated twice
Noticed this when investigating #3741. Looks like main sends INIT while content mounts `Base` triggering NEW_TAB_STATE_REQUEST then content sees INIT come by to resend NEW_TAB_STATE_REQUEST. Turns out main already has strings and immediately responds with NEW_TAB_INITIAL_STATE to each REQUEST.

Should be mostly harmless unless messages can arrive out of order? (Pretty sure they stay in order.)

For potential fixes, it seems that we can't do much about the parallel main-to-content INIT and content-to-main REQUEST, and content has no idea (?) if the initial REQUEST was never received. So either main remembers that it already responded to a target and skips or the target content doesn't rehydrate twice? @k88hudson

Mardak:
Fix #3747. r?@k88hudson Special case early tabs by marking them as `simulated` and only `reply`ing once.

### DIFF
--- a/system-addon/lib/ActivityStreamMessageChannel.jsm
+++ b/system-addon/lib/ActivityStreamMessageChannel.jsm
@@ -142,7 +142,7 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
   simulateMessagesForExistingTabs() {
     // Some pages might have already loaded, so we won't get the usual message
     for (const target of this.channel.messagePorts) {
-      const simulatedMsg = {target};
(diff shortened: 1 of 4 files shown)


as-pine-proxy:
[Treeherder: 702e29168046fd34f1411879b476afe8fd92286a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=702e29168046fd34f1411879b476afe8fd92286a)

PR #3781, opened 2017-10-31
fix(highlights): Use appropriate short url fallbacks

### LINKED ISSUE #3460: Top site shows "http://" protocol and "H" letter fallback
In today's Nightly... I was trying to get React Developer Tools working locally and had http://localhost:8097 in my history, and noticed I had a top site for "http://localhost:...", which seemed odd:

new_tab

### LINKED ISSUE #3716: Card hostname can overflow
If [ShortURL.jsm](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/ShortURL.jsm) can't process the hostname of a url properly it falls back to using the page title. If the page title is too long then the hostname of the highlights card overflows as shown here.

screen shot 2017-10-16 at 1 42 05 am

For reference, the url is [this](https://www.gov.uk/countersigning-passport-applications/accepted-occupations-for-countersignatories).

This issue is probably quite niche as it (presumably) has to be an unusual tld for the hostname to not be parsed properly (in this case `gov.uk`), and the page title has to be long enough.

One other issue the fallback causes is that highlights are no longer deduped by hostname, so all nine cards end up being from the same host, though that's less of a problem.

Probably a straightforward fix to limit the hostname to one line? Though I'm not sure what would happen if a domain itself happened to be really long.

Mardak:
Fix #3716 and fix #3460. Fall back to hostname and clean up the code a little. (Remove duplicate `newURI`/`URL`, lowercase, and only use url-related fallbacks for `shortURL`.)

before/after:
screen shot 2017-10-31 at 4 26 06 pm

### DIFF
--- a/system-addon/lib/ShortURL.jsm
+++ b/system-addon/lib/ShortURL.jsm
@@ -22,12 +22,13 @@ function handleIDNHost(hostname) {
 }
 
 /**
- * Returns the public suffix of a URL or empty string in case of error.
- * @param {string} url The url to be analyzed.
+ * Get the effective top level domain of a host.
+ * @param {string} host The host to be analyzed.
+ * @return {str} The suffix or empty string if there's no suffix.
  */
-function getETLD(url) {
+function getETLD(host) {
   try {
-    return Services.eTLD.getPublicSuffix(Services.io.newURI(url));
+    return Services.eTLD.getPublicSuffixFromHost(host);
   } catch (err) {
     return "";
   }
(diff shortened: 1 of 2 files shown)


piatra:
oh haha
```
Services.eTLD.getPublicSuffixFromHost("http://gov.uk")
"uk"
Services.eTLD.getPublicSuffixFromHost("http://www.gov.uk")
"gov.uk"
```
Do you want to strip the www part before and fix this?

Mardak:
It's because the eTLD is "gov.uk" for "www.gov.uk". You're right that "uk" is also a eTLD. Also, `FromHost` methods just want the host not a url.
```
Services.eTLD.getPublicSuffixFromHost("gov.uk")
gov.uk
```

Mardak:
Services.eTLD.getPublicSuffix(Services.io.newURI("http://gov.uk"))
gov.uk

as-pine-proxy:
[Treeherder: 084d04d16b51b7f783f9bde91f3f53e31cf1e800](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=084d04d16b51b7f783f9bde91f3f53e31cf1e800)

Issue #3745, opened 2017-10-19
Use a 1px border width for the search field on Windows only


aaronrbenson:
We should match the native OS when styling the search field on New Tab. The focused state should look like the attached:

![screen shot 2017-10-19 at 3 42 16 pm](https://user-images.githubusercontent.com/139451/31793193-30f20e92-b4e4-11e7-9f04-159c4e82e188.png)

Related, I've filed a bug to change the placeholder text on Windows to non-italicized: https://bugzilla.mozilla.org/show_bug.cgi?id=1410233

Mardak:
So it should be 2px on mac and 1px on windows?

aaronrbenson:
Is it 2px on mac now? Seems a bit thicker but I could be wrong ..

aaronrbenson:
I guess it doesn't matter .. the current state of macOS looks correct.

Mardak:
Oh indeed. It's currently 1px border with 2px box-shadow, so a total 3px.

aaronrbenson:
Hey @Mardak, was there anything else needed here for design feedback?

Mardak:
Sounds like this should be good. Just to summarize: change total border width to 1px for windows and leave unchanged elsewhere.

PR #3775, opened 2017-10-31
fix(prerender): Package all localized prerendered files

Mardak:
Fix #3758. r?@k88hudson Moves the `.html` and `.js` files to a `prerendered/` directory.
The generated m-c result: https://reviewboard.mozilla.org/r/194760/diff/1#index_header

PR #3784, opened 2017-11-01
style(content): #3745 Add per-platform css files

k88hudson:
Includes a windows-specific search focus border style

### DIFF
--- a/system-addon/jar.mn
+++ b/system-addon/jar.mn
@@ -17,6 +17,13 @@
   content/vendor/redux.js (./vendor/redux.js)
   content/vendor/react-redux.js (./vendor/react-redux.js)
   content/data/ (./data/*)
+#ifdef XP_MACOSX
+  content/css/activity-stream.css (./css/activity-stream-mac.css)
+#elifdef XP_WIN
+  content/css/activity-stream.css (./css/activity-stream-windows.css)
+#else
+  content/css/activity-stream.css (./css/activity-stream-linux.css)
+#endif
 #ifndef RELEASE_OR_BETA
   content/data/content/activity-stream-debug.html (./prerendered/en-US/activity-stream-debug.html)
   content/data/content/activity-stream-prerendered-debug.html (./prerendered/en-US/activity-stream-prerendered-debug.html)
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,8 @@ dist/
 firefox/
 *.update.rdf
 system-addon/data/content/activity-stream.bundle.js
-system-addon/data/content/activity-stream.css
 system-addon/data/locales.json
+system-addon/css/
 system-addon/prerendered/
 bin/prerender.js
 bin/prerender.js.map
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -209,7 +209,7 @@ function templateHTML(options, html) {
     <title>${options.title}</title>
     <link rel="icon" type="image/png" id="favicon" href="chrome://branding/content/icon32.png"/>
     <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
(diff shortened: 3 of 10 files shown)


system-addon/jar.mn
```diff
@@ -17,6 +17,13 @@
   content/vendor/redux.js (./vendor/redux.js)
   content/vendor/react-redux.js (./vendor/react-redux.js)
   content/data/ (./data/*)
+#ifdef XP_MACOSX
+  content/data/content/activity-stream.css (./data/content/activity-stream-mac.css)
```
Mardak: The `content/data/ (./data/*)` line is going to package all of them, so most likely leading to unreferenced file test failure. Do we make a new "top level" directory outside of data? `css/` ?

system-addon/content-src/activity-stream-windows.scss
```diff
@@ -0,0 +1,7 @@
+/* This is the windows variant */
```
Mardak: Just to be explicit, this comment *block* will appear in the generated file. Which I suppose would be handy for debugging?

system-addon/jar.mn
Mardak: I suppose you can create a new directory in this PR, and I'll update it in https://github.com/mozilla/activity-stream/pull/3755 if we don't want to put css in `prerendered/static/activity-stream-mac.css`.

Alternatively, the only thing inside `data/` will be `data/content/` after I remove `locales.json` from the other PR #3755. So we could have a `data/css/*.css` that we jar.mn package appropriately… and now that only activity-stream.bundle.js is the only single file:
```
  content/data/content/activity-stream.bundle.js (./data/content/activity-stream.bundle.js)
  content/data/content/assets/ (./data/content/assets/*)
  content/data/content/tippytop/ (./data/content/tippytop/*)
#ifdef …
  content/data/content/activity-stream.css (./data/css/activity-stream-mac.css)
```

And I suppose with that change, I could move prerendered to `data/prerendered/`.

(I suppose we could move things "up" a directory, i.e., ./data/{activity-stream.bundle.js,assets,css,tippytop,prerendered} if we're going to CLOBBER anyway although history might be broken. Separately, we could expose the `resource://activity-stream/data/content/*` as just `resource://activity-stream/content/*`…)

Mardak: On a related note, should I be updating https://github.com/mozilla/activity-stream/pull/3775 (wow #3775 = package all locales vs #3755 = remove locales.json) and the related mozreview patch to use `resource://activity-stream/data/content/prerendered/locales/en-US` instead of `resource://activity-stream/prerendered/locales/en-US` ? if we move things to `./data/prerendered` ?

system-addon/content-src/activity-stream-windows.scss
k88hudson: Yep, that's the idea

system-addon/content-src/activity-stream-windows.scss
```diff
@@ -0,0 +1,7 @@
...
+@import './activity-stream';
+
+// Search border should only have a 1-pixel blue border, so no box-shadow is needed.
+.search-wrapper input:focus {
+  box-shadow: none;
```
Mardak: The change from #3785 removes the border, so we'll have to fix this a different way.

system-addon/jar.mn
k88hudson: I like the idea of moving stuff to top-level directories (content/data/content was kind of weird anyway)

k88hudson:
Ok, I moved everything to a `css` directory and updated relevant paths. What do you think?

as-pine-proxy:
[Treeherder: cd0862e479f2c171a0a0f64d6413decea96cab37](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=cd0862e479f2c171a0a0f64d6413decea96cab37)

Issue #3679, opened 2017-10-10
The Search bar border around the search arrow is wrongly darker when the arrow is highlighted


SoftVision-CiprianMuresan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1407244

[Affected versions]:
- Firefox 58.0a1 Build ID 20171009220104

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Start the browser with the profile from prerequisites and open a New Tab.
2. Hover the Search bar arrow.
3. Observe the border around the arrow.

[Expected results]:
- The border has the same color as the rest of the Search bar border.

[Actual results]:
- The border around the Search arrow is slightly darker.

[Notes]:
- Attached a screen shot of the issue:
![search border is also highlighted](https://user-images.githubusercontent.com/18304511/31391389-0f2a18c4-addf-11e7-8dfe-3bc86075edfc.png)

Mardak:
The address bar has a hover state that darkens the whole box:

no hover:
<img width="818" alt="image" src="https://user-images.githubusercontent.com/438537/31415023-a87256ea-add5-11e7-9fef-03f350063311.png">

box hover:
<img width="819" alt="image" src="https://user-images.githubusercontent.com/438537/31415031-b1949e72-add5-11e7-8ba8-ed1702eb6246.png">

button hover:
<img width="821" alt="image" src="https://user-images.githubusercontent.com/438537/31415044-c5d69d18-add5-11e7-9435-943ea66840aa.png">

aaronrbenson:
It looks like the hover background extends into the border area making it look like the border is darker? Can we put some margin on the bg so it doesn't look that way @Mardak?

Mardak:
Clarified on slack that the border near the arrow should not be darker than the rest. And that the whole box's border should get darker on hover like the address bar.

sarracini:
Fixed with #3785

PR #3785, opened 2017-11-01
style(search): #3679 Fix search hover border styles

k88hudson:
This patch makes the border a consistent color when hovering over the submit button, and adds a slightly darker border color when hovering over the input.

### DIFF
--- a/system-addon/content-src/components/Search/_Search.scss
+++ b/system-addon/content-src/components/Search/_Search.scss
@@ -1,7 +1,7 @@
 .search-wrapper {
   $search-border-radius: 3px;
   $search-focus-color: $blue-50;
-  $search-height: 36px;
+  $search-height: 35px;
   $search-input-left-label-width: 35px;
   $search-button-width: 36px;
   $search-glyph-image: url('chrome://browser/skin/search-glass.svg');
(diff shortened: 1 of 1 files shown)


system-addon/content-src/components/Search/_Search.scss
```diff
@@ -60,7 +64,17 @@
...
     &:focus,
     &:hover {
-      background-color: $grey-90-10;
+      &::before {
+        content: "";
+        position: absolute;
+        top: 0;
+        offset-inline-start: 0;
+        width: 100%;
+        height: 100%;
```
Mardak: nit: Might as well just do top/right/bottom/left:0 in case it seems like this is rtl-special.

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -60,7 +64,17 @@
...
+        border-radius: 0 $border-radius $border-radius 0;
```
Mardak: nit: Let's share this inside `.search-button` (?) as a variable

Mardak: Or.. if we have a slightly smaller radius, will the whites in the corner be filled in? !

Mardak: Oh indeed! With 2px. ! ? !
screen shot 2017-11-01 at 2 44 05 pm

k88hudson:
Ok, I thought of a better way to do this that would prevent the gap and is slightly less hacky (basically changing the border/box-shadow combo to box-shadow only)

as-pine-proxy:
[Treeherder: 0cbe2f11e023c866f1bda49baeeb4274b9505867](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0cbe2f11e023c866f1bda49baeeb4274b9505867)

Issue #3749, opened 2017-10-20
The New Tab Preferences menu's width doesn't adapt to the window's width


IuliaCristescu4:
https://bugzilla.mozilla.org/show_bug.cgi?id=1410450

[Note]:
- The New Tab Preferences menu is not entirely visible at minimum window width 

[Affected versions]:
- 58.0a1 (2017-10-20)
- 57.0b10 build1 (20171019140425)

[Affected platforms]:
- Windows 10 x64
- Ubuntu 16.04 x64
- Mac OS X 10.11.6

[Steps to reproduce]:
1. Launch Firefox using a clean profile
2. Access the AS New Tab page and open the New Tab Preferences menu
3. Resize the browser window to minimum width 

[Expected result]:
- The New Tab Preferences menu adapt to the window's width so that all its elements are visible 

[Actual result]:
- The New Tab Preferences menu is not entirely visible

bryanbell:
Surely this can be better, but we should recognize that the browser itself isn't useful at that size. Arguably the browser min-width should be adjusted wider. 

![narrow-browser](https://user-images.githubusercontent.com/228979/32012277-2cc76164-b96c-11e7-9d85-29d825721b2c.gif)

Besides this whole menu is slated to be removed and relocated into the application preferences. Its inclusion on this page was always a temporary move. 

let's not fix, and instead work on its replacement.

Issue #3273, opened 2017-08-29
Move line separators in preferences and differently style confirmation


Mardak:
See https://mozilla.invisionapp.com/share/NYDAMWECB#/251367746_Preferences_Spec

Between search / top sites
<img width="353" alt="image" src="https://user-images.githubusercontent.com/438537/29804191-df695c64-8c35-11e7-8b5e-6be6c9121152.png">

and between highlights / snippets
<img width="364" alt="image" src="https://user-images.githubusercontent.com/438537/29804199-e3a647b0-8c35-11e7-8fbc-9c17be3fc796.png">

and background of "Done"
<img width="423" alt="image" src="https://user-images.githubusercontent.com/438537/29806019-c44eef1e-8c41-11e7-8801-72fd9f4a8345.png">

and lighter intro text:
<img width="341" alt="image" src="https://user-images.githubusercontent.com/438537/29884682-09f22b76-8d69-11e7-959b-71f2140947f0.png">

aaronrbenson:
More details for updated styles to the prefs panel: https://mozilla.invisionapp.com/share/NYDAMWECB#/251367746_Preferences_Spec

Mardak:
@aaronrbenson is the done button area anchored to the bottom or just under the last item? And if the window is too small, what is scrolling?

aaronrbenson:
My mistake, left that part and the spacing out! Updated the spec above.

PR #3786, opened 2017-11-01
Use flex-wrap to force a new line for small width.

### LINKED ISSUE #3630: Edit top sites "Done" button is displayed outside the parent window on narrow windows
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1404374

image

piatra:
Closes #3630 

image

Quickest fix. Could use some margins but this seems to obscure to write media queries for. And we've closed quite a few issues recently where we decided this width doesn't make too much sense to support. https://github.com/mozilla/activity-stream/issues/3749#issuecomment-339401334

### DIFF
--- a/system-addon/content-src/components/ManualMigration/_ManualMigration.scss
+++ b/system-addon/content-src/components/ManualMigration/_ManualMigration.scss
@@ -44,6 +44,7 @@
   button {
     align-self: center;
     height: 26px;
+    margin: 0;
     margin-inline-start: 20px;
     padding: 0 12px;
   }
--- a/system-addon/content-src/components/PreferencesPane/_PreferencesPane.scss
+++ b/system-addon/content-src/components/PreferencesPane/_PreferencesPane.scss
@@ -115,7 +115,6 @@
     border-left: $border-secondary;
     bottom: 0;
     offset-inline-end: 0;
-    padding: $prefs-spacing;
     position: fixed;
     width: $prefs-width;
 
--- a/system-addon/content-src/styles/_activity-stream.scss
+++ b/system-addon/content-src/styles/_activity-stream.scss
@@ -79,8 +79,9 @@ a {
   display: flex;
   flex-direction: row;
   margin: 0;
-  padding: 15px 25px;
+  padding: 15px 25px 0 25px;
   justify-content: flex-start;
+  flex-wrap: wrap;
 
   button {
     background: $input-secondary;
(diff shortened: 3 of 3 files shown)


Mardak:
Adding a `margin-bottom` to the button and removing the `padding-bottom` for `.actions` seems simple enough and looks much nicer than having the buttons touching:

@aaronrbenson ?

no margin
image

with margin
image

piatra:
Made the margin changes only apply to the modal component because we use `.actions` for the migration message and the preference panel done button.

Mardak:
Manual migration stuff is a pretty special case, so it overriding with a `padding: 0` and `margin: 0` is probably fine.

The prefs stuff didn't explicitly specify a 20px padding other than that's how much spacing was shown in the design: https://github.com/mozilla/activity-stream/issues/3273#issuecomment-326413878

I don't think it'll be a big deal to make `.actions` consistent with 15px above/below the buttons.

Otherwise with the current PR, the details of the spacing from generic `.actions` is unnecessarily duplicated to `ConfirmDialog` quite likely leading to bugs if the top level style changes.

Mardak:
Manual migration stuff is a pretty special case, so it overriding with a `padding: 0` and `margin: 0` is probably fine.

Edit: ~~The prefs stuff didn't explicitly specify a 20px padding other than that's how much spacing was shown in the design:~~ https://github.com/mozilla/activity-stream/issues/3273#issuecomment-326413878
@aaronrbenson Would it be okay to have the pref buttons be 15px spacing around to be consistent with all the other action buttons?
image

Otherwise with the current PR, the details of the spacing from generic `.actions` is unnecessarily duplicated to `ConfirmDialog` quite likely leading to bugs if the top level style changes.

piatra:
I have that patch ready, waiting for the 👍

sarracini:
If you add the following media query to the .done class in ```activity-stream.scss``` you can get it to look good:
```css
    &.done {
      background: $input-primary;
      border: solid 1px $blue-60;
      color: $white;
      margin-inline-start: auto;

      @media (max-width: $break-point-small) {
        margin-inline-end: 15px;
      }
    }
```
screen shot 2017-11-02 at 11 17 40 am

Mardak:
The right edge of the top row of buttons will depend on the translated text

piatra:
@sarracini  maybe change that to align with the TopSites then?

image

sarracini:
@Mardak like this? 
screen shot 2017-11-02 at 1 06 05 pm

as-pine-proxy:
[Treeherder: 368646815a5cfd35680f716ebe74ff36d0d837c9](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=368646815a5cfd35680f716ebe74ff36d0d837c9)

Issue #3803, opened 2017-11-08
CollapsibleSection URLs that come from extensions may not work


dmose:
This was turned up by a React warning on a test.  From reading the code, I _think_ the consequence of the warning is that CollapsibleSection URLs that come from extensions may not work. PR forthcoming.

PR #3804, opened 2017-11-08
fix(addons): #3803 CollapsibleSection icons from add-ons appear to fail

dmose:
How to test: need to actually write a webextension that does tries to do this to see if this is the consequence of the test warning that was fixed.  That said, it looks pretty straightforward...

r? @rlr


as-pine-proxy:
[Treeherder: 8891415b2a8b2f27160cea265b15b5fe62f4159d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8891415b2a8b2f27160cea265b15b5fe62f4159d)

PR #3812, opened 2017-11-09
Add platform specific styles.

### LINKED ISSUE #3792: Use appropriate platform-specific panelarrow image and sizing
With #3784, we now have per-platform css, so the double-background-image from #3644 to get an available icon isn't necessary as we can just directly reference the icon for that platform.

This also means we can get rid of the `parsable_css` whitelist entry:
https://searchfox.org/mozilla-central/source/browser/base/content/test/static/browser_parsable_css.js#91-99

Also, turns out the size of the image is slightly different depending on the platform.

piatra:
Closes #3792.

OSX.
screen shot 2017-11-09 at 15 38 06111
screen shot 2017-11-09 at 15 37 54423

Linux.
screen shot 2017-11-09 at 15 37 38887
screen shot 2017-11-09 at 15 37 21804

I don't have access to a windows machine with a build environment set up but the output is identical to linux.

RTL OSX

screen shot 2017-11-09 at 15 41 48989

And I guess a follow up bugzilla bug to remove https://searchfox.org/mozilla-central/source/browser/base/content/test/static/browser_parsable_css.js#91-99.
> https://bugzilla.mozilla.org/show_bug.cgi?id=1415892

When looking at these it seems to me that the image background and panel background are slightly different. Thoughts?

screen shot 2017-11-09 at 15 42 49920


system-addon/content-src/styles/activity-stream-linux.scss
```diff
@@ -1,2 +1,20 @@
 /* This is the linux variant */
 @import './activity-stream';
+
+// Custom image and size of the arrow from the (?) panel that shows more info
+// about a section.
+.section-info-option .info-option {
+  $arrow-height: 10px;
```
Mardak: Duplicating these 3x and away from where the rest of it's defined is going to make this quite hard to maintain. I think we should have a different pattern of something like:
```css
$os-info-arrow-height: 10px;
…
@import './activity-stream'; 
```

system-addon/content-src/styles/activity-stream-windows.scss
```diff
@@ -5,3 +5,21 @@
 .search-wrapper input:focus {
   box-shadow: 0 0 0 1px $blue-50;
```
Mardak: Similarly, can you make this `$os-search-focus-shadow-radius: 1px` and set it to `3px` for the other platforms?

system-addon/content-src/styles/activity-stream-linux.scss
piatra: @Mardak if the platform-specific-css is to go where the `...` are then they will not be able to take advantage of the variables defined in _variables.scss.

piatra: I guess we can import them again at the top of each platform specific file.

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -32,6 +32,11 @@
     padding-inline-start: $search-input-left-label-width;
     width: 100%;
     font-size: 15px;
+
+    // Search border should only have a 1-pixel blue border, so no box-shadow is needed.
+    &:focus {
+      box-shadow: 0 0 0 $os-search-focus-shadow-radius $blue-50;
```
Mardak: There should be an existing style that this was trying to override. Update that instead of adding a new one.

system-addon/content-src/styles/activity-stream-linux.scss
```diff
@@ -1,2 +1,8 @@
 /* This is the linux variant */
+
+$os-arrow-offset-end: 6px;
+$os-arrow-width: 20px;
+$os-search-focus-shadow-radius: 3px;
+$os-section-info-option-image: 'chrome://global/skin/arrow/panelarrow-vertical-themed.svg';
```
Mardak: The arrow stuff is directly referring to the panelarrow. These variable names make it seem like they're unrelated.

system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
```diff
@@ -55,8 +55,6 @@
 
     .section-info-option .info-option {
       $arrow-height: 10px;
```
Mardak: This happens to be 10px for all platforms, but let's keep it next to where the other `$os-info-arrow-` variables

system-addon/content-src/styles/activity-stream-linux.scss
```diff
@@ -1,2 +1,9 @@
 /* This is the linux variant */
+
+$os-arrow-height: 10px;
+$os-arrow-offset-end: 6px;
+$os-arrow-panel-url: 'chrome://global/skin/arrow/panelarrow-vertical-themed.svg';
+$os-arrow-width: 20px;
```
Mardak: If "os-arrow" is the scope/context, a "panel-url" doesn't really make sense? Also, "os-arrow" is a bit generic for this top level scope where these specific values aren't useful for other possible "os-arrow"s.

piatra:
Updated.

as-pine-proxy:
[Treeherder: 77edfd35cb884728d7f9cfe762f3626b3071bff8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=77edfd35cb884728d7f9cfe762f3626b3071bff8)

PR #3815, opened 2017-11-09
fix(tests): closes #3814, ConfirmDialog propType errors

### LINKED ISSUE #3814: fix ConfirmDialog.test.jsx propType errors
This is in preparation for making propType errors in the tests cause the tests to fail.

dmose:
This is in preparation for making all React warnings fail in order to have more confidence in the React 16 upgrade.  

How to test: both without and with the patch, do this:

`npm run testmc | grep FormattedMessage`

with the patch, there should be no matches.

### DIFF
--- a/system-addon/test/unit/content-src/components/ConfirmDialog.test.jsx
+++ b/system-addon/test/unit/content-src/components/ConfirmDialog.test.jsx
@@ -12,7 +12,11 @@ describe("<ConfirmDialog>", () => {
     dispatch = sinon.stub();
     ConfirmDialogProps = {
       visible: true,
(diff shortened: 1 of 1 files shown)


dmose:
Mardak, that's true.  Interestingly, the failures only seem to show up one at a time.  There were actually four errors that needed to be worked through.

as-pine-proxy:
[Treeherder: 4aa5960ae15e2533b5cfcad1edd0c55621e2ad86](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4aa5960ae15e2533b5cfcad1edd0c55621e2ad86)

PR #3817, opened 2017-11-09
React error on key warnings during tests

dmose:
The warnings are now all gone; this patch will keep them from returning.

How to test: after applying this patch, edit a test somewhere to be missing a required prop while `npm run tddmc` is going and save the file.  The tests should break.

### DIFF
--- a/system-addon/test/unit/unit-entry.js
+++ b/system-addon/test/unit/unit-entry.js
@@ -1,6 +1,17 @@
 const {GlobalOverrider, FakePrefs, FakePerformance, EventEmitter} = require("test/unit/utils");
 const {chaiAssertions} = require("test/schemas/pings");
 
+// Cause React warnings to make tests that trigger them fail
+const origConsoleError = console.error; // eslint-disable-line no-console
+console.error = function(msg, ...args) { // eslint-disable-line no-console
+  if (/(Invalid prop|Failed prop type|Check the render method)/.test(msg)) {
+    throw new Error(msg);
+  }
+
+  // eslint-disable-next-line no-console
+  origConsoleError.apply(console, [msg, ...args]);
+};
+
 const req = require.context(".", true, /\.test\.jsx?$/);
 const files = req.keys();
 
--- a/system-addon/test/unit/unit-entry.js
+++ b/system-addon/test/unit/unit-entry.js
@@ -4,12 +4,12 @@ const {chaiAssertions} = require("test/schemas/pings");
 // Cause React warnings to make tests that trigger them fail
 const origConsoleError = console.error; // eslint-disable-line no-console
 console.error = function(msg, ...args) { // eslint-disable-line no-console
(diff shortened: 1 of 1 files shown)


system-addon/test/unit/unit-entry.js
```diff
@@ -1,6 +1,17 @@
...
 
+// Cause React warnings to make tests that trigger them fail
+const origConsoleError = console.error; // eslint-disable-line no-console
+console.error = function(msg, ...args) { // eslint-disable-line no-console
+  if (/(Invalid prop|Failed prop type|Check the render method)/.test(msg)) {
+    throw new Error(msg);
+  }
+
+  // eslint-disable-next-line no-console
+  origConsoleError.apply(console, [msg, ...args]);
```
Mardak: nit: could just do this to avoid the triple disable line:
```js
Object.assign(console, { 
  _error: console.error, // eslint-disable-line no-console
  error(...args) {
    this._error(...args);
    if (/(Invalid prop|Failed prop type|Check the render method)/.test(args[0])) {
      throw new Error(args);
    }
  }
});
```

Also, should we always `console._error` and conditionally `throw`? Might be useful to see it "inline" where it would have appeared. Also, unclear if `msg` is sufficient or if there's interesting data in `args` (as you've originally defined) that would be lost if we only throw the first argument.

dmose: To me, the trick to avoid the triple disable line makes the code unnecessarily hard to follow, so I think i'm going to leave that as is.  That said, I think you're totally right that we should always error and conditionally throw; I'll do that.

dmose:
> I'm not familiar with what console.errors get called with 

This does not support the complete semantics of console.error, but I think it handles the most common case.  I'm inclined to start here, and only enhance if we see specific reasons to.

> in particular if only the first arg is useful in an Error?

This depends entirely on who is catching the error and what they do with it.  In this case, I believe that it's going to be mocha itself that is catching the error.  Error types are generally user-defined, I believe, so the only thing we could really do would be JSON.stringify all the stuff into one big thing to be the Error message, since Mocha is unlikely to know about any fields we define.

>  But quick tests seem to show it throwing the FormattedMessage stuff (when reverting to that)

Yeah, given that we now unconditionally log to console, I think that's good enough for a first pass, particularly since mocha doesn't seem to document any sort of stuff about doing this sort of thing.

as-pine-proxy:
[Treeherder: 1977ddf9762e75d23c7337b0354a979a0ddf3635](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1977ddf9762e75d23c7337b0354a979a0ddf3635)

as-pine-proxy:
[Treeherder: d6a4421e704afe9fcae256a7871e0eb3e2966ed9](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=d6a4421e704afe9fcae256a7871e0eb3e2966ed9)

PR #3828, opened 2017-11-11
Fix Bug 1421844 - Add implementation to Activity Stream mochitests/head.js

### LINKED ISSUE #3820: Perma-fail: debug browser_highlights_section.js | there should be 1 of [data-mochitest='highlights-card-placeholder'] found 3
@piatra this seems to be failing on linux and windows:

 FAIL | browser/extensions/activity-stream/test/functional/mochitest/browser_highlights_section.js | there should be 1 of [data-mochitest='highlights-card-placeholder'] found 3
Stack trace: chrome://mochitests/content/browser/browser/extensions/activity-stream/test/functional/mochitest/head.js:check_highlights_elements:125
https://treeherder.mozilla.org/logviewer.html#?job_id=143511024&repo=pine&lineNumber=14901

As well as a bunch of
 FAIL | browser/extensions/activity-stream/test/functional/mochitest/browser_topsites_section.js | A promise chain failed to handle a rejection: cache is undefined - stack: null Rejection date: Fri Nov 10 2017 01:04:34 GMT+0000 (UTC) - false == true
JS frame :: resource://testing-common/PromiseTestUtils.jsm :: assertNoUncaughtRejections :: line 253 
https://treeherder.mozilla.org/logviewer.html#?job_id=143511024&repo=pine&lineNumber=15536

That latter one is failing on all platforms:
mac: https://treeherder.mozilla.org/logviewer.html#?job_id=143510757&repo=pine&lineNumber=27770
win32-noe10s: https://treeherder.mozilla.org/logviewer.html#?job_id=143516756&repo=pine&lineNumber=28571
win32-e10s: https://treeherder.mozilla.org/logviewer.html#?job_id=143516761&repo=pine&lineNumber=40261
win64: https://treeherder.mozilla.org/logviewer.html#?job_id=143518254&repo=pine&lineNumber=38405

piatra:
I removed the failing assert and tests seem fine.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=f56ac6ac3f2f94a7a75308f10fb1cd86df6d4b2e&selectedJob=143776229
I did a re-trigger on bc4 previously failing.
Test still checks for the correct number of highlights bookmarks.

Closes #3820

### DIFF
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -50,7 +50,7 @@ this.TopSitesFeed = class TopSitesFeed {
   _dedupeKey(site) {
     return site && site.hostname;
   }
-  refreshDefaults(sites, options = {}) {
+  refreshDefaults(sites) {
     // Clear out the array of any previous defaults
     DEFAULT_TOP_SITES.length = 0;
 
@@ -63,11 +63,6 @@ this.TopSitesFeed = class TopSitesFeed {
         };
         site.hostname = shortURL(site);
         DEFAULT_TOP_SITES.push(site);
-      }
-
-      // Update Top Sites if the pref for default sites changes
-      if (options.broadcast) {
-        this.refresh({broadcast: true});
       }
     }
   }
@@ -286,7 +281,7 @@ this.TopSitesFeed = class TopSitesFeed {
         break;
       case at.PREF_CHANGED:
         if (action.data.name === DEFAULT_SITES_PREF) {
-          this.refreshDefaults(action.data.value, {broadcast: true});
+          this.refreshDefaults(action.data.value);
         }
         break;
       case at.PREFS_INITIAL_VALUES:
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -50,7 +50,7 @@ this.TopSitesFeed = class TopSitesFeed {
   _dedupeKey(site) {
     return site && site.hostname;
   }
-  refreshDefaults(sites) {
+  refreshDefaults(sites, options = {}) {
     // Clear out the array of any previous defaults
     DEFAULT_TOP_SITES.length = 0;
 
@@ -63,6 +63,11 @@ this.TopSitesFeed = class TopSitesFeed {
         };
         site.hostname = shortURL(site);
         DEFAULT_TOP_SITES.push(site);
+      }
+
+      // Update Top Sites if the pref for default sites changes
+      if (options.broadcast) {
+        this.refresh({broadcast: true});
       }
     }
   }
@@ -281,7 +286,7 @@ this.TopSitesFeed = class TopSitesFeed {
         break;
       case at.PREF_CHANGED:
         if (action.data.name === DEFAULT_SITES_PREF) {
-          this.refreshDefaults(action.data.value);
+          this.refreshDefaults(action.data.value, {broadcast: true});
         }
         break;
       case at.PREFS_INITIAL_VALUES:
--- a/system-addon/test/functional/mochitest/browser_highlights_section.js
+++ b/system-addon/test/functional/mochitest/browser_highlights_section.js
@@ -1,0 +1,50 @@
+"use strict";
+
+/**
+ * Helper for setup and cleanup of Highlights section tests.
+ * @param bookmarkCount Number of bookmark higlights to add
+ * @param test The test case
+ */
+function test_highlights(bookmarkCount, test) {
+  test_newtab({
+    async before({tab}) {
+      if (bookmarkCount) {
+        await addHighlightsBookmarks(bookmarkCount);
+        // Wait for HighlightsFeed to update and display the items.
+        await ContentTask.spawn(tab.linkedBrowser, null, async () => {
+          await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".card-outer:not(.placeholder)"),
+            "No highlights cards found.");
+        });
+      }
+    },
+    test,
+    async after() {
+      await clearHistoryAndBookmarks();
+    }
+  });
+}
+
+test_highlights(
+  2, // Number of highlights cards
+  function check_highlights_cards() {
+    let found = content.document.querySelectorAll(".card-outer:not(.placeholder)").length;
+    is(found, 2, "there should be 2 highlights cards");
+
+    found = content.document.querySelectorAll(".section-list .placeholder").length;
+    is(found, 1, "there should be 1 highlights placeholder");
+
+    found = content.document.querySelectorAll(".card-context-icon.icon-bookmark-added").length;
+    is(found, 2, "there should be 2 bookmark icons");
+  }
+);
+
+test_highlights(
+  1, // Number of highlights cards
+  function check_highlights_context_menu() {
+    const menuButton = content.document.querySelector(".section-list .context-menu-button");
+    // Open the menu.
+    menuButton.click();
+    const found = content.document.querySelector(".context-menu");
+    ok(found && !found.hidden, "Should find a visible context menu");
+  }
+);
--- a/system-addon/test/functional/mochitest/browser_topsites_section.js
+++ b/system-addon/test/functional/mochitest/browser_topsites_section.js
@@ -1,0 +1,55 @@
+"use strict";
+
+// Check TopSites edit modal and overlay show up.
+test_newtab(
+  // it should be able to click the topsites edit button to reveal the edit topsites modal and overlay.
+  function topsites_edit() {
+    const topsitesEditBtn = content.document.querySelector(".edit-topsites-button button");
+    topsitesEditBtn.click();
+
+    let found = content.document.querySelector(".edit-topsites");
+    ok(found && !found.hidden, "Should find a visible topsites edit menu");
+
+    found = content.document.querySelector(".modal-overlay");
+    ok(found && !found.hidden, "Should find a visible overlay");
+  }
+);
+
+// Test pin/unpin context menu options.
+test_newtab({
+  async before({pushPrefs}) {
+    // The pref for TopSites is empty by default.
+    await pushPrefs(["browser.newtabpage.activity-stream.default.sites", "https://www.youtube.com/,https://www.facebook.com/,https://www.amazon.com/,https://www.reddit.com/,https://www.wikipedia.org/,https://twitter.com/"]);
+    // Forces TopsitesFeed to refresh with the new default sites pref.
+    await clearHistoryAndBookmarks();
+  },
+  // it should pin the website when we click the first option of the topsite context menu.
+  test: async function topsites_pin_unpin() {
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".context-menu-button"),
+      "Topsite context menu not found");
+    // There are only topsites on the page, the selector with find the first topsite menu button.
+    const topsiteContextBtn = content.document.querySelector(".context-menu-button");
+    topsiteContextBtn.click();
+
+    const contextMenu = content.document.querySelector(".context-menu");
+    ok(contextMenu && !contextMenu.hidden, "Should find a visible topsite context menu");
+
+    const pinUnpinTopsiteBtn = contextMenu.querySelector(".context-menu-item a");
+    // Pin the topsite.
+    pinUnpinTopsiteBtn.click();
+
+    // Need to wait for pin action.
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".icon-pin-small"),
+      "No pinned icon found");
+
+    let pinnedIcon = content.document.querySelectorAll(".icon-pin-small").length;
+    is(pinnedIcon, 1, "should find 1 pinned topsite");
+
+    // Unpin the topsite.
+    pinUnpinTopsiteBtn.click();
+
+    // Need to wait for unpin action.
+    await ContentTaskUtils.waitForCondition(() => !content.document.querySelector(".icon-pin-small"),
+      "Topsite should be unpinned");
+  }
+});
--- a/system-addon/test/functional/mochitest/browser_topsites_section.js
+++ b/system-addon/test/functional/mochitest/browser_topsites_section.js
@@ -20,6 +20,8 @@ test_newtab({
   async before({pushPrefs}) {
     // The pref for TopSites is empty by default.
     await pushPrefs(["browser.newtabpage.activity-stream.default.sites", "https://www.youtube.com/,https://www.facebook.com/,https://www.amazon.com/,https://www.reddit.com/,https://www.wikipedia.org/,https://twitter.com/"]);
+    await pushPrefs(["browser.newtabpage.activity-stream.enabled", false]);
+    await pushPrefs(["browser.newtabpage.activity-stream.enabled", true]);
   },
   // it should pin the website when we click the first option of the topsite context menu.
   test: async function topsites_pin_unpin() {
--- a/system-addon/test/functional/mochitest/browser_topsites_section.js
+++ b/system-addon/test/functional/mochitest/browser_topsites_section.js
@@ -20,8 +20,6 @@ test_newtab({
   async before({pushPrefs}) {
     // The pref for TopSites is empty by default.
     await pushPrefs(["browser.newtabpage.activity-stream.default.sites", "https://www.youtube.com/,https://www.facebook.com/,https://www.amazon.com/,https://www.reddit.com/,https://www.wikipedia.org/,https://twitter.com/"]);
-    // Forces TopsitesFeed to refresh with the new default sites pref.
-    await clearHistoryAndBookmarks();
   },
   // it should pin the website when we click the first option of the topsite context menu.
   test: async function topsites_pin_unpin() {
--- a/system-addon/test/functional/mochitest/browser_topsites_section.js
+++ b/system-addon/test/functional/mochitest/browser_topsites_section.js
@@ -20,8 +20,9 @@ test_newtab({
   async before({pushPrefs}) {
     // The pref for TopSites is empty by default.
     await pushPrefs(["browser.newtabpage.activity-stream.default.sites", "https://www.youtube.com/,https://www.facebook.com/,https://www.amazon.com/,https://www.reddit.com/,https://www.wikipedia.org/,https://twitter.com/"]);
-    await pushPrefs(["browser.newtabpage.activity-stream.enabled", false]);
-    await pushPrefs(["browser.newtabpage.activity-stream.enabled", true]);
+    // Toggle the feed off and on as a workaround to read the new prefs.
+    await pushPrefs(["browser.newtabpage.activity-stream.feeds.topsites", false]);
+    await pushPrefs(["browser.newtabpage.activity-stream.feeds.topsites", true]);
   },
   // it should pin the website when we click the first option of the topsite context menu.
   test: async function topsites_pin_unpin() {
--- a/system-addon/test/functional/mochitest/head.js
+++ b/system-addon/test/functional/mochitest/head.js
@@ -1,5 +1,8 @@
 "use strict";
 
+XPCOMUtils.defineLazyModuleGetter(this, "PlacesTestUtils",
+  "resource://testing-common/PlacesTestUtils.jsm");
+
 function popPrefs() {
   return SpecialPowers.popPrefEnv();
 }
@@ -13,6 +16,11 @@ const ACTIVITY_STREAM_PREF = "browser.newtabpage.activity-stream.enabled";
 pushPrefs([ACTIVITY_STREAM_PREF, true]);
 gBrowser.removePreloadedBrowser();
 
+async function clearHistoryAndBookmarks() { // eslint-disable-line no-unused-vars
+  await PlacesUtils.bookmarks.eraseEverything();
+  await PlacesUtils.history.clear();
+}
+
 /**
  * Helper to wait for potentially preloaded browsers to "load" where a preloaded
  * page has already loaded and won't trigger "load", and a "load"ed page might
@@ -23,6 +31,36 @@ async function waitForPreloaded(browser) {
   if (readyState !== "complete") {
     await BrowserTestUtils.browserLoaded(browser);
   }
+}
+
+/**
+ * Helper to force the HighlightsFeed to update.
+ */
+function refreshHighlightsFeed() {
+  // Toggling the pref will clear the feed cache and force a places query.
+  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
+  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", true);
+}
+
+/**
+ * Helper to populate the Highlights section with bookmark cards.
+ * @param count Number of items to add.
+ */
+async function addHighlightsBookmarks(count) { // eslint-disable-line no-unused-vars
+  const bookmarks = new Array(count).fill(null).map((entry, i) => ({
+    parentGuid: PlacesUtils.bookmarks.unfiledGuid,
+    title: "foo",
+    url: `https://mozilla${i}.com/nowNew`
+  }));
+
+  for (let placeInfo of bookmarks) {
+    await PlacesUtils.bookmarks.insert(placeInfo);
+    // Bookmarks need at least one visit to show up as highlights.
+    await PlacesTestUtils.addVisits(placeInfo.url);
+  }
+
+  // Force HighlightsFeed to make a request for the new items.
+  refreshHighlightsFeed();
 }
 
 /**
--- a/system-addon/test/functional/mochitest/browser.ini
+++ b/system-addon/test/functional/mochitest/browser.ini
@@ -6,3 +6,5 @@ support-files =
 [browser_as_load_location.js]
 [browser_as_render.js]
 [browser_getScreenshots.js]
+[browser_highlights_section.js]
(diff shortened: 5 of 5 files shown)


system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -31,15 +31,11 @@ add_task(async function getHighlights() {
...
 
   let selector = "[data-mochitest='highlights-card']";
-  let selectorPlaceholder = "[data-mochitest='highlights-card-placeholder']";
   let message = "found cards";
 
   // it should display two highlights cards instead of placeholders.
   await check_highlights_elements(selector, count, message);
 
-  // it should display a placeholder card.
-  await check_highlights_elements(selectorPlaceholder, 1, "found placeholder");
```
Mardak: Why is this failing in the first place? An initial glance would make it seem quite odd why one assert would pass and not the other because it would seem to imply there's 2 highlights and 3 placeholders. But looking at `check_highlights_elements` shows that it's effectively doing a completely new test with opening a second new tab. Is that intended? If we checked that the page has 2 highlights and 1 placeholder, would it pass?

But even if a page is actually consistent in that there's 3 cards shown, it's quite suspicious that the `head` test functions results in this inconsistency. Is it necessary that the pref needs to be toggled each time a test runs?
```js
Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", true);
```

It's also a bit concerning that there's 3 very similar `head` functions that seem to have copy/pasted some open/wait/check/cleanup that each need to be fixed up if there is an underlying problem leading to this inconsistency.

piatra: > If we checked that the page has 2 highlights and 1 placeholder, would it pass?
Haven't tested, locally you can see it showing up correctly.
The idea was to have functions that did simple assertions that work independently (like commenting out the other code in the test case).
>  Is it necessary that the pref needs to be toggled each time a check runs?
No. Only once, after history/bookmarks have been updated. But the goal was to have the assert function work regardless of any other statements/changes. This is the workaround for getting the feeds to update.
> there is an underlying problem leading to this inconsistency.
I ran all the tests with `--verify` which is supposed to specifically find these problems
```
::: 1. Run each test 10 times in one browser. : Pass
::: 2. Run each test 5 times in a new browser each time. : Pass
::: 3. Run each test 10 times in one browser, in chaos mode. : Pass
::: 4. Run each test 5 times in a new browser each time, in chaos mode. : Pass 
```
It helped me find issues regarding cleanup in these tests.
I'm not saying the code is great, just that the choices were intentional. I'm currently getting a debug build to test this.

piatra: I was able to reproduce this locally on a debug build with `--disable-e10s`. I was expecting `--verify` included this.

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -21,10 +19,13 @@ async function add_highlights_bookmark(count) {
   for (let placeInfo of bookmarks) {
     await PlacesTestUtils.addVisits(placeInfo.url);
   }
+
+  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
```
Mardak: Moving this code doesn't make it any easier to understand. Probably better to put it in a documented function that can be reimplemented later to not need the pref changes.

system-addon/test/functional/mochitest/head.js
```diff
@@ -115,13 +113,9 @@ async function check_highlights_elements(selector, length, message) { // eslint-
...
   await waitForPreloaded(browser);
 
-  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
-  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", true);
-
   await BrowserTestUtils.waitForCondition(() => content.document.querySelector(selector), `No elements matching ${selector} were found`);
 
-  let found = await ContentTask.spawn(browser, selector, arg =>
-    content.document.querySelectorAll(arg).length);
+  let found = content.document.querySelectorAll(selector).length;
```
Mardak: Where did you see this type of `content` usage? https://bugzilla.mozilla.org/show_bug.cgi?id=1127393 seems like it's saying this is a bad pattern ?

Mardak: Actually, this just happens to slip through the eslint `no-cpows-in-tests` rule because it's too complicated to test accurately whether `content` is in a content task.
https://searchfox.org/mozilla-central/source/tools/lint/docs/linters/eslint-plugin-mozilla.rst#no-cpows-in-tests

system-addon/test/functional/mochitest/browser_highlights_section.js
piatra: The move (triggering pref changes only once) actually fixes the crash. (that's why I moved it)
https://treeherder.mozilla.org/#/jobs?repo=try&revision=c57070ed4152c44f76ad339b3b0306b04c79ad63

system-addon/test/functional/mochitest/head.js
piatra: I think a lot of tests use helper functions that spawn the ContentTask and I never got the difference. Will update.

system-addon/test/functional/mochitest/head.js
```diff
@@ -115,13 +113,9 @@ async function check_highlights_elements(selector, length, message) { // eslint-
   let browser = gBrowser.selectedBrowser;
   await waitForPreloaded(browser);
 
-  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
-  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", true);
-
   await BrowserTestUtils.waitForCondition(() => content.document.querySelector(selector), `No elements matching ${selector} were found`);
```
Mardak: This `content.` use is from main and triggering CPOWs. And you need to update this everywhere. This wait aross the different "check"s (probably best to rename those to "test" as it's not a simple check) seems like it could be part of some helper that does waitForCondition as pat of `before`?

system-addon/test/functional/mochitest/head.js
```diff
@@ -115,13 +113,9 @@ async function check_highlights_elements(selector, length, message) { // eslint-
...
 
-  let found = await ContentTask.spawn(browser, selector, arg =>
-    content.document.querySelectorAll(arg).length);
+  const found = await ContentTask.spawn(browser, selector, arg => content.document.querySelectorAll(arg).length);
   ok(found === length, `there should be ${length} of ${selector} found ${found}`);
```
Mardak: It's probably better to use `is` as when `ok` fails, it'll just say it wasn't truthy. Although I suppose you at least include the values in the printed message.

piatra:
@Mardak In my latest commit I rewrote the tests in `browser_highlights_section.js` using the `test_newtab` helper. There are still some nits with the code, wanted f? if I should move forward with the others as well.

piatra:
@Mardak updated to use test_newtab everywhere.

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -0,0 +1,69 @@
+"use strict";
+
+registerCleanupFunction(async () => {
+  await clearHistoryAndBookmarks();
+});
+
+function refreshHighlightsFeed() {
```
Mardak: This looks like it could be generally useful for tests that involve highlights, so move it to head.js with docs.

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -0,0 +1,69 @@
...
+  // Toggling the pref will clear the feed cache and force a places query.
+  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
+  Services.prefs.setBoolPref("browser.newtabpage.activity-stream.feeds.section.highlights", true);
+}
+
+async function add_highlights_bookmark(count) {
```
Mardak: nit: the naming convention seems to be camcelCase as usual with test_names or things that create_tests with underscores.

Also, this could be generally useful as well from head.js

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -0,0 +1,69 @@
...
+    url: `https://mozilla${i}.com/nowNew`
+  }));
+
+  for (let placeInfo of bookmarks) {
+    await PlacesUtils.bookmarks.insert(placeInfo);
+  }
+
+  // Add bookmark visits
+  for (let placeInfo of bookmarks) {
+    await PlacesTestUtils.addVisits(placeInfo.url);
```
Mardak: Does this second loop need to run strictly after the previous or can the `insert` happen then `addVisits` right away?

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -0,0 +1,69 @@
...
+  await add_highlights_bookmark(count);
+  await ContentTask.spawn(browser, null, async () => {
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".card-outer:not(.placeholder)"),
+      "No highlights cards found.");
+  });
+}
+
+test_newtab({
+  async before({tab}) {
+    await addHighlightsCards(tab.linkedBrowser, 2);
```
Mardak: Given the previous note about `after` and the pattern with these 2 `test_newtab` calls, did you try something like
```js
function test_highlights(bookmarkCount, test) {
  test_newtab({
    async before({tab}) {
      await addHighlightsBookmark(bookmarkCount);
      await ContentTask…
    },
    test,
    async after() {
      await clearHistoryAndBookmarks();
    }
  });
}
```

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -0,0 +1,69 @@
...
+  // Force HighlightsFeed to make a request for the new items.
+  refreshHighlightsFeed();
+}
+
+async function addHighlightsCards(browser, count) {
+  await clearHistoryAndBookmarks();
+  await add_highlights_bookmark(count);
+  await ContentTask.spawn(browser, null, async () => {
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".card-outer:not(.placeholder)"),
+      "No highlights cards found.");
```
Mardak: A description of what we're expecting to change would be useful. And the failure message isn't entirely right I think? Can we end up showing empty state instead of 3 placeholders? E.g., `test_highlights(0, test_empty_state)`

system-addon/test/functional/mochitest/browser_topsites_section.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
+
+registerCleanupFunction(async () => {
+  Services.prefs.clearUserPref("browser.newtabpage.activity-stream.default.sites");
+  Services.prefs.clearUserPref("browser.newtabpage.pinned");
+  await clearHistoryAndBookmarks();
```
Mardak: Similar to the previous file, we should be able to clean up more closely to the actual test than generally for the test file. I know the cleanup function pattern is typical of mochitests, but we'll try to do better.

system-addon/test/functional/mochitest/browser_topsites_section.js
```diff
@@ -0,0 +1,64 @@
...
+});
+
+// Check TopSites edit modal and overlay show up.
+test_newtab({
+  async before() {
+    await clearHistoryAndBookmarks();
```
Mardak: Does this test actually need to clear stuff?

system-addon/test/functional/mochitest/browser_topsites_section.js
```diff
@@ -0,0 +1,64 @@
...
+    found = content.document.querySelector(".modal-overlay");
+    ok(found && !found.hidden, "Should find a visible overlay");
+  }
+});
+
+// Test pin/unpin context menu options.
+test_newtab({
+  async before() {
+    // The pref for TopSites is empty by default.
+    Services.prefs.setStringPref("browser.newtabpage.activity-stream.default.sites", "https://www.youtube.com/,https://www.facebook.com/,https://www.amazon.com/,https://www.reddit.com/,https://www.wikipedia.org/,https://twitter.com/");
```
Mardak: You should be able to `pushPrefs(["browser.newtabpage.activity-stream.default.sites", …], ["browser.newtabpage.pinned", ""])` from `before` to have it set the pref and clean up at the end of the test.

system-addon/test/functional/mochitest/browser_topsites_section.js
```diff
@@ -0,0 +1,64 @@
...
+    const contextMenu = content.document.querySelector(".context-menu");
+    ok(contextMenu && !contextMenu.hidden, "Should find a visible topsite context menu");
+
+    const pinUnpinTopsiteBtn = contextMenu.querySelector(".context-menu-item a");
+    // Pin the topsite.
+    pinUnpinTopsiteBtn.click();
+
+    // Need to wait for pin action.
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".icon-pin-small"),
+      "No highlights cards found.");
```
Mardak: Wrong failure message.

This makes me wonder if we want some helper that gets injected into each of our content tests, so we could just do `await waitFor(".icon-pin-small");` or `await waitForNot(…)` Looks like `ContentTaskUtils` is directly imported: https://searchfox.org/mozilla-central/source/testing/mochitest/BrowserTestUtils/content/content-task.js#10

We don't need to fix this refactoring here. I'll try asking around about best practices in adding helper functions to the content task.

system-addon/test/functional/mochitest/browser_topsites_section.js
```diff
@@ -0,0 +1,64 @@
...
+
+    // Unpin the topsite.
+    pinUnpinTopsiteBtn.click();
+
+    // Need to wait for unpin action.
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelectorAll(".icon-pin-small").length === 0,
+      "No highlights cards found.");
+
+    pinnedIcon = content.document.querySelectorAll(".icon-pin-small").length;
+    is(pinnedIcon, 0, "should find no pinned topsites");
```
Mardak: This check seems to be implied by the previous `wait`. At least the earlier check for 1 `length` is to make sure we didn't get multiple pinned. Also, the condition should be able to just be `!document.querySelector(".icon-pin-small")` with appropriate failure message.

system-addon/test/functional/mochitest/head.js
```diff
@@ -13,6 +16,11 @@ const ACTIVITY_STREAM_PREF = "browser.newtabpage.activity-stream.enabled";
 pushPrefs([ACTIVITY_STREAM_PREF, true]);
 gBrowser.removePreloadedBrowser();
 
+async function clearHistoryAndBookmarks() { // eslint-disable-line no-unused-vars
+  await PlacesTestUtils.clearHistory();
```
Mardak: Apparently this is deprecated:
> > +  await PlacesTestUtils.clearHistory();
> nit: Please await PlacesUtils.history.clear(); instead, the PlacesTestUtils shortcut is deprecated.

https://bugzilla.mozilla.org/show_bug.cgi?id=1410240#c17

system-addon/test/functional/mochitest/browser_highlights_section.js
```diff
@@ -1,47 +1,13 @@
...
-}
-
 test_newtab({
   async before({tab}) {
-    await addHighlightsCards(tab.linkedBrowser, 2);
+    await addHighlightsBookmarks(2);
+    // Wait for HighlightsFeed to update and display the items.
+    await ContentTask.spawn(tab.linkedBrowser, null, async () => {
+      await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".card-outer:not(.placeholder)"),
+        "No highlights cards found.");
```
Mardak: Did the `test_highlights` approach not work?

piatra: I misread your comment. So just keep 1 generic test_highlights and pass in additional assertions in a fn.

piatra:
Updated + try looks good https://treeherder.mozilla.org/#/jobs?repo=try&revision=d8ab7318f01496418b5f56015994863ca20d33cc

system-addon/test/functional/mochitest/browser_topsites_section.js
```diff
@@ -0,0 +1,53 @@
+"use strict";
+
+// Check TopSites edit modal and overlay show up.
+test_newtab({
+  // it should be able to click the topsites edit button to reveal the edit topsites modal and overlay.
+  test: function topsites_edit() {
```
Mardak: nit: should be able to use just a function callback instead of an object with a test method

piatra:
I got it to fail locally. Ubuntu64 debug build, but the special combination that did it is `--disable-e10s --verify --run-until-failure`. I'll try to see what might be causing it.

piatra:
Looking at the logs 
```
[task 2017-12-05T01:43:43.113Z] 01:43:43     INFO - Browser Chrome Test Summary
[task 2017-12-05T01:43:43.115Z] 01:43:43     INFO - Passed:  21
[task 2017-12-05T01:43:43.117Z] 01:43:43     INFO - Failed:  0
````
If the code throws an exception it could skip an assert and show 0 failures but in this case a successful run shows 21 tests passing. 
So what's left is that something fails in the `after`, specifically `clearHistoryAndBookmarks`.

**Comment by Mardak** [comment]
In a successful run on try, it shows these two:
```
[task 2017-12-05T01:32:01.940Z] 01:32:01     INFO - Browser Chrome Test Summary
[task 2017-12-05T01:32:01.941Z] 01:32:01     INFO - Passed:  21
[task 2017-12-05T01:32:01.943Z] 01:32:01     INFO - Failed:  0
[task 2017-12-05T01:32:01.945Z] 01:32:01     INFO - Todo:    0
[task 2017-12-05T01:32:06.697Z] 01:32:06     INFO -  Browser Chrome Test Summary
[task 2017-12-05T01:32:06.698Z] 01:32:06     INFO -  	Passed: 21
[task 2017-12-05T01:32:06.700Z] 01:32:06     INFO -  	Failed: 0
```

And when it fails the leak test:
```
[task 2017-12-05T01:43:43.113Z] 01:43:43     INFO - Browser Chrome Test Summary
[task 2017-12-05T01:43:43.115Z] 01:43:43     INFO - Passed:  21
[task 2017-12-05T01:43:43.117Z] 01:43:43     INFO - Failed:  0
[task 2017-12-05T01:43:43.118Z] 01:43:43     INFO - Todo:    0
[task 2017-12-05T01:43:47.943Z] 01:43:47     INFO -  Browser Chrome Test Summary
[task 2017-12-05T01:43:47.944Z] 01:43:47     INFO -  	Passed: 21
[task 2017-12-05T01:43:47.945Z] 01:43:47     INFO -  	Failed: 1
[task 2017-12-05T01:43:47.945Z] 01:43:47     INFO -  	Todo: 0
```

Are you saying the additional `Failed: 1` is from the `after`? It's counting the FAILS, which includes:
[task 2017-12-05T01:43:47.774Z] 01:43:47    ERROR - TEST-UNEXPECTED-FAIL | browser/extensions/activity-stream/test/functional/mochitest/browser_topsites_section.js | leaked 2 window(s) until shutdown [url = about:newtab]

**Comment by piatra** [comment]
I meant to say that the leak happens in the `after` not the actual `Failed: 1`.

Searching through other tests where both functions are called I've noticed that the order of operations is always:
```
await PlacesUtils.bookmarks.eraseEverything();
await PlacesUtils.history.clear();
```
I have it the other way around. I haven't read too closely how the two work, but that's my next step. While I also do a try run with the calls inverted.

[0] https://searchfox.org/mozilla-central/rev/f5f1c3f294f89cfd242c3af9eb2c40d19d5e04e7/browser/components/places/tests/browser/browser_toolbar_overflow.js#37-38

**Comment by Mardak** [comment]
Leaks are checked when Firefox is shutting down, so it checks if anything during the test at any point ended up keeping a reference too long. I don't really have a good idea on what would be causing it, so that adjusting of clearing places is at least something.

**Comment by piatra** [comment]
The change also required an extra [`waitForCondition`](https://github.com/mozilla/activity-stream/pull/3828/commits/a2ba7d80560513946cdcb00b0be6116b56b5d25f#diff-0db811b2fcc35b788615a1043ee52665R28)
Try run https://treeherder.mozilla.org/#/jobs?repo=try&revision=ca06dcde9d4ad4f168c707656bbac5c3932152da&selectedJob=150770007
I wasn't able to get it running Activity Stream only tests using the syntax from your try push.

**Comment by piatra** [comment]
https://treeherder.mozilla.org/#/jobs?repo=try&revision=7ce1e25a775fbc93166e6a9c30342a2995c7c749
Now doing a refresh only on pref change (not pref init) and not having to clear history or bookmarks to force that refresh. 15 runs all green.

**Comment by Mardak** [comment]
@piatra actually hold on, does the pref change cause multiple refreshes -> broadcasts when starting up Firefox? E.g., init -> broadcast then pref change -> broadcast?

**Comment by piatra** [comment]
@Mardak I made another try push where I revert the previous broadcast change and use prefs to turn AS off and on to refresh the feed.
https://treeherder.mozilla.org/#/jobs?repo=try&revision=e53d9d0b25a35e46cf1dbfaf4f6fa6a161dc11f2

File: system-addon/test/functional/mochitest/browser_topsites_section.js [hunk 19]
```diff
@@ -20,6 +20,8 @@ test_newtab({
   async before({pushPrefs}) {
     // The pref for TopSites is empty by default.
     await pushPrefs(["browser.newtabpage.activity-stream.default.sites", "https://www.youtube.com/,https://www.facebook.com/,https://www.amazon.com/,https://www.reddit.com/,https://www.wikipedia.org/,https://twitter.com/"]);
+    await pushPrefs(["browser.newtabpage.activity-stream.enabled", false]);
```
Mardak: If we're going for a hacky way to force a refresh, we probably shouldn't do it by completely restarting activity stream:
1) this particular pref might go away soon as activity stream will just always be enabled
2) the full shutdown then startup of activity stream could lead to other issues similar to the issues related to "restarting" places data

We should comment that we're trying to force top sites to refresh if there's a better way to do this down the line.

A potentially more limited action that hopefully is safe is PLACES_LINK_BLOCKED. MIGRATION_COMPLETED could be an alternative, but the only way to trigger that would be to `notifyObservers` of the appropriate event, but other observers might do unexpected stuff…

piatra: I was expecting TopSitesFeed to behave more like the Sections. Shouldn't it react to `showTopSites` pref changes?  
https://github.com/mozilla/activity-stream/blob/8e5857c1f8b102c410892708dba9bb6f6bc3fcf6/system-addon/lib/SectionsManager.jsm#L330-L335

Mardak: I suppose toggling the feed should be better than toggling all of activity stream. If it passes tests without intermittents, I guess it's okay with comments.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -283,6 +284,9 @@ this.TopSitesFeed = class TopSitesFeed {
         if (action.data.name === DEFAULT_SITES_PREF) {
           this.refreshDefaults(action.data.value);
         }
+        if (action.data.name === TOP_SITES_PREF && action.data.value) {
+          this.refresh({broadcast: true});
```
Mardak: Oh, that's not what I meant or thought you were talking about. The feed has `browser.newtabpage.activity-stream.feeds.topsites` to control whether it's on or off, so when that gets toggled, it should cause the section to refresh, no?

piatra: Yes that works locally. Made another push try to make sure https://treeherder.mozilla.org/#/jobs?repo=try&revision=e23c9c961ef7b896594995bba8c0ddcc5be16fa6

This was confusing, the sections seem to use just 1 pref `feeds.section.*` while TopSites have both `feeds.topsites` and `showTopSites`. Do you know if there is intended?

piatra:
Not sure what happened in that previous try push (consistent failures on OSX debug) but central has been updated (2 backouts) and now everything is green https://treeherder.mozilla.org/#/jobs?repo=try&revision=16e50ca3732e0f3b77f9aec7201104a1d767c4a9

as-pine-proxy:
[Treeherder: ed01c14cbbc9424a653898072acd6358aca1af13](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=ed01c14cbbc9424a653898072acd6358aca1af13)

piatra:
There seems to be a failure on pine 
https://treeherder.mozilla.org/#/jobs?repo=pine&revision=ed01c14cbbc9424a653898072acd6358aca1af13&selectedJob=155401506

Looking at the failure screenshot https://public-artifacts.taskcluster.net/CA78UVLJS6m9GvQKgV-RiA/0/public/test_info/mozilla-test-fail-screenshot_ntfvfh.png I have a an idea what happened:
* The test wait condition is for topsites context menus
* A recent PR made changes so that empty topsite placeholders have a context menu
So it ends up clicking the wrong thing. This can be easily fixed by changing the wait condition.

Mardak:
This was reverted for the failure

Mardak:
This was reverted for the failure https://github.com/mozilla/activity-stream/commit/878ad8ebf31bed387be312bdace7a4adf0540dfc

PR #3839, opened 2017-11-13
New client-side rotation for personalization mode, and minor fixes for Pocket

csadilek:
Introduces a new client-side rotation mechanism for personalization mode:

Similar to frequency capping, we record impressions in a pref. Main difference is that for recs (stories) we only record the first impression. After fetching and sorting stories we check that stories don't have an impression older than 90 mins (configurable server-side). If they do we'll sort them to the end. The impression tracking pref is cleaned up after every fetch to remove impressions for stories no longer in the feed.

Other minor fixes included here are:
- Adding a check so domain affinity calculation is never calculated more than twice a day in case of idle-daily firing too often. Really just a safe-guard to avoid unneeded computation in case of a problem.
- Changing the View More Topics link to not use the CDN as requested by Pocket.

### DIFF
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -22,6 +22,7 @@ const STORIES_UPDATE_TIME = 30 * 60 * 1000; // 30 minutes
 const TOPICS_UPDATE_TIME = 3 * 60 * 60 * 1000; // 3 hours
 const STORIES_NOW_THRESHOLD = 24 * 60 * 60 * 1000; // 24 hours
 const MIN_DOMAIN_AFFINITIES_UPDATE_TIME = 12 * 60 * 60 * 1000; // 12 hours
+const DEFAULT_RECS_EXPIRE_TIME = 60 * 60 * 1000; // 1 hour
 const SECTION_ID = "topstories";
 const SPOC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.spoc.impressions";
 const REC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.rec.impressions";
@@ -232,15 +233,15 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.cache.set("domainAffinities", {});
   }
 
-  // If personalization is turned on we have to rotate stories on the client.
-  // Items only be on top for the time configured in settings.
-  // Once an item is expired it will be sorted to the back.
+  // If personalization is turned on, we have to rotate stories on the client so that
+  // active stories are at the front of the list, followed by stories that have expired
+  // impressions i.e. have been displayed for longer than recsExpireTime.
   rotate(items) {
     if (!this.personalized || items.length <= 3) {
       return items;
     }
 
-    const maxImpressionAge = (this.recsExpireTime || 5400) * 60 * 1000;
+    const maxImpressionAge = Math.max(this.recsExpireTime * 1000 || DEFAULT_RECS_EXPIRE_TIME, DEFAULT_RECS_EXPIRE_TIME);
     const impressions = this.readImpressionsPref(REC_IMPRESSION_TRACKING_PREF);
     const expired = [];
     const active = [];
@@ -377,14 +378,21 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.cleanUpImpressionPref(id => !activeStories.has(id), REC_IMPRESSION_TRACKING_PREF);
   }
 
-  cleanUpImpressionPref(expired, pref) {
+  /**
+   * Cleans up the provided impression pref (spocs or recs).
+   *
+   * @param isExpired predicate (boolean-valued function) that returns whether or not
+   * the impression for the given key is expired.
+   * @param pref the impression pref to clean up.
+   */
+  cleanUpImpressionPref(isExpired, pref) {
     const impressions = this.readImpressionsPref(pref);
     let changed = false;
 
     Object
       .keys(impressions)
       .forEach(id => {
-        if (expired(id)) {
+        if (isExpired(id)) {
           changed = true;
           delete impressions[id];
         }
@@ -482,7 +490,9 @@ this.TopStoriesFeed = class TopStoriesFeed {
             });
           }
           if (this.personalized) {
-            const topRecs = payload.tiles.filter(t => !this.spocCampaignMap.has(t.id)).map(t => t.id);
+            const topRecs = payload.tiles
+              .filter(t => !this.spocCampaignMap.has(t.id))
+              .map(t => t.id);
             this.recordTopRecImpressions(topRecs);
           }
         }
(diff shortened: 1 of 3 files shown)


system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -230,36 +233,25 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
-    const guid = items[0].guid;
-    if (!this.topItems.has(guid)) {
-      this.topItems.set(guid, 0);
-    } else {
-      const val = this.topItems.get(guid) + 1;
-      this.topItems.set(guid, val);
-      if (val >= 2) {
-        items.push(items.shift());
-        this.rotate(items);
+    const maxImpressionAge = (this.recsExpireTime || 5400) * 60 * 1000;
```
ncloudioj: nit: let's make a constant for `5400 * 60 * 1000`

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -374,43 +366,73 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+  }
+
+  // Clean up rec impression pref by removing all stories that are no
+  // longer part of the response.
+  cleanUpTopRecImpressionPref() {
+    const activeStories = new Set(this.stories.map(s => `${s.guid}`));
+    this.cleanUpImpressionPref(id => !activeStories.has(id), REC_IMPRESSION_TRACKING_PREF);
+  }
+
+  cleanUpImpressionPref(expired, pref) {
```
ncloudioj: This function wants some comments, at the very least for the first argument. It takes a function, whereas `expired` sounds more like a boolean.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -451,12 +473,18 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+        if (payload.tiles && viewImpression) {
+          if (this.shouldShowSpocs()) {
+            payload.tiles.forEach(t => {
+              if (this.spocCampaignMap.has(t.id)) {
+                this.recordCampaignImpression(this.spocCampaignMap.get(t.id));
+              }
+            });
+          }
+          if (this.personalized) {
+            const topRecs = payload.tiles.filter(t => !this.spocCampaignMap.has(t.id)).map(t => t.id);
```
ncloudioj: nit: could be easier to read if wrapping this line as
```
const topRecs = payload.tiles
  .filter(t => !this.spocCampaignMap.has(t.id))
  .map(t => t.id);
```

system-addon/test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -339,43 +341,87 @@ describe("Top Stories Feed", () => {
...
       assert.deepEqual(items, rotated);
 
+      // Recent impression should leave items unchanged
+      instance._prefs.get = pref => (pref === REC_IMPRESSION_TRACKING_PREF) && JSON.stringify({"g1": 1, "g2": 1, "g3": 1});
       rotated = instance.rotate(items);
-      assert.deepEqual(new Map([["g1", 1]]), instance.topItems);
       assert.deepEqual(items, rotated);
 
+      // Impression older than expiration time should rotate items
+      clock.tick(5400 * 60 * 60 * 1000 + 1);
```
ncloudioj: Ditto, `5400 * ...` should be a constant

system-addon/test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -339,43 +341,87 @@ describe("Top Stories Feed", () => {
...
-    it("should stop rotating if all items have been on top", () => {
-      let items = [{"guid": "g1"}, {"guid": "g2"}, {"guid": "g3"}, {"guid": "g4"}];
-      instance.topItems = new Map([["g1", 2], ["g2", 2], ["g3", 2], ["g4", 2]]);
+    it("should record top story impressions", async () => {
+      instance._prefs = {get: pref => undefined, set: sinon.spy()};
       instance.personalized = true;
 
-      const rotated = instance.rotate(items);
-      assert.deepEqual(items, rotated);
+      clock.tick(1);
```
ncloudioj: Nice tests 👍

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -190,6 +195,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.spocsPerNewTabs = settings.spocsPerNewTabs;
     this.timeSegments = settings.timeSegments;
     this.domainAffinityParameterSets = settings.domainAffinityParameterSets;
+    this.recsExpireTime = settings.recsExpireTime;
```
ncloudioj: Hmm, you've mentioned in the comment that

>  that stories don't have an impression older than 90 mins (configurable server-side)

Looks like it doesn't have any server-side configuration involved here, is there anything missing in here or that'd be added in a future patch?

ncloudioj: Oh, after another look at the code, it does call the `updateSettings` below. Perhaps we should put a guard here in case the server passes down a `unreasonable` value.

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -224,36 +233,25 @@ this.TopStoriesFeed = class TopStoriesFeed {
   }
 
   // If personalization is turned on we have to rotate stories on the client.
-  // An item can only be on top for two iterations (1hr) before it gets moved
-  // to the end. This will later be improved based on interactions/impressions.
+  // Items only be on top for the time configured in settings.
+  // Once an item is expired it will be sorted to the back.
```
ncloudioj: nit: let's rephrase this comment a bit. Perhaps "If personalization is turned on, we have to rotate all the stores on the client so that the active stories are relocated to the front of the list, followed by those expired ones."?

ncloudioj:
Have finished my first pass, it looks good.

Could you take a look at those comments (mostly are nits)? @csadilek 

I'd like to take another look once they are resolved. Thanks!

system-addon/lib/TopStoriesFeed.jsm
csadilek: Done, made sure value can't be lower than default, as that's the main risk i.e. rotating too often.

system-addon/lib/TopStoriesFeed.jsm
csadilek: Done, actually a bad bug there too, as 90 mins are 5400 * 1000, not 5400 * 60 * 1000.

as-pine-proxy:
[Treeherder: 2f2622bb033d806f189e5aa5809c73a13c812c93](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2f2622bb033d806f189e5aa5809c73a13c812c93)

Mardak:
https://hg.mozilla.org/releases/mozilla-beta/rev/2796c4cd23e3

PR #3840, opened 2017-11-13
fix (faviconfeed): #3834 exponential backoff for tippytop manifest fetch

### LINKED ISSUE #3834: Better handling of failure cases for remote tippy top fetch
Right now, the `FaviconFeed` can get into a state where it tries to fetch on every `SYSTEM_TICK` event. This will happen if there is no cache or the cached data is old and the remote endpoint returns a status code that isn't 200 (ok) or 304 (not modified).

We should probably add some sort of backoff algorithm.

rlr:
@Mardak exponential backoff?

Fixes #3834

### DIFF
--- a/system-addon/lib/FaviconFeed.jsm
+++ b/system-addon/lib/FaviconFeed.jsm
@@ -18,26 +18,30 @@ XPCOMUtils.defineLazyModuleGetter(this, "PlacesUtils",
 XPCOMUtils.defineLazyModuleGetter(this, "Services",
   "resource://gre/modules/Services.jsm");
 
-const TIPPYTOP_UPDATE_TIME = 24 * 60 * 60 * 1000; // 24 hours
+const FIVE_MINUTES = 5 * 60 * 1000;
+const ONE_DAY = 24 * 60 * 60 * 1000;
+const TIPPYTOP_UPDATE_TIME = ONE_DAY;
+const TIPPYTOP_RETRY_DELAY = FIVE_MINUTES;
 
 this.FaviconFeed = class FaviconFeed {
   constructor() {
-    this.tippyTopLastUpdated = 0;
+    this.tippyTopNextUpdate = 0;
     this.cache = new PersistentCache("tippytop", true);
     this.prefs = new Prefs();
     this._sitesByDomain = null;
+    this.numRetries = 0;
   }
 
   async loadCachedData() {
     const data = await this.cache.get("sites");
-    if (data && data._timestamp) {
+    if (data && "_timestamp" in data) {
       this._sitesByDomain = data;
-      this.tippyTopLastUpdated = data._timestamp;
+      this.tippyTopNextUpdate = data._timestamp + TIPPYTOP_UPDATE_TIME;
     }
   }
 
   async maybeRefresh() {
-    if (Date.now() - this.tippyTopLastUpdated >= TIPPYTOP_UPDATE_TIME) {
+    if (Date.now() >= this.tippyTopNextUpdate) {
       await this.refresh();
     }
   }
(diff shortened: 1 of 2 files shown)


system-addon/lib/FaviconFeed.jsm
```diff
@@ -50,15 +54,20 @@ this.FaviconFeed = class FaviconFeed {
...
       sitesByDomain._etag = etag;
-      this.tippyTopLastUpdated = sitesByDomain._timestamp = Date.now();
+      sitesByDomain._timestamp = Date.now();
       this.cache.set("sites", sitesByDomain);
       this._sitesByDomain = sitesByDomain;
+      this.tippyTopNextUpdate = sitesByDomain._timestamp + TIPPYTOP_UPDATE_TIME;
+      this.numRetries = 0;
+    } else {
+      // Something didn't go quite right. Retry with an exponential backoff.
+      let delay = Math.max(TIPPYTOP_UPDATE_TIME, TIPPYTOP_RETRY_DELAY * Math.pow(2, this.numRetries++));
```
rlr: can use `const`

system-addon/test/unit/lib/FaviconFeed.test.js
```diff
@@ -163,6 +163,16 @@ describe("FaviconFeed", () => {
       assert.deepEqual(feed._sitesByDomain, data);
       assert.notCalled(feed.cache.set);
     });
+    it("should handle server errors", async () => {
+      feed.loadFromURL = sinon.spy(url => ({data: [], status: 500}));
+      await feed.refresh();
+      assert.equal(1, feed.numRetries);
```
rlr: is there a good way to test that the delay is set properly? Monkeypatch (erm, mock) `Date.now()`?

system-addon/lib/FaviconFeed.jsm
Mardak: I'm thinking you wanted a `Math.min` to cap it at 24 hours of `delay`?

rlr: lol. yes. #fail

system-addon/lib/FaviconFeed.jsm
```diff
@@ -50,15 +54,20 @@ this.FaviconFeed = class FaviconFeed {
     let {data, etag, status} = await this.loadFromURL(this.prefs.get("tippyTop.service.endpoint"), headers);
     if (status === 304) {
       // Not modified, we are done. No need to dispatch actions or update cache. We'll check again in 24+ hours.
-      this.tippyTopLastUpdated = Date.now();
-      return;
-    }
-    if (status === 200 && data) {
+      this.tippyTopNextUpdate = Date.now() + TIPPYTOP_UPDATE_TIME;
+      this.numRetries = 0;
```
Mardak: This got me wondering if we should be storing some of this state in the cache? In particular, in the common case of `304`, `_timestamp` isn't updated, so every restart (after 24 hours from first `200`) will trigger a refresh?

system-addon/test/unit/lib/FaviconFeed.test.js
Mardak: The `clock = sinon.useFakeTimers()` will have `Date.now()` be `0` and you can advance it however you want. You probably don't need to advance and just check that the "next" time is the expected value, e.g., 0 + 5 minutes = 5 minutes.

rlr: ah, cool. If I had known, I would've caught my math fail 😉

system-addon/lib/FaviconFeed.jsm
rlr: yeah, good point. We should probably store the next update time at least on 304 and 200. Not sure if it's worth writing to cache on every failed retry though?

Mardak: Ehh.. mm.. I suppose just updating `_timestamp` on `304` should be enough and not bother caching with `numRetries`, i.e., a restart resets the backoff to 5 minutes.

rlr:
@Mardak ^ fixes pushed

system-addon/lib/FaviconFeed.jsm
```diff
@@ -49,16 +53,23 @@ this.FaviconFeed = class FaviconFeed {
...
-      this.tippyTopLastUpdated = sitesByDomain._timestamp = Date.now();
+      sitesByDomain._timestamp = Date.now();
       this.cache.set("sites", sitesByDomain);
       this._sitesByDomain = sitesByDomain;
+      this.tippyTopNextUpdate = sitesByDomain._timestamp + TIPPYTOP_UPDATE_TIME;
+      this.numRetries = 0;
+    } else {
+      // Something didn't go quite right. Retry with an exponential backoff.
+      const delay = Math.min(TIPPYTOP_UPDATE_TIME, TIPPYTOP_RETRY_DELAY * Math.pow(2, this.numRetries++));
+      this.tippyTopNextUpdate = Date.now() + delay;
```
Mardak: nit: let's avoid the duplicate code:
```js
let failedUpdate = false;
if (status === 200) {
  this._sitesByDomain = this._sitesArrayToObjectByDomain(data);
  this._sitesByDomain._etag = …
} else if (status !== 304) {
  failedUpdate = true;
}
let delay = TIPPYTOP_UPDATE_TIME;
if (failedUpdate) {
  delay = …
} else {
  this._sitesByDomain._timestamp = …
  this.cache.set(…
  this.numRetries = 0;
}
this.tippyTopNextUpdate = Date.now() + delay;
```

as-pine-proxy:
[Treeherder: 2adc00aed6d67c8baad3f9db4d849e04f7e5abfe](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2adc00aed6d67c8baad3f9db4d849e04f7e5abfe)

PR #3843, opened 2017-11-14
fix(favicons): Fall back to higher domain's icon if no subdomain match

### LINKED ISSUE #3842: Make sure go.twitch.tv gets a tippy top icon
Either we need to include "go.twitch.tv" as a domain, or be less strict on domain matching similar to how we allow "www.domain" and "domain" to use the same icon. Maybe it should try for an exact match first then fall back to a higher domain match? @rlr 

This would reduce the likelihood of https://bugzilla.mozilla.org/show_bug.cgi?id=1412505

Mardak:
Fix #3842. r?@rlr Not entirely sure if this is desired… but it would fix the go.twitch.tv vs twitch.tv situation. However, as in the unit test, it would set bugzilla.mozilla.org's icon to mozilla.org. Or similarly mail.google.com would get google.com.

### DIFF
--- a/system-addon/lib/FaviconFeed.jsm
+++ b/system-addon/lib/FaviconFeed.jsm
@@ -112,10 +112,25 @@ this.FaviconFeed = class FaviconFeed {
     }));
   }
 
+  /**
+   * Get a domain from an available list of domains preferring exact matches.
+   * @param {string} url The url to extract a domain from
+   * @param {array} domains An array of strings of acceptable domains
+   * @returns A domain or empty string if no match
+   */
+  getBestDomain(url, domains) {
+    let domain = getDomain(url);
+    while (domain && !domains.includes(domain)) {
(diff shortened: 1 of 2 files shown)


Mardak:
Here's some screenshots…

Before tippy:
screen shot 2017-11-13 at 5 38 44 pm

Tippy without this PR:
screen shot 2017-11-13 at 5 40 30 pm

After:
screen shot 2017-11-13 at 5 41 38 pm

as-pine-proxy:
[Treeherder: 8a4d27e65e9acf944f454d92d747bd27fdb65c51](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8a4d27e65e9acf944f454d92d747bd27fdb65c51)

Mardak:
This was reverted due to concerns of hoised rich icons replacing lower res page-provided icons resulting in the page not being able to replace it.

PR #3844, opened 2017-11-14
fix(prefs): Use MOZILLA_OFFICIAL instead of checking channel for dev-ness

### LINKED ISSUE #3769: Linux repacks detected as unofficial builds
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1411867

> const IS_UNOFFICIAL_BUILD = Services.prefs.getStringPref("app.update.channel") === "default";

The only pref using a custom _dev value right now:
https://github.com/mozilla/activity-stream/blob/653ef03329d0a2997d4a5b8f3f589602532f7fae/system-addon/lib/ActivityStream.jsm#L124-L127

So without a fix, we miss out on telemetry from these types of repacks.

Mardak:
Fix #3769. r?@k88hudson Also adds test that we don't use `value_local_dev` when not dev.

### DIFF
--- a/system-addon/lib/ActivityStreamPrefs.jsm
+++ b/system-addon/lib/ActivityStreamPrefs.jsm
@@ -4,6 +4,7 @@
 "use strict";
 
 const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/AppConstants.jsm");
 Cu.import("resource://gre/modules/Preferences.jsm");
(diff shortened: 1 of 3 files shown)


as-pine-proxy:
[Treeherder: a79a1bb60a57df9eb00e77708789642e7a5f1085](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a79a1bb60a57df9eb00e77708789642e7a5f1085)

PR #3845, opened 2017-11-14
fix (tippytop): URL for remote tippytop

rlr:
r? @Mardak @ncloudioj

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -137,7 +137,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["tippyTop.service.endpoint", {
     title: "Tippy Top service manifest url",
-    value: "https://s3.amazonaws.com/activitystream-dev-default-resources-s3bucket-1qw8m6s29v3dq/tippytop/icons.json"
+    value: "https://activity-stream-icons.services.mozilla.com/v1/icons.json.br"
   }]
 ]);
 


system-addon/lib/ActivityStream.jsm
```diff
@@ -137,7 +137,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["tippyTop.service.endpoint", {
     title: "Tippy Top service manifest url",
-    value: "https://s3.amazonaws.com/activitystream-dev-default-resources-s3bucket-1qw8m6s29v3dq/tippytop/icons.json"
+    value: "https://activity-stream-icons.services.mozilla.com/nightly/icons.json.br"
```
Mardak: If we expect "nightly" to be dynamically set based on the channel, we should use "%CHANNEL%" and have `FaviconFeed` call `Services.urlFormatter.formatURLPref("browser.newtabpage.activity-stream.tippyTop.service.endpoint")` (and get rid of the anti-pattern `new Prefs()`).

Mardak: Although for local/try/autoland builds, %CHANNEL% is "default" which doesn't seem to be a valid endpoint right now.

rlr:
@Mardak ^ I guess we can land this and ask @jbuck to add a */default/* endpoint that just mirrors nightly or something.

Mardak:
I guess it'll be a good test of handling:
```
$ curl -I https://activity-stream-icons.services.mozilla.com/default/icons.json.br
HTTP/1.1 403 Forbidden
Content-Type: application/xml
Connection: keep-alive
Date: Tue, 14 Nov 2017 23:22:45 GMT
Server: AmazonS3
X-Cache: Error from cloudfront
Via: 1.1 73b5e08c6138e8537ea4c8edfef5e6df.cloudfront.net (CloudFront)
X-Amz-Cf-Id: POgbPIzrBv1-ALrMcgmhXuBdKPctXKatWH6d2X0CSQx1tmEZECy2cA==
```

Although seems like we might not export today anyway…

Mardak:
@ncloudioj if we want to match tiles, do we want a version and/or locale in the path? Although I don't quite understand what we expect to get out of including the channel for this data. Versioning could be useful for format changes. Locale could target the tippytop data although potentially introduces some slight privacy concerns.

ncloudioj:
@Mardak I don't think locale is necessary in this case. As long as we have a version added, we can also pull out channel indeed. I'll ask @jbuck to make the endpoint change as such.

ncloudioj:
Hey @rlr , here is the new endpoint `https://activity-stream-icons.services.mozilla.com/v1/icons.json.br`

rlr:
OK yay, back to a simple pref. Updated ^

as-pine-proxy:
[Treeherder: a2ce9fc3b97a282b27f17543f309cb0ecf32984a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=a2ce9fc3b97a282b27f17543f309cb0ecf32984a)

Mardak:
https://hg.mozilla.org/releases/mozilla-beta/rev/6ff9a23d3212

PR #3852, opened 2017-11-15
chore(systemaddon): Use a single method to dispatch TopSites icon stats.

### LINKED ISSUE #3851: Extract TopSites telemetry call into a single method
https://github.com/mozilla/activity-stream/pull/3652#discussion_r151012061

piatra:
Closes #3851

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -41,7 +41,10 @@ function countTopSitesIconsTypes(topSites) {
 }
 
 class TopSites extends React.PureComponent {
-  componentDidUpdate() {
+  /**
+   * Dispatch session statistics about the quality of TopSites icons.
+   */
+  _storeTopSitesIconStats() {
     const realTopSites = this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
 
     const topSitesIconsStats = countTopSitesIconsTypes(realTopSites);
@@ -52,15 +55,12 @@ class TopSites extends React.PureComponent {
     }));
   }
 
-  componentDidMount() {
-    const realTopSites = this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
+  componentDidUpdate() {
+    this._storeTopSitesIconStats();
+  }
 
-    const topSitesIconsStats = countTopSitesIconsTypes(realTopSites);
-    // Dispatch telemetry event with the count of TopSites images types.
-    this.props.dispatch(ac.SendToMain({
-      type: at.SAVE_SESSION_PERF_DATA,
-      data: {topsites_icon_stats: topSitesIconsStats}
-    }));
+  componentDidMount() {
+    this._storeTopSitesIconStats();
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -41,7 +41,10 @@ function countTopSitesIconsTypes(topSites) {
 }
 
 class TopSites extends React.PureComponent {
-  componentDidUpdate() {
+  /**
+   * Dispatch session statistics about the quality of TopSites icons.
+   */
+  _storeTopSitesIconStats() {
```
Mardak: "Store" has an existing usage/meaning, so let's use either "save" or "dispatch"

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -41,7 +41,10 @@ function countTopSitesIconsTypes(topSites) {
...
     const realTopSites = this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
```
Mardak: What does "real" mean here? Are the ones sliced away not real? If it's supposed to be the same functionality as that of `render`, then let's refactor that.

system-addon/test/unit/content-src/components/TopSites.test.jsx
```diff
@@ -76,33 +76,31 @@ describe("<TopSites>", () => {
...
+      assert.calledOnce(storeStatsSpy);
     });
-    it("should dispatch session perf data on update", () => {
+    it("should call _storeTopSitesIconStats on componentDidUpdate", () => {
       wrapper.instance().componentDidUpdate(DEFAULT_PROPS);
 
+      assert.calledOnce(storeStatsSpy);
+    });
+    it("should dispatch SAVE_SESSION_PERF_DATA", () => {
+      wrapper.instance()._storeTopSitesIconStats(DEFAULT_PROPS);
```
Mardak: This method doens't take any parameters, so passing in an argument is confusing

system-addon/content-src/components/TopSites/TopSites.jsx
piatra: The way I would do this is set `this.realTopSites` in the `render` method. Alternatively there is `componentWillUpdate` but docs say:
> This method is not called for the initial render.

Mardak: Saving the property is almost like setting state, which shouldn't be done from `render`. And I believe the react pattern is to not set state based on props and instead just provide a helper to compute the value from props.

Mardak: Also, still doesn't answer what "real" means here.

piatra: That was the name used in the TopSites render method. We have two sets of top sites: the ones you regularly see and the ones in the top sites edit modal (that’s where the distinction came from I think). @r1cky knows more about this.

as-pine-proxy:
[Treeherder: aadfc4db1c7252184e69d4d6d5d3e88ca7e4882b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=aadfc4db1c7252184e69d4d6d5d3e88ca7e4882b)

as-pine-proxy:
[Treeherder: 1b14fb4f6cc34daf8dbf66a7a8acf3779ae4074f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1b14fb4f6cc34daf8dbf66a7a8acf3779ae4074f)

PR #3857, opened 2017-11-16
Fix Bug 1421426 - Upgrade React-Intl and React and switch to prop-types in prep for React 16

### LINKED ISSUE #3836: Upgrade React Intl to latest version
We need to do this in order to upgrade to React 16.  As part of it, we should either file an issue to make React Intl warnings break our tests, or we should just fix that error.

dmose:


dmose:
Please don't land this after review.  My intent is to do a partial, but not complete squash, and then extend the main commit message to have the [required license information](https://www.mozilla.org/en-US/foundation/licensing/).  I'll be doing the license.html patching in the bug for the next export, or one of its dependents.

package.json
```diff
@@ -13,8 +13,10 @@
     "moment": "2.15.2",
     "object-sizeof": "1.1.1",
     "page-metadata-parser": "0.6.0",
+    "prop-types": "15.6.0",
     "react": "15.3.2",
```
k88hudson: You'll need to update `react` here as well since it is used by node to generate the preloaded versions

dmose: It's not obvious to me why having react use its own prop-types internally with react-intl using the new stuff would be a problem, since the prop-types themselves aren't (shouldn't be?) passed around.  That said, it's not an unreasonable request; I'll poke...

dmose:
Reminder to self: move react-intl out of devDependencies into dependencies.

dmose:
If you look at the CI failure, it looks like pulling in even React 15.5.4 then requires a new version of Enzyme, which is itself a bit messy.  I've started a separate PR for that.

@k88hudson Unless we know of some specific reason why we can't keep the existing React (which uses its own internal proptypes) and have the updated react-intl use the new prop-types, I propose that we drop these last two commits.

Mardak:
With the current set of commits, this means we need to upgrade react twice? Once to this intermediate version then another to 16? I think the complexity is caused by our using of react for both 1) testing and prerendering and 2) in-firefox usage. ?? If so, would it make it any easier to focus on either testing+prerendering vs in-firefox?

dmose:
> With the current set of commits, this means we need to upgrade react twice? Once to this intermediate version then another to 16? 

Yes. Given the amount of churn in React but also the various other files, staging the changes like this and getting them in front of users in pieces has the value of both mitigating risk and making the individual pieces easier to manage.  I don't expect this to slow us down too much, and may even be a net win by avoiding too coupling too many changes (see the Subtle Surprises section in this [blog entry](https://blog.discordapp.com/lessons-from-migrating-a-large-codebase-to-react-16-e60e49102aa6) for some of the concerns).

> I think the complexity is caused by our using of react for both 1) testing and prerendering and 2) in-firefox usage. ??

Kinda. Here are some causes of complexity that I see:

* The fact that we pre-render might make some difference, but the only thing that has stood out so far is that the failure in the CI build didn't happen locally for me; I haven't yet tried to track down why.

* The way that we do our vendor bundling adds some, because it's easy to make inadvertent errors in the by-hand steps, and sometimes hard to guess which copies of the library are being run and whether they're consistent across development & production steps.  That said, I'm now pretty on top of them.

Most of all, I think, is that a lot of various dependent packages got upgraded in semi-lockstep after the point at which we stopped upgrading our dependencies.  Trying to jump over this makes ordering the changes hard, and staples a lot of big changes together.

>  If so, would it make it any easier to focus on either testing+prerendering vs in-firefox?

I don't think so, because I think those three things are coupled together fairly tightly.  What I've also been discovering is that the various dependent package APIs are inter-coupled in ways that are a bit hard to see until one tries various orderings of changes.  

My gut is that our best bet is to bite the bullet here and do it in chunks, and stay on top of our dependency updates in the future to avoid this kind of pain.

My suggestion for the rest looks something like this:

* we land this patch without the react changes now (feel free to steal the review :-)
* i'll work on the enzyme changes, which use an adapter setup and are less tightly coupled and land those.
* I'll put together small patches to update to React 15.5.4, and the redux-changes.
* Finally I'll update to React 16.

If you want to talk more, I'd suggest Vidyo to hammer through any remaining issues.

dmose:
I've rebased this on top of the Enzyme PR, and added some tweaks that will perhaps make the CI happy.  That said, I'd still prefer to extract the React upgrade and these tweaks to a different PR and land them separately.

dmose:
This does fix the React tests, but there's still a prop-types warning, which goes away when react-redux is also upgraded.

k88hudson:
I'm seeing a source map reference to `react-intl.min.js.map` in react-intl.js that does not resolve. Did you want to include this source map, or else remove the line? Otherwise we'll get a warning.

Re: the new enzyme default, I'm in favour of adopting it and setting the skip lifecycle option where necessary.

Feel free to land this as two (or more) commits for easier bisecting.

k88hudson:
I'm noticing a `Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead` warning when running this in debug mode. Is this expected? Interestingly this isn't caught by our tests since the filter doesn't pick it up.

dmose:
> I'm seeing a source map reference to react-intl.min.js.map in react-intl.js that does not resolve. Did you want to include this source map, or else remove the line? Otherwise we'll get a warning.

I'll remove the line.  I'm not convinced we want sourcemaps in the production builds themselves given the added size, and we don't yet have logic for a separate react-intl-dev.js build for nightly or debug.

> I'm noticing a Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead warning when running this in debug mode. Is this expected? Interestingly this isn't caught by our tests since the filter doesn't pick it up.

That will be fixed by [upgrading the version of react-redux that the tests use](https://bugzilla.mozilla.org/show_bug.cgi?id=1423652).  I'll poke at why our filters aren't picking this up.

dmose:
> Interestingly this isn't caught by our tests since the filter doesn't pick it up.

It does use console.error. My guess is that what's happening is that since this is triggered by the use of react-redux, and react-redux almost certainly loads before unit-entry.js, the warning is emitted  before console.error has been monkeypatched.

as-pine-proxy:
[Treeherder: 38e248d52101c97c36a0fddaef63932c0f621126](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=38e248d52101c97c36a0fddaef63932c0f621126)

PR #3862, opened 2017-11-17
chore(system-addon): closes #3861, remove unused react-json-inspector…

### LINKED ISSUE #3861: remove react-json-inspector in prep for React 16
The version we've got isn't compatible with React 16, and, it turns out, we're not actually using it at all. 

PR forthcoming...

dmose:
… dep.  The version we've got in package.json is not compatible with React 16, but, as it turns out, we're not using it at all.


Mardak:
Fixed by #3869 instead

PR #3751, opened 2017-10-22
Fix #3409 - Sass Linting Errors

SeanPrashad:
**Resolves #3409** - Sass linting was raising multiple issues (**96 issues - 22 errors & 74 warnings**).

To resolve this bug, I performed the following steps:

- Installed [`sass-lint` globally](https://www.npmjs.com/package/sass-lint) via npm
  - Utilized the built-in CLI debugging to diagnose and fix each issue via the following command:
`sass-lint -c .sass-lint.yml 'system-addon/content-src/**/*.scss' -v -q`
  (**Note**: the `-v` argument specifies *verbose output*)
- Ran `npm run startmc` in order to build copies of my changes to `mozilla-central` locally
- Referred to the `sass-lint` documentation ([/docs/rules](https://github.com/sasstools/sass-lint/tree/develop/docs/rules)) to identify solutions with respect to each error
- After all errors & warnings were resolved, I ran `npm run testmc` which produced the following output:
runningunittests

PR #3869, opened 2017-11-21
chore(package): Remove unused dependencies and add minimist

### LINKED ISSUE #3537: Audit unused dependencies
### Steps to reproduce:

```sh
$ npx depcheck

Unused dependencies
* classnames
* fancy-dedupe
* lodash.debounce
* moment
* object-sizeof
* page-metadata-parser
* react-json-inspector
* redux-logger
* redux-thunk
* redux-watch
* reselect
* tippy-top-sites
```

Was looking through a recent PR and noticed we didn't seem to be using **classnames** package anymore (but it still appears to be in the package.json). We may want to quickly troll through the possibly accurate list from **depcheck** and see if there are any other modules which we aren't using but still appear in package.json.

Note that there are some false-positives, such as **tippy-top-sites** which is used in a `tippymc:*` npm script in package.json (via _yamscripts.yml_).

Mardak:
Fix #3537. r?@k88hudson Removes a bunch of dependencies by testing with `buildmc`/`startmc`/`testmc`

`minimist` is being used by `bin/render-activity-stream-html.js`

The `require("…")` comments don't automatically do anything, but maybe it'll help track down usage or not of a dependency ?

@rlr says we don't need `tippymc` stuff anymore (and it's out of sync anyway as we had manual edits to the checked in files anyway).

Wasn't entirely sure about `babel` but looks like it was originally added for `webpack.module.loaders.loader: "babel"` whereas now we have `webpack.module.rules.loader: "babel-loader"`. At least it seems to build okay when removed…

Also `requirejs` isn't needed anymore for `karma`?


Mardak:
Oh, I also didn't remove `sass-lint` even though "unused" as the `yamscripts` is just commented out with the intention of turning it back on #3751.

as-pine-proxy:
[Treeherder: 2e8cf45e2f9ab35142341fa921980e615ae330ab](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2e8cf45e2f9ab35142341fa921980e615ae330ab)

PR #3870, opened 2017-11-21
chore(package): Update dependencies that pass unit tests using next-update

### LINKED ISSUE #2321: Green up our JS dependencies

### LINKED ISSUE #3743: Messages when running eslint: can't resolve reference #/definitions/basicConfig
@Standard8 looks like this started with the eslint 4 update from #3705:
```
> esw --ext=.js,.jsm,.json,.jsx .

can't resolve reference #/definitions/basicConfig from id #
can't resolve reference #/definitions/basicConfigOrBoolean from id #
can't resolve reference #/definitions/basicConfigOrBoolean from id #
✓ Clean (1:15:59 PM)
```

Before:
```
> esw --ext=.js,.jsm,.json,.jsx .

✓ Clean (1:15:16 PM)
```

Mardak:
Fix #2321 and fix #3743. As suggested by @k88hudson that a single commit updating should be fine as any unexpected failures should be easy to spot what dependency broke. Blocked on #3869 landing.


Mardak:
Updated with latest `next-update` run (ignoreing those that already failed) `-w react,react-dom,chai,enzyme,eslint-plugin-mozilla,istanbul-instrumenter-loader,karma-sinon,loader-utils,react-addons-test-utils`

Additionally updated `github 12.0.3 -> 12.0.4`

Mardak:
I suppose we should separately track updating each of react,react-dom,chai,enzyme,eslint-plugin-mozilla,istanbul-instrumenter-loader,karma-sinon,loader-utils,react-addons-test-utils ?

Mardak:
Maybe we should remove the react-redux/redux update as that should be covered by the react 16 update? maybe? This also doesn't actually update the vendor files right now.

dmose:
I would still be inclined to split this out into at least a larger number of PRs, because the failure modes here are not just "causes correct failures in the tests".  The failure modes may also be "causes false positives in the tests" and "breaks integration/functional tests that we haven't written but wish we had", and the ability to bisect can alleviate pain there.  That said, I totally get that 1 PR per package is likely overkill.

Another thing that I'm a bit concerned about with this patch is that a bunch of the changes here are across major versions, some of which are likely to have semantic changes to their APIs.
 
I also suspect that react-intl real wants to live in dependencies rather than devDependencies.  That's covered by #3857; I'll fix it there, since it requires vendoring fun.

dmose:
That said, maybe I'm being too conservative here, I'm not sure.  At the very least, I'd appreciate leaving out react-intl and the redux-related stuff.

dmose:
And, really, when I said "separate PRs", I meant "separate commits".

Mardak:
Additionally updated eslint 4.11.0 -> 4.12.0, github 12.0.4 -> 12.1.0, simple-git 1.82.0 -> 1.83.0.

PR #3879, opened 2017-11-25
Fix Bug 1421440 - There should not be double scrollbars when the Preferences Pane is open

mmBabol:


Mardak:
Thanks for the PR. We'll soon be removing the preference sidebar in https://bugzilla.mozilla.org/show_bug.cgi?id=1432672 so let's close this for now.

PR #3881, opened 2017-11-28
fix(search): Close Bug 1417943 - in-page search box gets focus after entering input in location bar (only on initial tab at browser startup)

Mardak:
Also fix Bug 1400585 - With activity-stream.aboutHome.autoFocus set, initil focus is in the address bar then moves to content search
Also fix Bug 1421417 - Focus moves to about:home search box when turning on the search preference

r?@sarracini Basically reverts #3452


as-pine-proxy:
[Treeherder: ec7c0d2beba2dce31eef3fdc6835c1cb556111c3](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=ec7c0d2beba2dce31eef3fdc6835c1cb556111c3)

PR #3888, opened 2017-11-30
Fix Bug 1421857 - fix test-merges regression to un-bust pine

dmose:
Tested by-hand in place, it should fix the problem (and can't possibly make it worse :-).

### DIFF
--- a/bin/test-merges.js
+++ b/bin/test-merges.js
@@ -98,7 +98,7 @@ function getPRMergeCommitId(prNumber) {
     owner: AS_REPO_OWNER,
     repo: AS_REPO_NAME,
     issue_number: prNumber
-  }).then(data => {
+  }).then(({data}) => {
     if (data.incomplete_results) {
       // XXX should handle this case theoretically, but since we'll be running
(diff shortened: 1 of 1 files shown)


dmose:
Although, to be fair, pine is closed right now, so the more normal code path has not yet been exercised, since pushing fails.

bin/test-merges.js
```diff
@@ -98,7 +98,9 @@ function getPRMergeCommitId(prNumber) {
     owner: AS_REPO_OWNER,
     repo: AS_REPO_NAME,
     issue_number: prNumber
-  }).then(data => {
+  }).then(wrappedData => {
+    let data = wrappedData.data;
```
Mardak: nit: could just do `.then(({data}) => {`

as-pine-proxy:
[Treeherder: f292dbdfc13b9cabd829ce88f5eff8ec4b65699e](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f292dbdfc13b9cabd829ce88f5eff8ec4b65699e)

PR #3891, opened 2017-12-04
Fix Bug 1402654 - Allow custom image / thumbnail for an edited top site

piatra:
Closes https://bugzilla.mozilla.org/show_bug.cgi?id=1402654

## How it works

* when the user inputs a url and makes a preview request we dispatch `SCREENSHOT_REQUEST` that gets handled in the TopSitesFeed `getScreenshotPreview`. this makes a `Screemshot.getScreenshotForURL` request and dispatches a SCREENSHOT_UPDATE with the response (or SCREENSHOT_FAILED).
* the TopSites reducer now handles SCREENSHOT_UPDATE (for different kinds of paylods)
  * it can set `screenshotPreview` or `customScreenshot` in addition to `screenshot` to a Topsite.
* when you update a topsite that includes a custom url, we have to clear the previous cached value from the LinksCache object
* when we query for topsites we only fetch the custom screenshot if it is set (so no rich icons or regular screenshot or favicon)
* the url value is stored as customScreenshotURL in the pinned pref
* because you have to click “save” to set a custom screenshot, all topsites with a custom screenshot are pinned sites
* The preview form uses the same component we use to display a Topsite item. There are 2 extra cases to handle: 
  * request failure -> letter fallback
  * custom screenshot `screenshotPreview || customScreenshot`
* for the Form inputs refactoring, the new component TopSiteEditInputs includes:
  * the url, title and custom screenshot inputs
  * The component handles the logic to display the error labels and to focus the inputs. Also the toggle to show hide the custom screenshot input is here. All the validation and the input change handler are still in the TopsiteForm.

----------

## UI/UX

There is no one source of information but I will try to list everything.
Latest design https://mozilla.invisionapp.com/share/7DEP1VMYX#/screens/266461200_Top_Site_-_Edit_Image_
In addition to that:
* Adding the tab loading animation when fetching a url
* Adding an X button at the end to clear the input
  * spec here https://design.firefox.com/photon/components/input-fields.html#behaviors
* Change the input focus behavior to match the same spec from design.firefox.com (blue border&shadow)
* The behavior we want for the screenshot input is:
  * hidden by default (unless user value is set)
  * click the link to reveal it (no option/action to hide back)
* Add input button to default site context menu (without having to go through the "Edit Mode")

Desired but not implemented:
* Adding a fade effect for text overflow (similar how location bar is doing it)
  * https://searchfox.org/mozilla-central/rev/f6f1731b1b7fec332f86b55fa40e2c9ae67ac39b/browser/base/content/browser.css#561
  * but there is no spec for how the "x" and the fade would work together
  * the location bar achieves the effect without actually using an input element
  * not clear what other input fields should get the fade effect

No required:
* the new design of the buttons
  * in the invision spec the cancel/save buttons look different
  * that design is not final yet so we are still using our current buttons

Let me know if I skipped something.


Mardak:
@aaronrbenson What's the expected behavior for various interactions and does this current PR look correct?

This used to have the page's screenshot, but when the image failed, it switched to just the letter fallback. The change to letter could be useful as an extra indication that it attempted. Also, is the failure message text and sizing correct -- I don't see that message in the invision doc:
screen shot 2017-12-04 at 8 09 57 pm

The preview link should not be interact-able and should have the image cropped differently? (Currently you can click it, and it navigates.)
screen shot 2017-12-04 at 8 10 25 pm

This shows up when deleting the content of the input box:
screen shot 2017-12-04 at 8 11 33 pm

I guess we shouldn't show the old screenshot when the url is cleared? Also, the sizing of the text and the string for the custom image should have ellipsis…? Also, what should cause the custom image input to disappear? Right now moving focus away from the image input box when there's no value makes it disappear.
screen shot 2017-12-04 at 8 12 00 pm

Is this the desired layout with the URL red error message?
screen shot 2017-12-04 at 8 12 58 pm

We probably want a non-text-selection cursor for hovering… and the color should be?
screen shot 2017-12-04 at 8 32 24 pm

The size of the confirmation button changes widths when it goes from Preview to Save, but probably not too much we can do there while not causing oddness for l10n?
screen shot 2017-12-04 at 8 35 05 pm

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -293,6 +335,9 @@ this.TopSitesFeed = class TopSitesFeed {
       case at.TOP_SITES_UNPIN:
         this.unpin(action);
         break;
+      case at.SCREENSHOT_REQUEST:
+        this.setCustomTopsiteScreenshot(action.data);
```
Mardak: This is called on a preview, so calling a `set` would seem to be more permanent than a preview. So either the method is not actually permanent or it shouldn't be as permanent?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -190,6 +197,33 @@ this.TopSitesFeed = class TopSitesFeed {
...
+    const pinnedLinks = await this.pinnedCache.request();
+    const frecentLinks = await this.frecentCache.request();
+    // Get the LinksCache entry that represents the topsite to update.
+    const link = pinnedLinks.concat(frecentLinks).find(topsite =>
+      topsite && topsite.url === url);
+
+    if (link) {
+      // If previous request failed, we need to clear any cached result
+      link.screenshotPreview = undefined;
+      Screenshots.maybeCacheScreenshot(link, customScreenshotURL, "screenshotPreview",
```
Mardak: Why do we want to cache the preview image?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -190,6 +197,33 @@ this.TopSitesFeed = class TopSitesFeed {
...
+        // When done, broadcast the result of the request.
+        screenshotPreview => {
+          this.store.dispatch(ac.BroadcastToContent({
```
Mardak: We probably don't need to broadcast to all pages when one form on a page is requesting the image? This would also change the main store's state which doesn't seem right for a transient image preview?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -165,19 +169,22 @@ this.TopSitesFeed = class TopSitesFeed {
    * Get an image for the link preferring tippy top, rich favicon, screenshots.
    */
   async _fetchIcon(link) {
+    const hasCustomScreenshot = !!link.customScreenshotURL;
```
Mardak: We should probably short circuit first if it has a custom screenshot and do some custom behavior. Or maybe fetchIcon just shouldn't be called if we aren't using an icon anyway?

system-addon/lib/Screenshots.jsm
```diff
@@ -63,20 +65,32 @@ this.Screenshots = {
...
+  /**
+  * Update image cache so that it will use the custom screenshot when
+  * looking up the site url.
+  *
+  * @param newScreenshotURL {string} URL pointing to the custom screenshot
+  * @param url {string} Site url used to lookup the screenshot.
+  */
+  async replaceScreenshot(newScreenshotURL, url) {
+    // .copy() will overwrite any existing thumbnail associated with `url`
+    await PageThumbsStorage.copy(newScreenshotURL, url);
```
Mardak: This doesn't seem right. There are other consumers of `PageThumbs`, so having it show some custom image that is *not* the page's thumbnail will probably lead to bugs. Highlights doesn't replace a url's thumbnail with the `preview_image_url` by requesting the thumbnail for that image instead of the page. Maybe something like that could be used for these custom preview images for top sites?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -1,24 +1,50 @@
...
+      screenshotPreview: null,
+      showCustomScreenshotForm: false,
+      screenshotRequestFailed: false
     };
     this.onLabelChange = this.onLabelChange.bind(this);
     this.onUrlChange = this.onUrlChange.bind(this);
     this.onCancelButtonClick = this.onCancelButtonClick.bind(this);
+    this.onCustomScreenshotUrlChange = this.onCustomScreenshotUrlChange.bind(this);
+    this.onEnableScreenshotUrlForm = this.onEnableScreenshotUrlForm.bind(this);
+    this.onHideCustomScreenshotForm = this.onHideCustomScreenshotForm.bind(this);
```
Mardak: With the 3 of these label/inputs, it would seem natural to move to a component

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -15,27 +15,39 @@ const TopSiteLink = props => {
...
-    imageStyle = {
-      backgroundColor: link.backgroundColor,
-      backgroundImage: `url(${tippyTopIcon || link.favicon})`
-    };
-  } else {
-    // styles and class names for top sites with screenshot + small icon in top left corner
-    imageClassName = `screenshot${link.screenshot ? " active" : ""}`;
-    imageStyle = {backgroundImage: link.screenshot ? `url(${link.screenshot})` : "none"};
+  // If the screenshot request failed use letter fallback.
+  if (!screenshotRequestFailed) {
```
Mardak: Something about this seems unnecessarily complex,but I'll need to think about it a bit more. I suppose we should be able to avoid all the lines of change by having a empty
```js
if (screenshotRequestFailed) {
  // Just use the letter fallback
} else if (screenshotPreview…) {
} else if (tippyTopIcon …) {
```

aaronrbenson:
Could I get a build to do some UX quality checking for this? There are a lot of little interactions and behaviors that @Mardak raises above and it would be helpful if I could test it out locally to provide notes.

system-addon/lib/TopSitesFeed.jsm
piatra: @Mardak I agree with the comment but I'm not sure how to address this. Any advice on how to pass the message?

locales/en-US/strings.properties
```diff
@@ -146,6 +146,13 @@ topsites_form_add_button=Add
 topsites_form_save_button=Save
 topsites_form_cancel_button=Cancel
 topsites_form_url_validation=Valid URL required
+topsites_form_label_title=Title
+topsites_form_label_url=URL
+topsites_form_label_screenshot_url=Custom Image
+topsites_form_screenshot_request_error=Image failed to load. Try a different URL.
+topsites_form_activate_custom_screenshot=Use a custom image
+topsites_form_custom_screenshot_placeholder=Paste a URL here
```
pdehaan: Why does one placeholder say "Type or paste" and the other is just "Paste"? Can't you type into the topsites URL?

```
topsites_form_url_placeholder=Type or paste a URL
topsites_form_custom_screenshot_placeholder=Paste a URL here
```

system-addon/lib/TopSitesFeed.jsm
Mardak: I believe @sarracini was looking at adding something like `SendOnlyToContent` in https://bugzilla.mozilla.org/show_bug.cgi?id=1415491 but before then, I suppose the closest would be `SendToContent`, but that would still update main store.

@k88hudson is there a preferred way to send transient data to a single content page without updating main store?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
Mardak: The 3 here I was referring to the 3 label/inputs: title label/input, url label/input, image label/input

Mardak:
@piatra could you comment on each review comment on what changed? It's hard to track down what's going on with so many commits with it already being a large change. Also, it would have been helpful to have a summary of what the approach is in the initial PR message.

Did you get UX feedback? And if so, what are the decisions?

piatra:
@Mardak will do.
No UX feedback yet, will be doing that next week in person.

system-addon/common/Reducers.jsm
```diff
@@ -95,18 +95,25 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       return Object.assign({}, prevState, {initialized: true, rows: action.data});
     case at.TOP_SITES_EDIT:
-      return Object.assign({}, prevState, {editForm: {visible: true, site: action.data}});
+      return Object.assign({}, prevState, {editForm: {visible: true, site: action.data}}, {screenshotRequestFailed: false});
```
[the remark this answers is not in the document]
piatra: Changed.

system-addon/content-src/components/TopSites/TopSite.jsx
piatra: Changed as suggested.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -1,24 +1,50 @@
...
+      url: Topsite ? Topsite.url : "",
+      customScreenshotUrl: Topsite ? Topsite.customScreenshotURL : "",
+      screenshotUrlValidationError: false,
+      validationError: false,
+      pendingScreenshotUpdate: false,
+      // Keep the preview screenshot data in the state to avoid showing
+      // cached results from previous requests.
+      screenshotPreview: null,
+      showCustomScreenshotForm: false,
+      screenshotRequestFailed: false
```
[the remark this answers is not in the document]
piatra: Current implementation adds:
In the TopSiteForm:
* customScreenshotUrl - this is more or less required for the input field to work + to compare with the prop and see if the input value changed
* pendingScreenshotUpdate - to shows the loading indicator
* screenshotPreview - this is a side effect of the `BroadcastToContent` (maybe I'm missing something): the user can exit the preview/edit flow by just refreshing the page so we don't have a chance to clear `props.Topsite.screenshotPreview` set in the Reducer. If you click edit again you will be presented with the previous preview image but the url associated is lost. Having this locally means it automatically gets cleared when the component dismounts.
* topsiteUrlValidationError & screenshotUrlValidationError - This way I can notify the TopSiteEditInputs component of validation errors (to show error labels & focus the elements) while at the same time keep the validation methods in the TopSiteForm component where they are needed more (add/save/preview functionality).

In TopSiteEditInputs
* showCustomScreenshotForm - to force the input to show

system-addon/content-src/components/TopSites/TopSiteForm.jsx
piatra: Created TopSiteEditInputs.jsx
* includes the url, title and custom url inputs
* handles the logic to display the error labels and to focus the inputs
* toggle to show hide the custom screenshot input 
* All the validation and the input change handler are still in the TopsiteForm.

system-addon/lib/Screenshots.jsm
piatra: Changed this. Now a preview request will trigger a `getScreenshotForURL` and we attach that in the Reducer to the Topsite as `screenshotPreview`.
If the user saves the form (it pins the site) we save the URL in the pinned pref as customScreenshotURL and upon the feed refresh we trigger `maybeCacheScreenshot` which adds `customScreenshot` to the LinksCache object that represents that Topsite. The `customScreenshot` is also migrated between links so we need to clear the value if the user updated the customScreenshotURL for the website.

system-addon/lib/TopSitesFeed.jsm
piatra: Now this never gets called if `customScreenshotURL` is set, instead if the customScreenshot is missing it will trigger a `maybeCacheScreenshot`.

system-addon/lib/TopSitesFeed.jsm
piatra: Replaced with `getScreenshotForURL`.

system-addon/lib/TopSitesFeed.jsm
piatra: the Broadcast has remained the same atm.

system-addon/lib/TopSitesFeed.jsm
piatra: s/set/get since the function now calls `getScreenshotForURL`.

locales/en-US/strings.properties
piatra: I'll bring this up for discussion.

k88hudson:
@Mardak @piatra I am still reading through, but I would think that any kind of preview-related  state change should stay in the local redux context (using plain actions without SendToMain etc.). It's possible that maybe a content-only reducer/part of a reducer is needed?)

system-addon/lib/Screenshots.jsm
```diff
@@ -12,6 +12,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "BackgroundPageThumbs",
   "resource://gre/modules/BackgroundPageThumbs.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "PageThumbs",
   "resource://gre/modules/PageThumbs.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PageThumbsStorage",
+  "resource://gre/modules/PageThumbs.jsm");
```
sarracini: where are you using PageThumbsStorage?

piatra: @sarracini the implementation changed a few times around and I forgot to remove it. Done. Thanks.

Mardak:
Are there updated designs or a description of how things should be changed?

piatra:
There is no one source of information but I will try to list everything.
Latest design https://mozilla.invisionapp.com/share/7DEP1VMYX#/screens/266461200_Top_Site_-_Edit_Image_
In addition to that:
* Adding the tab loading animation when fetching a url
* Adding an X button at the end to clear the input
  * spec here https://design.firefox.com/photon/components/input-fields.html#behaviors
* The behavior we want for the screenshot input is:
  * hidden by default (unless user value is set)
  * click the link to reveal it (no option/action to hide back)
* Add input button to default site context menu (without having to go through the "Edit Mode")

Desired but not implemented:
* Adding a fade effect for text overflow (similar how location bar is doing it)
  * https://searchfox.org/mozilla-central/rev/f6f1731b1b7fec332f86b55fa40e2c9ae67ac39b/browser/base/content/browser.css#561
  * but there is no spec for how the "x" and the fade would work together
  * the location bar achieves the effect without actually using an input element
  * not clear what other input fields should get the fade effect

Let me know if I skipped something.

piatra:
I've rebased the patch against the latest drag and drop work. Waiting on review.

I still think there is a bug in the LinksCache causing an issue with this PR, I've added a test (the one failing) that tries to capture the behavior.
https://github.com/mozilla/activity-stream/blob/84fd28714022c660a54ea1b3649efd259c4dcd8e/system-addon/lib/LinksCache.jsm#L109-L111
I think the problem is that `updateLink` always references the first `newLink` object created (even after the property is migrated) so calling `expire` and then `updateLink` will not update the newest object.

@Mardak / @k88hudson can you help validate this? What I’ve done is tag every link object with an incrementing id and then track who is being updated in the `updateLink` call. 

In this video https://ed.agadak.net/as/activity-stream.edit-preview.webm you do not see the bug because the sequence of events is:
* set custom avatar
  * (we are still working with the first link object created)
  * inside the pin method we calll `updateLink` to set `customScreenshot` to some "falsy" value to trigger a fetch
  * pinning happens that triggers an `expire` call
* refresh is called/getLinksWithDefaults
  * migration will copy over the "falsy" `customScreenshot` value
  * calls to `maybeCacheScreenshot` will not set the `customScreenshot` value on the correct object, but it will broadcast the image and refresh the UI
The custom image actually changes because the link object will always have an `undefined` custom screenshot (thus triggering a fetch every time). But if a different topsite is edited it won't work.

When you add a new Topsite `maybeCacheScreenshot` works because it sets the screenshot property on the newly created link object.

Here is another video actually showing it fail when we first pin a site (to trigger `expire`) and then try to edit https://drive.google.com/open?id=1KRyOoqb_Rg2UgFbktYZDhLlotFdKTOko

Mardak:
>  I think the problem is that updateLink always references the first newLink object created (even after the property is migrated) so calling expire and then updateLink will not update the newest object.

Ah indeed. Looks like we want to keep the original `__sharedCache` object (to maintain sharing) while updating the function to refer to the new link.

piatra:
Should this PR fix it or have it done separately?

Mardak:
This will need to be rebased on top of #3965.

Mardak:
Are there updated designs to reference? The previous "Latest design https://mozilla.invisionapp.com/share/7DEP1VMYX#/screens/266461200_Top_Site_-_Edit_Image_" no longer works

@aaronrbenson What's the color/decoration of the link/hover? And spacing between URL/custom should be the same as between title/url?

screen shot 2018-02-22 at 3 48 16 pm

aaronrbenson:
Yeah, sorry was cleaning up old projects in InVision. I sent @piatra the specs directly in slack .. attached here, too.

screen shot 2018-02-22 at 5 57 52 pm

screen shot 2018-02-22 at 5 58 45 pm

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -100,7 +106,7 @@ this.TopSitesFeed = class TopSitesFeed {
         {isDefault: !!notBlockedDefaultSites.find(finder)}, link, {hostname: shortURL(link)});
 
       // Add in favicons if we don't already have it
-      if (!copy.favicon) {
+      if (!copy.favicon && !copy.customScreenshotURL) {
```
Mardak: What is this guarding? This code path should be for pinned links inheriting properties from frecent links, but frecent links shouldn't have a custom screenshot anyway?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -131,7 +137,12 @@ this.TopSitesFeed = class TopSitesFeed {
     // Now, get a tippy top icon, a rich icon, or screenshot for every item
     for (const link of withPinned) {
       if (link) {
-        this._fetchIcon(link);
+        // If there is a custom screenshot this is the only image we display
+        if (link.customScreenshotURL && !link.customScreenshot) {
+          this._fetchCustomScreenshot(link);
+        } else {
```
Mardak: Does the previous condition actually want nested `if` blocks? In particular, if we have a custom screenshot url and already have the screenshot, we would end up fetching the icon?

Mardak: Alternatively, the existence check should happen inside `_fetch` similar to `_fetchIcon`

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -190,6 +201,31 @@ this.TopSitesFeed = class TopSitesFeed {
...
 
+/**
+ * Fetch, cache and broadcast a custom screenshot for a specific topsite.
+ * @param link cached topsite object
+ */
+  async _fetchCustomScreenshot(link) {
+    await Screenshots.maybeCacheScreenshot(link, link.customScreenshotURL, "customScreenshot",
+      customScreenshot => this.store.dispatch(ac.BroadcastToContent({
+        data: {customScreenshot, url: link.url},
+        type: at.SCREENSHOT_UPDATED
```
Mardak: This seems quite similar to the screenshot request from `_fetchIcon`, so maybe worth combining unless they actually should be doing something different.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -210,23 +246,41 @@ this.TopSitesFeed = class TopSitesFeed {
...
     const toPin = {url};
     if (label) {
       toPin.label = label;
     }
+    if (customScreenshotURL) {
+      toPin.customScreenshotURL = customScreenshotURL;
+    }
     NewTabUtils.pinnedLinks.pin(toPin, index);
+
+    await this._updateLinkCustomScreenshot({customScreenshotURL, url});
```
Mardak: Why not just pass in `toPin`?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -210,23 +246,41 @@ this.TopSitesFeed = class TopSitesFeed {
...
+  }
+
+  async _updateLinkCustomScreenshot(site) {
+    // If screenshot url changed or was removed we need to update the cached link obj
+    if (site.customScreenshotURL || site.customScreenshotURL === null) {
```
Mardak: Could this just be `!== undefined`?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -210,23 +246,41 @@ this.TopSitesFeed = class TopSitesFeed {
...
+    if (customScreenshotURL) {
+      toPin.customScreenshotURL = customScreenshotURL;
+    }
     NewTabUtils.pinnedLinks.pin(toPin, index);
+
+    await this._updateLinkCustomScreenshot({customScreenshotURL, url});
+  }
+
+  async _updateLinkCustomScreenshot(site) {
+    // If screenshot url changed or was removed we need to update the cached link obj
```
Mardak: This comment says "update" but more accurately it's clearing?

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -339,6 +331,58 @@ $half-base-gutter: $base-gutter / 2;
...
+      input:dir(rtl) {
+        padding-left: 32px;
+      }
+    }
+
+    .enable-custom-image-input {
+      display: inline-block;
+      font-size: 13px;
+      margin-top: 4px;
+      text-decoration: underline;
```
Mardak: Does this actually want underline? Our other link-like things don't seem to have that styling.

system-addon/common/Reducers.jsm
```diff
@@ -91,18 +91,26 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       return Object.assign({}, prevState, {initialized: true, rows: action.data});
     case at.TOP_SITES_EDIT:
-      return Object.assign({}, prevState, {editForm: {index: action.data.index}});
+      return Object.assign({}, prevState, {editForm: {index: action.data.index}, screenshotRequestFailed: false, screenshotPreview: null});
     case at.TOP_SITES_CANCEL_EDIT:
-      return Object.assign({}, prevState, {editForm: null});
+      return Object.assign({}, prevState, {editForm: null, screenshotRequestFailed: false});
+    case at.SCREENSHOT_FAILED:
+      return Object.assign({}, prevState, {screenshotRequestFailed: true});
```
Mardak: I wonder if some confusion is from a general `SCREENSHOT_FAILED` that is only really used for *preview* screenshots -> `PREVIEW_FAILED`?

system-addon/content-src/components/TopSites/_TopSites.scss
piatra: That's how the spec is.. or was?
image

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -67,7 +67,17 @@ export class TopSiteLink extends React.PureComponent {
     let showSmallFavicon = false;
     let smallFaviconStyle;
     let smallFaviconFallback;
-    if (tippyTopIcon || faviconSize >= MIN_RICH_FAVICON_SIZE) {
+    if (link.screenshotRequestFailed) {
+      showSmallFavicon = false;
+      smallFaviconFallback = false;
+    } else if (link.screenshotPreview || link.customScreenshot) {
```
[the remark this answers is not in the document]
piatra: Maybe I'm missing something, you should need it if you try a new preview for a link that has a custom screenshot. You want the preview to take precedence but then fallback to the custom one if the user removes the url used for preview.

system-addon/content-src/components/TopSites/_TopSites.scss
Mardak: I was also trying to get a recent spec last week, and aaron said he sent them to you on slack?

system-addon/common/Reducers.jsm
```diff
@@ -91,18 +91,26 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       return Object.assign({}, prevState, {initialized: true, rows: action.data});
     case at.TOP_SITES_EDIT:
-      return Object.assign({}, prevState, {editForm: {index: action.data.index}});
+      return Object.assign({}, prevState, {editForm: {index: action.data.index}, screenshotRequestFailed: false, screenshotPreview: null});
```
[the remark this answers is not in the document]
piatra: > the url being requested to be associated with the preview somehow
If you add a new topsite you can do a "preview request" without adding in the URL first.
Instead of having two different ways to handle for this I had this capture all scenario.
Should we still try and fix it?

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -67,7 +67,17 @@ export class TopSiteLink extends React.PureComponent {
     let showSmallFavicon = false;
     let smallFaviconStyle;
     let smallFaviconFallback;
-    if (tippyTopIcon || faviconSize >= MIN_RICH_FAVICON_SIZE) {
+    if (link.screenshotRequestFailed) {
+      showSmallFavicon = false;
+      smallFaviconFallback = false;
```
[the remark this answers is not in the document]
piatra: The prop could use a rename.
The purpose is to have this check first and prevent any className/style properties from being set. A topsite could have a tippytop icon but we don't want to show it for a failed request.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -12,13 +12,35 @@ export class TopSiteForm extends React.PureComponent {
...
+    this.onEnableScreenshotUrlForm = this.onEnableScreenshotUrlForm.bind(this);
+    this.validateUrl = this.validateUrl.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if ((nextProps.screenshotPreview && this.state.pendingScreenshotUpdate) ||
+        nextProps.screenshotRequestFailed) {
+      this.setState({
+        pendingScreenshotUpdate: false,
+        validationError: nextProps.screenshotRequestFailed,
```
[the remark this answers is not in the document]
piatra: You're right I can probably make use of just the `state.screenshotRequestFailed`

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -12,13 +12,35 @@ export class TopSiteForm extends React.PureComponent {
...
+        screenshotRequestFailed: nextProps.screenshotRequestFailed,
+        screenshotPreview: nextProps.screenshotPreview
```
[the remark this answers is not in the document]
piatra: The problem I was trying to work around here is that I want to use`props.screenshotRequestFailed` but if the user changes the value of the URL that prop value is obsolete and there is no way to clear it/know that it was used for a different request.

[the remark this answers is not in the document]
piatra: Scenario
1. User has a custom screenshot
2. User previews a different screenshot and request fails
3. User resets to the initial custom url value
The button should say save, but because `props.screenshotRequestFailed` is true it will say "Preview" or prevent saving.
And there might be other scenarios (like user removes custom url completely). You could work around (and I tried) but I ended up having a lot of boolean operations that were not easy to reason about.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -84,15 +150,65 @@ export class TopSiteForm extends React.PureComponent {
...
+    const {customScreenshotUrl} = this.state;
+    const {site} = this.props;
     // For UI purposes, editing without an existing link is "add"
-    const showAsAdd = !this.props.site;
+    const showAsAdd = !site;
+    const changedCustomScreenshotUrl = site && customScreenshotUrl &&
+      (site.customScreenshotURL !== this.cleanUrl(customScreenshotUrl));
+    // Preview mode enables the preview button and prevents saving or adding a topsite.
+    const previewMode = (customScreenshotUrl && !this.state.screenshotPreview) &&
+                        (showAsAdd || changedCustomScreenshotUrl);
```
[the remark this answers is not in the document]
piatra: Maybe the condition could use some refactoring. This is what I was trying to say:
* If we have a custom url and no preview it could mean that:
  * we are `edit`ing a topsite, so enable "Preview" if the value of the custom url changed
  * we are `edit`ing a topsite but no change to the url happend so show "Save"
  * we are `add`ing a topsite, we always want a "Preview" button if we don't have a preview

I couldn't find a way to include the last case in the other two (while keeping a single `previewMode` check/variable). We could further look at the site.customScreenshot value and see if it's set, but it would just replace one boolean for another.
Any thoughts?

system-addon/lib/TopSitesFeed.jsm
piatra: You're right. This is done as a map over pinned items and I think I just got confused.

system-addon/content-src/components/TopSites/_TopSites.scss
piatra: I can double check with Aaron.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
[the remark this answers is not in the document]
piatra: Just noticed `this.state.screenshotRequestFailed` is not currently used to decide the previewMode as I described. Fixing this.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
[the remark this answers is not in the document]
piatra: Maybe
```
customScreenshotUrl && !this.state.screenshotPreview && (site && !site.customScreenshot);
```
I'll looking into it some more.

locales/en-US/strings.properties
```diff
@@ -153,10 +153,15 @@ topsites_form_title_label=Title
 topsites_form_title_placeholder=Enter a title
 topsites_form_url_label=URL
 topsites_form_url_placeholder=Type or paste a URL
+topsites_form_image_label=Custom Image
+topsites_form_image_placeholder=Paste a URL here
```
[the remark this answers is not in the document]
piatra: image

system-addon/common/Reducers.jsm
```diff
@@ -91,18 +91,26 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
...
+          const {screenshot, customScreenshot} = action.data;
+          if (screenshot) {
+            return Object.assign({}, row, {screenshot});
+          }
+          return Object.assign({}, row, {customScreenshot});
         }
         return row;
       });
-      return hasMatch ? Object.assign({}, prevState, {rows: newRows}) : prevState;
+      return hasMatch ? Object.assign({}, prevState, {rows: newRows, screenshotRequestFailed: false}) : prevState;
```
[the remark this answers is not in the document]
piatra: I was initially handling preview requests as you described in [this comment](https://github.com/mozilla/activity-stream/pull/3891#discussion_r170839055) but then ran into the issue that I described where "Add" doesn't have an associated row entry/link object. This was just left over from that.

locales/en-US/strings.properties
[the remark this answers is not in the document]
Mardak: Did you get confirmation that this new one should have "here" and the other one should not? It's hard to review let alone implement without an accurate spec and keep track of what was decided or happened to be "implemented".

For example the screenshots from https://github.com/mozilla/activity-stream/pull/3891#issuecomment-367864260 have both "Custom Image" and "Custom Image URL" that needed to be clarified.

system-addon/content-src/components/TopSites/TopSite.jsx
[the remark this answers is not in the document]
Mardak: It looks like `screenshotPreview` is only set via the `Form`, so why not just have `_getPreviewScreenshot()` set `customScreenshot`? I.e., why does a `TopSite` component need to know that one is a preview and the other is a custom screenshot?

system-addon/content-src/components/TopSites/TopSite.jsx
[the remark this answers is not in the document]
Mardak: Similar to the previous comment of "why does a `TopSite` component need to know" or expose anything about screenshot requesting, yeah a rename of the property to something like "showLetter" would make more sense as a component API assuming that's what we want. Alternatively, the caller could just not pass in the properties (e.g., favicon) that would have triggered the other code paths to not complicate this component.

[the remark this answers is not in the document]
Mardak: Actually, given that we're already generating a custom site object to pass in via `_getPreviewScreenshot()`, it's probably better to not set things than overriding with an additional property. (The overriding is somewhat similar to "fixing" CSS with `!important`.)

system-addon/common/Reducers.jsm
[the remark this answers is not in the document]
Mardak: In another comment https://github.com/mozilla/activity-stream/pull/3891#discussion_r170874864, you mentioned issues around needing to know whether the user has reverted back to the original custom image after a failed screenshot. If we tracked the preview url request / response failure and had the original site object, we would be able to determine that the failed screenshot is for a different preview url and that the current typed in preview url is the original value.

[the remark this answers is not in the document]
piatra: > Should the screenshot stuff not be part of editForm? 

Yes

> I was referring to the preview url being requested not the url of the site

Sanity check: when replying with the screenshot preview data we should send the preview URL as well to check that we are replying to the correct request

I have replied to something completely different previously.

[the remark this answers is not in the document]
Mardak: > Yes

What's the reasoning for not having screenshot/preview stuff inside `editForm` state?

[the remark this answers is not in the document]
Mardak: > Sanity check: when replying with the screenshot preview data we should send the preview URL as well to check that we are replying to the correct request?

Right. I'm not sure if we need to save both the request and response as theoretically we could just ignore responses that aren't for the current request. E.g., when making a request, the page's redux state updates with the requested preview url and also sends to main. If there's a second request to main, the page's state again updates, and now there's 2 outstanding preview requests. I don't think backgroundpagethumbs exposes a way to cancel easily, so let's just let both requests finish for now, and the response has the thumbnail data and the requested url, but only the response that we now care about (the second url, which might have responded sooner).

Mardak:
Looking through previous comments from @k88hudson https://github.com/mozilla/activity-stream/pull/3891#issuecomment-350368439:
> @Mardak @piatra I am still reading through, but I would think that any kind of preview-related state change should stay in the local redux context (using plain actions without SendToMain etc.). It's possible that maybe a content-only reducer/part of a reducer is needed?)

I think we'll probably need a content-only reducer now, as we don't want preview related state to be persisted in main and then sent to other content.

That would mean adding a content-only reducer somewhere to the reducer chain?
https://github.com/mozilla/activity-stream/blob/f3e33da50e0bea2fad0a38a985c5eab1d98eb38a/system-addon/content-src/activity-stream.jsx#L11

system-addon/content-src/components/TopSites/TopSiteFormInput.jsx
```diff
@@ -19,17 +31,22 @@ export class TopSiteFormInput extends React.PureComponent {
...
+    const {typeUrl} = this.props;
+    const {validationError} = this.state;
 
     return (<label><FormattedMessage id={this.props.titleId} />
       <div className={`field ${typeUrl ? "url" : ""}${validationError ? " invalid" : ""}`}>
-        {showClearButton &&
+        {this.props.loading && <div className="loading-container">
+          <div className="loading-animation" />
+        </div>}
+        {!this.props.loading && showClearButton &&
```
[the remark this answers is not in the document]
piatra: > [eslint] Do not nest ternary expressions. (no-nested-ternary)

But I can get a similar style.

piatra:
@Mardak isn't that what [OnlyToOneContent](https://github.com/mozilla/activity-stream/blob/f3e33da50e0bea2fad0a38a985c5eab1d98eb38a/system-addon/common/Actions.jsm#L189) does? It was added after that comment.

Mardak:
There's still some open questions, but if we want to land this for 60, we'll want to get the strings in sooner, but that seems like it's still waiting confirmation from design?

Mardak:
> @Mardak isn't that what OnlyToOneContent does? It was added after that comment.

Depending on how we handle saving the request url / response url, we wouldn't want the main state to update via `Reducers.jsm` because we happen to want content to update when it makes a request. I suppose normally a component state could be used to remember what was requested, but in this case, we're also waiting on a response from main to decide if we should "accept" the response.

locales/en-US/strings.properties
```diff
@@ -153,10 +153,15 @@ topsites_form_title_label=Title
 topsites_form_title_placeholder=Enter a title
 topsites_form_url_label=URL
 topsites_form_url_placeholder=Type or paste a URL
+topsites_form_image_label=Custom Image
```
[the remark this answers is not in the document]
Mardak: @flodolo Just a heads up that there's these 5 strings that we'll try to get in for 60 and hopefully at least land the strings in activity-stream-l10n today/Friday. These and the ones from #4015 should be the last strings for activity-stream-l10n in 60.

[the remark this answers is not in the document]
flodolo: Can you add a CODEOWNERS file? See [this example](https://github.com/mozilla/testpilot/blob/master/CODEOWNERS). That way I'll get pinged for any PR trying to change the .properties file

You might need to add me as a collaborator (not completely sure), even with read access should be enough.

locales/en-US/strings.properties
```diff
@@ -153,10 +153,15 @@ topsites_form_title_label=Title
...
+topsites_form_image_placeholder=Paste a URL here
+topsites_form_image_button=Preview
```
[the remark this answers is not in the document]
flodolo: It would also be useful to add a comment explaining this is a verb, and used as a button label, just in case the ID is not obvious enough

locales/en-US/strings.properties
```diff
@@ -153,10 +153,15 @@ topsites_form_title_label=Title
...
+topsites_form_image_enable_button=Use a custom image…
```
Mardak: I ended up renaming this in a WIP if we were to have landed strings sooner. But sounds like we'll do it Monday.

```diff
 topsites_form_url_placeholder=Type or paste a URL
+topsites_form_use_image_link=Use a custom image…
+topsites_form_image_label=Custom Image
+topsites_form_image_placeholder=Paste a URL here
+# LOCALIZATION NOTE (topsites_form_*_button): These are verbs/actions.
+topsites_form_preview_button=Preview
 topsites_form_add_button=Add
 topsites_form_save_button=Save
 topsites_form_cancel_button=Cancel
 topsites_form_url_validation=Valid URL required
+topsites_form_image_validation=Image failed to load. Try a different URL.
 
```

system-addon/content-src/components/TopSites/_TopSites.scss
Mardak: Bryan said this should always be blue whether hovered or not. And no underline normally but underlined on hover.

locales/en-US/strings.properties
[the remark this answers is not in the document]
Mardak: Bryan said this should be the same as the URL's placeholder.

locales/en-US/strings.properties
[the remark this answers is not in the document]
Mardak: Bryan also confirmed this should be "Custom Image URL"

system-addon/lib/TopSitesFeed.jsm
piatra: we need to pass the explicit `null` value of `customScreenshotURL`

Mardak:
Looks like there's still comments unaddressed or replies to how it was addressed.

Here's a video of the current state of the PR of doing a preview request for "google.comm" then midway through changing it to "com" then having the preview fail resulting in blank:
mar-12-2018 13-14-24

piatra:
@Mardak Regarding [the comment](https://github.com/mozilla/activity-stream/pull/3891#issuecomment-369796199)
> Depending on how we handle saving the request url / response url [...]

I don't understand the concerns: when replying to the request we use `OnlyToOneContent` that skips the main reducer and `component state could be used to remember` is not needed because the `target` parameter takes care of that. Could you explain the problem because I'm not sure what I'm suppose to fix.

system-addon/common/Reducers.jsm
piatra: Renamed this to PREVIEW_FAILED

system-addon/content-src/components/TopSites/_TopSites.scss
piatra: Removed the underline.

system-addon/lib/TopSitesFeed.jsm
piatra: Renamed to `_clearLinkCustomScreenshot`

system-addon/lib/TopSitesFeed.jsm
piatra: created `_fetchScreenshot` used to get the custom urls and the regular urls (in `_fetchIcon`)

system-addon/lib/TopSitesFeed.jsm
piatra: Kept the check for customScreenshotURL,`_fetchIcon` is not needed if a custom url is present.

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -339,6 +331,58 @@ $half-base-gutter: $base-gutter / 2;
...
+        @keyframes tab-throbber-animation-rtl {
+          100% { transform: translateX(960px); }
+        }
+
+        width: 960px;
+        height: 16px;
+        -moz-context-properties: fill;
+        fill: $teal-70;
+        background-image: url('chrome://browser/skin/tabbrowser/loading.svg');
+        animation: tab-throbber-animation 1.05s steps(60) infinite;
```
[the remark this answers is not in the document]
piatra: Added bug number in a comment.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -84,16 +152,61 @@ export class TopSiteForm extends React.PureComponent {
...
   render() {
+    const {customScreenshotUrl} = this.state;
     // For UI purposes, editing without an existing link is "add"
     const showAsAdd = !this.props.site;
-
+    const changedCustomScreenshotUrl = customScreenshotUrl &&
+      (showAsAdd || (this.props.site.customScreenshotURL !== this.cleanUrl(customScreenshotUrl)));
+    // Preview mode enables the preview button and prevents saving or adding a topsite.
+    const previewMode = (changedCustomScreenshotUrl && !this.state.screenshotPreview) ||
+      this.state.screenshotRequestFailed;
```
Mardak: Probably just have the comment describe the condition we're trying to match especially because it seems like the logic had to be fixed up here a few times to get right, but I'm not even sure if this is right?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -12,13 +12,36 @@ export class TopSiteForm extends React.PureComponent {
...
+
+  componentWillReceiveProps(nextProps) {
+    // Check the response URL against the current state of the form
+    const pendingRequest = this.state.customScreenshotUrl && this.state.pendingScreenshotUpdate &&
+      nextProps.customScreenshotURL === this.cleanUrl(this.state.customScreenshotUrl);
+    if (pendingRequest || nextProps.screenshotRequestFailed) {
+      this.setState({
+        pendingScreenshotUpdate: false,
+        screenshotRequestFailed: nextProps.screenshotRequestFailed,
+        screenshotPreview: nextProps.screenshotPreview
```
Mardak: I've mentioned that having a copy of a prop as a state doesn't seem right and makes it hard to reason about when to use the value from state vs from prop especially when they're named exactly the same. It looks like the reason why it's in the state is so that it can be cleared from `onClearScreenshotInput` to get out of preview mode. Why couldn't `onClear` trigger something that makes the prop no longer be set, i.e., clear the local redux state.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -68,6 +120,22 @@ export class TopSiteForm extends React.PureComponent {
     }
   }
 
+  onPreviewButtonClick(event) {
+    event.preventDefault();
+    if (this.validateForm()) {
+      this.setState({pendingScreenshotUpdate: true});
```
Mardak: Any reason to have this in the component state vs setting true in redux probably by PREVIEW_REQUEST and cleared on SCREENSHOT_PREVIEW/PREVIEW_FAILED?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
Mardak: In particular, I would have expected some check for pending? And pending should be set only if we haven't received a preview/failure for the most recently requested url?

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -120,6 +124,9 @@ export class _TopSites extends React.PureComponent {
               <div className="modal">
                 <TopSiteForm
                   site={props.TopSites.rows[editForm.index]}
+                  screenshotPreview={this.props.TopSites.editForm.screenshotPreview}
+                  screenshotRequestFailed={this.props.TopSites.editForm.screenshotRequestFailed}
+                  customScreenshotURL={this.props.TopSites.editForm.customScreenshotURL}
```
Mardak: Maybe this seems to point to just directly connecting TopSiteForm to the store instead of passing it through this component?

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -337,20 +337,69 @@ $half-base-gutter: $base-gutter / 2;
...
+        width: 16px;
+        height: 16px;
+        overflow: hidden;
+        position: absolute;
+        transform: translateY(-50%);
+        top: 50%;
+        offset-inline-end: 8px;
+      }
+
+      // See Bug 1352119
```
Mardak: This comment isn't particularly useful as is. What's the bug for? Is it to fix something undesired here?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
piatra: I'm no longer copying the props in the state and added a new action creator to modify the local reducer.
Removed `pendingScreenshotUpdate` and replaced it with `requestScreenshotURL` which is `true` at the same time + I can use it to validate the correct request is received.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
piatra: >  I would have expected some check for pending

You have to be in `PreviewMode` in order to trigger a request and get to pending. Unless you meant something else.

piatra:
Updated to address the comments about state vs props. Moved part of the props in the reducer and using actions to clear/set.
I did not update the tests to save time in case there are more changes to the implementation.

system-addon/common/Actions.jsm
```diff
@@ -139,6 +140,18 @@ function AlsoToMain(action, fromTarget, skipLocal) {
   });
 }
 
+/**
+ * OnlyToThisContent - Creates a message that will only be dispatched locally.
+ *
+ * @param {object} action Any redux action (required)
+ */
+function OnlyToThisContent(action) {
```
Mardak: Is this necessary? Dispatch by default doesn't go elsewhere.

system-addon/common/Reducers.jsm
```diff
@@ -97,16 +97,43 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       return Object.assign({}, prevState, {
         editForm: {
           index: prevState.editForm.index,
-          screenshotRequestFailed: true
+          screenshotRequestFailed: true,
+          requestedScreenshotURL: null
```
Mardak: Why do we need to clear the requestedScreenshotURL? I would have expected this FAILED action to check if the failing preview was actually for the screenshot being requested. Similar to the check for SCREENSHOT_PREVIEW.

system-addon/common/Reducers.jsm
```diff
@@ -97,16 +97,43 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
...
+            index: prevState.editForm ? prevState.editForm.index : -1,
+            screenshotRequestFailed: false,
+            screenshotPreview: action.data.screenshotPreview,
+            customScreenshotURL: action.data.customScreenshotURL,
+            requestedScreenshotURL: null
+          }
+        });
+      }
+      return prevState;
+    case at.PREVIEW_REQUEST:
```
Mardak: Because these Reducers are also run in main, are there concerns with the main redux state updating and sending editForm state in rehydration messages?

system-addon/common/Actions.jsm
piatra: oops I didn't realize that.

system-addon/common/Reducers.jsm
piatra: I don't think that should happen. `PREVIEW_REQUEST` checks `prevState.editForm` which is null for the main reducer so `editForm` should always be null. Do you think we should have separate `request` actions (one to trigger the fetch in main, another to update the local reducer) to make the separation clearer?

system-addon/common/Reducers.jsm
piatra: As long as `requestedScreenshotURL` is set, we show the loading/pending state. Added the check as mentioned.

system-addon/common/Reducers.jsm
Mardak: Ah good point. I think this is a neat way to avoid updating main. Where TOP_SITES_EDIT to trigger reducing these actions is only seen in content.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -108,32 +199,40 @@ export class TopSiteForm extends React.PureComponent {
...
                 titleId="topsites_form_url_label"
                 typeUrl={true}
                 placeholderId="topsites_form_url_placeholder"
                 errorMessageId="topsites_form_url_validation"
                 intl={this.props.intl} />
+              {this._renderCustomScreenshotInput()}
             </div>
-            <TopSiteLink link={this.props.site || {}} title={this.state.label} />
+            <TopSiteLink link={this._getPreviewScreenshot()}
+              defaultStyle={this.props.screenshotRequestFailed}
```
Mardak: Did design confirm that a failed preview should show the letter fallback vs revert to what it showed before the preview attempt?

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -337,20 +337,70 @@ $half-base-gutter: $base-gutter / 2;
...
+        width: 16px;
+        height: 16px;
+        overflow: hidden;
+        position: absolute;
+        transform: translateY(-50%);
+        top: 50%;
+        offset-inline-end: 8px;
+      }
+
+      // See https://searchfox.org/mozilla-central/rev/3abf6fa7e2a6d9a7bfb88796141b0f012e68c2db/browser/themes/shared/tabs.inc.css#191
```
Mardak: This comment still isn't useful. It still requires someone to check the reference to figure out what it's even talking about. Maybe something like "This animation is derived from Firefox's tab loading animation: …"

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -180,14 +191,36 @@ this.TopSitesFeed = class TopSitesFeed {
...
+  /**
+   * Dispatch screenshot preview to target or notify if request failed.
+   * @param customScreenshotURL {string} The URL used to capture the screenshot
+   * @param target {string} Id of content process where to dispatch the result
+   */
+  async getScreenshotPreview({customScreenshotURL}, target) {
+    const screenshotPreview = await Screenshots.getScreenshotForURL(customScreenshotURL);
+    this.store.dispatch(ac.OnlyToOneContent({
+      data: {screenshotPreview, customScreenshotURL},
+      type: screenshotPreview ? at.SCREENSHOT_PREVIEW : at.PREVIEW_FAILED
```
Mardak: To match PREVIEW_REQUEST's {url}, this should dispatch `data: {url, preview}, type: PREVIEW_RESPONSE` where `preview` is `""` for failed.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -339,6 +392,9 @@ this.TopSitesFeed = class TopSitesFeed {
       case at.TOP_SITES_INSERT:
         this.insert(action);
         break;
+      case at.PREVIEW_REQUEST:
+        this.getScreenshotPreview(action.data, action.meta.fromTarget);
```
Mardak: Just pass in `action.data.url`

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -67,7 +67,16 @@ export class TopSiteLink extends React.PureComponent {
     let showSmallFavicon = false;
     let smallFaviconStyle;
     let smallFaviconFallback;
-    if (tippyTopIcon || faviconSize >= MIN_RICH_FAVICON_SIZE) {
+    if (defaultStyle) { // Render the Topsite without any image or icon
+      smallFaviconFallback = false;
+    } else if (link.screenshotPreview || link.customScreenshotURL) {
+      // styles and class names for top sites with rich icons
```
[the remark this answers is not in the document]
Mardak: Or why we're using this for a screenshot?

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -120,10 +124,10 @@ export class _TopSites extends React.PureComponent {
               <div className="modal">
                 <TopSiteForm
                   site={props.TopSites.rows[editForm.index]}
-                  index={editForm.index}
                   onClose={this.onFormClose}
                   dispatch={this.props.dispatch}
-                  intl={this.props.intl} />
+                  intl={this.props.intl}
+                  {...editForm} />
```
Mardak: Ha. I guess this works. Could be dangerous, but fine for now…

system-addon/content-src/components/TopSites/TopSiteFormInput.jsx
```diff
@@ -19,17 +35,21 @@ export class TopSiteFormInput extends React.PureComponent {
...
+          showClearButton && <div className="icon icon-clear-input" onClick={this.props.onClear} />}
         <input type="text"
           value={this.props.value}
           ref={this.onMount}
-          onChange={this.props.onChange}
-          placeholder={this.props.intl.formatMessage({id: this.props.placeholderId})} />
+          onChange={this.onChange}
+          placeholder={this.props.intl.formatMessage({id: this.props.placeholderId})}
+          autoFocus={this.props.shouldFocus}
+          disabled={this.props.loading} />
```
Mardak: Did design confirm that the input should be disabled? Otherwise the only way to interrupt seems to be to click cancel for the whole dialog. Why does this need to be disabled in the first place?

piatra: I thought that is what you were trying to show with the [gif from the comment](https://github.com/mozilla/activity-stream/pull/3891#issuecomment-372448047) but canceling the request also works. I'll check with design.

Mardak: I was showing that the wrong preview result could be shown, yes. But now that we're checking whether the response/screenshot/failure is for the requested url, it doesn't need to be disabled.

piatra: I mentioned the concerns about canceling, the decision was to keep the disabled state.

> i think disabling it is the right thing to do
> unlikely that someone will want to change it mid-stream
> but disabling the input would prevent any accidental typos

Mardak: Sounds like design wants to prevent the user from changing it even though the user might want to change it before the 30 seconds is up.

Mardak:
A separate issue that design said to follow up later is a way to restore the original custom image url instead of only being able to clear it.

piatra:
I'm not sure if https://github.com/mozilla/activity-stream/pull/3891/commits/d7d40faa548eb7255c9db0b83cda0603cd91f136 is what you meant in [this comment](https://github.com/mozilla/activity-stream/pull/3891#discussion_r175989153) I didn't understand the reference to TOP_SITES_EDIT
> Ah good point. I think this is a neat way to avoid updating main. Where TOP_SITES_EDIT to trigger reducing these actions is only seen in content.

There might be a weird edge case here (I'm still thinking about it):
* `PREVIEW_RESPONSE` sets props `(previewResponse: "data", previewUrl: "url")` 
* Deleting a character will send `PREVIEW_REQUEST_CANCEL` => `(previewResponse: null, previewUrl: "url")`
* Add that character back => you match the loading condition & trigger the loading animation
So what I did was remove `previewUrl` in the  `PREVIEW_REQUEST_CANCEL` action

Still waiting to hear on the decision about [reverting from an error state](https://github.com/mozilla/activity-stream/pull/3891#discussion_r176071574)

system-addon/content-src/components/TopSites/TopSiteForm.jsx
Mardak: I showed Aaron a screenshot of a failed request (red message, letter fallback), and he said "perfect. seems clear to me" So this should be good.

system-addon/common/Reducers.jsm
Mardak: Ah I see the confusion. I meant "good point" referring to your first 2 sentences. I don't think there should be a separate request action for main.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -84,16 +131,66 @@ export class TopSiteForm extends React.PureComponent {
...
+    const {customScreenshotUrl} = this.state;
+    const requestFailed = this.props.previewResponse === "";
     // For UI purposes, editing without an existing link is "add"
     const showAsAdd = !this.props.site;
-
+    const previous = (this.props.site && this.props.site.customScreenshotURL) || "";
+    const changed = customScreenshotUrl && this.cleanUrl(customScreenshotUrl) !== previous;
+    // Preview mode if changes were made to the custom screenshot URL and no preview was received yet
+    // or the request failed
+    const previewMode = (changed && this.props.previewResponse === null) || requestFailed;
```
Mardak: This should be able to be just `previewMode = changed && !this.props.previewResponse` If the value wasn't changed, we shouldn't be looking for a preview request.

Mardak: (The input change dispatching PREVIEW_REQUEST_CANCEL causes requestFailed to get cleared.)

system-addon/common/Reducers.jsm
```diff
@@ -90,9 +90,47 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
...
     case at.TOP_SITES_CANCEL_EDIT:
       return Object.assign({}, prevState, {editForm: null});
+    case at.PREVIEW_RESPONSE:
+      if (prevState.editForm &&
+        prevState.editForm.previewUrl === action.data.url) {
+        return Object.assign({}, prevState, {
+          editForm: {
+            index: prevState.editForm.index,
+            previewResponse: action.data.preview,
+            previewUrl: action.data.url
```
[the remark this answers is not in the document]
piatra: Had that at first but makes it more difficult to read. I also think it helps in this case to be explicit about what we want our props to be (for debugging/refactoring reasons).

PR #3894, opened 2017-12-05
Fix Bug 1421424 - chore(testing): update Enzyme to v3 in prep for React 16

dmose:
Updates to latest version of Enzymem which we need for more React 16 work.  I'll email out a copy of the [migration guide](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md) to the activity-stream-dev mailing list once this merges, because it contains breaking API changes, some of which are rather subtle.


Mardak:
The individual changes seem fine as per the enzyme migration guide. But one testing change here is that by default lifecycle methods are run:
https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#lifecycle-methods

There's an option to change the default to be disabled as we have been testing, and we do have some explicit `instance().componentDidMount()` tests.

@k88hudson any preference on lifecycle running by default or not?

dmose:
FWIW, my gut is that going with the new default is a better choice, since it means that the default is to test something that acts more like the real thing.  But I don't feel strongly about that.

dmose:
Admittedly, though, doing it the new way like I have in the PR does change the unit of code under test mostly behind the back of the existing tests.

Mardak:
Right, where lifecycle on by by default means it should catch any bugs in didMount just by mounting instead of when we explicitly call it.

It would also probably mean higher reported code coverage (?) for components as in the code got executed, but maybe that means people might skip writing a unit test that explicitly checks the lifecycle method? But we don't get code coverage for .jsx (?) files right now…

dmose:
> Right, where lifecycle on by by default means it should catch any bugs in didMount just by mounting instead of when we explicitly call it.

Well, I dunno about any bugs, but at least ones where things explode.

> It would also probably mean higher reported code coverage (?) for components as in the code got executed, but maybe that means people might skip writing a unit test that explicitly checks the lifecycle method? But we don't get code coverage for .jsx (?) files right now…

I'd be awfully curious to know how many people look at the code coverage info much at all, since it's not something that just appears in github PRs...

Another nice thing about the new behavior is that, presumably, existing docs and examples on the web will migrate towards assuming that's the config that tests will be written with, merely because it's the default.

Mardak:
Our tests currently fail if there's not 100% code coverage of the things being measured. Turns out components aren't being measured. ;)

dmose:
@k88hudson said that [she's in favor of adopting the new convention](https://github.com/mozilla/activity-stream/pull/3857#issuecomment-349741028).  Since that's how the patch is already written, @Mardak, do you have any more comments/requests on this, or is it ready to merge?

as-pine-proxy:
[Treeherder: 3cfa46f6a3ee7fffe6480190918f6ebd41e70325](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=3cfa46f6a3ee7fffe6480190918f6ebd41e70325)

PR #3895, opened 2017-12-06
fix (topsites): dont collapse holes in Top Sites

rlr:
I separated this out from my drag and drop changes. I wasn't sure if this is a bug or a feature, but I think we need to "honor" holes for drag and drop to work properly (so the user can pin a site over any placeholder past their frecent+default top sites).

r? @Mardak

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -180,6 +180,30 @@ TopSite.defaultProps = {
 
 const TopSitePlaceholder = () => <TopSiteLink className="placeholder" />;
 
+const TopSiteList = props => {
+  const topSites = props.TopSites.rows.slice(0, props.TopSitesCount);
+  const topSitesUI = [];
+  for (let i = 0, l = props.TopSitesCount; i < l; i++) {
+    const link = topSites[i];
+    topSitesUI.push(!link ? <TopSitePlaceholder key={i} /> : <TopSite
+      key={link.guid || link.url}
+      dispatch={props.dispatch}
+      link={link}
+      index={i}
+      intl={props.intl}
+      onEdit={props.onEdit}
+      editMode={props.editMode} />);
+  }
+  return (<ul className="top-sites-list">
+    {topSitesUI}
+  </ul>);
+};
+TopSiteList.defaultProps = {
+  editMode: false,
+  onEdit() {}
+};
+
 module.exports.TopSite = TopSite;
 module.exports.TopSiteLink = TopSiteLink;
 module.exports.TopSitePlaceholder = TopSitePlaceholder;
+module.exports.TopSiteList = TopSiteList;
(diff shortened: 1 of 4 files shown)


rlr:
Before:

screen shot 2017-12-06 at 1 35 26 pm

After:

screen shot 2017-12-06 at 1 35 02 pm

system-addon/content-src/components/TopSites/TopSitesEdit.jsx
```diff
@@ -71,14 +71,29 @@ class TopSitesEdit extends React.PureComponent {
...
+          dispatch={this.props.dispatch}
+          link={link}
+          index={i}
+          intl={this.props.intl}
+          onEdit={this.onEdit}
+          editMode={true} />);
+      } else {
+        topSitesUI.push(<TopSitePlaceholder key={i} />);
+      }
+    }
```
rlr: Thinking about how to DRY this up a little. I think a new `TopSiteList` component that just does this should work. *tries*

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -71,22 +71,28 @@ class TopSites extends React.PureComponent {
...
+      const link = topSites[i];
+      if (link) {
+        topSitesUI.push(<TopSite
+          key={link.guid || link.url}
+          dispatch={props.dispatch}
+          link={link}
+          index={i}
+          intl={props.intl} />);
+      } else {
+        topSitesUI.push(<TopSitePlaceholder key={i} />);
```
Mardak: nit: could do `topSitesUI.push(!link ? <TopSitePlaceholder… > : …)`

A quick test also verifies that this does what it should do, but yeah some refactoring would be much nicer

rlr:
@Mardak ^ way nicer now. That should help me refactor the DnD stuff nicely too.

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -180,6 +180,30 @@ TopSite.defaultProps = {
...
+      onEdit={props.onEdit}
+      editMode={props.editMode} />);
+  }
+  return (<ul className="top-sites-list">
+    {topSitesUI}
+  </ul>);
+};
+TopSiteList.defaultProps = {
+  editMode: false,
+  onEdit() {}
```
Mardak: Without a `defaultProps` for `TopSiteList`, if there's no `onEdit` provided, it should probably imply `editMode: false`? In particular, `TopSiteList` could not need `editMode` and just take an option `onEdit` to unconditionally pass over to `TopSite` which would be updated to conditionally render `LinkMenu` or the `edit-menu` (without its own `defaultProps`) ?

rlr: 👍 I'll do this in a followup PR.

as-pine-proxy:
[Treeherder: 8946aa5c0ef7d3b786ad3edfbc069d311119b85a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8946aa5c0ef7d3b786ad3edfbc069d311119b85a)

PR #3898, opened 2017-12-07
Fix Bug 1423506 - Tippytop does not honor preference browser.chrome.site_icons

Mardak:
r?@rlr Check prefs before trying to fetch. Also removes `ActivitySteramPrefs` usage to `getState` instead. Also fixed up not quite right test that was still checking state.TippyTop.


as-pine-proxy:
[Treeherder: 6cb06fa548ba7eaf44a823fedd828c7974debac3](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6cb06fa548ba7eaf44a823fedd828c7974debac3)

PR #3901, opened 2017-12-08
Fix Bug 1424292 - Pocket referrer only sent when story has HTTPS URL

csadilek:
Basically, lots of German publishers don't use HTTPS. The current referrer isn't sent along for these publishers, so switching the referrer to plain HTTP. This will make sure it's sent for both HTTP and HTTPS stories.

Details: https://bugzilla.mozilla.org/show_bug.cgi?id=1424292


as-pine-proxy:
[Treeherder: c12f4a74eb9d3937baabf24e626c8c51d10c8612](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c12f4a74eb9d3937baabf24e626c8c51d10c8612)

PR #3902, opened 2017-12-13
feat (topsites): Fix Bug 1402192 - Top Sites Drag n Drop

rlr:
I need to add some more content tests, but this is ready for some 👀 

r? @Mardak @k88hudson


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -259,6 +266,18 @@ this.TopSitesFeed = class TopSitesFeed {
...
+  /**
+   * Handle a Top Site drop into new slot.
+   */
+  drop(action) {
+    const {site, index} = action.data;
+
+    // Dropping a top site pins it in the slot it was dropped, pushing over any link already
+    // pinned in the slot (unless it's the last slot, then it replaces).
+    this._insertPin(site, index);
+    this._broadcastPinnedSitesUpdated();
```
Mardak: I was thinking this looked familiar… ;) Perhaps we should have whatever would have called `add` triggering `_insertPin(site, 0)` to dispatch "DROP" although maybe the action should be more generically "TOP_SITES_INSERT" or just reuse "TOP_SITES_ADD" but have it take an `index` where content will either dispatch with a hardcoded 0 for the existing add behavior (or have a default index of 0?) and the dynamic index for the drop behavior.

Mardak:
I put up a video attachment for uiwanted https://bugzilla.mozilla.org/attachment.cgi?id=8937572

rlr:
> I put up a video attachment for uiwanted

I *think* I can easily fix the drop in between tiles by using left padding instead of margins. I played with having a `setTimeout` to expire the preview to deal with the drop issue but I removed it because it wasnt always noticeable on my machine. I'll play with that more to see.

The hover getting stuck I have no idea how to fix. It's an issue with dragging no "releasing" the hover state.

Mardak:
For dragging over whitespace, I believe the underlying issue is that we have the listeners on the individual sites instead of the parent? It's a bit of a different approach, so I don't know how complex it will get, but if we maybe got something related to the coordinates of the current drag position and do some math?

With either approach of padding or calculations, we'll want to make sure the whitespace is treated equally on each side of the site. (Probably?)

rlr:
^ Fixed the first two issues and the hover issue magically disappeared with those changes. lol

system-addon/lib/TopSitesFeed.jsm
rlr: Ah, good point. No need to create new event type. Will fix

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -175,24 +253,139 @@ export class TopSite extends React.PureComponent {
...
+    this.setState({topSitesPreview: null});
+  }
+  onDrop(index) {
+    this.props.dispatch(ac.SendToMain({
+      type: at.TOP_SITES_DROP,
+      data: {site: {url: this.state.draggedSite.url, label: this.state.draggedTitle}, index}
+    }));
+    // Add a delay to the state reset to avoid ui jank going back to original state
+    // until the pinning happens and the state gets updated and pushed back down.
+    setTimeout(() => this.setState(this.DEFAULT_STATE), 300);
```
Mardak: Typically magic timeouts lead to random intermittent failures in mochitests, but we can at least mock the timer in our tests. For slow machines, the user could see the state reverting then updating when the pin propagates.

Do we need this to reset to default vs optimistically assuming the pin will succeed?

Also, what exactly happens when this gets called and redux state has updated? Does the `setState` end up doing nothing? Setting the state of the site component that happens to reuse it?

@k88hudson any suggestions?

rlr: Oh, hmm. I guess I can listen for prop updates and clear out the preview once at the point when I get an update with the site properly pinned.

rlr: I don't completely understand your questions. But basically the preview reorders the sites from the redux store and keeps them in the local state while needed. The sites in the local state takes precedence over the redux store.

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -175,24 +253,148 @@ export class TopSite extends React.PureComponent {
...
+    this.onDragLeave = this.onDragLeave.bind(this);
+    this.onDrop = this.onDrop.bind(this);
+  }
+  componentWillUpdate(nextProps) {
+    if (this.state.draggedSite) {
+      const prevTopSites = this.props.TopSites && this.props.TopSites.rows;
+      const newTopSites = nextProps.TopSites && nextProps.TopSites.rows;
+      if (prevTopSites && prevTopSites[this.state.draggedIndex] &&
+        prevTopSites[this.state.draggedIndex].url === this.state.draggedSite.url &&
+        (!newTopSites[this.state.draggedIndex] || newTopSites[this.state.draggedIndex].url !== this.state.draggedSite.url)) {
```
rlr: This check kind of got out of hand. Can it be simpler maybe?

system-addon/content-src/components/TopSites/TopSite.jsx
Mardak: The thing I was concerned about is related to the lifecycle of the component. Could `this.setState` be called after the props / other state have been changed and something unexpected might happen if we just set it to DEAFULT after 300ms?

system-addon/content-src/components/TopSites/TopSite.jsx
Mardak: I don't know if it breaks redux flat object something something. But if the .rows aren't the same Array object, maybe, it's different enough? Otherwise, maybe something like `rows.map(r => r.url).toString()` checking might be good enough to see if the urls are different?

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -8,59 +8,125 @@ import {
...
-    imageStyle = {backgroundImage: link.screenshot ? `url(${link.screenshot})` : "none"};
+export class TopSiteLink extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.onDragStart = this.onDragStart.bind(this);
+    this.onDragEnd = this.onDragEnd.bind(this);
+    this.onDragOver = this.onDragOver.bind(this);
+    this.onDragEnter = this.onDragEnter.bind(this);
+    this.onDragLeave = this.onDragLeave.bind(this);
+    this.onDrop = this.onDrop.bind(this);
```
Mardak: We have this pattern in many places now, but I wonder if having a central handler to dispatch would be any nicer. E.g.,
```js
constructor() {
  this.onEvent = this.onEvent.bind(this);
}
onEvent(event) {
  switch (event.type) {
    case "drop": …
}
```

Eh.. probably too much :p Maybe I'm thinking more of the `handleEvent` method on the `EventListener` interface, so one could just do `addEventListener("drop", this);` but I don't think React handles that? https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -8,59 +8,125 @@ import {
...
+  }
+  onDragLeave(e) {
+    if (this._allowDrop(e, this.props.index)) {
+      this.props.onDragLeave(this.props.index);
+    }
+  }
+  onDrop(e) {
+    if (this._allowDrop(e, this.props.index)) {
+      e.preventDefault();
+      this.props.onDrop(this.props.index);
```
Mardak: Although with a common `onEvent` could do…
```js
onEvent(event) {
  switch (event.type) {
    case "dragover":
    case "dragenter":
    case "dragleave":
    case "drop":
      if (this._allowDrop(event, index)) {
        event.preventDefault();
        this.props.onEvent(event, index);
```

? maybe…

system-addon/content-src/components/TopSites/TopSite.jsx
rlr: yeah, I’m definitely “tired” of the .bind() boilerplate

system-addon/content-src/components/TopSites/TopSite.jsx
rlr: I think that's better? 422ea358690cab7217b3fc586ccabb5a85032ab6

rlr: (except I broke something)

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -8,59 +8,113 @@ import {
...
+      case "dragenter":
+      case "dragleave":
+      case "drop":
+        if (this._allowDrop(event, this.props.index)) {
+          event.preventDefault();
+          this.props.onDragEvent(event, this.props.index);
+        }
+        break;
+      default:
+        break;
```
Mardak: nit: no need `default:`

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -8,59 +8,113 @@ import {
...
+    }
+  }
+  render() {
+    const {props} = this;
+    const {link, title, className, isDraggable} = props;
```
Mardak: nit: include `{children, onClick, …} = this.props;` without pulling `props` out the line before

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -175,24 +239,159 @@ export class TopSite extends React.PureComponent {
...
+        break;
+      case "dragover":
+      case "dragenter":
+        this.onDragEnter(index);
+        break;
+      case "dragleave":
+        this.onDragLeave();
+        break;
+      case "drop":
+        this.onDrop(index);
```
Mardak: nit: these all seem to be inline-able? Unless we want to expose them for slightly easier testing instead of passing a dummy event into this method?

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -175,24 +239,159 @@ export class TopSite extends React.PureComponent {
...
+   * list, if any. Otherwise leave an empty slot.
+   */
+  _fillOrLeaveHole(sites, index) {
+    sites[index] = null;
+    for (let i = index + 1, l = sites.length; i < l; i++) {
+      const site = sites[i];
+      if (site && !site.isPinned) {
+        this._fillOrLeaveHole(sites, i);
+        sites[index] = site;
+        break;
```
Mardak: Does this need to be recursive? The inner call is just doing the same thing starting from the first non-pinned site, so wouldn't just letting the loop continue just work? ??
```js
for (let i = index + 1; i < sites.length; i++) {
  if (site…) {
    sites[index] = site;
    // Update the index to fill to be the spot we just grabbed a site from
    index = i;
  }
}
```

??

Mardak: Oh, and probably set the `sites[i] = null;`

rlr: ah. yep. weird that I thought of recursion first 😮

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -175,24 +236,161 @@ export class TopSite extends React.PureComponent {
...
+    this._insertSite(preview, Object.assign({}, this.state.draggedSite, {isPinned: true}), index);
+    return preview;
+  }
+
+  /**
+   * Fill in the slot at the specified index with a non pinned site further down the
+   * list, if any. Otherwise leave an empty slot.
+   */
+  _fillOrLeaveHole(sites, index) {
+    let slotIndex = index;
```
Mardak: Could just define the method to have `slotIndex` as its parameter? ;)

rlr: this was because linter was complaining about reassigning to index. I think it looks funny this way though. already using index and i in the function.

rlr:
The other day I had an idea for a new algorithm to try out but I didn't write it down. Now I can't remember it but it will come back some day in the shower. 😄

rlr:
okie dokie. what could possibly go wrong? here goes this *gulp*

as-pine-proxy:
[Treeherder: 27b66f5acba4a992d25590897d85d99f02f2b2e8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=27b66f5acba4a992d25590897d85d99f02f2b2e8)

PR #3904, opened 2017-12-14
Use headless mode for unit tests

piatra:
You've seen it at the lightning talks, now in our repo!
This makes it so that `testmc:unit` runs in headless mode but `tddmc` still starts up regular firefox.
I like it because it doesn't steal the focus to start firefox, I don't know of any negative. 
Lmk if you think it's a good idea and I'll go ahead and make a bug for it too.

### DIFF
--- a/karma.mc.config.js
+++ b/karma.mc.config.js
@@ -26,9 +26,16 @@ preprocessors[PATHS.testFilesPattern] = [
 
 module.exports = function(config) {
   const isTDD = config.tdd;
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: c69d1c691316a6e6761560b46d62ff08fc20f007](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c69d1c691316a6e6761560b46d62ff08fc20f007)

PR #3905, opened 2017-12-18
Fix Bug 1425489 - Default to two rows of Top Sites

rlr:
tiny r? @Mardak 

My only concern is the first run experience will now have a row of placeholders. Is that OK @aaronrbenson @bryanbell ? Or do we need to add 6 more default sites?

screen shot 2017-12-18 at 4 21 49 pm

### DIFF
--- a/system-addon/common/PrerenderData.jsm
+++ b/system-addon/common/PrerenderData.jsm
@@ -51,7 +51,7 @@ this.PrerenderData = new _PrerenderData({
     "migrationExpired": true,
     "showTopSites": true,
     "showSearch": true,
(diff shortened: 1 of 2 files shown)


Mardak:
design says to not do this until we get https://bugzilla.mozilla.org/show_bug.cgi?id=1426464

as-pine-proxy:
[Treeherder: 4b0f81a5eac37a62c0c847825348ef6e2799f940](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4b0f81a5eac37a62c0c847825348ef6e2799f940)

PR #3907, opened 2017-12-19
Port Bug 1425868 - Activity Stream changes to info panel

nt1m:


system-addon/content-src/styles/activity-stream-linux.scss
```diff
@@ -2,7 +2,7 @@
 
 $os-infopanel-arrow-height: 10px;
 $os-infopanel-arrow-offset-end: 6px;
-$os-infopanel-arrow-url: 'chrome://global/skin/arrow/panelarrow-vertical-themed.svg';
+$os-infopanel-arrow-url: 'chrome://global/skin/arrow/panelarrow-vertical.svg';
```
piatra: It seems to be the same file everywhere so we can just specify it once here
https://github.com/mozilla/activity-stream/blob/8e5857c1f8b102c410892708dba9bb6f6bc3fcf6/system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss#L77
And drop the variable.

Mardak:
I would not have the "Fix Bug 1425868" in the title as it will cause the bug to be resolved fixed.

as-pine-proxy:
[Treeherder: 0495b7c88cd73bde180658651b57c7ea3dae9dd9](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0495b7c88cd73bde180658651b57c7ea3dae9dd9)

PR #3908, opened 2017-12-19
Prepare for Bug 1369722 - Use non-browser environment for jsms

Standard8:
In [bug 1369722](https://bugzilla.mozilla.org/show_bug.cgi?id=1369722) I'm working on updating the environment for .jsm files so that it is more accurate (currently we assume 'browser' which is too much).

This patch adds a simple ignore for something that is already checked.

Whilst here, I'm also upgrading e-p-m to 0.4.10 and as it now specifies mozilla/use-services by default, removing that from the activity stream definition.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,7 +28,6 @@ module.exports = {
     "plugin:mozilla/recommended" // require("eslint-plugin-mozilla")
   ],
   "rules": {
-    "mozilla/use-services": 2,
     "promise/param-names": 2,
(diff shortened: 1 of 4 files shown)


as-pine-proxy:
[Treeherder: 5bc0911bd1034b057bdfd40ec9438f13dd98007d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5bc0911bd1034b057bdfd40ec9438f13dd98007d)

PR #3914, opened 2017-12-22
Fix Bug 1402864 - Default Top Sites do not have the "Edit" option

Mardak:
Was bothering me that you could drag a default top site but then not edit it… Also that a placeholder had more edit-ability than a default site. But you could still do so via edit mode. r?@rlr


as-pine-proxy:
[Treeherder: cd143b430999af0bad83518129629bb1fa4cd447](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=cd143b430999af0bad83518129629bb1fa4cd447)

piatra:
I was doing this in the custom image picker PR.

Mardak:
@piatra was there something in the image picker PR that was needed before this should have been fixed?

PR #3917, opened 2018-01-04
Fix Bug 3431: Clean up usage of ActivityStreamPrefs to use Prefs from the store

### LINKED ISSUE #3431: Clean up usage of ActivityStreamPrefs to use Prefs from the store
Some places are creating a new `ActivityStreamPrefs` just to `.get()` an activity stream pref. That data is available from `this.store.getState().Prefs.values`

E.g.,
https://github.com/mozilla/activity-stream/blob/fc64a8ba1bb60f19646189ac5279840f763d2878/system-addon/lib/ManualMigration.jsm#L21
https://github.com/mozilla/activity-stream/blob/fe0dff9fee8dc41d970c436a49dd4cd30391ee55/system-addon/lib/TopStoriesFeed.jsm#L161

This will require some test changes too..

SeanPrashad:
Fixes #3431: Update redundant `.get()` calls to ActivityStreamPrefs to use `this.store.getState().Prefs.values`. Also updating the respective unit tests to accomodate this new behaviour.

**Note**: Currently a WIP as of Jan 4, 2018

### DIFF
--- a/system-addon/lib/ManualMigration.jsm
+++ b/system-addon/lib/ManualMigration.jsm
@@ -5,7 +5,6 @@
 
 const {utils: Cu} = Components;
 const {actionCreators: ac, actionTypes: at} = Cu.import("resource://activity-stream/common/Actions.jsm", {});
-const {Prefs} = Cu.import("resource://activity-stream/lib/ActivityStreamPrefs.jsm", {});
 const MIGRATION_ENDED_EVENT = "Migration:Ended";
 const MS_PER_DAY = 86400000;
 
@@ -18,7 +17,24 @@ XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge", "resource://gre/modules/Pr
 this.ManualMigration = class ManualMigration {
   constructor() {
     Services.obs.addObserver(this, MIGRATION_ENDED_EVENT);
-    this._prefs = new Prefs();
+  }
+
+  get getMigrationLastShownDate() {
+    return this.store.getState().Prefs.values.migrationLastShownDate;
+  }
+
+  set setMigrationLastShownDate(newDate) {
+    this.store.dispatch(ac.SetPref("migrationLastShownDate", newDate));
+    // this.store.getState().Prefs.values.migrationLastShownDate = newDate;
+  }
+
+  get getMigrationRemainingDays() {
+    return this.store.getState().Prefs.values.migrationRemainingDays;
+  }
+
+  set setMigrationRemainingDays(newDate) {
+    this.store.dispatch(ac.SetPref("migrationRemainingDays", newDate));
+    // this.store.getState().Prefs.values.migrationRemainingDays = newDate;
   }
 
   uninit() {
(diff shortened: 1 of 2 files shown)


system-addon/test/unit/lib/ManualMigration.test.js
```diff
@@ -38,17 +42,23 @@ describe("ManualMigration", () => {
       }
     };
 
-    const {ManualMigration} = injector({"lib/ActivityStreamPrefs.jsm": {Prefs: fakePrefs}});
+    // const {ManualMigration} = injector({"lib/ActivityStreamPrefs.jsm": {Prefs: fakePrefs}});
+    const {ManualMigration} = injector({"lib/ActivityStreamPrefs.jsm": {}});
```
Mardak: The `injector` was only needed to pass in a fake prefs, so because we're no longer using that, we should be able to just `import` ManualMigration as a regular module at the top of the test file.

system-addon/lib/ManualMigration.jsm
```diff
@@ -18,7 +17,22 @@ XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge", "resource://gre/modules/Pr
...
+  set setMigrationLastShownDate(newDate) {
+    this.store.getState().Prefs.values.migrationLastShownDate = newDate;
+  }
+
+  get getMigrationRemainingDays() {
+    return this.store.getState().Prefs.values.migrationRemainingDays;
+  }
+
+  set setMigrationRemainingDays(newDate) {
+    this.store.getState().Prefs.values.migrationRemainingDays = newDate;
```
Mardak: We don't want to directly modify the store without notifying others. See `expireMigration` in this file.

system-addon/lib/ManualMigration.jsm
```diff
@@ -18,7 +17,22 @@ XPCOMUtils.defineLazyModuleGetter(this, "ProfileAge", "resource://gre/modules/Pr
...
+  get getMigrationLastShownDate() {
+    return this.store.getState().Prefs.values.migrationLastShownDate;
+  }
+
+  set setMigrationLastShownDate(newDate) {
+    this.store.getState().Prefs.values.migrationLastShownDate = newDate;
+  }
+
+  get getMigrationRemainingDays() {
+    return this.store.getState().Prefs.values.migrationRemainingDays;
```
Mardak: Thanks for cleaning this up. I think instead of having individual getter/setters for each pref, it would be simple enough to have a helper that takes in the pref name. Although at that point, it's nothing specific to manual migration, so maybe it should be available for all Feeds?

@k88hudson any thoughts on having `this.store.getPref(name)` and similar `setPref` helper that dispatches?

SeanPrashad:
@Mardak I've updated the ManualMigrations `import` statement as requested and modified both setters to update the store as appropriate.

I'm not attempting to switch over the old `fakePrefs` statements in `ManualMigration.test.js` but I'm unsure what the old `fakePrefs.get.returns(today);` call would look like in the updated version - any suggestions?

k88hudson:
@SeanPrashad This is looking great so far! For the tests, instead of using the `fakePrefs` helper,  you'll want to set whatever value you need on your `instance.store.state` object. For example, instead of `fakePrefs.get.returns(today)`, you would set `instance.store.state.migrationLastShownDate = today`. Does that make sense?

system-addon/lib/ManualMigration.jsm
k88hudson: I can see that being helpful, yeah; I guess it is a bit verbose to do `this.store.getState().Prefs.values` every time.

I think extending `store` to have `store.getPref` and `store.setPrefs` as helpers would be ok, but if it's out of the scope of this PR a single getter/setter in `ManualMigration.jsm` would be fine by me too.

SeanPrashad:
Thank you @k88hudson! That makes much more sense - I will try to take a crack at it sometime soon!

Mardak:
Thanks for putting together this initial PR! Superseded by #4071.

k88hudson:
Thanks for your work on this! Your patch landed in this commit https://github.com/mozilla/activity-stream/commit/b634b4b0a438c0ad1f51b9d20cc0fc25b6d8c26d with some additional test fixes here: (https://github.com/mozilla/activity-stream/commit/2c8e2422be4a1db5802bc20207d34a22a4aeb42f)

PR #3920, opened 2018-01-05
Fix Bug 1423703 - Use negotiated locales, rather than requested in Activity Stream

Mardak:
Switch from `getRequestedLocale` to `getAppLocaleAsLangTag` (and `getAppLocalesAsLangTags` for negotiate in aboutNewTabService). Skips the mochitest for now as per bug as fixing it would depend on some other changes for 60. r?@k88hudson

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -339,7 +339,7 @@ function main() { // eslint-disable-line max-statements
   }
 
   // Provide some help to copy/paste locales if tests are failing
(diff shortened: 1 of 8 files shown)


as-pine-proxy:
[Treeherder: 329f9da55bd050b8da2dd6a8f400a1d984b2efe2](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=329f9da55bd050b8da2dd6a8f400a1d984b2efe2)

PR #3922, opened 2018-01-08
Fix Bug 1423256 - Add title and link elements back.

piatra:
https://bugzilla.mozilla.org/show_bug.cgi?id=1423256

This PR adds title and link elements back.
* This fixes cases like session restore where title is completely missing. Related to the icon it looks like https://bugzilla.mozilla.org/show_bug.cgi?id=1422074 was not needed.
* Adding it back fixes the case for "Duplicate Tab" when the icon would not show up.
* It also fixes the case where navigating New Tab -> Website -> (Back Button) would display the favicon of `website`

There was also an event listener ([see PR](https://github.com/mozilla/activity-stream/commit/20b392dd7e35d2a369ced71971f7fdbde782a5ef#diff-92b55e4a266df72d234dd8551ab6ca10)) that was removed when the link element was removed. I did not add it back in this PR because I'm not sure of the use case it covered (I'm not seeing any case where it would be needed). It could be that it is handled by https://reviewboard.mozilla.org/r/201000/diff/2#index_header

There is still a favicon flicker on about:home/new window, this is tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1403648 because the mechanism to find out if we are on about:home is async

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -187,7 +187,7 @@ function templateHTML(options, html) {
   <head>
     <meta charset="utf-8">
     <meta http-equiv="Content-Security-Policy-Report-Only" content="script-src 'unsafe-inline'; img-src http: https: data: blob:; style-src 'unsafe-inline'; child-src 'none'; object-src 'none'; report-uri https://tiles.services.mozilla.com/v4/links/activity-stream/csp">
-    <title>${options.strings.title}</title>
+    <title>${options.strings.newtab_page_title}</title>
     <link rel="icon" type="image/png" href="chrome://branding/content/icon32.png"/>
     <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
     <link rel="stylesheet" href="${options.baseUrl}css/activity-stream.css" />
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -187,6 +187,8 @@ function templateHTML(options, html) {
   <head>
     <meta charset="utf-8">
     <meta http-equiv="Content-Security-Policy-Report-Only" content="script-src 'unsafe-inline'; img-src http: https: data: blob:; style-src 'unsafe-inline'; child-src 'none'; object-src 'none'; report-uri https://tiles.services.mozilla.com/v4/links/activity-stream/csp">
+    <title>${options.title}</title>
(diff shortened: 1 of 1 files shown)


bin/render-activity-stream-html.js
```diff
@@ -187,6 +187,8 @@ function templateHTML(options, html) {
   <head>
     <meta charset="utf-8">
     <meta http-equiv="Content-Security-Policy-Report-Only" content="script-src 'unsafe-inline'; img-src http: https: data: blob:; style-src 'unsafe-inline'; child-src 'none'; object-src 'none'; report-uri https://tiles.services.mozilla.com/v4/links/activity-stream/csp">
+    <title>${options.title}</title>
+    <link rel="icon" type="image/png" id="favicon" href="chrome://branding/content/icon32.png"/>
```
Mardak: Is the `id` necessary?

bin/render-activity-stream-html.js
```diff
@@ -304,6 +306,7 @@ function main() { // eslint-disable-line max-statements
     const options = Object.assign({}, baseOptions, {
       direction: getTextDirection(locale),
       locale,
+      title: strings.newtab_page_title,
```
Mardak: Looks like `strings` from the next line can just be reused instead of additionally passing in the title?

piatra:
Locally I tested with the sessionstore PR reverted. I wanted to make that as a separate follow up bug.

Mardak:
How did you test locally? I'm seeing this with new window then 2 new tabs:

screen shot 2018-01-08 at 4 36 12 pm

piatra:
I used the obj key when I applied the feedback comments. It should be fixed now.
I tested by [running on the reference hardware](https://drive.google.com/open?id=17idxUse7P-zeUE4TnAWKWLxgaaQubw7n) to see if I catch any flashes. It fixes the duplicate tab scenario (now there is an icon) although the icon is not available on first paint. 

Follow up for session restore: https://bugzilla.mozilla.org/show_bug.cgi?id=1429119

as-pine-proxy:
[Treeherder: c1fc6ebdf686e57d6cc6fa8125837b14edffe7f5](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c1fc6ebdf686e57d6cc6fa8125837b14edffe7f5)

PR #3923, opened 2018-01-08
Fix Bug 1426787 - Top sites that match defaults lose 'Remove from History' option when they are pinned

sarracini:
Only add ```isDefault``` attribute to pinned sites that are not frecent, meaning we don't care if it is a default if it's a frecent. Prioritize frecent sites. This puts the 'remove from history' item back in the context menu of sites that are pinned + frecent + coincidentally also in the list of default sites


as-pine-proxy:
[Treeherder: 1afec9b6ea5a3aba880365786a757b328f4cde91](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1afec9b6ea5a3aba880365786a757b328f4cde91)

PR #3924, opened 2018-01-09
Fix Bug 1428108 - Enable activity-stream.debug by default in local builds

piatra:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1428108

@dmose Is this what you had in mind? Local builds will now load react-dev and react-dom-dev.

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -134,6 +134,11 @@ const PREFS_CONFIG = new Map([
   ["tippyTop.service.endpoint", {
     title: "Tippy Top service manifest url",
     value: "https://activity-stream-icons.services.mozilla.com/v1/icons.json.br"
+  }],
+  ["debug", {
+    title: "Enable debug mode",
+    value: false,
+    value_local_dev: true
   }]
 ]);
 


Mardak:
How does that interact with https://searchfox.org/mozilla-central/source/browser/app/profile/firefox.js#1287 ?

piatra:
I think when AS is loaded it will overwrite the default prefs set there. Could we remove that line altogether from firefox.js?

dmose:
I believe removing that line would make local developer builds on release or beta stop working, because they would try to use the debug version, and that [doesn't get packaged on release or beta](https://github.com/mozilla/activity-stream/blob/e219cc7f2da1fa6558a1c57e89f18b4ed0bd09d7/system-addon/jar.mn#L12).

So we're going to have figure out what to do about that.  It's possible to get the "release or beta" info from AppConstants.jsm at runtime; I suspect we'll need to use that.  I'm not sure whether that code would want to live in this repo, or in mozilla-central, [perhaps here](https://searchfox.org/mozilla-central/source/browser/components/newtab/aboutNewTabService.js#119).

piatra:
[We can define the same IS_UNOFFICIAL_BUILD](https://github.com/mozilla/activity-stream/blob/406b094ead23fb132578e78820a08d4231282186/system-addon/lib/ActivityStreamPrefs.jsm#L80) in aboutNewTabServices (where we define the path to the debug or regular html file) because it includes `AppConstants`.
We can make it so that for UNOFFICIAL builds we initialize debug to be `true` by default and you would be able to switch back because there is an even listener registered for updates.
Is that preferable?

A different solution was suggested to me, switching to actual firefox `debug` builds. We would always build for debug locally and our `firefox.js` might look like this
```
#ifndef RELEASE_OR_BETA
pref("browser.newtabpage.activity-stream.debug", false);
#endif
#ifdef DEBUG
pref("browser.newtabpage.activity-stream.debug", true);
#endif
```

Mardak:
The debug flag only works for not-`RELEASE_OR_BETA`, so you should be able to additionally set the pref `true` or `false` based on `MOZILLA_OFFICIAL`, e.g., https://searchfox.org/mozilla-central/source/modules/libpref/init/all.js#1049-1061

piatra:
Discussing with @dmose we decided that it would be best if the decision is based on `AppConstants.MOZILLA_OFFICIAL` so that it matches how we [set prefs in AS](https://github.com/mozilla/activity-stream/blob/a48d0d3901666ab387d3c6fed03b2b9932e7e1e4/system-addon/lib/ActivityStreamPrefs.jsm#L85).
It would be done in [aboutNewTabService](https://searchfox.org/mozilla-central/rev/03877052c151a8f062eea177f684a2743cd7b1d5/browser/components/newtab/aboutNewTabService.js#212-225) that's where the production/debug path is served from and it already includes `AppConstants`.

piatra:
Looks like it might have to be done in `firefox.js`. I tried setting it in `AboutNewTabService` which throws an error (it might happen because prefs are not yet initialized for writing). Waiting for prefs to be ready wouldn't help, first about:home will not get the debug version.
@Mardak do you mean something like
```
#ifndef RELEASE_OR_BETA
#ifndef MOZILLA_OFFICIAL
pref("browser.newtabpage.activity-stream.debug", true);
#else
pref("browser.newtabpage.activity-stream.debug", false);
#endif
#endif
```

piatra:
Fixed by https://bugzilla.mozilla.org/show_bug.cgi?id=1428108.

PR #3925, opened 2018-01-09
Fix Bug 1428356 - dismissing/deleting Top Sites when some are pinned

rlr:
I removed the optimistic optimization that doesn't work right when there are pinned sites.

r? @Mardak

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -134,13 +134,6 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
         return site;
       });
       return Object.assign({}, prevState, {rows: newRows});
-    case at.BLOCK_URL:
-    case at.DELETE_HISTORY_URL:
-      // Optimistically update the UI by responding to the context menu action
-      // events and removing the site that was blocked/deleted with an empty slot.
-      // Once refresh() finishes, we update the UI again with a new site
-      newRows = prevState.rows.filter(val => val && val.url !== action.data.url);
-      return Object.assign({}, prevState, {rows: newRows});
     default:
       return prevState;
   }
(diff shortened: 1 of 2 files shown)


sarracini:
I think this is fine too, it might be a tad slower since we have to wait for places to finish before blocking but we can always add other optimizations later 👍

rlr:
thanks! Oh travis is unhappy about something /me checks

rlr:
```
FAILED TESTS:
    SnippetsMap
      #set
        ✖ should set an item in indexedDB when connected
          Firefox 54.0.0 (Linux 0.0.0)
        Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      #delete
        ✖ should delete an item from indexedDB when connected
          Firefox 54.0.0 (Linux 0.0.0)
        expected false to be true
        AssertionError@node_modules/chai/chai.js:9320:13
        [3]</module.exports/Assertion.prototype.assert@node_modules/chai/chai.js:239:13
        [5]</module.exports/<@node_modules/chai/chai.js:1020:5
        propertyGetter@node_modules/chai/chai.js:7784:22
        proxyGetter@node_modules/chai/chai.js:9128:14
        [6]</module.exports/assert.isTrue@node_modules/chai/chai.js:4427:5
        @webpack:///system-addon/test/unit/content-src/lib/snippets.test.js:59:0 <- system-addon/test/unit/unit-entry.js:95570:7
```

Two indexedDB tests failing. Should be unrelated to these changes. Any ideas? @k88hudson ?

rlr:
(The tests pass locally for me.)

rlr:
After restarting travis twice, it's fixed. 🕺

as-pine-proxy:
[Treeherder: d0e21e98f10251f47963c8b55ed477e29f7b91a8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=d0e21e98f10251f47963c8b55ed477e29f7b91a8)

PR #3926, opened 2018-01-10
Fix Bug 1429489 - Package user event and session pings for Events Telemetry.

emtwo:


emtwo:
Note: In order to test this patch, Events.yaml from https://bugzilla.mozilla.org/show_bug.cgi?id=1429497 needs to be applied first and then a *non-artifact* build of toolkit/ should be done.

Then a good way to check if pings are making it into the events telemetry system is as follows:

```
      const OPTIN = Ci.nsITelemetry.DATASET_RELEASE_CHANNEL_OPTIN;
      // <Record an events telemetry event>
      let snapshot = Services.telemetry.snapshotEvents(OPTIN, false);
      Services.console.logStringMessage(`snapshot: ${JSON.stringify(snapshot.parent)}`);
```

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -296,11 +296,27 @@ this.TelemetryFeed = class TelemetryFeed {
...
+
+    let mainFields = ["event", "source", "action_position"];
+    let eventFields = mainFields.map(field => String(pingCentrePing[field]) || null);
+    let eventsTelemetryPing = [
+      "activity_stream",
+      ...eventFields,
+      utExtras
+    ];
+
+    return {pingCentrePing, eventsTelemetryPing};
```
emtwo: This will attempt to turn *any* user event ping into an events telemetry ping but at the moment it should only send `CLICK` pings. This should instead only create `pingCentrePing` and exit early if it's not a `CLICK` ping.

emtwo: The schema definition (https://bugzilla.mozilla.org/show_bug.cgi?id=1429497) has been updated to include **all** event types not just `CLICK` so this is no longer an issue.

ncloudioj:
I've noticed two major concerns so far:

* The payload conversion between PC and UT in `TelemetryFeed.jsm`, it's scattering among the whole module, and will become harder to add new metrics as well as maintain the old ones
* Those UT specific operations in `PingCentre.jsm`, which appears to be wasteful for those PingCentre users who don't need to deal with UT events

I'd recommend to implement this in a different way. Specifically, add a separate module in Activity-Stream, say `UTEventReporting.jsm`, and wrap all those details into it. Let TelemetryFeed simply talk to this module without changing anything for the PingCentre client.

What do you think? @emtwo

emtwo:
> The payload conversion between PC and UT in TelemetryFeed.jsm, it's scattering among the whole module, and will become harder to add new metrics as well as maintain the old ones

I agree we can pull out that code to separate it.

> Those UT specific operations in PingCentre.jsm, which appears to be wasteful for those PingCentre users who don't need to deal with UT events

The original motivation here was that Ping Centre client should maintain the same simple interface (i.e. `sendPing()`) but send the payload out to both pipelines. After discussion on irc, we've decided this would have only been beneficial if the payloads were identical. Since the events telemtry interface is fairly simple, it can be called directly outside of PingCentre client to simplify the code.

system-addon/lib/TelemetryFeed.jsm
```diff
@@ -118,6 +120,14 @@ this.TelemetryFeed = class TelemetryFeed {
     return this.pingCentre;
   }
 
+  /**
+   * Lazily initialize UTEventReporting to send pings
+   */
+  get utEvents() {
+    Object.defineProperty(this, "utEvents", {value: new UTEventReporting()});
```
ncloudioj: Does this need a guard here to check if `utEvents` already exists?

system-addon/lib/UTEventReporting.jsm
```diff
@@ -0,0 +1,56 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+/* globals Services */
+
+"use strict";
+
+const {interfaces: Ci, utils: Cu} = Components;
+Cu.import("resource://gre/modules/Services.jsm");
+
+const EXTRAS_FIELD_NAMES = ["addon_version", "session_id", "page", "user_prefs", "action_position"];
```
ncloudioj: nit: let's add a link here to reference the AS event schema in MC.

system-addon/lib/TelemetryFeed.jsm
emtwo: I think by default when you use `Object.defineProperty()` and set its `value` it can no longer have a getter or setter. So next time `this.utEvents` is called, it will access the property value directly and will not try to create a `new UTEventReporting()`. We do the same thing for `PingCentre` here:

https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/TelemetryFeed.jsm#L110

Dexterp37:
I had a look at the usage of the Telemetry API: looks good to me!

Mardak:
This is failing automation with recent eslint changes:
https://travis-ci.org/mozilla/activity-stream/builds/338135814

```
/home/travis/build/mozilla/activity-stream/system-addon/lib/TelemetryFeed.jsm (1/0)
  ✖  19:1  Please use ChromeUtils.defineModuleGetter instead of XPCOMUtils.defineLazyModuleGetter  mozilla/use-chromeutils-import
/home/travis/build/mozilla/activity-stream/system-addon/lib/UTEventReporting.jsm (1/0)
  ✖  9:1  Please use ChromeUtils.import instead of Cu.import  mozilla/use-chromeutils-import
/home/travis/build/mozilla/activity-stream/system-addon/test/schemas/pings.js (1/0)
  ✖  54:21  A linebreak is required before ']'  array-bracket-newline
✖ 3 errors (6:25:38 PM)
```
I'll fix them up.

as-pine-proxy:
[Treeherder: 8f0e724c81149ea0a0065b0453acb866f7ce085f](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=8f0e724c81149ea0a0065b0453acb866f7ce085f)

PR #3930, opened 2018-01-12
Fix Bug 1428727 - better shifting of pinned sites during drag n drop

rlr:
@Mardak this is what I was thinking. And no recursion! I haven't made the corresponding changes to the feed yet. Thoughts?

I'll show UX tomorrow. But I am pretty sure this is closer to what tiles was doing.

Before:

After:

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -300,7 +300,7 @@ export class _TopSiteList extends React.PureComponent {
         if (index !== this.state.draggedIndex) {
           this.props.dispatch(ac.SendToMain({
             type: at.TOP_SITES_INSERT,
-            data: {site: {url: this.state.draggedSite.url, label: this.state.draggedTitle}, index}
+            data: {site: {url: this.state.draggedSite.url, label: this.state.draggedTitle}, index, draggedFromIndex: this.state.draggedIndex}
           }));
           this.userEvent("DROP", index);
         }
@@ -308,7 +308,10 @@ export class _TopSiteList extends React.PureComponent {
     }
   }
   _getTopSites() {
-    return this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
+    // Make a copy of the sites to truncate or extend to desired length
+    let topSites = this.props.TopSites.rows.slice();
+    topSites.length = this.props.TopSitesCount;
+    return topSites;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


Mardak:
Looking at the After, it looks like the behavior is initially shifting the pinned sites into the hole. But when dragging reddit over wikipedia, it creates the hole again for a frecent site? I would have expected  facebook in reddit's original position when dragging over each top site until netflix:

- over youtube -> swap with reddit
- over facebook -> swap with reddit
- over amazon -> shift facebook + amazon left one to place reddit
- over wikipedia -> shift facebook + amazon + wikipedia left one to place reddit
- over twitter -> shift facebook + amazon + wikipedia + twitter left one to place reddit
- over netflix -> swap with reddit
- over bbc -> netflix moves to reddit's hole, bbc moves to netflix's hole, reddit moves to bbc's hole

rlr:
Yeah, I am shifting towards the closest hole (favoring the right side if there is a tie). Once I get to wikipedia, the closest hole would be on the second row so that's why it moves twitter there. But if there was only one row (or the second row had more pinned sites), it would do what you described:

Mardak:
I think it should shift towards the hole of the site being dragged. But I suppose UX can provide that feedback. ;)

rlr:
Hmmm. That’s definitely another way to do it. And probably would simplify
the code a little.

rlr:
I vidyo'd it out with @bryanbell and he said it should work just like the dragging of iphone apps where "there is pressure to move the apps/sites towards the upper left". From what I can tell, that just means the sites just shift towards the site that is being dragged. I'll try to get that implemented.

rlr:
OK, I think it's good for r? now. I will test it out a bunch locally for sanity.

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -312,67 +312,61 @@ export class _TopSiteList extends React.PureComponent {
...
           this.userEvent("DROP", index);
         }
         break;
     }
   }
   _getTopSites() {
-    return this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
+    let topSites = this.props.TopSites.rows.slice(0, this.props.TopSitesCount);
+    if (topSites.length < this.props.TopSitesCount) {
+      topSites[this.props.TopSitesCount - 1] = null;
```
Mardak: I think it would be clearer / simpler to just do
```js
// Make a copy of the sites to truncate or extend to desired length
let topSites = this.props.TopSites.rows.slice();
topSites.length = this.props.TopSitesCount;
return topSites;
```

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -241,20 +241,43 @@ this.TopSitesFeed = class TopSitesFeed {
   /**
    * Insert a site to pin at a position shifting over any other pinned sites.
    */
-  _insertPin(site, index) {
+  _insertPin(site, index, draggedFromIndex) {
     // Don't insert any pins past the end of the visible top sites. Otherwise,
     // we can end up with a bunch of pinned sites that can never be unpinned again
     // from the UI.
-    if (index >= this.store.getState().Prefs.values.topSitesCount) {
+    const topSitesCount = this.store.getState().Prefs.values.topSitesCount;
```
Mardak: nit: `const {topSitesCount} = …`

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -312,67 +312,61 @@ export class _TopSiteList extends React.PureComponent {
...
+    const siteToInsert = Object.assign({}, this.state.draggedSite, {isPinned: true});
+    if (!pinnedOnly[index]) {
+      pinnedOnly[index] = siteToInsert;
+    } else {
+      // Find the hole to shift the pinned site(s) towards. We shift towards the
+      // hole left by the site being dragged.
+      let holeIndex = index;
+      const indexStep = index > this.state.draggedIndex ? -1 : 1;
+      while (pinnedOnly[holeIndex]) {
+        holeIndex += indexStep;
```
Mardak: There was some shared `common/Reducers.jsm` pinning code before that we can try to clean up in a followup.

as-pine-proxy:
[Treeherder: 4952c254d9d3c62242ad712e23d948e8a3c379a8](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4952c254d9d3c62242ad712e23d948e8a3c379a8)

PR #3938, opened 2018-01-17
Fix Bug 1430130 - fix top site flickering on drop

rlr:
I already had all this code here https://github.com/rlr/activity-streams/blob/47951e8dae81183d7e7e968844b616d69ef0e36e/system-addon/content-src/components/TopSites/TopSite.jsx#L272-L283 to handle this. But I didn't realize dragend was called when dropping. Is this too hacky? Any better ideas?

@Mardak @k88hudson r?

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -290,6 +290,7 @@ export class _TopSiteList extends React.PureComponent {
   onDragEvent(event, index, link, title) {
     switch (event.type) {
       case "dragstart":
+        this.dropped = false;
         this.setState({
           draggedIndex: index,
           draggedSite: link,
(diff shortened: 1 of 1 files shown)


k88hudson:
@rlr This works but results in some weird drag state being left around which could cause bugs / unexpected behaviour and is probably not the best idea

rlr:
@k88hudson what drag state gets left around? It should all get cleared out in the `componentWillUpdate` code when new top sites come in.

rlr:
Closing in favor of #3939 3939

k88hudson:
A

Mardak:
From some quick searching, it looks like `drop` will happen before `dragend`. https://stackoverflow.com/a/43863083

Mardak:
Also, I quickly poked around at setting data on the data transfer from "drop" but that's not allowed.

as-pine-proxy:
[Treeherder: d0c6bf7e4a1c2c786b7d61975a7160324780293b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=d0c6bf7e4a1c2c786b7d61975a7160324780293b)

PR #3939, opened 2018-01-17
Fix Bug 1430130 - Prevent flashing on drag

k88hudson:


rlr:
I'm cool with this. My fix depends on the ordering of the drop and dragend events, which I don't really know if we should rely on that. We might be able to remove the `componentWillUpdate` check? And travis is fail.

k88hudson:
Still needs tests.

@rlr `drop` is always before `dragend`, so that shouldn't be an issue (if it occurs)

PR #3940, opened 2018-01-17
Fix Bug 1430640 - Button labels are cut in 'Delete from history' dialog on localized build

sarracini:
Verified with Aaron that this is the desirable look for long strings


as-pine-proxy:
[Treeherder: 152cafa406201550349dca4a98f9d0030863b8bc](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=152cafa406201550349dca4a98f9d0030863b8bc)

PR #3941, opened 2018-01-17
Bug 1429825 - Dragging and Dropping a Top Site in the same position with the trackpad wrongly opens the site

Mardak:
r?@rlr When dumping out various mouse events and barely dragging a top site, sometimes I see:
```
mousedown
dragstart
dragenter
drop
dragend
mouseup
click
```

And other times I get:
```
mousedown
dragstart
dragenter
drop
dragend
```

So I just made it cancel any `click`s after a `dragstart`.

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -25,7 +25,14 @@ export class TopSiteLink extends React.PureComponent {
 
   onDragEvent(event) {
     switch (event.type) {
+      case "click":
+        // Stop any link clicks if we started any dragging
+        if (this.dragged) {
+          event.preventDefault();
+        }
+        break;
       case "dragstart":
+        this.dragged = true;
         event.dataTransfer.effectAllowed = "move";
         event.dataTransfer.setData("text/topsite-index", this.props.index);
         event.target.blur();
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -80,9 +91,10 @@ export class TopSiteLink extends React.PureComponent {
     let draggableProps = {};
     if (isDraggable) {
       draggableProps = {
-        draggable: true,
```
Mardak: this wasn't being used ?

rlr: interesting, does it still work? https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable

Mardak:
Oh interesting. Looks like with and without `draggable` attribute, the events:

```
DIV mousedown
A dragstart
DIV dragenter
DIV dragover
DIV drop
A dragend
```

So yeah looks like "auto" works fine for us here.

as-pine-proxy:
[Treeherder: 4c4a9012cf0caf00ac13b84dafdd9d8560374664](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4c4a9012cf0caf00ac13b84dafdd9d8560374664)

PR #3942, opened 2018-01-18
fix(TopSites): Keep placeholder edit button from opening two modals, …

dmose:
…Bug 1431078.  Basically just clones the behavior from non-placeholder cards.

r?@Mardak

### DIFF
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -240,9 +240,17 @@ export class TopSitePlaceholder extends React.PureComponent {
     super(props);
     this.onEditButtonClick = this.onEditButtonClick.bind(this);
   }
+
   onEditButtonClick() {
-    this.props.onEdit(this.props.index);
+    if (this.props.onEdit) {
+      this.props.onEdit(this.props.index);
+      return;
+    }
+
+    this.props.dispatch(
+      {type: at.TOP_SITES_EDIT, data: {index: this.props.index}});
   }
+
   render() {
     return (<TopSiteLink className="placeholder" isDraggable={false} {...this.props}>
       <button className="context-menu-button edit-button icon"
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -241,10 +241,7 @@ export class TopSitePlaceholder extends React.PureComponent {
     this.onEditButtonClick = this.onEditButtonClick.bind(this);
   }
   onEditButtonClick() {
-    this.props.dispatch({
-      type: at.TOP_SITES_EDIT,
-      data: {index: this.props.index}
-    });
+    this.props.onEdit(this.props.index);
```
Mardak: Probably just make it conditionally call `this.props.onEdit` or do the existing `dispatch` ?

Mardak: @rlr ^^ ?

rlr: yeah, right now clicking on the edit icon from outside the modal doesn't work. I wonder if there is a way we could just kill the whole edit modal concept. I'll discuss with @bryanbell next week.

dmose: @Mardak @rlr: I'm confused.  If I'm understanding what you guys have said correctly, something should be busted with the patch as written.  In my local tree, clicking the hovering edit icon brings up a functional edit modal both when done in the normal topsites or in the edit mode modal.   Am I missing something?

dmose: The hovering edit icon over the giant image for a given site, that is.

rlr: This is what I see (and in the console I see: `JavaScript error: resource://activity-stream/data/content/activity-stream.bundle.js, line 2887: TypeError: this.props.onEdit is not a function`):

dmose: @rlr thanks!   I was clicking the wrong thing.

as-pine-proxy:
[Treeherder: 712df8f61d71c211cb37e97df328fb88d33cdfbb](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=712df8f61d71c211cb37e97df328fb88d33cdfbb)

PR #3945, opened 2018-01-18
Fix Bug 1428583 - Disable thumbnails if all open windows are in Private Browsing mode

sarracini:
Bails out of getting a screenshot of highlights and top sites if the most recent window is a private window. I think doing it here rather than in ```BackgroundPageThumbs.jsm``` has 3 advantages that I can think of...
1. If there are other components of ffx that depend on BackgroundPageThumbs to get a screenshot (even in the future) and maybe don't  care about private browsing, we shouldn't change the implementation for them
2. If we do it earlier can avoid some unnecessary code execution by doing it as soon as we know we don't need to fetch them
3. We have finer control over when to get or not get screenshots

Disadvantages:
1. More code, in more places. We have to remember to explicitly tell any feeds that need a screenshot to check for private browsing in the future. 

So here's the problem with the patch... right now if you run it with ```./mach run --private-window``` it bails out of fetching images which is what we want, but then opening a new non private window still doesn't trigger a refresh to get the images. It will wait in the next system_tick.

Any suggestions on how to deal with this @Mardak ?

### DIFF
--- a/system-addon/lib/Screenshots.jsm
+++ b/system-addon/lib/Screenshots.jsm
@@ -18,6 +18,10 @@ XPCOMUtils.defineLazyServiceGetter(this, "MIMEService",
   "@mozilla.org/mime;1", "nsIMIMEService");
 XPCOMUtils.defineLazyModuleGetter(this, "OS",
   "resource://gre/modules/osfile.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PrivateBrowsingUtils",
+  "resource://gre/modules/PrivateBrowsingUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "Services",
+  "resource://gre/modules/Services.jsm");
 
 const GREY_10 = "#F9F9FA";
 
@@ -73,6 +77,11 @@ this.Screenshots = {
    @ @param onScreenshot {function} Callback for when the screenshot loads
    */
   async maybeCacheScreenshot(link, url, property, onScreenshot) {
+    const win = Services.wm.getMostRecentWindow(null);
+    const isPrivate = PrivateBrowsingUtils.isWindowPrivate(win);
+    if (isPrivate) {
+      return;
+    }
     // Nothing to do if we already have a pending screenshot or
     // if a previous request failed and returned null.
     const cache = link.__sharedCache;
--- a/system-addon/lib/Screenshots.jsm
+++ b/system-addon/lib/Screenshots.jsm
@@ -68,6 +68,22 @@ this.Screenshots = {
   },
 
   /**
+   * Checks if all the open windows are private browsing windows. If so, we do not
+   * want to collect screenshots. If there exists at least 1 non-private window,
+   * we are ok to collect screenshots.
+   */
+  _shouldGetScreenshots() {
+    const windows = Services.wm.getEnumerator("navigator:browser");
+    while (windows.hasMoreElements()) {
+      if (!PrivateBrowsingUtils.isWindowPrivate(windows.getNext())) {
+        // As soon as we encounter 1 non-private window, screenshots are fair game.
+        return true;
+      }
+    }
+    return false;
+  },
+
+  /**
    * Conditionally get a screenshot for a link if there's no existing pending
    * screenshot. Updates the cached link's desired property with the result.
    *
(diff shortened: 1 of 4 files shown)


Mardak:
Did you consider our `Screenshots.jsm`? Seems like a middleground between doing it in each Feed vs in `BackgroundPageThumbs`.

Looking at only the most recent window doesn't seem right either. If the user initially had a non-private then opened a private, we would end up stopping, but that's not that different from your example of starting with private then opening a non-private… ?

sarracini:
Yeah, I'll put it at the top of ```maybeCacheScreenshot()``` and just return early? 

The only other way I can think of it is if we somehow listen on a new window event with the nsIWindowMediator, and attempt to get screenshots then?

If the request for the screenshot was coming from content we could access ```action._target.browser.ownerGlobal``` and look it up from there the request for a screenshot comes from ```INIT``` which doesn't come from content which doesn't have a ```_target``` on it's action from what I saw

Mardak:
Right. Our screenshots are independent of any window as we're effectively a background service. The pages being screenshotted should not contain those from private windows as those shouldn't trigger places events, etc.

To check all windows, you would use `Services.wm.getEnumerator("navigator:browser")`, e.g., https://searchfox.org/mozilla-central/source/browser/base/content/browser.js#295-298

There's even an existing usage looking for private windows ;) https://searchfox.org/mozilla-central/source/browser/base/content/browser.js#6706-6708

Maybe a step back is what urls are actually attempting screenshots when opening firefox with a single private window?

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -102,4 +119,15 @@ describe("Screenshots", () => {
...
+
+  describe("#_shouldGetScreenshots", () => {
+    it("should use private browsing utils to determine if a window is private", () => {
+      Screenshots._shouldGetScreenshots();
+      assert.calledOnce(global.PrivateBrowsingUtils.isWindowPrivate);
+    });
+    it("should return true if there exists at least 1 non-private window", () => {
+      global.PrivateBrowsingUtils.isWindowPrivate = () => false;
+      assert.isTrue(Screenshots._shouldGetScreenshots());
+    });
```
Mardak: A test for the `false` case would be good

system-addon/lib/Screenshots.jsm
```diff
@@ -63,6 +67,24 @@ this.Screenshots = {
...
+   * we are ok to collect screenshots.
+   */
+  _shouldGetScreenshots() {
+    const windows = Services.wm.getEnumerator("navigator:browser");
+    let shouldGetScreenshots = false;
+    while (windows.hasMoreElements()) {
+      if (!PrivateBrowsingUtils.isWindowPrivate(windows.getNext())) {
+        // As soon as we encounter 1 non-private window, screenshots are fair game.
+        shouldGetScreenshots = true;
+        break;
```
Mardak: nit: could just conditionally `return true;` and `return false;` at the end without the variable or `break`.

system-addon/test/unit/lib/Screenshots.test.js
sarracini: Still figuring out how to break out of that while(true) loop if since it'll go on forever in this test due to ```.hasMoreElements()``` being mocked to be forever true 😞

Mardak: Could stub `getEnumerator` to return something different for a single test?
```js
let more = 2;
sandbox.stub(Services.wm, "getEnumerator").returns({hasMoreElements() { return more--; }});
```

Set more based on testing 2 private windows, 1, 0, etc.

sarracini:
@Mardak  You're right about the performance concerns. Maybe I can put this check after we've decided we need a screenshot, but before we actually get it? That way we can delay it until after the cache? Might reduce the amount of look ups

sarracini:
Or we can actually see if perf takes a hit and then do optimizations later 🎉

system-addon/test/unit/lib/Screenshots.test.js
```diff
@@ -102,4 +131,23 @@ describe("Screenshots", () => {
...
+    it("should return false if there exists private windows", () => {
+      global.PrivateBrowsingUtils.isWindowPrivate = () => true;
```
Mardak: Let's check that this gets called the expected number of times. Then good to land!

Mardak:
FYI, this landed with 2 commits with the latter one basically undoing the former.

as-pine-proxy:
[Treeherder: 095ddb5ad86c1104d4c9d4a55293f01fc3c4274d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=095ddb5ad86c1104d4c9d4a55293f01fc3c4274d)

PR #3946, opened 2018-01-19
Fix Bug 1400962 - [RTL] Flip some icons on the Activity Stream

vikasmahato:


Mardak:
At a quick glance, I think this is what we want. Although I would think there should be some `sass`-y way to not need to just copy/paste the whole style (even if it's just one selector + rule). Probably `@include` ?

vikasmahato:
@Mardak I have updated the pull request

Mardak:
I took a couple screenshots with this PR:

LTR:
screen shot 2018-01-19 at 9 20 21 pm

RTL:
screen shot 2018-01-19 at 9 20 50 pm

Looks good! I'll double check with the reporter about the highlights icon.

as-pine-proxy:
[Treeherder: f5a3ce3a00945a68c4ce64a9a20320136a2c4718](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f5a3ce3a00945a68c4ce64a9a20320136a2c4718)

PR #3949, opened 2018-01-20
fix(TopSites) avoid setState race in TopSites , use eslint to prevent…

dmose:
similar, fix bug 1431566.


as-pine-proxy:
[Treeherder: 190d0e4ec049d6d87ed27d6bbf9fb56872f252dc](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=190d0e4ec049d6d87ed27d6bbf9fb56872f252dc)

PR #3950, opened 2018-01-20
changed scroll to auto: Closes #3696

### LINKED ISSUE #3696: Vertical scrollbar shows up all the time, even when there is nothing to scroll
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1407770

Testing on v58.0a1 (2017-10-11) on a 1920x1080 screen at 125% scaling. Should be reproducible regardless of the window size, resolution etc. though.

STR: Most easily noticeable if you have unchecked all items under 'New Tab Preferences' so that the page is blank, except for the Firefox logo and the cog icon.

AR: disabled scrollbar is visible:
capture

ER: it shouldn't be.

Probably because `body` has `overflow-y: scroll;`. Changing it to `overflow-y: auto;` fixes things.

shikhar-scs:
As suggested by @FatOrangutan and @shaweglea at #3696, I've made the required changes.
Please review.

Also solves [bug 1418515](https://bugzilla.mozilla.org/show_bug.cgi?id=1418515) at bugzilla

### DIFF
--- a/system-addon/content-src/styles/_activity-stream.scss
+++ b/system-addon/content-src/styles/_activity-stream.scss
@@ -13,7 +13,7 @@ body {
   color: $text-primary;
   font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Ubuntu', 'Helvetica Neue', sans-serif;
   font-size: 16px;
-  overflow-y: scroll;
+  overflow-y: auto;
 }
 
 h1,


Mardak:
The `overflow: scroll` was added to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1400920 so we don't want to regress that bug when fixing this bug. Could you test to see if this breaks that other bug again?

shikhar-scs:
@Mardak I tried setting up `activity-stream` but its taking forever (already been 3.5 hrs) to set up (I might be doing it wrong). So, as this is just a one line change and you already have (assuming) all of it set up, can you please try and check this out
>Could you test to see if this breaks that other bug again?

If not, please tag someone else from the community to check this. Would just take a minute or two (I've already spent so much time) to check the same and this would greatly save either of our time. :smile: 
This is just a small request :)

shikhar-scs:
@Mardak ?

shikhar-scs:
@Mardak any updates in this please ?

Mardak:
If someone can test that this fixes bug 1407770 without regressing bug 1400920, we can create a new PR and merge it.

PR #3958, opened 2018-01-30
Fix Bug 1434372 - Refactor code that adds a new Topsite, save vs add

piatra:


system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -127,28 +125,19 @@ export class _TopSites extends React.PureComponent {
...
-                  editMode={true}
-                  onClose={this.onFormClose}
-                  dispatch={this.props.dispatch}
-                  intl={this.props.intl} />
-              </div>
+          {(showAddForm || editIndex >= 0) &&
+          <div className="edit-topsites">
+            <div className="modal-overlay" onClick={this.onFormClose} />
+            <div className="modal">
+              <TopSiteForm
```
Mardak: @rlr probably out of scope here, but the modal including `<TopSiteForm>` probably don't even need to be within `TopSites` anymore as we don't need to overlay one list over another?

rlr: Yeah, this can probably be cleaned up more. Although `TopSites` is currently keeping track of the state that shows/hides this. Also, all that modal boilerplate should probably just go inside the component itself now too.

Mardak: Actually.. what's the `TopSites` component state vs redux `TopSites.editForm` state?

rlr: yeah, that's to trigger edit from the `ContextMenu`. So we could switch to only using that and break out the `TopSiteForm` into somewhere else. Or keep it here and just get rid of the state.

piatra:
Updated to use appropriate headers.
> if you click on an empty placeholder is that "editing a topsite" or "adding a topsite"?
> it should be add action

Mardak:
Not sure what was expected for the title vs button:
add-edit

system-addon/content-src/components/TopSites/TopSites.jsx
Mardak: Let's just leave this inline modal for now and get rid of the `TopSites` component state to just use the redux `TopSites.editForm` state instead.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -127,28 +125,19 @@ export class _TopSites extends React.PureComponent {
...
-                  label={editSite.label || editSite.hostname || ""}
-                  url={editSite.url || ""}
-                  index={editIndex}
-                  editMode={true}
-                  onClose={this.onFormClose}
-                  dispatch={this.props.dispatch}
-                  intl={this.props.intl} />
-              </div>
+          {(showAddForm || editIndex >= 0) &&
+          <div className="edit-topsites">
```
Mardak: Should be able to keep the original indentation of the `<div>` and below

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -145,12 +147,12 @@ export class TopSiteForm extends React.PureComponent {
...
+          {editMode &&
             <button className="done save" type="submit" onClick={this.onSaveButtonClick}>
-              <FormattedMessage id="topsites_form_save_button" />
+              <FormattedMessage id={this.props.url ? "topsites_form_save_button" : "topsites_form_add_button"} />
             </button>
           }
-          {!this.props.editMode &&
+          {!editMode &&
             <button className="done add" type="submit" onClick={this.onAddButtonClick}>
               <FormattedMessage id="topsites_form_add_button" />
```
Mardak: Looks like these buttons could be cleaned up by combining the `!editMode` and `editMode` cases and using a shared click handler where `onAddButtonClick` and `onSaveButtonClick` has a bunch of shared code already.

Mardak: @rlr Some reason I thought we discused `TOP_SITES_PIN` vs `TOP_SITES_INSERT` where PIN without an index or index of -1 would mean INSERT ??

rlr: @Mardak I *think* that was `TOP_SITES_ADD`+`TOP_SITES_EDIT` vs TOP_SITES_INSERT`? Or maybe it was `TOP_SITES_DROP`+`TOP_SITES_INSERT`? Gah, don't remember. But, the difference between those two is that pin replaces and insert potentially causes a shift.

rlr: But hmmm, I guess if you are pinning, the slot should be available and insert would do the same thing. Yeah, I think they could be combined.

Mardak: Oh right. we use INSERT for dropping to a specific index. It was ADD vs INSERT

Mardak: Just to clarify, we need to keep both `TOP_SITES_PIN` and `TOP_SITES_INSERT` actions in general to allow for adding/editing to specific positions vs shifting over pinned items. Right now this Edit UI uses PIN for edit/Save and INSERT for Add. We can keep it like that.

There is maybe a event handling simplification by making the `TOP_SITES_PIN` without an index (or as -1) behave like `TOP_SITES_INSERT` without an index (i.e., insert at position 0). This makes `TopSitesFeed` a little more complex, but it would allow the component code to always just dispatch `{type: at.TOP_SITES_PIN, data: {site, index: this.props.index}}`

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -117,7 +119,7 @@ export class TopSiteForm extends React.PureComponent {
         <section className="edit-topsites-inner-wrapper">
           <div className="form-wrapper">
             <h3 className="section-title">
-              <FormattedMessage id={this.props.editMode ? "topsites_form_edit_header" : "topsites_form_add_header"} />
+              <FormattedMessage id={this.props.url ? "topsites_form_edit_header" : "topsites_form_add_header"} />
```
Mardak: Let's pull out `this.props.url` to be clear that this decision matches up with the button text below.
```js
// For UI purposes, editing without an existing link is "add"
const showAsAdd = !this.props.url;
```

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -34,45 +33,48 @@ export class TopSiteForm extends React.PureComponent {
...
-    }
-  }
 
-  onSaveButtonClick(ev) {
-    ev.preventDefault();
-    if (this.validateForm()) {
-      let site = {url: this.cleanUrl()};
-      if (this.state.label !== "") {
-        site.label = this.state.label;
+      if (this.props.index >= 0) {
```
Mardak: Similarly we want a comment here as why we're using a check different from the UI bits.

as-pine-proxy:
[Treeherder: 58c6958957343f7008ea37ffd10a062cd56ee85a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=58c6958957343f7008ea37ffd10a062cd56ee85a)

PR #3961, opened 2018-01-31
Bug 1428106 - Add skipLocal to ComponentPerfTimer

k88hudson:
This fix is to prevent redux from triggering additional renders

### DIFF
--- a/system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
+++ b/system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
@@ -120,7 +120,9 @@ export class ComponentPerfTimer extends React.Component {
       this.props.dispatch(ac.SendToMain({
         type: at.SAVE_SESSION_PERF_DATA,
         // highlights_data_late_by_ms, topsites_data_late_by_ms.
-        data: {[`${this.props.id}_data_late_by_ms`]: value}
+        data: {[`${this.props.id}_data_late_by_ms`]: value},
+        // we don't want to trigger a redux update since this is triggered by render
+        meta: {skipLocal: true}
       }));
     } catch (ex) {
       // If this failed, it's likely because the `privacy.resistFingerprinting`
@@ -144,7 +146,9 @@ export class ComponentPerfTimer extends React.Component {
 
       this.props.dispatch(ac.SendToMain({
         type: at.SAVE_SESSION_PERF_DATA,
-        data
+        data,
+        // we don't want to trigger a redux update since this is triggered by render
+        meta: {skipLocal: true}
       }));
     } catch (ex) {
       // If this failed, it's likely because the `privacy.resistFingerprinting`
--- a/system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
+++ b/system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
@@ -117,12 +117,10 @@ export class ComponentPerfTimer extends React.Component {
       // value has to be Int32.
       const value = parseInt(this.perfSvc.getMostRecentAbsMarkStartByName(dataReadyKey) -
                              this.perfSvc.getMostRecentAbsMarkStartByName(firstRenderKey), 10);
-      this.props.dispatch(ac.SendToMain({
+      this.props.dispatch(ac.OnlyToMain({
         type: at.SAVE_SESSION_PERF_DATA,
         // highlights_data_late_by_ms, topsites_data_late_by_ms.
-        data: {[`${this.props.id}_data_late_by_ms`]: value},
-        // we don't want to trigger a redux update since this is triggered by render
-        meta: {skipLocal: true}
+        data: {[`${this.props.id}_data_late_by_ms`]: value}
       }));
     } catch (ex) {
       // If this failed, it's likely because the `privacy.resistFingerprinting`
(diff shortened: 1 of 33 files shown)


system-addon/test/unit/lib/init-store.test.js
```diff
@@ -64,6 +64,13 @@ describe("initStore", () => {
     store.dispatch(action);
     assert.calledWith(global.sendAsyncMessage, OUTGOING_MESSAGE_NAME, action);
   });
+  it("should not call next if meta.skipLocal is true", () => {
+    const action = ac.SendToMain({type: "FOO", meta: {skipLocal: true}});
+    const spy = globals.sandbox.spy();
+    store.subscribe(spy);
+    store.dispatch(action);
+    assert.notCalled(spy);
+  });
```
dmose: I'd suggest splitting this into (blank-line delimited) setup/execute/verify sections for readabilty.

system-addon/content-src/lib/init-store.js
```diff
@@ -38,10 +38,13 @@ function mergeStateReducer(mainReducer) {
...
  */
 const messageMiddleware = store => next => action => {
+  const skipLocal = action.meta && action.meta.skipLocal;
   if (au.isSendToMain(action)) {
     sendAsyncMessage(OUTGOING_MESSAGE_NAME, action);
   }
-  next(action);
+  if (!skipLocal) {
+    next(action);
+  }
```
dmose: If we stick with this solution, update the JSDoc here?

system-addon/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
```diff
@@ -144,7 +146,9 @@ export class ComponentPerfTimer extends React.Component {
 
       this.props.dispatch(ac.SendToMain({
         type: at.SAVE_SESSION_PERF_DATA,
-        data
+        data,
+        // we don't want to trigger a redux update since this is triggered by render
+        meta: {skipLocal: true}
```
dmose: It seems like SendToMain sending local by default is footgun due to it's non-intuitiveness.  Would it be a better solution to rename the current action SendToLocalAndMain and make SendToMain never send to local?

k88hudson: I was actually just thinking about that, it would require some broader changes though.  How would you feel if I file a follow-up bug to address that?

dmose: Isn't it just "search and replace" and then "add an action"?  Or is there more to the broader changes you describe?

k88hudson: There are some actions which do actually need to be copied, so I'll need to make sure that those that do are changed appropriately and tested

k88hudson: For example, blocking

dmose: I guess the reason I'm hesistant to spin it off is that it's not like this thing has been a huge push, as far as I'm away, and all the mental context for coding and reviewing is here now.  I'll let you make the call.  If it's truly more work and worth the context-switch to pick up again later, than go ahead and spin it off.

Mardak: If we go down this route, then maybe `ac.OnlyToMain` and `ac.AlsoToMain`?

And do we actually want to use "skipOrigin"? It was added when Actions.jsm was initially implemented, but never used, but seems like what we want to use via various `actionCreators`: https://github.com/mozilla/activity-stream/blob/master/system-addon/common/Actions.jsm#L104

There is a skipMain only used by TopStoriesFeed: https://github.com/mozilla/activity-stream/search?q=skipMain and that could probably be `ac.OnlyToContent` (this happens to be a single content target.. but for broadcast… `ac.BroadcastToContent` -> `ac.AlsoToAllContent` and `ac.OnlyToAllContent`? Unless we want to specially call out single targets, e.g., `ac.OnlyToOneContent`, where that seems to match our current usage of preferring broadcast in the common case.)

k88hudson: I like Also and Only since they don't conflict with the old names and clearly indicate what's going on. For Broadcast, we haven't really run into a usecase where that has to skip the main process, so dmose and I were thinking we can just leave as is

k88hudson:
Ok, let me know what you think about this - I've made the following changes with appropriate JS doc / description updates: 

SendToMain becomes AlsoToMain
SendToMain with skipLocal=true becomes OnlyToMain
SendToContent becomes AlsoToOneContent
SendToContent with skipMain=true becomes OnlyToOneContent
SendToPreloaded becomes AlsoToPreloaded
BroadcastToContent stays as is

dmose:
I :heart: these naming choices.  Looking through the modified code, it feels much more readable to me, in part because Also refers to the dispatch call itself, making it clearer that all your really doing here is some of extended dispatch.  I'll probably review this tomorrow.

as-pine-proxy:
[Treeherder: aba14bb1b376403f869cd8ed0d348e794c7af758](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=aba14bb1b376403f869cd8ed0d348e794c7af758)

PR #3962, opened 2018-01-31
chore (prefs): migrate topSitesCount to topSitesRows

rlr:
@Mardak, this was more tedious then I thought but it *should* help make it easier to implement the wider top sites with 8 across. And it's a better match to how we are allowing users to change # of rows. Thoughts? Did I do the migration right? r?

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -146,9 +146,13 @@ function onBrowserReady() {
     if (rows <= 0) {
       Services.prefs.setBoolPref("browser.newtabpage.activity-stream.showTopSites", false);
     } else {
-      // Assume we want a full row (6 sites per row)
-      Services.prefs.setIntPref("browser.newtabpage.activity-stream.topSitesCount", rows * 6);
+      Services.prefs.setIntPref("browser.newtabpage.activity-stream.topSitesRows", rows);
     }
+  });
+
+  // Old activity stream topSitesCount pref showed 6 per row
+  migratePref("browser.newtabpage.activity-stream.topSitesCount", count => {
+    Services.prefs.setIntPref("browser.newtabpage.activity-stream.topSitesRows", Math.ceil(count / 6));
   });
 }
 
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -6,8 +6,8 @@
 const {actionTypes: at} = Components.utils.import("resource://activity-stream/common/Actions.jsm", {});
 const {Dedupe} = Components.utils.import("resource://activity-stream/common/Dedupe.jsm", {});
 
-const TOP_SITES_DEFAULT_LENGTH = 6;
-const TOP_SITES_SHOWMORE_LENGTH = 12;
+const TOP_SITES_DEFAULT_ROWS = 2;
+const TOP_SITES_MAX_SITES_PER_ROW = 6;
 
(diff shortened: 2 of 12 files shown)


system-addon/common/Reducers.jsm
```diff
@@ -6,8 +6,8 @@
 const {actionTypes: at} = Components.utils.import("resource://activity-stream/common/Actions.jsm", {});
 const {Dedupe} = Components.utils.import("resource://activity-stream/common/Dedupe.jsm", {});
 
-const TOP_SITES_DEFAULT_LENGTH = 6;
-const TOP_SITES_SHOWMORE_LENGTH = 12;
+const TOP_SITES_DEFAULT_ROWS = 2;
+const TOP_SITES_MAX_SITES_PER_ROW = 6;
```
Mardak: Seems like we actually want a constants file that is usable by `lib/` and `content-src/`, but.. oh well…? There is `content-src/components/TopSites/TopSitesConstants.js` as well as `content-src/lib/constants.js` but not quite what we need here. @k88hudson any suggestions? I suppose most straightforward would just be a `common/Constants.jsm` with a bunch of `this.FOO = val;` and `this.EXPORTED_SYMBOLS = ["FOO, …];` ?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -246,8 +245,8 @@ this.TopSitesFeed = class TopSitesFeed {
     // Don't insert any pins past the end of the visible top sites. Otherwise,
     // we can end up with a bunch of pinned sites that can never be unpinned again
     // from the UI.
-    const {topSitesCount} = this.store.getState().Prefs.values;
-    if (index >= topSitesCount) {
+    const {topSitesRows} = this.store.getState().Prefs.values;
```
Mardak: This changes for this method might be cleaner as just
```js
const topSitesCount = this.store.getState().Prefs.values.topSitesRows * TOP_SITES_MAX_SITES_PER_ROW;
```

system-addon/bootstrap.js
```diff
@@ -150,6 +150,12 @@ function onBrowserReady() {
       Services.prefs.setIntPref("browser.newtabpage.activity-stream.topSitesCount", rows * 6);
```
Mardak: We should migrate the old `newtabpage.rows` pref directly to `topSiteRows`.

system-addon/bootstrap.js
```diff
@@ -150,6 +150,12 @@ function onBrowserReady() {
...
     }
   });
+
+  // Do a one time migration of browser.newtabpage.activity-stream.topSitesCount pref
```
Mardak: nit: combine the comment here and below to this line
```js
// Old activity stream topSitesCount pref showed 6 per row
```

system-addon/bootstrap.js
Mardak: (Yes, I realize there's a conflict if both prefs need to be migrated, but oh well. :p)

rlr:
Yeah the plan is to bump up the pref. But then still need to fix telemetry somehow. And... ugh, didn't think about deduping. But hopefully we can punt on that?

as-pine-proxy:
[Treeherder: 662e137c4cff47ac9ea914b46ef5e96161a81fec](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=662e137c4cff47ac9ea914b46ef5e96161a81fec)

PR #3963, opened 2018-02-01
Fix Bug 1434937 topsites refactor

piatra:
Follow up fixes

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -25,7 +25,7 @@ const INITIAL_STATE = {
     // The history (and possibly default) links
     rows: [],
     // Used in content only to dispatch action from
-    // context menu to TopSites.
+    // context menu to TopSiteForm.
     editForm: null
   },
   Prefs: {
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -24,8 +24,7 @@ const INITIAL_STATE = {
     initialized: false,
     // The history (and possibly default) links
     rows: [],
-    // Used in content only to dispatch action from
-    // context menu to TopSiteForm.
+    // Used in content only to dispatch action to TopSiteForm.
     editForm: null
   },
   Prefs: {
@@ -92,7 +91,7 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       return Object.assign({}, prevState, {initialized: true, rows: action.data});
     case at.TOP_SITES_EDIT:
-      return Object.assign({}, prevState, {editForm: {visible: true, index: action.data.index}});
+      return Object.assign({}, prevState, {editForm: {index: action.data.index}});
     case at.TOP_SITES_CANCEL_EDIT:
       return Object.assign({}, prevState, {editForm: null});
     case at.SCREENSHOT_UPDATED:
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -25,11 +25,8 @@ const INITIAL_STATE = {
     // The history (and possibly default) links
     rows: [],
     // Used in content only to dispatch action from
-    // context menu to TopSitesEdit.
-    editForm: {
-      visible: false,
-      index: -1
-    }
+    // context menu to TopSites.
+    editForm: null
   },
   Prefs: {
     initialized: false,
@@ -95,7 +92,7 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       return Object.assign({}, prevState, {initialized: true, rows: action.data});
     case at.TOP_SITES_EDIT:
-      return Object.assign({}, prevState, {editForm: {index: action.data.index}});
+      return Object.assign({}, prevState, {editForm: {visible: true, index: action.data.index}});
     case at.TOP_SITES_CANCEL_EDIT:
       return Object.assign({}, prevState, {editForm: null});
     case at.SCREENSHOT_UPDATED:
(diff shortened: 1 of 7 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -284,10 +289,16 @@ this.TopSitesFeed = class TopSitesFeed {
    * Handle an insert (drop/add) action of a site.
    */
   insert(action) {
+    let {index} = action.data;
+    // Negative index will prepend the TopSite in the first position
+    if (index === -1) {
+      index = 0;
```
Mardak: You can skip the `|| 0` below with:
```js
// Treat invalid pin index values (e.g., -1, undefined) as insert in the first position
if (!(index > 0)) {
  index = 0;
}
```

(NB: The `>` implicitly converts `undefined` to `NaN` and `null` to `0`. And comparisons with `NaN` is `false`, so that's why the negation.)

system-addon/common/Reducers.jsm
```diff
@@ -25,11 +25,8 @@ const INITIAL_STATE = {
     // The history (and possibly default) links
     rows: [],
     // Used in content only to dispatch action from
-    // context menu to TopSitesEdit.
-    editForm: {
-      visible: false,
-      index: -1
-    }
+    // context menu to TopSiteForm.
```
Mardak: This comment isn't right as it'll be triggered from the Add link/button too.

system-addon/common/Reducers.jsm
```diff
@@ -95,7 +92,7 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       return Object.assign({}, prevState, {initialized: true, rows: action.data});
     case at.TOP_SITES_EDIT:
-      return Object.assign({}, prevState, {editForm: {index: action.data.index}});
+      return Object.assign({}, prevState, {editForm: {visible: true, index: action.data.index}});
```
Mardak: `visible` shouldn't be necessary as we should only be showing the form if `editForm` is not `null`?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -38,44 +39,25 @@ export class TopSiteForm extends React.PureComponent {
 
     if (this.validateForm()) {
       const site = {url: this.cleanUrl()};
+      const index = this.props.index || -1;
```
Mardak: This will incorrectly make edits to the `0` index become inserts. We should just `const {index} = this.props;`

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -38,44 +39,25 @@ export class TopSiteForm extends React.PureComponent {
...
-        this.insertTopSite(site);
-      }
+      this.props.dispatch(ac.SendToMain({
+        type: at.TOP_SITES_PIN,
+        data: {site, index}
+      }));
+      this.props.dispatch(ac.UserEvent({
+        source: TOP_SITES_SOURCE,
+        event: "TOP_SITES_EDIT",
+        action_position: index
```
Mardak: We'll probably want to confirm with @ncloudioj (?) that it's okay to get instead a user event for "TOP_SITES_ADD", it'll be a "TOP_SITES_EDIT" with `-1` index.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -161,7 +143,6 @@ export class TopSiteForm extends React.PureComponent {
 }
 
 TopSiteForm.defaultProps = {
-  label: "",
-  url: "",
+  TopSite: null,
   index: -1
```
Mardak: I don't think we need to keep around `defaultProps` anymore as we pass in the values from the one usage in `TopSites`. But it should also be safe to just keep it…

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -126,14 +113,13 @@ export class _TopSites extends React.PureComponent {
...
-          {(showAddForm || editIndex >= 0) &&
+          {(editForm && editForm.visible) &&
             <div className="edit-topsites">
               <div className="modal-overlay" onClick={this.onFormClose} />
               <div className="modal">
                 <TopSiteForm
-                  label={editSite.label || editSite.hostname || ""}
-                  url={editSite.url || ""}
-                  index={editIndex}
+                  TopSite={props.TopSites.rows[editForm.index]}
```
Mardak: nit: to match with the existing style, it should probably just be `site={…}`?

system-addon/test/unit/content-src/components/TopSites.test.jsx
```diff
@@ -533,16 +533,16 @@ describe("<TopSiteForm>", () => {
...
+          data: {site: {url: "http://valid.com"}, index: -1},
           meta: {from: "ActivityStream:Content", to: "ActivityStream:Main"},
-          type: at.TOP_SITES_INSERT
+          type: at.TOP_SITES_PIN
         }
       );
     });
   });
 
   describe("edit existing Topsite", () => {
```
Mardak: let's add a test for editing index 0 getting the correct action

system-addon/content-src/components/TopSites/TopSiteForm.jsx
ncloudioj: Yes, you can either of them here, and -1 could be used as a valid "action_position". Please consider adding a comment in the `data_dictionary.md` to explain what action_position -1 means for this event.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -107,13 +99,8 @@ export class _TopSites extends React.PureComponent {
       header: {id: "settings_pane_topsites_header"},
       body: {id: "settings_pane_topsites_body"}
     };
-    const {showAddForm} = this.state;
-    const {editForm} = this.props.TopSites;
-    let {editIndex} = this.state;
-    if (editIndex < 0 && editForm) {
-      editIndex = editForm.index;
-    }
-    const editSite = this.props.TopSites.rows[editIndex] || {};
```
Mardak: 👍 @rlr ^^ !

PR #3964, opened 2018-02-01
Upgrade to React 16.2 and add error boundaries (fix bug 1421441)

dmose:
This upgrades React to version 16 and adds error boundaries.   I propose to land all of the commits unsquashed, except the last one, which will be removed, as it's only there to help testing.  In the future, I think we're gonna want storybook views of the various error states to keep them from regressing; adding storybook support might want to be a spinoff bug from this.

The intent is for telemetry to happen not as part of this landing, but as part of the Sentry landing, which I'll be working on soon.

The last commit is what's preventing a clean travis run.  All the tests, including our mochitests, pass on my machine; once this gets r+, I'll remove that last commit and force push, which will cause travis to then run and (with any luck!) sign off for landing.

How to test:

1. pull this branch
2. `git checkout HEAD^` # this gets rid of the faked errors to test the working thing as it will land.
3. `rm -fr node_modules` # not sure why we need this, but...
4. `npm install`
5. `npm run buildmc && npm run startmc`
6.  `./mach build && ./mach run` # in the mercurial tree

When you're done testing the working stuff, you can then checkout the entire branch so that you get the faked error commit too, and then you'll see the various sections as they will look if something blows up at runtime; opening the preferences gear will show what bustage there does too.  

Be careful not to run "buildmc", however, as this will explode trying to regenerate the (artificially busted) pre-rendered files.

Once you're satisfied with this, uncomment the "throw" on line 66 of CollapsibleSection.jsx to see what happens if something gets caught outside the main error boundaries.

Error boundaries pix:

all sub-chunks of the page throwing exceptions:
all sub-chunks of the page throwing exceptions 

explosions not captured by the sub-chunks crash the whole page:
explosions not captured by the sub-chunks crash the whole page

r?@rlr


system-addon/test/unit/content-src/components/Search.test.jsx
```diff
@@ -31,9 +31,7 @@ describe("<Search>", () => {
 
     const wrapper = mountWithIntl(<Search {...DEFAULT_PROPS} />);
 
-    assert.calledOnce(spy);
-    assert.equal(spy.firstCall.args[0], "ContentSearchClient");
-    assert.equal(spy.firstCall.args[1], wrapper.instance());
+    assert.calledOnce(spy.withArgs("ContentSearchClient", wrapper.instance()));
```
rlr: i dont think i had seen this done like this before. cool!

system-addon/test/unit/content-src/components/ErrorBoundary.test.jsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  it("should call window.location.reload(true) if .reload-button clicked", () => {
+    const fakeWindow = {location: {reload: sinon.spy()}};
+    const wrapper = shallow(<ErrorBoundaryFallback windowObj={fakeWindow} />);
+
+    wrapper.find(".reload-button").simulate("click");
+
+    assert.calledOnce(fakeWindow.location.reload);
+    assert.calledWithExactly(fakeWindow.location.reload, true);
```
rlr: can you do the same trick here? i.e. could these two lines be `assert.calledOnce(fakeWindow.location.reload.withArgs(true));` instead?

rlr:
I did a first pass at the changes and it looks pretty straightforward and good. I'm going to go through the testing steps next. Stay tuned....

system-addon/test/unit/content-src/components/ErrorBoundary.test.jsx
dmose: I would expect that to work, but to me, that feels harder to read.  Maybe it's just because I'm stockholm-syndromed into the old way.  But I think I'll leave it as is.

as-pine-proxy:
[Treeherder: 71e750ae526aa36cdfcf1006728ae0646b5c0150](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=71e750ae526aa36cdfcf1006728ae0646b5c0150)

PR #3968, opened 2018-02-05
Fix Bug 1435863 - Add source map support for css, js in Activity Stream

k88hudson:
I've added source maps such that they will be included in all builds for Firefox (they are external, so won't be loaded until devtools are opened, but they do add ~`360kb` of additional files to be downloaded). What do you think @sarracini @Mardak? Or should we perhaps only include them for non-release channels?

### DIFF
--- a/system-addon/jar.mn
+++ b/system-addon/jar.mn
@@ -17,13 +17,27 @@
   content/vendor/react-intl.js (./vendor/react-intl.js)
   content/vendor/redux.js (./vendor/redux.js)
   content/vendor/react-redux.js (./vendor/react-redux.js)
-  content/data/ (./data/*)
+  content/data/content/assets/ (./data/content/assets/*)
+  content/data/content/tippytop/ (./data/content/tippytop/*)
+  content/data/content/activity-stream.bundle.js (./data/content/activity-stream.bundle.js)
+#ifndef RELEASE_OR_BETA
+  content/data/content/activity-stream.bundle.js.map (./data/content/activity-stream.bundle.js.map)
+#endif
 #ifdef XP_MACOSX
   content/css/activity-stream.css (./css/activity-stream-mac.css)
+#ifndef RELEASE_OR_BETA
+  content/css/activity-stream-mac.css.map (./css/activity-stream-mac.css.map)
+#endif
 #elifdef XP_WIN
   content/css/activity-stream.css (./css/activity-stream-windows.css)
+#ifndef RELEASE_OR_BETA
+  content/css/activity-stream-windows.css.map (./css/activity-stream-windows.css.map)
+#endif
 #else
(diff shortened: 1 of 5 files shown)


k88hudson:
Ok, I adjusted jar.mn to only include maps on non-release/beta

system-addon/jar.mn
```diff
@@ -17,13 +17,28 @@
   content/vendor/react-intl.js (./vendor/react-intl.js)
   content/vendor/redux.js (./vendor/redux.js)
   content/vendor/react-redux.js (./vendor/react-redux.js)
-  content/data/ (./data/*)
+  content/data/content/locales.json (./data/locales.json)
```
Mardak: `locales.json` doesn't actually make it to mozilla-central

system-addon/jar.mn
```diff
@@ -17,13 +17,28 @@
...
+  content/data/content/assets/ (./data/content/assets/*)
+  content/data/content/tippytop/ (./data/content/tippytop/*)
+  content/data/content/activity-stream.bundle.js (./data/content/activity-stream.bundle.js)
+#ifndef RELEASE_OR_BETA
+  content/data/content/activity-stream.bundle.js.map (./data/content/activity-stream.bundle.js.map)
+#endif
 #ifdef XP_MACOSX
   content/css/activity-stream.css (./css/activity-stream-mac.css)
+  #ifndef RELEASE_OR_BETA
```
Mardak: Not sure if indenting actually works?  At least I don't see other places indenting…

k88hudson: Based on manual testing, it looks like this does work

k88hudson: Hm, nevermind. The inner block seems to get ignored

system-addon/jar.mn
k88hudson: whoops, wrong path!

k88hudson:
Ok, this should be good now 👍

system-addon/jar.mn
```diff
@@ -17,13 +17,28 @@
   content/vendor/react-intl.js (./vendor/react-intl.js)
   content/vendor/redux.js (./vendor/redux.js)
   content/vendor/react-redux.js (./vendor/react-redux.js)
-  content/data/ (./data/*)
+  content/data/locales.json (./data/locales.json)
```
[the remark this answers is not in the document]
k88hudson: Oh right, that's only used in tests 🐳

k88hudson:
*Actually* good now this time

as-pine-proxy:
[Treeherder: 9850ee4c059f2805f7f3462bda936c6cff9bba30](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=9850ee4c059f2805f7f3462bda936c6cff9bba30)

PR #3969, opened 2018-02-06
Fix Bug 1385944 - support wider layout

rlr:
@Mardak turns out we don't dedupe from Top Sites anymore 😅 

before:
screen shot 2018-02-05 at 7 22 13 pm

after:
screen shot 2018-02-05 at 7 21 12 pm

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -6,8 +6,8 @@
 const {actionTypes: at} = ChromeUtils.import("resource://activity-stream/common/Actions.jsm", {});
 const {Dedupe} = ChromeUtils.import("resource://activity-stream/common/Dedupe.jsm", {});
 
-const TOP_SITES_DEFAULT_ROWS = 2;
-const TOP_SITES_MAX_SITES_PER_ROW = 6;
+const TOP_SITES_DEFAULT_ROWS = 1;
+const TOP_SITES_MAX_SITES_PER_ROW = 8;
 
 const dedupe = new Dedupe(site => site && site.url);
 
--- a/system-addon/content-src/components/Sections/_Sections.scss
+++ b/system-addon/content-src/components/Sections/_Sections.scss
@@ -54,3 +54,19 @@
     }
   }
 }
+
+.wide-layout-enabled {
+  .sections-list {
+    .section-list {
+      @media (min-width: $break-point-widest) and (max-width: $break-point-widest + 2 * $card-width) {
+        :nth-child(3n) {
+          @include context-menu-open-left;
+        }
+      }
+
+      @media (min-width: $break-point-widest) {
+        grid-template-columns: repeat(auto-fit, $card-width-large);
+      }
+    }
+  }
+}
--- a/system-addon/content-src/components/TopSites/TopSite.jsx
+++ b/system-addon/content-src/components/TopSites/TopSite.jsx
@@ -161,7 +161,7 @@ export class TopSite extends React.PureComponent {
     const {link} = props;
     const isContextMenuOpen = this.state.showContextMenu && props.activeIndex === props.index;
     const title = link.label || link.hostname;
-    return (<TopSiteLink {...props} onClick={this.onLinkClick} onDragEvent={this.props.onDragEvent} className={isContextMenuOpen ? "active" : ""} title={title}>
+    return (<TopSiteLink {...props} onClick={this.onLinkClick} onDragEvent={this.props.onDragEvent} className={`${props.className || ""}${isContextMenuOpen ? " active" : ""}`} title={title}>
         <div>
           <button className="context-menu-button icon" onClick={this.onMenuButtonClick}>
             <span className="sr-only">
@@ -197,7 +197,7 @@ export class TopSitePlaceholder extends React.PureComponent {
   }
 
   render() {
-    return (<TopSiteLink className="placeholder" isDraggable={false} {...this.props}>
+    return (<TopSiteLink {...this.props} className={`placeholder ${this.props.className || ""}`} isDraggable={false}>
       <button className="context-menu-button edit-button icon"
        title={this.props.intl.formatMessage({id: "edit_topsites_edit_button"})}
        onClick={this.onEditButtonClick} />
@@ -349,12 +349,20 @@ export class _TopSiteList extends React.PureComponent {
     // drag and drop reordering and the underlying DOM nodes are reused.
     // This mostly (only?) affects linux so be sure to test on linux before changing.
     let holeIndex = 0;
+
+    // On narrow viewports, we only show 6 sites per row. We'll mark the rest as
+    // .hide-for-narrow to hide in CSS via @media query.
+    const maxNarrowVisibleIndex = props.TopSitesRows * 6;
+
     for (let i = 0, l = topSites.length; i < l; i++) {
       const link = topSites[i];
       const slotProps = {
         key: link ? link.url : holeIndex++,
         index: i
       };
+      if (i >= maxNarrowVisibleIndex) {
+        slotProps.className = "hide-for-narrow";
+      }
       topSitesUI.push(!link ? (
         <TopSitePlaceholder
           {...slotProps}
--- a/system-addon/content-src/components/TopSites/TopSites.jsx
+++ b/system-addon/content-src/components/TopSites/TopSites.jsx
@@ -51,7 +51,7 @@ export class _TopSites extends React.PureComponent {
    * Dispatch session statistics about the quality of TopSites icons and pinned count.
    */
   _dispatchTopSitesStats() {
-    const topSites = this._getTopSites();
+    const topSites = this._getVisibleTopSites();
     const topSitesIconsStats = countTopSitesIconsTypes(topSites);
     const topSitesPinned = topSites.filter(site => !!site.isPinned).length;
     // Dispatch telemetry event with the count of TopSites images types.
@@ -62,10 +62,16 @@ export class _TopSites extends React.PureComponent {
   }
 
   /**
-   * Return the TopSites to display based on prefs.
+   * Return the TopSites that are visible based on prefs and window width.
    */
-  _getTopSites() {
-    return this.props.TopSites.rows.slice(0, this.props.TopSitesRows * TOP_SITES_MAX_SITES_PER_ROW);
+  _getVisibleTopSites() {
+    // We hide 2 sites per row when not in the wide layout.
+    let sitesPerRow = TOP_SITES_MAX_SITES_PER_ROW;
+    // $break-point-widest = 1072px (from _variables.scss)
+    if (!global.matchMedia(`(min-width: 1072px)`).matches) {
+      sitesPerRow -= 2;
+    }
+    return this.props.TopSites.rows.slice(0, this.props.TopSitesRows * sitesPerRow);
   }
 
   componentDidUpdate() {
--- a/system-addon/content-src/components/TopSites/_TopSites.scss
+++ b/system-addon/content-src/components/TopSites/_TopSites.scss
@@ -220,6 +220,38 @@
   }
 }
 
+// Always hide .hide-for-narrow if wide layout is disabled
+.wide-layout-disabled {
+  .top-sites-list {
+    .hide-for-narrow {
+      display: none;
+    }
+  }
+}
+
+.wide-layout-enabled {
+  .top-sites-list {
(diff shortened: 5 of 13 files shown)


Mardak:
@bryanbell should the default switch to 1 row of top sites (for everybody?)

Mardak:
Looks like there's some edge case issues:
screen shot 2018-02-05 at 11 15 56 pm

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -43,6 +43,7 @@ function countTopSitesIconsTypes(topSites) {
 export class _TopSites extends React.PureComponent {
   constructor(props) {
     super(props);
+    this.windowObj = this.props.windowObj || window;
```
Mardak: I believe we should be using `global` and overriding it in tests. (for prerendering)

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -62,10 +63,13 @@ export class _TopSites extends React.PureComponent {
...
   /**
-   * Return the TopSites to display based on prefs.
+   * Return the TopSites that are visible based on prefs and window width.
    */
-  _getTopSites() {
-    return this.props.TopSites.rows.slice(0, this.props.TopSitesRows * TOP_SITES_MAX_SITES_PER_ROW);
+  _getVisibleTopSites() {
+    // We hide 2 sites per row when not in the wide layout.
+    const WIDE_LAYOUT_BREAKPOINT = 1056;
+    const sitesPerRow = this.windowObj.innerWidth < WIDE_LAYOUT_BREAKPOINT ? TOP_SITES_MAX_SITES_PER_ROW - 2 : TOP_SITES_MAX_SITES_PER_ROW;
```
Mardak: Given that we should have a live document, there should be a dynamic way to query what's shown… I'll think about it a bit more..

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -218,6 +218,54 @@
       @include context-menu-button-hover;
     }
   }
+
+  &.top-sites-1rows {
```
Mardak: I don't really like the approach of needing to declare in css ahead of time each of the `#rows` as it wouldn't allow for about:config custom rows... but maybe oh well. We probably don't want to do `max-height` stuff as `overflow: hidden` has odd cropping at edges.

Alternatively is to have javascript calculate and skip rendering some on resize. What to do for prerendered... mmmm... Maybe with react 16, we prerender with 8 and on mount, re-render without the last 2? Although that will probably have some jank with a flash of a partial row for narrow windows…

@k88hudson any suggestions?

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -218,6 +218,54 @@
...
+    @media (max-width: $break-point-widest - 1) {
+      .top-site-outer {
+        display: none;
+
+        &:nth-child(-n+6) {
```
Mardak: If we go with this approach, instead of `none` for all then "unhiding" the first 6, you could just hide not-the-first-6 with `.top-site-outer:nth-child(n+7) { display: none; }`

system-addon/content-src/components/TopSites/TopSites.jsx
rlr: I tried that but `window` is a getter-only property for globals. I can do `global.window.innerWidth = 1056;` in the tests and it works but that seems hacky (?).

system-addon/content-src/components/TopSites/_TopSites.scss
rlr: Somebody setting more than 4 rows *probably* has a wider layout? And they can probably figure out how to inject their own CSS 😉

system-addon/content-src/components/TopSites/_TopSites.scss
rlr: I'll try this but for some reason I thought that would just not show any child.

rlr: oh nice, it works. how did you trigger that edge case you screenshoted?

Mardak:
@rlr Oh oh oh! An idea just popped in! We can render each top site with an extra `<top-site class="hide-for-narrow">` for those that should only be shown in wider view. Then the css rule is just to hide those when it's not wide media. ? ! ?

rlr:
@bryanbell says we can go back to 1 row for the default.

rlr:
OK, cleaned up a some more. But I still can't repro the edge case when resizing window. You are on Mac? I guess there must be a difference between `window.innerWidth` and the `@media width`?

rlr:
oh looks like I can do something like `if (window.matchMedia('(max-width: 1055px)').matches) {}`. 💃

rlr:
hmmm. I fixed that but I don't think that's related to your issue 🤔

Mardak:
Here's a video from the latest changes:
https://ed.agadak.net/as/activity-stream.wider.webm

rlr:
@Mardak I see, it must be a fractional pixel thing on retina I am guessing? But I still can't repro. I think that last commit ^^ might fix hopefully.

Mardak:
Latest change doesn't seem to fix. The odd wrapping happens on both my retina screen and external monitor. It starts causing problems at width 1056px and fixes when it shrinks 16px more to 1040px.

If I change the media queries to be 16px more for 1072px, I don't see the problem………

Interestingly, this is not a new problem apparently, so we can do a separate bug. When the window is 785px, the `@media (min-width: 800px) {` still applies for me. But at 784px.. again 16px off, it switches to the next smaller media query.

Mardak:
screen shot 2018-02-06 at 5 11 47 pm

Mardak:
The 16px off thing might be nightly60 only. Seems okay in beta59.

Mardak:
Oh wait? 16px matches up with the scrollbar's width ? !

rlr:
that makes sense. do you change your scrollbar settings? they are different than what I think is standard/default (they don't auto hide)

Mardak:
Ha ha! ? Attaching an external monitor seems to make the scrollbars always visible… Detaching fixes the bug :p

Mardak:
Oh. It's probably more likely because I have a USB C hub that has hdmi and usb keyboard/mouse combo attached, so it might assume because I have a mouse (with potentially no mousewheel?) it should show the scrollbar.

system-addon/common/Reducers.jsm
```diff
@@ -6,8 +6,8 @@
 const {actionTypes: at} = ChromeUtils.import("resource://activity-stream/common/Actions.jsm", {});
 const {Dedupe} = ChromeUtils.import("resource://activity-stream/common/Dedupe.jsm", {});
 
-const TOP_SITES_DEFAULT_ROWS = 2;
-const TOP_SITES_MAX_SITES_PER_ROW = 6;
+const TOP_SITES_DEFAULT_ROWS = 1;
```
Mardak: After looking at the remaining use of `TOP_SITES_DEFAULT_ROWS`, it looks like it's either wrong (`HighlightsFeed`) or only used in tests, where we could just separately define. Let's clean this up separate from this PR

system-addon/content-src/components/TopSites/TopSite.jsx
```diff
@@ -349,12 +349,20 @@ export class _TopSiteList extends React.PureComponent {
     // drag and drop reordering and the underlying DOM nodes are reused.
     // This mostly (only?) affects linux so be sure to test on linux before changing.
     let holeIndex = 0;
+
+    // On narrow viewports, we only show 6 sites per row. We'll mark the rest as
+    // .hide-for-narrow to hide in CSS via @media query.
+    let maxNarrowVisibleIndex = props.TopSitesRows * 6;
```
Mardak: nit: `const`

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -218,6 +218,14 @@
       @include context-menu-button-hover;
     }
   }
+
+  .hide-for-narrow {
+    display: none;
+
+    @media (min-width: $break-point-widest) {
```
Mardak: Sorry for all the churn here. Let's do a single rule:
```css
@media not all and (min-width: $break-point-widest) {
  .hide-for-narrow {
    display: none;
  }
}
```

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -62,10 +62,16 @@ export class _TopSites extends React.PureComponent {
...
 
   /**
-   * Return the TopSites to display based on prefs.
+   * Return the TopSites that are visible based on prefs and window width.
    */
-  _getTopSites() {
-    return this.props.TopSites.rows.slice(0, this.props.TopSitesRows * TOP_SITES_MAX_SITES_PER_ROW);
+  _getVisibleTopSites() {
+    // We hide 2 sites per row when not in the wide layout.
+    const WIDE_LAYOUT_BREAKPOINT = 1072; // 1056 + 16 for scrollbar
```
Mardak: The `1056` is seemingly arbitrary, so let's just have the comment be `// $break-point-widest from _variables.scss`

Mardak: And I suppose might as well just make the `1072` directly part of the `matchMedia` string arg.

system-addon/content-src/components/Sections/_Sections.scss
```diff
@@ -20,6 +20,10 @@
         @include context-menu-open-left;
```
Mardak: Looks like we'll need a similar rule/include for widest:
screen shot 2018-02-06 at 7 51 24 pm

system-addon/content-src/styles/_variables.scss
```diff
@@ -70,10 +70,12 @@ $wrapper-default-width: $grid-unit * 2 + $base-gutter * 1; // 2 top sites
 $wrapper-max-width-small: $grid-unit * 3 + $base-gutter * 2; // 3 top sites
 $wrapper-max-width-medium: $grid-unit * 4 + $base-gutter * 3; // 4 top sites
 $wrapper-max-width-large: $grid-unit * 6 + $base-gutter * 5; // 6 top sites
-$break-point-small: $wrapper-max-width-small + $base-gutter * 2;
-$break-point-medium: $wrapper-max-width-medium + $base-gutter * 2;
-$break-point-large: $wrapper-max-width-large + $base-gutter * 2;
-$wrapper-card-outer: 768px;
+$wrapper-max-width-widest: $grid-unit * 8 + $base-gutter * 7; // 8 top sites
+$scrollbar-width: 16px;
```
Mardak: A comment about why we need this and that we're just using a single value for all os

system-addon/content-src/components/Sections/_Sections.scss
Mardak: Also for top sites too.

Mardak:
Also, I'm assuming @bryanbell wanted to keep the aspect ratio of cards the same, but now short description cards are pretty empty:
screen shot 2018-02-06 at 8 25 13 pm

system-addon/content-src/components/TopSites/TopSites.jsx
rlr: Interesting- I am seeing examples of people importing variables from .scss into js but I couldn't get it to work on first try so I'll punt on it for now.

system-addon/common/Reducers.jsm
rlr: Hmm. Looking at `HighlightsFeed` again I was wrong about Top Sites not getting deduped from Highlights. And yeah, the use there of the default rows is wrong. I think I'll ask one more time if we can remove that deduping 😄 . Seems like it should only come into play on newish profiles anyway.

rlr:
dare i ask... should this be behind a pref?

rlr:
@Mardak want to 👀 that last commit (616f160) ?

rlr:
I'll keep around a branch with a revert to that one commit and hope that it applies cleanly 20 weeks from now 😄 

I'm excited to reorganize my pinned top sites!

as-pine-proxy:
[Treeherder: 504f63c615d66fbe9ed938db97caaceea4ae5b55](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=504f63c615d66fbe9ed938db97caaceea4ae5b55)

PR #3975, opened 2018-02-07
Fix Bug 1432657 - Add pocket card type to highlight types

sarracini:
1. Allows Pocket items to be shown in highlights
2. Adds a "remove from pocket list" option in context menu (includes string + icon)
3. Adds functionality to actually remove the item from the pocket list via context menu
4. Adds a "saved to pocket" highlight  context  type (includes string + small version of regular pocket icon)
5. Preemptively updates UI when removing a pocket item via the reducer 

Has the same behaviour as when you bookmark a trending story - update the card itself then next time you refresh it shows it in the highlights section

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -33,6 +33,7 @@ type_label_visited=Visited
 type_label_bookmarked=Bookmarked
 type_label_synced=Synced from another device
 type_label_recommended=Trending
+type_label_pocket=Saved to Pocket
 # LOCALIZATION NOTE(type_label_open): Open is an adjective, as in "page is open"
 type_label_open=Open
 type_label_topic=Topic
@@ -58,6 +59,8 @@ confirm_history_delete_p1=Are you sure you want to delete every instance of this
 # page from history.
 confirm_history_delete_notice_p2=This action cannot be undone.
 menu_action_save_to_pocket=Save to Pocket
+menu_action_delete_pocket=Delete from Pocket
+menu_action_archive_pocket=Archive in Pocket
 
 # LOCALIZATION NOTE (search_for_something_with): {search_term} is a placeholder
 # for what the user has typed in the search input field, e.g. 'Search for ' +
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -265,6 +265,22 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           return item;
         })
       }));
+    case at.PLACES_SAVED_TO_POCKET:
+      if (!action.data) {
+        return prevState;
+      }
+      return prevState.map(section => Object.assign({}, section, {
+        rows: section.rows.map(item => {
+          if (item.url === action.data.url) {
+            return Object.assign({}, item, {
+              pocket_id: action.data.pocket_id,
+              title: action.data.title,
+              type: "pocket"
+            });
+          }
+          return item;
+        })
+      }));
     case at.PLACES_BOOKMARK_REMOVED:
       if (!action.data) {
         return prevState;
@@ -291,6 +307,10 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
     case at.PLACES_LINK_BLOCKED:
       return prevState.map(section =>
         Object.assign({}, section, {rows: section.rows.filter(site => site.url !== action.data.url)}));
+    case at.DELETE_FROM_POCKET:
+    case at.ARCHIVE_FROM_POCKET:
+      return prevState.map(section =>
+        Object.assign({}, section, {rows: section.rows.filter(site => site.pocket_id !== action.data.pocket_id)}));
     default:
       return prevState;
   }
--- a/system-addon/content-src/components/ContextMenu/ContextMenu.jsx
+++ b/system-addon/content-src/components/ContextMenu/ContextMenu.jsx
@@ -34,7 +34,7 @@ export class ContextMenu extends React.PureComponent {
       <ul role="menu" className="context-menu-list">
         {this.props.options.map((option, i) => (option.type === "separator" ?
           (<li key={i} className="separator" />) :
-          (<ContextMenuItem key={i} option={option} hideContext={this.hideContext} />)
+          (option.type !== "empty" && <ContextMenuItem key={i} option={option} hideContext={this.hideContext} />)
         ))}
       </ul>
     </span>);
--- a/system-addon/content-src/lib/link-menu-options.js
+++ b/system-addon/content-src/lib/link-menu-options.js
@@ -7,6 +7,7 @@ import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
  */
 export const LinkMenuOptions = {
   Separator: () => ({type: "separator"}),
+  EmptyItem: () => ({type: "empty"}),
   RemoveBookmark: site => ({
     id: "menu_action_remove_bookmark",
     icon: "bookmark-added",
@@ -48,7 +49,7 @@ export const LinkMenuOptions = {
     icon: "dismiss",
     action: ac.AlsoToMain({
       type: at.BLOCK_URL,
-      data: site.url
+      data: {url: site.url, pocket_id: site.pocket_id}
     }),
     impression: ac.ImpressionStats({
       source: eventSource,
@@ -77,7 +78,7 @@ export const LinkMenuOptions = {
       type: at.DIALOG_OPEN,
       data: {
         onConfirm: [
-          ac.AlsoToMain({type: at.DELETE_HISTORY_URL, data: {url: site.url, forceBlock: site.bookmarkGuid}}),
+          ac.AlsoToMain({type: at.DELETE_HISTORY_URL, data: {url: site.url, pocket_id: site.pocket_id, forceBlock: site.bookmarkGuid}}),
           ac.UserEvent({event: "DELETE", source: eventSource, action_position: index})
         ],
         eventSource,
@@ -121,6 +122,24 @@ export const LinkMenuOptions = {
     }),
     userEvent: "SAVE_TO_POCKET"
   }),
+  DeleteFromPocket: site => ({
+    id: "menu_action_delete_pocket",
+    icon: "delete",
+    action: ac.AlsoToMain({
+      type: at.DELETE_FROM_POCKET,
+      data: {pocket_id: site.pocket_id}
+    }),
+    userEvent: "DELETE_FROM_POCKET"
+  }),
+  ArchiveFromPocket: site => ({
+    id: "menu_action_archive_pocket",
+    icon: "check",
+    action: ac.AlsoToMain({
+      type: at.ARCHIVE_FROM_POCKET,
+      data: {pocket_id: site.pocket_id}
+    }),
+    userEvent: "ARCHIVE_FROM_POCKET"
+  }),
   EditTopSite: (site, index) => ({
     id: "edit_topsites_button_text",
     icon: "edit",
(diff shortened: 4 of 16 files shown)


locales/en-US/strings.properties
```diff
@@ -54,6 +55,7 @@ confirm_history_delete_p1=Are you sure you want to delete every instance of this
 # page from history.
 confirm_history_delete_notice_p2=This action cannot be undone.
 menu_action_save_to_pocket=Save to Pocket
+menu_action_remove_pocket=Remove from Pocket List
```
Mardak: Probably confirm with pocket(?) about this string. In particular the opposite string is just "Save to Pocket" (no "List")

locales/en-US/strings.properties
```diff
@@ -29,6 +29,7 @@ type_label_visited=Visited
 type_label_bookmarked=Bookmarked
 type_label_synced=Synced from another device
 type_label_recommended=Trending
+type_label_pocket=Saved To Pocket
```
Mardak: Probably also confirm with pocket people on their terminology.

system-addon/lib/PlacesFeed.jsm
```diff
@@ -262,6 +264,39 @@ class PlacesFeed {
...
+  async saveToPocket(action) {
+    const {url, title} = action.data.site;
+    // If the user is not logged in, show the panel to prompt them to log in
+    if (!pktApi.isUserLoggedIn()) {
+      Pocket.savePage(action._target.browser, url, title);
+      return;
+    }
+    try {
+      // If the user is logged in, just save the link to Pocket and update the page
+      let {item_id} = await this.addLinkToPocket(url, title);
```
Mardak: What is the expected behavior of saving to pocket? Before, it would open up the tagging doorhanger and trigger the library animation.

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -112,14 +111,20 @@ this.HighlightsFeed = class HighlightsFeed {
         this.fetchImage(page);
       }
 
+      // When we bookmark something it adds a history entry - so adjust the type
```
Mardak: This comment is a bit misleading. The reason why the type switches to "bookmark" is that a highlight that comes in as history/visited might happen to be bookmarked, so show it as bookmarked even though the source was from history.

system-addon/lib/PlacesFeed.jsm
Mardak: Note that the bookmark behavior is to open the bookmark doorhanger that triggers the library animation.

system-addon/common/Reducers.jsm
```diff
@@ -289,6 +305,7 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
       return prevState.map(section => Object.assign({}, section,
         {rows: section.rows.filter(site => !action.data.includes(site.url))}));
     case at.PLACES_LINK_BLOCKED:
+    case at.REMOVE_FROM_POCKET:
       return prevState.map(section =>
         Object.assign({}, section, {rows: section.rows.filter(site => site.url !== action.data.url)}));
```
Mardak: Hmm.. Removing a specific pocket item isn't the same as blocking a url.. where this code would hide any other bookmarks of the same url when unpocketing.

system-addon/content-src/lib/link-menu-options.js
```diff
@@ -120,6 +120,14 @@ export const LinkMenuOptions = {
     }),
     userEvent: "SAVE_TO_POCKET"
   }),
+  RemoveFromPocket: (site, eventSource) => ({
+    id: "menu_action_remove_pocket",
+    icon: "remove-pocket",
+    action: ac.AlsoToMain({
+      type: at.REMOVE_FROM_POCKET,
+      data: {pocket_id: site.pocket_id, url: site.url}
```
Mardak: If we make the reducer remove by `pocket_id` then we won't need `url` here anymore.

system-addon/lib/PlacesFeed.jsm
sarracini: While we wait for the ui wanted for the correct behaviour, I'll explain my reasoning behind this change..
calling ```Pocket.savePage``` although does drop the doorhanger down doesn't have any indication of a successful or not successful save once it goes off and does it's stuff. With the pktApi we get the pocket_id back when we're successful so we can update the card we just saved, similar to how we do that for bookmark (i.e changing the context menu + highlight context) So I traded off one kind of confirmation feedback for a different one. Ideally we would call ```addLinkToPocket```, and then also call something in Pocket.jsm to *just* show the doorhanger + library animation.

system-addon/lib/HighlightsFeed.jsm
sarracini: These were my steps to reproduce for this that caused me to write this comment (on a clean profile):
1. Bookmark a trending pocket story from the context menu
2. Open debugger and set a breakpoint in HighlightsFeed.jsm right after we dedupe against top sites
3. Force a refresh in Higlights (I dismissed another trending story)
4. Observe ```this.store.getState().TopSites.rows``` and see that the first item in the array is the bookmarked trending story with type 'history' ... even though I've never actually visited it.

So it's a bit confusing that it's there even though I never actually visited the page. How about "Adjust the type for 'history' items that are also 'bookmarked'" for the comment?

piatra: https://github.com/mozilla/activity-stream/pull/3734

sarracini:
Just adding a note here so i don't forget:
Change the update to 1 hour instead of 5 minutes to not overload pocket servers

locales/en-US/strings.properties
sarracini: Designers said "Delete on Pocket" and "Archive on Pocket"

system-addon/lib/PlacesFeed.jsm
sarracini: The designers say don't bother with the doorhanger :D

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -86,8 +89,7 @@ this.HighlightsFeed = class HighlightsFeed {
 
     // Request more than the expected length to allow for items being removed by
     // deduping against Top Sites or multiple history from the same domain, etc.
-    // Until bug 1425496 lands, do not include saved Pocket items in highlights
-    const manyPages = await this.linksCache.request({numItems: MANY_EXTRA_LENGTH, excludePocket: true});
+    const manyPages = await this.linksCache.request({numItems: MANY_EXTRA_LENGTH, excludePocket: options.excludePocket});
```
Mardak: Mmm… Doesn't this mean we only include pocket items for the one refresh out of the ~12 each hour? So most highlight updates result in no pocket items shown?

sarracini: Other way around, no? We want to update pocket on every refresh except the system_tick unless it's been over an hour... so if options.excludePocket is null for all refreshes, then !excludePocket should be true and then we'd go fetch them in NewTabUtils

sarracini:
blocked on bug 1439684: https://bugzilla.mozilla.org/show_bug.cgi?id=1439684

Mardak:
Given the Pocket team's desire to have sync instead of polling, we probably should have a pref that toggles `excludePocket`. We'll want it eventually anyway when we have Highlights sub option.

system-addon/lib/ActivityStream.jsm
```diff
@@ -130,6 +130,10 @@ const PREFS_CONFIG = new Map([
     title: "Boolean flag that decides whether or not to show the topstories disclaimer.",
     value: true
   }],
+  ["section.highlights.excludePocket", {
```
sarracini: Not sure about what namespace we want for the "sub-option" prefs

Mardak: I guess it matches "section.highlights.collapsed" although we also have "browser.newtabpage.activity-stream.collapseTopSites"…

Let's keep it next to the other highlights pref. And perhaps rename it to "includePocket" as that's what the UI would expose.

sarracini:
New design specs say pocket items should not be able to delete from history
image

sarracini:
Blocked on: https://bugzilla.mozilla.org/show_bug.cgi?id=1440389

system-addon/lib/SectionsManager.jsm
```diff
@@ -66,7 +66,7 @@ const BUILT_IN_SECTIONS = {
     icon: "highlights",
     title: {id: "header_highlights"},
     maxRows: 3,
-    availableContextMenuOptions: ["CheckBookmark", "SaveToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl", "DeleteUrl"],
+    availableContextMenuOptions: ["CheckBookmarkOrArchive", "CheckSavedToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl", "CheckDeleteHistoryOrEmpty"],
```
Mardak: We want these "Check" versions for `topstories.availableContextMenuOptions` too for the in-page update. (Fortunately, topstories already doesn't have "DeleteUrl" :p)

system-addon/common/Reducers.jsm
```diff
@@ -265,6 +265,22 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           return item;
         })
       }));
+    case at.PLACES_SAVED_TO_POCKET:
+      if (!action.data) {
+        return prevState;
+      }
+      return prevState.map(section => Object.assign({}, section, {
+        rows: section.rows.map(item => {
+          if (item.url === action.data.url) {
```
Mardak: There's a slight bug / accidental(?) fix here in that if you pocket a story, stories' `rows` gets deduped via `SECTION_UPDATE`. A similar thing happens when bookmarking that @piatra and @csadilek worked on.

The bug is that even when Highlights no longer has the deduping url, e.g., archived pocket or unbookmarked page+delete history, `rows` is still deduped until it gets refreshed with an update or restart.

Simplest steps to reproduce:
1) save to pocket a story
2) archive that card
3) refresh page see that the card is still gone
4) toggle stories section from preferences, and see the story come back

Probably not really worth a fix as the user wanted it gone anyway and would most likely only run into it on restarting firefox, and by then, hopefully new stories would have come in…

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -34,7 +34,7 @@ export class ContextMenu extends React.PureComponent {
       <ul role="menu" className="context-menu-list">
         {this.props.options.map((option, i) => (option.type === "separator" ?
           (<li key={i} className="separator" />) :
-          (<ContextMenuItem key={i} option={option} hideContext={this.hideContext} />)
+          (option.type !== "empty" && <ContextMenuItem key={i} option={option} hideContext={this.hideContext} />)
```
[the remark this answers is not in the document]
sarracini: I'm assuming some of this is going to change anyways with weighted highlights, since we'll have to have a way to assign context menu options on a per card basis and we won't need this empty stuff anymore

system-addon/common/Reducers.jsm
sarracini: Ah I see, yeah that's unfortunate but I think you're right, not super important to fix. Unless we added the saved to pocket trending card to the block list when they archive or save it or something. But probably not worth fixing right now

as-pine-proxy:
[Treeherder: 676200a7f29b2ccf29583c226d0edd1f1f49bd01](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=676200a7f29b2ccf29583c226d0edd1f1f49bd01)

PR #3976, opened 2018-02-08
chore (polish): change drag and drop target to a grey box

rlr:
@aaronrbenson requested and approved these changes.

Before:

After:

### DIFF
--- a/system-addon/content-src/components/TopSites/_TopSites.scss
+++ b/system-addon/content-src/components/TopSites/_TopSites.scss
@@ -207,6 +207,21 @@
         display: none;
       }
     }
+
+    &.dragged {
+      .tile {
+        background: $grey-20;
+        box-shadow: none;
+
+        * {
+          display: none;
+        }
+      }
+
+      .title {
+        visibility: hidden;
+      }
+    }
   }
 
   &:not(.dnd-active) {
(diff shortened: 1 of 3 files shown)


Mardak:
This probably warrants some extra testing on other OSes. It looks like this approach changes the DOM nodes being dragged, but at least on mac, it happens to keep the original visual state of thing being dragged.

rlr:
good point. i'll fire up linux. and travis 🤔

rlr:
After pulling my hair out some, I managed to get it tested on linux. Do you have a Windows dev machine or should I bite that bullet now? I'll fix the tests too.

system-addon/content-src/components/TopSites/_TopSites.scss
```diff
@@ -207,6 +207,24 @@
...
+    &.dragged {
+      .tile {
+        background: $grey-20;
+        box-shadow: none;
+
+        &::before,
+        .screenshot,
+        .rich-icon,
+        .top-site-icon {
+          display: none;
```
Mardak: Not sure if it would be better to have the component just render fewer items, e.g., just a `.tile` with no contents? Seems like this approach could easily result in structure render changes to accidentally appear when dragged. Or would a `*` descendant match be too much? Hmm...

PR #3978, opened 2018-02-09
chore(mc): Port Bug 1427674 - Unify FxA content server URL preferences. r=markh,tcsc

Mardak:
`git -C ../../mozilla-central sh 089fb03cc734 browser/extensions/activity-stream | patch -p4`
https://hg.mozilla.org/mozilla-central/rev/b014201f939f


Mardak:
Durr. I was thinking there was something like this already. I checked the PRs list but it didn't jump out for some reason :p

Mardak:
Merged the other instead. Thanks!

PR #3979, opened 2018-02-09
chore (prefs): change default # of rows back to 1 for Firefox 59

rlr:
oh hey. that was easy! (TM)


Mardak:
We need to update the prerendered values, yeah?

rlr:
gah! good call. fixed ^

rlr:
looks like I didnt change that either for switching back to 1 on the wide layout patch. PR coming

PR #3986, opened 2018-02-18
Fix Bug 1438565 - Send source and action_position with DELETE ping

piatra:
I initially thought this was a lot worse but it looks like we "didn't lose" all the clicks.

`DeleteUrl` has a userEvent of [`DIALOG_OPEN`](https://github.com/mozilla/activity-stream/blob/a050adee0eebcc47c310ed9d9fb9973f1a07e1bd/system-addon/content-src/lib/link-menu-options.js#L89) defined which is [handled by the `LinkMenu`](https://github.com/mozilla/activity-stream/blob/a050adee0eebcc47c310ed9d9fb9973f1a07e1bd/system-addon/content-src/components/LinkMenu/LinkMenu.jsx#L23-L28).

If you sum up the DIALOG_OPEN events from Highlights and from Topsites they roughly add up to the DELETE count with n/a source (+/- some CANCEL events I suppose).

update: The bug mentioned the DELETE event but looks like DIALOG_CANCEL also has no source

### DIFF
--- a/system-addon/content-src/components/ConfirmDialog/ConfirmDialog.jsx
+++ b/system-addon/content-src/components/ConfirmDialog/ConfirmDialog.jsx
@@ -31,7 +31,7 @@ export class _ConfirmDialog extends React.PureComponent {
 
   _handleCancelBtn() {
     this.props.dispatch({type: actionTypes.DIALOG_CANCEL});
-    this.props.dispatch(ac.UserEvent({event: actionTypes.DIALOG_CANCEL}));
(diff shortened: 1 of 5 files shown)


as-pine-proxy:
[Treeherder: 91ef5c1bbc9dc8228e1c2a0f4baa28d38d25c38a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=91ef5c1bbc9dc8228e1c2a0f4baa28d38d25c38a)

PR #3987, opened 2018-02-19
Fix Bug 1422079 - sections context menu

rlr:
Most of the +green+ is offset by -red- due to the removal of the infoOption `<Info/>` component stuff. yay

* Remove `<Info/>` component and related tests, styles, etc.
* Implement new `<SectionMenu/>` component which is a `ContextMenu` specific for sections.
** It is missing the "Move Up" and "Move Down" options until we implement reordering
** Some icons are still being manufactured @bryanbell 

Screenshots:
screen shot 2018-02-19 at 5 56 34 pm

screen shot 2018-02-19 at 5 56 42 pm

screen shot 2018-02-19 at 5 56 49 pm


rlr:
@csadilek want to take a look at this PR too? specifically the stuff Section related changes.

Mardak:
@aaronrbenson @bryanbell Should all the menu options be actions, e.g., "Learn More" vs "About Pocket" and "View Privacy Notice" vs "Privacy Notice"?

Also, we don't call the top stories section "Pocket" right now… It's "Recommended by Pocket", so should the menu actions be "Manage Recommended by Pocket"?

aaronrbenson:
Much of this is about brevity. See the attached example. I don't think the "About Pocket" link needs to necessarily be an action.

screen shot 2018-02-20 at 5 42 08 pm

system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -135,12 +51,10 @@ export class _CollapsibleSection extends React.PureComponent {
   constructor(props) {
     super(props);
     this.onBodyMount = this.onBodyMount.bind(this);
-    this.onInfoEnter = this.onInfoEnter.bind(this);
-    this.onInfoLeave = this.onInfoLeave.bind(this);
```
Mardak: Were these not actually used by `CollapsibleSection`? Looks like it was added twice but only needed in `Info`?

rlr: oh weird I didnt even notice that! I guess it was some copy/pasta fail when creating `<Disclaimer/>`

rlr: oh, I guess this isn't `<Disclaimer/>`. I fail at reading diffs. Yeah, this must've been a fail when I broke `<Info/>` into it's own component or something.

system-addon/content-src/lib/section-menu-options.js
```diff
@@ -0,0 +1,46 @@
...
+  AddTopSite: section => ({
+    id: "menu_action_add_topsite",
+    icon: "add",
+    action: {type: at.TOP_SITES_EDIT, data: {index: -1}},
+    userEvent: "SECTION_MENU_ADD_TOPSITE"
+  }),
+  AboutPocket: section => ({
+    id: "menu_action_about_pocket",
+    icon: "about",
+    redirectUrl: "https://getpocket.cdn.mozilla.net/firefox/new_tab_learn_more",
```
Mardak: We should pass in the `disclaimer_link` as part of the section to use here.

system-addon/lib/SectionsManager.jsm
```diff
@@ -41,7 +41,8 @@ const BUILT_IN_SECTIONS = {
       button: {id: options.disclaimer_buttontext || "section_disclaimer_topstories_buttontext"}
     },
     maxRows: 1,
-    availableContextMenuOptions: ["CheckBookmark", "SaveToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl"],
+    sectionMenuOptions: ["RemoveSection", "CheckCollapsed", "Separator", "AboutPocket", "ManageSection"],
```
Mardak: I wonder if we want to have a common list where "About" gets added in dynamically?

locales/en-US/strings.properties
```diff
@@ -22,6 +22,11 @@ header_stories_from=from
 # the site.
 context_menu_button_sr=Open context menu for {title}
 
+# LOCALIZATION NOTE(section_context_menu_button_sr): This is for screen readers when
+# the section edit context menu button is focused/active. Title is the name of
+# the section (Top Sites, Highlights, etc).
+section_context_menu_button_sr=Open context menu for {title} section
```
flodolo: This might not work with all languages, because you're reusing a stand-alone string (title) in the middle of another one. Not sure if there's a better way to do it, besides removing the `{title}` completely.

locales/en-US/strings.properties
```diff
@@ -193,3 +198,15 @@ error_fallback_default_info=Oops, something went wrong loading this content.
 # that failed to render, and suggests the likely way to deal with the failure:
 # reloading the page.
 error_fallback_default_refresh_suggestion=Refresh page to try again.
+
+# LOCALIZATION NOTE (menu_action_*).  These strings are displayed in the section
+# context menu and are meant as a call to action for the given section.
```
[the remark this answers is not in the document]
flodolo: `section_menu_*` sounds good

locales/en-US/strings.properties
```diff
@@ -193,3 +198,15 @@ error_fallback_default_info=Oops, something went wrong loading this content.
...
+menu_action_remove_section=Remove Section
+menu_action_collapse_section=Collapse Section
+menu_action_expand_section=Expand Section
+menu_action_manage_section=Manage {section}
```
flodolo: Same note  about `{section}`, you'd be using a string (name of section) in the middle of a sentence, with potential problems (case, grammatical case and declension). Does it need to be there?

locales/en-US/strings.properties
```diff
@@ -193,3 +198,15 @@ error_fallback_default_info=Oops, something went wrong loading this content.
...
+menu_action_add_topsite=Add Top Site
+menu_action_about_pocket=About Pocket
+menu_action_move_up=Move Up
+menu_action_move_down=Move Down
+menu_action_privacy_notice=Privacy Notice
```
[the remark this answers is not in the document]
flodolo: Since you plan to remove the other one, it wouldn't hurt to have this string added.

locales/en-US/strings.properties
```diff
@@ -193,3 +198,15 @@ error_fallback_default_info=Oops, something went wrong loading this content.
...
 error_fallback_default_refresh_suggestion=Refresh page to try again.
+
+# LOCALIZATION NOTE (menu_action_*).  These strings are displayed in the section
+# context menu and are meant as a call to action for the given section.
+menu_action_remove_section=Remove Section
+menu_action_collapse_section=Collapse Section
+menu_action_expand_section=Expand Section
+menu_action_manage_section=Manage {section}
+menu_action_add_topsite=Add Top Site
+menu_action_about_pocket=About Pocket
```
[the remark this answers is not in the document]
csadilek: @mardak yes, agreed on keeping this generic, if possible. menu_action_about_section = About {section}.

system-addon/content-src/lib/section-menu-options.js
csadilek: Could we change this to AboutSection instead? Section could hold an object that further describes this i.e. section.about? SectionManager can provide this about object based on topstories.options.

system-addon/lib/SectionsManager.jsm
```diff
@@ -41,12 +41,8 @@ const BUILT_IN_SECTIONS = {
       button: {id: options.disclaimer_buttontext || "section_disclaimer_topstories_buttontext"}
     },
     maxRows: 1,
-    availableContextMenuOptions: ["CheckBookmark", "SaveToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl"],
-    infoOption: {
-      header: {id: options.provider_header || "pocket_feedback_header"},
-      body: {id: options.provider_description || "pocket_feedback_body"},
-      link: {href: options.info_link, id: "section_info_privacy_notice"}
-    },
```
csadilek: Here's where I thought we could configure the "about object" based on topstories.options, which we could then access in section-menu-options.js.

system-addon/content-src/lib/section-menu-options.js
Mardak: This just in from the design meeting. This should be "Privacy Notice" and link to https://www.mozilla.org/en-US/privacy/firefox/#suggest-relevant-content which is `info_link` but needs the `#anchor` updated.

locales/en-US/strings.properties
Mardak: Design feedback was to just switch to "Manage Section"

locales/en-US/strings.properties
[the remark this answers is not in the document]
Mardak: Design feedback is that this is specifically the privacy notice link, so just make it "Privacy Notice" and point to the appropriate url from the option.

locales/en-US/strings.properties
rlr: I like 👍 😄

system-addon/lib/SectionsManager.jsm
rlr: I guess we could have a `"CheckAbout"`. I'll try that.

system-addon/content-src/lib/section-menu-options.js
rlr: cool. So we'll have a `PrivacyNotice` option and no `About*` section and just conditionally show it based on if it has the url in the section options.

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -30,6 +26,10 @@ export class Section extends React.PureComponent {
     }
   }
 
+  getFormattedMessage(message) {
+    return typeof message === "string" ? <span>{message}</span> : this.props.intl.formatMessage({id: message.id}, message.values);
```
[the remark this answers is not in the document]
rlr: I can undo these changes now that we will just have the link be "Manage Section".

rlr:
I think I addressed all the feedback so far. Icons still pending...

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+const DEFAULT_SECTION_MENU_OPTIONS = ["RemoveSection", "CheckCollapsed", "Separator", "ManageSection"];
+
+export class _SectionMenu extends React.PureComponent {
+  getOptions() {
+    const {props} = this;
+
+    const propOptions = props.options || DEFAULT_SECTION_MENU_OPTIONS;
+
+    const options = propOptions.map(o => SectionMenuOptions[o](props)).map(option => {
```
Mardak: This would seem like it could be shared with `LinkMenu.jsx`'s functionality as higher order function probably from a shared `content-src/lib/menu-options`, e.g., `makeOptions(propOptions, SectionMenuOptions)`? @sarracini 

Although now I realize React is probably rerendering this component like crazy because there's all these dynamic function bindings for `onClick` even though it probably could just be defined once and refer to `this` to get at `options`?

I guess all of this HOF and more cleanup should be a separate bug. Would be interesting to see if this is actually has performance impacts given how many context menus we have now… ~25 menus * ~10 options -> 250 `ContextMenuItem` being rendered every time a prop updates?

system-addon/lib/SectionsManager.jsm
```diff
@@ -40,13 +40,10 @@ const BUILT_IN_SECTIONS = {
       },
       button: {id: options.disclaimer_buttontext || "section_disclaimer_topstories_buttontext"}
     },
+    privacyNoticeURL: "https://www.mozilla.org/en-US/privacy/firefox/#suggest-relevant-content",
     maxRows: 1,
-    availableContextMenuOptions: ["CheckBookmark", "SaveToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl"],
-    infoOption: {
-      header: {id: options.provider_header || "pocket_feedback_header"},
-      body: {id: options.provider_description || "pocket_feedback_body"},
```
Mardak: These strings are now unused I believe (so add to the commit comment). Although _body is already missing… ???

Mardak:
Oh, I thought I commented about this earlier but maybe I forgot… Do we need to support the "open to the right" vs "open to the left when there's not enough space on the right"?

system-addon/lib/SectionsManager.jsm
```diff
@@ -40,13 +40,10 @@ const BUILT_IN_SECTIONS = {
       },
       button: {id: options.disclaimer_buttontext || "section_disclaimer_topstories_buttontext"}
     },
+    privacyNoticeURL: "https://www.mozilla.org/en-US/privacy/firefox/#suggest-relevant-content",
```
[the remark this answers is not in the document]
csadilek: @Mardak @rlr Yes, let's remove info_link from topstories.options and add privacy_link instead to make this configurable. We can also delete provider_header (from options) and the corresponding string pocket_feedback_header now that we're no longer using it.

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
sarracini: Yeah this little chunk here could be refactored to a more general "menu options" instead of a link menu and a sections menu. But I think you're right @Mardak, we should evaluate the performance and do it as a follow up

system-addon/lib/SectionsManager.jsm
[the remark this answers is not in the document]
csadilek: If you prefer not having to configure this link we can add: 
`options.privacy_link || "https://www.mozilla.org/en-US/privacy/firefox/#suggest-relevant-content"`

This way we can override in case we ever need to, but it will work by default.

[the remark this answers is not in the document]
rlr: makes sense. will fix. thanks!

system-addon/lib/SectionsManager.jsm
rlr: I think I saw them used in the Preferences panel. I’ll double check.

rlr: I think "pocket_feedback_body" was a bug there. Probably was meant to be "pocket_description"? We use that in preferences. I'll fix it in `SectionManager.jsx`. But the pocket_feedback_header can die.

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
rlr: One thing we can easily do is instead of having:

```
  render() {
    return (<ContextMenu
      visible={this.props.visible}
      onUpdate={this.props.onUpdate}
      options={this.getOptions()} />);
  }
```

we should have:

```
  render() {
    return this.props.visible && (<ContextMenu
      visible={this.props.visible}
      onUpdate={this.props.onUpdate}
      options={this.getOptions()} />);
  }
```

That we we render on demand. Should save all the cpu cycles and kittens

rlr: I think the "react way" is something like this anyway:

```
{this.props.visible &&
  <Component />
}
```

rlr: Yeah. I played with it but it requires changes in `<ContextMenu/>` because it currently assumes that it will be mounted initially with `visible=false`. But should be easyish fix. Just need to do `LinkMenu` and `SectionMenu` changes together, so I'd rather do as followup instead of piling on unrelated `LinkMenu` changes here.

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
```diff
@@ -0,0 +1,53 @@
...
+import React from "react";
+import {SectionMenuOptions} from "content-src/lib/section-menu-options";
+
+const DEFAULT_SECTION_MENU_OPTIONS = ["RemoveSection", "CheckCollapsed", "Separator", "ManageSection"];
+
+export class _SectionMenu extends React.PureComponent {
+  getOptions() {
+    const {props} = this;
+
+    const propOptions = props.options || DEFAULT_SECTION_MENU_OPTIONS;
```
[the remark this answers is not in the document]
rlr: I could go either way with this one. 🤔 👍 👎 ? coin toss?

rlr:
@Mardak Everything should be addressed except for https://github.com/mozilla/activity-stream/pull/3987#discussion_r169874509 and the unnecessary rendering of all the context menus on every page load. Icons are coming. I might start hacking on the perf/rendering issue and keep the changes separate.

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
rlr: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1440456

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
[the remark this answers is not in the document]
Mardak: I think the main thing is that `DEFAULT_SECTION_MENU_OPTIONS` is basically duplicated in 5 places right now, so avoiding bugs when accidentally changing a subset of them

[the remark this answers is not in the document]
rlr: That's a good reason 😄

rlr:
@Mardak done ^^. One think I am not excited about is a bunch of props are passed to `<CollapsibleSection/>` just to be passed along to `<SectionMenu/>`. Makes me wonder if there is a better way to factor it so that they are more like siblings instead. On the other hand, it's nice that we only create SectionMenu from that one component instead of from each of the parent components. 6 one way, half a dozen the other.

Mardak:
I believe the way to avoid pass-through props is having `getChildContext()`, although we haven't really used it ourselves so far… @k88hudson might know more about when to use or avoid `context`.

system-addon/content-src/lib/section-menu-options.js
```diff
@@ -0,0 +1,46 @@
...
+  AddTopSite: section => ({
+    id: "section_menu_action_add_topsite",
+    icon: "add",
+    action: {type: at.TOP_SITES_EDIT, data: {index: -1}},
+    userEvent: "SECTION_MENU_ADD_TOPSITE"
+  }),
+  PrivacyNotice: section => ({
+    id: "section_menu_action_privacy_notice",
+    icon: "privacy",
+    action: {type: at.OPEN_LINK, data: {url: section.privacyNoticeURL}},
```
Mardak: This wants an `OnlyToMain` (right now it sends only to itself/content, so it does nothing.)

system-addon/content-src/components/SectionMenu/SectionMenu.jsx
```diff
@@ -0,0 +1,58 @@
...
+    if (props.privacyNoticeURL) {
+      propOptions.splice(-1, 0, "PrivacyNotice");
+    }
+
+    const options = propOptions.map(o => SectionMenuOptions[o](props)).map(option => {
+      const {action, id, type, userEvent} = option;
+      if (!type && id) {
+        option.label = props.intl.formatMessage({id});
+        option.onClick = () => {
+          if (action) {
```
Mardak: All options have actions, so this check isn't necessary.

system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -228,25 +124,64 @@ export class _CollapsibleSection extends React.PureComponent {
...
+              onClick={this.onMenuButtonClick}
+              onMouseEnter={this.onMenuButtonMouseEnter}
+              onMouseLeave={this.onMenuButtonMouseLeave}>
+              <span className="sr-only">
+                <FormattedMessage id="section_context_menu_button_sr" />
+              </span>
+            </button>
+            <SectionMenu
+              options={menuOptions}
+              name={name}
```
Mardak: `name` is unused here too.

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -101,24 +91,22 @@ export class _TopSites extends React.PureComponent {
...
-    };
     const {editForm} = props.TopSites;
 
     return (<ComponentPerfTimer id="topsites" initialized={props.TopSites.initialized} dispatch={props.dispatch}>
-      <CollapsibleSection className="top-sites" icon="topsites" title={<FormattedMessage id="header_top_sites" />} infoOption={infoOption} prefName="collapseTopSites" Prefs={props.Prefs} dispatch={props.dispatch}>
+      <CollapsibleSection
+        className="top-sites"
+        icon="topsites"
+        title={props.intl.formatMessage({id: "header_top_sites"})}
+        name={props.intl.formatMessage({id: "header_top_sites"})}
```
Mardak: `name` is unused here too.

system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
Mardak: Oh odd. This is indeed outdated. I wonder what view github was showing me earlier…

Mardak:
Oh, and this adds to the `UTEventReporting` errors:
`Unknown event: ["activity_stream", "event", "SECTION_MENU_MANAGE"]`

https://bugzilla.mozilla.org/show_bug.cgi?id=1438239

system-addon/content-src/lib/section-menu-options.js
rlr: oops. guess I didn't test that 🤦‍♂️

rlr:
I squashed into two commits. Just waiting on the icons now (*cough* @bryanbell *cough*)

Mardak:
We could probably land without icons for now to get strings exported sooner

bryanbell:
[Icons needed.zip](https://github.com/mozilla/activity-stream/files/1750162/Icons.needed.zip)

Here are the icons needed for this menu.

Mardak:
Same icon for collapse and expand?

Mardak:
There's an existing `glyph-info-16.svg` but it has a thinner border… ? https://searchfox.org/mozilla-central/source/browser/extensions/activity-stream/data/content/assets/glyph-info-16.svg

rlr:
YAY! @bryanbell gave me an expand icon too. Also confirmed that we should use the new -information icon. 

screen shot 2018-02-22 at 8 54 43 pm

screen shot 2018-02-22 at 8 54 54 pm

screen shot 2018-02-22 at 8 55 00 pm

system-addon/data/content/assets/glyph-information-16.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><path d="M8 16a8 8 0 1 1 8-8 8.009 8.009 0 0 1-8 8zM8 2a6 6 0 1 0 6 6 6.006 6.006 0 0 0-6-6z"/><path d="M8 7a1 1 0 0 0-1 1v3a1 1 0 0 0 2 0V8a1 1 0 0 0-1-1z"/><circle cx="8" cy="5" r="1.188"/></svg>
```
Mardak: If the old one isn't used, let's just replace `glyph-info-16.svg`

system-addon/data/content/assets/glyph-info-option-12.svg
```diff
@@ -1 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 12"><path fill="context-fill" d="M6 0a6 6 0 1 0 6 6 6 6 0 0 0-6-6zm.7 10.26a1.13 1.13 0 0 1-.78.28 1.13 1.13 0 0 1-.78-.28 1 1 0 0 1 0-1.42 1.13 1.13 0 0 1 .78-.28 1.13 1.13 0 0 1 .78.28 1 1 0 0 1 0 1.42zM8.55 5a3 3 0 0 1-.62.81l-.67.63a1.58 1.58 0 0 0-.4.57 2.24 2.24 0 0 0-.12.74H5.06a3.82 3.82 0 0 1 .19-1.35 2.11 2.11 0 0 1 .63-.86 4.17 4.17 0 0 0 .66-.67 1.09 1.09 0 0 0 .23-.67.73.73 0 0 0-.77-.86.71.71 0 0 0-.57.26 1.1 1.1 0 0 0-.23.7h-2A2.36 2.36 0 0 1 4 2.47a2.94 2.94 0 0 1 2-.65 3.06 3.06 0 0 1 2 .6 2.12 2.12 0 0 1 .72 1.72 2 2 0 0 1-.17.86z"/></svg>
```
Mardak: Looks like all the existing icons have `viewBox` instead of `width/height`:
`svgo --multipass --disable=removeViewBox --enable={removeDimensions,removeStyleElement,sortAttrs} *.svg`

system-addon/data/content/assets/glyph-info-16.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path fill="context-fill" d="M8 1a7 7 0 1 0 7 7 7.008 7.008 0 0 0-7-7zm0 13a6 6 0 1 1 6-6 6.007 6.007 0 0 1-6 6zm0-7a1 1 0 0 0-1 1v3a1 1 0 1 0 2 0V8a1 1 0 0 0-1-1zm0-3.188A1.188 1.188 0 1 0 9.188 5 1.188 1.188 0 0 0 8 3.812z"/></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path d="M8 16a8 8 0 1 1 8-8 8.009 8.009 0 0 1-8 8zM8 2a6 6 0 1 0 6 6 6.006 6.006 0 0 0-6-6z"/><path d="M8 7a1 1 0 0 0-1 1v3a1 1 0 0 0 2 0V8a1 1 0 0 0-1-1z"/><circle cx="8" cy="5" r="1.188"/></svg>
```
Mardak: Oh I forgot to mention this, but it's easier to tell now via this diff. It's missing fill="context-fill"

rlr: ah, and only that one.

Mardak: `fill: red !important` ;)
image

Mardak: I suppose this would have missed the white fill too, but that's harder to glance quickly ;)

image

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -101,24 +91,21 @@ export class _TopSites extends React.PureComponent {
...
+        showPrefName="showTopSites"
+        eventSource={TOP_SITES_SOURCE}
+        Prefs={props.Prefs}
+        dispatch={props.dispatch}>
         <TopSiteList TopSites={props.TopSites} TopSitesRows={props.TopSitesRows} dispatch={props.dispatch} intl={props.intl} />
         <div className="edit-topsites-wrapper">
-          <div className="add-topsites-button">
-            <button
-              className="add"
-              title={this.props.intl.formatMessage({id: "edit_topsites_add_button_tooltip"})}
```
Mardak: I guess that was short lived. string already unused!

Mardak: Actually, this was only added in this 60 cycle? It should be safe to just remove now as we don't need to keep it longer for beta

rlr: oh right. that was at the work week which was 60 already. nice

rlr: and I guess "edit_topsites_add_button" can be removed in a few releases too

system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
```diff
@@ -15,150 +18,53 @@
...
-      margin: 0;
-      margin-top: 12px;
-    }
-
-    .info-option-link {
-      color: $link-primary;
-      margin-left: 7px;
-    }
+  @media (max-width: $break-point-widest + $card-width * 1.5) {
+    @include context-menu-open-left;
```
rlr: this isn't specific enough and is causing link menus to open to the left too.

dmose:
[The context API is changing in some upcoming version of React](https://medium.com/dailyjs/reacts-%EF%B8%8F-new-context-api-70c9fe01596b).   It might make sense to wait for this to solve the "prop-drilling" problem.

PR #3988, opened 2018-02-21
Fix Bug 1438523 - add enableWideLayout to prerender prefs

rlr:
The "wireframe" term threw me off but I eventually figured it out.

### DIFF
--- a/system-addon/common/PrerenderData.jsm
+++ b/system-addon/common/PrerenderData.jsm
@@ -56,7 +56,8 @@ this.PrerenderData = new _PrerenderData({
     "section.highlights.collapsed": false,
     "section.topstories.collapsed": false,
     "feeds.section.topstories": true,
-    "feeds.section.highlights": true
+    "feeds.section.highlights": true,
+    "enableWideLayout": true
   },
   // Prefs listed as invalidating will prevent the prerendered version
   // of AS from being used if their value is something other than what is listed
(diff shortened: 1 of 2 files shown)


rlr:
Hmm, I thought the issue was that we flash 6 topsites then change it to 8. Also the search bar starts at one size then grows. `.section-empty-state` only applies to sections?

rlr:
Yeah, I dont think `.section-empty-state` has anything to do with the screenshot on the bug:

rlr:
(ie, the issue, as I see it, is that we load the page with the `div.wide-layout-disabled` and then remove the class once react renders`

Mardak:
That video also shows the smaller placeholder highlights outline becoming big.

rlr:
Is that ^^ what you are referring to?

rlr:
fancy screenshots! how you get those magical powers? ha

Mardak:
For the screenshot, I ended up doing screen recording to get the exact frame then some image editing (ended up using ScreenFlow and GravitDesigner, but any recording and editing should do)

as-pine-proxy:
[Treeherder: 2cf0e0f5a4f57f8b4992829118f66caec8b6072c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2cf0e0f5a4f57f8b4992829118f66caec8b6072c)

PR #3989, opened 2018-02-22
Fix Bug 1425457 - Include some extra info to user event ping for top sites

piatra:


ncloudioj:
We'd like to generalize this in order to cover other interests on event pings. For example, a click on a pinned top site, or a click on a bookmarked highlight, etc.

I'd recommend to add a "value" field to the current event ping to do this. In this case, it just adds a `value = "PINNED"` to the payload. Also, we can use a composite to record multiple things `value = "PINNED,TIPPYTOP"`.

@piatra what do you think?

ncloudioj:
Let's also update the events.md and data_dictionary.md for those changes.

piatra:
@ncloudioj the values we could possibly send out in the value field are: `icon_type`, `is_pinned` and `is_bookmark` do we want to send this as a JSON string?

ncloudioj:
Despite that we usually prefer to use a "flat" schema in the payload schema, it's a sound idea to use JSON for the "value" field here, since apparently a comma separate string won't be easy to extend and query in this case.

We can serialized the following object and set it as the `value`:
 ```json
{
  "icon_type": "[tippytop|rich_icon|...]",
  "card_type": "[history, bookmark, trending]",
  "is_pinned": [true|false]
}
```

Again, let's make sure all the fields are properly documented in the telemetry documents.

**Comment by ncloudioj** [comment]
Actually, let's stick with the uppercase for those values in the JSON so that it's consistent with other metrics.

File: docs/v2-system-addon/data_dictionary.md [hunk 1]
```diff
@@ -147,6 +147,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
 | `action` | [Required] Either `activity_stream_event`, `activity_stream_session`, or `activity_stream_performance`. | :one:
 | `addon_version` | [Required] The version of the Activity Stream addon. | :one:
 | `client_id` | [Required] An identifier for this client. | :one:
+| `card_type` | Either ("bookmark", "history", "trending", "pinned", "not_pinned") | :one:
```
ncloudioj: Let's add a qualifier, either `Optional` or `Required` to all the new fields. I guess all should be `Optional` since they're not available for the old versions.

ncloudioj: nit: `not_pinned` is not of the interest in this case, let's omit it and keep the payload compact.

docs/v2-system-addon/data_dictionary.md
```diff
@@ -170,7 +171,8 @@ and losing focus. | :one:
 | `ua` | [Auto populated by Onyx] The user agent string. | :two:
 | `unload_reason` | [Required] The reason the Activity Stream page lost focus. | :one:
 | `url` | [Optional] The URL of the recommendation shown in one of the highlights spots, if any. | :one:
-| `value` | [Required] An integer that represents the measured performance value. Can store counts, times in milliseconds, and should always be a positive integer.| :one:
+| `value` (performance) | [Required] An integer that represents the measured performance value. Can store counts, times in milliseconds, and should always be a positive integer.| :one:
+| `value` (event) | Has {"icon_type", "card_type"}, when user action was triggered from a topsite, or {"card_type"} otherwise | :one:
```
ncloudioj: How about `... | [Optional] An object with keys "icon_type" and "card_type" to record the extra information of event ping |`?

docs/v2-system-addon/data_dictionary.md
```diff
@@ -191,6 +193,7 @@ and losing focus. | :one:
 | `user_prefs` | [Required] The encoded integer of user's preferences. | :one: & :four:
 | `is_prerendered` | [Required] A boolean to signify whether the page is prerendered or not | :one:
 | `is_preloaded` | [Required] A boolean to signify whether the page is preloaded or not | :one:
+| `icon_type` | Either ("tippytop", "rich_icon", "screenshot_with_icon", "screenshot", "no_image") | :one:
```
ncloudioj: Let's add a `[Optional]` as well as a note to say that it's only available in the "click" event ping of Top sites.

docs/v2-system-addon/data_events.md
```diff
@@ -123,8 +135,11 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as
 ```js
 {
   "event": "BOOKMARK_ADD",
-  "source": "TOP_SITES",
+  "source": "HIGHLIGHTS",
```
ncloudioj: Nice catch!

docs/v2-system-addon/data_events.md
```diff
@@ -123,8 +135,11 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as
...
   "action_position": 2,
+  "value": {
+    "card_type": ["history" | "trending"]
```
ncloudioj: nit: perhaps s/trending/pocket/ ?

system-addon/content-src/components/TopSites/TopSites.jsx
```diff
@@ -9,25 +9,33 @@ import {TOP_SITES_MAX_SITES_PER_ROW} from "common/Reducers.jsm";
 import {TopSiteForm} from "./TopSiteForm";
 import {TopSiteList} from "./TopSite";
 
+function topSiteIconType(link) {
+  if (!link) {
+    return null;
```
ncloudioj: Be careful about this case, cause the following `acc[topSiteIconType(link)]++` might generate some funky results.

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
```diff
@@ -13,8 +13,15 @@ export class _LinkMenu extends React.PureComponent {
 
     // Handle special case of default site
     const propOptions = !site.isDefault ? props.options : DEFAULT_SITE_MENU_OPTIONS;
-
-    const options = propOptions.map(o => LinkMenuOptions[o](site, index, source)).map(option => {
+    // Extra telemetry information about the item
+    const siteInfo = {};
+    if (site.iconType) { // Topsite
+      siteInfo.icon_type = site.iconType;
+      siteInfo.card_type = site.isPinned ? "pinned" : "not_pinned";
```
ncloudioj: As mentioned earlier, we can leave out the card_type if it's "not_pinned".

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
```diff
@@ -24,7 +31,8 @@ export class _LinkMenu extends React.PureComponent {
             props.dispatch(ac.UserEvent({
               event: userEvent,
               source,
-              action_position: index
+              action_position: index,
+              value: siteInfo
```
ncloudioj: Since "value" is an optional field, let's simply leave it out if `Object.keys(siteInfo).length === 0`.

system-addon/content-src/lib/link-menu-options.js
```diff
@@ -71,15 +71,15 @@ export const LinkMenuOptions = {
       action_position: index
     })
   }),
-  DeleteUrl: (site, index, eventSource) => ({
+  DeleteUrl: (site, index, eventSource, value) => ({
```
ncloudioj: How about other menu options here? Such as `BLOCK`. Although I am not sure if we wanted to do this for all of them?

ncloudioj:
@piatra Just finished my first scan, please see the review comments for more details.

Also, apart from the "CLICK" event, I'm not sure if we wanted to apply this to all other events, such as block, delete etc. Perhaps do them as a followup only if needed in the future? Thoughts?

system-addon/content-src/lib/link-menu-options.js
piatra: BLOCK defines a user event here
https://github.com/mozilla/activity-stream/blob/f7bbd56c77ac849b5c6f5c702d63753c1b5e96bf/system-addon/content-src/lib/link-menu-options.js#L59
Which is handled here
https://github.com/mozilla/activity-stream/blob/f7bbd56c77ac849b5c6f5c702d63753c1b5e96bf/system-addon/content-src/components/LinkMenu/LinkMenu.jsx#L23-L28
So you should see the same topsite information as for the others. Delete was "special" and needed a different change.

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
piatra: Will that ever be true? All cards must have a "type" history, bookmark or trending and all topsites have an icon type attribute associated with them.

system-addon/content-src/components/TopSites/TopSites.jsx
piatra: There is actually a mistake here. That shouldn't be in. There is always a link object being passed in.

docs/v2-system-addon/data_events.md
piatra: The current type of the cards from the Recommended by Pocket section is "trending". This is not the same as the "items you have in your pocket list" that just landed. Maybe I should update to include that as well.

docs/v2-system-addon/data_dictionary.md
piatra: It should go out with any kind of action you take on a topsite (click, block, delete, pin etc).

docs/v2-system-addon/data_dictionary.md
piatra: If we consider that the "card_type" for Topsite items is send for more than just click do we still want to omit the "not_pinned" attribute?

ncloudioj: I don't think the "not_pinned" attribute is necessary here. If we were to fetch all the "not_pinned" event records from top sites, we can just do `WHERE source = 'TOPSITES' and card_type <> 'pinned'`

docs/v2-system-addon/data_events.md
ncloudioj: Sounds good, let's stick to "trending" then, leave the "pocket" for the future use (such as highlights).

system-addon/content-src/components/LinkMenu/LinkMenu.jsx
ncloudioj: Yep, as long as we can guarantee that `siteInfo` always contains at least one key, that's fine.

system-addon/content-src/lib/link-menu-options.js
ncloudioj: Oh, "DELETE" is a special case. So siteInfo will be attached to all the events then, correct?

piatra: Yes

piatra:
Updated. Now all of the sections:
* Topsites
* Top stories
* Highlights

Will send information on CLICK events, as well as for all of the context menu options.

"visited" highlights cards won't send any extra information because it's the default card type.
Topsites that are not pinned do not send "not_pinned" as a type but they do send their "icon_type" information: for example a tippytop website that you click on.

ncloudioj:
@fmarier Could you do a data review for this patch, please?

In a nutshell,  we add two fields to the event ping,
* card_type
* icon_type

For more details, see the change in `data_event.md` and `data_dictionary.md`.

Do we need to fill out the data review request form for this? Let us know.

Thanks!

fmarier:
Yes, if you could fill out the data review request form, that would be great. Attach it on bug 1425457 as a text file and r? me on it.

ncloudioj:
@fmarier here you go. https://bugzilla.mozilla.org/show_bug.cgi?id=1425457#c2

ncloudioj:
Data review is already done and tracked in the bugzilla bug.

@piatra 🚢 !!! It needs some rebase work.

as-pine-proxy:
[Treeherder: 1d0da51cb8b44221a3fff9e03897e939f15a4c02](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1d0da51cb8b44221a3fff9e03897e939f15a4c02)

PR #3993, opened 2018-02-23
fix(tests): Update mochitests for section context menus

Mardak:
r?@sarracini Needed to fix these for export because they were assuming only context menus on sites/cards. cc @rlr


Mardak:
@rlr The add top site button test still assumes the link/option is available even when the menu isn't open. Also noticed that prerendered files are a bit larger because each section has a context menu with all options prerendered instead of lazily. for https://bugzilla.mozilla.org/show_bug.cgi?id=1440456

as-pine-proxy:
[Treeherder: 226f340b244ab504cd13006252be6d138890830d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=226f340b244ab504cd13006252be6d138890830d)

PR #3999, opened 2018-02-24
 Fix Bug 1440781 - Refreshing preloaded tabs interferes with spoc placement

csadilek:
Fixes a regression, introduced in 59, but plan is to fix for 60 only.

This PR introduces a way to mark cards as pinned, so they don't get refreshed as part of a SECTION_UPDATEs.

This is useful to avoid losing spoc cards when preloaded tabs refresh based on the latest feed data. A preloaded tab with a spoc will still refresh and all its cards will update except the spoc (pinned card) which will remain in place. This is to make sure we keep the spoc frequency as configured in the feed.

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -207,6 +207,18 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           // If the action is updating rows, we should consider initialized to be true.
           // This can be overridden if initialized is defined in the action.data
           const initialized = action.data.rows ? {initialized: true} : {};
+
+          if (action.data.rows && section.rows.find(card => card.pinned)) {
+            // Make sure that pinned cards stay at their current position
+            const rows = Array.from(action.data.rows);
+            section.rows.forEach((card, index) => {
+              if (card.pinned) {
+                rows.splice(index, 0, card);
+              }
+            });
+            return Object.assign({}, section, initialized, Object.assign({}, action.data, {rows}));
+          }
+
           return Object.assign({}, section, initialized, action.data);
         }
         return section;
--- a/system-addon/lib/TopStoriesFeed.jsm
+++ b/system-addon/lib/TopStoriesFeed.jsm
@@ -311,7 +311,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
         // Create a new array with a spoc inserted at index 2
         const position = SectionsManager.sections.get(SECTION_ID).order;
         let rows = this.store.getState().Sections[position].rows.slice(0, this.stories.length);
-        rows.splice(2, 0, spocs[0]);
+        rows.splice(2, 0, Object.assign(spocs[0], {pinned: true}));
 
         // Send a content update to the target tab
         const action = {type: at.SECTION_UPDATE, data: Object.assign({rows}, {id: SECTION_ID})};
--- a/system-addon/test/unit/common/Reducers.test.js
+++ b/system-addon/test/unit/common/Reducers.test.js
@@ -291,6 +291,21 @@ describe("Reducers", () => {
       const updatedSection = newState.find(section => section.id === "foo_bar_2");
       assert.propertyVal(updatedSection, "initialized", true);
     });
+    it("should retain pinned cards on SECTION_UPDATE", () => {
+      const ROW = {id: "row"};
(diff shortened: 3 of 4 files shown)


system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -311,7 +311,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
         // Create a new array with a spoc inserted at index 2
         const position = SectionsManager.sections.get(SECTION_ID).order;
         let rows = this.store.getState().Sections[position].rows.slice(0, this.stories.length);
-        rows.splice(2, 0, spocs[0]);
+        rows.splice(2, 0, Object.assign(spocs[0], {pinned: true}));
 
         // Send a content update to the target tab
         const action = {type: at.SECTION_UPDATE, data: Object.assign({rows}, {id: SECTION_ID})};
```
Mardak: Might not be an issue now, but potentially this `SECTION_UPDATE` could result in multiple spocs trying to be in index 2 if some reason `maybeAdd` is called multiple times for a given page?

csadilek: That would be the case if NEW_TAB_REHYDRATED is fired multiple times for a tab, yes, but we do have a check to only fire this event once: https://github.com/mozilla/activity-stream/blob/master/system-addon/content-src/components/Base/Base.jsx#L46

Mardak: Oh I was thinking more of in the future if we had multiple callers of `maybeAdd` where we would need to be extra careful. But we're not there now, so maybe not a real issue! ;)

csadilek: OK, I guess let's wait and see if the spoc feature sticks around and how, once experimentation ends. Thanks for your help!

system-addon/common/Reducers.jsm
```diff
@@ -207,6 +207,18 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           // If the action is updating rows, we should consider initialized to be true.
           // This can be overridden if initialized is defined in the action.data
           const initialized = action.data.rows ? {initialized: true} : {};
+
+          if (action.data.rows && section.rows.find(card => card.pinned)) {
```
Mardak: Why the additional `find` check? Doesn't seem necessary?

system-addon/lib/TopStoriesFeed.jsm
```diff
@@ -311,7 +311,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
         // Create a new array with a spoc inserted at index 2
         const position = SectionsManager.sections.get(SECTION_ID).order;
         let rows = this.store.getState().Sections[position].rows.slice(0, this.stories.length);
-        rows.splice(2, 0, spocs[0]);
+        rows.splice(2, 0, Object.assign(spocs[0], {pinned: true}));
```
Mardak: This modifies the `spoc[0]` object making future references also `pinned`, which I guess is fine. Would it make more sense to just treat all as pinned when it first gets assigned?

system-addon/common/Reducers.jsm
csadilek: Just so we don't copy the rows, and add the additional processing (iterating over all rows), if there's no pinned card (almost always).

system-addon/lib/TopStoriesFeed.jsm
csadilek: We could do it here for all spocs: https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/TopStoriesFeed.jsm#L97

But, I guess, only a subset of spocs in the feed will ever have a chance of being displayed for a given client. If you prefer, I can move it up and mark them all as pinned.

system-addon/common/Reducers.jsm
Mardak: Eh. I don't think this is particularly performance critical section, so simpler code is preferable, but this is fine.

system-addon/lib/TopStoriesFeed.jsm
Mardak: We've typically avoided modifying objects with `Object.assign` (and mostly to create a new object) as they're harder to track down, but this should be okay.

as-pine-proxy:
[Treeherder: dc2ce899777cd38065228fb4295211512456dd5c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=dc2ce899777cd38065228fb4295211512456dd5c)

PR #4003, opened 2018-02-26
fix(topstories): Pinned cards should be cleared when section is re-initialized

csadilek:
This refines the earlier fix to *not* retain pinned cards when the section is disabled and enabled again i.e. by an .options change or when the user toggles the checkbox in the pref pane.

I think in these cases we do want the cards to update. Disabling the section causes a SECTION_UPDATE event with an empty rows array.

### DIFF
--- a/system-addon/common/Reducers.jsm
+++ b/system-addon/common/Reducers.jsm
@@ -208,8 +208,9 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           // This can be overridden if initialized is defined in the action.data
           const initialized = action.data.rows ? {initialized: true} : {};
 
-          if (action.data.rows && section.rows.find(card => card.pinned)) {
-            // Make sure that pinned cards stay at their current position
+          // Make sure pinned cards stay at their current position when rows are updated.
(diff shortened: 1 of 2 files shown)


system-addon/common/Reducers.jsm
```diff
@@ -208,7 +208,7 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
           // This can be overridden if initialized is defined in the action.data
           const initialized = action.data.rows ? {initialized: true} : {};
 
-          if (action.data.rows && section.rows.find(card => card.pinned)) {
+          if (action.data.rows && action.data.rows.length > 0 && section.rows.find(card => card.pinned)) {
             // Make sure that pinned cards stay at their current position
```
k88hudson: Can you just update this comment to reflect the change you just made?

csadilek: OK, thanks, updated.

as-pine-proxy:
[Treeherder: 2b797073eeeab37c81d0ec2d79813a76a98316fd](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2b797073eeeab37c81d0ec2d79813a76a98316fd)

PR #4005, opened 2018-02-27
chore(mc): Backport of Bug 1440276 - Activity Stream spikes after Bookmark activity, triggering lots of main thread work

sarracini:
Adds a ```skipTags: true``` property to the observer so it knows that it should not notify observers when adding/removing tags during a sync.
Also removes unused code for ```onItemChanged()```

This doesn't actually work yet with ```onItemChanged```.... posted my findings on https://bugzilla.mozilla.org/show_bug.cgi?id=1440276

_Update: Since this bug now involves some changes in ```nsNavBookmarks.cpp```, I'm going to make the changes to ```PlacesFeed.jsm``` also in m-c and this will become a backport PR of those changes_

### DIFF
--- a/system-addon/lib/PlacesFeed.jsm
+++ b/system-addon/lib/PlacesFeed.jsm
@@ -90,6 +90,7 @@ class HistoryObserver extends Observer {
 class BookmarksObserver extends Observer {
   constructor(dispatch) {
     super(dispatch, Ci.nsINavBookmarkObserver);
+    this.skipTags = true;
   }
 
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: 90727f978bebec1aa00f8ebd3dee5a56008d9532](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=90727f978bebec1aa00f8ebd3dee5a56008d9532)

PR #4011, opened 2018-03-01
Bug 1441984 - Only show OpenInPrivateBrowsing option if private browsing enabled

k88hudson:
This should fix the issue of only showing OpenInPrivateBrowsing when private browsing is enabled, I decided to add the value into `Prefs` since `SnippetsFeed` could be turned off. Alternatively, I could add it to a completely new file/reducer, but i don't think overhead is worth it

cc @Mardak 

@rlr what do you think about connecting the `LinkComponent` again? I had to refactor tests to get it work

### DIFF
--- a/system-addon/content-src/lib/link-menu-options.js
+++ b/system-addon/content-src/lib/link-menu-options.js
@@ -1,5 +1,15 @@
 import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 
+const _OpenInPrivateWindow = site => ({
+  id: "menu_action_open_private_window",
+  icon: "new-window-private",
+  action: ac.OnlyToMain({
+    type: at.OPEN_PRIVATE_WINDOW,
+    data: {url: site.url, referrer: site.referrer}
+  }),
+  userEvent: "OPEN_PRIVATE_WINDOW"
+});
+
 /**
  * List of functions that return items that can be included as menu options in a
  * LinkMenu. All functions take the site as the first parameter, and optionally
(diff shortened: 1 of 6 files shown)


system-addon/lib/PrefsFeed.jsm
```diff
@@ -63,6 +66,9 @@ this.PrefsFeed = class PrefsFeed {
       values[name] = this._prefs.get(name);
     }
 
+    // Not a pref, but we need this to determine whether to show private-browsing-related stuff
+    values.isPrivateBrowsingEnabled = PrivateBrowsingUtils.enabled;
```
k88hudson: @mkaply does this look right to you? Also, how would we go about watching for changes to this?

system-addon/content-src/lib/link-menu-options.js
```diff
@@ -1,5 +1,15 @@
 import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 
+const _OpenInPrivateWindow = site => ({
+  id: "menu_action_open_private_window",
+  icon: "new-window-private",
+  action: ac.AlsoToMain({
```
rlr: you didn't change this, but should it be `ac.OnlyToMain(...` instead?

system-addon/lib/PrefsFeed.jsm
k88hudson: Ok perfect, thanks!

as-pine-proxy:
[Treeherder: 32a0da06f91dec8ea7640b98b0706bdadfb60971](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=32a0da06f91dec8ea7640b98b0706bdadfb60971)

PR #4014, opened 2018-03-01
Fix Bug 1438368 - about:home about:newtab snippet blocking

piatra:
The problem this PR works around is that if `about:home` and `about:newtab` are not both open at the same time, then that about page (that is not open) would not receive the blocked snippets updates.

Most of the work was done, there were a few changes needed
* added ActivityStreamStorage.jsm to store the blocklist
* prevent the indexdDB on the content side from updating the blocklist
  * use the blocklist passed in the init instead

Need to add tests. The way I tried this locally was calling `gSnippetsMap.blockSnippetById` by hand. You can clear by calling `.clear()`.
This requires an appropriate change on the snippets side (I will follow up with that PR).


piatra:
Regarding the PR for snippets-service: we can't simply check if `blockSnippetById` exists and call it. A scenario where this fails:
* Open the browser (default page about:home) no other tabs open
* Block snippet
* Open a new tab and block the snippet you get on about:newtab
* about:newtab blocklist will overwrite about:home blocklist on sync losing the info about the first blocked snippet

Maybe we can implement a version number in `snippets.js` that we could target from [the snippets js file](https://github.com/mozmeao/snippets-service/blob/master/snippets/base/templates/base/includes/snippet_as.js).
Snippets do version targeting but the js file provided is unique.

docs/v2-system-addon/snippets.md
```diff
@@ -30,7 +30,7 @@ async function setFoo() {
 
 `.blockList`: (Array) An array of snippet IDs that have been blocked by the users, or an empty array if none have been blocked.
 
-`.blockSnippetById(id)`: (func) A function that adds an id to a blockList and returns
+`.blockSnippet(id)`: (func) A function that adds an id to a blockList and returns
```
k88hudson: yep this sounds fine to me, we can bump the version if a more major change is made

k88hudson:
Re: indexedDB v.s. using file-based storage, I'm under the impression indexedDB is the better solution for this based on some documents about moving away from prefs; maybe we can consider getting rid of / migrating the other file-based storage mechanism when we are more confident about the reliability of indexedDB?

dmose:
One thing that _might_ complicate things, or might not be a problem, is
that about:home and about:newtab are not "real" origins in the eyes of the
browser.  If that does turn out to be a problem, there's a bug on file
which could make them be real origins.

2018-03-06 7:32 GMT-08:00 Kate Hudson <notifications@github.com>:

system-addon/common/Reducers.jsm
```diff
@@ -334,6 +334,8 @@ function Snippets(prevState = INITIAL_STATE.Snippets, action) {
   switch (action.type) {
     case at.SNIPPETS_DATA:
       return Object.assign({}, prevState, {initialized: true}, action.data);
+    case at.SNIPPET_BLOCKED:
```
k88hudson: The original implementation was kind of a stop-gap measure, but I think it might be a better  if instead of passing the entire block list around every time, we just send a single ID and append it here.

system-addon/content-src/lib/snippets.js
```diff
@@ -35,6 +35,7 @@ export class SnippetsMap extends Map {
 
   clear() {
     super.clear();
+    this._dispatch(ac.AlsoToMain({type: at.SNIPPETS_BLOCKLIST_UPDATED, data: []}));
```
k88hudson: This isn't necessarily what we want to do here, you might have an out-of-date indexedDB version or something on about:home/about:newtab that shouldn't necesarrily affect the main store

system-addon/lib/SnippetsFeed.jsm
```diff
@@ -126,6 +128,14 @@ this.SnippetsFeed = class SnippetsFeed {
     this.store.dispatch(ac.BroadcastToContent({type: at.SNIPPETS_DATA, data}));
   }
 
+  _saveBlockList(blockList) {
+    this._storage.set("blockList", blockList);
```
k88hudson: Probably best to return this `Promise` in case it needs to be awaited

k88hudson:
@dmose this is implementing IndexedDB in the main process, not in `about:home`/`about:newtab`

piatra:
@k88hudson updated!

system-addon/content-src/lib/snippets.js
```diff
@@ -212,6 +214,7 @@ export class SnippetsProvider {
 
     if (cachedVersion !== this.appData.version) {
       this.snippetsMap.clear();
+      this.snippetsMap._dispatch(ac.AlsoToMain({type: at.SNIPPETS_BLOCKLIST_CLEARED}));
```
k88hudson: can you move this to the .clear method so it can be called from the console?

k88hudson:
Thanks, good to go after that change! 👍 🎆

docs/v2-system-addon/snippets.md
glogiotatidis: As stated in https://github.com/mozmeao/snippets-service/pull/365 from the snippets-service side it's better that we keep the same name.

as-pine-proxy:
[Treeherder: 43d9b2dbf63220435742eb5dfb2b63eded6c6064](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=43d9b2dbf63220435742eb5dfb2b63eded6c6064)

PR #4015, opened 2018-03-02
Fix Bug 1404890 - Move new tab settings for sections to about:preferences

Mardak:
r?@k88hudson Added a new feed (only so it can access `Sections` state -- originally I had it as part of `PrefsFeed`, but turns out `Preferences` https://searchfox.org/mozilla-central/source/toolkit/content/preferencesBindings.js manages all the updating just fine.)

There is an unused string ` +prefs_restore_defaults_button=Restore Defaults ` which would be needed for bug 1434751. Pretty sure there aren't others to add to activity stream, although there are others that would be needed as part of mozilla-central: https://docs.google.com/document/d/1uI7xtTuI9pw9rbOzN7q1kjUeIOEH3wfJtFr-4-degzQ

Design: https://mozilla.invisionapp.com/share/EGFI16P8Y36#/screens

Screenshot (with testing a locked search pref):
screen shot 2018-03-01 at 2 59 23 pm

There are no tests for the new jsm right now, but I'll try adding some to at least make sure it handles actions/events. But for testing the details of what gets rendered, that could get tricky and content-src's coverage is different, so not sure if we should treat it differently from other jsms.

### DIFF
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -10,6 +10,9 @@ const {actionCreators: ac, actionTypes: at} = ChromeUtils.import("resource://act
 
 ChromeUtils.defineModuleGetter(this, "PlacesUtils", "resource://gre/modules/PlacesUtils.jsm");
 
+// We need to support both content sidebar and about:preferences for experiments
+let ABOUT_PREFERENCES = true;
+
 /*
  * Generators for built in sections, keyed by the pref name for their feed.
  * Built in sections may depend on options stored as serialised JSON in the pref
@@ -20,10 +23,10 @@ const BUILT_IN_SECTIONS = {
     id: "topstories",
     pref: {
       titleString: {id: "header_recommended_by", values: {provider: options.provider_name}},
-      descString: {id: options.provider_description || "pocket_description"},
+      descString: {id: ABOUT_PREFERENCES ? "prefs_topstories_description" : options.provider_description || "pocket_description"},
       nestedPrefs: options.show_spocs ? [{
         name: "showSponsored",
-        titleString: {id: "settings_pane_topstories_options_sponsored"},
+        titleString: {id: ABOUT_PREFERENCES ? "prefs_topstories_show_sponsored_label" : "settings_pane_topstories_options_sponsored", values: {provider: options.provider_name}},
         icon: "icon-info"
       }] : []
     },
@@ -54,7 +57,7 @@ const BUILT_IN_SECTIONS = {
     id: "highlights",
     pref: {
       titleString: {id: "settings_pane_highlights_header"},
-      descString: {id: "settings_pane_highlights_body2"}
+      descString: {id: ABOUT_PREFERENCES ? "prefs_highlights_description" : "settings_pane_highlights_body2"}
     },
     shouldHidePref:  false,
     eventSource: "HIGHLIGHTS",
@@ -76,6 +79,7 @@ const SectionsManager = {
   initialized: false,
   sections: new Map(),
   init(prefs = {}) {
+    ABOUT_PREFERENCES = prefs["feeds.aboutpreferences"];
     for (const feedPrefName of Object.keys(BUILT_IN_SECTIONS)) {
       const optionsPrefName = `${feedPrefName}.options`;
       this.addBuiltInSection(feedPrefName, prefs[optionsPrefName]);
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -10,9 +10,6 @@ const {actionCreators: ac, actionTypes: at} = ChromeUtils.import("resource://act
 
 ChromeUtils.defineModuleGetter(this, "PlacesUtils", "resource://gre/modules/PlacesUtils.jsm");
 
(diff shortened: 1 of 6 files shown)


system-addon/lib/SectionsManager.jsm
```diff
@@ -78,6 +81,7 @@ const SectionsManager = {
   initialized: false,
   sections: new Map(),
   init(prefs = {}) {
+    ABOUT_PREFERENCES = prefs["feeds.aboutpreferences"];
```
Mardak: This should be temporary only for experimentation. The idea is for implementing bug 1432672 to open about:preferences instead of sidebar, it would also look at this same feed pref.

locales/en-US/strings.properties
```diff
@@ -110,7 +110,22 @@ time_label_minute={number}m
 time_label_hour={number}h
 time_label_day={number}d
 
-# LOCALIZATION NOTE (settings_pane_*): This is shown in the Settings Pane sidebar.
+# LOCALIZATION NOTE (prefs_*, settings_*): These are shown in about:preferences
+# for a "Firefox Home" section. "Home" should be localized matching the sidebar.
```
flodolo: Let's add at the end: ", while Firefox should be treated as a brand and kept in English"

BTW, where is the string for the sidebar? There is no pane for the home page at this point

locales/en-US/strings.properties
```diff
@@ -110,7 +110,22 @@ time_label_minute={number}m
...
+prefs_home_description=Choose what content you want on your Firefox Home screen.
+prefs_restore_defaults_button=Restore Defaults
+# LOCALIZATION NOTE (prefs_section_rows_option): This is a semi-colon list of
+# plural forms used in a drop down of multiple row options (1 row, 2 rows).
+# See: http://developer.mozilla.org/en/docs/Localization_and_Plurals
+prefs_section_rows_option={num} row;{num} rows
+prefs_search_header=Web Search
+prefs_topsites_description=The sites you visit most
+prefs_topstories_description=High-quality content you might otherwise miss
+prefs_topstories_show_sponsored_label={provider} Sponsored Stories
```
flodolo: Add a note for `{provider}`? E.g. `{provider} is going to be replaced by the name of a new provider (e.g. 'Pocket')`

locales/en-US/strings.properties
flodolo: I guess the answer to the latter is in

> …  although there are others that would be needed as part of mozilla-central

Mardak: Yeah, I suppose it's confusing in that the mozilla-central string doesn't exist yet, so I tried to be a bit more descriptive:
```
# LOCALIZATION NOTE (prefs_*, settings_*): These are shown in about:preferences
# for a "Firefox Home" section. "Firefox" should be treated as a brand and kept
# in English, while "Home" should be localized matching the about:preferences
# sidebar mozilla-central string for the panel that has preferences related to
# what is shown for the homepage, new windows, and new tabs.
```

locales/en-US/strings.properties
Mardak: Added

k88hudson:
This is pretty much good to go, I will update after we land https://bugzilla.mozilla.org/show_bug.cgi?id=1443337

Mardak:
This can actually land before bug 1443337 creates the "Home" panel as this injected content will appear after "Home page" group. So before the other bug lands, this content will just appear in General.

Mardak:
@k88hudson I rebased this on latest master with strings, and I think this should be fine to land right now for early 61, and the section will just jump to the right spot when the Home panel lands.

Landing this now means @sarracini is unblocked for the gear work in https://bugzilla.mozilla.org/show_bug.cgi?id=1432672

system-addon/lib/SectionsManager.jsm
```diff
@@ -20,10 +23,10 @@ const BUILT_IN_SECTIONS = {
     id: "topstories",
     pref: {
       titleString: {id: "header_recommended_by", values: {provider: options.provider_name}},
-      descString: {id: options.provider_description || "pocket_description"},
+      descString: {id: ABOUT_PREFERENCES ? "prefs_topstories_description" : options.provider_description || "pocket_description"},
```
Mardak: I suppose this condition can be removed because we're no longer experimenting. Separately, should we still just leave about preferences as a feed that someone could disable.. and somewhat break things? I guess that user would just not see activity stream prefs in about:preferences…

Mardak:
I added a second commit that removes the `ABOUT_PREFERENCES` check from sections that should be squashed when merging.

Mardak:
Added prefs with 💯% coverage!

Mardak:
Added tests with 💯% coverage! Doesn't particularly test the layout/structure, but it stubs enough to trigger every code path.

@k88hudson a quick look at the code changes to get the testing working? Moving to `init` instead of `constructor` to simplify testing. Similarly checking for `!== "object"` works for `undefined` string ids to simplify dummy test values.

as-pine-proxy:
[Treeherder: b91bc8a97f7ee72377067ae29972d5f57679dad7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b91bc8a97f7ee72377067ae29972d5f57679dad7)

PR #4018, opened 2018-03-05
Fix Bug 1433209 - Remove unused strings to simplify translating

piatra:


Mardak:
The purpose of the bug was to remove strings to simplify translating -- not just the strings that became unused in 60. There's probably various other strings that came over from test pilot version too.

Mardak:
E.g., time_label seems unused. Although unclear if we will need them in the near future although I don't think I've seen designs with those recently.

piatra:
Some more unused strings:

* default_label_loading
* header_visit_again
* header_bookmarks
* header_bookmarks_placeholder
* header_stories_from
* search_for_something_with
* search_settings
* section_info_send_feedback
* time_label_*

I wrote a script to return the number of file occurrences. Not super precise, I just compare with the number of known unused strings. Usually unused strings are also grouped together.

```bash
# select lines where a string id is defined
cat ./locales/en-US/strings.properties | grep = | while read line
do
  # get the id of the string from that line
  sid=$(echo $line | cut -d "=" -f 1)
  # get the number of files where it occurs
  echo "$sid $(grep $sid -Rl . | wc -l)"
done
```

Not sure how to decide on this though.

Mardak:
Why was this merged? It looks like there's quite a few more strings that can be removed.

as-pine-proxy:
[Treeherder: 446f7cf1a751d9ba7f14dbebb4c0622748561bec](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=446f7cf1a751d9ba7f14dbebb4c0622748561bec)

PR #4019, opened 2018-03-05
 Fix Bug 1443243 - Preloaded tab is sending impression ping after section was disabled

csadilek:
Fixed this by removing the visibility change listener when the section is removed. A new one will be attached anyway when a section is (re-)created. The listener was firing for sections that no longer exist.


as-pine-proxy:
[Treeherder: f1ea13c53c156530f28a34d7319be004d132bb30](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=f1ea13c53c156530f28a34d7319be004d132bb30)

PR #4020, opened 2018-03-05
Fix Bug 1441278 - Missing line numbers in failing unit tests

piatra:
WFM now

>  @webpack:///system-addon/test/unit/content-src/components/PreferencesPane.test.jsx:101:4 <- system-addon/test/unit/unit-entry.js:89146:5

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "husky": "0.14.3",
     "istanbul-instrumenter-loader": "0.2.0",
     "joi-browser": "13.0.1",
-    "karma": "2.0.0",
+    "karma": "1.7.1",
     "karma-chai": "0.1.0",
     "karma-coverage": "1.1.1",
     "karma-firefox-launcher": "1.1.0",


Mardak:
Is there a good way to fail a unit test so we don't accidentally upgrade karma again? Would checking something in `Error().stack` work?

dmose:
Not sure if this PR is entirely analogous to the thing that caused the
Karma failure, but here is an example of something like the checking that
Ed is suggestion: https://github.com/getsentry/raven-js/pull/837/files

2018-03-06 7:19 GMT-08:00 Ed Lee <notifications@github.com>:

piatra:
Error().stack seems to contain the non-sourced-map version of the line numbers.
Conversion seems to be done separate to running the tests https://github.com/karma-runner/karma/blob/e7b9ab89ac919eb5b4a479dfc0ce0e352d1273c7/lib/reporter.js#L10-L26

dmose:
Maybe you could just require("karma/lib/reporter") and call the API directly to do the conversion by hand and see if it's what you want?  Maybe that's more fragile than it's worth, though...

dmose:
I mean, really, this is only a problem until Karma 2.next gets released, and one would hope that won't be too long.  So I'm kind of inclined to say we shouldn't overrotate here.

as-pine-proxy:
[Treeherder: b384e28030322bd94d4013ccef9a8cfb708285e3](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b384e28030322bd94d4013ccef9a8cfb708285e3)

PR #4022, opened 2018-03-06
Fix Bug 1433212 - Implement section reordering

### LINKED ISSUE #3251: Clicking context menu separator wrongly dismisses the menu
[Affected versions]:
- Firefox 57.0a1 Build ID 	20170827100428

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:

[Steps to reproduce]:
2. Click the "..." button for a Top Site.
3. Click the line between two options and observe the behavior.
 
[Expected results]:
- Nothing happens.

[Actual results]:
- The menu is dismissed.

[Notes]:
- The issue is not reproducible with any other context menu.
top sites context menu dismissed

rlr:
r? @k88hudson Fix #3251 too

Screenshots:
screen shot 2018-03-06 at 10 17 24 am
screen shot 2018-03-06 at 10 17 37 am
screen shot 2018-03-06 at 10 17 41 am


system-addon/common/Reducers.jsm
```diff
@@ -177,28 +177,11 @@ function Sections(prevState = INITIAL_STATE.Sections, action) {
         }
         return section;
       });
-
-      // Invariant: Sections array sorted in increasing order of property `order`.
-      // If section doesn't exist in prevState, create a new section object. If
-      // the section has an order, insert it at the correct place in the array.
-      // Otherwise, prepend it and set the order to be minimal.
+      // Otherwise, append it
```
Mardak: I suppose @k88hudson is taking a look, but I believe for webextensions, we might want to prepend. Or insert after topsites. Or.. only worry about it when webextension stuff lands. ;)

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
```diff
@@ -50,6 +50,11 @@
             color: $white;
           }
         }
+
+        &.disabled {
+          opacity: 0.4;
+          cursor: default;
```
Mardak: I believe if we make it `pointer-events: none` there would be no click event to specially handle anyway?

system-addon/common/Reducers.jsm
rlr: the order here isn't significant because we order them when rendering. On section register, we do prepend to the order pref. https://github.com/mozilla/activity-stream/pull/4022/files#diff-46c4b10ce28f1463cea556e4020e95f4R284

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
rlr: I had it like that but then the click would still bubble up to window and the context menu would hide. And I don't think that's the expected behavior.

system-addon/common/Reducers.jsm
rlr: But that's a good point. I am currently prepending to the top above topsites. But maybe we should check if the first slot is topsites then insert in the second slot?

rlr: And I guess it doesn't matter until that stuff lands yeah 👍

Mardak: Oh nevermind, I see the prepending happening in `SectionsManager`.

system-addon/lib/SectionsManager.jsm
```diff
@@ -282,6 +280,13 @@ class SectionsFeed {
   onAddSection(event, id, options) {
     if (options) {
       this.store.dispatch(ac.BroadcastToContent({type: at.SECTION_REGISTER, data: Object.assign({id}, options)}));
+
+      // Make sure the section is in sectionOrder pref. Otherwise, prepend it.
+      const orderedSections = this.orderedSectionIds;
+      if (!orderedSections.includes(id)) {
+        orderedSections.splice(0, 0, id);
```
Mardak: nit: `orderedSections.unshift(id)` ;)

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -186,13 +189,46 @@ Section.defaultProps = {
...
+        key: section.id,
+        isFirst: i === 0,
+        isLast: i === sectionsData.length - 1
+      };
+      if (section.id === "topsites") {
+        sections.push(<TopSites {...extraProps} />);
+      } else {
+        sections.push(<SectionIntl {...section} Prefs={this.props.Prefs} dispatch={this.props.dispatch} {...extraProps} />);
+      }
+    }
```
rlr: while you're looking, if you have any ideas on how to simplify these two functions above 😄 

The issue is that I need to figure out all the sections I am going to render and their order first, to then apply the `isFirst` and `isLast` props. Initially, I was throwing the `<components/>` into the array in one loop but then there was no way to go update the props on them. So I have to loop twice. If that makes sense?

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
Mardak: Ah okay. Although.. there is a separate bug #3251 about clicking the separator causing the menu to dismiss..... I suppose a larger fix would be to eat events on the menu that aren't for actionable options.

system-addon/lib/SectionsManager.jsm
rlr: *mindblown* I wonder who names these things

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
rlr: Oh hmm. Maybe the context menu container can stop propagation on all clicks. And clicks to valid items can call `hideContext()` directly. Will try

system-addon/content-src/components/Sections/Sections.jsx
Mardak: You should be able to do something like this from the first/only loop?
```js
{
  isFirst: sectionsData.length === 0,
  isLast: sectionsData.length == enabledSections.length - 1
}
```

And instead of pushing temporary objects, just push the jsx?

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
rlr: da15d8a  seems to work 💃

system-addon/content-src/components/Sections/Sections.jsx
rlr: ooh. hmm. yeah. I guess that works as long as the user doesn't manually mess up the sectionOrder pref. I'm fine with that.

system-addon/content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -39,9 +44,11 @@ export class ContextMenuItem extends React.PureComponent {
     this.onKeyDown = this.onKeyDown.bind(this);
   }
 
-  onClick() {
-    this.props.hideContext();
-    this.props.option.onClick();
+  onClick(event) {
+    if (!this.props.option.disabled) {
```
Mardak: This change (`event` prop and `disabled` check) shouldn't be needed after the `pointer-events: none` as the click never goes to the option.

system-addon/lib/SectionsManager.jsm
Mardak: Yeah! We should have `push` and `unpush` or `pop` and `unpop` for some consistency! ;) I would think `shift` is from https://www.unix.com/man-page/linux/1/shift/ and `unshift` is the pattern following `unget` http://heirloom.sourceforge.net/devtools/unget.1.html

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -186,13 +189,35 @@ Section.defaultProps = {
...
+        key: sectionId,
+        isFirst: sections.length === 0,
+        isLast: sections.length === expectedCount - 1
+      };
+      if (sectionId === "topsites" && showTopSites) {
+        sections.push(<TopSites {...commonProps} />);
+      } else {
+        const section = enabledSections.find(s => s.id === sectionId);
+        if (section) {
+          sections.push(<SectionIntl {...section} Prefs={this.props.Prefs} dispatch={this.props.dispatch} {...commonProps} />);
```
k88hudson: Personally, I would just connect `SectionIntl` to redux instead of passing in `Prefs` and `dispatch`

system-addon/test/unit/lib/SectionsManager.test.js
```diff
@@ -493,5 +515,52 @@ describe("SectionsFeed", () => {
       assert.calledOnce(stub);
       assert.calledWith(stub, "Foo", "bar.com");
     });
+    it("should call the feed's uninit on UNINIT", () => {
```
k88hudson: this test name needs to be updated

system-addon/test/unit/lib/SectionsManager.test.js
```diff
@@ -493,5 +515,52 @@ describe("SectionsFeed", () => {
...
+      sinon.stub(feed, "moveSection");
+      const id = "topsites";
+      const direction = +1;
+      feed.onAction({type: "SECTION_MOVE", data: {id, direction}});
+
+      assert.calledOnce(feed.moveSection);
+      assert.calledWith(feed.moveSection, id, direction);
+    });
+  });
+  describe("#moveSection", () => {
```
k88hudson: can you add a test to make sure items are moved correctly when there is a disabled item inbetween?

e.g. if your sections are:

```
 [{id: "A", enabled: true}, {id: "B", enabled: false}, {id: "C", enabled: true}]
```

and you move `C` up, the order should be `C`, `A`, `B`, not `A`, `C`, `B`

rlr:
fixes pushed ^ thanks!

k88hudson:
🔥🔥🔥 This is great! Thanks 👍

as-pine-proxy:
[Treeherder: 1c3d9009e6e8a12598748a950a8b61559e488d28](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=1c3d9009e6e8a12598748a950a8b61559e488d28)

PR #4024, opened 2018-03-07
Bug 1410828 - Add a test to verify that the next top site in queue replaces a dismissed top site

icrisan:
Bug 1410828 - Add a test to verify that the next top site in queue replaces a dismissed top site


piatra:
This should probably go in the `browser_topsites_contextMenu_options.js` since it's testing one of the options there.

system-addon/test/functional/mochitest/browser_topsites_dismissItem.js
```diff
@@ -0,0 +1,34 @@
...
+  before: setDefaultTopSites,
+  test: async function defaultTopSites_dismiss() {
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".top-site-icon"),
+      "Topsite tippytop icon not found");
+
+    let defaultTopSitesNumber = content.document.querySelector(".top-sites-list").querySelectorAll("[class=\"top-site-outer\"]").length;
+    Assert.equal(defaultTopSitesNumber, 6, "6 top sites are loaded by default");
+
+    let firstTopSite = content.document.querySelector(".top-sites-list li:first-child a").getAttribute("href");
+    Assert.equal(firstTopSite, "https://www.youtube.com/");
```
piatra: Let's not test for exact string matches. We can get the `href` of the second topsite and check that it moved to the first position.

icrisan:
Hi @piatra! I've just added the changes requested.

piatra:
Tests fail for me locally when running with `--run-until-failure` the problem is that the test sets the blocked value in `browser.newtabpage.blocked` pref it doesn't get cleared. The test needs to clean both the pref and `NewTabUtils.blockedLinks._links` which does not get cleared on pref change. 
I'll update when I have a solution for this.

Mardak:
There's a `NewTabUtils.restoreAll()` https://searchfox.org/mozilla-central/rev/33cc9e0331da8d9ff3750f1e68d72d61201176cb/toolkit/modules/NewTabUtils.jsm#2228-2238

or not sure if calling directly `NewTabUtils.blockedLinks.resetCache()` is good enough…

piatra:
I needed to do
```js
  async after() {
    await SpecialPowers.pushPrefEnv({set: [["browser.newtabpage.blocked", ""]]});
    await new Promise(resolve => NewTabUtils.undoAll(resolve));
    await setDefaultTopSites();
  }
```

Not sure if it's something specific to the way `--verify` runs tests but I needed to restore DEFAULT_TOPSITES at the end of the test as well.

Mardak:
Superseded by #4036. @icrisan, curious, why not just update this PR instead of creating a new one?

PR #4027, opened 2018-03-07
(nobug): Add strings for New Windows and Tabs section of prefs

k88hudson:
These are additional strings for the New Windows and Tabs section of the new about:preferences section. Designs are here: https://mozilla.invisionapp.com/share/65FPNL9RH4V#/screens/280785532

@flodolo UX has decided these are necessary in order to land the other sections of  about:preferences, however due to the fact we don't have much time left in 60 to implement it there's a good chance we will push landing the feature to 61. In your opinion would it be OK to export these now or is it better to wait?

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -115,6 +115,19 @@ time_label_day={number}d
 # in English, while "Home" should be localized matching the about:preferences
 # sidebar mozilla-central string for the panel that has preferences related to
 # what is shown for the homepage, new windows, and new tabs.
+prefs_new_windows_tabs_header=New Windows and Tabs
+prefs_new_windows_description=Choose what you see when you open your homepage, new windows, and new tabs.
+
+prefs_homepage_choice_label=Homepage and new windows:
+prefs_newtab_choice_label=New tabs:
+prefs_homepage_newtab_choice_1_firefox_home=Firefox Home
+prefs_homepage_newtab_choice_2_custom=Custom URL(s)
+prefs_homepage_newtab_choice_3_blank=Blank Page
+
+prefs_homepage_custom_url_placeholder=Paste a URL
+prefs_homepage_custom_url_current_pages_button=Use Current Pages
+prefs_homepage_custom_url_use_bookmark_button=Use Bookmark
+
 prefs_home_header=Firefox Home Content
 prefs_home_description=Choose what content you want on your Firefox Home screen.
 prefs_restore_defaults_button=Restore Defaults
@@ -155,6 +168,14 @@ settings_pane_snippets_header=Snippets
 settings_pane_snippets_body=Read short and sweet updates from Mozilla about Firefox, internet culture, and the occasional random meme.
 settings_pane_done_button=Done
 settings_pane_topstories_options_sponsored=Show Sponsored Stories
+
+# LOCALIZATION NOTE(extension_notice_*): These are shown in the new Home section in about:preferences for configuring
+# the page shown on about:home and about:newtab. Each notice has a "Manage Extensions" button next to it that links to about:addons
+extension_notice_new_windows_home=An extension is controlling the page you see on new windows and home.
+extension_notice_new_tabs=An extension is controlling the page you see on new tabs.
+extension_notice_new_windows_tabs_home=An extension is controlling the page you see on new tabs, new windows, and home.
+extension_notice_multiple=There are extensions controlling the pages you see on new windows, new tabs, and home.
+extension_notice_manage_extensions_button=Manage Extensions
 
 # LOCALIZATION NOTE (edit_topsites_*): This is shown in the Edit Top Sites modal
 # dialog.


flodolo:
> @flodolo UX has decided these are necessary in order to land the other sections of about:preferences, however due to the fact we don't have much time left in 60 to implement it there's a good chance we will push landing the feature to 61. In your opinion would it be OK to export these now or is it better to wait?

It can creates some churn, but if they're finalized (approved), the sooner they land, the higher are the chances that they're localized by many locales.

locales/en-US/strings.properties
```diff
@@ -115,6 +115,19 @@ time_label_day={number}d
...
+prefs_new_windows_description=Choose what you see when you open your homepage, new windows, and new tabs.
+
+prefs_homepage_choice_label=Homepage and new windows:
+prefs_newtab_choice_label=New tabs:
+prefs_homepage_newtab_choice_1_firefox_home=Firefox Home
+prefs_homepage_newtab_choice_2_custom=Custom URL(s)
+prefs_homepage_newtab_choice_3_blank=Blank Page
+
+prefs_homepage_custom_url_placeholder=Paste a URL
+prefs_homepage_custom_url_current_pages_button=Use Current Pages
```
flodolo: Do we need a plural here? Firefox settings have "Use Current Page" and "Use Current Pages", depending on the number of tabs.

locales/en-US/strings.properties
```diff
@@ -115,6 +115,19 @@ time_label_day={number}d
 # in English, while "Home" should be localized matching the about:preferences
 # sidebar mozilla-central string for the panel that has preferences related to
 # what is shown for the homepage, new windows, and new tabs.
+prefs_new_windows_tabs_header=New Windows and Tabs
+prefs_new_windows_description=Choose what you see when you open your homepage, new windows, and new tabs.
+
+prefs_homepage_choice_label=Homepage and new windows:
+prefs_newtab_choice_label=New tabs:
+prefs_homepage_newtab_choice_1_firefox_home=Firefox Home
+prefs_homepage_newtab_choice_2_custom=Custom URL(s)
```
flodolo: Do we really need the (s)?

locales/en-US/strings.properties
```diff
@@ -156,6 +169,14 @@ settings_pane_snippets_body=Read short and sweet updates from Mozilla about Fire
 settings_pane_done_button=Done
 settings_pane_topstories_options_sponsored=Show Sponsored Stories
 
+# LOCALIZATION NOTE(extension_notice_*): These are shown in the new Home section in about:preferences for configuring
+# the page shown on about:home and about:newtab. Each notice has a "Manage Extensions" button next to it that links to about:addons
+extension_notice_new_windows_home=An extension is controlling the page you see on new windows and home.
+extension_notice_new_tabs=An extension is controlling the page you see on new tabs.
+extension_notice_new_windows_tabs_home=An extension is controlling the page you see on new tabs, new windows, and home.
+extension_notice_multiple=There are extensions controlling the pages you see on new windows, new tabs, and home.
```
flodolo: Why "the pages"?

locales/en-US/strings.properties
k88hudson: In this case, users can enter more than one, so I think it's needed?

locales/en-US/strings.properties
k88hudson: Good point, I see code for that in the existing UI actually, I will update this

locales/en-US/strings.properties
k88hudson: This message is shown if a different extension controls about:home v.s. about:newtab. I'll add a comment to that effect

flodolo: What's the difference between this and `extension_notice_new_windows_tabs_home`, besides the fact that there are multiple extensions? I don't think it's safe to assume that one extension means one page. Also the order changes (tab-window-home, window-tab-home).

locales/en-US/strings.properties
flodolo: We currently don't have [any UI](https://transvision.mozfr.org/?recherche=%28s%29&repo=gecko_strings&sourcelocale=en-US&locale=en-US&search_type=strings_entities) using `(s)` in Firefox (/browser). Maybe just use `Custom URLs…` would work?

k88hudson: Ok, that sounds find to me

locales/en-US/strings.properties
k88hudson: Hm, good point. Let me check this one

locales/en-US/strings.properties
```diff
@@ -115,6 +115,19 @@ time_label_day={number}d
 # in English, while "Home" should be localized matching the about:preferences
 # sidebar mozilla-central string for the panel that has preferences related to
 # what is shown for the homepage, new windows, and new tabs.
+prefs_new_windows_tabs_header=New Windows and Tabs
+prefs_new_windows_description=Choose what you see when you open your homepage, new windows, and new tabs.
+
+prefs_homepage_choice_label=Homepage and new windows:
+prefs_newtab_choice_label=New tabs:
+prefs_homepage_newtab_choice_1_firefox_home=Firefox Home
```
Mardak: The design has a grayed out " (Default)" after this "Firefox Home" text. Last I chatted with design, it sounded like it's okay with it just being regular text. Not sure if you were planning on some other way to add that text? Or is it not desired anymore?

locales/en-US/strings.properties
```diff
@@ -115,6 +115,19 @@ time_label_day={number}d
...
+prefs_homepage_newtab_choice_2_custom=Custom URL(s)
+prefs_homepage_newtab_choice_3_blank=Blank Page
+
+prefs_homepage_custom_url_placeholder=Paste a URL
+prefs_homepage_custom_url_current_pages_button=Use Current Pages
+prefs_homepage_custom_url_use_bookmark_button=Use Bookmark
```
Mardak: The existing string has … as well as the design.
https://searchfox.org/mozilla-central/rev/efce4ceddfbe5fe4e2d74f1aed93894bada9b273/browser/locales/en-US/browser/preferences/preferences.ftl#118-120

Although sounds like we'll just directly reuse that now?

zbraniecki:
@k88hudson - I'm working on the next weave of Preferences code to be transitioned to Fluent. One of the bugs (https://bugzilla.mozilla.org/show_bug.cgi?id=1438375) has a number of strings related to locking down control by extensions.

In this PR I see similar strings. Are you planning to land that into mozilla-central? Is it going to be a separate pool of strings maintained by activity stream? Is there any chance to make sure they're consistent?

k88hudson:
@zbraniecki We decided after discussing these string will land in MC in the patch attached to https://bugzilla.mozilla.org/show_bug.cgi?id=1417155 (with the rest of preferences)

PR #4028, opened 2018-03-07
Fix Bug 1433209 - Remove unused strings

piatra:
Document tracking unused strings https://docs.google.com/document/d/1o7cZ2xmNOJNmyzGxbLzYq-w-AYp8TQyLhU9u-z3XOzs/edit

More deletions to go with [the previous PR that landed](https://github.com/mozilla/activity-stream/commit/1d0e3079ce17b23ac77bd0ec2513205fa9c11dd0).

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -5,9 +5,6 @@ header_highlights=Highlights
 # LOCALIZATION NOTE(header_recommended_by): This is followed by the name
 # of the corresponding content provider.
 header_recommended_by=Recommended by {provider}
-# LOCALIZATION NOTE(header_stories_from): This is followed by a logo of the
-# corresponding content (stories) provider
-header_stories_from=from
 
 # LOCALIZATION NOTE(context_menu_button_sr): This is for screen readers when
 # the context menu button is focused/active. Title is the label or hostname of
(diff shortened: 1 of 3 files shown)


Mardak:
@flodolo how should we approach removing unused strings? Should it be pretty aggressive in that if we don't have some functionality that requires the string in the next version / before we switch to fluent, it should be removed from `strings.properties`? The [unused strings doc](https://docs.google.com/document/d/1o7cZ2xmNOJNmyzGxbLzYq-w-AYp8TQyLhU9u-z3XOzs/edit) from @piatra's script https://github.com/mozilla/activity-stream/pull/4018#issuecomment-370781898 has some more strings that could be removed but aren't currently in this PR:
```ini
header_stories_from=from
type_label_open=Open
type_label_topic=Topic
time_label_less_than_minute=<1m
time_label_minute={number}m
time_label_hour={number}h
time_label_day={number}d
```

I'm not aware of any designs that will use those strings as is. But if we do remove them and need them again, I believe pontoon UI has suggestion features to reuse the previous translation?

flodolo:
> Should it be pretty aggressive in that if we don't have some functionality that requires the string in the next version / before we switch to fluent, it should be removed from strings.properties?

That sounds like a good idea in general. When doing the migration, we would likely catch them since you need to go and assign l10n-IDs to elements, and they wouldn't be there.

> I'm not aware of any designs that will use those strings as is. But if we do remove them and need them again, I believe pontoon UI has suggestion features to reuse the previous translation?

For these specific strings, we're likely going to have matches from other projects as suggestions. I'm all for removing them.

as-pine-proxy:
[Treeherder: b1331d010b85adfd3699070b8e272ef0620d5641](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=b1331d010b85adfd3699070b8e272ef0620d5641)

PR #4035, opened 2018-03-08
Fix Bug 1442383 - Activity stream spends a lot time fetching highlights after bookmark sync

sarracini:
This patch does 2 things as per the conversation in [bug 1442383](https://bugzilla.mozilla.org/show_bug.cgi?id=1442383):
1. it ignores all bookmark added/removed notifications that came from sync by checking `SOURCES.SYNC`
2. when sync is finished and we receive the notification `weave:engine:sync:applied`, it refreshes highlights to check if there were some important bookmarks either added or removed that should show up in activity stream

In addition, there were some tests for `onItemChanged` that were being skipped that are now removed as per [this comment from bug 1440276](https://github.com/mozilla/activity-stream/pull/4005#pullrequestreview-100580633)

### DIFF
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
+ChromeUtils.import("resource://gre/modules/Services.jsm");
 
 const {actionTypes: at} = ChromeUtils.import("resource://activity-stream/common/Actions.jsm", {});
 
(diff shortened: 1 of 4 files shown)


system-addon/lib/HighlightsFeed.jsm
```diff
@@ -26,6 +27,7 @@ ChromeUtils.defineModuleGetter(this, "PageThumbs",
 const HIGHLIGHTS_MAX_LENGTH = 9;
 const MANY_EXTRA_LENGTH = HIGHLIGHTS_MAX_LENGTH * 5 + TOP_SITES_DEFAULT_ROWS * TOP_SITES_MAX_SITES_PER_ROW;
 const SECTION_ID = "highlights";
+const SYNC_BOOKMARKS_FINISHED_EVENT = "weave:engine:sync:finish";
```
mhammond: I should have mentioned this in the bug, but listening for "weave:engine:sync:applied" (and still checking data=="bookmarks") would be slightly more efficient - that means there was almost certainly new bookmarks added by the engine (whereas sync:finish will be fired even if the engine did nothing)

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -4,6 +4,7 @@
 "use strict";
 
 ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
```
k88hudson: Unrelated, but can't we remove this?

sarracini: yeah actually

as-pine-proxy:
[Treeherder: 6aedfe6d1e90968583661d07d14aa0bfe581ee4b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6aedfe6d1e90968583661d07d14aa0bfe581ee4b)

PR #4026, opened 2018-03-07
Bug 1441570 - Add a test to verify that a topsite can be edited from the context menu.

icrisan:
Bug 1441570 - Add a test to verify that a topsite can be edited from the context menu.

PR #4036, opened 2018-03-09
Fix Bug 1410828 - Add a test to verify that the next top site in queue replaces a dismissed top site

icrisan:
Bug 1410828 - Add a test to verify that the next top site in queue replaces a dismissed top site

### DIFF
--- a/system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
+++ b/system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
@@ -48,8 +48,6 @@ test_newtab({
     Assert.equal(defaultTopSitesNumber, 5, "5 top sites are displayed after one of them is dismissed");
   },
   async after() {
-    await SpecialPowers.pushPrefEnv({set: [["browser.newtabpage.blocked", ""]]});
     await new Promise(resolve => NewTabUtils.undoAll(resolve));
-    await setDefaultTopSites();
   }
 });
--- a/system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
+++ b/system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
@@ -4,12 +4,9 @@
 
 "use strict";
 
-/**
- * Test verifies the menu options for a default top site.
- */
-
 test_newtab({
   before: setDefaultTopSites,
+  // Test verifies the menu options for a default top site.
   test: async function defaultTopSites_menuOptions() {
     await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".top-site-icon"),
       "Topsite tippytop icon not found");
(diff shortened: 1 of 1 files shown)


sarracini:
@icrisan i had to revert your PR #4026, since it was failing tests when run in the suite. It seemed to not be clearing the pinned site after it pins it. I think a similar approach as what you're doing here would fix it, or perhaps even just manually unpinning it at the end of the test. Would you be able to open up a PR again for #4026 with a fix for that? Thank you!

system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
```diff
@@ -25,3 +22,34 @@ test_newtab({
...
+
+    // Need to wait for dismiss action.
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".top-sites-list li:first-child a").getAttribute("href") === secondTopSite,
+      "First topsite was dismissed");
+
+    defaultTopSitesNumber = content.document.querySelector(".top-sites-list").querySelectorAll("[class=\"top-site-outer\"]").length;
+    Assert.equal(defaultTopSitesNumber, 5, "5 top sites are displayed after one of them is dismissed");
+  },
+  async after() {
+    await SpecialPowers.pushPrefEnv({set: [["browser.newtabpage.blocked", ""]]});
```
Mardak: This doesn't seem right to call `SpecialPowers` here. There's already a `pushPrefs` globally, but the `before` should actually be using the scoped one instead. The scoped one cleans up after a test runs whereas the global one cleans up only when the test file is completed I believe.

system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
```diff
@@ -25,3 +22,34 @@ test_newtab({
...
+    await new Promise(resolve => NewTabUtils.undoAll(resolve));
+    await setDefaultTopSites();
```
Mardak: As noted in the previous comment, fixing up `setDefaultTopSites` to use the scoped one might avoid needing to do this here. If we do need it here for `--verify`, we should figure out why it's needed to figure out a better way to not need it.

system-addon/test/functional/mochitest/browser_topsites_contextMenu_options.js
Mardak: Sure, but using the scoped one from `before` should clear that pref when the test finishes.

piatra:
Updated.

as-pine-proxy:
[Treeherder: 97e5ad49b32f5c5b09f5432aab367543af01eaac](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=97e5ad49b32f5c5b09f5432aab367543af01eaac)

PR #4037, opened 2018-03-09
Fix Bug 1444518 - Ignore bookmarks added and removed during an import or restore

kitcambridge:
I think I did this right, but please let me know if there's anything that needs changing! 😄

### DIFF
--- a/system-addon/lib/HighlightsFeed.jsm
+++ b/system-addon/lib/HighlightsFeed.jsm
@@ -27,6 +27,8 @@ const HIGHLIGHTS_MAX_LENGTH = 9;
 const MANY_EXTRA_LENGTH = HIGHLIGHTS_MAX_LENGTH * 5 + TOP_SITES_DEFAULT_ROWS * TOP_SITES_MAX_SITES_PER_ROW;
 const SECTION_ID = "highlights";
 const SYNC_BOOKMARKS_FINISHED_EVENT = "weave:engine:sync:applied";
+const BOOKMARKS_RESTORE_SUCCESS_EVENT = "bookmarks-restore-success";
+const BOOKMARKS_RESTORE_FAILED_EVENT = "bookmarks-restore-failed";
 
 this.HighlightsFeed = class HighlightsFeed {
   constructor() {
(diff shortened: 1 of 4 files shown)


sarracini:
Thanks for doing this Kit! Maybe just add a test in PlacesFeed.test.js as well, similar to this: https://github.com/mozilla/activity-stream/blob/master/system-addon/test/unit/lib/PlacesFeed.test.js#L387

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -55,11 +59,18 @@ this.HighlightsFeed = class HighlightsFeed {
...
 
   observe(subject, topic, data) {
-    // When we receive a notification that a sync has happened for bookmarks, refresh highlights
-    if (topic === SYNC_BOOKMARKS_FINISHED_EVENT && data === "bookmarks") {
+    // When we receive a notification that a sync has happened for bookmarks,
+    // or Places finished importing or restoring bookmarks, refresh highlights
+    const manyBookmarksChanged =
+      (topic === SYNC_BOOKMARKS_FINISHED_EVENT && data === "bookmarks") ||
+      topic === BOOKMARKS_RESTORE_SUCCESS_EVENT ||
+      topic === BOOKMARKS_RESTORE_FAILED_EVENT;
```
sarracini: I'm not sure we actually _do_ want to do a refresh if an restore has failed. Maybe you can give some more context about when this event is fired, and why we would want to fetch new highlights in this case?

kitcambridge: That makes sense. If a restore fails, it'll erase all your existing bookmarks, but not add any new ones...so I guess the only issue is that existing highlights will be stale. But, Places tries its best to validate and fix up malformed backups before erasing everything, so do you think it's worth removing `BOOKMARKS_RESTORE_FAILED_EVENT`?

sarracini: Does the event still fire even if Places manages to fix things up and not erase everything?

kitcambridge: Nope, it'll be `SUCCESS` in that case.

sarracini: Ok then I think it's ok to leave the failed even. If it really truly does fail and the bookmarks are all gone then we should reflect that in Activity Stream

as-pine-proxy:
[Treeherder: 4a0ce47e822703a4aa3939f26ea792cc9833f959](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=4a0ce47e822703a4aa3939f26ea792cc9833f959)

PR #4040, opened 2018-03-14
Fix Bug 1445157 - Move isCollapsed pref to IndexedDB

piatra:


system-addon/lib/PrefsFeed.jsm
```diff
@@ -17,12 +18,17 @@ this.PrefsFeed = class PrefsFeed {
...
+    this._storage = new ActivityStreamStorage("sectionPrefs");
   }
 
   // If the any prefs are set to something other than what the prerendered version
   // of AS expects, we can't use it.
-  _setPrerenderPref(name) {
-    this._prefs.set("prerender", PrerenderData.arePrefsValid(pref => this._prefs.get(pref)));
+  async _setPrerenderPref() {
+    if (!this._storage.intialized) {
+      await this._storage.init();
```
k88hudson: initializing the same store in two places could produce a race condition where `storage.init()` is called a second time while the first one is still finishing resolving – maybe use a promise that can be awaited or cache a reference to the initialization as a promise?

system-addon/lib/ActivityStream.jsm
```diff
@@ -261,12 +262,14 @@ this.ActivityStream = class ActivityStream {
...
     this.feeds = FEEDS_CONFIG;
     this._defaultPrefs = new DefaultPrefs(PREFS_CONFIG);
+    this._storage = new ActivityStreamStorage(["sectionPrefs", "snippets"]);
   }
 
   init() {
     try {
       this._updateDynamicPrefs();
       this._defaultPrefs.init();
+      this._storage.init();
```
k88hudson: what's your reasoning for initializing here?

system-addon/lib/ActivityStreamStorage.jsm
```diff
@@ -24,13 +28,25 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
...
+  }
+
   set(key, value) {
     return this.getStore().put(value, key);
   }
 
   _openDatabase() {
     return IndexedDB.open(this.dbName, {version: this.dbVersion}, db => {
-      db.createObjectStore(this.storeName);
+      if (Array.isArray(this.storeName)) {
```
k88hudson: can you maybe document this behaviour in the constructor or here?

system-addon/lib/ActivityStream.jsm
piatra: No specific reason just because there's already an `init` function.

system-addon/lib/PrefsFeed.jsm
piatra: There might not be a good reason for calling `.init` twice. Need to think about it but when the preference feed is initialized we call `_setPrerenderPref` I should just call it once in `PrefsFeed.init`.
If both are required I'll cache the reference to init as you suggested.

k88hudson:
I wonder if we need to migrate users' existing settings, that's probably something we want to do

piatra:
@k88hudson I had to add a new action `MIGRATE_PREFS` because the existing one `UPDATE_SECTION_PREFS` would trigger actions from other feeds as well.
I tested this by setting the initial collapsed pref values in `ActivityStream.jsm` to true.
Section code was kinda tricky: section prefs are handled by the SectionsManager (reads initial values etc) but it doesn't have access to the store and can't dispatch actions.

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -151,7 +154,23 @@ this.TopSitesFeed = class TopSitesFeed {
     }
 
     const links = await this.getLinksWithDefaults();
-    const newAction = {type: at.TOP_SITES_UPDATED, data: links};
+    const newAction = {type: at.TOP_SITES_UPDATED, data: {links}};
+
+    if (!this._storage.intialized) {
+      await this._storage.init();
+      const storedPrefs = await this._storage.get(SECTION_ID) || {};
+      // Migrate prefs from Prefs if IndexedDB is empty
```
k88hudson: I think it might make more sense to check if there is a user-set value in preferences and migrate it rather than checking indexedDB first. This avoids necessary migrations for users who haven't changed their defaults and ensures that a race condition won't accidentally overwrite unmigrated data

system-addon/common/Actions.jsm
```diff
@@ -36,6 +36,7 @@ for (const type of [
   "DIALOG_OPEN",
   "DISABLE_ONBOARDING",
   "INIT",
+  "MIGRATE_PREFS",
```
k88hudson: I don't think this is being used anymore

system-addon/lib/ActivityStreamStorage.jsm
k88hudson: A comment here could be good

as-pine-proxy:
[Treeherder: 009b7c976f2a31af3eb1cfc071e4a3b30beaba2b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=009b7c976f2a31af3eb1cfc071e4a3b30beaba2b)

PR #4041, opened 2018-03-14
Fix Bug 1432672 - Remove sidebar preferences and have gear open about:preferences

sarracini:
1. Removes sidebar 
2. Makes gear open about:preferences


locales/en-US/strings.properties
```diff
@@ -88,25 +88,6 @@ prefs_topstories_sponsored_learn_more=Learn more
 prefs_highlights_description=A selection of sites that you’ve saved or visited
 prefs_snippets_description=Updates from Mozilla and Firefox
 settings_pane_button_label=Customize your New Tab page
-settings_pane_header=New Tab Preferences
```
Mardak: These strings can't be removed yet as we will likely uplift strings to beta

Mardak: Post a comment to https://bugzilla.mozilla.org/show_bug.cgi?id=1445090 with these string ids that should be removed.

sarracini: How about pocket_description, settings_pane_topstories_options_sponsored, settings_pane_highlights_body2 based on Bug 1404890? Can I remove those?

rlr:
Basically, any strings that are used in 60 need to stay another ~6-8 weeks until after the last string export to beta. :+1:

as-pine-proxy:
[Treeherder: 228d9bc2d46e1b87c2d7d64425f3df613a4a1aa4](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=228d9bc2d46e1b87c2d7d64425f3df613a4a1aa4)

PR #4044, opened 2018-03-19
Fix Bug 1402312 - Implement dark theme

rlr:
r? @Mardak / @k88hudson 

In action:

Screenshots:
screen shot 2018-03-19 at 15 56 29-fullpage

screen shot 2018-03-19 at 3 57 22 pm

screen shot 2018-03-19 at 3 57 47 pm

screen shot 2018-03-19 at 4 01 24 pm


system-addon/lib/ThemeFeed.jsm
```diff
@@ -0,0 +1,68 @@
...
+
+  updateTheme(data) {
+    if (data && data.window) {
+      // We only update newtab theme if the theme activated isn't window specific.
+      // We'll be able to do better in the future: see Bug 1444459
+      return;
+    }
+
+    // Jump through some hoops to check if the current theme has light or dark
+    // text. If light, then we enable our dark (background) theme.
```
rlr: NOTE: inspiration for this comes from https://searchfox.org/mozilla-central/source/toolkit/modules/LightweightThemeConsumer.jsm#108-112

system-addon/lib/ThemeFeed.jsm
```diff
@@ -0,0 +1,68 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+ChromeUtils.import("resource://gre/modules/Services.jsm");
+
+const {actionCreators: ac, actionTypes: at} = ChromeUtils.import("resource://activity-stream/common/Actions.jsm", {});
+
+const THEME_UPDATE_EVENT = "lightweight-theme-styling-update";
+
+this.ThemeFeed = class ThemeFeed {
```
rlr: This feed will be obsolete or need to change once https://bugzilla.mozilla.org/show_bug.cgi?id=1444459 is implemented and lands.

system-addon/lib/ThemeFeed.jsm
```diff
@@ -0,0 +1,68 @@
...
+  _parseRGB(colorString) {
+    let rgb = colorString.match(/^rgba?\((\d+), (\d+), (\d+)/);
+    rgb.shift();
+    return rgb.map(x => parseInt(x, 10));
+  }
+
+  updateTheme(data) {
+    if (data && data.window) {
+      // We only update newtab theme if the theme activated isn't window specific.
+      // We'll be able to do better in the future: see Bug 1444459
```
rlr: Fun fact, different windows can be assigned different themes via web extensions, for example. Also private browsing. Out of scope for now :)

Mardak:
Looking!

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
...
+  --newtab-icon-secondary-color: $grey-90-60;
+  --newtab-icon-tertiary-color: $grey-30;
+  --newtab-textbox-border: $grey-90-20;
+  --newtab-textbox-color: inherit;
+  --newtab-button-primary-color: $blue-60;
+  --newtab-button-secondary-color: inherit;
+  --newtab-contextmenu-button-color: $white;
+  --newtab-modal-color: $white;
+  --newtab-overlay-color: $grey-20;
+  --newtab-searchbox-color: $grey-10;
```
Mardak: nit: `--newtab-searchbox-background-color` as there's a bunch of colors ;)

Also, could you make this `$white` here to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1447752 :)

system-addon/content-src/components/Base/_Base.scss
```diff
@@ -93,19 +99,15 @@ main {
...
     top: 15px;
     z-index: 12001;
 
     &:hover {
-      background-color: $background-secondary;
-    }
-
-    &:active {
-      background-color: $background-primary;
+      background-color: $grey-20;
```
Mardak: This should be grey-20 when clicking always?

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
...
+  --newtab-searchbox-border-color: $grey-10-40;
+  --newtab-searchbox-text-color: $grey-10-60;
+  --newtab-searchbox-icon-color: $grey-10-60;
+  --newtab-section-header-text-color: $grey-10-80;
+  --newtab-section-navigation-text-color: $grey-10-80;
+  --newtab-section-active-color: $grey-60;
+  --newtab-section-active-contextmenu-color: $white;
+  --newtab-topsites-label-color: $grey-10-80;
+  --newtab-card-background-color: $grey-90;
+  --newtab-inner-box-shadow-color: $grey-60;
```
Mardak: I'm still looking through this, but are there any conceptually related uses of a color that are used for more than one variable? I don't think I spotted any yet, but for example if `--newtab-text-primary-color` and `--newtab-textbox-color` were the same, there could be a local scss $var to share the two. But then, why would we have two different css --var? (I suppose the css var could exist if say dark shared a color but light had different colors -- but then again, why are they not consistent between the two themes?)

rlr:
Basically, I went through every color spec'd out in the dark theme spec that needed to be changed and started breaking them out into variables. And then I tried to consolidate into as few variables as possible in some cases I had to get UX (Amy) to nod some changes/compromises to save a variable and keep more generic. But it's possible we can reduce it more?

See https://mozilla.invisionapp.com/share/H4FW3RYFXVK#/screens/279681634_New_Tab_Dark_Theme_Spec and:

dark theme spec - outstanding items

rlr:
Afterwards, I wondered if I should've kept all the scss `$vars` and just make them point to `var(--vars)`. I could switch back if it makes it easier to read or maintain.

system-addon/content-src/components/Base/_Base.scss
rlr: This is the hover on the prefs gear icon. For overall consistency, this should probably match `--newtab-section-active-color` so I should rename that.

rlr:
Also, there are cases that two variables are the same in the default theme but different in the dark theme, and vice-versa.

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
...
+  --newtab-icon-secondary-color: $grey-90-60;
+  --newtab-icon-tertiary-color: $grey-30;
+  --newtab-textbox-border: $grey-90-20;
+  --newtab-textbox-color: inherit;
+  --newtab-active-element-color: $grey-20-60;
+  --newtab-button-primary-color: $blue-60;
+  --newtab-button-secondary-color: inherit;
+  --newtab-contextmenu-button-color: $white;
+  --newtab-modal-color: $white;
+  --newtab-overlay-color: $grey-20;
```
rlr: I guess up to here, the variables are pretty generic. And all the ones below are pretty specific to different sections.

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
...
+  --newtab-searchbox-icon-color: $grey-10-60;
+  --newtab-section-header-text-color: $grey-10-80;
+  --newtab-section-navigation-text-color: $grey-10-80;
+  --newtab-section-active-contextmenu-color: $white;
+  --newtab-topsites-label-color: $grey-10-80;
+  --newtab-card-background-color: $grey-90;
+  --newtab-inner-box-shadow-color: $grey-60;
+}
+
+// scss variables related to the theme.
```
rlr: I moved these variables below into this file because some depend on the CSS variables defined above. But I guess they can go back to _variables.scss even though that's imported first because the vars don't get evaluated until runtime anyway.

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
...
+$border-primary: 1px solid var(--newtab-border-primary-color);
+$border-secondary: 1px solid var(--newtab-border-secondary-color);
+$input-primary: $blue-60;
+$input-secondary: transparent;
+$shadow-primary: 0 0 0 5px $grey-30;
```
Mardak: I would have expected any shade of black to be different for light vs dark. Default background is $grey-10, so this $grey-30 shadow is 20 degrees away. To maintain the same amount of contrast on the dark theme's $grey-80 background, I would have expected $grey-60 for the shadow:

60 screen shot 2018-03-23 at 4 40 01 pm

30 screen shot 2018-03-23 at 4 40 22 pm

@bryanbell ?

rlr: ah, Amy (is she on github?) said the box shadow on hover could stay the same.

screen shot 2018-03-23 at 7 48 51 pm

Mardak: bryan said "Top is better by far" with +1 from aaron

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Dark theme
+.outer-wrapper.dark-theme {
+  --newtab-background-color: $grey-80;
+  --newtab-text-primary-color: $grey-10;
+  --newtab-text-secondary-color: $grey-10-40;
+  --newtab-text-conditional-color: $grey-10;
+  --newtab-link-primary-color: $blue-40;
+  --newtab-link-secondary-color: $blue-40;
```
Mardak: The "secondary" link color is currently used for Pocket stuff, and I'm pretty sure it's teal for default theme specifically because Pocket uses teal for its links. Should get design to confirm that it was intentional to make Pocket links not teal anymore. And if so, why keep it teal for default theme?

system-addon/content-src/styles/_activity-stream.scss
```diff
@@ -22,12 +21,7 @@ h2 {
 }
 
 a {
-  color: $link-primary;
   text-decoration: none;
-
-  &:hover {
-    color: $link-secondary;
```
Mardak: Removing the :hover -secondary is right, so we can get rid of this overriding color too:
https://github.com/mozilla/activity-stream/blob/837cea5f88ceac381a95b5c6bdb0f1e3106d2ad2/system-addon/content-src/components/TopSites/_TopSites.scss#L364-L366

I seem to recall design wanting links to underline on hover.. but I guess leave it this way for now.

system-addon/content-src/styles/_activity-stream.scss
```diff
@@ -22,12 +21,7 @@ h2 {
 }
 
 a {
-  color: $link-primary;
```
Mardak: Why did this need to move into Base? I suppose technically there could be links under `body` that aren't under `.outer-wrapper` but there's other things in this file that are using `var()`…?

system-addon/content-src/styles/_activity-stream.scss
```diff
@@ -84,7 +78,7 @@ a {
   padding: 15px 25px 0;
 
   button {
-    background-color: $input-secondary;
+    background: var(--newtab-button-secondary-color);
```
Mardak: Was there a need to make this `background` and not more scoped to `background-color`? It would probably be better consistency / expectation to use `background-color` wherever we use a `var(…color)`

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
```diff
@@ -1,5 +1,5 @@
 .context-menu {
-  background: $background-primary;
+  background: var(--newtab-background-color);
   border-radius: $context-menu-border-radius;
   box-shadow: $context-menu-shadow;
```
Mardak: This context menu box-shadow / border should probably be a themed value. Not entirely sure what it should be.. At least the second 1px shadow should be something closer to --newtab-border-primary-color.

--newtab-border-primary-color screen shot 2018-03-23 at 5 46 47 pm

current shadow screen shot 2018-03-23 at 5 42 44 pm

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -22,9 +21,11 @@
   width: 100%;
 
   input {
-    border: 0;
+    background: var(--newtab-searchbox-background-color);
+    border: solid 1px var(--newtab-searchbox-border-color);
     border-radius: $search-border-radius;
     box-shadow: $shadow-secondary, 0 0 0 1px $black-15;
```
Mardak: There's various blacks here that probably need to be themed. E.g., search button hover, etc.

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,80 @@
+// Default theme
+.outer-wrapper {
```
Mardak: We can set `:not(.dark-theme)` to avoid having both styles apply with the latter overriding for dark theme. Not sure if there's a good way to make sure both lists define the same variables though…

system-addon/content-src/styles/_activity-stream.scss
rlr: no need, just saving bytes 😝 . kind of my personal preference but dont really care either way.

system-addon/content-src/styles/_theme.scss
rlr: is there a performance gain to doing that? or just a little less clutter in dev tools inspector?

system-addon/content-src/styles/_activity-stream.scss
rlr: I moved to `.outer-wrapper` because that's the first (topmost?) node that we control from react. I can't (cleanly) set classes or css vars to body or anything above `.outer-wrapper`. I positive side effect of moving the themed stuff to be sandboxed in `.outer-wrapper` is that this doesn't break snippets styles.

rlr:
I just got off vidyo with @bryanbell @aaronrbenson +Amy. Those were the changes we came up with ^^ 0253625 

* lighter teal for better contrast copy pasted off pocket website
* darker separators for context menu
* leave the lack of visible border and drop shadow as is on context menu
* no visible hover state for search box (this is consistent with awesome bar in dark theme)
* themeify the searchbox focus outline. use blue-40 in dark theme
* tweaked placeholder and empty section borders

And they want it in their hands ASAP. I guess we can further tweak it in production^Wnightly if necessary.

rlr:
oops i hit the wrong button. lol

system-addon/content-src/components/Base/_Base.scss
```diff
@@ -1,4 +1,6 @@
 .outer-wrapper {
+  background: var(--newtab-background-color);
```
Mardak: `background` vs `background-color` does have behavior differences in that setting `background: <color>` will cause other properties of `background-*` to be overridden. We've run into bugs in the past (I don't recall exactly which) where unintended background properties were being set… but potentially there are the opposite bugs where not setting the other `background-*` properties can cause problems…

system-addon/content-src/styles/_activity-stream.scss
Mardak: I haven't actually used css variables, but looking at m-c usage/declarations, they seem to typically be under a `:root { --var: … }` selector.

@k88hudson do you have any preference or know of desired usage patterns?

system-addon/content-src/styles/_activity-stream.scss
```diff
@@ -99,13 +93,14 @@ a {
     }
 
     &.dismiss {
+      background-color: transparent;
```
Mardak: Why not use `$input-secondary` here?

system-addon/lib/PrefsFeed.jsm
```diff
@@ -32,11 +33,13 @@ this.PrefsFeed = class PrefsFeed {
...
-  // of AS expects, we can't use it.
-  async _setPrerenderPref() {
+  // If any prefs or the theme are set to something other than what the
+  // prerendered version of AS expects, we can't use it.
+  async _setPrerenderPref(theme) {
     const indexedDBPrefs = await this._storage.getAll();
-    this._prefs.set("prerender", PrerenderData.arePrefsValid(pref => this._prefs.get(pref), indexedDBPrefs));
+    const prefsAreValid = PrerenderData.arePrefsValid(pref => this._prefs.get(pref), indexedDBPrefs);
+    const themeIsDefault = (theme || this.store.getState().Theme).className === INITIAL_STATE.Theme.className;
+    this._prefs.set("prerender", prefsAreValid && themeIsDefault);
```
Mardak: This means dark theme is never prerendered, so potentially the page will flash all white before switching to dark?

@k88hudson any suggestions here? I suppose we could maybe have `aboutNewTabService` point to a default vs dark theme prerendered version? Although that seems a bit excessive to set a single class name or not.

@rlr also, how are theme changes going to be handled going forwards (for the later in 61 vs later)?

system-addon/common/Reducers.jsm
```diff
@@ -31,6 +31,7 @@ const INITIAL_STATE = {
     initialized: false,
     values: {}
   },
+  Theme: {className: ""},
```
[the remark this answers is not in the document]
rlr: I *think* the way it will probably work is there will be a list of css variables here that we expose to themes + web extensions so they can override. So yeah, I envision more properties added here.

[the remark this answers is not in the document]
rlr: And I think we'll probably keep the feed around but it might be different? I haven't thought much about how we would make newtab themeable on a window by window basis. It depends on the implementation details of bug 1444459 I guess?

system-addon/content-src/styles/_activity-stream.scss
rlr: nice catch 👍

system-addon/lib/PrefsFeed.jsm
rlr: > also, how are theme changes going to be handled going forwards (for the later in 61 vs later)?

You mean the design process or something else?

> This means dark theme is never prerendered, so potentially the page will flash all white before switching to dark?

There are so many things that invalidate prerender now that I'm not sure it is even useful other than first run.

rlr:
Basically, I was thinking of this as just a first step towards exposing our theme variables to the browser theme system. So, we'd have our default light + dark css variables and those can get overriden by individual themes at the browser level in the future. I was planning on prototyping that but haven't gotten around to it.

system-addon/content-src/styles/_activity-stream.scss
Mardak: mconley points out that there is some movement to more appropriately scope css variables for performance reasons, e.g., https://bugzilla.mozilla.org/show_bug.cgi?id=1435122

system-addon/content-src/styles/_theme.scss
Mardak: John-Galt seems to suggest it probably won't make much of a difference performance-wise. So for developer / QA checking, I think we can stick **without** `:not` as the default theme `var`s will always be set, so if we accidentally forget to set it in `.dark`, it should be more obvious when a light theme's color shows up incorrectly.

system-addon/lib/PrefsFeed.jsm
Mardak: I was referring to the not needing ThemeFeed. Sounds like going forwards, we won't need to invalidate prerendering with dark theme as each content page will be getting an async message anyway to set the theme.

Hopefully preloading will avoid most of the potential flashing of wrong-theme-styling.

So with that, maybe we don't need to complicate prerendering related stuff with this PR?

k88hudson:
I'm noticing the search box text is not visible / the dark theme isn't applied to it properly
image

k88hudson:
Extra space without the bg color at bottom if the page is <100% height of the window (easier to see if you collapse some stuff)
image

system-addon/lib/PrefsFeed.jsm
k88hudson: Invalidation is probably the right call, yep. 

I'm definitely noticing some flashing, and even prerendering aside if the message doesn't come through immediately you might see some flashing of a white background

I guess you could have a *single* variant (like an unprerendered, dark background version of the html) for dark theme... but I agree, preloading will take care of a lot of the problem.

Maybe we should check the delay on a slower/reference machine to see if it's acceptable

rlr:
^ fixed text in search dropdown widget. theming that is out of scope as it isnt themed elsewhere yet:

screen shot 2018-03-27 at 12 25 43 pm

k88hudson:
@Mardak re:css variable patterns (I can't reply on that comment for some reason), I agree that placing global css variables (like colors) under `:root` is the way to go, i.e.

```css
:root {
  --blue-50: #123456
}
```

rlr:
amazing catch with the bg color, yikes. fixord ^

nt1m:
@rlr /re the search popup, this is going to be themed starting tomorrow's Nightly by https://bugzilla.mozilla.org/show_bug.cgi?id=1417883

rlr:
@nt1m oh, awesome! 💃

nt1m:
@rlr Just my personal opinion about this, I think it would be better to invest in the theming API feature (rather than having a specific dark theme fix), since the dark theme can leverage it very easy afterwards.

nt1m:
@rlr Though the work done here with the CSS variables will be super useful for students :)

rlr:
> @rlr Just my personal opinion about this, I think it would be better to invest in the theming API feature (rather than having a specific dark theme fix), since the dark theme can leverage it very easy afterwards.

I agree that the theming API is the way to go long term. But that is blocked on other work and UX/product decided this would be a quick win for dark theme users at it is requested a lot. And most of the tedious work here (breaking out the theme-able variables) had to be done anyway.

Mardak:
> /re the search popup this is going to be themed starting tomorrow's Nightly by https://bugzilla.mozilla.org/show_bug.cgi?id=1417883

@nt1m I just built latest m-c, and I'm not sure if we're referring to the same search popup. I do see that address bar popup is dark themed, but the in-content one styled by https://searchfox.org/mozilla-central/source/browser/base/content/contentSearchUI.css seems unchanged.

nt1m:
@Mardak I was talking about the main browser searchbar, not the one in about:newtab.

Mardak:
Oh, I see. @rlr, activity stream is the only consumer of contentSearchUI now, so most likely we'll need to update it. Or more likely set the appropriate styling for default/dark from our css.

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -84,3 +85,8 @@
     transform: translateY(2px);
   }
 }
+
+// The search dropdown isn't themed (yet), so we need to keep dark text.
+#searchSuggestionTable {  // sass-lint:disable-line no-ids
```
Mardak: There's already a `.contentSearchSuggestionTable` just a few lines higher that we can reuse.

```diff
@@ -79,7 +80,25 @@
   }
+}
 
+@at-root {
   // Adjust the style of the contentSearchUI-generated table
-  .contentSearchSuggestionTable { // sass-lint:disable-line class-name-format
+  // sass-lint:disable-block class-name-format
+  .contentSearchSuggestionTable {
+    background-color: var(--newtab-searchbox-background-color);
     border: 0;
-    transform: translateY(2px);
+    transform: translateY(3px);
+
+    .contentSearchHeader {
+      background-color: var(--newtab-background-color);
+      color: var(--newtab-text-secondary-color);
+    }
+
+    .contentSearchHeader,
+    .contentSearchSuggestionsList {
+      border-color: var(--newtab-border-primary-color);
+    }
+
+    .contentSearchSearchWithHeaderSearchText {
+      color: var(--newtab-text-primary-color);
+    }
   }
```
screen shot 2018-03-27 at 9 30 15 pm
screen shot 2018-03-27 at 9 30 24 pm

Mardak: Oh, and I wonder if we should name the `--var`s as `--newtab-search-…` (no "box") ?

system-addon/content-src/components/Sections/_Sections.scss
```diff
@@ -39,14 +39,14 @@
...
-        fill: $fill-secondary;
+        fill: var(--newtab-icon-secondary-color);
         height: 50px;
         margin: 0 auto;
         width: 50px;
       }
 
       .empty-state-message {
-        color: $text-secondary;
+        color: var(--newtab-text-primary-color);
```
Mardak: For default theme, the text is no longer faded, but I think we don't want it too light anyway for contrast…

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,84 @@
...
+  --newtab-topsites-label-color: $grey-10-80;
+  --newtab-card-background-color: $grey-90;
+  --newtab-card-active-outline-color: $grey-60;
+  --newtab-inner-box-shadow-color: $grey-10-20;
+}
+
+// scss variables related to the theme.
+$border-primary: 1px solid var(--newtab-border-primary-color);
+$border-secondary: 1px solid var(--newtab-border-secondary-color);
+$input-primary: $blue-60;
```
Mardak: Similar to the comment about getting rid of `$input-secondary` as there's a similar `--var`, get rid of this scss var and update the last usage in _ContextMenu to use the css var.

system-addon/lib/PrefsFeed.jsm
Mardak: Maybe an alternative to invalidating prerendering or extra prerendering dark theme.. What if we used `localStorage` or something in the page to add `.dark-theme` to `<body>` as the page loads? The storage gets updated when `Theme.className` arrives so the next load of the page should synchronously render the appropriate background color. Although this would prevent different themes per window…

I guess we'll just keep the invalidation for now…

system-addon/lib/ThemeFeed.jsm
```diff
@@ -0,0 +1,68 @@
...
+const {actionCreators: ac, actionTypes: at} = ChromeUtils.import("resource://activity-stream/common/Actions.jsm", {});
+
+const THEME_UPDATE_EVENT = "lightweight-theme-styling-update";
+
+this.ThemeFeed = class ThemeFeed {
+  init(LightweightThemeManager) {
+    Services.obs.addObserver(this, THEME_UPDATE_EVENT);
+
+    let temp = {LightweightThemeManager};
+    ChromeUtils.import("resource://gre/modules/LightweightThemeManager.jsm", temp);
```
Mardak: This seems odd… I guess for injecting custom LightweightThemeManager for testing? That's what the injector is for? Although in this case, we would just override the global if this was imported normally?

system-addon/test/unit/lib/ThemeFeed.test.js
```diff
@@ -0,0 +1,102 @@
...
+  let ThemeFeed;
+  let THEME_UPDATE_EVENT;
+  let feed;
+  let globals;
+  let sandbox;
+
+  beforeEach(() => {
+    globals = new GlobalOverrider();
+    sandbox = globals.sandbox;
+    ({ThemeFeed, THEME_UPDATE_EVENT} = injector({}));
```
Mardak: If nothing is being injected, why not just import the module normally?

system-addon/content-src/components/ContextMenu/_ContextMenu.scss
```diff
@@ -1,5 +1,5 @@
 .context-menu {
-  background: $background-primary;
+  background: var(--newtab-background-color);
```
Mardak: Do we also need to set the text color for the context menu items? Dark theme doesn't have icons that match up with the text color anymore.. unless it's hovered:
image
image

system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
```diff
@@ -91,12 +91,13 @@
...
     a {
-      color: $link-secondary;
+      color: var(--newtab-link-secondary-color);
+      font-weight: bold;
       padding-left: 3px;
     }
 
     button {
-      background: $grey-10;
+      background: $input-secondary;
```
Mardak: Looks like this should be `var(--newtab-button-secondary-color)` instead of `transparent`
image
image

system-addon/content-src/components/Search/_Search.scss
Mardak: (I used `@at-root` mostly to create a new block for the sass disable.)

Mardak:
I'm guessing this missing background is different from what kate pointed out earlier as I think you fixed that. Scrolling down the page shows `<body>`'s white background:
screen shot 2018-03-28 at 12 59 27 am

system-addon/content-src/components/Search/_Search.scss
rlr: awesome, thanks!. and --newtab-search- sounds good. fixing

system-addon/content-src/components/Base/_Base.scss
```diff
@@ -93,19 +99,17 @@ main {
...
     top: 15px;
     z-index: 12001;
 
-    &:hover {
-      background-color: $background-secondary;
-    }
-
+    &:hover,
+    &:focus,
     &:active {
```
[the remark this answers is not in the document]
rlr: currently, when you click the background goes away. I guess we can keep that behavior. I dont think it really makes a difference either way since it's behavior is more like a link than a button.

system-addon/content-src/components/Sections/_Sections.scss
rlr: so, leave it or?

system-addon/content-src/styles/_theme.scss
rlr: to confirm, we'd be using `--newtab-link-primary-color` for this correct?

system-addon/lib/ThemeFeed.jsm
rlr: I was assuming whoever did this knew what they were doing: https://searchfox.org/mozilla-central/source/toolkit/modules/LightweightThemeConsumer.jsm#29-30

But I could make it global if there is no reason to have done that.

system-addon/test/unit/lib/ThemeFeed.test.js
rlr: In case you asked me to make `LightweightThemeManager` global? 😉

rlr:
I guess I'll have to fix merge conflicts if this is ready to land before pull #4058

system-addon/content-src/styles/_theme.scss
```diff
@@ -0,0 +1,83 @@
...
+  --newtab-modal-color: $white;
+  --newtab-overlay-color: $grey-20;
+  --newtab-search-background-color: $white;
+  --newtab-search-border-color: transparent;
+  --newtab-search-focus-border-color: $blue-50;
+  --newtab-search-text-color: inherit;
+  --newtab-search-icon-color: $grey-90-40;
+  --newtab-section-header-text-color: $grey-50;
+  --newtab-section-navigation-text-color: $grey-50;
+  --newtab-section-navigation-link-color: $blue-40;
```
rlr: this was missing a dark theme value. but it was actually unused! removing

system-addon/content-src/styles/_theme.scss
Mardak: Not sure… I suppose it's a choice between $blue-40 link-primary vs $blue-60 button-primary.

Although, I wonder if it's intended that they're different for dark:
image

rlr:
Snippets/onboarding was out of scope for this I thought. @k88hudson had mentioned something something or the other.

system-addon/content-src/styles/_theme.scss
rlr: A problem with having the variables under `.outer-wrapper` is that snippets/onboarding won't be able to see them because they live outside of this in the DOM tree. So, we probably wan't to assign them to `:root` at that point, but then we will need a new way to override them for each theme because we can't set a `.dark-theme` class to `:root` can we? I guess we can cross that bridge when we get there. Maybe https://bugzilla.mozilla.org/show_bug.cgi?id=1444459 will be fixed soonish and we can just wait for that.

rlr:
Filed followup bugs for onboarding and snippets:
https://bugzilla.mozilla.org/show_bug.cgi?id=1449651
https://bugzilla.mozilla.org/show_bug.cgi?id=1449653

system-addon/content-src/styles/_activity-stream.scss
```diff
@@ -99,13 +96,14 @@ a {
...
       border: 0;
       padding: 0;
       text-decoration: underline;
     }
 
     &.done {
-      background: $input-primary;
+      background-color: var(--newtab-button-primary-color);
       border: solid 1px $blue-60;
```
Mardak: not sure why I didn't use $input-primary instead of this $blue-60 for border.. but it should probably just be var(--newtab-button-primary-color) now

Mardak:
From the feedback on slack, given that we'll make all dark's blues blue-40 and default's blue-60, should we have a different var name:
```diff
+  --newtab-button-primary-color: $blue-60;
+  --newtab-link-primary-color: $blue-60;
+  --newtab-search-focus-border-color: $blue-50;

+  --newtab-button-primary-color: $blue-60;
+  --newtab-link-primary-color: $blue-40;
+  --newtab-search-focus-border-color: $blue-40;
```

Mardak:
Although keeping individual css variables allows them to be individually overridden by a theme if desired…

rlr:
Having one var for all the blues sounds compelling. But yeah maybe they should remain separate because I can't think a name for a variable that overrides them. `--theme-blue`? but then what if I wanted pink 😆 . `--theme-action-color` 🤔

Mardak:
`--newtab-interaction-primary-color` ? ;)

rlr:
I guess we can still get stingier with variables when we start to expose them to theming api.

system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
Mardak: @rlr looks like this was skipped? The disclaimer's button is now just `transparent` instead of a var

system-addon/content-src/components/Base/Base.jsx
```diff
@@ -83,13 +83,13 @@ export class BaseContent extends React.PureComponent {
...
     const {props} = this;
-    const {App} = props;
+    const {App, Theme} = props;
     const {initialized} = App;
     const prefs = props.Prefs.values;
 
     const shouldBeFixedToTop = PrerenderData.arePrefsValid(name => prefs[name]);
 
-    const outerClassName = `outer-wrapper${shouldBeFixedToTop ? " fixed-to-top" : ""} ${prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"}`;
+    const outerClassName = `outer-wrapper ${Theme.className}${shouldBeFixedToTop ? " fixed-to-top" : ""} ${prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"}`;
```
Mardak: This ends up causing all the prerendered files to have an extra space:
```
"outer-wrapper  fixed-to-top…"
```

PR #4045, opened 2018-03-19
Fix Bug 1446395 - Assign context menus in Highlights on a per card type basis

sarracini:
Instead of assigning context menus to each card for an entire section type, this allows you to specify context menu options for each card type *only* for highlights.

### DIFF
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -61,7 +61,6 @@ const BUILT_IN_SECTIONS = {
     icon: "highlights",
     title: {id: "header_highlights"},
     maxRows: 3,
-    availableLinkMenuOptions: ["CheckBookmarkOrArchive", "CheckSavedToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl", "CheckDeleteHistoryOrEmpty"],
     emptyState: {
       message: {id: "highlights_empty_state"},
       icon: "highlights"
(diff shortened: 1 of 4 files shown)


system-addon/lib/SectionsManager.jsm
```diff
@@ -185,12 +189,46 @@ const SectionsManager = {
...
+   * Sets each card in highlights' context menu options based on the card's type.
+   * (See types.js for a list of types)
+   *
+   * @param rows section rows containing a type for each card
+   */
+  _addCardTypeLinkMenuOptions(rows) {
+    for (let card of rows) {
+      switch (card.type) {
+        case "history":
+          card.contextMenuOptions = CONTEXT_MENU_OPTIONS_FOR_HIGHLIGHT_TYPES.history;
```
k88hudson: Can't this just be:
```js
for (let card of rows) {
  card.contextMenuOptions = CONTEXT_MENU_OPTIONS_FOR_HIGHLIGHT_TYPES[card.type];
}
```

You may also want to throw an error saying something like `No context menu for highlight type ${card.type} is configured` if you're not providing a default case.

system-addon/lib/SectionsManager.jsm
```diff
@@ -185,12 +189,46 @@ const SectionsManager = {
...
+          break;
+        case "bookmark":
+          card.contextMenuOptions = CONTEXT_MENU_OPTIONS_FOR_HIGHLIGHT_TYPES.bookmark;
+          break;
+        case "pocket":
+          card.contextMenuOptions = CONTEXT_MENU_OPTIONS_FOR_HIGHLIGHT_TYPES.pocket;
+          break;
+      }
+
+      // Remove any options that shouldn't be there
```
k88hudson: Why would this happen?

system-addon/lib/SectionsManager.jsm
```diff
@@ -61,18 +61,22 @@ const BUILT_IN_SECTIONS = {
...
-    availableLinkMenuOptions: ["CheckBookmarkOrArchive", "CheckSavedToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl", "CheckDeleteHistoryOrEmpty"],
     emptyState: {
       message: {id: "highlights_empty_state"},
       icon: "highlights"
     },
     shouldSendImpressionStats: false
   })
 };
+const CONTEXT_MENU_OPTIONS_FOR_HIGHLIGHT_TYPES = {
+  history: ["CheckBookmark", "CheckSavedToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl", "DeleteUrl"],
```
k88hudson: personally I'd like a test ensuring all of the possible highlight types that get produced by highlights feed match up with these to handle possible future breakages... but that might be hard to do?

system-addon/lib/SectionsManager.jsm
sarracini: If the pocket extension is not enabled we have to remove the 'CheckSavedToPocket' option in each card

system-addon/lib/SectionsManager.jsm
k88hudson: although that's sounding more like an integration test

sarracini: could write a mochitest for it as a follow up?

sarracini:
Made review changes + added a test (and a better comment) about the filtering bit

as-pine-proxy:
[Treeherder: 7e89bb89eacd09517e5e58ed2ea3d5a4d2aac523](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7e89bb89eacd09517e5e58ed2ea3d5a4d2aac523)

PR #4046, opened 2018-03-20
Fix Bug 1445158 - Implement persistence for isEnabled using IndexedDB

piatra:
Still in progress. Having to work around the fact that TopSites are not a section and about:preferences would work with a combination of preferences and indexeddb storage which might mean there is some prerequisite work needed before this can land.


piatra:
It's not certain this is currently needed: 
* Needs some amount of refactoring to get TopSites into Sections
  * the move into `about:preferences` made this PR more complicated
* Would require managing both IndexedDB stored preferences as well as regular prefs in `about:preference`
* Moving a pref to IndexedDB means losing the ability to sync across devices

Moving this work to backlog.

PR #4047, opened 2018-03-20
Fix Bug 1408026 - Render layers for the preloaded newtab after hydration

mikeconley:
This is still WIP, because I'd like to know if:

1. This is the right place to handle this action, and
2. Whether or not there's a way I can only do this for preloaded tabs

@sarracini, if you have a second, can you help me with those questions?

### DIFF
--- a/system-addon/lib/ActivityStreamMessageChannel.jsm
+++ b/system-addon/lib/ActivityStreamMessageChannel.jsm
@@ -153,11 +153,23 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
   getPreloadedBrowser() {
     let preloadedPorts = [];
     for (let port of this.channel.messagePorts) {
-      if (port.browser.getAttribute("preloadedState") === "preloaded") {
+      if (this.isPreloadedBrowser(port.browser)) {
         preloadedPorts.push(port);
       }
     }
     return preloadedPorts.length ? preloadedPorts : null;
+  }
+
+  /**
+   * isPreloadedBrowser - Returns true if the passed browser has been preloaded
+   *                      for faster rendering of new tabs.
+   *
+   * @param {<browser>} A <browser> to check.
+   * @return {bool} True if the browser is preloaded.
+   *                      if there aren't any preloaded browsers
+   */
+  isPreloadedBrowser(browser) {
+    return browser.getAttribute("preloadedState") === "preloaded";
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mikeconley:
Thanks to @sarracini, I think we've found a place to do this that makes sense.

So, what do I need to do to make your CI happy?

system-addon/lib/ActivityStreamMessageChannel.jsm
```diff
@@ -220,6 +232,15 @@ this.ActivityStreamMessageChannel = class ActivityStreamMessageChannel {
    * @param  {obj} msg The messsage from a page that was just loaded
    */
   onNewTabLoad(msg) {
+    let {browser} = msg.target;
+    if (this.isPreloadedBrowser(browser)) {
```
sarracini: Do you need to do this across windows as well? The `getPreloadedBrowser` function returns an array of all the preloaded browsers across all windows, and you can iterate over them to add `renderLayers` to each

mikeconley: I don't think so, no - I think I only ever want to do this on a preloaded tab that just loaded. This is less about the _contents_ of the preloaded tab, and more about the preloadability of that one preloaded tab, if that makes sense. Once we receive the message that it has loaded, I'm making the parent say, "oh, okay, so we can paint and upload layers for this one tab then, since it just loaded", which increases the likelihood that when we do want to switch to it, it'll be more or less immediate.

sarracini: Ah makes sense. Thanks!

sarracini:
So for CI changes, all you need to do is add the right mocks to the right tests:
here: https://github.com/mozilla/activity-stream/blob/master/system-addon/test/unit/lib/ActivityStreamMessageChannel.test.js#L94, here: https://github.com/mozilla/activity-stream/blob/master/system-addon/test/unit/lib/ActivityStreamMessageChannel.test.js#L100, here: https://github.com/mozilla/activity-stream/blob/master/system-addon/test/unit/lib/ActivityStreamMessageChannel.test.js#L109, and here: https://github.com/mozilla/activity-stream/blob/master/system-addon/test/unit/lib/ActivityStreamMessageChannel.test.js#L215

add `browser: {getAttribute() {}}` to the objects and it should fix the test failures. 
In addition you'll have to write a test to give line coverage to the new code:
```js
it("should add renderLayers to preloaded browsers", () => {
  RPmessagePorts.push({loaded: true, portID: "foo", browser: {getAttribute: () => "preloaded"}});
  mm.simulateMessagesForExistingTabs();
  assert.equal(RPmessagePorts[0].browser.renderLayers, true);
});
```

sarracini:
You can even add a test to make sure it doesn't add it to non-preloaded browsers:
```js
it("should not add renderLayers to non-preloaded browsers", () => {
  RPmessagePorts.push({loaded: true, portID: "foo", browser: {getAttribute: () => "consumed"}});
  mm.simulateMessagesForExistingTabs();
  assert.isUndefined(RPmessagePorts[0].browser.renderLayers);
});
```

as-pine-proxy:
[Treeherder: 436f3fe9ae74a4b575d6d68a012af662501ee07b](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=436f3fe9ae74a4b575d6d68a012af662501ee07b)

PR #4048, opened 2018-03-21
Fix Bug 1431469 - expose previous session end (UNINIT) to snippets

rlr:
I was wondering if we really need to pass the current session start? Snippets could just use Date.now() and that should be good enough?

r? @k88hudson 
cc: @piatra

### DIFF
--- a/system-addon/lib/SnippetsFeed.jsm
+++ b/system-addon/lib/SnippetsFeed.jsm
@@ -155,7 +155,8 @@ this.SnippetsFeed = class SnippetsFeed {
       defaultBrowser: this.isDefaultBrowser(),
       isDevtoolsUser: this.isDevtoolsUser(),
       addonInfo: await this.getAddonInfo(),
-      blockList: await this._getBlockList() || []
+      blockList: await this._getBlockList() || [],
+      previousSessionEnd: this._previousSessionEnd
     };
     this._dispatchChanges(data);
   }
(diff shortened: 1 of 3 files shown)


system-addon/lib/SnippetsFeed.jsm
```diff
@@ -192,6 +196,17 @@ this.SnippetsFeed = class SnippetsFeed {
     browser.loadURI(url);
   }
 
+  _onSessionStart() {
```
k88hudson: Yeah, I think you're right – let's remove this and just provide session end

system-addon/lib/SnippetsFeed.jsm
```diff
@@ -213,6 +228,12 @@ this.SnippetsFeed = class SnippetsFeed {
       case at.TOTAL_BOOKMARKS_REQUEST:
         this.getTotalBookmarksCount(action._target.browser);
         break;
+      case at.NEW_TAB_INIT:
+        this._onSessionStart();
+        break;
+      case at.NEW_TAB_UNLOAD:
```
k88hudson: I'm wondering if maybe we should be look at a more general *browser* session, like maybe the `UNINIT` event, since `NEW_TAB_UNLOAD` happens frequently within the same browser session

rlr: oh yeah! that's great. should we keep the currentSessionStart or dump it? or set it to `INIT` instead of `NEW_TAB_INIT`?

rlr:
@k88hudson fixed ^^

k88hudson:
Thanks!

as-pine-proxy:
[Treeherder: 7d6037d3d454ea8d24dc451420ca9bf51d5bf438](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7d6037d3d454ea8d24dc451420ca9bf51d5bf438)

PR #4050, opened 2018-03-22
Fix Bug 1384094 - High CPU. network traffic and memory usage when open Newtab page if you have opened / bookmarked a problematic/ huge game page in the past

Mardak:
r?@sarracini This somewhat conflicts with #3891 which expects a `null` return value. So depending on which one lands first, there needs to be some fixup. But that's also why I expose `EMPTY_SCREENSHOT`. @piatra 

Tested with dummy pages, e.g., add "localhost", as well as the problematic pages from the bugs: https://bugzilla.mozilla.org/attachment.cgi?id=8889875

Also tested that expiration also doesn't get rid of these empty files if they're a top site.

screen shot 2018-03-22 at 1 54 55 pm

### DIFF
--- a/system-addon/lib/Screenshots.jsm
+++ b/system-addon/lib/Screenshots.jsm
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-const EXPORTED_SYMBOLS = ["Screenshots"];
+const EXPORTED_SYMBOLS = ["Screenshots", "EMPTY_SCREENSHOT"];
 
 ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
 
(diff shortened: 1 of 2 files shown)


Mardak:
Added "Bug 1384094" to the comment!

as-pine-proxy:
[Treeherder: cedb0b0d64b97a345e35b7af3ba0918e985dda1e](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=cedb0b0d64b97a345e35b7af3ba0918e985dda1e)

Mardak:
Actually. I'm going to revert this. Pretty sure this is causing pine failures: TEST-UNEXPECTED-FAIL | toolkit/components/extensions/test/mochitest/test-oop-extensions/test_ext_webrequest_basic.html | unexpected request image/png;base64, 

Maybe getScreenshot should just return null. And if it detects that it's an empty thumbnail, also return null. This way instead of showing a white thumbnail and a small letter fallback, it'll show a large letter fallback.

PR #4057, opened 2018-03-24
Fix Bug 1421428 - Link rich icon for redirected url

ncloudioj:


ncloudioj:
This patch tries to fix the issue of missing rich icons for the redirected URLs in Top Sites. Specifically, if a top sites url can't find a rich icon from neither favicon.sqlite nor TippyTop, as the last resort, it queries the Places to see if it's a redirected url, if so and its final url has a rich icon in the favicon database, it then links to that icon so AS can use it next time.

Note:
* In this implementation, it does all those in the main thread idle handler in order to minimize the performance impact
* I put all the code in the `faviconfeed.jsm`, although we could also move it to newtabutils if necessary

@Mardak Would be great to hear your thoughts on this before kicking it to r?

ncloudioj:
Tested this with a few sites with redirects, it did help the icon quality of top sites icons. In this case, tippytop served the icon for "python.org", and this patch found the rich icons from favicon.sqlite for "nike.com", "fastmail.com", and "wealthsimple.com".

screen shot 2018-03-26 at 3 18 19 pm

Mardak:
@rlr wrote the tippy top icon stuff, so he can probably comment more. I think the high level approach should be fine: i.e., on missing rich icon, check local favicons for a better one.

We probably need to be careful about unnecessarily checking over and over again, e.g., 1) in short bursts of repeated refresh calls (startup?) and 2) periodically (every 5 minutes).

Also, for tippytop, we ended up setting `#tippytop` as the ref just to distinguish that we specially set them -- just in case we needed to identify them and remove / clean up, etc.

ncloudioj:
> We probably need to be careful about unnecessarily checking over and over again, e.g., 1) in short bursts of repeated refresh calls (startup?) and 2) periodically (every 5 minutes).

Makes sense. It can maintain a set of already-dispatched urls to avoid the unnecessary checking.

rlr:
The code looks good, pretty straightforward except for that SQL which is giving me a headache. I'm testing it out now with various cases. stay tuned.

rlr:
using tinyurl.com to redirect to python.org...

before:
screen shot 2018-03-27 at 12 49 35 pm

after:
screen shot 2018-03-27 at 12 53 11 pm

system-addon/lib/FaviconFeed.jsm
```diff
@@ -10,23 +10,122 @@ Cu.importGlobalProperties(["fetch", "URL"]);
...
+async function fetchIconFromRedirects(url) {
+  const visitPaths = await fetchVisitPaths(url);
+  if (visitPaths.length > 1) {
+    const lastVisit = visitPaths.pop();
+    const redirectedUri = Services.io.newURI(lastVisit.url);
+    const iconInfo = await getFaviconInfo(redirectedUri);
+    if (iconInfo && iconInfo.faviconSize >= MIN_FAVICON_SIZE) {
+      PlacesUtils.favicons.setAndFetchFaviconForPage(
+        Services.io.newURI(url),
+        iconInfo.iconUri,
```
Mardak: Does this use the original icon uri? I would have expected something like tippytop's ` iconUri = iconUri.mutate().setRef("tippytop").finalize();`

ncloudioj: You mean, something like `.setRef("redirect").finalize();`?

I thought that if we'd only use that ref to identify the favicon type in telemetry (as it's commented [here](https://github.com/mozilla/activity-stream/blob/master/system-addon/lib/FaviconFeed.jsm#L233)), then this really is just a regular rich icon.

ncloudioj:
Another note on @Mardak's [comment](https://github.com/mozilla/activity-stream/pull/4057#discussion_r177533807) of setting a ref for the favicon. Firefox treats them as two different icons even though they share the same url, hence it will download the icon again instead of re-using the existing one. That's perhaps another reason of keeping the icon url as it is.

Mardak:
Just more that there's "no going back" or at least it'll be really hard to identify which redirected urls we got favicons for with this change vs were originally on the page.

And I guess do we want telemetry of how many redirected pages we fixed or lump them into the rich icons?

as-pine-proxy:
[Treeherder: 15cbe7820ce3b20966b414fa3cbd78cccfe11fff](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=15cbe7820ce3b20966b414fa3cbd78cccfe11fff)

ncloudioj:
> Just more that there's "no going back" or at least it'll be really hard to identify which redirected urls we got favicons for with this change vs were originally on the page.

Yea, linking without tagging will make it impossible to tell the difference. The only way to "going back" would be delete that url from history.

I believe that lumping them into the rich icon should be fine for now from the telemetry's perspective. I can file a separate bug to discuss if there is any interest of knowing how many redirected pages get fixed by this.

system-addon/lib/FaviconFeed.jsm
```diff
@@ -134,6 +241,14 @@ this.FaviconFeed = class FaviconFeed {
         Services.scriptSecurityManager.getSystemPrincipal()
       );
     }
+
+    if (!fetchedByTippyTop) {
+      const md5 = md5Hash(url);
+      if (!this._queryForRedirects.has(md5)) {
+        this._queryForRedirects.add(md5);
```
Mardak: Curious, why complicate with hashing instead of just checking/adding the url?

ncloudioj: Because this will sit in memory, just a preventive move in case the size of URL goes too far (Places caps it as [65536](https://searchfox.org/mozilla-central/source/toolkit/components/places/PlacesUtils.jsm#193)).

Mardak: We have plenty of urls and copies of this exact url in various in-memory caches, etc. And now we have an additional jsm to load in to memory… oh well

ncloudioj: Oh, I didn't know that we wanted to limit the number of jsm modules. Let me revert this part of change then.

PR #4059, opened 2018-03-26
 Fix Bug 1421682 - Activity Streams exhausts file descriptors capturing screenshots when large numbers of bookmarks are added at once

sarracini:
@Mardak what do you think about this as a proof of concept for a fix of [Bug 1421682](https://bugzilla.mozilla.org/show_bug.cgi?id=1421682) ? Just debounce the `onItemAdded`/ `onItemRemoved` calls until the last one finishes, so that large amounts of bookmarks don't refresh highlights a bunch and cause the thumbnailer to go crazy.


Mardak:
So... It's true that `HighlightsFeed` doesn't actually look at the contents of the action, but it looks like `SectionsManager` does. So only dispatching the first url's data after waiting for the last bookmark isn't quite right.

I wonder if we just want a broader "PLACES_STUFF_CHANGED" type action ;) Notice in `PlacesFeed`, `HistoryObserver.onDeleteURI` does some batching too, but nothing really looks at the array of deleted links either? We probably want to update that to use some shared `PlacesFeed.dispatchStuffChangedAfterDelay`. And then TopSites and Highlights, etc would expire/refresh on those actions instead.

In terms of sharing something from `PlacesFeed` to the two `{History,Bookmarks}Observer`, both constructors take a custom `dispatch` already:
https://github.com/mozilla/activity-stream/blob/837cea5f88ceac381a95b5c6bdb0f1e3106d2ad2/system-addon/lib/PlacesFeed.jsm#L174-L175

So we could have the custom dispatch specially detect at.PLACES_STUFF_CHANGED to debounce before actually calling `this.store.dispatch(Broadcast)` ?

This way, the timer exists on `PlacesFeed`, so you don't need a special "deconstructor" ;) and clean things up in UNINIT/removeObservers

sarracini:
@Mardak something like this? If this approach is ok I'll go ahead and fix the tests

system-addon/common/Reducers.jsm
```diff
@@ -166,6 +166,12 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
         return site;
       });
       return Object.assign({}, prevState, {rows: newRows});
+    case at.PLACES_LINKS_DELETED:
```
sarracini: Without picking this up in the reducer, it takes a full second to update the page after deleting something, and it looks slow and unresponsive

system-addon/lib/PlacesFeed.jsm
```diff
@@ -171,8 +176,9 @@ class BookmarksObserver extends Observer {
 
 class PlacesFeed {
   constructor() {
-    this.historyObserver = new HistoryObserver(action => this.store.dispatch(ac.BroadcastToContent(action)));
-    this.bookmarksObserver = new BookmarksObserver(action => this.store.dispatch(ac.BroadcastToContent(action)));
+    this.placesChangedTimer = null;
+    this.historyObserver = new HistoryObserver(action => this.customDispatch(action));
+    this.bookmarksObserver = new BookmarksObserver(action => this.customDispatch(action));
```
Mardak: Probably just bind once and pass in:
```js
this.customDispatch = this.customDispatch.bind(this);
… new HistoryObserver(this.customDispatch);
…
```

system-addon/lib/PlacesFeed.jsm
```diff
@@ -187,7 +193,40 @@ class PlacesFeed {
...
+  customDispatch(action) {
+    // If we are changing many links at once, debounce this action and only dispatch
+    // at the end
+    if (action.type === at.PLACES_LINKS_CHANGED) {
+      if (this.placesChangedTimer) {
+        this.placesChangedTimer.delay = PLACES_LINKS_CHANGED_DEBOUNCE_TIME;
+      } else {
+        this.placesChangedTimer = this.setTimeout(() => {
+          this.placesChangedTimer = null;
+          this.store.dispatch(ac.OnlyToMain(action));
```
Mardak: Oh only to main is kinda tricky, but that's fine. Hopefully callers don't accidentally use some other action creator…

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -366,7 +366,7 @@ this.TopSitesFeed = class TopSitesFeed {
       // All these actions mean we need new top sites
       case at.MIGRATION_COMPLETED:
       case at.PLACES_HISTORY_CLEARED:
-      case at.PLACES_LINKS_DELETED:
+      case at.PLACES_LINKS_CHANGED:
```
Mardak: Ah hrm.. Also we end up unnecessarily refreshing top sites now for bookmark triggered CHANGED now… oh well?

But main issue is if it needs to broadcast anymore, and seems like no?

sarracini: should it still expire the frecentCache?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -366,10 +366,14 @@ this.TopSitesFeed = class TopSitesFeed {
       // All these actions mean we need new top sites
       case at.MIGRATION_COMPLETED:
       case at.PLACES_HISTORY_CLEARED:
-      case at.PLACES_LINKS_DELETED:
         this.frecentCache.expire();
         this.refresh({broadcast: true});
         break;
+      case at.PLACES_LINKS_CHANGED:
+        this.frecentCache.expire();
+        this.pinnedCache.expire();
```
Mardak: I don't think pinned cache needs to expire from thees history/bookmark places changes… ? I.e., are there any PLACES_LINKS_CHANGED events that would cause the pinned cache to be wrong?

as-pine-proxy:
[Treeherder: 3600de8ce4aa428547ae38496f114a766768465d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=3600de8ce4aa428547ae38496f114a766768465d)

PR #4061, opened 2018-03-27
Fix Bug 1448911 - Fail early for file protocol customScreenshotUrl

piatra:


piatra:
Do you think there are any advantages if alternatively we would check `new URL(...).protocol` ?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -123,17 +123,27 @@ export class TopSiteForm extends React.PureComponent {
...
+    return urlObj && validProtocols.includes(urlObj.protocol);
+  }
+
   validateCustomScreenshotUrl() {
+    const invalidProtocols = ["file:"];
     const {customScreenshotUrl} = this.state;
-    return !customScreenshotUrl || this.validateUrl(customScreenshotUrl);
+
+    return !customScreenshotUrl ||
+      (this.validateUrl(customScreenshotUrl) && !invalidProtocols.some(p => customScreenshotUrl.startsWith(p)));
```
Mardak: `invalidProtocols` is redundant code now. `validateUrl` will only be truthy if it's http/s, so the `.some` check will always be `false`.

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -123,17 +123,27 @@ export class TopSiteForm extends React.PureComponent {
...
-      return !!new URL(this.cleanUrl(url));
+      return new URL(url);
     } catch (e) {
       return false;
     }
   }
 
+  validateUrl(url) {
+    const validProtocols = ["http:", "https:"];
+    const urlObj = this._tryParseUrl(url) || this._tryParseUrl(`https://${url}`);
```
Mardak: Any reason to add "https" here but "http" for `cleanUrl`? Why not just reuse `cleanUrl` for the fallback?

system-addon/content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -123,14 +123,21 @@ export class TopSiteForm extends React.PureComponent {
     return url;
   }
 
-  validateUrl(url) {
+  _tryParseUrl(url) {
     try {
-      return !!new URL(this.cleanUrl(url));
+      return new URL(url);
     } catch (e) {
       return false;
```
Mardak: nit: this should return `null` as we're returning objects now.

as-pine-proxy:
[Treeherder: 075fa4b4de7fc51181fa4b4ddcad8f0d89290814](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=075fa4b4de7fc51181fa4b4ddcad8f0d89290814)

PR #4062, opened 2018-03-27
Fix Bug 1449292 - TopSites don't update their custom screenshot

piatra:
This scenario didn't really show up in development because it required some amount of history.

You can have frecent topsites which are low quality (have a screenshot not a hi-res icon). If you were to set a screenshot for these, we would not do a lookup in the frecent cache to clear the old screenshot and because of 
https://github.com/mozilla/activity-stream/blob/e8ef6e23b38c4eed36c6d0bf3f04516290b86ad9/system-addon/lib/TopSitesFeed.jsm#L203-L205
We would not fetch the new screenshot.

### DIFF
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -103,13 +103,15 @@ this.TopSitesFeed = class TopSitesFeed {
       // Copy all properties from a frecent link and add more
       const finder = other => other.url === link.url;
 
+      // Remove frecent link's screenshot if pinned link has a custom one
       const frecentSite = frecent.find(finder);
+      if (frecentSite && link.customScreenshotURL) {
+        delete frecentSite.screenshot;
+      }
       // If the link is a frecent site, do not copy over 'isDefault', else check
       // if the site is a default site
       const copy = Object.assign({}, frecentSite ||
-        {isDefault: !!notBlockedDefaultSites.find(finder)}, link, {hostname: shortURL(link)},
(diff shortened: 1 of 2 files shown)


Mardak:
Why not avoid using the screenshot from frecent if the pinned link has a custom screenshot?

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -100,10 +100,13 @@ this.TopSitesFeed = class TopSitesFeed {
...
 
+      const frecentSite = frecent.find(finder);
       // If the link is a frecent site, do not copy over 'isDefault', else check
       // if the site is a default site
-      const copy = Object.assign({}, frecent.find(finder) ||
-        {isDefault: !!notBlockedDefaultSites.find(finder)}, link, {hostname: shortURL(link)});
+      const copy = Object.assign({}, frecentSite ||
+        {isDefault: !!notBlockedDefaultSites.find(finder)}, link, {hostname: shortURL(link)},
+        // If the pinned site has a customScreenshotURL the frecent screenshot will be incorrect
+        (frecentSite && !link.screenshot) && {screenshot: link.customScreenshotURL ? undefined : frecentSite.screenshot});
```
Mardak: This would be much easier to read if you remove the screenshot above:
```js
// Remove frecent link's screenshot if pinned link has a custom one
const frecentSite = frecent.find(finder);
if (frecentSite && link.customScreenshotURL) {
  delete frecentSite.screenshot;
```

PR #4064, opened 2018-03-28
chore (codequality) #nobug Remove unnecessary usage of injector

rlr:
r? @Mardak

PR #4066, opened 2018-03-28
chore(tests): Clean up unnecessary injector and wrong global setting

Mardak:
r? @rlr Followup for #4064. Make unit-entry have a TEST_GLOBAL for the various Cc and alias PlacesUtils to get those Cc. Also fixes PlacesFeed test to not change global directly, which wasn't getting restored.

Basically adding dummy methods/properties on unit-entry to later sandbox.stub/spy.

### DIFF
--- a/system-addon/test/unit/lib/UserDomainAffinityProvider.test.js
+++ b/system-addon/test/unit/lib/UserDomainAffinityProvider.test.js
@@ -1,5 +1,5 @@
 import {GlobalOverrider} from "test/unit/utils";
-import injector from "inject!lib/UserDomainAffinityProvider.jsm";
+import {UserDomainAffinityProvider} from "lib/UserDomainAffinityProvider.jsm";
 
 const TIME_SEGMENTS = [
   {"id": "hour", "startTime": 3600, "endTime": 0, "weightPosition": 1},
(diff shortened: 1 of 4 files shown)


Mardak:
Also, instead of globals.set replacing whole objects, stub individual methods/properties.

system-addon/test/unit/lib/UserDomainAffinityProvider.test.js
```diff
@@ -30,27 +30,20 @@ const PARAMETER_SETS = {
...
-      getService() {
-        return global.PlacesUtils.history;
-      }
-    };
 
-    ({UserDomainAffinityProvider} = injector());
+    const testUrl = "www.somedomain.com";
+    globals.sandbox.stub(global.Services.io, "newURI").returns({host: testUrl});
+
+    const history = global.Cc["@mozilla.org/browser/nav-history-service;1"];
```
rlr: cool trick. Any reason that `const history = Cc["@mozilla.org/browser/nav-history-service;1"].getService(Ci.nsINavHistoryService);` isn't done inside the contructor instead of at import time. Or even more lazily? Or does it not matter?

rlr: Sorry for distracting with this kind of unrelated stuff 🙈

Mardak: Yeah, it should probably be lazy. Not sure why it was written to be access at the top level…

as-pine-proxy:
[Treeherder: c7bb4057a6f4ebcd68d8bb5231ce8bd51b27ce5d](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=c7bb4057a6f4ebcd68d8bb5231ce8bd51b27ce5d)

PR #4058, opened 2018-03-26
Fix Bug 1445157 - Move isCollapsed pref to IndexedDB

piatra:
This reverts commit e6c25770e88b3c639111491451f17a1e9e4e6881 (that reverted and backed out the patch). 

The issue was that I did not increment the `dbVersion` so the `onupgradeneeded` never updated the schema for profiles that needed to migrate.

PR #4060, opened 2018-03-27
Fix bug 1449223 - Add Telemetry for failed IndexedDB Transactions

piatra:
WIP This depends on #4058 landing first.

PR #4068, opened 2018-03-28
fix(storage): Lazily open database on first access instead of requiri…

Mardak:
Fixing up problem introduced with the combination of #4044 and #4058.

r?@piatra I don't actually know how IndexedDB stuff works… The objectStore.methods seem to return something promise-like that can be awaited?

### DIFF
--- a/system-addon/lib/ActivityStream.jsm
+++ b/system-addon/lib/ActivityStream.jsm
@@ -280,7 +280,11 @@ this.ActivityStream = class ActivityStream {
     try {
       this._updateDynamicPrefs();
       this._defaultPrefs.init();
-      this._storage.init();
+
+      // Accessing the db causes the object stores to be created / migrated.
+      // This needs to happen before other instances try to access the db, which
+      // would update only a subset of the stores to the latest version.
+      this._storage.db; // eslint-disable-line no-unused-expressions
 
       // Hook up the store and let all feeds and pages initialize
       this.store.init(this.feeds, ac.BroadcastToContent({
--- a/system-addon/lib/ActivityStreamStorage.jsm
+++ b/system-addon/lib/ActivityStreamStorage.jsm
@@ -9,35 +9,26 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
     this.dbName = "ActivityStream";
     this.dbVersion = 3;
     this.storeName = storeName;
-
-    this._db = null;
   }
 
   get db() {
-    if (!this._db) {
-      throw new Error("It looks like the db connection has not initialized yet. Are you use .init was called?");
-    }
-    return this._db;
+    return this._db || (this._db = this._openDatabase());
   }
 
-  get intialized() {
-    return this._db !== null;
+  async getStore() {
+    return (await this.db).objectStore(this.storeName, "readwrite");
   }
 
-  getStore() {
-    return this.db.objectStore(this.storeName, "readwrite");
+  async get(key) {
+    return (await this.getStore()).get(key);
   }
 
-  get(key) {
-    return this.getStore().get(key);
+  async getAll() {
+    return (await this.getStore()).getAll();
   }
 
-  getAll() {
-    return this.getStore().getAll();
-  }
-
-  set(key, value) {
-    return this.getStore().put(value, key);
+  async set(key, value) {
+    return (await this.getStore()).put(value, key);
   }
 
   _openDatabase() {
@@ -54,10 +45,6 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
         db.createObjectStore(this.storeName);
       }
     });
-  }
-
-  async init() {
-    this._db = await this._openDatabase();
   }
 };
 
--- a/system-addon/lib/PrefsFeed.jsm
+++ b/system-addon/lib/PrefsFeed.jsm
@@ -88,7 +88,7 @@ this.PrefsFeed = class PrefsFeed {
     }
   }
 
-  async init() {
+  init() {
     this._prefs.observeBranch(this);
 
     // Get the initial value of each activity stream pref
@@ -103,10 +103,7 @@ this.PrefsFeed = class PrefsFeed {
     // Set the initial state of all prefs in redux
     this.store.dispatch(ac.BroadcastToContent({type: at.PREFS_INITIAL_VALUES, data: values}));
 
-    if (!this._storage.intialized) {
-      await this._storage.init();
-      this._migratePrefs();
-    }
+    this._migratePrefs();
     this._setPrerenderPref();
     this._initOnboardingPref();
   }
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -81,7 +81,7 @@ const SectionsManager = {
   initialized: false,
   sections: new Map(),
   async init(prefs = {}) {
-    await this.initStorage();
+    this._storage = new ActivityStreamStorage("sectionPrefs");
 
     for (const feedPrefName of Object.keys(BUILT_IN_SECTIONS)) {
       const optionsPrefName = `${feedPrefName}.options`;
@@ -103,10 +103,6 @@ const SectionsManager = {
 
     this.initialized = true;
     this.emit(this.INIT);
-  },
-  initStorage() {
-    this._storage = new ActivityStreamStorage("sectionPrefs");
-    return this._storage.init();
   },
   observe(subject, topic, data) {
     switch (topic) {
--- a/system-addon/lib/SnippetsFeed.jsm
+++ b/system-addon/lib/SnippetsFeed.jsm
@@ -169,7 +169,6 @@ this.SnippetsFeed = class SnippetsFeed {
   }
 
   async init() {
-    await this._storage.init();
     this._previousSessionEnd = await this._storage.get("previousSessionEnd");
     await this._refresh();
     Services.prefs.addObserver(ONBOARDING_FINISHED_PREF, this._refresh);
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -167,11 +167,8 @@ this.TopSitesFeed = class TopSitesFeed {
     const links = await this.getLinksWithDefaults();
     const newAction = {type: at.TOP_SITES_UPDATED, data: {links}};
 
-    if (!this._storage.initialized) {
-      await this._storage.init();
-      const storedPrefs = await this._storage.get(SECTION_ID) || {};
-      newAction.data.pref = getDefaultOptions(storedPrefs);
-    }
+    const storedPrefs = await this._storage.get(SECTION_ID) || {};
+    newAction.data.pref = getDefaultOptions(storedPrefs);
 
     if (options.broadcast) {
       // Broadcast an update to all open content pages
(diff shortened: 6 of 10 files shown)


rlr:
I saw a `dump` in there! 😈

rlr:
I had just got it working too 😜 

```
diff --git a/system-addon/lib/ActivityStreamStorage.jsm b/system-addon/lib/ActivityStreamStorage.jsm
index 18261071..47438eee 100644
--- a/system-addon/lib/ActivityStreamStorage.jsm
+++ b/system-addon/lib/ActivityStreamStorage.jsm
@@ -9,35 +9,25 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
     this.dbName = "ActivityStream";
     this.dbVersion = 3;
     this.storeName = storeName;
-
-    this._db = null;
   }

-  get db() {
+  async getStore() {
     if (!this._db) {
-      throw new Error("It looks like the db connection has not initialized yet. Are you use .init was called?");
+      this._db = await this._openDatabase();
     }
-    return this._db;
-  }
-
-  get intialized() {
-    return this._db !== null;
+    return this._db.objectStore(this.storeName, "readwrite");
   }

-  getStore() {
-    return this.db.objectStore(this.storeName, "readwrite");
+  async get(key) {
+    return (await this.getStore()).get(key);
   }

-  get(key) {
-    return this.getStore().get(key);
+  async getAll() {
+    return (await this.getStore()).getAll();
   }

-  getAll() {
-    return this.getStore().getAll();
-  }
-
-  set(key, value) {
-    return this.getStore().put(value, key);
+  async set(key, value) {
+    return (await this.getStore()).put(value, key);
   }

   _openDatabase() {
@@ -55,10 +45,6 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
       }
     });
   }
-
-  async init() {
-    this._db = await this._openDatabase();
-  }
 };

 function getDefaultOptions(options) {
diff --git a/system-addon/lib/PrefsFeed.jsm b/system-addon/lib/PrefsFeed.jsm
index 5a4faa2d..ee73c99f 100644
--- a/system-addon/lib/PrefsFeed.jsm
+++ b/system-addon/lib/PrefsFeed.jsm
@@ -88,7 +88,7 @@ this.PrefsFeed = class PrefsFeed {
     }
   }

-  async init() {
+  init() {
     this._prefs.observeBranch(this);

     // Get the initial value of each activity stream pref
@@ -103,10 +103,7 @@ this.PrefsFeed = class PrefsFeed {
     // Set the initial state of all prefs in redux
     this.store.dispatch(ac.BroadcastToContent({type: at.PREFS_INITIAL_VALUES, data: values}));

-    if (!this._storage.intialized) {
-      await this._storage.init();
-      this._migratePrefs();
-    }
+    this._migratePrefs();
     this._setPrerenderPref();
     this._initOnboardingPref();
   }
diff --git a/system-addon/lib/SectionsManager.jsm b/system-addon/lib/SectionsManager.jsm
index 300e802e..48fcc40c 100644
--- a/system-addon/lib/SectionsManager.jsm
+++ b/system-addon/lib/SectionsManager.jsm
@@ -81,7 +81,7 @@ const SectionsManager = {
   initialized: false,
   sections: new Map(),
   async init(prefs = {}) {
-    await this.initStorage();
+    this._storage = new ActivityStreamStorage("sectionPrefs");

     for (const feedPrefName of Object.keys(BUILT_IN_SECTIONS)) {
       const optionsPrefName = `${feedPrefName}.options`;
@@ -104,10 +104,6 @@ const SectionsManager = {
     this.initialized = true;
     this.emit(this.INIT);
   },
-  initStorage() {
-    this._storage = new ActivityStreamStorage("sectionPrefs");
-    return this._storage.init();
-  },
   observe(subject, topic, data) {
     switch (topic) {
       case "nsPref:changed":
diff --git a/system-addon/lib/SnippetsFeed.jsm b/system-addon/lib/SnippetsFeed.jsm
index c985f15b..c6d51b75 100644
--- a/system-addon/lib/SnippetsFeed.jsm
+++ b/system-addon/lib/SnippetsFeed.jsm
@@ -169,7 +169,6 @@ this.SnippetsFeed = class SnippetsFeed {
   }

   async init() {
-    await this._storage.init();
     this._previousSessionEnd = await this._storage.get("previousSessionEnd");
     await this._refresh();
     Services.prefs.addObserver(ONBOARDING_FINISHED_PREF, this._refresh);
diff --git a/system-addon/test/unit/lib/ActivityStreamStorage.test.js b/system-addon/test/unit/lib/ActivityStreamStorage.test.js
index ea9ce783..d1b4deee 100644
--- a/system-addon/test/unit/lib/ActivityStreamStorage.test.js
+++ b/system-addon/test/unit/lib/ActivityStreamStorage.test.js
@@ -16,25 +16,18 @@ describe("ActivityStreamStorage", () => {
   afterEach(() => {
     sandbox.restore();
   });
-  it("should throw an error if you try to use it without init", () => {
-    assert.throws(() => storage.db);
-  });
-  it("should not throw an error if you called init", async () => {
-    await storage.init();
-    assert.ok(storage.db);
-  });
-  it("should revert key value parameters for put", () => {
+  it("should revert key value parameters for put", async () => {
     const stub = sandbox.stub();
-    sandbox.stub(storage, "getStore").returns({put: stub});
+    sandbox.stub(storage, "getStore").resolves({put: stub});

-    storage.set("key", "value");
+    await storage.set("key", "value");

     assert.calledOnce(stub);
     assert.calledWith(stub, "value", "key");
   });
   it("should create a db with the correct store name", async () => {
     const dbStub = {createObjectStore: sandbox.stub(), objectStoreNames: {contains: sandbox.stub().returns(false)}};
-    await storage.init();
+    await storage._openDatabase();

     // call the cb with a stub
     indexedDB.open.args[0][2](dbStub);
@@ -45,7 +38,7 @@ describe("ActivityStreamStorage", () => {
   it("should handle an array of object store names", async () => {
     storage = new ActivityStreamStorage(["store1", "store2"]);
     const dbStub = {createObjectStore: sandbox.stub(), objectStoreNames: {contains: sandbox.stub().returns(false)}};
-    await storage.init();
+    await storage._openDatabase();

     // call the cb with a stub
     indexedDB.open.args[0][2](dbStub);
@@ -57,20 +50,11 @@ describe("ActivityStreamStorage", () => {
   it("should skip creating existing stores", async () => {
     storage = new ActivityStreamStorage(["store1", "store2"]);
     const dbStub = {createObjectStore: sandbox.stub(), objectStoreNames: {contains: sandbox.stub().returns(true)}};
-    await storage.init();
+    await storage._openDatabase();

     // call the cb with a stub
     indexedDB.open.args[0][2](dbStub);

     assert.notCalled(dbStub.createObjectStore);
   });
-  it("should be initialized after calling init", async () => {
-    const dbStub = {createObjectStore: sandbox.stub(), objectStoreNames: {contains: sandbox.stub().returns(false)}};
-    await storage.init();
-
-    // call the cb with a stub
-    indexedDB.open.args[0][2](dbStub);
-
-    assert.isTrue(storage.intialized);
-  });
 });
diff --git a/system-addon/test/unit/lib/SectionsManager.test.js b/system-addon/test/unit/lib/SectionsManager.test.js
index 57edead0..9375cdf2 100644
--- a/system-addon/test/unit/lib/SectionsManager.test.js
+++ b/system-addon/test/unit/lib/SectionsManager.test.js
@@ -207,13 +207,6 @@ describe("SectionsManager", () => {

       assert.calledOnce(SectionsManager.updateSections);
     });
-    it("should call initStorage on init", async () => {
-      sandbox.stub(SectionsManager, "initStorage").returns(Promise.resolve());
-
-      await SectionsManager.init();
-
-      assert.calledOnce(SectionsManager.initStorage);
-    });
   });
   describe("#_addCardTypeLinkMenuOptions", () => {
     const addCardTypeLinkMenuOptionsOrig = SectionsManager._addCardTypeLinkMenuOptions;
```

rlr:
Main difference is I removed `.db` from the API since it wasn't really used.

rlr:
easier to see: https://github.com/mozilla/activity-stream/compare/master...rlr:storage-init

So if we both wrote same-ish code on our own, is that an automatic r+? I guess @piatra 👀

system-addon/lib/ActivityStreamStorage.jsm
```diff
@@ -9,35 +9,26 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
...
   get db() {
-    if (!this._db) {
-      throw new Error("It looks like the db connection has not initialized yet. Are you use .init was called?");
-    }
-    return this._db;
-  }
-
-  get intialized() {
-    return this._db !== null;
+    return this._db || (this._db = this._openDatabase());
```
rlr: oh right. my patch has a race condition here. 👍

system-addon/lib/ActivityStream.jsm
```diff
@@ -280,7 +280,9 @@ this.ActivityStream = class ActivityStream {
     try {
       this._updateDynamicPrefs();
       this._defaultPrefs.init();
-      this._storage.init();
+
+      // Accessing the db causes the tables to be created / migrated
+      this._storage.db; // eslint-disable-line no-unused-expressions
```
Mardak: @piatra should this `await`?

piatra: That's what I expected too but no. The callback function passed in is [`onupgradeneeded`](https://searchfox.org/mozilla-central/rev/7e663b9fa578d425684ce2560e5fa2464f504b34/toolkit/modules/IndexedDB.jsm#245-249)
```
when the database needs to be created, or its schema needs to be upgraded.
```
So the very first call sets up the objectStores, the rest will only connect to them.

Mardak: But isn't a new `IndexedDB.open` being called for every new instance of `ActivityStreamStorage` in each of the feeds that use it? I suppose a related question is why do we need so many instances?

piatra: 1) Yes `open` is called but the cb we pass in is not. After the first call subsequent requests will only [open the db](https://searchfox.org/mozilla-central/rev/7e663b9fa578d425684ce2560e5fa2464f504b34/toolkit/modules/IndexedDB.jsm#254) requested (and expect it to be there).
2) Initially it was only needed by snippets when the code was added and the next consumer would have been the Sections. Because the SectionsManager (where the sections prefs are built) is not connected to the store I did not think to make the Storage a feed (I'm not sure if this is what you had in mind). Also this way there is also separation between the storages of different consumers.
* It's not great as I'm finding out in #4060 where to get access to the TelemetryFeed and report IndexedDB errors I have to pass the feed around to the storage.

Mardak: Oh, I guess the version is shared among all the.. tables/objects? So the earliest instance/open/request is going to trigger the upgrade, and it will need to upgrade the schema for each thing we care about?

piatra: Yes.

To answer the initial question: also yes. I wasn't sure if there are any negative side effect of awaiting in our main `init` function. The init call happens before any of the feeds get initialized and also testing it out seemed that it could work without await-ing on storage.

as-pine-proxy:
[Treeherder: 72c311f62fe5b1fc9c498cea99049d11ccb3edd1](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=72c311f62fe5b1fc9c498cea99049d11ccb3edd1)

PR #3797, opened 2017-11-06
fix(topstories): #3796 Pocket disclaimer not clickable for some users

csadilek:
Closes #3796 

This is adding a fallback in case the disclaimer link is not configured and the defaults can't be overridden i.e. due to an active shield study.

Issue #3796, opened 2017-11-06
Pocket disclaimer link is not clickable


piatra:
<img width="757" alt="image" src="https://user-images.githubusercontent.com/810040/32438856-6b71df28-c2ec-11e7-936a-6a5a6e9eda77.png">
Pocket disclaimer link does not have an href.

csadilek:
The reasons for this not being clickable for some users is that we launched a Shield study last week targeting the topstories.options pref just before we got this link. So, for users in this study the href won't be configured. I will prepare a PR with a client-side fallback in case the config is missing.

PR #4070, opened 2018-03-29
Fix Bug 1446195 - Update story preferences description and link

Mardak:
r?@sarracini Moved the link to always be shown for stories instead of conditionally for sponsored. The old string can be replaced/removed because it wasn't actually used in 60. Also removed the option fallback stuff as it was confusingly out of sync for the disclaimer link href and the other options weren't being set (originally fixing #3796 in #3797)

Before sponsored off:
screen shot 2018-03-28 at 9 55 32 pm
Before sponsored on:
screen shot 2018-03-28 at 9 54 39 pm

After sponsored off:
screen shot 2018-03-28 at 9 57 18 pm
After sponsored on:
screen shot 2018-03-28 at 9 53 08 pm

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -80,7 +80,7 @@ prefs_restore_defaults_button=Restore Defaults
 prefs_section_rows_option={num} row;{num} rows
 prefs_search_header=Web Search
 prefs_topsites_description=The sites you visit most
-prefs_topstories_description=High-quality content you might otherwise miss
+prefs_topstories_description2=Great content from around the web, personalized for you
 # LOCALIZATION NOTE (prefs_topstories_show_sponsored_label): {provider} is
 # replaced by the name of the content provider for this section, e.g., "Pocket"
 prefs_topstories_show_sponsored_label={provider} Sponsored Stories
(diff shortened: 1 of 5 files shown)


locales/en-US/strings.properties
```diff
@@ -80,7 +80,7 @@ prefs_restore_defaults_button=Restore Defaults
 prefs_section_rows_option={num} row;{num} rows
 prefs_search_header=Web Search
 prefs_topsites_description=The sites you visit most
-prefs_topstories_description=High-quality content you might otherwise miss
+prefs_topstories_description2=Great content from around the web, personalized for you
```
sarracini: So you don't need to add `prefs_topstories_description` to [Bug 1445090](https://bugzilla.mozilla.org/show_bug.cgi?id=1445090) then?

Mardak: Nah.. unless we expect to uplift the about:preferences stuff to beta 60.. Although if we do that, we could uplift strings too ;)

as-pine-proxy:
[Treeherder: 0357ce6d42cc41acc115734f45d3cc0bfa97414a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=0357ce6d42cc41acc115734f45d3cc0bfa97414a)

PR #4073, opened 2018-03-30
chore(cleanup): fix some missed r? comments from 997f422

rlr:
ref:
https://github.com/mozilla/activity-stream/pull/4044#discussion_r178161891
https://github.com/mozilla/activity-stream/pull/4044#discussion_r178194126

r? @Mardak

### DIFF
--- a/system-addon/content-src/components/Base/Base.jsx
+++ b/system-addon/content-src/components/Base/Base.jsx
@@ -94,7 +94,12 @@ export class BaseContent extends React.PureComponent {
 
     const shouldBeFixedToTop = PrerenderData.arePrefsValid(name => prefs[name]);
 
-    const outerClassName = `outer-wrapper ${Theme.className}${shouldBeFixedToTop ? " fixed-to-top" : ""} ${prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"}`;
+    const outerClassName = [
+      "outer-wrapper",
+      Theme.className,
+      shouldBeFixedToTop && "fixed-to-top",
+      prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"
+    ].filter(v => v).join(" ");
 
     return (
         <div className={outerClassName}>
--- a/system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
+++ b/system-addon/content-src/components/CollapsibleSection/_CollapsibleSection.scss
@@ -109,7 +109,7 @@
     }
 
     button {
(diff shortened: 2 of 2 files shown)


system-addon/content-src/components/Base/Base.jsx
```diff
@@ -94,7 +94,7 @@ export class BaseContent extends React.PureComponent {
 
     const shouldBeFixedToTop = PrerenderData.arePrefsValid(name => prefs[name]);
 
-    const outerClassName = `outer-wrapper ${Theme.className}${shouldBeFixedToTop ? " fixed-to-top" : ""} ${prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"}`;
+    const outerClassName = `outer-wrapper${Theme.className ? ` ${Theme.className}` : ""}${shouldBeFixedToTop ? " fixed-to-top" : ""} ${prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"}`;
```
Mardak: eh. mrhm.m.m.... I know this is how we do it elsewhere.... What if we just sneak in a simple filter/join..
```js
const outerClassName = [
  "outer-wrapper",
  Theme.className,
  shouldBeFixedToTop && "fixed-to-top",
  prefs.enableWideLayout ? "wide-layout-enabled" : "wide-layout-disabled"
].filter(v => v).join(" ");
```

I thought @k88hudson had written something like this somewhere already…

Maybe we sneak it in here and add it to some `content-src/lib` later ;)

Mardak:
Amy had some default / dark theme menu matching with the chrome comments in slack.

mismatch default
image

mismatch dark
image

Not sure if we want to address those just in a separate PR / followup bug to get better details on what they should be…

Mardak:
Ah this is ready, so let's split off the followup menu / dark theme bugs.

rlr:
There is also this: 

screen shot 2018-03-30 at 5 50 59 pm

🤔

as-pine-proxy:
[Treeherder: 41e79952d38cf6522b27ac9b168c5c39bca8f05c](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=41e79952d38cf6522b27ac9b168c5c39bca8f05c)

rlr:
I wonder if those should all be combined with `--newtab-element-active-color`... Also, there is another color when you are typing or if you mouse down on an item (on awesomebar only for now it seems):

screen shot 2018-03-30 at 6 18 32 pm
screen shot 2018-03-30 at 6 18 41 pm

PR #4075, opened 2018-04-03
Fix Bug 1433215 - Add blockList persistance to Message Center

piatra:
This uses the same object storage as snippets since both block based on string ids. Modified the Message Center blocklist to be an array.

### DIFF
--- a/system-addon/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/system-addon/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -33,7 +33,7 @@ export class ASRouterAdmin extends React.PureComponent {
 
   renderMessageItem(msg) {
     const isCurrent = msg.id === this.state.currentId;
-    const isBlocked = this.state.blockList[msg.id];
+    const isBlocked = this.state.blockList.includes(msg.id);
 
     let itemClassName = "message-item";
     if (isCurrent) { itemClassName += " current"; }
--- a/system-addon/lib/ASRouter.jsm
+++ b/system-addon/lib/ASRouter.jsm
@@ -61,7 +61,7 @@ class _ASRouter {
   constructor(initialState = {}) {
     this.initialized = false;
     this.messageChannel = null;
-    this._storage = new ActivityStreamStorage("snippets");
+    this._storage = null;
     this._state = Object.assign({
       currentId: null,
       blockList: [],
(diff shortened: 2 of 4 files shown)


system-addon/lib/MessageCenterRouter.jsm
```diff
@@ -132,18 +138,22 @@ class _MessageCenterRouter {
         await this.sendNextMessage(target);
         break;
       case "BLOCK_MESSAGE_BY_ID":
-        await this.setState(state => {
-          const newState = Object.assign({}, state.blockList);
-          newState[action.data.id] = true;
-          return {blockList: newState};
+        await this.setState(async state => {
+          const blockList = state.blockList.slice();
```
k88hudson: personally I think `[...state.blockList]` more clearly indicates you're making a copy

system-addon/lib/MessageCenterRouter.jsm
```diff
@@ -132,18 +138,22 @@ class _MessageCenterRouter {
...
         });
         await this.clearMessage(target, action.data.id);
         break;
       case "UNBLOCK_MESSAGE_BY_ID":
-        await this.setState(state => {
-          const newState = Object.assign({}, state.blockList);
-          delete newState[action.data.id];
-          return {blockList: newState};
+        await this.setState(async state => {
+          const blockList = state.blockList.slice();
```
k88hudson: same comment about `[...state.blockList]`

system-addon/lib/MessageCenterRouter.jsm
```diff
@@ -132,18 +138,22 @@ class _MessageCenterRouter {
...
         break;
       case "BLOCK_MESSAGE_BY_ID":
-        await this.setState(state => {
-          const newState = Object.assign({}, state.blockList);
-          newState[action.data.id] = true;
-          return {blockList: newState};
+        await this.setState(async state => {
+          const blockList = state.blockList.slice();
+          blockList.push(action.data.id);
+          await this._storage.set("blockList", blockList);
```
k88hudson: I'm not sure we want to wait on the indexedDB to actually set the state here, we probably just want to do that in the background and throw an error if there was a problem with the transaction

system-addon/lib/MessageCenterRouter.jsm
```diff
@@ -132,18 +138,22 @@ class _MessageCenterRouter {
...
         break;
       case "UNBLOCK_MESSAGE_BY_ID":
-        await this.setState(state => {
-          const newState = Object.assign({}, state.blockList);
-          delete newState[action.data.id];
-          return {blockList: newState};
+        await this.setState(async state => {
+          const blockList = state.blockList.slice();
+          blockList.splice(blockList.indexOf(action.data.id), 1);
+          await this._storage.set("blockList", blockList);
```
k88hudson: Again, I don't think we actually want to wait on the transaction here

piatra:
@k88hudson updated. regarding
> throw an error if there was a problem with the transaction

Sounds like this can just be handled by #4060

k88hudson:
🔥🔥 🔥

PR #4076, opened 2018-04-03
Fix Bug 1433230 - Add recent downloads to Highlights

sarracini:
Adds a single, most recent download to highlights. Right now just appends it to the end of all highlights, so the easiest way to test this is on a new profile, with little history and maybe just a bookmark.
Some notes:
1. Clicking on the card opens it up in the application that the OS can open it in
2. Show in Finder string is determined based on platform, and is only shown if there is a path to show
3. Go to Download page option in the context menu is disabled if the download doesn't have a referrer.
4. Clicking on the magnifying glass should do the same as Show in Finder
5. We refresh highlights when a new download gets added, or when a download gets deleted, so that it can show up immediately (like bookmarks)
6. We only care about downloads that are newer than 36 hours
7. Highlights feed needs to make an instance of DownloadsManager, but DownloadsManager needs know about actions that are dispatched to it. This means it would also need to be instantiated in ActivityStream.jsm so it can be hooked up to the store. Since we don't want 2 instances of DownloadsManager, HighlightsFeed relays actions to DownloadsManager.

Designs: https://mozilla.invisionapp.com/share/J6GEU4EHTYC#/screens 

Here's a picture!
screen shot 2018-04-16 at 4 47 02 pm


system-addon/test/unit/lib/DownloadsManager.test.js
```diff
@@ -0,0 +1,155 @@
...
+    download = downloadsManager._viewableDownloadItems.get(DOWNLOAD_URL);
+  });
+  afterEach(() => {
+    downloadsManager._viewableDownloadItems.clear();
+    globals.restore();
+  });
+  describe("#onAction", () => {
+    it("should add a DownloadsCommon view on NEW_TAB_LOAD", () => {
+      let stub = global.DownloadsCommon.getData().addView;
+      stub.reset();
```
sarracini: I think stub.reset() may be removed from their API soon......

locales/en-US/strings.properties
```diff
@@ -44,6 +45,22 @@ menu_action_save_to_pocket=Save to Pocket
...
+menu_action_show_file_windows=Open containing folder
+menu_action_show_file_linux=Open containing folder
+menu_action_show_file_default=Show File
+
+# LOCALIZATION NOTE (menu_action_copy_download_link, menu_action_go_to_download_page):
+# "Download" here, in both cases, is not a verb, it is a noun. As in, "Copy the
+# link that belongs to this downloaded item"
+menu_action_copy_download_link=Copy Download link
+menu_action_go_to_download_page=Go to Download page
+menu_action_remove_download=Remove from history
```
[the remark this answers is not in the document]
sarracini: Yeah I know, I've been back and forth with design and they told me to be consistent with our context menus in activity stream

system-addon/content-src/lib/link-menu-options.js
```diff
@@ -91,6 +98,40 @@ export const LinkMenuOptions = {
     },
     userEvent: "DIALOG_OPEN"
   }),
+  ShowFile: (site, index, eventSource, isEnabled, siteInfo, platform) => ({
+    id: _GetPlatformString(platform),
+    icon: "search",
+    action: ac.AlsoToMain({
```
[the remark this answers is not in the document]
sarracini: Ah good call, these should be `Only`

system-addon/lib/HighlightsFeed.jsm
```diff
@@ -260,6 +274,18 @@ this.HighlightsFeed = class HighlightsFeed {
       case at.UNINIT:
         this.uninit();
         break;
+
+      // Relay the downloads actions to DownloadsManager
+      case at.NEW_TAB_LOAD:
```
k88hudson: You don't need to do this, you can just call `this.downloadsManager.onAction(action);` on every action

sarracini:
Chatted with Bryan, and the following changes are to be made:
1. remove the little hover search glass thing in the corner
2. clicking on the card opens it in finder
3. Show in Finder in context menu then becomes Open File
4. add an icon to the card to make it clear that it's a download

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,154 @@
...
+        this.goToDownloadPage(action.data.url);
+        break;
+      case at.REMOVE_DOWNLOAD_FILE:
+        this.removeDownload(action.data.url);
+        break;
+      case at.SHOW_DOWNLOAD_FILE:
+        this.showFile(action.data.url);
+        break;
+      case at.OPEN_DOWNLOAD_FILE:
+        this.openFile(action.data.url);
```
Mardak: Eh.. these 5 actions could just just do something like
```js
  case OPEN:
    downloadsCmd = "open";
    break;
…
if (downloadsCmd) {
  this.findDownloadedItem(action.data.url)[`downloadsCmd_${downloadsCmd}`]();
}
```

Too clever? ;)

system-addon/content-src/components/Card/Card.jsx
```diff
@@ -139,14 +169,15 @@ export class Card extends React.PureComponent {
...
+    return (<li className={`card-outer${isContextMenuOpen ? " active" : ""}${props.placeholder ? " placeholder" : ""}`} onMouseLeave={this.onMouseOut} onMouseEnter={this.onMouseOver}>
+      <a href={link.type === "pocket" ? link.open_url : link.url} onClick={!props.placeholder ? this.onLinkClick : undefined} >
         <div className="card">
           {hasImage && <div className="card-preview-image-outer">
             <div className={`card-preview-image${this.state.imageLoaded ? " loaded" : ""}`} style={imageStyle} />
           </div>}
           <div className={`card-details${hasImage ? "" : " no-image"}`}>
-            {link.hostname && <div className="card-host-name">{link.hostname}</div>}
+            {link.type === "download" && <div className="card-download-icon icon icon-download-folder" />}
+            {link.hostname && <div className="card-host-name">{this.computeHostname()}</div>}
```
Mardak: The Card component has been relatively generic allowing custom behavior via additional optional properties as opposed to type checking.

Alternatively, it looks like the Download Card looks and behaves a bit differently from existing Cards, so maybe it should just be its own Component?

system-addon/lib/PlacesFeed.jsm
```diff
@@ -313,7 +313,9 @@ class PlacesFeed {
         this.saveToPocket(action.data.site, action._target.browser);
         break;
       case at.OPEN_LINK: {
-        this.openLink(action);
+        if (action.data.type !== "download") {
```
Mardak: This check / change shouldn't be needed anymore.

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,168 @@
...
+
+  addStore(store) {
+    this._store = store;
+  }
+
+  formatDownload(element) {
+    const downloadedItem = element.download;
+    return {
+      hostname: new URL(downloadedItem.source.url).hostname,
+      url: downloadedItem.source.url,
```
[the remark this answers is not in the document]
Mardak: Did design have any comments on what should show in the status bar on hover?

system-addon/content-src/components/Card/Card.jsx
Mardak: @k88hudson any thoughts on if we should care about how generic Card is? It also would potentially affect web extensions that happen to have use a "download" type.. and more generally of special type behaviors.

sarracini:
A note for myself: Make sure the private browsing downloads don't show up

system-addon/content-src/lib/link-menu-options.js
```diff
@@ -10,6 +10,13 @@ const _OpenInPrivateWindow = site => ({
   userEvent: "OPEN_PRIVATE_WINDOW"
 });
 
+export const GetPlatformString = platform => {
+  if (platform === "win") { return "menu_action_show_file_windows"; }
+  if (platform === "macosx") { return "menu_action_show_file_mac_os"; }
+  if (platform === "linux") { return "menu_action_show_file_linux"; }
```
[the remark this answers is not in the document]
sarracini: i honestly don't know why i can never remember that switch cases are a thing. will switch it!

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,154 @@
...
+      path: downloadedItem.target.path,
+      title: element.displayName,
+      description: `${element.sizeStrings.stateLabel} \u2014 ${element.fileType}`,
```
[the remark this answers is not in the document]
sarracini: Yeah just a long dash and strings on either side so i think this should be fine?

system-addon/lib/DownloadsManager.jsm
sarracini: Yeah I thought about this too. I don't mind being too clever ;)

sarracini:
Also @Mardak design said they didn't care about how the status bar shows the url

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,154 @@
...
+  get download() {
+    return this._download;
+  }
+
+  get downloadedSince() {
+    return (Date.now() - this.download.endTime);
+  }
+
+  get fileType() {
+    return this.download.target.path.split(".")[1].toUpperCase();
```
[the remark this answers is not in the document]
sarracini: I couldn't find anywhere that actually got the file type for the download. there's also nothing available on the downloaded item to determine what the type is. We can always take the last element in the split? at least we would get an extension that way?

system-addon/data/content/assets/glyph-download-icon.svg
```diff
@@ -0,0 +1,16 @@
...
+        <path d="M27,16 L21.151,16 L19.584,14.538 C19.2135838,14.1922221 18.7257252,13.9999379 18.219,14 L15,14 C13.8954305,14 13,14.8954305 13,16 L13,26 C13,27.1045695 13.8954305,28 15,28 L27,28 C28.1045695,28 29,27.1045695 29,26 L29,18 C29,16.8954305 28.1045695,16 27,16 Z M18.219,16 L19.291,17 L15,17 L15,16 L18.219,16 Z M27,26 L15,26 L15,18 L21,18 L21,17.986 C21.05,17.986 21.1,18 21.151,18 L27,18 L27,26 Z" id="path-1"></path>
+    </defs>
+    <g id="Icons--/-Download-Folder" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <rect id="Circle" fill="#0A84FF" x="0" y="0" width="42" height="42" rx="21"></rect>
+        <mask id="mask-2" fill="white">
+            <use xlink:href="#path-1"></use>
+        </mask>
+        <use id="Shape" fill="#FFFFFF" fill-rule="nonzero" xlink:href="#path-1"></use>
+    </g>
+</svg>
```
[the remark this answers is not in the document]
sarracini: I showed it to Bryan with dark theme and he said it looks good!

sarracini:
Recent discoveries blocking this bug:
1. [A bug in Places](https://bugzilla.mozilla.org/show_bug.cgi?id=1453353) means that we receive `onDownloadAdded` notifications for every page that we visit. This will most likely show up in profiles, so until that's been optimized for we should hold off on that
2. As per [the second part of this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1453353#c2), we should not be initializing the Download system (i.e calling `DownloadsCommon.getData` on startup, since it does a lot of work and will most likely be bad news. A possible solution would be to dispatch a "downloaded added" notification in `History.cpp -> AddDownload` and listen for that in `DownloadsManager.jsm` and only initialize once we've received it

locales/en-US/strings.properties
```diff
@@ -22,6 +22,7 @@ type_label_visited=Visited
 type_label_bookmarked=Bookmarked
 type_label_recommended=Trending
 type_label_pocket=Saved to Pocket
+type_label_download=Downloaded
```
Mardak: nit: we have `type_label_bookmarked` so let's do "downloaded" in the id.

system-addon/content-src/components/Card/_Card.scss
```diff
@@ -49,6 +49,9 @@
     .card-title {
       color: var(--newtab-link-primary-color);
     }
+
+    .alternate ~ .card-host-name { display: none; }
+    .card-host-name.alternate { display: block; }
```
Mardak: nit: bracing new lines

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  onAction(action) {
+    let downloadsCmd;
+    switch (action.type) {
+      case at.COPY_DOWNLOAD_LINK:
+        downloadsCmd = "copyLocation";
+        break;
+      case at.GO_TO_DOWNLOAD_PAGE:
+        downloadsCmd = "openReferrer";
```
Mardak: This seems to trigger an exception `this.element.ownerGlobal.openURL is not a function DownloadsViewUI.jsm:453`﻿ 
https://searchfox.org/mozilla-central/rev/59a9a86553e9bfd9277202748ff791fd9bc0713b/browser/components/downloads/DownloadsViewUI.jsm#452-453

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,146 @@
...
+    this._downloadData = DownloadsCommon.getData(this._browser.ownerGlobal, true, false, true);
+    this._downloadData.addView(this);
+  }
+
+  onDownloadAdded(download) {
+    const elem = new DownloadElement(download, this._browser);
+    const downloadedItem = elem.download;
+    if (!this._viewableDownloadItems.has(downloadedItem.source.url)) {
+      this._viewableDownloadItems.set(downloadedItem.source.url, elem);
+      this._store.dispatch({type: at.DOWNLOAD_CHANGED});
```
Mardak: Looks like this is called when it first starts a download or an existing download. We probably want handle downloads that were just started and then completed to get the correct file size, otherwise:
image

Dismissing some page forces a refresh and it correctly shows:
image

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,146 @@
...
+    let downloadsCmd;
+    switch (action.type) {
+      case at.COPY_DOWNLOAD_LINK:
+        downloadsCmd = "copyLocation";
+        break;
+      case at.GO_TO_DOWNLOAD_PAGE:
+        downloadsCmd = "openReferrer";
+        break;
+      case at.REMOVE_DOWNLOAD_FILE:
+        downloadsCmd = "delete";
```
Mardak: Looks like it's possible for this to throw if the file was deleted.. although not sure why it doesn't show up as disabled?
```
[Exception... "Component returned failure code: 0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST) [nsIFile.isExecutable]"  nsresult: "0x80520006 (NS_ERROR_FILE_TARGET_DOES_NOT_EXIST)"  location: "JS frame :: resource:///modules/DownloadsCommon.jsm :: openDownloadedFile :: line 423"  data: no]
```
screen shot 2018-04-18 at 2 39 19 pm

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,146 @@
...
+        break;
+      case at.UNINIT:
+        this.uninit();
+        break;
+    }
+    // Call the appropriate downloads command function based on the event we received
+    if (downloadsCmd) {
+      let elem = this._viewableDownloadItems.get(action.data.url);
+      if (elem) {
+        elem[`downloadsCmd_${downloadsCmd}`]();
```
Mardak: I've learned that dynamically building keys for strings, methods, etc. is nice looking but makes it hard to find, e.g., from searchfox. "Are there any more callers of downloadsCmd_delete? No? Baleeted!" Sorry for giving the example with string building in the previous comment. :(

system-addon/content-src/components/Card/Card.jsx
```diff
@@ -146,6 +154,8 @@ export class Card extends React.PureComponent {
             <div className={`card-preview-image${this.state.imageLoaded ? " loaded" : ""}`} style={imageStyle} />
           </div>}
           <div className={`card-details${hasImage ? "" : " no-image"}`}>
+            {link.type === "download" && <div className="card-download-icon icon icon-download-folder" />}
+            {link.type === "download" && <div className="card-host-name alternate"><FormattedMessage id={GetPlatformString(this.props.platform)} defaultMessage={GetPlatformString("default")} /></div>}
```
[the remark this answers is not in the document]
sarracini: Yup we can remove this, I think it was laying around from before when I didn't have it always showing an alternate host 👍

system-addon/lib/DownloadsManager.jsm
sarracini: 😂 no worries!

sarracini:
Latest changes doesn't fix the whole deleted from system still showing up in highlights situation, but addresses all the other changes

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,179 @@
...
+    this._browser = Services.appShell.hiddenDOMWindow;
+    this._downloadData = DownloadsCommon.getData(this._browser.ownerGlobal, true, false, true);
+    this._downloadData.addView(this);
+  }
+
+  onDownloadAdded(download) {
+    const elem = new DownloadElement(download, this._browser);
+    const downloadedItem = elem.download;
+    if (!this._viewableDownloadItems.has(downloadedItem.source.url)) {
+      if (downloadedItem.succeeded && downloadedItem.target.exists) {
```
Mardak: Between these two if blocks, add `await downloadedItem.refresh();` and that should get us the correct `exists` value.

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,179 @@
...
+    if (this._viewableDownloadItems.has(download.source.url)) {
+      this._viewableDownloadItems.delete(download.source.url);
+      this._store.dispatch({type: at.DOWNLOAD_CHANGED});
+    }
+  }
+
+  get downloads() {
+    let results = [];
+    for (const url of this._viewableDownloadItems.keys()) {
+      const elem = this._viewableDownloadItems.get(url);
```
Mardak: nit: `for (const elem of .values())` as we don't use the url

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,179 @@
...
+    if (this._downloadData) {
+      this._downloadData.removeView(this);
+      this._downloadData = null;
+    }
+  }
+
+  copyLocation(url) {
+    let elem = this._viewableDownloadItems.get(url);
+    if (elem && elem.download.target.exists) {
+      elem.downloadsCmd_copyLocation();
```
Mardak: As noted on irc, we can do the method strings just not string concatenation. And we shouldn't need a separate exists check. If the user does delete the file while the card is already showing... oh well?

system-addon/lib/DownloadsManager.jsm
Mardak: Probably add a comment too about why

system-addon/lib/DownloadsManager.jsm
```diff
@@ -98,17 +96,31 @@ this.DownloadsManager = class DownloadsManager {
...
-      if (elem && (elem.downloadedSince < RECENT_DOWNLOAD_THRESHOLD)) {
-        const formattedDownloadForHighlights = this.formatDownload(elem);
-        results.push(formattedDownloadForHighlights);
+    for (const elem of this._viewableDownloadItems.values()) {
+      // Only get downloads within the time threshold specified
+      if (elem && (elem.downloadedSince < threshold)) {
+        // If we only care about downloads where the file still exists, refresh them
+        // to ensure the 'exists' attribute is up to date
+        if (onlyExists) {
+          await elem.download.refresh();
```
Mardak: As discussed on irc, I'm assuming `refresh`, which results in file IO, is significantly slower than anything else we do in this method whether that's iterating or sorting, so we want to minimize the number of times we call it. This can be done by only checking the first few most recent downloads in order, so if the first one exists, we're done.

system-addon/lib/DownloadsManager.jsm
```diff
@@ -98,17 +96,31 @@ this.DownloadsManager = class DownloadsManager {
...
     let results = [];
-    for (const url of this._viewableDownloadItems.keys()) {
-      const elem = this._viewableDownloadItems.get(url);
-      // We only want downloads that are less than 36 hours old
-      if (elem && (elem.downloadedSince < RECENT_DOWNLOAD_THRESHOLD)) {
-        const formattedDownloadForHighlights = this.formatDownload(elem);
-        results.push(formattedDownloadForHighlights);
+    for (const elem of this._viewableDownloadItems.values()) {
+      // Only get downloads within the time threshold specified
+      if (elem && (elem.downloadedSince < threshold)) {
```
Mardak: Also just documenting from irc. `elem` shouldn't ever be falsey now. And this comparison expands to `Date.now() - endTime < threshold` and gets computed for every download. We can shuffle things around to be `Date.now() - threshold < endTime` which makes the math loop invariant, so compute a `dateThreshold = Date.now() - threshold` once before the loop and check `endTime > dateThreshold`

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,183 @@
...
+                      .filter(elem => elem.download.endTime > Date.now() - threshold)
+                      .sort((elem1, elem2) => elem1.download.endTime < elem2.download.endTime);
+
+    // Iterate over this list checking for ones that exist, and formatting the ones
+    // to return
+    for (const elem of downloads) {
+      if (onlyExists) {
+        // Refresh download to ensure the 'exists' attribute is up to date
+        await elem.download.refresh();
+        if (elem.download.target.exists && elem.download.succeeded) {
```
Mardak: you can `if (!exists) continue;` and share the `formatDownload` / `else` stuff. Both paths want only succeeded downloads too?

sarracini: maybe pass `succeeded` as a parameter too?

Mardak: I suppose an optional succeeded parameter would match and doesn't need to add an extra branch for test coverage

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,183 @@
...
+    return items[items.length - 1].toUpperCase();
+  }
+
+  downloadsCmd_copyLocation() {
+    let clipboard = Cc["@mozilla.org/widget/clipboardhelper;1"].getService(Ci.nsIClipboardHelper);
+    clipboard.copyString(this.download.source.url);
+  }
+
+  downloadsCmd_openReferrer() {
+    this.element.openNewTabWith(this.download.source.referrer);
```
Mardak: Looks like this opens in a background tab. Maybe just pass in true as the second argument for shifted?

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,183 @@
...
+    this._downloadData.addView(this);
+  }
+
+  onDownloadAdded(download) {
+    const elem = new DownloadElement(download, this._browser);
+    const downloadedItem = elem.download;
+    if (!this._viewableDownloadItems.has(downloadedItem.source.url)) {
+      this._viewableDownloadItems.set(downloadedItem.source.url, elem);
+
+      // Debounce incoming onDownloadAdded notifications
```
Mardak: Probably note this is for startup where all existing downloads are added

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,183 @@
...
+
+  async getDownloads({threshold, numItems = this._viewableDownloadItems.size, onlyExists = false}) {
+    if (!threshold) {
+      return [];
+    }
+    let results = [];
+
+    // Only get downloads within the time threshold specified and sort by recency
+    let downloads = [...this._viewableDownloadItems.values()]
+                      .filter(elem => elem.download.endTime > Date.now() - threshold)
```
Mardak: nit: `Date.now() - threshold` defined outside of the callback.

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  onDownloadRemoved(download) {
+    if (this._viewableDownloadItems.has(download.source.url)) {
+      this._viewableDownloadItems.delete(download.source.url);
+      this._store.dispatch({type: at.DOWNLOAD_CHANGED});
+    }
+  }
+
+  async getDownloads({threshold, numItems = this._viewableDownloadItems.size, onlyExists = false}) {
```
Mardak: If `threshold` is required, probably make it as the first parameter and have the second the options object

system-addon/lib/DownloadsManager.jsm
```diff
@@ -0,0 +1,183 @@
...
+      }
+      if (results.length === numItems) {
+        break;
+      }
+    }
+
+    return results;
+  }
+
+  uninit() {
```
Mardak: Probably want to cancel the timer to be clean

system-addon/content-src/lib/link-menu-options.js
```diff
@@ -91,6 +104,49 @@ export const LinkMenuOptions = {
...
     userEvent: "DIALOG_OPEN"
   }),
+  ShowFile: (site, index, eventSource, isEnabled, siteInfo, platform) => ({
+    id: GetPlatformString(platform),
+    icon: "search",
+    action: ac.OnlyToMain({
+      type: at.SHOW_DOWNLOAD_FILE,
+      data: {url: site.url}
+    }),
+    disabled: !site.path
```
Mardak: I think we don't need this disabled anymore? Same for OpenFile

PR #4081, opened 2018-04-05
Fix Bug 1451419 - set background of Top Site icons to $white

rlr:
I am confirming with Amy that this should also apply to the default theme.

Before:
screen shot 2018-04-05 at 12 24 44 pm
screen shot 2018-04-05 at 12 24 08 pm

After:
screen shot 2018-04-05 at 12 24 58 pm
screen shot 2018-04-05 at 12 23 58 pm

### DIFF
--- a/system-addon/content-src/components/TopSites/_TopSites.scss
+++ b/system-addon/content-src/components/TopSites/_TopSites.scss
@@ -144,7 +144,7 @@ $half-base-gutter: $base-gutter / 2;
 
   // Some common styles for all icons (rich and default) in top sites
   .top-site-icon {
-    background-color: var(--newtab-background-color);
+    background-color: $white;
     background-position: center center;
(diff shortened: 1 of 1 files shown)


rlr:
Amy confirmed the change to white for default theme too.

piatra:
@rlr does this mean we want the background color of cards to also be white?

rlr:
The card image background color changed this week but not to white 30238bea90cb819cda0bee57cb23a0b812746274

Are there any cases where those images have a transparent background? If so, it's probably pretty rare?

as-pine-proxy:
[Treeherder: 2a10c65cdf1ad7c7bc580687c8c00523bae1828a](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=2a10c65cdf1ad7c7bc580687c8c00523bae1828a)

PR #4082, opened 2018-04-05
Fix Bug 1451089 - make context menu styles consistent with doorhangers

rlr:
Before:
screen shot 2018-04-05 at 2 39 32 pm
screen shot 2018-04-05 at 2 39 16 pm

After:
screen shot 2018-04-05 at 2 32 43 pm
screen shot 2018-04-05 at 2 43 01 pm

PR #4083, opened 2018-04-05
Fix Bug 1451474 - color updates for search drop down

rlr:
Before:
screen shot 2018-04-05 at 4 14 58 pm

After:
screen shot 2018-04-05 at 4 13 30 pm

### DIFF
--- a/system-addon/content-src/components/Search/_Search.scss
+++ b/system-addon/content-src/components/Search/_Search.scss
@@ -1,6 +1,6 @@
 .search-wrapper {
   $search-border-radius: 3px;
-  $search-focus-color: var(--newtab-search-focus-border-color);
+  $search-focus-color: $blue-60;
   $search-height: 35px;
   $search-input-left-label-width: 35px;
   $search-button-width: 36px;
@@ -25,7 +25,6 @@
     border: solid 1px var(--newtab-search-border-color);
     border-radius: $search-border-radius;
     box-shadow: $shadow-secondary, 0 0 0 1px $black-15;
-    color: var(--newtab-search-text-color);
     font-size: 15px;
     padding: 0;
     padding-inline-end: $search-button-width;
@@ -84,8 +83,9 @@
   // Adjust the style of the contentSearchUI-generated table
   // sass-lint:disable-block class-name-format
   .contentSearchSuggestionTable {
-    background-color: var(--newtab-search-background-color);
+    background-color: var(--newtab-modal-color);
     border: 0;
+    box-shadow: $context-menu-shadow;
     transform: translateY(3px);
 
     .contentSearchHeader {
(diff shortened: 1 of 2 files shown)


system-addon/content-src/styles/_theme.scss
```diff
@@ -53,7 +53,7 @@
   --newtab-background-color: $grey-80;
   --newtab-border-primary-color: $grey-10-80;
   --newtab-border-secondary-color: $grey-10-20;
-  --newtab-button-primary-color: $blue-40;
+  --newtab-button-primary-color: $blue-60;
```
rlr: I sneaked this button color change in here. Basically we realized we need blue-60 when it is background color for light text in order to hit the contrast minimum.

system-addon/content-src/styles/_theme.scss
```diff
@@ -34,7 +34,7 @@
   // Search
   --newtab-search-background-color: $white;
   --newtab-search-border-color: transparent;
-  --newtab-search-focus-border-color: $blue-50;
+  --newtab-search-focus-border-color: $blue-60;
```
rlr: we could potentially take away this variable now 😈

Mardak: And just use `--newtab-button-primary-color`? There seems to be an unused `$input-primary` now that can be removed too. Let's remove both and… leave the "button" name? Unless you want to rename them to `--newtab-input-primary-color` (and button-secondary too)?

system-addon/content-src/styles/_theme.scss
```diff
@@ -82,10 +82,10 @@
 
   // Search
   --newtab-search-background-color: $grey-60;
-  --newtab-search-border-color: $grey-10-40;
-  --newtab-search-focus-border-color: $blue-40;
+  --newtab-search-border-color: $grey-10-20;
+  --newtab-search-focus-border-color: $blue-60;
   --newtab-search-icon-color: $grey-10-60;
-  --newtab-search-text-color: $grey-10-60;
+  --newtab-search-text-color: $grey-10;
```
Mardak: `--newtab-search-text-color` can also just be removed now too and use `--newtab-text-primary-color`

Mardak:
Oh, I see #4082 is removing the `$input-primary`

system-addon/content-src/styles/_theme.scss
Mardak: Actually, there's a normalize style that already sets to inherit, so just remove!

system-addon/content-src/styles/_theme.scss
rlr: I was thinking of just setting it to `$blue-60` directly since both themes have the same color now. Good?

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -1,6 +1,6 @@
 .search-wrapper {
   $search-border-radius: 3px;
-  $search-focus-color: var(--newtab-search-focus-border-color);
+  $search-focus-color: $blue-60;
```
rlr: @Mardak cool with this or any better ideas?

Mardak: Part of it is if a theme wanted to actually change it.. but that's fine for now.

as-pine-proxy:
[Treeherder: 7fd026a989dff54e7f0b3785b22da423ae1477c7](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=7fd026a989dff54e7f0b3785b22da423ae1477c7)

PR #4089, opened 2018-04-12
Use moz-osx-font-smoothing: grayscale for dark theme

k88hudson:
What do we think about using `moz-osx-font-smoothing: grayscale` when the dark theme is enabled? it makes some of the type look a lot sharper, especially with the reversed text. Is there any reason we might not want to do this?

image

### DIFF
--- a/system-addon/content-src/styles/_theme.scss
+++ b/system-addon/content-src/styles/_theme.scss
@@ -47,6 +47,7 @@ body {
 
 // Dark theme
 .dark-theme {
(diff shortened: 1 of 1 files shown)


as-pine-proxy:
[Treeherder: 6db3f3e27625827ce0e08823844e5b2b05364c45](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=6db3f3e27625827ce0e08823844e5b2b05364c45)

PR #4090, opened 2018-04-13
Bug 1448918 - Migrate Firefox Desktop's first run experience from a w…

ericawright:
…eb page to an in-product experience.

- Adds an overlay containing an image and prompt to sign in to sync to activity stream when the url is "about:welcome"
- Overlay fades to reveal activity stream when user clicks skip or signs in.
- Overlay should look almost identical to https://www.mozilla.org/en-US/firefox/57.0/firstrun/ (to view properly, make sure you are not signed in to fxa when visiting this page).
- snippets have been disabled on this page

r? @sarracini, @k88hudson 

Bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1448918

TODO: 
- ~tests~
- ~set up with l10n~
- ~if possible, skip if user is already signed in (apparently some users may see this on upgrading)?~ I don't think this is necessary

Notes:
- to test it requires some changes in m-c or, manually change `isFirstrun = true` in `system-addon/content-src/activity-stream.jsx` line 24.


system-addon/content-src/activity-stream.jsx
```diff
@@ -21,6 +21,7 @@ if (!global.gActivityStreamPrerenderedState) {
 
 ReactDOM.hydrate(<Provider store={store}>
   <Base
+    isFirstrun={new global.URL(global.document.location).href === "about:welcome"}
```
sarracini: no need to do URL here, you can just access href directly off location

system-addon/content-src/components/Base/Base.jsx
```diff
@@ -137,7 +140,9 @@ export class BaseContent extends React.PureComponent {
             </div>
             <ConfirmDialog />
           </main>
-        </div>);
+        </div>
+        {showOverlay}
```
sarracini: Just do:
` {props.isFirstRun && <StartupOverlay />}`
and remove the showOverlay stuff

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,87 @@
+import {actionCreators as ac} from "common/Actions.jsm";
+import {connect} from "react-redux";
+import {injectIntl} from "react-intl";
+import React from "react";
+
+export class _StartupOverlay extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.locations = {};
```
sarracini: What is this used for? Doesn't seem to be referenced anywhere

system-addon/content-src/components/StartupOverlay/_StartupOverlay.scss
```diff
@@ -0,0 +1,241 @@
...
+  float: left;
+  clear: both;
+  width: 441px;
+}
+
+.firstrun-content {
+  line-height: 1.5;
+  margin-bottom: 48px;
+  max-width: 352px;
+  background: url('#{$image-path}sync-devices.svg') bottom center no-repeat;
```
sarracini: Run this svg  though `svgo`: https://github.com/svg/svgo

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+  clickSkip() {
+    this.props.dispatch(ac.UserEvent({event: "CLICK_SKIP"}));
+    this.removeOverlay();
+  }
+
+  render() {
+    return (
+      <div className="overlay-wrapper">
+        <canvas className="canvas">
```
sarracini: Is this supposed to be here?

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,87 @@
+import {actionCreators as ac} from "common/Actions.jsm";
+import {connect} from "react-redux";
```
sarracini: I don't think this component needs to be connected

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,87 @@
...
+    this.state = {emailInput: ""};
+  }
+
+  initScene(element) {
+    document.body.dataset.welcome = "true";
+    this.scene = element;
+    // Timout to allow the scene to render once before attaching the attribute
+    // to trigger the animation.
+    setTimeout(() => {
+      this.scene.dataset.content = "true";
```
sarracini: Instead of using data attributes here, can you not just set state here and then add a class in the render function based on the state? And then in the css just check for the class?
```js
this.setState({content: true});
```
and then do:
```html
<div className={"firstrun-scene ${this.state.content ? "show" : "" } "} ref={this.initScene}>
```
Same with `sign-in` data attribute. I say this because generally data attributes are used for stuff that needs actual data, whereas these seem to just be either true or false, which is just as good as adding a class or not.

We do this pattern a bunch: https://github.com/mozilla/activity-stream/blob/master/system-addon/content-src/components/Card/Card.jsx#L142

sarracini: Yeah might as well just make it a div

sarracini:
Just tested out the overlay, it seems like the z-index doesn't beat out snippets: 
screen shot 2018-04-13 at 1 38 36 pm

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
sarracini: Oh nvm you're dispatching

sarracini:
You'll have to change the margins and stuff to use margin-inline-end/start to fix RTL issues:
screen shot 2018-04-13 at 1 50 53 pm

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,87 @@
...
+          <div className="fxaccounts-container">
+            <div className="firstrun-left-divider">
+              <h1 className="firstrun-title">Take Firefox with You</h1>
+              <p className="firstrun-content">Get your bookmarks, history, passwords and other settings on all your devices.</p>
+              <a className="firstrun-link" href="https://www.mozilla.org/firefox/features/sync/" target="_blank" rel="noopener noreferrer">Learn more about Firefox Accounts</a>
+            </div>
+            <div className="firstrun-sign-in">
+              <p className="form-header">Enter your email <span>to continue to Firefox Sync</span></p>
+              <form method="get" action="https://accounts.firefox.com" target="_blank" rel="noopener noreferrer" onSubmit={this.onSubmit}>
+                <input name="service" type="hidden" value="sync" />
```
sarracini: All these inputs are hidden, any reason we can't remove them?

ericawright: Good question, I'll have to ask @vladikoff about that.

vladikoff: Heya, remove if you want, just make sure you pass the params to fxa in some other way so users can login to Sync

sarracini:
Can you use the logo in at `chrome://branding/content/about-logo.png` rather than committing the png?

ericawright:
sure, interesting that chrome://branding/content/about-logo.png changes depending on which build you're using though - not sure if that's a problem or not. Do you know the location of a logo which is only firefox?

ericawright:
@ncloudioj I'm told I should ask you about adding new telemetry events? How'd I do?

locales/en-US/strings.properties
```diff
@@ -171,3 +171,24 @@ section_menu_action_add_topsite=Add Top Site
...
+
+# LOCALIZATION NOTE (firstrun_form_header and firstrun_form_sub_header):
+# firstrun_form_sub_header is a continuation of firstrun_form_header, they are one sentance.
+# firstrun_form_header is displayed more boldly as the call to action.
+firstrun_form_header=Enter your email
+firstrun_form_sub_header=to continue to Firefox Sync.
+
+# LOCALIZATION NOTE (firstrun_extra_legal_links): {terms} is equal to firstrun_terms_of_service, and
+# {privacy} is equal to firstrun_privacy_notice. {terms} and {privacy} are clickable links.
+firstrun_extra_legal_links=By proceding you agree to the {terms} and {privacy}.
```
vladikoff: typo: `proceding` -> `proceeding`

FxA also has a comma after the word: 
 

Not sure if you want that here

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,86 @@
...
+    this.onInputChange = this.onInputChange.bind(this);
+    this.onSubmit = this.onSubmit.bind(this);
+    this.clickSkip = this.clickSkip.bind(this);
+    this.initScene = this.initScene.bind(this);
+
+    this.state = {emailInput: ""};
+  }
+
+  initScene() {
+    // Timout to allow the scene to render once before attaching the attribute
```
vladikoff: typo: `Timout` -> `Timeout`

locales/en-US/strings.properties
ericawright: yeah, I want it as identical as possible

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,86 @@
...
+              <h1 className="firstrun-title"><FormattedMessage id="firstrun_title" /></h1>
+              <p className="firstrun-content"><FormattedMessage id="firstrun_content" /></p>
+              <a className="firstrun-link" href="https://www.mozilla.org/firefox/features/sync/" target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_learn_more_link" /></a>
+            </div>
+            <div className="firstrun-sign-in">
+              <p className="form-header"><FormattedMessage id="firstrun_form_header" /><span><FormattedMessage id="firstrun_form_sub_header" /></span></p>
+              <form method="get" action="https://accounts.firefox.com" target="_blank" rel="noopener noreferrer" onSubmit={this.onSubmit}>
+                <input name="service" type="hidden" value="sync" />
+                <input name="forceExperiment" type="hidden" value="emailFirst" />
+                <input name="forceExperimentGroup" type="hidden" value="treatment" />
```
vladikoff: @shane-tomlinson we can replace `forceExperiment` and `forceExperimentGroup` with `action=email` right? confirm?

ncloudioj:
@ericawright Looks like you've already added two events "SKIPPED_SIGNIN" and "SUBMIT_EMAIL". If these are all you wanted to capture in telemetry, then it's done. Since this only adds two more values to the existing event ping, it doesn't even need the data review. You're all set :)

ericawright:
@sarracini https://treeherder.mozilla.org/#/jobs?repo=try&revision=5fc7903445b7d565956db612be50cf5aae1bd03b
most recent try. (in progress at the moment)

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
shane-tomlinson: Yup, I'm a bit surprised forceExperiment and forceExperimentGroup are still being used in fact.

ericawright: @vladikoff can you give me an example of what you mean by `action=email`?

ericawright:
@sarracini, @dmose pointed out in the bugzilla bug that there may be some more telemetry to change. "there's some code in TelemetryFeed.jsm that assumes that assumes that if you're not on about:home, that means you must be on about:newtab."
Do you know of any additional telemetry that should be changed?
In TelemetryFeed.jsm line 446 I'm suspicious of, but not sure if I should change.

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
shane-tomlinson: @ericawright - The `force*` query parameters were meant for testing, sorry if I never told you to update to `action=email` before! Instead of having the two inputs with the name of `forceExperiment` and `forceExperimentGroup`, use a single input with `name=action` and `value=email`.

```html
<input name="action" type="hidden" value="email" />
```

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+  initScene() {
+    // Timeout to allow the scene to render once before attaching the attribute
+    // to trigger the animation.
+    setTimeout(() => {
+      this.setState({show: true});
+    }, 10);
+  }
+
+  removeOverlay() {
```
sarracini: You have to bind this function in the constructor, `this.setState` won't work

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,86 @@
...
+    }, 400);
+  }
+
+  onInputChange(e) {
+    this.setState({emailInput: e.target.value});
+  }
+
+  onSubmit() {
+    this.props.dispatch(ac.UserEvent({event: "SUBMIT_EMAIL"}));
+    window.addEventListener("focus", this.removeOverlay);
```
sarracini: can you use `visibilitychange` event instead of `focus` ? and also remove the `eventListener` once you've removed the overlay, or else this will fire every time you focus back to the page

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
ericawright: 👍

dmose:
@ericawright 446 was indeed the line I was thinking of.   Also relevant, we'll probably have to decide how to send these pages as well as the timing, I could imagine we'd want to send them as about:welcome, but perhaps @ncloudioj has some other thoughts.  If so, we'll need to update our data docs (see https://github.com/mozilla/activity-stream/blob/1818ebe6385b8ce96d7be0fbbb83e03adab43a62/docs/v2-system-addon/telemetry.md for details).  That might make sense to do in a separate PR, however.

However, I should have been more clear: besides just in telemetry, I suspect there are other things in the code base that make assumptions about this.   I'd suggest using searchfox to search for both about:home and about:newtab, as that should cover code in this repo (as browser/extensions/activity-stream/) as well as stuff that lives elsewhere in mozilla-central.

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
sarracini: `focus` isn't really used much anymore since HTML5 uses the page visibility API. Plus `focus` could be fired if the browser loses focus other ways, but that doesn't mean it's not visible. `visibilitychange` feels like the more suited event

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,86 @@
...
+  initScene() {
+    // Timeout to allow the scene to render once before attaching the attribute
+    // to trigger the animation.
+    setTimeout(() => {
+      this.setState({show: true});
+    }, 10);
+  }
+
+  removeOverlay() {
+    this.setState({signIn: true});
```
sarracini: Is it possible to re-use the `show` state here and just set it to false? Removing the overlay is basically just like not showing it anymore, so re-using the same `show` state seems to make sense

ericawright: I think so, will require some css changes to get the animations right

ericawright:
The csp workaround has been removed after conversation in https://bugzilla.mozilla.org/show_bug.cgi?id=1448918, it will be added to the whitelist and wait until the csp is fixed for activity stream.

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,87 @@
...
+            <div className="firstrun-left-divider">
+              <h1 className="firstrun-title"><FormattedMessage id="firstrun_title" /></h1>
+              <p className="firstrun-content"><FormattedMessage id="firstrun_content" /></p>
+              <a className="firstrun-link" href="https://www.mozilla.org/firefox/features/sync/" target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_learn_more_link" /></a>
+            </div>
+            <div className="firstrun-sign-in">
+              <p className="form-header"><FormattedMessage id="firstrun_form_header" /><span><FormattedMessage id="firstrun_form_sub_header" /></span></p>
+              <form method="get" action="https://accounts.firefox.com" target="_blank" rel="noopener noreferrer" onSubmit={this.onSubmit}>
+                <input name="service" type="hidden" value="sync" />
+                <input name="action" type="hidden" value="email" />
```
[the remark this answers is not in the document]
ericawright: I am suspicious some of this may still change depending on fxa analytics in the next few weeks anyway... once they change moz.org to email-first-flow.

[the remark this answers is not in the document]
sarracini: just leave them then

system-addon/content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -0,0 +1,89 @@
...
+                <input name="context" type="hidden" value="fx_desktop_v3" />
```
shane-tomlinson: I realize this was already merged and I'm late to the party, some additional parameters would be great here so we can track where users are coming from, things like `entrypoint` and `utm_*` parameters.

shane-tomlinson: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1462569

PR #4094, opened 2018-04-16
Fix Bug 1454423 - update secondary text color to pass contrast test

rlr:
* also add a tertiary text color (only currently used for card context text)
* secondary and tertiary text color are the same in default theme for now

Before:
screen shot 2018-04-16 at 3 00 27 pm
screen shot 2018-04-16 at 3 00 19 pm

After:
screen shot 2018-04-16 at 2 59 31 pm
screen shot 2018-04-16 at 2 59 39 pm

### DIFF
--- a/system-addon/content-src/styles/_theme.scss
+++ b/system-addon/content-src/styles/_theme.scss
@@ -26,6 +26,7 @@ body {
   --newtab-text-conditional-color: $grey-60;
   --newtab-text-primary-color: $grey-90;
   --newtab-text-secondary-color: $grey-50;
+  --newtab-text-tertiary-color: $grey-50;
   --newtab-textbox-background-color: $white;
   --newtab-textbox-border: $grey-90-20;
   @include textbox-focus($blue-60); // sass-lint:disable-line mixins-before-declarations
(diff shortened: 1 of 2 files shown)


system-addon/content-src/styles/_theme.scss
```diff
@@ -77,7 +78,8 @@ body {
   --newtab-link-secondary-color: $pocket-teal;
   --newtab-text-conditional-color: $grey-10;
   --newtab-text-primary-color: $grey-10;
-  --newtab-text-secondary-color: $grey-10-40;
+  --newtab-text-secondary-color: $grey-10-80;
+  --newtab-text-tertiary-color: $grey-10-40;
```
Mardak: I would think whatever contrast test is failing for existing usage of secondary-color would then continue to fail for the context now using tertiary color. If it's a usability issue, why is it okay to let the context disappear?

rlr:
Here is the list of what comes next (I had to look it up!): 

> The sequence continues with quaternary, quinary, senary, septenary, octonary, nonary, and denary, although most of these terms are rarely used.

as-pine-proxy:
[Treeherder: add99548f490eeec1afe882e2bb88c02eae67b77](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=add99548f490eeec1afe882e2bb88c02eae67b77)

PR #4096, opened 2018-04-17
WIP - Bug 1449338 - Replace search icon with logo of user's search engine

daleharvey:
People want this bug to land this week for uplift so looking for some early feedback

This seems to mostly work, the only issue I have found is when wikipedia is set, the icon url is `resource://search-plugins/images/wikipedia.ico` which doesnt seem to be able to be accessed from content?

Also needs tests, but would like confirmation that I am on the right path before adding them


daleharvey:
Closed as prefer other patch

PR #4097, opened 2018-04-17
Fix Bug 1449338 - Show currently selected engine in newtabs search input

daleharvey:
This updates the CSS to handle the change in https://reviewboard.mozilla.org/r/237204/diff/#index_header

### DIFF
--- a/system-addon/content-src/components/Search/_Search.scss
+++ b/system-addon/content-src/components/Search/_Search.scss
@@ -1,12 +1,11 @@
 .search-wrapper {
   $search-height: 35px;
+  $search-icon-size: 18px;
+  $search-icon-padding: 8px;
+  $search-icon-width: 2 * $search-icon-padding + $search-icon-size;
   $search-input-left-label-width: 35px;
   $search-button-width: 36px;
-  $search-glyph-image: url('chrome://browser/skin/search-glass.svg');
   $glyph-forward: url('chrome://browser/skin/forward.svg');
-  $search-glyph-size: 16px;
-  // This is positioned so it is visually (not metrically) centered. r=abenson
-  $search-glyph-left-position: 12px;
 
   cursor: default;
   display: flex;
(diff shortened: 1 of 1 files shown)


Mardak:
Testing with the patch from bug 1449338 applied with default and dark theme:
screen shot 2018-04-18 at 10 37 31 am
screen shot 2018-04-18 at 10 40 27 am
screen shot 2018-04-18 at 10 41 02 am
screen shot 2018-04-18 at 10 40 01 am
screen shot 2018-04-18 at 10 43 54 am
screen shot 2018-04-18 at 10 39 25 am

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -2,11 +2,7 @@
   $search-height: 35px;
   $search-input-left-label-width: 35px;
```
Mardak: This should be renamed to something referring to icon. And is the sizing still correct given that the icon size is increasing from 16px to 18px? Perhaps we want something like
```scss
$search-icon-size: 18px;
$search-icon-padding: 8px;
$search-icon-width: 2 * $search-icon-padding + $search-icon-size;
```

daleharvey:
Thanks for the review, updated and used the new vars in their appropriate places, looks cleaner now too, cheers

system-addon/content-src/components/Search/_Search.scss
```diff
@@ -16,13 +15,15 @@
   width: 100%;
 
   input {
-    background: var(--newtab-textbox-background-color);
+    background: var(--newtab-textbox-background-color) var(--newtab-search-icon) $search-icon-padding center / $search-icon-size no-repeat;
```
Mardak: Oh oops! Almost forgot that the icon was centered before in the label, but now this is anchored to the left offset by $padding. For rtl, we'll want to offset from the right with the 2-value form: `background-position-x: right $search-icon-padding;`

image

as-pine-proxy:
[Treeherder: efca435a916232d0de94ab217100519e2af727d5](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=efca435a916232d0de94ab217100519e2af727d5)

PR #4106, opened 2018-04-18
Fix Bug 1455081 - update icon and text tertiary colors in dark theme

rlr:
Only variable value tweaking now. I see the light at the end of the tunnel 😅 

Before:
screen shot 2018-04-18 at 4 00 38 pm

After:
screen shot 2018-04-18 at 3 40 45 pm


rlr:
Yeah. Lots of potential for consolidation. We kind of need a tool that allows reassigning variables and changing their values while seeing how in real time how it affects both themes? Otherwise, we end up with this explosion of variables.

PR #4110, opened 2018-04-19
Add telemetry pings (part 1 - UI instrumentation)

k88hudson:
This patch adds instrumentation for user event telemetry to AS router templates – nothing is actually done with the events so far, it just adds a simple ping action type.
The first commit is just some style and functionality changes to support actual links.


system-addon/content-src/message-center/components/SnippetBase.jsx
```diff
@@ -6,11 +6,14 @@ const defaultStyles = {
     bottom: 0,
     left: 0,
     right: 0,
-    padding: "20px",
-    backgroundColor: "white",
+    padding: "12px",
+    backgroundColor: "var(--newtab-card-background-color)",
```
k88hudson: @rlr do you think this is a good option for the snippets background so it works with dark theme?

ncloudioj:
r+ for the telemetry part. Thanks @k88hudson  for taking the first crack at it, it looks great!

@rlr Since you have been tagged for the UI change, I'll pass it to you for the UI changes :)

k88hudson:
Thanks nan! I will split off the UI commit in another PR so it's easier for ricky to review :+1:

system-addon/content-src/message-center/components/SnippetBase.jsx
rlr: 👍 as long as you are OK with the color being coupled to the card color.

PR #4111, opened 2018-04-19
Fix Bug 1432595 - Add section in prefs to turn off parts of Highlights

Mardak:
This should probably land after #4076 but doesn't really need to. The patches don't actually conflict because the other PR adds "includeDownloads" in a non-conflicting area.

r? @sarracini This doesn't have the logic where if everything is unchecked, Highlights should get unchecked... for a followup!

screen shot 2018-04-19 at 2 48 10 pm


as-pine-proxy:
[Treeherder: 5fd675125d53d53ba5ceb12c829599ff9171ad89](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=5fd675125d53d53ba5ceb12c829599ff9171ad89)

locales/en-US/strings.properties
```diff
@@ -102,6 +102,9 @@ prefs_topstories_description2=Great content from around the web, personalized fo
 prefs_topstories_options_sponsored_label=Sponsored Stories
 prefs_topstories_sponsored_learn_more=Learn more
 prefs_highlights_description=A selection of sites that you’ve saved or visited
+prefs_highlights_options_visited_label=Visited Pages
+prefs_highlights_options_download_label=Most Recent Download
```
flodolo: Question from a localizer: should this be downloads (plural)?

Mardak: This is singular as it shows only one download

flodolo: Ah, thanks! (cc @ TonnesM)

TonnesM: Yeah, I noticed by testing meanwhile but thanks for the reminder. ;)

Side note: we used l10n for "Most recently downloaded" which a) covers both singular and plural (at least for my locale), and b) circumvents the use of "download" as a noun. Note that some locales (de, es) currently do use translations for "downloads" (noun, plural).

PR #4113, opened 2018-04-20
Disable downloads in highlights

sarracini:
Just in case


sarracini:
i fixed the leak so no need to disable!

PR #4107, opened 2018-04-18
Fix Bug 1454985 - use ::after for search provider dividers

rlr:
And... this is LTR friendly now 😄 

Before:
screen shot 2018-04-18 at 4 56 48 pm

After:
screen shot 2018-04-18 at 4 55 51 pm

PR #4115, opened 2018-04-22
fix(styles): Switch to border-image instead of ::after for search engine separator

Mardak:
Followup as suggested in https://github.com/mozilla/activity-stream/pull/4107#issuecomment-383393619

r? @rlr Turns out the icons weren't actually centered with `::after` but now they're exactly 32px from the line on each side!

before:
screen shot 2018-04-22 at 11 29 04 am

after:
screen shot 2018-04-22 at 11 44 29 am

(Open up both images in tabs and switch between the two tabs to check! ;))

### DIFF
--- a/system-addon/content-src/components/Search/_Search.scss
+++ b/system-addon/content-src/components/Search/_Search.scss
@@ -127,7 +127,12 @@
     }
 
     .contentSearchOneOffItem {
+      // Make the border slightly shorter by offsetting from the top and bottom
+      $border-offset: 18%;
+
       background-image: none;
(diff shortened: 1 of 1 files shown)


Mardak:
https://developer.mozilla.org/en-US/docs/Web/CSS/border-image
https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient

At least I wasn't familiar with the border-image/-slice part

Mardak:
I wonder if design wants a fancy gradiented border now instead of the sharp cutoff ;) :p

as-pine-proxy:
[Treeherder: 18a081bf8ffe9c33d8c58b5484544869f5f62dde](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=18a081bf8ffe9c33d8c58b5484544869f5f62dde)

PR #4117, opened 2018-04-23
Fix modal icon width

sarracini:
Before:
screen shot 2018-04-23 at 1 47 06 pm

After:
screen shot 2018-04-23 at 2 06 36 pm

### DIFF
--- a/system-addon/content-src/styles/_icons.scss
+++ b/system-addon/content-src/styles/_icons.scss
@@ -39,6 +39,7 @@
   }
 
   &.icon-modal-delete {
+    flex-shrink: 0;
     background-image: url('#{$image-path}glyph-modal-delete-32.svg');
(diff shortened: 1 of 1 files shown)


system-addon/content-src/styles/_icons.scss
```diff
@@ -39,6 +39,8 @@
   }
 
   &.icon-modal-delete {
+    flex-basis: auto;
```
rlr: It looks like this is already the default so you can remove it. The line below alone fixes it for me.

as-pine-proxy:
[Treeherder: 615804de25dd3d5e7def5bce9547e59c790c9c03](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=615804de25dd3d5e7def5bce9547e59c790c9c03)

PR #4122, opened 2018-04-24
Fix Bug 1456445 - The 'Go to download page' context menu option does not do anything on Windows and Linux

sarracini:
Apparently, in Windows and Linux, there is no `openNewTabWith` function available on hidden windows. Which is what `this.element` is. This way should work on all platforms


sarracini:
Yes, I'll need to make some changes so that it can pass "tabshifted" as part of the action so we can focus to the tab we opened

sarracini:
Design says it's ok that opening the referrer url for a download item replaces activity stream - no extra changes to `openLink` necessary then

as-pine-proxy:
[Treeherder: 07d34bb3faa164cbd804660677e9772b8be52b43](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=07d34bb3faa164cbd804660677e9772b8be52b43)

PR #4126, opened 2018-04-26
Fix tests that would never fail

ericawright:
When writing tests recently I noticed a few that are written to never fail. The `wrapper`, even when empty has a truthy value and will pass an `assert.ok`, this is also the case for `wrapper.find('foo')`.

In `test/unit/content-src/components/Card.test.jsx`, line 67, this was passing because of the `wrapper` having a value, when in fact the link menu has never existed. This is directly contradicted by line 71, which ensures it does not exist. This is why I removed it.

`test/unit/content-src/components/ConfirmDialog.test.jsx` line 26, it fails to find ConfirmDialog, but will succeed in finding '.confirmation-dialog` - I'm assuming those are intended as the same thing.

### DIFF
--- a/system-addon/test/unit/content-src/components/Card.test.jsx
+++ b/system-addon/test/unit/content-src/components/Card.test.jsx
@@ -64,8 +64,7 @@ describe("<Card>", () => {
 
     assert.lengthOf(wrapper.find(".card-preview-image"), 0);
   });
-  it("should have a link menu", () => assert.ok(wrapper.find(LinkMenu)));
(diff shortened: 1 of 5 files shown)


k88hudson:
Good catch! Thanks for fixing this 👍

as-pine-proxy:
[Treeherder: 723305d8a7dfc4dd194e7948089aa8684d4793dd](https://treeherder.mozilla.org/#/jobs?repo=pine&revision=723305d8a7dfc4dd194e7948089aa8684d4793dd)

PR #4128, opened 2018-04-30
fix(prerender): Check top before loading more scripts

Mardak:
r?@k88hudson Just short circuit the rest of the top level script

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -200,7 +200,13 @@ function templateHTML(options, html) {
     <div id="snippets-container">
       <div id="snippets"></div>
     </div>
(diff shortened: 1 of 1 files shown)


bin/render-activity-stream-html.js
```diff
@@ -200,7 +200,13 @@ function templateHTML(options, html) {
     <div id="snippets-container">
       <div id="snippets"></div>
     </div>
-    <script>
+    <script>(() => {
+// Clear out any prerendered content if we shouldn't continue loading
```
k88hudson: When does this actually occur?

k88hudson:
Ok, this works as expected 👍

PR #4130, opened 2018-05-01
[WIP] Fix Bug 1449062 - Avoid requesting thumbnails if a rich icon might be found

rlr:
@Mardak thoughts on this approach?

One problem with it is that it doesn't wait for the redirect query to run. I guess I could dispatch from `fetchIconFromRedirects()` if it fails to find a rich icon?

Also, I was struggling to port some of the tests so I figured I'd see if the approach is OK before pulling more hair out with those.


Mardak:
The link object being passed around is from the cache, and it has functions of it via __sharedCache. I guess it might happen to work in Main as the same object is used?? but if it needed to go to content, it would be serialized? Also, TopSiesFeed is deleting stuff from the link before dispatching, so not sure how that interacts...

rlr:
Let's get on vidyo sometime today. I'm not fully understanding. I *think* what I've tried to implement has the same end effect of what you proposed below?

from IRC:

> r1cky: hmm.. so the competing aspects is that faviconfeed is normally lazy, so we don't want to always wait for it to initialize
> 1:19 PM i wonder if we can keep the lazy aspect (not refreshing sites by domain until RICH_ICON_MISSING) and dispatch an action to update redux state when faviconfeed is actually initialized
> 1:20 PM topsitesfeed can still dispatch RICH_ICON_MISSING and skips screenshots while favicons are not initialized?
> 1:20 PM although just because favicons are initialized doesn't mean we should get screenshots quite yet...
> 1:20 PM also, we probably don't want to store the whole list of rich icon domains in redux as it gets copied to each content page

system-addon/lib/TopSitesFeed.jsm
```diff
@@ -432,6 +429,9 @@ this.TopSitesFeed = class TopSitesFeed {
       case at.PREVIEW_REQUEST:
         this.getScreenshotPreview(action.data.url, action.meta.fromTarget);
         break;
+      case at.SCREENSHOT_NEEDED:
+        this._fetchScreenshot(action.data.link, action.data.url);
```
rlr: > The link object being passed around is from the cache, and it has functions of it via __sharedCache. I guess it might happen to work in Main as the same object is used?? 

Oh, hmm. I see. This isn't working because the `__sharedCache` seems to be getting dropped somewhere. `TypeError: cache is undefined[Learn More] Screenshots.jsm:125:1 `

rlr:
closing for now. probably needs a new approach

PR #4135, opened 2018-05-02
Fix Bug 1436228 - Green up our JS dependencies for Firefox 61

Mardak:
r?@k88hudson Most upgrades are dev dependencies except prop-types, which didn't actually change any code as it was fixing up its license header for some files. I kept the changes to the single package.json with a final package-lock update if someone wants to bisect later.

These are not updated:
```
 react                         16.2.0  →  16.3.2 
 react-dom                     16.2.0  →  16.3.2 
 react-redux                    5.0.6  →   5.0.7 
 redux                          3.6.0  →   4.0.0 
 istanbul-instrumenter-loader   0.2.0  →   3.0.1 
 karma-sinon                    1.0.5  →   1.8.0 
 loader-utils                  0.2.16  →   1.1.0 
 node-sass                      4.7.2  →   4.9.0 
 sinon                          4.5.0  →   5.0.3 
 webpack                       3.10.0  →   4.6.0 
```

react/react-dom is split out for early in 62 cycle update.
redux/react-redux similarly for early 62.
istanbul-instrumenter-loader/karma-sinon/loader-utils have been skipped in the past
node-sass starting from the next available version 4.8.1 seems to cause dark theme $vars to not expand, so splitting out
sinon removed .reset, so that needs to be cleaned up
webpack seems to result in minified-ish/newline removed output, so maybe there's more configs we want?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "cpx": "1.5.0",
     "enzyme": "3.3.0",
     "enzyme-adapter-react-16": "1.1.1",
-    "eslint": "4.17.0",
+    "eslint": "4.19.1",
     "eslint-plugin-import": "2.8.0",
     "eslint-plugin-json": "1.2.0",
     "eslint-plugin-mozilla": "0.8.1",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "karma-mocha-reporter": "2.2.5",
     "karma-sinon": "1.0.5",
     "karma-sourcemap-loader": "0.3.7",
-    "karma-webpack": "2.0.9",
+    "karma-webpack": "3.0.0",
     "loader-utils": "0.2.16",
     "minimist": "1.2.0",
     "mocha": "5.0.0",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "eslint-plugin-no-unsanitized": "3.0.0",
     "eslint-plugin-promise": "3.7.0",
     "eslint-plugin-react": "7.7.0",
-    "eslint-watch": "3.1.3",
+    "eslint-watch": "3.1.4",
     "husky": "0.14.3",
     "istanbul-instrumenter-loader": "0.2.0",
     "joi-browser": "13.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,60 +7,60 @@
...
+    "eslint-plugin-mozilla": "0.11.0",
+    "eslint-plugin-no-unsanitized": "3.0.0",
+    "eslint-plugin-promise": "3.7.0",
+    "eslint-plugin-react": "7.7.0",
+    "eslint-watch": "3.1.4",
     "husky": "0.14.3",
     "istanbul-instrumenter-loader": "0.2.0",
     "joi-browser": "13.0.1",
-    "karma": "1.7.1",
+    "karma": "2.0.2",
```
k88hudson: The last time we were greening dependencies karma  2.0.0 had problems with source maps (see https://bugzilla.mozilla.org/show_bug.cgi?id=1441278) and we needed to downgrade it. Let me confirm this is still the case, it may be fixed now

Mardak: Yup, I confirmed it yesterday but good to make sure it wasn't just some local configuration of mine: https://bugzilla.mozilla.org/show_bug.cgi?id=1441278#c9

k88hudson: Excellent! it is indeed working 👍

k88hudson:
Ok, I tested all the yamscripts and some debugging etc., everything looks good. Do we need to file some follow-up tickets to handle some of the tasks you described (such as removing `.reset` from sinon for example)?

Mardak:
Filed the followup bugs:
       1459444: Upgrade dependency from react/react-dom 16.2.0
       1459445: Upgrade dependency from redux 3.6.0 / react-redux 5.0.6
       1459446: Upgrade dependency from node-sass 4.7.2
       1459447: Upgrade dependency from sinon 4.5.0
       1459449: Upgrade dependency from webpack 3.10.0

https://bugzilla.mozilla.org/showdependencytree.cgi?id=1436228&maxdepth=1&hide_resolved=0

PR #4136, opened 2018-05-02
Fix Bug 1458640 - Prevent any fatal indexedDB errors from crashing section

k88hudson:


system-addon/lib/ActivityStreamStorage.jsm
```diff
@@ -65,6 +65,27 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
...
+   *                  If it does not exist, create it.
+   *                  If an error occurs, deleted the db and attempt to
+   *                  re-create it.
+   * @returns Promise that resolves with a db instance
+   */
+  async createOrOpenDb() {
+    let db;
+    try {
+      db = await this._openDatabase();
+    } catch (e) {
```
Mardak: I was able to trigger a `UnknownError: The operation failed for reasons unrelated to the database itself and not covered by any other error code.` with a `sqlite3 ./storage/permanent/chrome/idb/1657114595AmcateirvtiSty.sqlite "UPDATE object_data SET data = RANDOM(), key = RANDOM(); UPDATE database SET name = RANDOM();"`

So I guess this is working? ;)

system-addon/lib/ActivityStreamStorage.jsm
```diff
@@ -65,6 +65,27 @@ this.ActivityStreamStorage = class ActivityStreamStorage {
...
+   * createOrOpenDb - Open a db (with this.dbName) if it exists.
+   *                  If it does not exist, create it.
+   *                  If an error occurs, deleted the db and attempt to
+   *                  re-create it.
+   * @returns Promise that resolves with a db instance
+   */
+  async createOrOpenDb() {
+    let db;
+    try {
+      db = await this._openDatabase();
```
Mardak: nit: this can just `return await this._openDatabase();` and similar below without needing `let db;`

PR #4143, opened 2018-05-07
Fix Bug 1432676 - Create AS Router UI surface / template for Onboarding overlay

sarracini:
This PR does the following:
1. Creates an `OnboardingMessages.jsx` component which houses the new onboarding tour
2. Creates a `ModalOverlay.jsx` component as a generic modal that can be used for other types of messages (but is also used for onboarding right now)
3. Creates a notion of "bundled messages". Onboarding needs 3 messages, all from the same provider, of the same template. So `ASRouter.jsm` can now detect if the message needs to be bundled with others of it's type, with how many, then goes through and bundles these up. It sends it down to the `asrouter-content.jsx` via a new message "SET_BUNDLED_MESSAGES" which saves a "bundle" in the state. 
4. Conditionally renders onboarding if there is a set of bundled messages, or the regular snippets template as a default.
4. Adds user telemetry to each "Try Now" action for onboarding messages
5. Adds a generic `executeAction` function which executes an action in main process from a list of new `ASRouterActions` found in `Actions.jsm`.
6. Changes the render root to be a more generic `asrouter-container` element (previously `snippets-container`)
7. Adds a new UI Surface type "NEWTAB_OVERLAY" for telemetry 

As a follow up I'm going to factor out the onboarding messages that live in `ASRouter.jsm` into their own file, but I'm thinking I'll do that as part of [bug 1433217](https://bugzilla.mozilla.org/show_bug.cgi?id=1433217)

Specs: https://mozilla.invisionapp.com/share/V4GS3HWWP9H#/screens


sarracini:
So either I don't get 100% coverage and these lines aren't covered (but all tests pass): 
`ASRouter.jsm  lines 320,327 `
Or the following test doesn't pass (but coverage is 100%):
`it("should send a message back to the to the target", async () => {`

system-addon/common/Actions.jsm
```diff
@@ -110,6 +110,15 @@ for (const type of [
   actionTypes[type] = type;
 }
 
+const ASRouterActions = {};
+for (const type of [
```
k88hudson: If this is intended to be a whitelist of actions available to external message providers, I think a comment describing at such here would be good

system-addon/content-src/asrouter/asrouter-content.jsx
```diff
@@ -47,15 +53,15 @@ export class ASRouterUISurface extends React.PureComponent {
...
     this.sendImpression = this.sendImpression.bind(this);
     this.sendUserActionTelemetry = this.sendUserActionTelemetry.bind(this);
-    this.state = {message: {}};
+    this.state = {message: {}, bundle: {}};
   }
 
-  sendUserActionTelemetry(extraProps = {}) {
-    const {message} = this.state;
-    const eventType =  `${message.provider}_user_event`;
+  sendUserActionTelemetry(extraProps = {}, bundledMessageID) {
```
k88hudson: I think it might be better to remove the`bundleMessageId` param and just add`message_id` to `extraProps` when you are calling the function in onboarding messages, i.e. 

OnboardingMessage.jsx:
```js
this.props.sendUserActionTelemetry({message_id: this.props.id});
```

Additionally, you could do something like this here:
```js
if (!message && !extraProps.message_id) {
  throw new Error(`You must provider a message_id for bundled messages...`);
}

File: system-addon/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx [hunk 3]
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+
+export class ModalOverlay extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.onClick = this.onClick.bind(this);
+  }
+
+  componentWillMount() {
+    this.setState({active: true});
```
k88hudson: Why not just set `this.state = {active: true}` in the constructor?

system-addon/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  componentWillReceiveProps() {
+    this.setState({active: true});
```
k88hudson: What's your reasoning behind this?

system-addon/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  onClick(e) {
```
k88hudson: can you make this a little bit more descriptive like, `onDoneButtonClick` or something like that

system-addon/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -0,0 +1,42 @@
...
+    this.setState({active: false});
+    document.body.classList.remove("modal-open");
```
k88hudson: You probably want to remove this on `componentWillUnmount` as well

system-addon/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  render() {
+    const {active} = this.state;
+    if (active) {
+      // This lets the body know that the modal is open so that we can hide the scroll bar
+      document.body.classList.add("modal-open");
```
k88hudson: I think what you want here is a `componentDidUpdate` function, in which you would check the active state and set `document.body.classList` to the appropriate thing

system-addon/content-src/asrouter/components/ModalOverlay/_ModalOverlay.scss
```diff
@@ -0,0 +1,90 @@
+.activity-stream {
+  &.modal-open {
+    overflow: hidden;
+  }
+}
+.modalOverlayOuter {
+  background: $white;
```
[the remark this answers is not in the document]
sarracini: let's do it as a follow up, since I don't have the design specs yet for dark theme

k88hudson:
The super small size doesn't look too great:
image

k88hudson:
I think the height of the middle section should probably be allowed to be variable since this text will be localized:

image

system-addon/content-src/asrouter/components/ModalOverlay/_ModalOverlay.scss
```diff
@@ -0,0 +1,90 @@
...
+  // if modal is short enough, add a vertical scroll bar
+  @media(max-width: 850px) and (max-height: 730px) {
+    overflow-y: scroll;
+  }
+
+  &.active {
+    display: block;
+  }
+
+  h2 {
```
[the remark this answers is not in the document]
sarracini: I've got line-height 24px and font-size 18px in the specs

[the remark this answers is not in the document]
sarracini: nvm!

k88hudson:
Ok, this looks good to go with the following fixes:
- [ ] rename div back to `snippets-container`
- [ ] fix that failing test to run with both a regular message and a bundled message via `.setState({messages: })`
- [ ] file a "design polish" follow-up ticket that includes flexible heights for localized text/really narrow widths

sarracini:
Follow up bug filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1462378

Mardak:
I fixed a `border-radius: none` in https://github.com/mozilla/activity-stream/commit/edc5578ede00cf2d8c820f9182aa645593722f0a to get browser_parsable_css passing. But it looks like there's more warnings from sasslint including:
```
system-addon/content-src/asrouter/templates/OnboardingMessage/_OnboardingMessage.scss
   33:15  warning  A value of `none` is not allowed. `0` must be used                                           border-zero
   80:13  warning  A value of `none` is not allowed. `0` must be used                                           border-zero
```

PR #4144, opened 2018-05-07
Bug 1401194 - Add data-section-id to sections for web extension tests

k88hudson:
This is to support tests that will be added in Bug 1401194 in order for web extension tests to identify the correct section more easily (see https://reviewboard.mozilla.org/r/213240/#comment299514)

### DIFF
--- a/system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
+++ b/system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
@@ -152,9 +152,11 @@ export class _CollapsibleSection extends React.PureComponent {
     const disclaimerPref = `section.${id}.showDisclaimer`;
     const needsDisclaimer = disclaimer && this.props.Prefs.values[disclaimerPref];
     const active = menuButtonHover || showContextMenu;
-
     return (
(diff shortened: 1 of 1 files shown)


system-addon/content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -152,9 +152,11 @@ export class _CollapsibleSection extends React.PureComponent {
     const disclaimerPref = `section.${id}.showDisclaimer`;
     const needsDisclaimer = disclaimer && this.props.Prefs.values[disclaimerPref];
     const active = menuButtonHover || showContextMenu;
-
     return (
-      <section className={`collapsible-section ${this.props.className}${enableAnimation ? " animation-enabled" : ""}${collapsed ? " collapsed" : ""}${active ? " active" : ""}`}>
+      <section
+        className={`collapsible-section ${this.props.className}${enableAnimation ? " animation-enabled" : ""}${collapsed ? " collapsed" : ""}${active ? " active" : ""}`}
+        // Note: data-section-id is used for web extension api tests in mozilla central
+        data-section-id={id}>
```
Mardak: Pretty sure data attribute values can have any string value, so we don't need to additionally clean up / sanitize… yeah?

k88hudson: yes, rendering the id as data attribute should always be a string, so it shouldn't be a problem here

PR #4145, opened 2018-05-09
Fix Bug 1460119 - gSnippetsMap.getTotalBookmarksCount() never resolves

rlr:
r? @k88hudson

### DIFF
--- a/system-addon/content-src/lib/snippets.js
+++ b/system-addon/content-src/lib/snippets.js
@@ -73,13 +73,13 @@ export class SnippetsMap extends Map {
 
   getTotalBookmarksCount() {
     return new Promise(resolve => {
-      this._dispatch(ac.OnlyToMain({type: at.TOTAL_BOOKMARKS_REQUEST}));
       global.addMessageListener("ActivityStream:MainToContent", function onMessage({data: action}) {
         if (action.type === at.TOTAL_BOOKMARKS_RESPONSE) {
           resolve(action.data);
           global.removeMessageListener("ActivityStream:MainToContent", onMessage);
         }
       });
+      this._dispatch(ac.OnlyToMain({type: at.TOTAL_BOOKMARKS_REQUEST}));
     });
   }
 
--- a/system-addon/content-src/lib/snippets.js
+++ b/system-addon/content-src/lib/snippets.js
@@ -73,13 +73,13 @@ export class SnippetsMap extends Map {
 
   getTotalBookmarksCount() {
     return new Promise(resolve => {
+      this._dispatch(ac.OnlyToMain({type: at.TOTAL_BOOKMARKS_REQUEST}));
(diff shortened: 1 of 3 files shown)


pmac:
If we're going to have to change this to get it to work, is it possible to just make the value available in gSnippetsMap like other values (e.g. `appData.isDevtoolsUser`)? I could be wrong, but I don't believe this number needs to be accurate up to the minute and can just refresh on session start or end.

@glogiotatidis please correct me if I'm wrong though.

glogiotatidis:
> If we're going to have to change this to get it to work, is it possible to just make the value available in gSnippetsMap like other values (e.g. appData.isDevtoolsUser)? I could be wrong, but I don't believe this number needs to be accurate up to the minute and can just refresh on session start or end.

Yes to having this as the other values. Refreshing async on every newtab load or every X minutes/days (as we want to do with addon list) would be ideal just to be able to get stats for long lived sessions.

k88hudson:
@pmac @glogiotatidis Personally I'd prefer to keep this as asynchronous function to prevent having to run this query at startup every time if it's not needed

system-addon/content-src/lib/snippets.js
```diff
@@ -73,13 +73,13 @@ export class SnippetsMap extends Map {
...
   getTotalBookmarksCount() {
     return new Promise(resolve => {
-      this._dispatch(ac.OnlyToMain({type: at.TOTAL_BOOKMARKS_REQUEST}));
       global.addMessageListener("ActivityStream:MainToContent", function onMessage({data: action}) {
         if (action.type === at.TOTAL_BOOKMARKS_RESPONSE) {
           resolve(action.data);
           global.removeMessageListener("ActivityStream:MainToContent", onMessage);
         }
       });
+      this._dispatch(ac.OnlyToMain({type: at.TOTAL_BOOKMARKS_REQUEST}));
```
k88hudson: This doesn't actually need to move

system-addon/lib/SnippetsFeed.jsm
```diff
@@ -212,7 +212,7 @@ this.SnippetsFeed = class SnippetsFeed {
         this._clearBlockList();
         break;
       case at.TOTAL_BOOKMARKS_REQUEST:
-        this.getTotalBookmarksCount(action._target.browser);
+        this.getTotalBookmarksCount(action.meta.fromTarget);
```
k88hudson: hm, I wonder if we can figure out a test to prevent this error in the future...

system-addon/content-src/lib/snippets.js
rlr: oh oops. i moved it when troubleshooting and forgot to move it back

system-addon/lib/SnippetsFeed.jsm
rlr: mochitests? 😖

PR #4148, opened 2018-05-11
Fix Bug 1457580 - Update strings-import to get strings from l10n-central

Mardak:
Waiting for Pike's feedback before review.

### DIFF
--- a/bin/strings-import.js
+++ b/bin/strings-import.js
@@ -2,74 +2,74 @@
 "use strict";
 
 /* eslint-disable no-console */
+const fetch = require("node-fetch");
 
-const DEFAULT_LOCALE = "en-US";
-
-/* globals cd, cp, ls, mkdir, rm */
+/* globals cd, ls, mkdir, rm, ShellString */
 require("shelljs/global");
 
-const args = process.argv.slice(2);
-if (args.length < 1) {
-  throw Error("Please provide the path to the target strings repository");
+const DEFAULT_LOCALE = "en-US";
+const L10N_CENTRAL = "https://hg.mozilla.org/l10n-central";
+const PROPERTIES_PATH = "raw-file/default/browser/chrome/browser/activity-stream/newtab.properties";
+const STRINGS_FILE = "strings.properties";
+
+// Get all the locales in l10n-central
+async function getLocales() {
+  console.log(`Getting locales from ${L10N_CENTRAL}`);
+
+  // Add all non-test sub repository locales
+  const locales = [];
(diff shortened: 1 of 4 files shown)


bin/strings-import.js
```diff
@@ -2,74 +2,78 @@
...
 
 /* eslint-disable no-console */
+const fetch = require("node-fetch");
+
+/* globals cd, ls, mkdir, rm, ShellString */
+require("shelljs/global");
 
 const DEFAULT_LOCALE = "en-US";
+const L10N_CENTRAL = "https://hg.mozilla.org/l10n-central";
+const PROPERTIES_PATH = "raw-file/tip/browser/extensions/activity-stream/activity-stream.properties";
```
flodolo: I think you want `default` here, instead of `tip`.

bin/strings-import.js
```diff
@@ -2,74 +2,78 @@
...
-/* globals cd, cp, ls, mkdir, rm */
-require("shelljs/global");
+// Get all the locales in l10n-central
+async function getLocales() {
+  console.log(`Getting locales from ${L10N_CENTRAL}`);
 
-const args = process.argv.slice(2);
-if (args.length < 1) {
-  throw Error("Please provide the path to the target strings repository");
+  const html = await (await fetch(L10N_CENTRAL)).text();
```
flodolo: You can use `https://hg.mozilla.org/l10n-central/?style=raw`

Pike: ... or json, depending on your preference.

flodolo: I didn't know about https://hg.mozilla.org/l10n-central/?style=json

That makes things a lot less hacky

bin/strings-import.js
```diff
@@ -2,74 +2,74 @@
...
 
-// Use the first script argument as the source of localized strings
-const [stringsPath] = args;
-const stringsRepo = require("simple-git")(stringsPath);
+// Save the properties file to the locale's directory
+async function saveProperties(locale) {
+  // Only save a file if the repository has the file
+  const url = `${L10N_CENTRAL}/${locale}/${PROPERTIES_PATH}`;
+  const response = await fetch(url);
+  if (response.status !== 200) {
```
k88hudson: `if (!response.ok)` might be a little less restrictive

Mardak: ok? ok!? We're not ok! Must be 200!!! ok… ok.

Mardak: Updated ;)

PR #4149, opened 2018-05-11
Fix Bug 1457582 - Replace strings-export and include .properties as part of buildmc

Mardak:
Waiting for Pike's feedback before review.


Mardak:
`cpx` copies multiple files/directories to a destination without renaming the copied file. That's why I needed to use `cp`

Mardak:
r?@k88hudson Updated PR to use the new destination:

browser/locales/en-US/chrome/browser/activity-stream/newtab.properties

yamscripts.yml
```diff
@@ -24,6 +24,7 @@ scripts:
     # Copy over all of the system-addon directory
     copy: cpx "system-addon/**/{,.}*" $npm_package_config_mc_dir/browser/extensions/activity-stream
     version: node ./bin/update-version.js $npm_package_config_mc_dir/browser/extensions/activity-stream
+    stringsExport: cp locales/en-US/strings.properties $npm_package_config_mc_dir/browser/locales/en-US/chrome/browser/activity-stream/newtab.properties
```
k88hudson: `cp` will fail on windows but we're probably using other unixy stuff elsewhere so oh well :P

k88hudson: also, do you want to check if this directory already exists? this would fail otherwise

Mardak: Yeah for directory existing.. but... except for the first export, the directory and file will already exist. ;) So I decided against an extra mkdir in the common case going forwards. 😛

PR #4151, opened 2018-05-15
Bug 1461787 - Add endpoint pref, button_type to ASRouter experiment

k88hudson:
These are some final changes for the 61 experiment, including making the as endpoint changeable via a pref, as well as adding "button_type" option for messages that allows them to be anchor-style buttons.

You can test this by:
- setting `browser.newtabpage.activity-stream.asrouter.snippetsUrl` to `https://gist.github.com/k88hudson/2d627d17ebd3e622e5b05e060d98b182/raw` and **restarting** the browser
- make sure the correct url shows up on about:newtab#asrouter
- make sure both types of snippets show up correctly

The two types of buttons look like this:

anchor
image

default/button
image

### DIFF
--- a/system-addon/lib/ASRouter.jsm
+++ b/system-addon/lib/ASRouter.jsm
@@ -4,20 +4,11 @@ Cu.importGlobalProperties(["fetch"]);
 const INCOMING_MESSAGE_NAME = "ASRouter:child-to-parent";
 const OUTGOING_MESSAGE_NAME = "ASRouter:parent-to-child";
 const ONE_HOUR_IN_MS = 60 * 60 * 1000;
-// This is a temporary endpoint until we have something for snippets
-const SNIPPETS_ENDPOINT = "https://activity-stream-icons.services.mozilla.com/v1/messages.json.br";
-
-const LOCAL_TEST_MESSAGES = [
-  {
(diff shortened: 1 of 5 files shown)


system-addon/lib/ASRouter.jsm
```diff
@@ -4,20 +4,10 @@ Cu.importGlobalProperties(["fetch"]);
...
-    template: "simple_snippet",
-    content: {
-      text: "Your browser is ready for a makeover. Don't worry, you've got tons of options.",
-      button_label: "Check them out here",
-      button_url: "https://addons.mozilla.org/en-US/firefox/themes"
-    }
-  }
-];
+const SNIPPETS_ENDPOINT_PREF = "browser.newtabpage.activity-stream.asrouter.snippetsUrl";
+const SNIPPETS_ENDPOINT = Services.prefs.getStringPref(SNIPPETS_ENDPOINT_PREF,
```
piatra: I think that without a pref change listener this would require browser restart in order to register the changes

k88hudson:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1462114 as a follow-up to add a listener for endpoint prefs.

PR #4152, opened 2018-05-15
Fix Bug 1460782 - Exception in TelemetryFeed uninit

piatra:
The problem described in [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1460782#c4) seems very likely. The feed is instantiated but the `initAction` is dispatched later because it waits for indexeddb. We could move the `addObserver` call in the constructor but I'm not sure if it has any other consequences cc @dmose ?
Guarding against exceptions than could be thrown by removeObserver seems like a common pattern.

### DIFF
--- a/system-addon/bootstrap.js
+++ b/system-addon/bootstrap.js
@@ -35,10 +35,9 @@ XPCOMUtils.defineLazyModuleGetter(this, "ActivityStream",
         (await (await fetch(uri)).text())
           .split("\n").slice(2).forEach(line => cb(line.split(" ").slice(1)));
       } catch (e) {
-        // Silently ignore any modules that fail to load.
+        // Silently ignore listings that fail to load
+        // probably because the resource: has been unloaded
       }
-
-      return null;
     };
 
     // Look for modules one level deeper than the top resource URI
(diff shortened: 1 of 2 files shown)


system-addon/bootstrap.js
```diff
@@ -30,8 +30,16 @@ let waitingForBrowserReady = true;
...
-      .split("\n").slice(2).forEach(line => cb(line.split(" ").slice(1)));
+    const processListing = async (uri, cb) => {
+      try {
+        (await (await fetch(uri)).text())
+          .split("\n").slice(2).forEach(line => cb(line.split(" ").slice(1)));
+      } catch (e) {
+        // Silently ignore any modules that fail to load.
+      }
+
+      return null;
```
Mardak: Is this `return` necessary?

system-addon/bootstrap.js
```diff
@@ -30,8 +30,16 @@ let waitingForBrowserReady = true;
...
   "resource://activity-stream/lib/ActivityStream.jsm", null, null, () => {
     // Helper to fetch a resource directory listing and call back with each item
-    const processListing = async (uri, cb) => (await (await fetch(uri)).text())
-      .split("\n").slice(2).forEach(line => cb(line.split(" ").slice(1)));
+    const processListing = async (uri, cb) => {
+      try {
+        (await (await fetch(uri)).text())
+          .split("\n").slice(2).forEach(line => cb(line.split(" ").slice(1)));
+      } catch (e) {
+        // Silently ignore any modules that fail to load.
```
Mardak: The modules aren't being loaded here, so this comment is confusing. Maybe just "Silently ignore listings that fail to load -- probably because the resource: has been unloaded"

system-addon/bootstrap.js
Mardak: Was there something that caused this to be added in the first place? linting?

PR #4158, opened 2018-05-18
Bug 1456954 - Support object rest spread in Activity Stream

k88hudson:
This bug allows us to use object rest spread syntax via a babel transform, which essentially converts it to `Object.assign`.

Unfortunately some nested dependency stuff (I think eslint is the problem at the moment) is blocking us using it natively, but I'll file a follow-up to check this in a few weeks to see if the issue has been resolved.

Does this seem reasonable @Mardak @piatra?


system-addon/common/Actions.jsm
```diff
@@ -127,7 +127,7 @@ for (const type of [
 // Helper function for creating routed actions between content and main
 // Not intended to be used by consumers
 function _RouteMessage(action, options) {
-  const meta = action.meta ? Object.assign({}, action.meta) : {};
+  const meta = action.meta ? {...action.meta} : {};
```
piatra: :clap:

package.json
```diff
@@ -22,6 +22,7 @@
     "babel-plugin-jsm-to-esmodules": "0.4.0",
     "babel-plugin-transform-async-to-module-method": "6.24.1",
     "babel-plugin-transform-es2015-modules-commonjs": "6.26.2",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
```
piatra: Do we want to lock this to a specific version?

piatra:
I think the plugin needs to be added to the webpack prerender file too.

k88hudson:
The prerender webpack config picks up all the transforms from the regular webpack config, so there is no need to edit them separately

PR #4159, opened 2018-05-21
Fix Bug 1459289 - compact cards for highlights

rlr:
Note the following changes that we (@bryanbell and myself) decided as we iterated on this. I'll add a comment to the bug as well so QA is aware.

* changed highlights to two rows instead of 3
* fill all the possible/max rows with placeholders instead of just the current row
* on narrow screens, we hide the last two highlights (like we do for top sites)
* no longer showing file type for downloads
* if a link has no image (all downloads), we now show a grey box where the image goes instead of pushing the text to the top

Before:
screen shot 2018-05-21 at 5 08 03 pm
screen shot 2018-05-21 at 5 08 22 pm

After:
screen shot 2018-05-17 at 7 29 19 pm
screen shot 2018-05-17 at 7 31 21 pm


rlr:
I guess @sarracini is on PTO this week. @Mardak ?

system-addon/content-src/components/Card/_Card.scss
```diff
@@ -190,56 +168,144 @@
...
     }
 
-    .card-download-icon {
-      &.icon-download-folder {
-        height: $large-download-folder-icon-size;
-        width: $large-download-folder-icon-size;
+    .card-text {
+      .card-title,
+      &:not(.no-description) .card-title {
+        font-size: 10px;
```
Mardak: This looks particularly small, but I suppose if that's what design wants… Even the site name for top sites under the images are 11px and those are usually a single word to glance at instead of several words to try to read. Maybe because next to Pocket titles, which are 17px, these titles are more than 40% smaller.

system-addon/content-src/components/Card/_Card.scss
```diff
@@ -60,6 +65,7 @@
   }
 
   .card-preview-image-outer {
+    background-color: $grey-30;
```
Mardak: Some reason I thought we had some gradient placeholder.. but I can't seem to find it even quickly looking through old versions.

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -108,42 +116,61 @@ export class Section extends React.PureComponent {
...
     // Otherwise, we should show placeholders.
     const shouldShowEmptyState = initialized && !rows.length;
 
+    const cards = [];
+    if (!shouldShowEmptyState) {
+      for (let i = 0; i < maxCards; i++) {
+        const link = realRows[i];
+        // On narrow viewports, we only show 3 cards per row. We'll mark the rest as
+        // .hide-for-narrow to hide in CSS via @media query.
+        const className = (i >= maxCardsOnNarrow) ? "hide-for-narrow" : "";
```
Mardak: I see this is basically what we do for TopSites, so I guess it's fine. Just thinking about this a bit more, it's a bit strange for the API to have Card just use whatever className given instead of some boolean flag.

system-addon/lib/DownloadsManager.jsm
```diff
@@ -52,14 +44,7 @@ this.DownloadsManager = class DownloadsManager {
...
     const downloadedItem = element.download;
-    let description;
-    if (element.fileType) {
-      // If we have a file type: '1.5 MB — PNG'
-      description = `${element.sizeStrings.stateLabel} \u2014 ${element.fileType}`;
-    } else {
-      // If we do not have a file type: '1.5 MB'
-      description = `${element.sizeStrings.stateLabel}`;
-    }
+    const description = `${element.sizeStrings.stateLabel}`;
```
Mardak: nit: just get rid of the backtick string and use the value? ;)

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -19,7 +20,14 @@ function getFormattedMessage(message) {
 export class Section extends React.PureComponent {
   _dispatchImpressionStats() {
     const {props} = this;
-    const maxCards = 3 * props.maxRows;
+    let cardsPerRow = CARDS_PER_ROW_DEFAULT;
+    if (props.compactCards && global.matchMedia(`(min-width: 1072px)`).matches) {
+      // If the section has compact cards and the viewport is wide enough, we show
+      // 4 columns instead of 3.
+      // $break-point-widest = 1072px (from _variables.scss)
+      cardsPerRow = CARDS_PER_ROW_COMPACT_WIDE;
```
Mardak: Something about this seems a bit confusing with some implied behavior. I'm not sure if something like `EXTRA_COMPACT_CARDS = 1` as maybe it's clearer that non-wide with compact would just use the default. I guess we'll leave this as is for now.

system-addon/lib/DownloadsManager.jsm
rlr: ehm. oops

system-addon/content-src/components/Card/_Card.scss
rlr: Yeah, it is pretty small 🤔 . This might be one of those that gets tweaked once it gets in front of more eyeballs 👀

system-addon/content-src/components/Sections/Sections.jsx
rlr: Make a `hideForNarrow` prop instead? I guess that would keep it more strict/focused/scoped/whateveryoucallit?

Mardak: I wonder if `optional` would be too vague ;)

system-addon/content-src/components/Sections/Sections.jsx
rlr: Yeah. it's icky. I need to play with CSS grids a little more. What happens if you specify a 3x3 grid but have 8 items? Do the last 2 get hidden automagically?

PR #4161, opened 2018-05-22
Bug 1457227 - convert currentId to lastMessageId

k88hudson:
This pull request fixes an issue where if a AS router message is blocked, other preloaded browsers with the message don't also get blocked.

### DIFF
--- a/system-addon/lib/ASRouter.jsm
+++ b/system-addon/lib/ASRouter.jsm
@@ -161,7 +161,7 @@ class _ASRouter {
     this._storage = null;
     this._resetInitialization();
     this._state = {
-      currentId: null,
+      lastMessageId: null,
       providers: [],
       blockList: [],
       messages: [],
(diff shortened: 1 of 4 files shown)


system-addon/lib/ASRouter.jsm
```diff
@@ -318,11 +318,13 @@ class _ASRouter {
     }
   }
 
-  async clearMessage(target, id) {
-    if (this.state.currentId === id) {
-      await this.setState({currentId: null});
-    }
-    target.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {type: "CLEAR_MESSAGE"});
+  async blockById(idOrIds) {
```
piatra: Could you add a test for this method? A lot of asserts from the old unit tests went away with the refactoring.

k88hudson: Same goes for testing the end state (e.g. `Router.state.blockList.includes("foo")`) rather than implementation details (e.g. `assert.calledOnce(Router._storage.set)`)

PR #4162, opened 2018-05-22
Fix Bug 1457233 - Allow AS router message content to include subset of HTML

piatra:
I've tried including fluent as external modules that resulted in some issues where the build step would complain about the syntax and fail. Alternatively using the [`compat version`](https://github.com/projectfluent/fluent.js/tree/master/fluent-react#compatibility) requires the `regenerator-runtime` and in order to build you would have to include this module before fluent, this would result in too many external modules.

Similar problems occur when bundling fluent with AS:
* syntax errors force to use `compat` which also implies a dependency on `regenerator-runtime`
  * maybe a different preprocessing step?
* [the spread syntax](https://github.com/projectfluent/fluent.js/blob/75f57aa43a95f7e2072c1066e44524af9a07e7a1/fluent-react/vendor/voidElementTags.js#L15) results in an error for me even if I [rebased on top of the PR which added the plugin support](https://github.com/mozilla/activity-stream/pull/4158)
* at build step it expects [document to be globally available](https://github.com/projectfluent/fluent.js/blob/75f57aa43a95f7e2072c1066e44524af9a07e7a1/fluent-react/src/markup.js#L3)

But it works if you get through all of that, although I haven't figured out how to render links yet.

The extra logic involved (LocalizationProvider) is only temporary until https://github.com/projectfluent/fluent.js/issues/202 is fixed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -392,14 +392,14 @@
       "dev": true
     },
     "asn1.js": {
-      "version": "4.9.2",
-      "resolved": "https://registry.npmjs.org/asn1.js/-/asn1.js-4.9.2.tgz",
-      "integrity": "sha512-b/OsSjvWEo8Pi8H0zsDd2P6Uqo2TK2pH8gNLSJtNLM2Db0v2QaAZ0pBQJXVjAn4gBuugeVDr7s63ZogpUIwWDg==",
+      "version": "4.10.1",
+      "resolved": "https://registry.npmjs.org/asn1.js/-/asn1.js-4.10.1.tgz",
+      "integrity": "sha512-p32cOF5q0Zqs9uBiONKYLm6BClCoBCM5O9JfeUSlnQLBTxYdTK+pW+nXflm8UkKd2UYlEbYz5qEi0JuZR9ckSw==",
       "dev": true,
       "requires": {
         "bn.js": "4.11.8",
         "inherits": "2.0.3",
-        "minimalistic-assert": "1.0.0"
+        "minimalistic-assert": "1.0.1"
       }
     },
     "assert": {
@@ -668,6 +668,12 @@
       "integrity": "sha1-ytnK0RkbWtY0vzCuCHI5HgZHvpU=",
       "dev": true
     },
+    "babel-plugin-syntax-async-generators": {
+      "version": "6.13.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-syntax-async-generators/-/babel-plugin-syntax-async-generators-6.13.0.tgz",
+      "integrity": "sha1-a8lj67FuzLrmuStZbrfzXDQqi5o=",
+      "dev": true
+    },
     "babel-plugin-syntax-flow": {
       "version": "6.18.0",
       "resolved": "https://registry.npmjs.org/babel-plugin-syntax-flow/-/babel-plugin-syntax-flow-6.18.0.tgz",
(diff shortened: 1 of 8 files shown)


system-addon/content-src/activity-stream-prerender.jsx
```diff
@@ -1,3 +1,4 @@
+import "regenerator-runtime/runtime";
```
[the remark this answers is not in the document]
k88hudson: After some more investigation async iterator support seemed to have landed in acorn 5.4.0 (see https://github.com/acornjs/acorn/releases/tag/5.4.0), so the same bug that is causing our object rest spread bug is also causing this :(

piatra:
I've only whitelisted `<b>` tags. If any other tag is found, it will be stripped, its text content will be used.

Right now anchors are rendered using `ASRouterAnchor` but snippets can contain links in arbitrary positions (not just at the end of the message). This requires some changes on the snippets side, anchors will be wrapped in `<cta>` tags and the URL will be provided separately in the payload like the endpoint demo we currently have.

grafik

system-addon/content-src/asrouter/asrouter-content.jsx
```diff
@@ -53,6 +55,45 @@ function shouldSendImpressionOnUpdate(nextProps, prevProps) {
...
+};
+
+/**
+ * Transform an object (tag name: {url}) into (tag name: anchor) where the url
+ * is used as href, in order to render links inside a Fluent.Localized component.
+ */
+function convertLinks(links) {
+  if (links) {
+    return Object.keys(links).reduce((acc, linkTag) => {
+      acc[linkTag] = <a href={links[linkTag].url} />;
```
k88hudson: Can you whitelist allowed protocols here, so at least `javascript:` isn't allowed?

webpack.prerender.config.js
```diff
@@ -24,6 +24,7 @@ module.exports = Object.assign({}, webpackConfig, {
     "react-dom": "commonjs react-dom"
   },
   plugins: [
-    new webpack.BannerPlugin(banner)
+    new webpack.BannerPlugin(banner),
+    new webpack.DefinePlugin({document: {createElement: () => {}}})
```
k88hudson: looool hax :P Can you add a comment about why this is needed?

system-addon/content-src/asrouter/asrouter-content.jsx
```diff
@@ -129,12 +170,15 @@ export class ASRouterUISurface extends React.PureComponent {
...
-          <SimpleSnippet
-            {...this.state.message}
-            UISurface="NEWTAB_FOOTER_BAR"
-            getNextMessage={ASRouterUtils.getNextMessage}
-            onBlock={this.onBlockById(this.state.message.id)}
-            sendUserActionTelemetry={this.sendUserActionTelemetry} />
+          <LocalizationProvider messages={generateMessages(this.state.message.content.text)}>
+            <SimpleSnippet
+              {...this.state.message}
+              richText={<RichText text={this.state.message.content} />}
```
k88hudson: I would do `text={this.state.message.content.text} links={this.state.message.content.links}` or just `{...this.state.message.content}`, passing message.content as text is kind of confusing

package-lock.json
```diff
@@ -392,14 +392,14 @@
       "dev": true
     },
     "asn1.js": {
-      "version": "4.9.2",
```
k88hudson: this seems like  it might be a lot of changes for just adding a couple of dependencies, what version of npm are you using out of curiosity?

system-addon/content-src/asrouter/asrouter-content.jsx
```diff
@@ -53,6 +55,45 @@ function shouldSendImpressionOnUpdate(nextProps, prevProps) {
   return (nextProps.message.id && (!prevProps.message || prevProps.message.id !== nextProps.message.id));
 }
 
+function* generateMessages(content) {
```
k88hudson: Any particular reason this is a generator and not an async function?

system-addon/content-src/asrouter/asrouter-content.jsx
```diff
@@ -53,6 +55,45 @@ function shouldSendImpressionOnUpdate(nextProps, prevProps) {
...
+    return Object.keys(links).reduce((acc, linkTag) => {
+      acc[linkTag] = <a href={links[linkTag].url} />;
+      return acc;
+    }, {});
+  }
+
+  return null;
+}
+
+function RichText(props) {
```
k88hudson: Can you add a small comment about what this is for

system-addon/content-src/asrouter/asrouter-content.jsx
k88hudson: actually you can use `safeURI` here for that

piatra:
Documentation here https://github.com/mozilla/activity-stream/blob/master/system-addon/content-src/asrouter/schemas/message-format.md

piatra:
Not sure what to do about the package-lock.json diff.
```
npm --version
5.6.0
```

PR #4166, opened 2018-05-24
Bug 1428274 - AS Router should handle 204, 302, 500 codes approrpiately

k88hudson:
I added some tests here to make sure AS Router handles different error codes correctly.

Does the behaviour in these tests look reasonable to you @glogiotatidis?


glogiotatidis:
> Does the behaviour in these tests look reasonable to you @glogiotatidis?

:+1:

PR #4167, opened 2018-05-24
chore(highlights): Update mochitest to check for 2 rows of placeholders

Mardak:
r?@k88hudson Followup to #4159


Mardak:
r1cky> Mardak: the diff looks good (on my phone at least :)

PR #4160, opened 2018-05-22
Bug 1440421 - Simple implementation of mirroring for feedback

dmose:
This works together with the other patch in bug 1440421.  That patches push down browser/extensions/activity-stream into browser/extensions/activity-stream/system-addon in a way that still builds and functions.  This one updates the exporting code to export everything.

PR #4169, opened 2018-05-25
Mirror directories between mc / github

k88hudson:
This PR is intended to make our github repository closer to a mirror of the browser/extensions/activity-stream/ folder in MC. It's based off @dmose's work in https://github.com/mozilla/activity-stream/pull/4160 with a slightly different approach:

- Instead of moving files in mc to a `system-addon/` subdirectory, I chose to move files in our github repository. The reason for this is that all the code that needs to be changed in mc for file moves (such as eslint, license files) are more distributed whereas all the changes to get rid of `system-addon/` in github are local to our repo.
- I added a `bundle` task that could be run from the mozilla central directory directly, `buildmc` still works as it does now
- Syncing is done with `rsync` and a `.mcignore` file listing all files and directories to ignore
- I wasn't sure what to do with the startmc copy watching task, I kept `cpx` with a glob there but having that in two places is not great

The first commit here is a directory move commit, the second contains all path changes and new build tasks.

To test this, I would recommend:
- Run buildmc from the github directory and see if the result runs correctly from the `mozilla-central` directory;
- Make a change in browser/extensions/activity-stream/ and run `npm install && npm run bundle && ./mach build browser/extensions/activity-stream && ./mach run` to see if your change was compiled correctly
- Run` importmc` from github directory to see if the change was correctly imported


karma.mc.config.js
```diff
@@ -120,19 +118,20 @@ module.exports = function(config) {
             test: /\.jsx$/,
             exclude: /node_modules/,
             loader: "babel-loader",
-            options: {
-              presets: ["react"], // require("babel-preset-react")
-              plugins: [["transform-object-rest-spread", {"useBuiltIns": true}]]
-            }
```
dmose: Is that supposed to be part of _this_ PR?

dmose:
> Instead of moving files in mc to a system-addon/ subdirectory, I chose to move files in our github repository. The reason for this is that all the code that needs to be changed in mc for file moves (such as eslint, license files) are more distributed whereas all the changes to get rid of system-addon/ in github are local to our repo.

That's a reasonable approach too.

> I added a bundle task that could be run from the mozilla central directory directly, buildmc still works as it does now

Nice; I like it.

> Syncing is done with rsync and a .mcignore file listing all files and directories to ignore
    
+1

> I wasn't sure what to do with the startmc copy watching task, I kept cpx with a glob there but having that in two places is not great

We could consider something like https://www.npmjs.com/package/watch-and-rsync I suppose.

karma.mc.config.js
k88hudson: sorry, bad rebase – should be fixed now

.gitignore
```diff
@@ -8,10 +8,10 @@ logs/
...
 firefox/
 *.update.rdf
-system-addon/data/content/activity-stream.bundle.js
-system-addon/data/content/*.js.map
-system-addon/data/locales.json
-system-addon/css/
-system-addon/prerendered/
+data/content/activity-stream.bundle.js
+data/content/*.js.map
+data/locales.json
```
Mardak: I noticed `locales.json` being copied over, but I was initially confused why it didn't show up in mozilla-central's `git status`, and it's because m-c git recursively looks at sub `.gitignore`s. We happen to be okay with existing checked-in-to-m-c files, but at least for right now, just because we don't want to check something in to activity-stream repository doesn't mean we also want to ignore it in mozilla-central, e.g., `prerendered/` files.

It does look like the only ignored and untracked file is `locales.json`, so maybe we just make sure that's added to `.mcignore`?

yamscripts.yml
```diff
@@ -10,54 +10,46 @@ scripts:
...
-    pre: =>cleanmc
-    locales: pontoon-to-json --src locales --dest system-addon/data
-    webpack: webpack --config webpack.system-addon.config.js --display-optimization-bailout
-    css: node-sass --source-map true --source-map-contents system-addon/content-src/styles -o system-addon/css
-    html: rimraf system-addon/prerendered && webpack --config webpack.prerender.config.js && node ./bin/render-activity-stream-html.js
-    # Copy over all of the system-addon directory
-    copy: cpx "system-addon/**/{,.}*" $npm_package_config_mc_dir/browser/extensions/activity-stream
+    pre: rimraf $npm_package_config_mc_dir/browser/extensions/activity-stream/
+    bundle: => bundle
+    copy: rsync --exclude-from .mcignore -a . $npm_package_config_mc_dir/browser/extensions/activity-stream/
```
Mardak: If we want the m-c version of the repository to be accurate, we do indeed want to copy over dot files including `.eslintrc`; however:
```
$ ./mach lint browser/extensions/activity-stream/
An error occurred running eslint. Please check the following error messages:


Oops! Something went wrong! :(

ESLint: 4.19.1.
ESLint couldn't find the plugin "eslint-plugin-import". This can happen for a couple different reasons:

1. If ESLint is installed globally, then make sure eslint-plugin-import is also installed globally. A globally-installed ESLint cannot find a locally-installed plugin.

2. If ESLint is installed locally, then it's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm i eslint-plugin-import@latest --save-dev
```

I don't think we want to require m-c eslint to require our additional plugins yet. I suppose one simple fix is to have `m-c/.eslintignore` ignore all of `browser/extensions/activity-stream` ?? I believe @Standard8 has mentioned it's more acceptable for us as we run it ourselves. Although for anyone who tries to modify activity-stream directly from m-c won't get linting coverage…

yamscripts.yml
```diff
@@ -10,54 +10,46 @@ scripts:
...
+    copyPingCentre: cpx "ping-centre/PingCentre.jsm" $npm_package_config_mc_dir/browser/modules
 
 # startmc: Start developing the bootstrapped add-on
   startmc:
     _parallel: true
-    pre: =>prebuildmc
-    copy: =>buildmc:copy -- -w
+    pre: =>buildmc
+    # This copies only the system addon sub-folder; changing anything outside of it will need a full rebuild.
+    copy: cpx "{{,.}*,!(dist|logs|node_modules|ping-centre)/**/{,.}*}" $npm_package_config_mc_dir/browser/extensions/activity-stream/ -w
```
Mardak: If `startmc` is primarily to streamline development process, I think it's okay if we're a bit excessive in copying. So ignoring just `node_modules` will probably be good enough to keep this fast. Otherwise we might try too hard to keep this ignore list exact.

yamscripts.yml
Mardak: Alternatively, we could just .mcignore our .eslintrc.js for now… @Standard8 any suggestions?

.gitignore
k88hudson: I think probably the right thing to do here is not to copy over the `.gitignore`

yamscripts.yml
k88hudson: I tried ignoring `.eslintrc.js`, although for some reason i'm having trouble properly excluding files in `.eslintignore` (either a local one or the global one) when running ./mach lint

k88hudson: Ok, for some reason it wasn't working before, applying this patch seems to do the trick:
```diff
diff --git a/.eslintignore b/.eslintignore
index a7846d0e5630..c19e4a2d497a 100644
--- a/.eslintignore
+++ b/.eslintignore
@@ -93,6 +93,10 @@ browser/extensions/pocket/content/panels/js/vendor/**
 browser/extensions/activity-stream/data/content/activity-stream.bundle.js
 browser/extensions/activity-stream/test/**
 browser/extensions/activity-stream/vendor/**
+browser/extensions/activity-stream/*.js
+browser/extensions/activity-stream/bin/**
+browser/extensions/activity-stream/loaders/**
+browser/extensions/activity-stream/content-src/**
 # The only file in browser/locales/ is pre-processed.
 browser/locales/**
 # imported from chromium
```

k88hudson: Ok, so looks like we also needed to update python lint configuration

```diff
diff --git a/.flake8 b/.flake8
index 4df557d37236..ff91d4f4e460 100644
--- a/.flake8
+++ b/.flake8
@@ -3,6 +3,7 @@
 ignore = E121, E123, E126, E129, E133, E226, E241, E242, E704, W503, E402, E741
 max-line-length = 99
 exclude =
+    browser/extensions/activity-stream/node_modules/,
     browser/extensions/mortar/ppapi/,
     build/pymake/,
     testing/mochitest/pywebsocket,
```

and cpplint?
```diff
diff --git a/tools/lint/cpp-virtual-final.yml b/tools/lint/cpp-virtual-final.yml
index 319e971ac68b..f8f8ccb12779 100644
--- a/tools/lint/cpp-virtual-final.yml
+++ b/tools/lint/cpp-virtual-final.yml
@@ -10,6 +10,7 @@ cpp-virtual-final:
         - gfx/skia
         - media/webrtc/trunk
         - security/sandbox/chromium
+        - browser/extensions/activity-stream/node_modules
     extensions: ['cpp', 'h', 'mm']
     type: regex
     #
```

k88hudson: we still lint all of `lib/` and `common/` I guess :P but if we're ok with ignoring the whole thing, that's probably fine too

Mardak:
Here's the files that are being linted from m-c:
```
common/… 5 files
lib/… 32 files
prerendered/locales/… 285 activity-stream{-prerendered.html,-strings.js,.html} files
prerendered/static/activity-stream-debug.html
prerendered/static/activity-stream-initial-state.js
prerendered/static/activity-stream-prerendered-debug.html
```

I don't think we want the prerendered stuff linted.

I see there's already:
```
# Screenshots and Follow-on search are imported as a system add-on and have
# their own lint rules currently.
browser/extensions/followonsearch/**
browser/extensions/screenshots/**
```

So I think we're good to just ignore completely:
`browser/extensions/activity-stream/**`

Mardak:
For the other linters and node_modules.. I don't think we want to officially support people running `npm install` from `browser/extensions/activity-stream` quite yet, so I think we can skip that for now.

We'll want to do this change in its own export commit to m-c, so we can do a normal export, export just this and fix up .eslintignore at the same time.

Mardak:
Here's the list of changes for the export commit, and I've highlighted a few items:
```diff
	new file:   browser/extensions/activity-stream/.mcignore
	new file:   browser/extensions/activity-stream/.nvmrc
	new file:   browser/extensions/activity-stream/.sass-lint.yml
	new file:   browser/extensions/activity-stream/.travis.yml
	new file:   browser/extensions/activity-stream/CHANGELOG.md
	new file:   browser/extensions/activity-stream/CODEOWNERS
	new file:   browser/extensions/activity-stream/LICENSE
	modified:   browser/extensions/activity-stream/README.md
	new file:   browser/extensions/activity-stream/bin/download-firefox-artifact
	new file:   browser/extensions/activity-stream/bin/download-firefox-travis.sh
	new file:   browser/extensions/activity-stream/bin/prepare-mochitests-dev
	new file:   browser/extensions/activity-stream/bin/process-system-addon-for-package.js
	new file:   browser/extensions/activity-stream/bin/render-activity-stream-html.js
	new file:   browser/extensions/activity-stream/bin/strings-import.js
	new file:   browser/extensions/activity-stream/bin/test-merges.js
	new file:   browser/extensions/activity-stream/bin/update-version.js
	new file:   browser/extensions/activity-stream/content-src/activity-stream-prerender.jsx
	new file:   browser/extensions/activity-stream/content-src/activity-stream.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/asrouter-content.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/Button/Button.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/Button/_Button.scss
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/ImpressionsWrapper/ImpressionsWrapper.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/ModalOverlay/_ModalOverlay.scss
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/SnippetBase/SnippetBase.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/components/SnippetBase/_SnippetBase.scss
	new file:   browser/extensions/activity-stream/content-src/asrouter/schemas/message-format.md
	new file:   browser/extensions/activity-stream/content-src/asrouter/schemas/provider-response.schema.json
	new file:   browser/extensions/activity-stream/content-src/asrouter/template-utils.js
	new file:   browser/extensions/activity-stream/content-src/asrouter/templates/OnboardingMessage/OnboardingMessage.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/templates/OnboardingMessage/_OnboardingMessage.scss
	new file:   browser/extensions/activity-stream/content-src/asrouter/templates/SimpleSnippet/SimpleSnippet.jsx
	new file:   browser/extensions/activity-stream/content-src/asrouter/templates/SimpleSnippet/SimpleSnippet.schema.json
	new file:   browser/extensions/activity-stream/content-src/asrouter/templates/SimpleSnippet/_SimpleSnippet.scss
	new file:   browser/extensions/activity-stream/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ASRouterAdmin/ASRouterAdmin.scss
	new file:   browser/extensions/activity-stream/content-src/components/Base/Base.jsx
	new file:   browser/extensions/activity-stream/content-src/components/Base/_Base.scss
	new file:   browser/extensions/activity-stream/content-src/components/Card/Card.jsx
	new file:   browser/extensions/activity-stream/content-src/components/Card/_Card.scss
	new file:   browser/extensions/activity-stream/content-src/components/Card/types.js
	new file:   browser/extensions/activity-stream/content-src/components/CollapsibleSection/CollapsibleSection.jsx
	new file:   browser/extensions/activity-stream/content-src/components/CollapsibleSection/_CollapsibleSection.scss
	new file:   browser/extensions/activity-stream/content-src/components/ComponentPerfTimer/ComponentPerfTimer.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ConfirmDialog/ConfirmDialog.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ConfirmDialog/_ConfirmDialog.scss
	new file:   browser/extensions/activity-stream/content-src/components/ContextMenu/ContextMenu.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ContextMenu/_ContextMenu.scss
	new file:   browser/extensions/activity-stream/content-src/components/ErrorBoundary/ErrorBoundary.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ErrorBoundary/_ErrorBoundary.scss
	new file:   browser/extensions/activity-stream/content-src/components/LinkMenu/LinkMenu.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ManualMigration/ManualMigration.jsx
	new file:   browser/extensions/activity-stream/content-src/components/ManualMigration/_ManualMigration.scss
	new file:   browser/extensions/activity-stream/content-src/components/Search/Search.jsx
	new file:   browser/extensions/activity-stream/content-src/components/Search/_Search.scss
	new file:   browser/extensions/activity-stream/content-src/components/SectionMenu/SectionMenu.jsx
	new file:   browser/extensions/activity-stream/content-src/components/Sections/Sections.jsx
	new file:   browser/extensions/activity-stream/content-src/components/Sections/_Sections.scss
	new file:   browser/extensions/activity-stream/content-src/components/StartupOverlay/StartupOverlay.jsx
	new file:   browser/extensions/activity-stream/content-src/components/StartupOverlay/_StartupOverlay.scss
	new file:   browser/extensions/activity-stream/content-src/components/TopSites/TopSite.jsx
	new file:   browser/extensions/activity-stream/content-src/components/TopSites/TopSiteForm.jsx
	new file:   browser/extensions/activity-stream/content-src/components/TopSites/TopSiteFormInput.jsx
	new file:   browser/extensions/activity-stream/content-src/components/TopSites/TopSites.jsx
	new file:   browser/extensions/activity-stream/content-src/components/TopSites/TopSitesConstants.js
	new file:   browser/extensions/activity-stream/content-src/components/TopSites/_TopSites.scss
	new file:   browser/extensions/activity-stream/content-src/components/Topics/Topics.jsx
	new file:   browser/extensions/activity-stream/content-src/components/Topics/_Topics.scss
	new file:   browser/extensions/activity-stream/content-src/lib/constants.js
	new file:   browser/extensions/activity-stream/content-src/lib/detect-user-session-start.js
	new file:   browser/extensions/activity-stream/content-src/lib/init-store.js
	new file:   browser/extensions/activity-stream/content-src/lib/link-menu-options.js
	new file:   browser/extensions/activity-stream/content-src/lib/section-menu-options.js
	new file:   browser/extensions/activity-stream/content-src/lib/snippets.js
	new file:   browser/extensions/activity-stream/content-src/styles/_activity-stream.scss
	new file:   browser/extensions/activity-stream/content-src/styles/_icons.scss
	new file:   browser/extensions/activity-stream/content-src/styles/_normalize.scss
	new file:   browser/extensions/activity-stream/content-src/styles/_theme.scss
	new file:   browser/extensions/activity-stream/content-src/styles/_variables.scss
	new file:   browser/extensions/activity-stream/content-src/styles/activity-stream-linux.scss
	new file:   browser/extensions/activity-stream/content-src/styles/activity-stream-mac.scss
	new file:   browser/extensions/activity-stream/content-src/styles/activity-stream-windows.scss
	new file:   browser/extensions/activity-stream/contributing.md
	modified:   browser/extensions/activity-stream/data/content/activity-stream.bundle.js
	modified:   browser/extensions/activity-stream/data/content/activity-stream.bundle.js.map
+	new file:   browser/extensions/activity-stream/data/locales.json
	new file:   browser/extensions/activity-stream/docs/ISSUE_TEMPLATE.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/1.GETTING_STARTED.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/data_dictionary.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/data_events.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/mochitests.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/preferences.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/sections.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/snippets.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/telemetry.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/test-merges.md
	new file:   browser/extensions/activity-stream/docs/v2-system-addon/unit_testing_guide.md
	modified:   browser/extensions/activity-stream/install.rdf.in
	new file:   browser/extensions/activity-stream/karma.mc.config.js
	modified:   browser/extensions/activity-stream/lib/ActivityStream.jsm
	new file:   browser/extensions/activity-stream/loaders/inject-loader.js
	new file:   browser/extensions/activity-stream/package-lock.json
	new file:   browser/extensions/activity-stream/package.json
-	deleted:    browser/extensions/activity-stream/test/.eslintrc.js
-	deleted:    browser/extensions/activity-stream/test/functional/mochitest/.eslintrc.js
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/Base.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/Card.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/CollapsibleSection.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/ComponentPerfTimer.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/ConfirmDialog.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/ContextMenu.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/ErrorBoundary.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/LinkMenu.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/ManualMigration.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/Search.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/SectionMenu.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/Sections.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/StartupOverlay.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/TopSites.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/components/Topics.test.jsx
	new file:   browser/extensions/activity-stream/test/unit/content-src/lib/detect-user-session-start.test.js
	new file:   browser/extensions/activity-stream/test/unit/content-src/lib/snippets.test.js
	new file:   browser/extensions/activity-stream/webpack.prerender.config.js
	new file:   browser/extensions/activity-stream/webpack.system-addon.config.js
	new file:   browser/extensions/activity-stream/yamscripts.yml
```

I suppose deleting the eslintrc files will be fine as they aren't being used in m-c anyway. The data/locales.json is a built file (like bin/prerender.js) that hasn't been checked in to m-c recently as it's not used by the add-on, so I think it should be skipped with mcignore as well.

.mcignore
```diff
@@ -0,0 +1,17 @@
+.git/
+node_modules/
+npm-debug.log
+.DS_Store
+*.sw[po]
+*.xpi
+*.pyc
+logs/
+dist/
+ping-centre/
```
Mardak: Just making sure, this is intended to ignore both `/ping-centre/` and `/test/unit/ping-centre`? I think it would be clearer if we were explicit about which things should be root vs matching anywhere. The `/locales/` fix avoided the prerendered missing files. As noted elsewhere, skipping all `.eslintrc.js` files happens to be fine too. But it's unclear if that was actually intentional or not.

Maybe have all the "match anywhere" patterns first then separately all the `/` root patterns (e.g., .git, node_modules, logs, dist)

.mcignore
```diff
@@ -0,0 +1,17 @@
...
+*.update.rdf
+bin/prerender.js
+bin/prerender.js.map
+/locales/
+.gitignore
+.eslintrc.js
+.eslintignore
```
Mardak: Actually, when we ignore all of `activity-stream/**`, we can copy over these eslint files now without breaking `./mach lint`

README.md
```diff
@@ -1,13 +1,16 @@
...
+The files in this directory, including vendor dependencies, are imported from the
+system-addon directory in https://github.com/mozilla/activity-stream.
+
+Read [docs/v2-system-addon](https://github.com/mozilla/activity-stream/tree/master/docs/v2-system-addon/1.GETTING_STARTED.md) for more detail.
 
 ## Where should I file bugs?
 
-Although we regularly check the ActivityStream:NewTab component on Bugzilla, we generally
-like to file bugs at https://github.com/mozilla/activity-stream/issues.
+Although we regularly check the ActivityStream:NewTab component on Bugzilla.
```
Mardak: This is just a fragment. I guess just remove "Although" ;)

.mcignore
k88hudson: Yep, that sounds good

k88hudson: I guess we do want to ignore ping centre tests if we aren't copy the source code

PR #4170, opened 2018-05-25
Fix Bug 1464467 - fix type icon in RTL

rlr:
r? @k88hudson 

Before:
screen shot 2018-05-25 at 3 14 40 pm

After:
screen shot 2018-05-25 at 3 13 28 pm


Mardak:
One thing to note is as made more obvious with the wrong positioning of the type icon, it's possible for it to overlap with the host name text. I guess file a bug with uiwanted.

Mardak:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1464536

PR #4171, opened 2018-05-25
Fix Bug 1464536 - Compact icon type circle can overlap host name

Mardak:
r?@rlr I don't think I've used so many characters to generate "4px" for `clip-path: inset(0 0 4px);`

Before:
screen shot 2018-05-25 at 4 39 49 pm

After:
screen shot 2018-05-25 at 4 39 09 pm

(gold background added to see the clipping)

### DIFF
--- a/system-addon/content-src/components/Card/_Card.scss
+++ b/system-addon/content-src/components/Card/_Card.scss
@@ -219,6 +219,8 @@
 }
 
 .compact-cards {
+  $card-detail-vertical-spacing: 12px;
+
   .card-outer {
     height: $card-height-compact;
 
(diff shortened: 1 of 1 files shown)


system-addon/content-src/components/Card/_Card.scss
```diff
@@ -255,10 +257,11 @@
       $container-size: 32px;
       background-color: var(--newtab-card-background-color);
       border-radius: $container-size / 2;
+      clip-path: inset(0 0 $container-size - ($card-height-compact - $card-preview-image-height-compact - 2 * $card-detail-vertical-spacing));
```
l-hedgehog: Hmmm, I understand this produces the necessary `4px`, but I think this should be `$container-size - $icon-size - $card-detail-vertical-spacing`, that is, height of `.card-context` minus height of overlap with `.card-preview-image` minus the top padding of `.card-details`?

PR #4173, opened 2018-05-29
Fix Bug 1463836 - AS Router Remote Provider Message Previewing

piatra:
Endpoint: https://gist.githubusercontent.com/piatra/70234f08696c0a0509d7ba5568cd830f/raw/68370f34abc134142c64b6f0a9b9258a06de7aa3/messages.json
Example url: `about:newtab?endpoint=https://gist.githubusercontent.com/piatra/70234f08696c0a0509d7ba5568cd830f/raw/68370f34abc134142c64b6f0a9b9258a06de7aa3/messages.json&snippetId=42`
When navigating to this URL it should automatically load the correct snippet. 
On every refresh we want to fetch data from the endpoint again. The workflow would be:
* snippet creator makes v0 of snippet
* copy-paste the url and open it in a new tab
* make changes to snippet -> refresh the same page, get the latest/updated content

ASRouterAdmin will add the endpoint (if found in the URL) only for debugging purposes, snippet creators are interested in landing on `about:newtab` and will probably never see the admin.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -298,7 +298,13 @@ class _ASRouter {
 
   async sendNextMessage(target) {
     const msgs = this._getUnblockedMessages();
-    let message = await ASRouterTargeting.findMatchingMessage(msgs, target);
+    let message = null;
+    const previewMsgs = this.state.messages.filter(item => item.provider === "preview");
+    if (!previewMsgs.length) {
+      message = await ASRouterTargeting.findMatchingMessage(msgs, target);
+    } else {
+      [message] = previewMsgs;
+    }
     await this.setState({lastMessageId: message ? message.id : null});
 
     await this._sendMessageToTarget(message, target);
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -327,12 +327,24 @@ class _ASRouter {
     }
   }
 
+  async _addEndpoint({url, id}) {
+    const providers = [...this.state.providers];
+    if (!providers.find(p => p.url === url)) {
+      // Set update cycle to 0 to fetch new content on every page refresh
+      providers.push({id, type: "remote", url, updateCycleInMs: 0});
+      await this.setState({providers});
+    }
+  }
+
   async onMessage({data: action, target}) {
     switch (action.type) {
       case "CONNECT_UI_REQUEST":
       case "GET_NEXT_MESSAGE":
         // Wait for our initial message loading to be done before responding to any UI requests
         await this.waitForInitialized;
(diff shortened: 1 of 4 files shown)


piatra:
Waiting on a decision https://bugzilla.mozilla.org/show_bug.cgi?id=1465557

lib/ASRouter.jsm
```diff
@@ -307,12 +319,33 @@ class _ASRouter {
     }
   }
 
+  _validPreviewEndpoint(url) {
+    try {
```
k88hudson: Can you maybe throw/log appropriate errors that output useful info (e.g. `The preview URL host ${endpoint.host} is not in the whitelist`) for debugging purposes?

lib/ASRouter.jsm
```diff
@@ -307,12 +319,33 @@ class _ASRouter {
...
+      const endpoint = new URL(url);
```
k88hudson: Should we also allow the SNIPPETS_ENDPOINT here since it's allowed for production?

piatra: I whitelisted both production and preview:
https://github.com/piatra/activity-stream/blob/ef1f34f803703919e61e95ce55faa2c965e8d45c/lib/ASRouter.jsm#L24-L25

We could whitelist whatever SNIPPETS_ENDPOINT points to but is that necessary? If you change the pref value it will fetch the snippets because it goes through a different flow then preview.

piatra: By SNIPPETS_ENDPOINT do you mean the exact URL or the hostname of the endpoint?

lib/ASRouter.jsm
piatra: I went with logging just `url`, I'm assuming the value might be invalid and constructing the URL object can fail/throw.

lib/ASRouter.jsm
k88hudson: oh right, ha! missed the production line somehow... :P

k88hudson: Having the pref be allowed would let you work with a local provider more easily, although can also just use the admin panel for that so this probably fine

lib/ASRouter.jsm
```diff
@@ -327,12 +333,36 @@ class _ASRouter {
...
+      if (!WHITELIST_HOSTS[endpoint.host]) {
+        Cu.reportError(`The preview URL host ${endpoint.host} is not in the whitelist`);
+      }
+      return (endpoint.protocol === "https:" && WHITELIST_HOSTS[endpoint.host]);
```
k88hudson: can you also report an error for if the protocol is not https?

PR #4176, opened 2018-05-29
Fix Bug 1464349 - truncate and nowrap card hostnames

rlr:
Also sliced the hostname to 100 characters which makes the page render waaaaay faster. Just try refreshing :)

r? @Mardak

### DIFF
--- a/system-addon/content-src/components/Card/Card.jsx
+++ b/system-addon/content-src/components/Card/Card.jsx
@@ -165,7 +165,7 @@ export class _Card extends React.PureComponent {
             {link.type === "download" && <div className="card-host-name alternate"><FormattedMessage id={GetPlatformString(this.props.platform)} /></div>}
             {link.hostname &&
               <div className="card-host-name">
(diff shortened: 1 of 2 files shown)


system-addon/content-src/components/Card/Card.jsx
```diff
@@ -165,7 +165,7 @@ export class _Card extends React.PureComponent {
             {link.type === "download" && <div className="card-host-name alternate"><FormattedMessage id={GetPlatformString(this.props.platform)} /></div>}
             {link.hostname &&
               <div className="card-host-name">
-                {link.hostname}{link.type === "download" && `  \u2014 ${link.description}`}
+                {link.hostname.slice(0, 100)}{link.type === "download" && `  \u2014 ${link.description}`}
```
rlr: Maybe the even better we can detect the hostname is a data URI (`startsWith('image/')` maybe?) and then just show the download size (`link.description`) without a hostname?

PR #4177, opened 2018-05-29
Fix Bug 1433217 - Create Onboarding message provider

sarracini:
Is this sort of what you had in mind @k88hudson?


system-addon/lib/OnboardingMessageProvider.jsm
```diff
@@ -0,0 +1,67 @@
...
+      title: "Add-ons",
+      text: "Add even more features that make Firefox work harder for you. Compare prices, check the weather or express your personality with a custom theme.",
+      icon: "addons",
+      button_label: "Try It Now",
+      button_action: "OPEN_ABOUT_PAGE",
+      button_action_params: "addons"
+    }
+  },
+  {
+    id: "ONBOARDING_4",
```
k88hudson: Would you mind making this a separate commit just cause it's easier to see / reference when we export or whatever 🙏

sarracini: Good idea 👍

sarracini:
Ran the image through ImageOptim, brought size from 20.1 kb to 13.8 kb

PR #4178, opened 2018-05-29
Bug 1433216	- Add AS Router targeting

k88hudson:
This patch adds basic support for targeting via JEXL, and a few basic targeting values. 

To test this, set `browser.newtabpage.activity-stream.asrouter.snippetsUrl` to https://gist.github.com/k88hudson/2d627d17ebd3e622e5b05e060d98b182 and try logging in / logging out of firefox accounts.

### DIFF
--- a/system-addon/lib/ASRouterTargeting.jsm
+++ b/system-addon/lib/ASRouterTargeting.jsm
@@ -1,0 +1,60 @@
+ChromeUtils.import("resource://gre/modules/components-utils/FilterExpressions.jsm");
+ChromeUtils.import("resource://gre/modules/Services.jsm");
+ChromeUtils.defineModuleGetter(this, "ProfileAge",
+  "resource://gre/modules/ProfileAge.jsm");
+ChromeUtils.import("resource://gre/modules/Console.jsm");
+
+const FXA_USERNAME_PREF = "services.sync.username";
+
+/**
+ * removeRandomItemFromArray - Removes a random item from the array and returns it.
+ *
+ * @param {Array} arr An array of items
+ * @returns one of the items in the array
+ */
+function removeRandomItemFromArray(arr) {
+  return arr.splice(Math.floor(Math.random() * arr.length), 1)[0];
+}
+
+const TargetingGetters = {
+  get profileAgeCreated() {
+    return new ProfileAge(null, null).created;
+  },
+  get profileAgeReset() {
+    return new ProfileAge(null, null).reset;
+  },
+  get hasFxAccount() {
+    return Services.prefs.prefHasUserValue(FXA_USERNAME_PREF);
+  }
+};
+
+this.ASRouterTargeting = {
+  Environment: TargetingGetters,
+
+  isMatch(filterExpression, context = this.Environment) {
+    return FilterExpressions.eval(filterExpression, context);
+  },
+
+  /**
+   * findMatchingMessage - Given an array of messages, returns one message
+   *                       whos targeting expression evaluates to true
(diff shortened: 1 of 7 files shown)


system-addon/lib/ASRouterTargeting.jsm
```diff
@@ -0,0 +1,52 @@
...
+};
+
+this.ASRouterTargeting = {
+  Environment: TargetingGetters,
+
+  isMatch(filterExpression, context = this.Environment) {
+    return FilterExpressions.eval(filterExpression, context);
+  },
+
+  async findMatchingMessage(originalArray, context) {
```
sarracini: Maybe rename `originalArray` to something more descriptive like `originalMessages`

system-addon/lib/ASRouterTargeting.jsm
```diff
@@ -0,0 +1,52 @@
...
+    const arrayOfItems = [...originalArray];
+    let match;
+    let candidate;
+    while (!match && arrayOfItems.length) {
+      candidate = removeRandomItemFromArray(arrayOfItems);
+      if (candidate && (!candidate.targeting || await this.isMatch(candidate.targeting, context))) {
```
sarracini: Why do we want to match if the candidate has no targeting?

system-addon/test/functional/mochitest/browser_asrouter_targeting.js
```diff
@@ -0,0 +1,68 @@
...
+  is(match, messages[0], "should match and return the correct message");
+});
+
+add_task(async function return_nothing_for_no_matching_message() {
+  const messages = [{id: "bar", targeting: "!FOO"}];
+  const context = {FOO: true};
+
+  const match = await ASRouterTargeting.findMatchingMessage(messages, context);
+
+  ok(typeof match === "undefined", "should match and return the correct message");
```
sarracini: nit: I think you can do `is(match, undefined, "should match and return the correct message")` here. Just for consistency and clarity

system-addon/test/functional/mochitest/browser_asrouter_targeting.js
```diff
@@ -0,0 +1,68 @@
...
+});
+
+add_task(async function checkProfileAgeReset() {
+  let profileAccessor = new ProfileAge();
+  is(await ASRouterTargeting.Environment.profileAgeReset, await profileAccessor.reset,
+    "should return correct profile age reset");
+
+  const message = {id: "foo", targeting: `profileAgeReset == ${await profileAccessor.reset}`};
+  is(await ASRouterTargeting.findMatchingMessage([message]), message,
+    "should select correct item by profile age reet");
```
sarracini: nit: typo here `reset`

system-addon/lib/ASRouterTargeting.jsm
k88hudson: If there's no targeting, presumably that means we could show it in all conditions (otherwise it would never be shown)

system-addon/test/functional/mochitest/browser_asrouter_targeting.js
k88hudson: reeeet

system-addon/lib/ASRouterTargeting.jsm
sarracini: Ah ok. So in the current state that we're in, it's ok that the onboarding tour still gets added to the rotation of messages to show, since it doesn't have any targeting yet?

PR #4179, opened 2018-05-29
Fix Bug 1400536 - add pref for number of rows in sections

rlr:
Whoa, I was super excited to open about:preferences and see this without touching any of that code (🎩 tip @Mardak):

screen shot 2018-05-29 at 5 23 51 pm

### DIFF
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -20,10 +20,7 @@ function getFormattedMessage(message) {
 export class Section extends React.PureComponent {
   get numRows() {
     const {rowsPref, maxRows, Prefs} = this.props;
-    if (rowsPref && Prefs.values[rowsPref] <= maxRows) {
-      return Prefs.values[rowsPref];
-    }
-    return maxRows;
+    return rowsPref ? Prefs.values[rowsPref] : maxRows;
   }
 
   _dispatchImpressionStats() {
--- a/system-addon/content-src/components/Sections/Sections.jsx
+++ b/system-addon/content-src/components/Sections/Sections.jsx
@@ -18,6 +18,14 @@ function getFormattedMessage(message) {
 }
 
 export class Section extends React.PureComponent {
+  get numRows() {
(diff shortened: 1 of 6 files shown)


rlr:
I am going to file a followup bug about increasing maxRows of Top Sites to 4 because that is totally separate code.

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -18,6 +18,14 @@ function getFormattedMessage(message) {
 }
 
 export class Section extends React.PureComponent {
+  get numRows() {
+    const {rowsPref, maxRows, Prefs} = this.props;
+    if (rowsPref && Prefs.values[rowsPref] <= maxRows) {
```
rlr: or should we let a user set the pref to 100 rows of highlights? 🤔

system-addon/content-src/components/Sections/Sections.jsx
```diff
@@ -18,6 +18,14 @@ function getFormattedMessage(message) {
 }
 
 export class Section extends React.PureComponent {
+  get numRows() {
```
rlr: I guess this calls for some unit tests.

system-addon/content-src/components/Sections/Sections.jsx
rlr: to clarify ^. If the user went through the trouble of going into `about:config` to try to get more rows of Highlights. Should we respect that? We might run out of highlights rows from the feed but in most cases we should have enough for more than the max I think.

Mardak: I think about:config overrides is fine. The about:preferences UI will show a blank entry, and something too big will just have a bunch of placeholders where someone adjusting about:config can adjust again. I guess for Pocket, it'll happen to work for 7 rows as we get 20 stories? (Now someone will file a bug saying there's too many empty cards!)

rlr:
^ I removed the pref value check and added some tests.

Mardak:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1465548 for the link location issue

PR #4184, opened 2018-05-30
Fix Bug 1465398 - change max rows for top sites to 4

rlr:
I think this is all we have to do? @Mardak r?

### DIFF
--- a/system-addon/lib/TopSitesFeed.jsm
+++ b/system-addon/lib/TopSitesFeed.jsm
@@ -30,6 +30,7 @@ const MIN_FAVICON_SIZE = 96;
 const CACHED_LINK_PROPS_TO_MIGRATE = ["screenshot", "customScreenshot"];
 const PINNED_FAVICON_PROPS_TO_MIGRATE = ["favicon", "faviconRef", "faviconSize"];
 const SECTION_ID = "topsites";
+const ROWS_PREF = "topSitesRows";
 
 this.TopSitesFeed = class TopSitesFeed {
   constructor() {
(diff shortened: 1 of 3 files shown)


system-addon/lib/TopSitesFeed.jsm
```diff
@@ -88,8 +88,8 @@ this.TopSitesFeed = class TopSitesFeed {
   }
 
   async getLinksWithDefaults() {
-    // Get at least 2 rows so toggling between 1 and 2 rows has sites
-    const numItems = Math.max(this.store.getState().Prefs.values.topSitesRows, 2) * TOP_SITES_MAX_SITES_PER_ROW;
+    // Get at least 4 rows so toggling between 1, 2, 3, 4 rows has sites
+    const numItems = Math.max(this.store.getState().Prefs.values.topSitesRows, 4) * TOP_SITES_MAX_SITES_PER_ROW;
```
rlr: I guess we could only fetch what we need and then refresh whenever the pref is changed. That might be better now that we go up to 4?

Mardak: Pretty sure this will cause a lot of extra storage, e.g,. thumbnails and icons, if the user only has the default 1 row. Perhaps now that adding/removing rows is done through a separate about:preferences page, a slight delay when it updates in the background off screen, it won't be an issue of not having an immediately responsive toggling.

rlr: Yep yep. I think we were more worried about that when preferences was the in-page side panel.

rlr:
ok, that seems to be working for me ^

rlr:
correct, we didnt want individual sites moving around while the user was watching. now it's less likely as they would have to have two windows open and it would be happening in a background window anyway.

PR #4185, opened 2018-05-31
Fix Bug 1465178 - Telemetry source in AS Router is null

piatra:
The events I tested for:
* snippet impressions
* snippet dismiss
* snippet click event (link)
* onboarding "Try it Now" button

"Start Browsing" the button that closes the tour doesn't have telemetry attached, not sure if that is intended or not.

### DIFF
--- a/system-addon/test/unit/asrouter/templates/SimpleSnippet.test.jsx
+++ b/system-addon/test/unit/asrouter/templates/SimpleSnippet.test.jsx
@@ -18,6 +18,13 @@ function mountAndCheckProps(content = {}) {
 }
 
 describe("SimpleSnippet", () => {
+  let sandbox;
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+  });
+  afterEach(() => {
+    sandbox.restore();
+  });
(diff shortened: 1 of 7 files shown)


system-addon/test/unit/asrouter/templates/SimpleSnippet.test.jsx
```diff
@@ -18,6 +18,13 @@ function mountAndCheckProps(content = {}) {
 }
 
 describe("SimpleSnippet", () => {
+  let sandbox;
```
sarracini: did you leave this in accidentally? tests seem to still pass when I remove this chunk

piatra: Oh. Yes.. moved the tests in asrouter-content.test.jsx and forgot to cleanup.

sarracini:
Also yes it's intended that "Start Browsing" doesn't have telemetry right now, because there's actually no other way to close the modal (unless you refresh or open a new tab) so we just assume that if they have other interactions on that page, they've clicked "Start Browsing"

PR #4190, opened 2018-06-04
[WIP] Fix Bug 1465553 - Add targeting to new Onboarding/First Run tour

sarracini:
First pass at some real onboarding targeting using JEXL!!!


system-addon/content-src/asrouter/components/ModalOverlay/_ModalOverlay.scss
```diff
@@ -11,7 +11,7 @@
   top: 0;
   width: 100%;
   display: none;
-  z-index: 100000;
+  z-index: 1000;
```
sarracini: Adjusting the z index here to not be so high - because it goes on top of the startup overlay for about:welcome

system-addon/lib/ASRouter.jsm
```diff
@@ -225,27 +225,38 @@ class _ASRouter {
...
+
+    // First, find all messages of same template. These are potential matching targeting candidates
+    let bundledMessagesOfSameTemplate = this._getUnblockedMessages()
+                                          .filter(msg => msg.bundled && msg.template === originalMessage.template && msg.id !== originalMessage.id);
+
+    while (bundledMessagesOfSameTemplate.length) {
+      // Find a message that matches the targeting context - or break if there are no matching messages
+      const message = await ASRouterTargeting.findMatchingMessage(bundledMessagesOfSameTemplate, target);
+      if (!message) {
+        /* istanbul ignore next */ // Code coverage in mochitests
```
sarracini: need to still write a mochitest for this.....

system-addon/lib/OnboardingMessageProvider.jsm
```diff
@@ -40,7 +42,8 @@ const ONBOARDING_MESSAGES = [
       button_label: "Try It Now",
       button_action: "OPEN_ABOUT_PAGE",
       button_action_params: "addons"
-    }
+    },
+    targeting: "isFirstRun && isInExperimentCohort == 1"
```
k88hudson: Wouldn't this be `0` and `1` based on your default value of `0` above?

sarracini: i was thinking of having 0 be like no experiment running kind of thing, and then 1 or 2 based on the cohort. that way if for some reason the pref gets read wrong or something then it won't accidentally get put into a cohort. there just won't be any messages shown

k88hudson: ah, that makes sense ok

system-addon/lib/ASRouterTargeting.jsm
```diff
@@ -25,14 +26,21 @@ const TargetingGetters = {
...
   get hasFxAccount() {
     return Services.prefs.prefHasUserValue(FXA_USERNAME_PREF);
+  },
+  get isInExperimentCohort() {
+    return Services.prefs.getIntPref(ONBOARDING_EXPERIMENT_PREF, 0);
   }
 };
 
+function EnvironmentTargeting(target) {
+  return {isFirstRun: target.url === "about:welcome"};
```
sarracini: The only thing that's unclear to me here is that doing it this way means that this experiment is dependent on the `about:welcome` experiment. That means that some of the users in the `about:welcome` cohort will be in these cohorts. I'm assuming that's fine? Forever confused about how these experiments are supposed to run together.....

lib/ASRouterTargeting.jsm
```diff
@@ -25,14 +26,21 @@ const TargetingGetters = {
   },
   get hasFxAccount() {
     return Services.prefs.prefHasUserValue(FXA_USERNAME_PREF);
+  },
+  get isInExperimentCohort() {
```
k88hudson: can you add a comment that this is a temporary targeting value for the purpose of the experiment?

sarracini:
@k88hudson If you could take just another quick look, I made the changes in ASRouter.jsm that we talked about in person last week about adding a `force` param so that we can show them without targeting in the `about:newtab#asrouter` admin page. When `force = true`, we skip targeting and just show the messages that have the same template + bundling criteria. Plus I added an additional test to check both code paths

sarracini:
Made a couple style adjustments:
1. fixed the z-index with the prefs-button; i was still clickable when the modal was open so i decreased the z-index and bumped the z-index of the modal
2. made some improvements on the very very narrow view by adding another media query for very very small screens:

very narrow:
screen shot 2018-06-19 at 10 48 40 am
very very narrow:
screen shot 2018-06-19 at 10 49 52 am

You get some spacing for the text to keep flowing under, and a scrollbar since it's much longer

sarracini:
@k88hudson ok check out just the last commit with the onboarding experiment pref... i'm not sure if that logic is right ..... the snippets stuff is confusing ................

content-src/lib/snippets.js
```diff
@@ -388,7 +388,7 @@ export function addSnippetsSubscriber(store) {
     // snippets.initialized:                  Is SnippetsProvider currently initialised?
     if (state.Prefs.values["feeds.snippets"] &&
       // If the message center experiment is enabled, don't show snippets
-      !state.Prefs.values.asrouterExperimentEnabled &&
+      (!state.Prefs.values.asrouterExperimentEnabled && state.Prefs.values.asrouterOnboardingCohort === 0) &&
```
k88hudson: To make this less confusing I would create a variable up top that's like `isASRouterEnabled`and used that instead

content-src/lib/snippets.js
```diff
@@ -383,12 +383,13 @@ export function addSnippetsSubscriber(store) {
 
   store.subscribe(async () => {
     const state = store.getState();
+    const isASRouterEnabled = !state.Prefs.values.asrouterExperimentEnabled && state.Prefs.values.asrouterOnboardingCohort === 0;
```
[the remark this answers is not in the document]
sarracini: and then check `!isASRouterEnabled` in the check?

PR #4194, opened 2018-06-06
fix: Update docs to include about:welcome

ericawright:
r= @sarracini 
r=@ncloudioj

I edited some telemetry documents to include about:welcome, since these events will also send on this page, and will include "about:welcome" as a string for the `page` key.

questions: 
- in `data_events` around line 32 - how should I reword that to include about:welcome, or should I? 
- should I have changed the line about`Joi`, in `test/schemas/pings.js`

### DIFF
--- a/docs/v2-system-addon/data_dictionary.md
+++ b/docs/v2-system-addon/data_dictionary.md
@@ -39,7 +39,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
   "addon_version": "1.0.0",
   "client_id": "374dc4d8-0cb2-4ac5-a3cf-c5a9bc3c602e",
   "locale": "en-US",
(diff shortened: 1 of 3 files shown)


ncloudioj:
> how should I reword that to include about:welcome, or should I?

@ericawright Could you just update the `page` field for those about:welcome specific events?

Otherwise, it LGTM. r=nanj

ericawright:
@ncloudioj it is possible all of these events could be occurring on about:welcome though. the only events that would not occur on about:welcome would be anything to do with snippets

ncloudioj:
I recall that we only added two new events "SUBMIT_EMAIL" and "SKIPPED_SIGNIN" for about:welcome, is that correct? Are there any other telemetry events from about:welcome? @sarracini

PR #4201, opened 2018-06-21
Fix Bug 1469823 - Listen for PLACES_LINK_DELETED actions and update Topsites

piatra:
Looking back at the [discussion in the PR](https://github.com/mozilla/activity-stream/pull/4184#discussion_r191934937) the original change so that we don't fetch `4 * TOP_SITES_MAX_SITES_PER_ROW` unnecessarily, but I didn't see this usecase (when the user removes a topsites) being discussed. 
This would fetch an extra 4 items to be used in case a hole needs to be filled in.
It also seems like Highlights deduping uses all the `TopSites.rows` in the store (not just the ones being displayed) so this would also need to be fixed. 

f? @Mardak

### DIFF
--- a/lib/TopSitesFeed.jsm
+++ b/lib/TopSitesFeed.jsm
@@ -395,6 +395,7 @@ this.TopSitesFeed = class TopSitesFeed {
       // All these actions mean we need new top sites
       case at.MIGRATION_COMPLETED:
       case at.PLACES_HISTORY_CLEARED:
+      case at.PLACES_LINK_DELETED:
         this.frecentCache.expire();
         this.refresh({broadcast: true});
         break;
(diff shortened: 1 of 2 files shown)


Mardak:
I would expect blocking a top site to cause a refresh to fill in new data?

piatra:
We do listen for blocked events https://github.com/mozilla/activity-stream/blob/master/lib/PlacesFeed.jsm#L181 but removing from history doesn't notify any observer events https://searchfox.org/mozilla-central/rev/d0a41d2e7770fc00df7844d5f840067cc35ba26f/toolkit/components/places/History.jsm#302

Mardak:
Oh, it's not dismissing/blocking. It's deleting from history. If it is calling the linked History.jsm method, the comment says it should trigger notifications:
https://searchfox.org/mozilla-central/rev/d0a41d2e7770fc00df7844d5f840067cc35ba26f/toolkit/components/places/History.jsm#281
   * Any change may be observed through nsINavHistoryObserver

I would guess PLACES_LINK_DELETED would be dispatched, and we just handling it from TopSitesFeed.

piatra:
Yes we can listen for LINK_DELETED and trigger a `refresh({broadcast: true})` I'll update the PR.

PR #4202, opened 2018-06-21
Fix Bug 1468978 - Update schema to use format: uri for links

piatra:
Documentation https://spacetelescope.github.io/understanding-json-schema/reference/string.html#format

@k88hudson do you know if there's a better way to do this?
@glogiotatidis will this update to the schema help you format things on the server side?

### DIFF
--- a/content-src/asrouter/templates/SimpleSnippet/SimpleSnippet.schema.json
+++ b/content-src/asrouter/templates/SimpleSnippet/SimpleSnippet.schema.json
@@ -1,16 +1,35 @@
 {
   "title": "SimpleSnippet",
   "description": "A simple template with an icon, text, and optional button.",
-  "version": "1.0.1",
+  "version": "1.1.1",
   "type": "object",
+  "definitions": {
+    "plainText": {
+      "description": "Plain text (no HTML allowed)",
+      "type": "string"
+    },
+    "richText": {
+      "description": "Text with HTML subset allowed: i, b, u, strong, em, br",
+      "type": "string"
+    },
+    "link_url": {
+      "description": "Target for links or buttons",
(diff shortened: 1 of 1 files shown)


k88hudson:
`format` seems to be OK for uris, although I'm not sure how many implementations actually support it. That also won't cover the html/plain text case.

You could use `$ref` and `definitions` (it's supported, although a bit cumbersome):

```json
{
  "properties": {
    "title": {
      "allOf": [
        {"$ref": "#/definitions/plainText"},
        {"description": "Snippet title displayed before snippet text"}
      ]
    },
    "text": {
      "allOf": [
        {"$ref": "#/definitions/richText"},
        {"description": "Main body text of snippet."}
      ]
    }
  },
  "definitions": {
    "plainText": {
      "description": "Plain text (no HTML allowed)",
      "type": "string",
      "pattern": "/...../"
    },
    "richText": {
      "description": "Text with HTML subset allowed: i, b, u, strong, em, br",
      "type": "string"
    }
  }
}
```

Alternatively, we could do something non-standard, like the web extension folks do (see [this schema](https://searchfox.org/mozilla-central/source/browser/components/extensions/schemas/history.json#287) for example), which does seem a lot nicer but might not have support in a validator

k88hudson:
If we don't care about actually programmatically validating `format`, we could also just define our own set too

piatra:
Looking at the example linked from components/extensions `url` fields also have a plain `string` type and a `description`.

piatra:
@k88hudson let me know if this is closer to what you had in mind.

glogiotatidis:
> @glogiotatidis will this update to the schema help you format things on the server side?

Looks good, thanks!

PR #4204, opened 2018-06-21
Bug 1458658 - Unblock sites that are added

sarracini:
Quick fix + a test for this bug. Easy win and makes unblocking a lower priority


piatra:
Nit: could you rename this to "Unblock sites that are added if they were previously blocked" bugzilla history shows there might have been some edit-merge-conflicts and the longer title makes it more clear.

PR #4210, opened 2018-06-22
Port Bug 1347204 - Implement Google Chrome New Tab Page color properties

nt1m:


piatra:
Thanks for your PR, always happy to remove some code. Unfortunatelly it seems that this breaks current behaviour: [the ThemeFeed](https://github.com/mozilla/activity-stream/blob/master/lib/ThemeFeed.jsm#L39) used to broadcast when the default dark theme was selected, AS content [would pick up on that](https://github.com/mozilla/activity-stream/blob/0287617d8a742bee8967aeff51fc0d19f364ab69/content-src/components/Base/Base.jsx#L31) and add a class name enabling dark theme. With your changes if you customize/change to dark theme AS will not update.

sarracini:
What is this blocked on?

piatra:
Waiting for the patch to land in m-c.

content-src/styles/_theme.scss
```diff
@@ -77,7 +77,7 @@ body {
 }
 
 // Dark theme
-.dark-theme {
+body[lwt-newtab-brighttext] {
```
piatra: We have a sass linter that only returns warnings (won't fail on travis) and it's complaining about this line. In this case I wouldn't recommend fixing it, instead you can just suppress the warnings by adding this at the end.
```
// sass-lint:disable-line force-attribute-nesting, no-qualifying-elements
```
You can run the linter by running `npm run lint` (notice that we have quite a few but we're trying not to add more).

Mardak:
https://hg.mozilla.org/mozilla-central/rev/ca8e1213f5ed has landed, so we'll need to merge this before the next export.

PR #4211, opened 2018-06-22
Bug 1448359 - Enable CSP for Activity Stream

k88hudson:
Alright, I think this should cover what we want. I changed a few things:
* images no longer allow `http` (I think that's OK? I don't think we use any external images right now anyway)
*  I added `default-src none` which should cover frames, object-src, etc.
* `connect-src` is `https` only (i.e. fetching). This is only used for snippets as far as I can think of right now.

The only thing I am a bit worried about is if the (legacy) snippets url is OK as `https` only – @glogiotatidis @pmac do you see this as being an issue possibly?


pmac:
HTTPS only should be fine for snippets.

PR #4212, opened 2018-06-25
Fix Bug 1470170 - Implement priority system for AS Router messages

sarracini:
This PR allows for a notion of "triggers" which act as a way to have priority in AS Router. This means that messages can be both targeted, for things that are long lasting (e.g locale, list of addons, prefs etc) but they can also be shown when a trigger happens - a one time event (e.g first new tab on a fresh profile, new addon installed, etc). Basically the content requests that a trigger message be shown under some condition (in this case we check the url to be about:welcome, indicating we are in a first run case) and then AS Router searches for messages that should be shown on a 'firstRun' trigger. Picks those, and shows them.

New method  `findMatchingMessageWithTrigger` called when content sends data 'trigger', and then fall back to the original `findMatchingMessage` if no messages with the trigger are found

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -232,7 +232,22 @@ class _ASRouter {
     this.messageChannel.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {type: "ADMIN_SET_STATE", data: state});
   }
 
-  async _getBundledMessages(originalMessage, target, force = false) {
+  async _findMessage(msgs, target, data = {}) {
+    let message;
+    let {trigger} = data;
+    if (trigger) {
+      // Find a message that matches the targeting context as well as the trigger context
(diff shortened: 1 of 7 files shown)


lib/ASRouter.jsm
```diff
@@ -251,7 +266,7 @@ class _ASRouter {
     } else {
       while (bundledMessagesOfSameTemplate.length) {
         // Find a message that matches the targeting context - or break if there are no matching messages
-        const message = await ASRouterTargeting.findMatchingMessage(bundledMessagesOfSameTemplate, target);
+        let message = await this._findMessage(bundledMessagesOfSameTemplate, target, data);
```
k88hudson: why the change to `let`

sarracini: because of copy pasting! i fix

PR #4218, opened 2018-06-28
Fix Bug 1471514 - Add top-of-funnel metrics for fxa

ericawright:
@sarracini @vladikoff

~To Test: change line 32 from "https://accounts.firefox.com/metrics-flow" to "https://latest.dev.lcip.org/metrics-flow"
Currently there is a CORs issue with accounts.firefox.com and the changes required will hopefully be live in a few days.~

re: https://github.com/mozilla/fxa-content-server/issues/6267
re: https://github.com/mozilla/fxa-content-server/issues/5882

### DIFF
--- a/content-src/components/StartupOverlay/StartupOverlay.jsx
+++ b/content-src/components/StartupOverlay/StartupOverlay.jsx
@@ -1,5 +1,5 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 import {FormattedMessage, injectIntl} from "react-intl";
-import {actionCreators as ac} from "common/Actions.jsm";
 import {connect} from "react-redux";
 import React from "react";
 
@@ -15,8 +15,29 @@ export class _StartupOverlay extends React.PureComponent {
 
     this.state = {
       emailInput: "",
-      overlayRemoved: false
+      overlayRemoved: false,
+      flowId: "",
+      flowBeginTime: ""
     };
+    this.didFetch = false;
+  }
+
+  async componentWillUpdate() {
+    if (this.props.fxa_endpoint && !this.didFetch) {
+      try {
+        this.didFetch = true;
+        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);
+        if (response.status === 200) {
+          const {flowId, flowBeginTime} = await response.json();
+          this.setState({flowId, flowBeginTime});
+        }
+      } catch (error) {
+        this.props.dispatch(ac.OnlyToMain({type: at.TELEMETRY_UNDESIRED_EVENT, data: {value: "FXA_METRICS_ERROR"}}));
+      }
+    }
+  }
+
+  componentDidMount() {
     this.initScene();
   }
 
@@ -70,8 +91,9 @@ export class _StartupOverlay extends React.PureComponent {
       return null;
     }
 
-    let termsLink = (<a href="https://accounts.firefox.com/legal/terms" target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_terms_of_service" /></a>);
-    let privacyLink = (<a href="https://accounts.firefox.com/legal/privacy" target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_privacy_notice" /></a>);
+    let termsLink = (<a href={`${this.props.fxa_endpoint}/legal/terms`} target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_terms_of_service" /></a>);
+    let privacyLink = (<a href={`${this.props.fxa_endpoint}/legal/privacy`} target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_privacy_notice" /></a>);
+
     return (
       <div className={`overlay-wrapper ${this.state.show ? "show " : ""}`}>
         <div className="background" />
@@ -84,13 +106,15 @@ export class _StartupOverlay extends React.PureComponent {
             </div>
             <div className="firstrun-sign-in">
               <p className="form-header"><FormattedMessage id="firstrun_form_header" /><span className="sub-header"><FormattedMessage id="firstrun_form_sub_header" /></span></p>
-              <form method="get" action="https://accounts.firefox.com" target="_blank" rel="noopener noreferrer" onSubmit={this.onSubmit}>
+              <form method="get" action={this.props.fxa_endpoint} target="_blank" rel="noopener noreferrer" onSubmit={this.onSubmit}>
                 <input name="service" type="hidden" value="sync" />
                 <input name="action" type="hidden" value="email" />
                 <input name="context" type="hidden" value="fx_desktop_v3" />
                 <input name="entrypoint" type="hidden" value="activity-stream-firstrun" />
                 <input name="utm_source" type="hidden" value="activity-stream" />
                 <input name="utm_campaign" type="hidden" value="firstrun" />
+                <input name="flow_id" type="hidden" value={this.state.flowId} />
+                <input name="flow_begin_time" type="hidden" value={this.state.flowBeginTime} />
                 <span className="error">{this.props.intl.formatMessage({id: "firstrun_invalid_input"})}</span>
                 <input className="email-input" name="email" type="email" required="true" onInvalid={this.onInputInvalid} placeholder={this.props.intl.formatMessage({id: "firstrun_email_input_placeholder"})} onChange={this.onInputChange} />
                 <div className="extra-links">
@@ -112,4 +136,5 @@ export class _StartupOverlay extends React.PureComponent {
   }
 }
 
-export const StartupOverlay = connect()(injectIntl(_StartupOverlay));
+const getState = state => ({fxa_endpoint: state.Prefs.values.fxa_endpoint});
+export const StartupOverlay = connect(getState)(injectIntl(_StartupOverlay));
--- a/lib/PrefsFeed.jsm
+++ b/lib/PrefsFeed.jsm
@@ -103,6 +103,9 @@ this.PrefsFeed = class PrefsFeed {
     values.isPrivateBrowsingEnabled = PrivateBrowsingUtils.enabled;
     values.platform = AppConstants.platform;
 
+    // Get the firefox accounts url to send firstrun metrics to.
+    values.fxa_endpoint = Services.prefs.getStringPref("identity.fxaccounts.remote.root", "https://accounts.firefox.com");
+
     // Set the initial state of all prefs in redux
     this.store.dispatch(ac.BroadcastToContent({type: at.PREFS_INITIAL_VALUES, data: values}));
 


content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -12,12 +12,26 @@ export class _StartupOverlay extends React.PureComponent {
...
 
   initScene() {
+    let req = new XMLHttpRequest();
+    req.onreadystatechange = () => {
+      if (req.readyState === 4 && req.status === 200) {
+        this.setFormFields(req.responseText);
+      }
+    };
+    req.onerror = e => {
+      console.error("fxa error", e);
```
piatra: Should we report these to telemetry?

ericawright: sure!

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -12,12 +12,26 @@ export class _StartupOverlay extends React.PureComponent {
...
     this.onInputInvalid = this.onInputInvalid.bind(this);
+    this.setFormFields = this.setFormFields.bind(this);
 
-    this.state = {emailInput: ""};
+    this.state = {emailInput: "", flowId: "", flowBeginTime: ""};
     this.initScene();
   }
 
   initScene() {
+    let req = new XMLHttpRequest();
```
sarracini: Is there any reason you can't use the regular fetch API here instead of XMLHttpRequest? https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -34,6 +48,11 @@ export class _StartupOverlay extends React.PureComponent {
     }, 400);
   }
 
+  setFormFields(r) {
```
sarracini: Why a new function? Why not just set the state in the callback of the request when you have the response handy?

content-src/components/StartupOverlay/StartupOverlay.jsx
ericawright: @vladikoff, would fxa use/want this data?

vladikoff: @ericawright not for now, worst case we can look at telemetry I guess!

ericawright: @vladikoff that's what I mean, if I store this event in telemety would you use it, since AS doesn't need it.

vladikoff: @ericawright oh, if it is easy enough and you have other errors already sent to telemetry let's do it! We can later look if the request fails for users

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -12,12 +12,26 @@ export class _StartupOverlay extends React.PureComponent {
...
+    req.onreadystatechange = () => {
+      if (req.readyState === 4 && req.status === 200) {
+        this.setFormFields(req.responseText);
+      }
+    };
+    req.onerror = e => {
+      console.error("fxa error", e);
+    };
+
+    req.open("GET", "https://accounts.firefox.com/metrics-flow");
```
vladikoff: can we use any existing FxA settings or setup a setting in activity stream that can be changed in `about:config` to alter this url?

ericawright: I found identity.fxaccounts.remote.root, but it sounds like you'd want to be able to change the whole thing.
I'll make a new one, how about identity.fxaccounts.remote.metrics?
(I also noticed there are urls in about:config for legal/terms and legal/privacy - I'll add those to a different PR)

ericawright: though perhaps adding a new pref for this url should also go in a separate PR since this one will need to be uplifted

ericawright: I take that back, I'll use a pref and fallback to the hardcoded url if there is no pref, then if the pref code doesn't get uplifted this will at least still work, and then enterprise folks or others who wish to block this could.
sound okay?

ericawright:
@sarracini I fixed the failing test, not sure if I've done it correctly with a  fetch stub. Also, added a pref for the metrics endpoint. I would prefer the pref gets set in product and not from AS - though that may be an irrational concern.

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -13,11 +15,20 @@ export class _StartupOverlay extends React.PureComponent {
...
   }
 
   initScene() {
+    window.fetch(this.props.fxa_metrics_endpoint)
+      .then(response => response.json())
+      .then(response => {
+        this.setState({flowId: response.flowId, flowBeginTime: response.flowBeginTime});
+      })
+      .catch(error => {
+        this.props.dispatch(ac.UndesiredEvent({type: "FXA_METRICS_ERROR", error}));
```
vladikoff: awesome! 👍

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -2,6 +2,8 @@ import {FormattedMessage, injectIntl} from "react-intl";
 import {actionCreators as ac} from "common/Actions.jsm";
 import {connect} from "react-redux";
 import React from "react";
+/* eslint-disable import/no-commonjs */
+// const fetch = require("node-fetch");
```
sarracini: leftover?

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -80,6 +89,8 @@ export class _StartupOverlay extends React.PureComponent {
                 <input name="entrypoint" type="hidden" value="activity-stream-firstrun" />
```
vladikoff: @ericawright this form above has `accounts.firefox.com` hardcoded vs using `this.props.fxa_metrics_endpoint` like the fetch script. Is it possible to fix that?

ericawright: yeah I think that's already in the prefs, I'll find it

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -13,11 +13,20 @@ export class _StartupOverlay extends React.PureComponent {
     this.removeOverlay = this.removeOverlay.bind(this);
     this.onInputInvalid = this.onInputInvalid.bind(this);
 
-    this.state = {emailInput: ""};
+    this.state = {emailInput: "", flowId: "", flowBeginTime: ""};
     this.initScene();
   }
 
   initScene() {
+    window.fetch(this.props.fxa_endpoint)
```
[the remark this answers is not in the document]
sarracini: This is actually not working as intended. If you log `this.props.fxa_endpoint` at this point, it's undefined. Therefore this actually throws all the time. You can put a log in the `.catch` and you can see that it throws:
`TypeError: "undefined is not a valid URL."`
The pref you're looking for is not ready in the constructor of this component unless you put it in the prerendered data's initial prefs values. The prefs get sent too late for this, and so you never get any of the data you're looking for, and it attempts to do a fetch on `undefined` which will throw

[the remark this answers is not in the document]
ericawright: I'm not getting that error, I was getting a cors error (since I accidentally lost the /metrics-flow portion of the URl when changing the contents of the pref.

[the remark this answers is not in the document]
sarracini: screen shot 2018-07-04 at 11 07 13 am

lib/ActivityStream.jsm
```diff
@@ -170,6 +170,10 @@ const PREFS_CONFIG = new Map([
   ["asrouter.snippetsUrl", {
     title: "A custom URL for the AS router snippets",
     value: ""
+  }],
```
ericawright: seems to need to be here and in the prerender, else it gets erased

common/PrerenderData.jsm
```diff
@@ -63,7 +63,8 @@ this.PrerenderData = new _PrerenderData({
     "feeds.section.topstories": true,
     "feeds.section.highlights": true,
     "sectionOrder": "topsites,topstories,highlights",
-    "collapsed": false
+    "collapsed": false,
+    "fxaccounts.endpoint": "https://accounts.firefox.com"
```
sarracini: So this won't be the right approach either. Remove this from here, remove the pref from ActivityStream.jsm, and add a look up for the pref: identity.fxaccounts.remote.root with a fallback value in PrefsFeed after values.platform:

```js
values.fxa_endpoint = Services.prefs.getStringPref("identity.fxaccounts.remote.root", "https://accounts.firefox.com");
```

Then in StartupOverlay.jsx connect Prefs:
```js
const getState = state => ({fxa_endpoint: state.Prefs.values.fxa_endpoint});
export const StartupOverlay = connect(getState)(injectIntl(_StartupOverlay));
```

And then you can use this.props.fxa_endpoint directly in the component

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -15,12 +15,24 @@ export class _StartupOverlay extends React.PureComponent {
...
-      overlayRemoved: false
+      overlayRemoved: false,
+      flowId: "",
+      flowBeginTime: ""
     };
     this.initScene();
   }
 
-  initScene() {
+  async initScene() {
```
sarracini: Call this in `componentDidMount`, I just noticed this was in the constructor and it shouldn't be there

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -15,12 +15,24 @@ export class _StartupOverlay extends React.PureComponent {
...
+    try {
+      let response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);
```
sarracini: Do this fetch stuff in componentWillUpdate, that's when the props from the connected Prefs will be ready. Any earlier than that and our store isn't ready.
Also add a flag to make sure we're not doing multiple fetchs when we update:

```js
  async componentWillUpdate() {
    try {
      if (!this.props.fxa_endpoint || this.didFetch) {
        return;
      }
      this.didFetch = true;
      const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);
      if (response.status === 200) {
        const {flowId, flowBeginTime} = await response.json();
        this.setState({flowId, flowBeginTime});
      }
    } catch (error) {
      this.props.dispatch(ac.OnlyToMain({type: at.TELEMETRY_UNDESIRED_EVENT, data: {value: "FXA_METRICS_ERROR"}}));
    }
  }
```

lib/ActivityStream.jsm
sarracini: You can remove this with the changes mentioned above

common/PrerenderData.jsm
vladikoff: Amazing!

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -28,6 +41,18 @@ export class _StartupOverlay extends React.PureComponent {
     }, 10);
   }
 
+  async setFxaMetrics() {
```
sarracini: Again, no need for a different function here, just adds an extra bind in the constructor that you don't need.

content-src/components/Base/Base.jsx
```diff
@@ -152,7 +152,7 @@ export class BaseContent extends React.PureComponent {
             <ConfirmDialog />
           </main>
         </div>
-        {this.props.isFirstrun && <StartupOverlay />}
+        {this.props.isFirstrun && <StartupOverlay fxa_endpoint={prefs["fxaccounts.endpoint"]} />}
```
sarracini: you don't need this prop here, you're getting the prop from the connected state in StartupOverlay.jsx

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -12,11 +12,24 @@ export class _StartupOverlay extends React.PureComponent {
...
       emailInput: "",
-      overlayRemoved: false
+      overlayRemoved: false,
+      flowId: "",
+      flowBeginTime: ""
     };
+  }
+
+  componentWillUpdate() {
+    if (this.props.fxa_endpoint && !this.didFetch) {
```
sarracini: You need to define this.didFetch in the constructor with value of false

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -12,11 +12,24 @@ export class _StartupOverlay extends React.PureComponent {
...
+      this.didFetch = true;
+      this.setFxaMetrics();
```
sarracini: do you need to `await` on this function here since it's async?

lib/PrefsFeed.jsm
```diff
@@ -102,6 +102,7 @@ this.PrefsFeed = class PrefsFeed {
     // computed in main process
     values.isPrivateBrowsingEnabled = PrivateBrowsingUtils.enabled;
     values.platform = AppConstants.platform;
+    values.fxa_endpoint = Services.prefs.getStringPref("identity.fxaccounts.remote.root", "https://accounts.firefox.com");
```
sarracini: a comment here would be helpful about why this is necessary

content-src/components/StartupOverlay/StartupOverlay.jsx
sarracini: if you remove this function and just put the logic in componentWillUpdate then you can disregard this comment

ericawright: I was going to put the logic in the componentWillUpdate, but then it runs multiple times being async, the flag does not get set in time to stop the second time from running. 
I was also considering, on failure, doing a setTimout and running this again up to 2 or 3 times (circumventing the flag) just in case there is a temporary error, like a bad connection or something.

content-src/components/StartupOverlay/StartupOverlay.jsx
ericawright: sure, I figured undefined worked well enough

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -15,8 +15,29 @@ export class _StartupOverlay extends React.PureComponent {
...
+      flowBeginTime: ""
     };
+    this.didFetch = false;
+  }
+
+  async componentWillUpdate() {
+    if (this.props.fxa_endpoint && !this.didFetch) {
+      try {
+        this.didFetch = true;
+        let response = await fetch(`/metrics-flow`);
```
sarracini: supposed to be 
```js
const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);`
```

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -15,8 +15,29 @@ export class _StartupOverlay extends React.PureComponent {
...
+      flowBeginTime: ""
     };
+    this.didFetch = false;
+  }
+
+  async componentWillUpdate() {
+    if (this.props.fxa_endpoint && !this.didFetch) {
+      try {
+        this.didFetch = true;
+        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);
```
Mardak: Looks like this is causing test failure timeout:
TEST-UNEXPECTED-FAIL | browser/base/content/test/urlbar/browser_urlbar_blanking.js | Found a tab after previous test timed out: about:welcome -
https://treeherder.mozilla.org/logviewer.html#?job_id=187870714&repo=try&lineNumber=1633

Where the test loads gInitialPages:
https://searchfox.org/mozilla-central/source/browser/base/content/browser.js#321

The test times out with about:welcome open and status bar showing "Performing TLS handshake to 127.0.0.1"

Mardak: Ah, during testing, it's set to {server} so 127.0.0.1, and the request silently fails:

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://127.0.0.1:8888//metrics-flow. (Reason: CORS request did not succeed).

Mardak: The following works because it looks like the test/metrics-flow endpoint just never resolves in time for the test:
```js
const controller = new AbortController();
setTimeout(() => controller.abort(), 5000);
const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`, {signal: controller.signal});
```

Unclear how long we should wait before aborting (and triggering a FXA_METRICS_ERROR…

ericawright: @mardak would you prefer something like that, or to edit the test?

sarracini: It's not just a single test, so you won't be able to edit the test

ericawright: true, it would fail anywhere this page is used wouldn't it

Mardak:
Reverted in https://github.com/mozilla/activity-stream/commit/245cf97a5847d7885328347f6335da32c5549008 for the mochitest failures

PR #4219, opened 2018-06-28
Backport Bug 1463663 - Rename sendAsyncMessage to RPMSendAsyncMessage

piatra:
There's still a eslint error because it doesn't know about this new function name and I haven't updated tests.


piatra:
Updated the tests. There needs to be another follow up in which we update eslint. The patch has landed but eslint-mozilla-plugin needs to make a new release.

sarracini:
Added a commit to allow `RPMSendAsyncMessage` to be globally defined for now, until there's a release which includes this function name. We will remove it once that is fixed

PR #4224, opened 2018-06-29
Bug 1471695 - add Addon targeting to ASRouter

AdamHillier:


lib/ASRouterTargeting.jsm
```diff
@@ -27,6 +29,10 @@ const TargetingGetters = {
   get hasFxAccount() {
     return Services.prefs.prefHasUserValue(FXA_USERNAME_PREF);
   },
+  get activeAddons() {
+    return AddonManager.getActiveAddons(["extension", "service"])
+      .then(({addons}) => addons.map(addon => addon.id));
```
k88hudson: Can you explicitly include the fields defined here, as well the `fullData` property?

lib/ASRouterTargeting.jsm
```diff
@@ -27,6 +29,10 @@ const TargetingGetters = {
   get hasFxAccount() {
     return Services.prefs.prefHasUserValue(FXA_USERNAME_PREF);
   },
+  get activeAddons() {
```
k88hudson: maybe call this `addonsInfo` to be more consistent with the old API

k88hudson:
🔥

PR #4225, opened 2018-07-02
Fix Bug 1472209 - Remove StartupOverlay if the tour is skipped

piatra:
issue described in https://github.com/piatra/activity-stream/blob/1cb7444a7bd1c653e87512d5b9e4ce83e454c9b2/content-src/components/StartupOverlay/StartupOverlay.jsx#L62-L63

### DIFF
--- a/content-src/components/StartupOverlay/StartupOverlay.jsx
+++ b/content-src/components/StartupOverlay/StartupOverlay.jsx
@@ -13,7 +13,10 @@ export class _StartupOverlay extends React.PureComponent {
     this.removeOverlay = this.removeOverlay.bind(this);
     this.onInputInvalid = this.onInputInvalid.bind(this);
 
-    this.state = {emailInput: ""};
+    this.state = {
+      emailInput: "",
+      overlayRemoved: false
+    };
     this.initScene();
   }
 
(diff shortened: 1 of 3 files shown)


content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -59,10 +59,16 @@ export class _StartupOverlay extends React.PureComponent {
...
+    // about:welcome, prop.isFirstrun is true and StartupOverlay is rendered
+    if (!this.state.show) {
+      return null;
+    }
+
     let termsLink = (<a href="https://accounts.firefox.com/legal/terms" target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_terms_of_service" /></a>);
     let privacyLink = (<a href="https://accounts.firefox.com/legal/privacy" target="_blank" rel="noopener noreferrer"><FormattedMessage id="firstrun_privacy_notice" /></a>);
     return (
-      <div className={`overlay-wrapper ${this.state.show ? "show " : ""}`}>
+      <div className="overlay-wrapper show">
```
Mardak: Isn't the non-show overlay-wrapper needed so that there's the transition? @ericawright ?

piatra:
Good catch! I updated the PR: 
* the overlay still needs to removed, I deferred that to happen after the animation ends
* we need to prevent the `welcome` class name from being added when `updateTheme` is called because it hides the overflow
  * the observation here is that we only need to add it once so I moved it to `componentWillMount`

PR #4229, opened 2018-07-04
Fix Bug 1459449 - Upgrate webpack

k88hudson:
In this patch I have upgraded webpack to 4.15.0. I have tested normal/debug versions of AS as well as buildmc, startmc, tddmc and test. Please verify everything is working as expected for you.

Here is a link to the [diff of the bundle](https://gist.github.com/k88hudson/66a01864c1de4541e190df8faaab79da)

The only real change that was needed was adding a [mode of "none"](https://webpack.js.org/concepts/mode/#mode-none) to preserve our existing settings. I'm assuming we're not interested in minifying our code so that the export stays relatively readable (at least while we're still exporting build assets) 

Additionally: 
- I needed to fix a few tests which erroneously used `this` in the wrong context (they were wrong anyway). 
- I removed the `--display-optimization-bailout` option since I have noticed that no one has actually been reading it and it's quite verbose. I would propose to create a monthly audit instead where we run `--display-optimization-bailout` and file any issues for modules that could be changed to be optimized. What do you think of that approach @dmose?
- I updated the `.npmrc` because iI noticed it was wrong
- The rest-spread issue is still not fixed. I think eslint might still be the blocker.


dmose:
Your display-optimiziation-bailout sounds like a fine plan; much better than the status quo.

dmose:
Out of curiosity, is the size or performance of the output from the new version meaningfully different than from the old version?

k88hudson:
@dmose I didn't look at the performance difference, the size is about -15kb (I posted a link to the diff above). So, doesn't seem like too much?

dmose:
Not huge, but I'm glad it's a win nonetheless.  :-)

.nvmrc
```diff
@@ -1 +1 @@
-7.*
```
dmose: Might be worth sending out an email about this before landing.  Probably not a huge deal, though.

PR #4233, opened 2018-07-06
Fix Bug 1471243: stop AS showing before first run overlay

AdamHillier:
These CSS changes will stop AS flashing before the startup overlay renders, but only if prerendering is disabled for the about:welcome load. With prerendering you'll still see the prerendered page flash before the overlay.

I have a patch to AboutRedirector and aboutNewTabService to disable prerendering on about:welcome; I'm not sure if there's a way to do that within AS that would be easier?

Kate and I talked about the possibility of having separate prerendered about:welcome files; [timing measurements](https://gist.github.com/AdamHillier/c32db37b66570f0e0d7f08a31b1849f6) suggest that these CSS changes should be sufficient by themselves.

### DIFF
--- a/content-src/components/Base/Base.jsx
+++ b/content-src/components/Base/Base.jsx
@@ -30,7 +30,7 @@ export class _Base extends React.PureComponent {
     this.sendNewTabRehydrated(App);
     addLocaleDataForReactIntl(locale);
     if (this.props.isFirstrun) {
-      global.document.body.classList.add("welcome");
+      global.document.body.classList.add("welcome", "hide-main");
     }
   }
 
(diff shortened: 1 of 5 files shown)


k88hudson:
It looks like the "Learn more about Firefox Accounts" link is no longer white for some reason:
image

content-src/components/Base/Base.jsx
```diff
@@ -128,6 +129,7 @@ export class BaseContent extends React.PureComponent {
     return (
       <div>
         <div className={outerClassName}>
+          {this.props.isFirstrun && <StartupOverlay />}
```
k88hudson: Was this change made for performance reasons?

AdamHillier: Hmm so I was originally moving it around to see if it would get it to render faster but in the end I don't think it makes a difference so I'll change it back

PR #4234, opened 2018-07-09
Fix Bug 1462114 - Set provider endpoint based on pref value

piatra:
This also fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1474244

https://github.com/piatra/activity-stream/blob/55005f9f8a61e5884cdd0152b17a7a567b1cc5dc/lib/ASRouterFeed.jsm#L37-L39
If we toggle the snippetsUrl pref then `enableOrDisableBasedOnPref` is called, the experiment pref is off and the Router gets disabled.

I've also moved the current fallback value for the `snippetsUrl` as a pref value in `ActivityStream.jsm` since the pref listener is set up to react for any provider pref change.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -123,11 +123,11 @@ class _ASRouter {
     this.initialized = false;
     this.messageChannel = null;
     this._storage = null;
+    this._prefs = null;
     this._resetInitialization();
     this._updateProviderEndpointUrl = this._updateProviderEndpointUrl.bind(this);
+    this._addASRouterPrefListener();
 
-    this._prefs = Services.prefs.getBranch(ASROUTER_PREF_BRANCH);
-    this._prefs.addObserver("", this);
     if (initialState.providers) {
       initialState.providers = initialState.providers.map(this._updateProviderEndpointUrl);
     }
@@ -140,6 +140,13 @@ class _ASRouter {
       ...initialState
     };
     this.onMessage = this.onMessage.bind(this);
+  }
+
+  _addASRouterPrefListener() {
+    if (this._prefs === null) {
+      this._prefs = Services.prefs.getBranch(ASROUTER_PREF_BRANCH);
+      this._prefs.addObserver("", this);
+    }
   }
 
   // Update provider endpoint and fetch new messages on pref change
@@ -230,6 +237,7 @@ class _ASRouter {
   async init(channel, storage) {
     this.messageChannel = channel;
     this.messageChannel.addMessageListener(INCOMING_MESSAGE_NAME, this.onMessage);
+    this._addASRouterPrefListener();
     await this.loadMessagesFromAllProviders();
     this._storage = storage;
 
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -15,8 +15,7 @@ ChromeUtils.defineModuleGetter(this, "ASRouterTargeting",
 const INCOMING_MESSAGE_NAME = "ASRouter:child-to-parent";
 const OUTGOING_MESSAGE_NAME = "ASRouter:parent-to-child";
 const ONE_HOUR_IN_MS = 60 * 60 * 1000;
-const ASROUTER_PREF_BRANCH = "browser.newtabpage.activity-stream.asrouter.";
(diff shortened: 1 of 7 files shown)


lib/ASRouter.jsm
```diff
@@ -125,7 +123,15 @@ class _ASRouter {
     this.initialized = false;
     this.messageChannel = null;
     this._storage = null;
+    this._prefs = null;
     this._resetInitialization();
+    this._updateProviderEndpointUrl = this._updateProviderEndpointUrl.bind(this);
+    this._addASRouterPrefListener();
```
k88hudson: You're calling this twice, once in the constructor and once in `init()`; i'm assuming you just meant to call it in init?

lib/ASRouterFeed.jsm
```diff
@@ -33,7 +33,10 @@ class ASRouterFeed {
    * (asrouterExperimentEnabled) and enable or disable ASRouter based on
    * its value.
    */
-  enableOrDisableBasedOnPref() {
+  enableOrDisableBasedOnPref(prefChanged) {
+    if (!["asrouterOnboardingCohort", "asrouterExperimentEnabled"].includes(prefChanged)) {
```
k88hudson: This doesn't cover the `INIT`case, this will only trigger from a pref change

lib/ASRouter.jsm
```diff
@@ -136,6 +142,38 @@ class _ASRouter {
...
+    await this.setState(prevState => {
+      const providers = [...prevState.providers];
+      this._updateProviderEndpointUrl(providers.find(p => p.endpointPref === aPrefName));
+      return {providers};
+    });
+
+    await this.loadMessagesFromAllProviders();
+  }
+
+  _updateProviderEndpointUrl(provider) {
```
k88hudson: Because this is running in the constructor, I think this is running into some race conditions with our AS-specific prefs getting added later (i.e. it basically returns the default `""` and never gets updated)

In order to make this less bug prone, I would maybe checking the pref / setting `provider.url` at the point at which messages are fetched (i.e. in `_remoteLoader`). Fetches don't happen so often so I wouldn't worry about the overhead. You can still reset `.lastUpdated` and reload messages in the observer.

lib/ASRouter.jsm
```diff
@@ -136,6 +142,38 @@ class _ASRouter {
     this.onMessage = this.onMessage.bind(this);
   }
 
+  _addASRouterPrefListener() {
+    if (this._prefs === null) {
+      this._prefs = Services.prefs.getBranch(ASROUTER_PREF_BRANCH);
```
k88hudson: Personally, I wouldn't use a pref branch here so we don't limit ourselves to what kinds of prefs can be used for endpoints; just use Services.prefs directly

lib/ASRouter.jsm
piatra: Wanted to add some `broadcast` method to update opened tabs when pref is changed but getting the next message is [tied to the target](https://github.com/mozilla/activity-stream/blob/c97d8464ddefc9109c06b0137f7dea48434b7419/lib/ASRouter.jsm#L323) and doesn't make it very straight forward if you want to broadcast the same message to all message channel ports.
Does the feature make sense? Should it be a followup? 
Only noticed it because ASRouter admin reacts to pref changes and updates the message provider header.

test/unit/asrouter/ASRouterFeed.test.js
```diff
@@ -73,20 +73,31 @@ describe("ASRouterFeed", () => {
...
   });
   describe("#onAction: PREF_CHANGE", () => {
+    it("should return early if the pref changed does not enable/disable the router", async () => {
+      // Router starts initialized
+      await Router.init(new FakeRemotePageManager(), storage);
+      sinon.stub(Router, "uninit");
+      prefs[EXPERIMENT_PREF] = false;
+
+      // call .onAction with INIT
+      feed.onAction({type: at.PREF_CHANGED, data: {name: "someOtherPref"}});
```
k88hudson: 👍

k88hudson:
Please split out the preferences fix in ASRouterFeed in case we want to uplift it. Thanks!

PR #4240, opened 2018-07-13
Fix Bug 1471514 - Add top-of-funnel metrics for fxa.

ericawright:
This is almost identical to #4218, except it now uses `identity.fxaccounts.remote.firstrun`  instead of `identity.fxaccounts.remote.root` as a pref. This new pref was created so that we can turn it off during tests without disrupting anything else that may depend on `identity.fxaccounts.remote.root`.

### DIFF
--- a/content-src/components/StartupOverlay/StartupOverlay.jsx
+++ b/content-src/components/StartupOverlay/StartupOverlay.jsx
@@ -1,5 +1,5 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 import {FormattedMessage, injectIntl} from "react-intl";
-import {actionCreators as ac} from "common/Actions.jsm";
 import {connect} from "react-redux";
 import React from "react";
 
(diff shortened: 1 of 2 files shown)


content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -15,8 +15,29 @@ export class _StartupOverlay extends React.PureComponent {
 
     this.state = {
       emailInput: "",
-      overlayRemoved: false
+      overlayRemoved: false,
+      flowId: "",
+      flowBeginTime: ""
```
sarracini: this should probably be a 0, since it's a timestamp. let's keep the "types" consistent

lib/PrefsFeed.jsm
```diff
@@ -101,6 +101,9 @@ this.PrefsFeed = class PrefsFeed {
     values.isPrivateBrowsingEnabled = PrivateBrowsingUtils.enabled;
     values.platform = AppConstants.platform;
 
+    // Get the firefox accounts url to send firstrun metrics to.
+    values.fxa_endpoint = Services.prefs.getStringPref("identity.fxaccounts.remote.firstrun", "https://accounts.firefox.com");
```
sarracini: just change the pref name to match the pref name in the m-c patch

PR #4241, opened 2018-07-13
Bug 1476309 - use RemoteSettings for Tippy Top

rlr:
@ncloudioj to test it out, you need to change the prefs for RS. I just run the following in the browser console:

```
const SERVER_STAGE = "https://settings.stage.mozaws.net/v1";
const HASH_STAGE = "DB:74:CE:58:E4:F9:D0:9E:E0:42:36:BE:6C:C5:C4:F6:6A:E7:74:7D:C0:21:42:7A:03:BC:2F:57:0C:8B:9B:90";
Services.prefs.setCharPref("services.settings.server", SERVER_STAGE);
Services.prefs.setCharPref("security.content.signature.root_hash", HASH_STAGE);
```

### DIFF
--- a/lib/FaviconFeed.jsm
+++ b/lib/FaviconFeed.jsm
@@ -5,11 +5,9 @@
 
 ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
 
-XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
-
 const {actionTypes: at} = ChromeUtils.import("resource://activity-stream/common/Actions.jsm", {});
-const {PersistentCache} = ChromeUtils.import("resource://activity-stream/lib/PersistentCache.jsm", {});
 const {getDomain} = ChromeUtils.import("resource://activity-stream/lib/TippyTopProvider.jsm", {});
+const {RemoteSettings} = ChromeUtils.import("resource://services-settings/remote-settings.js", {});
 
 ChromeUtils.defineModuleGetter(this, "PlacesUtils",
   "resource://gre/modules/PlacesUtils.jsm");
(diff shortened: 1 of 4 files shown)


ncloudioj:
LGTM!

@leplatrem, could you also take a look at this please (the client side change of [bug 1437671](https://bugzilla.mozilla.org/show_bug.cgi?id=1437671))?

The key change of this patch is that we use RemoteSettings to replace the old JSON Tippytop manifest (manually sync'd to its origin on S3).

Note that since the Tippytop functionality is not sensitive to the data latency (add/update/delete) in Activity Stream, we completely rely on the default syncing strategy of RS. We've also removed the memory cache for the Tippytop manifest, and implemented the lookup directly on top of `RemoteSettings.get()`. That at least considerably simplified the code : ).

Would love to hear your thoughts on those changes.

lib/FaviconFeed.jsm
```diff
@@ -246,21 +152,27 @@ this.FaviconFeed = class FaviconFeed {
     }
   }
 
+  async getSite(domain) {
+    const sites = await this.tippyTop.get({filters: {"domain": domain}});
```
leplatrem: nitpick: could also be shortened as `this.tippyTop.get({filters: {domain}});`

lib/FaviconFeed.jsm
```diff
@@ -223,10 +129,10 @@ this.FaviconFeed = class FaviconFeed {
       return;
     }
 
-    const sitesByDomain = await this.getSitesByDomain();
-    const domain = getDomain(url);
-    if (domain in sitesByDomain) {
-      let iconUri = Services.io.newURI(sitesByDomain[domain].image_url);
+    const site = await this.getSite(getDomain(url));
+
+    if (site) {
```
leplatrem: nitpick: since the rest of the function is useless when site is null, maybe:
```js
if (!site) {
  return;
}
```

test/unit/lib/FaviconFeed.test.js
```diff
@@ -258,35 +103,37 @@ describe("FaviconFeed", () => {
...
+      const get = () => Promise.resolve([]);
+      feed._tippyTop = {get};
+      const site = await feed.getSite("example.com");
+      assert.isNull(site);
+    });
+    it("should lazy init _tippyTop", async () => {
+      assert.isUndefined(feed._tippyTop);
+      await feed.getSite("example.com");
+      assert.ok(feed._tippyTop);
     });
```
leplatrem: An alternative to this would be to store records in the local db:
```js
const collection = await RemoteSettings("tippytop").openCollection();
await collection.create({
  domain,
  image_url: `${url}/icon.png`
});
```

But this may require to setup IndexedDB in your tests suite, which may not be obvious

PR #4242, opened 2018-07-13
Backport chanages from Bug 1474410

sarracini:
Backport changes which move activity stream init/uninit from boostrap.js to AboutNewTab.jsm in m-c. Also while here, deleting the now unused `uninstall` function in ActivityStream.jsm.

This needs to wait until changes in bug 1474410 land in mozilla central before landing


sarracini:
This is now unblocked: https://hg.mozilla.org/mozilla-central/rev/f98112e49582

PR #4250, opened 2018-07-17
Bug 1476381 - Fix outstanding sass lint warnings

k88hudson:
Woohoo, no more warnings!

To test:
- run `npm run lint:sasslint` and ensure no warnings are produced.
- please check `about:welcome` and dark theme since almost all changes involve those files.

Notes:
- I don't think there's actually a way to allow for more than 1 `class-name-format` convention, so I just removed the rule. Personally I don't particularly think it matters if people use camel case or hyphens anyway
- these changes should not result in any significant changes in the output and definitely not in the design

### DIFF
--- a/content-src/styles/_variables.scss
+++ b/content-src/styles/_variables.scss
@@ -53,6 +53,11 @@ $download-icon-fill: #12BC00;
 $pocket-icon-fill: #D70022;
 $email-input-focus: rgba($blue-50, 0.3);
 $email-input-invalid: rgba($red-60, 0.3);
+$aw-extra-blue-1: #004EC2;
+$aw-extra-blue-2: #0080FF;
+$aw-extra-blue-3: #00C7FF;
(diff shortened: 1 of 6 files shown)


content-src/styles/_variables.scss
```diff
@@ -53,6 +53,11 @@ $download-icon-fill: #12BC00;
 $pocket-icon-fill: #D70022;
 $email-input-focus: rgba($blue-50, 0.3);
 $email-input-invalid: rgba($red-60, 0.3);
+$aw-extra-blue-1: #004EC2;
+$aw-blue-2: #0080FF;
+$aw-3: #00C7FF;
```
sarracini: should be `$aw-extra-blue-2` and `$aw-extra-blue-3` as used below

dmose:
@k88hudson would it be more future-proof to set [max-warnings](https://github.com/sasstools/sass-lint/blob/master/docs/options/max-warnings.md) to `0` rather than calling out all the rules individually?

k88hudson:
@dmose I suppose that would be better, yes!

PR #4251, opened 2018-07-17
gSnippetsMap parity in ASRouter

k88hudson:
Adds additional targeting properties to achieve parity with existing snippets targeting fields.

To test:
- Please ensure mochi tests pass by running `./mach test browser/extensions/activity-stream/test/functional/mochitest/browser_asrouter_targeting.js`. See tests and docs for examples and JEXL

### DIFF
--- a/test/functional/mochitest/browser_asrouter_targeting.js
+++ b/test/functional/mochitest/browser_asrouter_targeting.js
@@ -109,6 +109,15 @@ add_task(async function checkisDefaultBrowser() {
     "should select correct item by isDefaultBrowser");
 });
 
+add_task(async function checkdevToolsOpenedCount() {
+  await pushPrefs(["devtools.selfxss.count", 5]);
+  is(ASRouterTargeting.Environment.devToolsOpenedCount, 5,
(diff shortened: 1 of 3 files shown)


test/functional/mochitest/browser_asrouter_targeting.js
```diff
@@ -71,6 +73,51 @@ add_task(async function checkhasFxAccount() {
...
+
+  const message2 = {id: "foo", targeting: `searchEngines[${Services.search.getVisibleEngines()[0].identifier} in .installed]`};
+  is(await ASRouterTargeting.findMatchingMessage([message2], {}), message2,
+    "should select correct item by searchEngines.installed");
+});
+
+add_task(async function checkisDefaultBrowser() {
+  const expected = ShellService.isDefaultBrowser();
+  const result = ASRouterTargeting.Environment.isDefaultBrowser;
+  is(typeof result, "boolean",
```
piatra: If checking for a default browser fails the result is null is it expected to always pass?

k88hudson: I guess I'm assuming it's not going to fail in the test environment, although perhaps that's not a 100% solid assumption

PR #4254, opened 2018-07-19
Fix Bug 1474372: Run mochitests on taskcluster

piatra:
screen shot 2018-07-19 at 16 44 12
Not sure where that request actually goes. It's needed so that taskcluster actually registers our repository.

Here are the logs of the mochitests running on my fork of AS https://tools.taskcluster.net/groups/FZ0dHWLVTq6q0sbQj2JEig/tasks/FZ0dHWLVTq6q0sbQj2JEig/runs/0/logs/public%2Flogs%2Flive.log#L1753

Some more info about the task steps include:
* pulling the image
* updating m-c 
* pulling the PR we want to test
* building AS and m-c (artifact)
* running the tests

Took 17 minutes. To achieve this the docker image `piatra/asmochitests` (specified in .taskcluster.yml) is built to include m-c repo + an artifact built before publishing in an attempt to speed things up.
As a follow up I am planning to set up a cron job on my machine that will rebuild + publish a new version of the image weekly to keep up with code changes. This does not mean the m-c is not updated (there is a `hg pull` step when running the tests) but it helps to have the image as up to date as possible.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -8,6 +8,7 @@ tasks:
           - pull_request.opened
           - pull_request.reopened
           - pull_request.synchronize
+          - push
     payload:
       maxRunTime: 7200
       image: piatra/asmochitests
@@ -17,7 +18,7 @@ tasks:
         - '-c'
         - >-
           git clone {{event.head.repo.url}} /activity-stream && cd /activity-stream &&
-          git checkout {{event.head.sha}} && npm install . && bash ./mochitest.sh
+          git checkout {{event.head.sha}} && bash ./mochitest.sh
     metadata:
       name: activitystream
       description: run mochitests for PRs
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,0 +1,30 @@
+version: 0
+tasks:
+  - provisionerId: '{{ taskcluster.docker.provisionerId }}'
+    workerType: '{{ taskcluster.docker.workerType }}'
+    extra:
+      github:
+        events:
+          - pull_request.opened
+          - pull_request.reopened
+          - pull_request.synchronize
+    payload:
+      maxRunTime: 7200
+      image: piatra/asmochitests
+      command:
+        - /bin/bash
+        - '--login'
+        - '-c'
+        - >-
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -90,7 +90,7 @@
     "mc_dir": "../mozilla-central"
   },
   "scripts": {
-    "mochitest": "(cd $npm_package_config_mc_dir && ./mach mochitest browser/extensions/activity-stream/test/functional/mochitest )",
+    "mochitest": "(cd $npm_package_config_mc_dir && ./mach mochitest browser/extensions/activity-stream/test/functional/mochitest --headless )",
```
k88hudson: can you edit the `yamscripts` file instead? Also, would it be worth it / possible to to run `browser/components/newtab/`  and maybe also `./mach test all_files_referenced | grep activity-stream`?

.taskcluster.yml
```diff
@@ -0,0 +1,27 @@
...
+    payload:
+      maxRunTime: 7200
+      image: piatra/asmochitests
+      command:
+        - /bin/bash
+        - '--login'
+        - '-c'
+        - >-
+          git clone {{event.head.repo.url}} /activity-stream && cd /activity-stream &&
+          git checkout {{event.head.sha}} && npm install . && bash ./mochitest.sh
```
k88hudson: why not add `npm install` to the bash script since that has the rest of your test-related setup?

package.json
piatra: oops true.

piatra:
`./mach test browser_parsable_css` seems to fail in headless mode, I have to look into using xvfb similar to how we run our tests on travis. `./mach test all_files_referenced` seems to fail whatever I try. Initially assumed it might be related to artifact build but I get the same results after doing a full build.

Mardak:
I believe you need to `./mach package` before all_files_referenced

piatra:
Followed instructions from https://searchfox.org/mozilla-central/rev/88199de427d3c5762b7f3c2a4860c10734abd867/browser/base/content/test/static/browser_all_files_referenced.js#5-6

piatra:
Tests are now running: 🎉 
grafik

PR #4255, opened 2018-07-22
Fix Bug 1471351: Add openURL trigger to ASRouter

AdamHillier:
In order to support an `openURL` trigger, triggers become (instead of being just strings) objects with an `id` and an optional `param` (in this case the url host name).

`ASRouterTriggerListeners.jsm` has a map of singleton listeners which are initialised or not depending on whether ASRouter has messages that have that particular trigger. Every time ASRouter updates its messages, it checks their triggers to do this un/initialisation of the trigger listeners.

At the moment the only listener is for `openURL`, which listens for webpages being opened with specific host names, but it's set up in a way that allows listeners for new triggers to be added very easily.

You can manually test that the patch works by adding a test message to `OnboardingMessageProvider` such as:
```js
{
  id: "TEST",
  trigger: {
    id: "openURL",
    params: ["www.amazon.com", "www.amazon.co.uk"]
  }
}
```
After putting a breakpoint in the `TRIGGER` case of ASRouter#onMessage in the browser debugger, and then navigating to www.amazon.com or www.amazon.co.uk you should be able to inspect the trigger message that gets sent.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -281,9 +281,8 @@ class _ASRouter {
     this.messageChannel.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {type: "ADMIN_SET_STATE", data: state});
   }
 
-  async _findMessage(messages, target, data = {}) {
+  async _findMessage(messages, target, trigger) {
     let message;
-    const {trigger} = data;
     const {impressions} = this.state;
     if (trigger) {
       // Find a message that matches the targeting context as well as the trigger context
@@ -300,7 +299,7 @@ class _ASRouter {
     return bundle.sort((a, b) => a.order - b.order);
   }
 
-  async _getBundledMessages(originalMessage, target, data, force = false) {
+  async _getBundledMessages(originalMessage, target, trigger, force = false) {
     let result = [{content: originalMessage.content, id: originalMessage.id, order: originalMessage.order || 0}];
 
     // First, find all messages of same template. These are potential matching targeting candidates
@@ -319,7 +318,7 @@ class _ASRouter {
     } else {
       while (bundledMessagesOfSameTemplate.length) {
         // Find a message that matches the targeting context - or break if there are no matching messages
-        const message = await this._findMessage(bundledMessagesOfSameTemplate, target, data);
+        const message = await this._findMessage(bundledMessagesOfSameTemplate, target, trigger);
         if (!message) {
           /* istanbul ignore next */ // Code coverage in mochitests
           break;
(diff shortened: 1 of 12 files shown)


lib/ASRouterTriggerListeners.jsm
```diff
@@ -0,0 +1,142 @@
...
+      }
+    },
+
+    observe(subject, topic, data) {
+      let win;
+      let onLoad;
+
+      switch (topic) {
+        case "domwindowopened":
+          win = subject.QueryInterface(Ci.nsIDOMWindow);
```
sarracini: Actually, as of recently, the subject is the window! no need to QueryInterface anymore

lib/ASRouterTriggerListeners.jsm
```diff
@@ -0,0 +1,142 @@
...
+            break;
+          }
+          onLoad = () => {
+            win.removeEventListener("load", onLoad);
+            // Ignore non-browser windows.
+            if (win.document.documentElement.getAttribute("windowtype") === "navigator:browser") {
+              win.gBrowser.addTabsProgressListener(this);
+            }
+          };
+          win.addEventListener("load", onLoad);
```
sarracini: You can use the options `{once: true}` on the event listener so that you don't have to do the "add an event listener and remove it after firing once" dance:
` win.addEventListener("load", onLoad, {once: true});`

lib/ASRouter.jsm
```diff
@@ -223,6 +226,26 @@ class _ASRouter {
           newState.messages = [...newState.messages, ...messages];
         }
       }
+
+      // Some triggers have trigger listeners that we might need to initialise.
+      for (const [triggerID, listener] of ASRouterTriggerListeners) {
```
sarracini: Does it make more sense to do this the other way around... iterate over `newState.messages` and only add the listeners to the messages that have triggers? I'm assuming we don't want to add listeners if we don't even plan on showing that message. That way you wouldn't have to filter out the messages that have triggers

lib/ASRouter.jsm
```diff
@@ -223,6 +226,26 @@ class _ASRouter {
...
+        const triggerMessages = newState.messages
+          .filter(m => m.trigger && m.trigger.id === triggerID);
+
+        if (triggerMessages.length) {
+          // If some messages have this trigger, initialise the trigger listener
+          const params = new Set();
+          for (const message of triggerMessages) {
+            ((message.trigger.params) || []).forEach(p => params.add(p));
```
sarracini: I think you can literally just make a `new Set()` directly from params if that's what you're looking to do... no need to manually add each item: `new Set(params || [])`

sarracini: Clarified with Adam about the use of Set and multiple parameters for different messages with the same trigger. As discussed over Slack we'll take the following approach:
> iterate over the messages in ASRouter and just blindly send the params and call init to the listener, and then internally the listener would know that it already has a running list of hosts, and to just add these new params to the set it already has

PR #4257, opened 2018-07-23
Fix Bug 1477631: Allow setting whitelist hosts via prefs

piatra:
With this PR you can add additional whitelist hosts via the pref `browser.newtab.activity-stream.asrouter.whitelistHosts`. They are added in addition to the already existing list.

Because this is used for development only the pref **does not come with a default value** (does not exist in about:config), it will have to be added manually. Example value `["gist.github.com", "gist.githubusercontent.com"]`
You also need to restart the Router by flipping `browser.newtabpage.activity-stream.asrouterExperimentEnabled` which I think is acceptable since it's used for development.
Every host added will be of type preview: when adding an endpoint its`updateCycle` will be 0 so that it refreshes on every page load.

If you are trying to add an endpoint that is not in the list you will get an error
screen shot 2018-07-23 at 16 30 10

There's a confirmation message on success
screen shot 2018-07-23 at 16 30 16

Then you will be able to add new endpoints which will show up in the ASRouter admin
screen shot 2018-07-23 at 16 31 12

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -484,7 +484,7 @@ class _ASRouter {
 
   _loadSnippetsWhitelistHosts() {
     let additionalHosts = [];
-    const whitelistPrefValue = Services.prefs.getStringPref(SNIPPETS_ENDPOINT_WHITELIST);
+    const whitelistPrefValue = Services.prefs.getStringPref(SNIPPETS_ENDPOINT_WHITELIST, "");
     try {
       additionalHosts = JSON.parse(whitelistPrefValue);
     } catch (e) {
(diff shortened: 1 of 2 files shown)


lib/ASRouter.jsm
```diff
@@ -468,18 +470,37 @@ class _ASRouter {
...
     } catch (e) {
       return false;
     }
   }
 
+  _loadSnippetsWhitelistHosts() {
+    let additionalHosts = [];
+    try {
+      additionalHosts = JSON.parse(Services.prefs.getStringPref(SNIPPETS_ENDPOINT_WHITELIST));
+    } catch (e) {}
```
k88hudson: it might be useful to surface a parsing error if someone is trying to debug this?

piatra:
cc @glogiotatidis

PR #4260, opened 2018-07-25
Fix Bug 1477766 - Display Simplified Onboarding Tour Cards in specific order

sarracini:
Adds an order property to the message, and sorts on that order. Defaults to 0 if no order provided - meaning this bundle doesn't care about this order

### DIFF
--- a/content-src/asrouter/schemas/provider-response.schema.json
+++ b/content-src/asrouter/schemas/provider-response.schema.json
@@ -21,6 +21,15 @@
             "description": "An id matching an existing Activity Stream Router template",
             "enum": ["simple_snippet"]
           },
+          "bundled": {
+            "type": "integer",
+            "description": "The number of messages of the same template this one should be shown with (optional)"
+          },
+          "order": {
+            "type": "integer",
+            "minimum": 0,
+            "description": "If bundled with other messages of the same template, which order should this one be placed in? (optional - defaults to 0)"
+          },
           "content": {
             "type": "object",
             "description": "An object containing all variables/props to be rendered in the template. See individual template schemas for details."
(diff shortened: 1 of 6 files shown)


lib/OnboardingMessageProvider.jsm
```diff
@@ -50,11 +53,12 @@ const ONBOARDING_MESSAGES = [
     id: "ONBOARDING_4",
     template: "onboarding",
     bundled: 3,
+    order: 1,
     content: {
-      title: "Extensions",
-      text: "Make browsing faster, smarter, or safer with browser apps. Protect passwords, find deals, download videos, and much more. You can even block annoying ads with extensions like Ghostery.",
+      title: "Block Ads with Ghostery",
```
sarracini: Some copy changes to the Ghostery extension that were requested prior to 62

content-src/asrouter/schemas/provider-response.schema.json
```diff
@@ -21,6 +21,14 @@
             "description": "An id matching an existing Activity Stream Router template",
             "enum": ["simple_snippet"]
           },
+          "bundled": {
+            "type": "integer",
+            "description": "The number of messages of the same template this one should be shown with (optional)"
+          },
+          "order": {
```
k88hudson: can you add to the docs as well?

lib/ASRouter.jsm
```diff
@@ -283,8 +283,12 @@ class _ASRouter {
     return message;
   }
 
+  _orderBundle(bundle) {
+    return bundle.sort((a, b) => a.order > b.order);
```
k88hudson: This should be `a.order - b.order`

content-src/asrouter/schemas/provider-response.schema.json
```diff
@@ -21,6 +21,14 @@
...
+            "type": "integer",
```
k88hudson: Should this maybe have `"minimum": 0` as well or are we allowing negative values?

PR #4261, opened 2018-07-25
Fix Bug 1476713 - Add action that triggers addon install from URL.

piatra:
To test you need a signed addon, anything from AMO works like: https://addons.mozilla.org/firefox/downloads/file/684007/emoji_cheatsheet-2.0.1-an+fx.xpi


lib/ASRouter.jsm
```diff
@@ -553,6 +562,9 @@ class _ASRouter {
       case "IMPRESSION":
         this.addImpression(action.data);
         break;
+      case "INSTALL_ADDON_FROM_URL":
```
k88hudson: you need to add this to `ASRouterActions` in Actions.jsm (see `ASRouterUtils.executeAction` in asrouter-content). This is for security reasons (i.e. so you can't just specify any action in a message and have it work)

lib/ASRouter.jsm
```diff
@@ -104,6 +105,14 @@ const MessageLoaderUtils = {
         .map(msg => ({...msg, provider: provider.id}));
     const lastUpdated = Date.now();
     return {messages, lastUpdated};
+  },
+
+  async installAddonFromURL(browser, url) {
+    const aUri = Services.io.newURI(url);
```
sarracini: Services.io.newURI might throw if the url is invalid, so let's wrap this in a try/catch to hand that, and maybe just send something to telemetry or return out of the function

test/unit/asrouter/MessageLoaderUtils.test.js
```diff
@@ -135,4 +136,34 @@ describe("MessageLoaderUtils", () => {
...
+      globals.restore();
+    });
+    it("should call installAddonFromWebpage with correct arguments", async () => {
+      getInstallStub.resolves(null);
+      installAddonStub.resolves(null);
+
+      await MessageLoaderUtils.installAddonFromURL({}, "foo.com");
+
+      assert.calledOnce(getInstallStub);
+      assert.calledOnce(installAddonStub);
```
sarracini: you're not actually checking that these were called with the correct arguments, despite what the test is named ;)

lib/ASRouter.jsm
```diff
@@ -584,6 +593,9 @@ class _ASRouter {
       case "IMPRESSION":
         this.addImpression(action.data);
         break;
+      case "INSTALL_ADDON_FROM_URL":
```
sarracini: let's access it from `ra.INSTALL_ADDON_FROM_URL`

lib/ASRouter.jsm
piatra: Wrapping this in a try catch and will follow up with telemetry in another bug. Currently the router doesn't have access to the telemetry feed.

test/unit/asrouter/MessageLoaderUtils.test.js
piatra: I will update the test description. The function is only calling apis and passing results without any transformations. Stubbing all the various functions in order to assert doesn't seem very useful.

PR #4262, opened 2018-07-25
Fix Bug 1478218: add Places cleanup to browser_asrouter_targeting.js

AdamHillier:


lib/ASRouterTargeting.jsm
```diff
@@ -215,4 +220,6 @@ this.ASRouterTargeting = {
   }
 };
 
-this.EXPORTED_SYMBOLS = ["ASRouterTargeting", "removeRandomItemFromArray"];
```
AdamHillier: `removeRandomItemFromArray` was being exported but then not used anywhere, so I removed it

PR #4263, opened 2018-07-25
Fix Bug 1475921 - remove text-transform usage on localized strings (section headers)

rlr:
There are two more instances of `text-transform: uppercase;` but they are on strings that aren't localized (domain letter fallback and domain names). We can take care of those with `/* eslint-disable */` or possibly `.toUpperCase()` when we add the eslint rule in https://bugzilla.mozilla.org/show_bug.cgi?id=1476450

r? @k88hudson

### DIFF
--- a/content-src/components/CollapsibleSection/_CollapsibleSection.scss
+++ b/content-src/components/CollapsibleSection/_CollapsibleSection.scss
@@ -8,7 +8,6 @@
     font-size: $section-title-font-size;
     font-weight: bold;
     margin: 0;
-    text-transform: uppercase;
 
     span {
       color: var(--newtab-section-header-text-color);
--- a/content-src/components/TopSites/_TopSites.scss
+++ b/content-src/components/TopSites/_TopSites.scss
@@ -299,7 +299,6 @@ $half-base-gutter: $base-gutter / 2;
     }
 
     .section-title {
-      text-transform: none;
       font-size: 16px;
       margin: 0 0 16px;
     }


locales/en-US/strings.properties
```diff
@@ -1,10 +1,13 @@
 newtab_page_title=New Tab
 
 header_top_sites=Top Sites
+header_top_sites_upper=TOP SITES
```
Mardak: @flodolo any preference on the name? Not sure if it needs a separate localization note or people will know to try to make it uppercase if appropriate?

flodolo: String ID make sense, I think we can skip the localization note.

common/PrerenderData.jsm
```diff
@@ -89,14 +89,14 @@ this.PrerenderData = new _PrerenderData({
       icon: "pocket",
       id: "topstories",
       order: 1,
-      title: {id: "header_recommended_by", values: {provider: "Pocket"}}
+      title: {id: "header_recommended_by_upper", values: {provider: "POCKET"}}
```
k88hudson: You're would also need to change this in the prefs config of `Activity-Stream.jsm` (see `feeds.section.topstories.options`) but I'm not sure that's the right thing to do given that in some languages the translation won't be all uppercase?

lib/SectionsManager.jsm
```diff
@@ -31,7 +31,7 @@ const BUILT_IN_SECTIONS = {
     shouldHidePref: options.hidden,
     eventSource: "TOP_STORIES",
     icon: options.provider_icon,
-    title: {id: "header_recommended_by", values: {provider: options.provider_name}},
+    title: {id: "header_recommended_by_upper", values: {provider: options.provider_name && options.provider_name.toUpperCase()}},
```
rlr: @k88hudson so here I added `toUpperCase()`.... 🤔 💭

common/PrerenderData.jsm
rlr: possibly just hardcoding pocket into the string is a reasonable option too since there will be no other providers?

Thoughts @Mardak ?

rlr: For reference:

screen shot 2018-07-25 at 5 41 55 pm

rlr:
@k88hudson good news! @bryanbell said to change it to title case and see if anybody even notices 😃 

before:
screen shot 2018-07-26 at 4 40 51 pm

after:
screen shot 2018-07-26 at 4 39 23 pm

PR #4264, opened 2018-07-25
Bug 1466888 - Empty scrollbar on blank page

leadrig:
Well, guys, I ve found a problem:
If you open New Tab, and open it's source code, you will see this line:
resource://activity-stream/css/activity-stream.css
In that file you will see this line under body element properories:
overflow-y: scroll;
To prevent the behavior, noticed in this bug, you should change it to auto.

### DIFF
--- a/content-src/styles/_activity-stream.scss
+++ b/content-src/styles/_activity-stream.scss
@@ -21,7 +21,6 @@ body {
   background-color: var(--newtab-background-color);
   font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Ubuntu', 'Helvetica Neue', sans-serif;
   font-size: 16px;
-  overflow-y: auto;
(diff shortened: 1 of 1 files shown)


piatra:
Afaik this was a limitation of our prerendering where if we didn't keep the scrollbar around it would slightly move the content around when loading.
Now with 1553375 closed could we merge this fix? Were there other reasons we had the scrollbar?

piatra:
cc @Mardak

Mardak:
Even without prerendering, the page width will change as content loads. But we can still take this given our better understanding of our performance target of settled time (which scrollbar appearing doesn't affect right now).

scroll auto

Notice the page shifts left as it centers before the top sites or article images load.

Mardak:
I rebased your patch on latest master

PR #4265, opened 2018-07-26
Backport Bug 1474414 - Move most of browser/extensions/activity-stream to browser/components/newtab

sarracini:
GOODBYE SYSTEM-ADDON LIFE 🎉


aboutNewTabService.js
```diff
@@ -0,0 +1,252 @@
...
+
+ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
+ChromeUtils.import("resource://gre/modules/Services.jsm");
+ChromeUtils.import("resource://gre/modules/AppConstants.jsm");
+
+ChromeUtils.defineModuleGetter(this, "AboutNewTab",
+                               "resource:///modules/AboutNewTab.jsm");
+
+const TOPIC_APP_QUIT = "quit-application-granted";
+const TOPIC_LOCALES_CHANGE = "intl:app-locales-changed";
```
Mardak: This file seems to be outdated? In particular missing some about:welcome and @imjching TOPIC_CONTENT_DOCUMENT_INTERACTIVE stuff

sarracini: i need to stop using dxr and make the switch to searchfox. will update

sarracini: or maybe i'll just pull latest mc and it from there

mochitest.sh
```diff
@@ -14,8 +14,8 @@ cd /activity-stream && npm install . && npm run buildmc
 # Build latest m-c with Activity Stream changes
 cd /mozilla-central && ./mach build \
   && ./mach test browser_parsable_css \
-  && ./mach lint -l eslint -l codespell browser/extensions/activity-stream \
-  && ./mach test browser/extensions/activity-stream --headless \
+  && ./mach lint -l eslint -l codespell browser/components/newtab \
+  && ./mach test browser/components/newtab --headless \
   && ./mach test browser/components/newtab/tests/browser --headless \
   && ./mach test browser/components/newtab/tests/xpcshell \
```
Mardak: It's no longer "tests" but I think the newly added "test browser/components/newtab" will cover both directories anyway? @piatra ?

piatra: That's right, it covers both directories but the `--headless` argument is invalid for xpcshell and the test exits.
Since there is a virtual display
https://github.com/mozilla/activity-stream/blob/1136adfa70bda22c48ecbd155fdd09334a960611/mochitest.sh#L5-L6
You could simply drop the `--headless` argument from all.

Mardak: If there's some benefit of it running headless, then we should probably just remove the `test browser/compnoents/newtab` and have the 2 lines for mochitest headless and xpcshell.

Mardak: Looks like we have green taskcluster in #4267 where I went with the 2 lines for mochitest headless and xpcshell. We'll need #4268 to get things passing

PR #4267, opened 2018-07-27
Component-ify

Mardak:
Like #4265 but mainly to trigger task cluster. This has conflicts as it's rebased to the last exported revision not latest master. It also includes #4268


Mardak:
#4265 merged

PR #4272, opened 2018-07-30
Fix Bug 1479413 - Remove preview provider and message after it has been displayed

piatra:
And add documentation about how preview works.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -226,7 +226,6 @@ class _ASRouter {
           newState.messages = [...newState.messages, ...messages];
         }
       }
-<<<<<<< HEAD
 
       // Some messages have triggers that require us to initalise trigger listeners
       const unseenListeners = new Set(ASRouterTriggerListeners.keys());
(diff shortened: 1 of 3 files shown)


lib/ASRouter.jsm
```diff
@@ -435,7 +436,12 @@ class _ASRouter {
       message = await this._findMessage(msgs, target, data);
     }
 
-    await this.setState({lastMessageId: message ? message.id : null});
+    if (previewMsgs.length) {
+      // We don't want to cache preview messages, remove them after we selected the message to show
+      await this.setState(state => this._removePreviewMessage(state, {lastMessageId: message.id}));
```
sarracini: are we sure there will be a `message` ? the else case seems to guard against message not existing, should we do that here too?

lib/ASRouter.jsm
```diff
@@ -511,10 +517,21 @@ class _ASRouter {
     }, {...DEFAULT_WHITELIST_HOSTS});
   }
 
+  _removePreviewMessage(state, lastMessageId) {
+    return {
+      messages: state.messages.filter(m => m.id !== lastMessageId.lastMessageId),
```
sarracini: is `lastMessageId` an object with a `lastMessageId` property? that seems weird.. can we re-name this to something less weird?

lib/ASRouter.jsm
piatra: if `previewMsgs.length` then we do have a message. We pick the first element of the array to dispatch.

PR #4276, opened 2018-08-01
Fix Bug 1479458 - Spocs updating when pref changes from a shield study

ScottDowne:
To test:

1. Set the pref "browser.newtabpage.activity-stream.feeds.section.topstories.options" to default.

2. load about:home take not of the third rec. This ensures that we have a preloaded tab generated with default recs. You can keep this tab open, or close it, both are viable and shouldn't matter.

3. Update "browser.newtabpage.activity-stream.feeds.section.topstories.options" pref to:
```
{"api_key_pref":"extensions.pocket.oAuthConsumerKey","hidden":false,"provider_icon":"pocket","provider_name":"Pocket","read_more_endpoint":"https://getpocket.com/explore/trending?src=fx_new_tab","stories_endpoint":"file:///home/scott/spoc.json","stories_referrer":"https://getpocket.com/recommendations","topics_endpoint":"https://getpocket.cdn.mozilla.net/v3/firefox/trending-topics?version=2&consumer_key=$apiKey&locale_lang=en-US","show_spocs":true,"personalized":true}
```

You'll also need a file saved on your local system and to update the stories_endpoint with the file. i used this file: https://bug1479458.bmoattachments.org/attachment.cgi?id=8997024

4. open about:home again (or another one).

5. Ensure the new tab updates the third rec with something from theringer.com titled "The Kawhi-DeRozan Trade Is a Win for Both the Spurs and Raptors" If you kept the initial tab open from step 2, that should also be updated.

This basically simulates shield updating the pref.

### DIFF
--- a/content-src/components/Base/Base.jsx
+++ b/content-src/components/Base/Base.jsx
@@ -48,9 +48,29 @@ export class _Base extends React.PureComponent {
     this.updateTheme();
   }
 
-  componentWillUpdate({App}) {
+  hasTopStoriesSectionChanged(nextProps) {
+    const nPropsSections = nextProps.Sections.find(section => section.id === "topstories");
+    const tPropsSections = this.props.Sections.find(section => section.id === "topstories");
+    if (nPropsSections && nPropsSections.options) {
+      if (!tPropsSections || !tPropsSections.options) {
+        return true;
+      }
+      if (nPropsSections.options.show_spocs !== tPropsSections.options.show_spocs) {
+        return true;
+      }
+      if (nPropsSections.options.stories_endpoint !== tPropsSections.options.stories_endpoint) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  componentWillUpdate(nextProps) {
     this.updateTheme();
-    this.sendNewTabRehydrated(App);
+    if (this.hasTopStoriesSectionChanged(nextProps)) {
+      this.renderNotified = false;
+    }
+    this.sendNewTabRehydrated(nextProps.App);
   }
 
   updateTheme() {
@@ -154,4 +174,4 @@ export class BaseContent extends React.PureComponent {
   }
 }
 
-export const Base = connect(state => ({App: state.App, Prefs: state.Prefs}))(_Base);
+export const Base = connect(state => ({App: state.App, Prefs: state.Prefs, Sections: state.Sections}))(_Base);
--- a/lib/SectionsManager.jsm
+++ b/lib/SectionsManager.jsm
@@ -422,7 +422,7 @@ class SectionsFeed {
     this.store.dispatch(ac.SetPref("sectionOrder", orderedSections.join(",")));
   }
 
-  onAction(action) {
+  async onAction(action) {
     switch (action.type) {
       case at.INIT:
         SectionsManager.onceInitialized(this.init);
@@ -435,7 +435,7 @@ class SectionsFeed {
         if (action.data) {
           const matched = action.data.name.match(/^(feeds.section.(\S+)).options$/i);
           if (matched) {
-            SectionsManager.addBuiltInSection(matched[1], action.data.value);
+            await SectionsManager.addBuiltInSection(matched[1], action.data.value);
             this.store.dispatch({type: at.SECTION_OPTIONS_CHANGED, data: matched[2]});
           }
         }
--- a/lib/TopStoriesFeed.jsm
+++ b/lib/TopStoriesFeed.jsm
@@ -35,33 +35,52 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this._prefs = new Prefs();
   }
 
-  init() {
-    const initFeed = () => {
-      SectionsManager.enableSection(SECTION_ID);
-      try {
-        const {options} = SectionsManager.sections.get(SECTION_ID);
-        const apiKey = this.getApiKeyFromPref(options.api_key_pref);
-        this.stories_endpoint = this.produceFinalEndpointUrl(options.stories_endpoint, apiKey);
-        this.topics_endpoint = this.produceFinalEndpointUrl(options.topics_endpoint, apiKey);
-        this.read_more_endpoint = options.read_more_endpoint;
-        this.stories_referrer = options.stories_referrer;
-        this.personalized = options.personalized;
-        this.show_spocs = options.show_spocs;
-        this.maxHistoryQueryResults = options.maxHistoryQueryResults;
-        this.storiesLastUpdated = 0;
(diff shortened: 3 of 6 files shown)


ScottDowne:
@k88hudson Possible to get some initial feedback on this while I start adding something in the tests.

content-src/components/Base/Base.jsx
```diff
@@ -48,9 +48,25 @@ export class _Base extends React.PureComponent {
     this.updateTheme();
   }
 
-  componentWillUpdate({App}) {
+  hasSectionChanged(nextProps) {
+    if (this.props.Sections[0] && this.props.Sections[0].options) {
```
piatra: Could we replace this with something like
```js
this.props.Sections.find(s => s.id === "topstories")
```

content-src/components/Base/Base.jsx
```diff
@@ -48,9 +48,25 @@ export class _Base extends React.PureComponent {
...
+  componentWillUpdate(nextProps) {
     this.updateTheme();
-    this.sendNewTabRehydrated(App);
+    if (this.hasSectionChanged(nextProps)) {
+      this.renderNotified = false;
+    }
+    this.sendNewTabRehydrated(nextProps.App);
+
+    /*
+{"api_key_pref":"extensions.pocket.oAuthConsumerKey","hidden":false,"provider_icon":"pocket","provider_name":"Pocket","read_more_endpoint":"https://getpocket.com/explore/trending?src=fx_new_tab","stories_endpoint":"file:///home/scott/spoc.json","stories_referrer":"https://getpocket.com/recommendations","topics_endpoint":"https://getpocket.cdn.mozilla.net/v3/firefox/trending-topics?version=2&consumer_key=$apiKey&locale_lang=en-US","show_spocs":true,"personalized":true}
```
ScottDowne: This is just a note for myself, I can remove.

ScottDowne:
This is the bugzilla ticket https://bugzilla.mozilla.org/show_bug.cgi?id=1479458

This includes strs.

ScottDowne:
Test is failing because it's saying lines 341,349,515,518 in TopStoriesFeed.jsm are not covered, but from what I can tell, they are? Not sure what it's looking for in terms of covered?

Mardak:
Looks like the line numbers aren't accurate because of webpack (?). You can run tests locally with `npm test` then `npm run debugcoverage`

The missing lines are starred lines:
```
  maybeAddSpoc(target) {
    const updateContent = () => {
      if (!this.shouldShowSpocs()) {
  *      return false;

     if (!this.spocs || !this.spocs.length) {
        // We have stories but no spocs so there's nothing to do and this update can be
        // removed from the queue.
*        return false;

  onAction(action) {
…
        case at.SYSTEM_TICK:
…
          if (_this5.stories) {
*            updateProps = { rows: _this5.stories };

          if (_this5.topics) {
*            updateProps = Object.assign({}, updateProps, { topics: _this5.topics, read_more_endpoint: _this5.read_more_endpoint });
```

ScottDowne:
@Mardak Awesome, thanks!

ScottDowne:
@Mardak Also feel free to take a look at these changes.

There are some smaller issues, but the main issue is a race condition where `dispatchUpdateEvent` was being called multiple times inside init, which caused the redux reducer to reset changes and cause a race condition. It was exposed via changing `browser.newtabpage.activity-stream.feeds.section.topstories.options` via a shield test.

ScottDowne:
Fixed, thanks!

lib/TopStoriesFeed.jsm
```diff
@@ -35,33 +35,64 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+        shouldBroadcast = true;
+        await this.fetchTopics();
+      }
+
+      let updateProps = {};
+      if (this.stories) {
+        updateProps = {rows: this.stories};
+      }
+      if (this.topics) {
+        updateProps = {...updateProps, topics: this.topics, read_more_endpoint: this.read_more_endpoint};
```
k88hudson: In this case you could just do `updateProps.rows = this.stories` and `Object.assign(updateProps, {topics: this.topics, read_more_endpoint: this.read_more_endpoint}`

lib/SectionsManager.jsm
```diff
@@ -435,7 +435,7 @@ class SectionsFeed {
         if (action.data) {
           const matched = action.data.name.match(/^(feeds.section.(\S+)).options$/i);
           if (matched) {
-            SectionsManager.addBuiltInSection(matched[1], action.data.value);
+            await SectionsManager.addBuiltInSection(matched[1], action.data.value);
```
k88hudson: It might be safer to remove the await in `storage.get` inside this function rather than this since it's not strictly necessary, although I think this might be fine too

lib/TopStoriesFeed.jsm
```diff
@@ -448,18 +477,27 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+          await this.fetchTopics();
+        }
+
+        if (this.stories) {
+          updateProps = {rows: this.stories};
+        }
+        if (this.topics) {
+          updateProps = {...updateProps, topics: this.topics, read_more_endpoint: this.read_more_endpoint};
         }
+        this.dispatchUpdateEvent(false, updateProps);
```
k88hudson: Is there anyway you could create a function to handle both the SYSTEM_TICK and INIT updates in one spot?

lib/TopStoriesFeed.jsm
```diff
@@ -73,6 +104,9 @@ this.TopStoriesFeed = class TopStoriesFeed {
   }
 
   uninit() {
+    this.storiesLoaded = false;
+    this.cache.set("stories", {});
```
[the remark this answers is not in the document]
ScottDowne: I think this was based on an assumption that the spocs pref wasn't going to change often, thus not need to re fetch the data, and thus could rely on cache/existing data?

content-src/components/Base/Base.jsx
```diff
@@ -48,9 +48,29 @@ export class _Base extends React.PureComponent {
     this.updateTheme();
   }
 
-  componentWillUpdate({App}) {
+  hasSectionChanged(nextProps) {
+    const [nPropsSections] = nextProps.Sections;
```
k88hudson: Can you do perhaps do `nextProps.Sections.find(section => section.id === "topstories")` to make this more resilient to future changes?

Additionally, it might be a good idea to call this `hasTopStoriesSectionChanged`

k88hudson:
Can you rename the commit message to include the bug number like `Bug 123123 ...`

PR #4288, opened 2018-08-06
Fix Bug 1480518 - Replace default and frecent sites with @search when they match given URL pattern

piatra:
grafik
grafik
grafik

This will replace any default site that would show up in the topsites (and pin it to the first position) and any frecent website (no pinning or position change).
It adds a new action `FILL_SEARCH_TERM` that populates the location bar with the `@{search-provider}` key term.

Not sure (not implemented) any context menu options: if you unpin a default topsite turned into search topsite it will revert your action and pin it back again.

If a frecent site is turned into a search site you can still see the origin url when you hover the tile 
screen shot 2018-08-06 at 16 48 36

No icon (just the fallback text) and using the regular tippytop or high res icon as a small favicon.


lib/TopSitesFeed.jsm
```diff
@@ -35,8 +35,8 @@ const ROWS_PREF = "topSitesRows";
 // Search experiment stuff
 const NO_DEFAULT_SEARCH_TILE_EXP_PREF = "improvesearch.noDefaultSearchTile";
 const SEARCH_HOST_FILTERS = [
-  {hostname: "google", identifierPattern: /^google/},
-  {hostname: "amazon", identifierPattern: /^amazon/}
+  {hostname: "google", identifierPattern: /^google|^www.google/},
```
AdamHillier: The identifierPattern was originally to match the identity of the search engine in Firefox, which is a string like 'google' or 'google-2018' or 'amazon-en-GB'. You can see it used that way on line 52. It's bit confusing to also use it to match a hostname on line 61, maybe make the hostname a regex with `/^google|^www.google/` instead?

lib/TopSitesFeed.jsm
```diff
@@ -188,17 +192,30 @@ this.TopSitesFeed = class TopSitesFeed {
...
-    const withPinned = insertPinned(checkedAdult, pinned).slice(0, numItems);
+    // and check if any match default search topsite rules
+    for (const site of insertPinned(dedupedUnpinned, pinned).slice(0, numItems)) {
+      // A default site that is also a search topsite should be pinned to
+      // the first position
+      if (site.searchTopSite && site.isDefault && !site.isPinned) {
+        await this._insertPin(site, 0);
+        // Refresh the cache because we just modified pinned sites
+        this.pinnedCache.expire();
+        pinned = await this.pinnedCache.request();
```
AdamHillier: Does this result in multiple Places sql queries? If so, does it work if we just expire the cache and re-request once after the for loop instead?

piatra:
Just noticed this: the website url is google.com/photos so it fetches the photos icon. Right now it just uses whatever icon it would normally display for the Topsite but it should probably fetch the icon associated to the URL origin.
grafik

lib/TopSitesFeed.jsm
k88hudson: Yeah I would keep this list separate so as not to affect the other experiment anyway

k88hudson:
I'm noticing missing icons for amazon/google on the first run of a new profile, but not for a custom added amazon.com pinned tile: 

image

The icon URL appears to be `"undefined"` upon inspection.

lib/TopSitesFeed.jsm
```diff
@@ -42,12 +42,22 @@ const SEARCH_FILTERS = [
   "ask",
   "duckduckgo"
 ];
+// List of sites we match against Topsites in order to identify sites
+// that should be converted to search Topsites
+const TOPSITE_SEARCH_PROVIDERS = [
+  {hostname: /^google|^www.google/},
```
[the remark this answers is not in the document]
k88hudson: (I am realizing `hostname` is a bad name for this btw since it's really a normalized hostname)

lib/TopSitesFeed.jsm
```diff
@@ -159,11 +169,11 @@ this.TopSitesFeed = class TopSitesFeed {
           return false;
         }
         return true;
-      });
+      }).map(this.topSiteToSearchTopSite);
```
k88hudson: This is kind of a nit but it's more efficient to combine `.filter` and `.map` into one `.reduce` like in the above treatment of `frecent`

lib/TopSitesFeed.jsm
```diff
@@ -198,24 +208,43 @@ this.TopSitesFeed = class TopSitesFeed {
...
-      pinned, frecent, notBlockedDefaultSites);
-    const dedupedUnpinned = [...dedupedFrecent, ...dedupedDefaults];
-
-    // Remove adult sites if we need to
-    const checkedAdult = this.store.getState().Prefs.values.filterAdult ?
-      filterAdult(dedupedUnpinned) : dedupedUnpinned;
+    const dedupedUnpinned = this.dedupeTopSites({
+      // Convert any frecent site that matches TOPSITE_SEARCH_PROVIDERS to a search topsite
+      // without pinning or changing position
+      frecent: frecent.map(this.topSiteToSearchTopSite),
```
k88hudson: I wonder if it would make more sense to move this to ~line 151 to be more consistent with how you are handling defaults and also the other search experiment?

lib/TopSitesFeed.jsm
```diff
@@ -198,24 +208,43 @@ this.TopSitesFeed = class TopSitesFeed {
...
 
     // Insert the original pinned sites into the deduped frecent and defaults
-    const withPinned = insertPinned(checkedAdult, pinned).slice(0, numItems);
+    // and check if any match default search topsite rules
+    let withPinned = insertPinned(dedupedUnpinned, pinned).slice(0, numItems);
+    for (const site of withPinned) {
+      // A default site that is also a search topsite should be pinned to
+      // the first position
+      if (site && site.searchTopSite && site.isDefault && !site.isPinned) {
+        this._insertPin(site, 0);
```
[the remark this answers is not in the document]
k88hudson: Also, how is the order controlled here?

lib/TopSitesFeed.jsm
```diff
@@ -259,6 +301,37 @@ this.TopSitesFeed = class TopSitesFeed {
...
+    return site;
+  }
+
+  /**
+   * Fetch the icon corresponding to the URL origin
+   * @param site {object} LinksCache object corresponding to a TopSite
+   */
+  async _fetchOriginTopSiteIcon(site) {
+    try {
+      const originSite = {...site, url: new URL(site.url).origin};
```
k88hudson: This is probably OK, although probably doesn't need to be as complex as this (we might want to just hardcode a url in the SEARCH_PROVIDERS config with which to do a tippy-top lookup for example)

lib/TopSitesFeed.jsm
```diff
@@ -42,12 +42,20 @@ const SEARCH_FILTERS = [
   "ask",
   "duckduckgo"
 ];
+const SEARCH_SHORTCUTS_EXPERIMENT = "improvesearch.topSiteSearchShortcuts";
+// List of sites we match against Topsites in order to identify sites
+// that should be converted to search Topsites
+const TOPSITE_SEARCH_PROVIDERS = [/^google/, /^amazon/];
```
k88hudson: change these to plain strings

lib/TopSitesFeed.jsm
```diff
@@ -42,12 +42,20 @@ const SEARCH_FILTERS = [
...
 
 function getShortURLForCurrentSearch() {
   const url = shortURL({url: Services.search.currentEngine.searchForm});
   return url;
 }
 
+function isSearchProvider(shortUrl) {
+  return TOPSITE_SEARCH_PROVIDERS.some(match => shortUrl.match(match));
```
k88hudson: let's just do `===`

lib/TopSitesFeed.jsm
```diff
@@ -145,21 +158,27 @@ this.TopSitesFeed = class TopSitesFeed {
     .reduce((validLinks, link) => {
       const hostname = shortURL(link);
       if (!this.isExperimentOnAndLinkFilteredSearch(hostname)) {
-        validLinks.push({...link, hostname});
+        validLinks.push({
+          hostname,
+          ...(searchShortcutsExperiment ? this.topSiteToSearchTopSite(link) : link)
```
k88hudson: can you reverse this order just in case

k88hudson:
🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 🥇 finish liiiiine YES!

Just fix up those things, squash, and this is good to merge;

I noticed an issue where pinning a searchysearchy causes the background of the tile to change form a `@` to a letter, which i suspect is unintended. However, this doesn't need to be a blocker and I will file as a follow-up.

Thank you for getting this done!

PR #4289, opened 2018-08-07
Fix Bug 1480508: Add search shortcuts for new and existing users

AdamHillier:
WIP, based off Andrei's PR. Will need to rebase when his lands and add more tests.

### DIFF
--- a/lib/TopSitesFeed.jsm
+++ b/lib/TopSitesFeed.jsm
@@ -43,17 +43,21 @@ const SEARCH_FILTERS = [
   "duckduckgo"
 ];
 const SEARCH_SHORTCUTS_EXPERIMENT = "improvesearch.topSiteSearchShortcuts";
+const SEARCH_SHORTCUTS_HAVE_PINNED_PREF = "improvesearch.topSiteSearchShortcuts.havePinned";
 // List of sites we match against Topsites in order to identify sites
 // that should be converted to search Topsites
-const TOPSITE_SEARCH_PROVIDERS = ["google", "amazon"];
+const SEARCH_SHORTCUTS = [
+  {keyword: "@google", shortURL: "google", url: "https://google.com", searchIdentifier: /^google/},
+  {keyword: "@amazon", shortURL: "amazon", url: "https://amazon.com", searchIdentifier: /^amazon/}
+];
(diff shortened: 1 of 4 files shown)


k88hudson:
Ok, after some manual testing this is working well in the default case, however for organic top sites, google/amazon are getting pinned to the first and second positions in all cases whereas they should be pinned to their organic positions. Looking at the code now

lib/TopSitesFeed.jsm
```diff
@@ -70,10 +71,13 @@ this.TopSitesFeed = class TopSitesFeed {
     PageThumbs.addExpirationFilter(this);
   }
 
-  init() {
+  async init() {
     // If the feed was previously disabled PREFS_INITIAL_VALUES was never received
     this.refreshDefaults(this.store.getState().Prefs.values[DEFAULT_SITES_PREF]);
     this._storage = this.store.dbStorage.getDbTable("sectionPrefs");
+    if (this.store.getState().Prefs.values[SEARCH_SHORTCUTS_EXPERIMENT]) {
```
k88hudson: I don't think it's sufficient to only run this on `init`, this also need needs to be run when the pref updates etc. (e.g. when shield sets the experiment pref, particularly if we decide to ship to release with this default-off)

Perhaps this should be checked on every refresh i.e. in `getLinksWithDefaults`?

AdamHillier:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1481680

lib/TopSitesFeed.jsm
```diff
@@ -146,6 +150,65 @@ this.TopSitesFeed = class TopSitesFeed {
...
+        Array(numberOfSlots - plainPinnedSites.length).fill(null)
+      );
+
+      await new Promise(resolve => Services.search.init(resolve));
+
+      const tryToInsertSearchShortcut = shortcut => {
+        const nextAvailable = pinnedSites.indexOf(null);
+        // Only add a search shortcut if the site isn't already pinned, we
+        // haven't previously inserted it, there's space to pin it, and the
+        // search engine is available in Firefox
```
k88hudson: :+1:

PR #4294, opened 2018-08-07
feat(experiments): add text taggers

jonathankoren:
### Update 2
* Will convert object to maps later under https://bugzilla.mozilla.org/show_bug.cgi?id=1487456
* For...ofs in tests except for one, see comment in https://github.com/mozilla/activity-stream/pull/4294#discussion_r214102213

### Update
* Switched to for..of loops
* Removed double-quote on keys
* **Did not** switch from Objects to Maps. We use Object in a lot of places in the follow on PR (which you haven't seen yet). This would be a big change. It would mean converting the fundamental data structure. Honestly, I'm not comfortable with making that change at this time. I'm open to doing it later, especially if we need a speed up, but not today.

### Description
This PR adds the ability to classify text. We define two different classifiers, a Naïve Bayes (NB) classifier, and a multiclass nonnegative matrix factorization (NMF) classifier. Both use a bag of words, TF-IDF vectors as features. The purpose of this code is to allow Firefox to classify pages into topics, by examining the text found on the page.

This code is part of the Pocket Personalization v2 experiment which uses content analysis to locally build interest profiles.

This code is dark.

### Testing
Unit tests
This code has no current consumers.

### Related
We reviewed this internally on PR https://github.com/Pocket/activity-stream/pull/1

### See Also
https://docs.google.com/document/d/12OtUZywivIvBnO3hmMNjptmIQ8cQLFOIOzJO4bRLqdg/edit
https://en.wikipedia.org/wiki/Naive_Bayes_classifier
https://en.wikipedia.org/wiki/Non-negative_matrix_factorization
https://en.wikipedia.org/wiki/Tf%E2%80%93idf


k88hudson:
@jonathankoren hey, when is the expected ship date for this experiment / i.e. how soon do you need this merged / reviewed?

jonathankoren:
@k88hudson We're aiming to land the feature before August 23 code freeze. There's more code to be written, but because it's a fairly large amount new code, I thought it was prudent to break it up into smaller PRs rather than hold on to one giant PR. The idea was that it makes it easier to understand, and lowers the overall risk. The feature is going to gated behind an A/B test and will launch dark, and then be turned on for some fraction of users through Shield.

jonathankoren:
If it's useful, I'm on slack, and have been working with Scott Downe, who I think sits next to you.

k88hudson:
Ok thanks! Just wanted to know if this was planned for an uplift or not.
We are a little under pressure this week with another feature that needs to get uplifted, but we'll try to get someone to look at this soon. Your approach to break it up into a bunch of PRs sounds great 👍

jonathankoren:
The plan is to try _really_ _really_ hard to not have to get uplift after the Sept 4 beta release, but if we miss the date, we'll need uplift to beta. Otherwise, we're stuck for a long time. Fortunately, we're going to be behind a flag, so hopefully it should possible.

ncloudioj:
Let's also track this on Bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1483667

lib/NaiveBayesTextTagger.jsm
```diff
@@ -0,0 +1,64 @@
...
+   * label. If the negative class is matched, then "label" is set to null.
+   */
+  tag(text) {
+    let fv = this.tokenizer.getTfIdfVector(text, this.model.vocab_idfs);
+
+    let bestLogProb = null;
+    let bestClassId = -1;
+    let bestClassLabel = null;
+    let logSumExp = 0.0; // will be P(x). Used to create a proper probability
+    for (let classId = 0; classId < this.model.classes.length; classId++) {
```
ncloudioj: Nit: [for..of statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is more succinct for array iteration. Consider replacing all the raw for loops with `for..of`?

lib/NaiveBayesTextTagger.jsm
```diff
@@ -0,0 +1,64 @@
...
+    if (bestClassId === this.model.positive_class_id) {
+      bestClassLabel = this.model.positive_class_label;
+    } else {
+      bestClassLabel = null;
+    }
+
+    let confident = ((bestClassId === this.model.positive_class_id) &&
+                      (bestLogProb > this.model.positive_class_threshold_log_prob));
+    return {
+      "label": bestClassLabel,
```
ncloudioj: Nit: redundant double quotes for keys.

lib/NaiveBayesTextTagger.jsm
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      if ((bestLogProb === null) || (classLogProb > bestLogProb)) {
+        bestLogProb = classLogProb;
+        bestClassId = classId;
+      }
+      logSumExp += Math.exp(classLogProb);
+    }
+
+    // now normalizae the probability by dividing by P(x)
```
ncloudioj: s/normalizae/normalize/

lib/NmfTextTagger.jsm
```diff
@@ -0,0 +1,62 @@
...
+      for (let j = 0; j < fve.length; j++) {
+        // eslint-disable-next-line prefer-destructuring
+        let [termId, tfidf] = fve[j];
+        fvDotTwv += tfidf * this.model.topic_word[topicId][termId];
+      }
+      toksInLatentSpace.push(fvDotTwv);
+    }
+
+    // now project toksInLatentSpace back into class space
+    let predictions = {};
```
ncloudioj: Nit: prefer to use [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+// Javascript doesn't support unicode page mnemonics out of the box, and so
+// we have to define them ourselves.
+const UNICODE_SPACE = "\x20\xA0\u1680\u2000-\u200A\u2028\u2029\u202F\u205F\u3000";
```
ncloudioj: All these constants here want some comments to clarify:
* What they are in general
* Some extra reference links, which pointing to more details about those unicodes, would be even better

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,74 @@
...
+
+const REGEXP_SPLITS = new RegExp(`[${UNICODE_SPACE}${UNICODE_SYMBOL}${UNICODE_PUNCT}]+`);
+const REGEXP_ALPHANUMS = new RegExp(`^[${UNICODE_NUMBER}${UNICODE_MARK}${UNICODE_LETTER}]+$`);
```
ncloudioj: Do we also want to filter other stopwords? Which could help reduce the size of result.

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,74 @@
...
+
+this.TfIdfVectorizer = class TfIdfVectorizer {
+  /**
+   * Downcases the text, and splits it into consecutive alphanumeric characters.
+   * This is locale aware, and so will not strip accents. This uses "word
+   * breaks", and os is not appropriate for languages without them
```
ncloudioj: what do you mean by "os"?

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,74 @@
...
+   * (e.g. Chinese).
+   */
+  tokenize(text) {
+    return text.toLocaleLowerCase().split(REGEXP_SPLITS).filter(tok => tok.match(REGEXP_ALPHANUMS));
```
ncloudioj: These two pieces of regex processing could be quite CPU intensive, we might want to keep an eye on it by profiling.

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * Converts a sequence of tokens into an L2 normed TF-IDF. Any terms that are
+   * not preindexed (i.e. does has a computed inverse document frequency) will
+   * be dropped.
+   */
+  toksTotfIdfVector(tokens, vocab_idfs) {
+    let tfidfs = {};
```
ncloudioj: Again, a `Map` would be better than object for this.

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,74 @@
...
+    // now multiply by the log inverse document frequencies, then take
+    // the L2 norm of this.
+    let l2Norm = 0.0;
+    Object.keys(tfidfs).forEach(tok => {
+      tfidfs[tok][1] *= vocab_idfs[tok][1];
+      l2Norm += tfidfs[tok][1] * tfidfs[tok][1];
+    });
+    l2Norm = Math.sqrt(l2Norm);
+    Object.keys(tfidfs).forEach(tok => {
+      tfidfs[tok][1] /= l2Norm;
```
ncloudioj: Any chance to see `l2Norm === 0`? If so, better add a guard to prevent the divided-by-zero error.

lib/TfIdfVectorizer.jsm
jonathankoren: Unicode specifies certain mnemonics for code pages and character classes. They call them "character properties" https://en.wikipedia.org/wiki/Unicode_character_property . These mnemonics are have been adopted by many regular expression libraries, however the standard regexp system doesn't support unicode character properties. If you want to use them, you have to pull in a library like XRegExp. 

Each of these sections contains the characters values / ranges for specific character property: Whitespace, Symbol (S), Punctuation (P), Number (N), Mark (M), and Letter (L). The tables are defined by Unicode. I think I reformatted the data from XRegExp to get it to work here.

We have to support unicode the tokenizer, as we expect to see (and have seen in our training data) nonenglish and multilingual documents. This allows us to use the same code for any language that uses spaces to separate words.

Interestingly enough, it looks like xregexp is pulled in via two or three levels of dependencies, but I have no idea how / should you'd import it directly.

lib/TfIdfVectorizer.jsm
jonathankoren: That wouldn't happen here. This is simply cutting up a stream of bytes into tokens to check against the dictionary. You have to do this even before you do stopword removal, because you don't even know what the words are yet.

We end up doing something much more aggressive than stop word removal because the tagging models use closed vocabulary. If the word isn't in the vocabulary, we simply drop it.

lib/TfIdfVectorizer.jsm
jonathankoren: That's just a typo. It looks like I accidentally doubled the word "is", but ended up typing it wrong the first time.

lib/TfIdfVectorizer.jsm
jonathankoren: What's the difference?

ncloudioj: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared

lib/TfIdfVectorizer.jsm
ncloudioj: Could you add this to the code?

lib/NaiveBayesTextTagger.jsm
ncloudioj: Did you miss this?

lib/NaiveBayesTextTagger.jsm
ncloudioj: ditto

test/unit/lib/NaiveBayesTextTagger.test.js
```diff
@@ -0,0 +1,106 @@
+import {NaiveBayesTextTagger} from "lib/NaiveBayesTextTagger.jsm";
+import {TfIdfVectorizer} from "lib/TfIdfVectorizer.jsm";
+
+const EPSILON = 0.00001;
+
+describe("Naive Bayes Tagger", () => {
+  describe("#tag", () => {
+    let model = {
+      "model_type": "nb",
```
ncloudioj: Let's also remove the double quotes in this files.

test/unit/lib/NaiveBayesTextTagger.test.js
```diff
@@ -0,0 +1,106 @@
...
+        assert.equal(tc.expected.label, actual.label);
+      });
+      it(`should give ${tc.input} the correct probability`, () => {
+        let delta = Math.abs(tc.expected.logProb - actual.logProb);
+        assert.isTrue(delta <= EPSILON);
+      });
+    };
+
+    // RELEASE THE TESTS!
+    for (let i = 0; i < testCases.length; i++) {
```
ncloudioj: Let's do `for..of` for this too.

test/unit/lib/NmfTextTagger.test.js
```diff
@@ -0,0 +1,115 @@
...
+        Object.keys(actual).forEach(tag => {
+          let delta = Math.abs(tc.expected[tag] - actual[tag]);
+          assert.isTrue(delta <= EPSILON);
+        });
+      });
+    };
+
+    // RELEASE THE TESTS!
+    for (let i = 0; i < testCases.length; i++) {
+      checkTag(testCases[i]);
```
ncloudioj: ditto

test/unit/lib/TfIdfVectorizer.test.js
```diff
@@ -0,0 +1,110 @@
...
+      it(`${tc.input} should have the correct tf-idf`, () => {
+        Object.keys(actual).forEach(actualTok => {
+          let delta = Math.abs(tc.expected[actualTok][1] - actual[actualTok][1]);
+          assert.isTrue(delta <= EPSILON);
+        });
+      });
+    };
+
+    // run the tests
+    for (let i = 0; i < testCases.length; i++) {
```
ncloudioj: ditto

lib/NmfTextTagger.jsm
```diff
@@ -0,0 +1,61 @@
...
+        fvDotTwv += tfidf * topicVect[termId];
+      }
+      toksInLatentSpace.push(fvDotTwv);
+    }
+
+    // now project toksInLatentSpace back into class space
+    let predictions = {};
+    Object.keys(this.model.document_topic).forEach(topic => {
+      let score = 0;
+      for (let i = 0; i < toksInLatentSpace.length; i++) {
```
ncloudioj: Let's `for..of` this.

jonathankoren: You cant do that here. It's the index is across two different arrays.

lib/TfIdfVectorizer.jsm
```diff
@@ -0,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+// Javascript doesn't support unicode page mnemonics out of the box, and so
+// we have to define them ourselves.
+const UNICODE_SPACE = "\x20\xA0\u1680\u2000-\u200A\u2028\u2029\u202F\u205F\u3000";
```
ncloudioj: Please add the comments that you added in 6eae5458e65458e17b8e4d7c003f0cfb6cca7ff5?

ncloudioj:
@jonathankoren There is an eslint error reported by Taskcluster, could you do a rebase against master to see if it fixes it?

PR #4298, opened 2018-08-08
Fix Bug 1481677 - persistent/stable search (aka fixed position search bar)

rlr:
Be sure to test scrolling at different window heights.
* At <= 700px it shouldn't become fixed and it has less padding.
* At > 700px it should turn into the persistent header search bar.

Verify dark theme too.

before:
screen shot 2018-08-08 at 2 24 47 pm

after:
screen shot 2018-08-08 at 2 24 00 pm

screen shot 2018-08-08 at 2 27 49 pm

screen shot 2018-08-08 at 2 28 32 pm

### DIFF
--- a/content-src/components/Base/Base.jsx
+++ b/content-src/components/Base/Base.jsx
@@ -24,6 +24,20 @@ function addLocaleDataForReactIntl(locale) {
   addLocaleData([{locale, parentLocale: "en"}]);
 }
 
+// Returns a function will not be continuously triggered when called. The
+// function will be triggered if called again after `wait` milliseconds.
+function debounce(func, wait) {
+  let timer;
+  return (...args) => {
+    if (timer) { return; }
+
+    let wakeUp = () => { timer = null; };
+
+    timer = setTimeout(wakeUp, wait);
+    func.apply(this, args);
+  };
+}
+
 export class _Base extends React.PureComponent {
   componentWillMount() {
     const {App, locale} = this.props;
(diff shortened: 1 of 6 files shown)


content-src/components/Base/Base.jsx
```diff
@@ -106,6 +106,25 @@ export class BaseContent extends React.PureComponent {
   constructor(props) {
     super(props);
     this.openPreferences = this.openPreferences.bind(this);
+    this.onWindowScroll = this.onWindowScroll.bind(this);
+    this.state = {fixedSearch: false};
+  }
+
+  componentDidMount() {
+    global.addEventListener("scroll", this.onWindowScroll);
```
rlr: before I add the debounce... i sit needed or is that a premature optimization?

content-src/components/Search/Search.jsx
```diff
@@ -64,23 +64,25 @@ export class _Search extends React.PureComponent {
...
-        title={this.props.intl.formatMessage({id: "search_web_placeholder"})}
-        type="search" />
-      <button
-        id="searchSubmit"
-        className="search-button"
-        onClick={this.onClick}
-        title={this.props.intl.formatMessage({id: "search_button"})}>
-        <span className="sr-only"><FormattedMessage id="search_button" /></span>
-      </button>
+      <div className="search-inner-wrapper">
```
rlr: i only added this extra wrapper here. the rest of the diff is indentation.

piatra:
There's a slight gap between the suggestions and the search bar when you scroll.

Top of the page
grafik
Scrolled down
grafik

rlr:
@piatra yep, I noticed that but it's only if you scroll with the dropdown open. Unlikely? If you type again it fixes itself. I don't think there is anything I can do from our side there.

content-src/components/Base/Base.jsx
k88hudson: nah definitely a good idea

k88hudson: scroll is already janky on some machines

content-src/components/Search/Search.jsx
k88hudson: That's fine

content-src/components/Base/Base.jsx
```diff
@@ -106,6 +120,25 @@ export class BaseContent extends React.PureComponent {
   constructor(props) {
     super(props);
     this.openPreferences = this.openPreferences.bind(this);
+    this.onWindowScroll = debounce(this.onWindowScroll.bind(this), 5);
```
rlr: I made the wait only 5ms because even then when you scroll fast you can see some jank when the header appears. Better than nothing?

rlr: (I guess we can tweak it based on feedback once it lands?)

PR #4300, opened 2018-08-08
Fix bug in manually pinned site appearing as a search shortcut

AdamHillier:
I noticed a bug where if I manually pinned a new site with the url "amazon.co.uk" and a custom label (e.g. "amazon uk"), then it would get transformed into a broken search shortcut. When clicking on the site the url bar would get focused and instead of inserting "@amazon" it inserted the custom label. This is a simple problem with the ordering of the properties in topSiteToSearchTopSite.


AdamHillier:
Oh oops was actually fixed in the commit I needed to rebase onto https://github.com/mozilla/activity-stream/commit/5de8ec083b2fc89e58204355e827a4806889255c

PR #4301, opened 2018-08-09
Add utm params to initial fetch call.

ericawright:
utm params are added to the initial fetch call to give more information to FxA
as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1478649#c5

fixes: Bug 1478649

### DIFF
--- a/content-src/components/StartupOverlay/StartupOverlay.jsx
+++ b/content-src/components/StartupOverlay/StartupOverlay.jsx
@@ -26,7 +26,8 @@ export class _StartupOverlay extends React.PureComponent {
     if (this.props.fxa_endpoint && !this.didFetch) {
       try {
         this.didFetch = true;
-        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);
+        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow?entrypoint=
+          activity-stream-firstrun&utm_source=activity-stream&utm_campaign=firstrun&form_type=email`);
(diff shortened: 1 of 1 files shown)


k88hudson:
R+, this URL matches the one posted in https://bugzilla.mozilla.org/show_bug.cgi?id=1478649#c5

@ericawright Would you mind editing your commit to include the bug number so we can export this more easily, i.e.

```
Bug 1478649 - Add utm params to initial fetch call
```

content-src/components/StartupOverlay/StartupOverlay.jsx
```diff
@@ -26,7 +26,8 @@ export class _StartupOverlay extends React.PureComponent {
     if (this.props.fxa_endpoint && !this.didFetch) {
       try {
         this.didFetch = true;
-        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow`);
+        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow?entrypoint=
+          activity-stream-firstrun&utm_source=activity-stream&utm_campaign=firstrun&form_type=email`);
```
vladikoff: Thanks for adding these @ericawright !  👍 

@k88hudson one thing to keep in mind, if the values in https://github.com/mozilla/activity-stream/blob/27696cb3c9d88e32c03b6040102de678c99f16a5/content-src/components/StartupOverlay/StartupOverlay.jsx#L115 ever change then will need to change in this file as well.

PR #4304, opened 2018-08-09
Fix Bug 1482184 - only apply fixed-search if search section is visible

rlr:
ez fixy. r? @k88hudson

### DIFF
--- a/content-src/components/Base/Base.jsx
+++ b/content-src/components/Base/Base.jsx
@@ -157,7 +157,7 @@ export class BaseContent extends React.PureComponent {
     const outerClassName = [
       "outer-wrapper",
       shouldBeFixedToTop && "fixed-to-top",
-      this.state.fixedSearch && "fixed-search"
+      prefs.showSearch && this.state.fixedSearch && "fixed-search"
     ].filter(v => v).join(" ");
 
     return (


content-src/components/Base/Base.jsx
```diff
@@ -157,7 +157,7 @@ export class BaseContent extends React.PureComponent {
     const outerClassName = [
       "outer-wrapper",
       shouldBeFixedToTop && "fixed-to-top",
-      this.state.fixedSearch && "fixed-search"
+      prefs.showSearch && this.state.fixedSearch && "fixed-search"
```
k88hudson: Do you need to remove the listeners on a pref change as well for scrolling?

rlr: well. right now I keep the scroll handlers on all the time. i guess i can optimize that to only listen if `showSearch` is true 🤔

k88hudson: oh nevermind that's in the search component

k88hudson: Let's leave it as is for now since the code is less risky, maybe we should file as a follow-up

content-src/components/Base/Base.jsx
```diff
@@ -133,6 +133,8 @@ export class BaseContent extends React.PureComponent {
   }
 
   onWindowScroll() {
+    if (!this.props.Prefs.values.showSearch) { return; }
```
rlr: @k88hudson ☝️ like this?

k88hudson: I'm testing that, I think it's ok

k88hudson: Yep, that works

k88hudson: Actually on second thought, this doesn't work properly if you flip search off in one state (i.e. sticky) and turn it back on in the non-sticky state. I think you should just remove it. Oh well!

PR #4305, opened 2018-08-09
Fix Bug 1482190: don't convert manually pinned sites to search shortcuts

AdamHillier:
Oops was caused by my patch. Solution is straightforward.


k88hudson:
Aha! Just manually testing, but this makes sense

PR #4312, opened 2018-08-10
fix(topsites) - TopSitesFeed would fail if it got more pinned sites than slots

piatra:
This resulted in AS not loading in Nightly on first run after update.
The issue happened because I had 19 pinned sites and 16 slots available. This might be a likely scenario if you sync your profile across two different devices one getting wide layout and one getting small layout with fewer topsite slots.
screen shot 2018-08-10 at 14 58 53

### DIFF
--- a/lib/TopSitesFeed.jsm
+++ b/lib/TopSitesFeed.jsm
@@ -173,8 +173,9 @@ this.TopSitesFeed = class TopSitesFeed {
       // The plainPinnedSites array is populated with pinned sites at their
       // respective indices, and null everywhere else, but is not always the
       // right length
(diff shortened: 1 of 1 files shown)


lib/TopSitesFeed.jsm
```diff
@@ -173,8 +173,9 @@ this.TopSitesFeed = class TopSitesFeed {
       // The plainPinnedSites array is populated with pinned sites at their
       // respective indices, and null everywhere else, but is not always the
       // right length
+      const emptySlots = numberOfSlots > plainPinnedSites.length ? numberOfSlots - plainPinnedSites.length : 0;
```
AdamHillier: Nit: might be cleaner to just do `Math.max(..., 0)`

PR #4314, opened 2018-08-10
Fix Bug 1482404 - remove unnecessary search background-color

rlr:
r? @Mardak or @k88hudson ?


rlr:
how did I break taskcluster?

PR #4316, opened 2018-08-10
Fix Bug 1481680 - Add mochitests for search shortcuts and fix existing mochitests

piatra:
f? @Mardak still working on this.

### DIFF
--- a/test/browser/browser.ini
+++ b/test/browser/browser.ini
@@ -5,8 +5,6 @@ support-files =
   head.js
 prefs =
   browser.newtabpage.activity-stream.debug=false
-# Temporarily disable the search improvements for mochitests until bug 1481680
-  browser.newtabpage.activity-stream.improvesearch.topSiteSearchShortcuts=false
 
 [browser_activity_stream_strings.js]
 [browser_as_load_location.js]
--- a/test/browser/browser_topsites_contextMenu_options.js
+++ b/test/browser/browser_topsites_contextMenu_options.js
@@ -5,16 +5,13 @@
 "use strict";
 
 test_newtab({
-  before: async () => {
-    await disableSearchImprovementsPrefs();
-    await setDefaultTopSites();
-  },
+  before: setDefaultTopSites,
   // Test verifies the menu options for a default top site.
   test: async function defaultTopSites_menuOptions() {
     await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".top-site-icon"),
       "Topsite tippytop icon not found");
 
-    let contextMenuItems = content.openContextMenuAndGetOptions(".top-sites-list li:first-child").map(v => v.textContent);
+    let contextMenuItems = content.openContextMenuAndGetOptions(".top-sites-list li:not(.search-shortcut)").map(v => v.textContent);
 
     Assert.equal(contextMenuItems.length, 5, "Number of options is correct");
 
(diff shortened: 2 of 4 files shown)


test/browser/browser.ini
```diff
@@ -5,8 +5,6 @@ support-files =
   head.js
 prefs =
   browser.newtabpage.activity-stream.debug=false
-# Temporarily disable the search improvements for mochitests until bug 1481680
-  browser.newtabpage.activity-stream.improvesearch.topSiteSearchShortcuts=false
```
sarracini: i think we should keep this, and then just enable it for the one test which we want to test the search shortcuts for... that way we don't have to keep disabling in the future when we add more mochitests that doesn't want this off, and the in the future when we enable this permanently we won't need to go in and remove all the disable code for each tests

Mardak: On the flip side, the default behavior is to have top site search shortcuts, so our tests should be closer to the default behavior

Mardak: Although looking at how the tests were fixed, they're disabling anyway… I wonder if the tests should work with it enabled instead

Mardak: Although then again... I would assume this pref is going away at some point.. soon? So we should not rely on being able to change the pref to false. ??

piatra: I agree with that last part, I'm going to keep the pref on and update the test to expect the search topsites feature to exist.

piatra:
Want to get this landed and then look into adding mochitests for our CFR panel.

test/browser/browser_topsites_section.js
```diff
@@ -116,3 +118,21 @@ test_newtab({
...
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelector(".search-shortcut .title.pinned"), "Wait for pinned search topsites");
+
+    const searchTopSites = content.document.querySelectorAll(".title.pinned");
+    ok(searchTopSites.length >= 2, "There should be at least 2 search topsites");
+
+    searchTopSites[0].click();
+
+    return searchTopSites[0].innerText;
+  },
+  after: searchTopSiteTag => {
```
Mardak: Aha! Passing data from test content to main. 👍 

I believe this can be a plain method: `after(searchTopSiteTag) {`

test/browser/browser_topsites_section.js
```diff
@@ -116,3 +118,21 @@ test_newtab({
...
+    ok(gURLBar.focused, "We clicked a search topsite the focus should be in location bar");
+    ok(gURLBar.value.includes(searchTopSiteTag), "Should contain the tag of the search topsite clicked");
```
Mardak: Not sure if we need to clean up the value as tests open up new tabs anyway and this one is going away. Also, unclear if we need to wait any longer than the content task finishing.

I suppose if tests and verify are happy, then no need to change.

test/browser/browser_topsites_contextMenu_options.js
```diff
@@ -30,24 +30,46 @@ test_newtab({
...
+
+    // Unpin
+    contextMenuItems[0].querySelector("a").click();
+
+    await ContentTaskUtils.waitForCondition(() => content.document.querySelectorAll(".search-shortcut").length === 1,
+      "1 search topsite displayed after we unpin the other one");
+  },
+  after: () => {
+    // This is modified when calling .unpin in `disableSearchImprovementsPrefs`
+    Services.prefs.clearUserPref("browser.newtabpage.activity-stream.improvesearch.topSiteSearchShortcuts.havePinned");
```
Mardak: Is this necessary? At least the comment is outdated. Potentially we want to clean up / prepare for fresh top search more generally?

piatra: The pref clear is required for multiple test runs in the same browser (`--verify`) which is what the TV mochitest suite (I think) is doing. On the second test run the TopSitesFeed wouldn't re-pin the unpinned topsite.
`disableSearchImprovements` would do a proper cleanup, you would need to toggle the `SEARCH_SHORCUTS_EXPERIMENT` in order to trigger that. I would do it as a follow up because it involves more code (wait & assert that cleanup happend).

PR #4317, opened 2018-08-10
Fix Bug 1480507 - Add/Edit new Top Search Shortcut

rlr:
This is ready to be looked at and tested. @k88hudson @piatra @Mardak 👀 ?

I still need to write tests.


common/Reducers.jsm
```diff
@@ -23,7 +23,14 @@ const INITIAL_STATE = {
...
     rows: [],
     // Used in content only to dispatch action to TopSiteForm.
-    editForm: null
+    editForm: null,
+    // Used in content only to open the SearchShortcutsForm modal.
+    showSearchShortcutsForm: false,
+    // The list of available search shortcuts.
+    searchShortcuts: [
+      {keyword: "@google", shortURL: "google", url: "https://www.google.com", icon: "resource://activity-stream/data/content/tippytop/images/google-com@2x.png"},
+      {keyword: "@amazon", shortURL: "amazon", url: "https://www.amazon.com", icon: "resource://activity-stream/data/content/tippytop/images/amazon@2x.png"}
```
rlr: This needs to be hooked up with real data from somewhere. @Mardak ?

content-src/components/TopSites/SearchShortcutsForm.jsx
```diff
@@ -0,0 +1,143 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
+import {FormattedMessage} from "react-intl";
+import React from "react";
+
+class SelectableSearchShortcut extends React.PureComponent {
+  render() {
+    const {shortcut, selected} = this.props;
+    const imageStyle = {backgroundImage: `url("${shortcut.icon}")`};
+    return (
+      <div className="top-site-outer">
```
rlr: There is a bit of duplication of HTML between this component and the `TopSite` component. But I didn't want to make a mess out of that component right now. I figured this duplication is less risky for now.

content-src/components/TopSites/SearchShortcutsForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+    const {rows} = this.props.TopSites;
+    const pinQueue = [];
+    const unpinQueue = [];
+    this.state.shortcuts.forEach(shortcut => {
+      const alreadyPinned = rows.find(({isPinned, searchTopSite, label}) => isPinned && searchTopSite && label === shortcut.keyword);
+      if (shortcut.isSelected && !alreadyPinned) {
+        pinQueue.push(this._searchTopSite(shortcut));
+      } else if (!shortcut.isSelected && alreadyPinned) {
+        unpinQueue.push({url: alreadyPinned.url});
+      }
```
rlr: All this pinning and unpinning logic might want to be in a feed instead. But.... good enough for now?

common/Reducers.jsm
Mardak: @sarracini @AdamHillier I would think this data needs to match the format of https://github.com/mozilla/activity-stream/blob/9ddc1713fab5dfbc7f90321ad6c4c69d38535b3e/lib/SearchShortcuts.jsm#L9-L12

We'll want to derive the data from something like `Services.search.getEngines().map(v => v.wrappedJSObject._internalAliases)` from https://hg.mozilla.org/integration/autoland/rev/1f187635f2e4#l3.12

k88hudson: We can add this to the reducer from `TopSitesFeed`

content-src/components/TopSites/SearchShortcutsForm.jsx
Mardak: Can we not just insert into the first position? Is the dialog supposed to show already pinned stuff?

rlr: I assumed it would always show all the options available?

common/Reducers.jsm
Mardak: Notably, the urls used here are different from those in SearchShortcuts, which will affect blocking

content-src/components/TopSites/SearchShortcutsForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+      const availableSlots = [];
+      rows.forEach(({isPinned, searchTopSite, url}, index) => {
+        if (!isPinned || (searchTopSite && unpinQueue.find(site => url === site.url))) {
+          availableSlots.push(index);
+        }
+      });
+
+      pinQueue.forEach(shortcut => {
+        if (availableSlots.length === 0) {
+          // if they’ve pinned all their top sites then we do nothing
```
k88hudson: This seems like it might result in unexpected behaviour for users; I wonder if there's a way we could  prevent this state?

k88hudson: From our discussion with aaron, it sounds like the intended behaviour is to insert it at the first position and shift everything to the right

AdamHillier:
Do we want to update the `improvesearch.topSitesSearchShortcuts.havePinned` pref when a user manually adds a site @k88hudson?

lib/TopSitesFeed.jsm
```diff
@@ -340,6 +342,20 @@ this.TopSitesFeed = class TopSitesFeed {
     if (!this._tippyTopProvider.initialized) {
       await this._tippyTopProvider.init();
     }
+    if (this.store.getState().Prefs.values[SEARCH_SHORTCUTS_EXPERIMENT] && !this.searchShortcutsInitialized) {
+      // Populate the state with available search shortcuts
+      const searchShortcuts = [];
+      SEARCH_SHORTCUTS.forEach(shortcut => {
```
k88hudson: Can you look through the available search engines in `Services.search.getEngines()` and match on `engine.identifier.match(shortcut.searchIdentifier)`

k88hudson: Can you also add an observer for engine changes as well to cause a refresh?

Mardak: `engine.identifier` can be null (for added engines), so check before `.match`-ing otherwise we run into bugs like https://bugzilla.mozilla.org/show_bug.cgi?id=1482464

lib/TopSitesFeed.jsm
```diff
@@ -90,6 +92,10 @@ this.TopSitesFeed = class TopSitesFeed {
       this._currentSearchHostname = getShortURLForCurrentSearch();
       this.refresh({broadcast: true});
     }
+
+    if (topic === "browser-search-engine-modified") {
+      this.updateSearchShortcuts();
```
rlr: ^ Good enough or do I need to look at the topics too? @Mardak @k88hudson

rlr: `s/topic/data/`

lib/TopSitesFeed.jsm
```diff
@@ -360,6 +367,32 @@ this.TopSitesFeed = class TopSitesFeed {
...
+
+    if (!this._tippyTopProvider.initialized) {
+      await this._tippyTopProvider.init();
+    }
+
+    // Populate the state with available search shortcuts
+    const searchShortcuts = [];
+    SEARCH_SHORTCUTS.forEach(shortcut => {
+      // Only add the shortcut if the engine is available.
+      if (!Services.search.getEngines().find(e => e.identifier && e.identifier.match(shortcut.searchIdentifier))) {
```
rlr: @Mardak ^ included the guard for `identifier` 😉

rlr:
@Mardak @k88hudson I still need to spend some time writing more tests, especially for the UI Component. But if we are short on time I can land in a followup.

lib/TopSitesFeed.jsm
Mardak: Looks like we want "engine-added" and "engine-removed". There can be a lot of unrelated "engine-changed"

Mardak: Ah, we don't care about the user adding/removing custom engines as they won't have an identifier. And hiding a packaged one... can still show up in the list for now.

lib/TopSitesFeed.jsm
```diff
@@ -360,6 +362,32 @@ this.TopSitesFeed = class TopSitesFeed {
...
+    if (!this.store.getState().Prefs.values[SEARCH_SHORTCUTS_EXPERIMENT]) {
+      return;
+    }
+
+    if (!this._tippyTopProvider.initialized) {
+      await this._tippyTopProvider.init();
+    }
+
+    // Populate the state with available search shortcuts
+    const searchShortcuts = [];
```
k88hudson: what about maybe

``` js
const searchShortcuts = SEARCH_SHORTCUTS.reduce((result, shortcut) => {
  // Only add the shortcut if the engine is available.
  if (!Services.search.getEngines().find(e => e.identifier && e.identifier.match(shortcut.searchIdentifier))) {
    return result;
  }
  result.push(this._tippyTopProvider.processSite({...shortcut}));
  return result;
}, []);
```

😁

rlr: love it!

lib/TopSitesFeed.jsm
k88hudson: I think you need to `await new Promise(resolve => Services.search.init(resolve));` here to make sure search is ready here

common/Reducers.jsm
```diff
@@ -172,6 +180,8 @@ function TopSites(prevState = INITIAL_STATE.TopSites, action) {
       }
       newRows = prevState.rows.filter(site => action.data.url !== site.url);
       return Object.assign({}, prevState, {rows: newRows});
+    case at.UPDATE_SEARCH_SHORTCUTS:
+      return Object.assign({}, prevState, action.data);
```
k88hudson: I would do `return {...prevState, searchShortcuts: action.data.searchShortcuts}` here just so this is easier to read / figure out what's going on

PR #4320, opened 2018-08-10
Fix Bug 1482479 - Update SEARCH_SHORTCUTS and top_sites.json to support the 9 desired search urls

rlr:


lib/SearchShortcuts.jsm
```diff
@@ -6,10 +6,15 @@
 // List of sites we match against Topsites in order to identify sites
 // that should be converted to search Topsites
 const SEARCH_SHORTCUTS = [
-  {keyword: "@google", shortURL: "google", url: "https://google.com", searchIdentifier: /^google/},
+  {keyword: "@amazon", shortURL: "amazon", url: "https://amazon.com", searchIdentifier: /^amazon/},
   {keyword: "@baidu", shortURL: "baidu", url: "https://baidu.com", searchIdentifier: /^baidu/},
-  {keyword: "@yandex", shortURL: "yandex", url: "https://yandex.com", searchIdentifier: /^yandex/},
-  {keyword: "@amazon", shortURL: "amazon", url: "https://amazon.com", searchIdentifier: /^amazon/}
+  {keyword: "@bing", shortURL: "bing", url: "https://bing.com", searchIdentifier: /^bing/},
+  {keyword: "@duckduckgo", shortURL: "duckduckgo", url: "https://duckduckgo.com", searchIdentifier: /^duckduckgo/},
```
k88hudson: the `searchIdentifier` here should be `ddg` I think

Mardak: If I change locally to ddg, I see:
image

lib/SearchShortcuts.jsm
```diff
@@ -6,10 +6,15 @@
...
+  {keyword: "@ebay", shortURL: "ebay", url: "https://ebay.com", searchIdentifier: /^ebay/},
+  {keyword: "@google", shortURL: "google", url: "https://google.com", searchIdentifier: /^google/},
+  {keyword: "@twitter", shortURL: "twitter", url: "https://twitter.com", searchIdentifier: /^twitter/},
+  {keyword: "@wikipedia", shortURL: "wikipedia", url: "https://wikipedia.com", searchIdentifier: /^wikipedia/},
+  {keyword: "@yandex", shortURL: "yandex", url: "https://yandex.com", searchIdentifier: /^yandex/}
```
Mardak: Sounds like the ordering is important. mkaply says to get the ordering from Services.search.. I guess we can put them in order when we do the .reduce

lib/SearchShortcuts.jsm
```diff
@@ -6,10 +6,15 @@
...
+  {keyword: "@amazon", shortURL: "amazon", url: "https://amazon.com", searchIdentifier: /^amazon/},
   {keyword: "@baidu", shortURL: "baidu", url: "https://baidu.com", searchIdentifier: /^baidu/},
-  {keyword: "@yandex", shortURL: "yandex", url: "https://yandex.com", searchIdentifier: /^yandex/},
-  {keyword: "@amazon", shortURL: "amazon", url: "https://amazon.com", searchIdentifier: /^amazon/}
+  {keyword: "@bing", shortURL: "bing", url: "https://bing.com", searchIdentifier: /^bing/},
+  {keyword: "@duckduckgo", shortURL: "duckduckgo", url: "https://duckduckgo.com", searchIdentifier: /^duckduckgo/},
+  {keyword: "@ebay", shortURL: "ebay", url: "https://ebay.com", searchIdentifier: /^ebay/},
+  {keyword: "@google", shortURL: "google", url: "https://google.com", searchIdentifier: /^google/},
+  {keyword: "@twitter", shortURL: "twitter", url: "https://twitter.com", searchIdentifier: /^twitter/},
+  {keyword: "@wikipedia", shortURL: "wikipedia", url: "https://wikipedia.com", searchIdentifier: /^wikipedia/},
```
k88hudson: this should be `.org`

lib/SearchShortcuts.jsm
Mardak: `Services.search.getDefaultEngines()` gives only defaults and in the expected order. Maybe we should iterate on that and then test each regex..... ?

rlr:
ship it?

screen shot 2018-08-10 at 5 38 38 pm

lib/TopSitesFeed.jsm
```diff
@@ -375,12 +375,12 @@ this.TopSitesFeed = class TopSitesFeed {
 
     // Populate the state with available search shortcuts
     await new Promise(resolve => Services.search.init(resolve));
-    const searchShortcuts = SEARCH_SHORTCUTS.reduce((result, shortcut) => {
-      // Only add the shortcut if the engine is available.
-      if (!Services.search.getEngines().find(e => e.identifier && e.identifier.match(shortcut.searchIdentifier))) {
-        return result;
+    const searchShortcuts = Services.search.getDefaultEngines().reduce((result, engine) => {
+      if (!engine.identifier) { return result; }
```
Mardak: test failure. also, you can maybe just nest to avoid the early single line return

Mardak:
```
        ✖ should dispatch UPDATE_SEARCH_SHORTCUTS on updateSearchShortcuts
          Firefox 54.0.0 (Linux 0.0.0)
        Services.search.getDefaultEngines is not a function
```

PR #4321, opened 2018-08-10
Bug 1482519 - Separate custom search shortcut options from organic ones

k88hudson:
Also fixed a small bug where null top sites were crashing the add search
shortcuts editor.

### DIFF
--- a/content-src/components/TopSites/SearchShortcutsForm.jsx
+++ b/content-src/components/TopSites/SearchShortcutsForm.jsx
@@ -38,7 +38,10 @@ export class SearchShortcutsForm extends React.PureComponent {
     const shortcuts = [];
     const {rows, searchShortcuts} = props.TopSites;
     searchShortcuts.forEach(shortcut => {
-      shortcuts.push(Object.assign({}, shortcut, {isSelected: !!rows.find(({isPinned, searchTopSite, label}) => isPinned && searchTopSite && label === shortcut.keyword)}));
+      shortcuts.push({
+        ...shortcut,
+        isSelected: !!rows.find(row => row && row.isPinned && row.searchTopSite && row.label === shortcut.keyword)
+      });
     });
     this.state = {shortcuts};
   }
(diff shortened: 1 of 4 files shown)


content-src/components/TopSites/SearchShortcutsForm.jsx
```diff
@@ -38,7 +38,10 @@ export class SearchShortcutsForm extends React.PureComponent {
     const shortcuts = [];
     const {rows, searchShortcuts} = props.TopSites;
     searchShortcuts.forEach(shortcut => {
-      shortcuts.push(Object.assign({}, shortcut, {isSelected: !!rows.find(({isPinned, searchTopSite, label}) => isPinned && searchTopSite && label === shortcut.keyword)}));
+      shortcuts.push({
```
k88hudson: Found this while I was testing

lib/SearchShortcuts.jsm
```diff
@@ -8,15 +8,20 @@
 const SEARCH_SHORTCUTS = [
```
k88hudson: I had to take a slightly different approach here since ursula's patch would need to be updated if I included a new property with which to filter the list

content-src/components/TopSites/SearchShortcutsForm.jsx
k88hudson: Basically if there is `null` because of a pin space, this throws

rlr: Yay! Someday I’ll get all that new syntax into my muscle memory.

rlr: Yep. the empty slots bite again.

Mardak: Do we also need to fix this in onSaveButtonClick?

k88hudson: yes we do

PR #4322, opened 2018-08-10
fix(tests): Skip taskcluster eslint as we have coverage with travis

Mardak:
r?@k88hudson @piatra untested.. but.. should.. fix ?


Mardak:
Oh, ha. I suppose by creating the PR, taskcluster will test this for me! :)

PR #4324, opened 2018-08-11
Bug 1479458 - Fixing broadcast of updated data from cache.

ScottDowne:
To test.

1. Checkout aab3771a50215d97ed4f34c7056f46a0f943a822 which is the pr before the patch that caused the bug.

2. Open about:home enough to generate some spocs and a preload and a cache.

3. Checkout my branch.

4. Go to about:home

5. Should see rec.

If you want to see the initial bug reproduced, instead of checking out my branch on step 3, checkout  be0ccddd3e6f7d7adffa55b0440f8a1104286ca5 and do everything else the same, and then on 5 should see no recs.

### DIFF
--- a/lib/TopStoriesFeed.jsm
+++ b/lib/TopStoriesFeed.jsm
@@ -40,8 +40,6 @@ this.TopStoriesFeed = class TopStoriesFeed {
     try {
       const {options} = SectionsManager.sections.get(SECTION_ID);
       const apiKey = this.getApiKeyFromPref(options.api_key_pref);
-      // Set this to true if we're not loading from cache.
-      let shouldBroadcast = false;
       this.stories_endpoint = this.produceFinalEndpointUrl(options.stories_endpoint, apiKey);
       this.topics_endpoint = this.produceFinalEndpointUrl(options.topics_endpoint, apiKey);
       this.read_more_endpoint = options.read_more_endpoint;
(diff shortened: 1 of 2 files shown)


ScottDowne:
I'm not sure how to read taskcluster or sure what it's actually testing. Think I need help in fixing it :)

piatra:
Issue running eslint not with your code. I've merged a PR that removes linting from our taskcluster CI (since it's covered by Travis) and retriggered your build for another test.

piatra:
oh. @ScottDowne you will need to pull the latest changes and update your PR otherwise you're still triggering the eslint check.

ScottDowne:
@piatra got it, thanks, updating ow.

lib/TopStoriesFeed.jsm
```diff
@@ -57,17 +55,14 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
       if (this.storiesLastUpdated === 0) {
-        shouldBroadcast = true;
         await this.fetchStories();
       }
       if (this.topicsLastUpdated === 0) {
-        shouldBroadcast = true;
         await this.fetchTopics();
       }
-      this.doContentUpdate(shouldBroadcast);
+      this.doContentUpdate(true);
```
ScottDowne: Essentially, this is how the old code worked, and I misinterpreted it when I was moving some stuff around.

PR #3588, opened 2017-09-23
fix(styles): Update search focus to Blue-50 shades of Grey-90

Mardak:
Fix #3583. r?@sarracini Removed the confusing / duplicate `search-input` mixin. Also removed unnecessary `z-index`

screen shot 2017-09-23 at 9 36 47 am

PR #4325, opened 2018-08-13
Fix Bug 1482871 - remove blue flash when clicking on div.search-wrapper

rlr:
not sure why that was there in the first place?

### DIFF
--- a/content-src/components/Search/_Search.scss
+++ b/content-src/components/Search/_Search.scss
@@ -51,7 +51,6 @@
     box-shadow: $shadow-secondary, 0 0 0 1px $black-25;
   }
 
-  &:active input,
   input:focus {
     border: $input-border-active;
     box-shadow: var(--newtab-textbox-focus-boxshadow);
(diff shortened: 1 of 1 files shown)


content-src/components/Search/_Search.scss
```diff
@@ -51,7 +51,6 @@
     box-shadow: $shadow-secondary, 0 0 0 1px $black-25;
   }
 
-  &:active input,
```
Mardak: This was added in #3588 along with a related style:
https://github.com/mozilla/activity-stream/blob/6a2252ccf8e0aea02857db1fb7848802ead4c491/content-src/components/Search/_Search.scss#L84-L85

This is so that when the user is clicking on the search button, it adds the blue border/shadow and the depressed button.

rlr: I guess i’ll use the new div.search-inner-wrapper

rlr: Fixed. Now it seems to behave just like release. @Mardak

PR #4326, opened 2018-08-13
Fix Bug 1482629 - Compare search topsites against search engine _internalAliases

piatra:


k88hudson:
I was thinking maybe changes to `NewTabUtils` might be necessary but it doesn't look like that's the case.

PR #4330, opened 2018-08-13
Fix Bug 1482853 - fix width of search bar in fixed header

rlr:


content-src/styles/_variables.scss
```diff
@@ -77,6 +77,11 @@ $icon-size: 16px;
 $smaller-icon-size: 12px;
 $larger-icon-size: 32px;
 
+$searchbar-width-smallest: $grid-unit * 2 + $base-gutter * 1;
```
k88hudson: This seems a little small but it's still useable, so I guess it's fine

PR #4331, opened 2018-08-13
Fix Bug 1482895 - scroll the modal on short windows

rlr:
screen shot 2018-08-13 at 1 08 30 pm


content-src/components/TopSites/_TopSites.scss
```diff
@@ -282,6 +282,8 @@ $half-base-gutter: $base-gutter / 2;
     box-shadow: $shadow-secondary;
     left: 0;
     margin: 0 auto;
+    max-height: calc(100% - 40px);
+    overflow: auto;
```
k88hudson: This is causing an overflow on the x-axis as well, is that perhaps unintended?

PR #4334, opened 2018-08-13
Fix Bug 1482246 - Correctly report mochitest failures

piatra:
I'm not sure if this is a bug or not but running `./mach test` or `./mach mochitest` seems to return 0 regardless of the success/failure of the actual test.
The only workaround I've found is outputing the logs in a file and grepping for the `TEST-UNEXPECTED` string, funny enough this method is also mentioned in a bug [about returning the correct status code from mochitests](https://bugzilla.mozilla.org/show_bug.cgi?id=875382#c3)
f? @Mardak

### DIFF
--- a/mochitest.sh
+++ b/mochitest.sh
@@ -15,9 +15,9 @@ cd /activity-stream && npm install . && npm run buildmc
 cd /mozilla-central && ./mach build \
   && ./mach test browser_parsable_css \
   && ./mach lint -l codespell browser/components/newtab \
-  && ./mach test browser/components/newtab/test/browser --headless \
+  && ./mach test browser/components/newtab/test/browser \
   && ./mach test browser/components/newtab/test/xpcshell \
-  && ./mach test browser/components/preferences/in-content/tests/browser_hometab_restore_defaults.js --headless \
-  && ./mach test browser/components/preferences/in-content/tests/browser_newtab_menu.js --headless \
-  && ./mach test browser/components/enterprisepolicies/tests/browser/browser_policy_set_homepage.js --headless \
-  && ./mach test browser/components/preferences/in-content/tests/browser_search_subdialogs_within_preferences_1.js --headless
+  && ./mach test browser/components/preferences/in-content/tests/browser_hometab_restore_defaults.js \
+  && ./mach test browser/components/preferences/in-content/tests/browser_newtab_menu.js \
+  && ./mach test browser/components/enterprisepolicies/tests/browser/browser_policy_set_homepage.js \
+  && ./mach test browser/components/preferences/in-content/tests/browser_search_subdialogs_within_preferences_1.js
--- a/mochitest.sh
+++ b/mochitest.sh
@@ -13,19 +13,12 @@ cd /activity-stream && npm install . && npm run buildmc
 
 # Build latest m-c with Activity Stream changes
 cd /mozilla-central && ./mach build \
-  && ./mach test browser_parsable_css \
   && ./mach lint -l codespell browser/components/newtab \
-  && ./mach test browser/components/newtab/test/browser \
-  browser/components/newtab/test/xpcshell \
+  && ./mach test --log-tbpl test_run_log \
+  browser_parsable_css \
+  browser/components/newtab \
   browser/components/preferences/in-content/tests/browser_hometab_restore_defaults.js \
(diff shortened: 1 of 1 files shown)


mochitest.sh
```diff
@@ -15,9 +15,17 @@ cd /activity-stream && npm install . && npm run buildmc
 cd /mozilla-central && ./mach build \
   && ./mach test browser_parsable_css \
```
Mardak: Probably include this test with the rest now

mochitest.sh
```diff
@@ -15,9 +15,17 @@ cd /activity-stream && npm install . && npm run buildmc
...
   && ./mach lint -l codespell browser/components/newtab \
-  && ./mach test browser/components/newtab/test/browser --headless \
-  && ./mach test browser/components/newtab/test/xpcshell \
-  && ./mach test browser/components/preferences/in-content/tests/browser_hometab_restore_defaults.js --headless \
-  && ./mach test browser/components/preferences/in-content/tests/browser_newtab_menu.js --headless \
-  && ./mach test browser/components/enterprisepolicies/tests/browser/browser_policy_set_homepage.js --headless \
-  && ./mach test browser/components/preferences/in-content/tests/browser_search_subdialogs_within_preferences_1.js --headless
+  && ./mach test browser/components/newtab/test/browser \
```
Mardak: Looks like mochitests failing have exit code 0.. probably file a bug for this

mochitest.sh
```diff
@@ -15,9 +15,17 @@ cd /activity-stream && npm install . && npm run buildmc
...
+  browser/components/preferences/in-content/tests/browser_newtab_menu.js \
+  browser/components/enterprisepolicies/tests/browser/browser_policy_set_homepage.js \
+  browser/components/preferences/in-content/tests/browser_search_subdialogs_within_preferences_1.js --log-tbpl test_run_log
+
+# Check test log for any error reports
+if [ $(cat test_run_log | grep "TEST-UNEXPECTED" | wc -l) -gt 0 ]
+then
+	false
+else
+	true
```
Mardak: This could be simplified to just `&& ! grep -q TEST-UNEXPECTED test_run_log` (chain the && unless it already failed a non-mochitest, negate the exit of `grep` where normally if it matches, it exits 0 and exits 1 if no match)

mochitest.sh
```diff
@@ -15,9 +15,17 @@ cd /activity-stream && npm install . && npm run buildmc
...
   && ./mach test browser_parsable_css \
   && ./mach lint -l codespell browser/components/newtab \
-  && ./mach test browser/components/newtab/test/browser --headless \
-  && ./mach test browser/components/newtab/test/xpcshell \
-  && ./mach test browser/components/preferences/in-content/tests/browser_hometab_restore_defaults.js --headless \
-  && ./mach test browser/components/preferences/in-content/tests/browser_newtab_menu.js --headless \
-  && ./mach test browser/components/enterprisepolicies/tests/browser/browser_policy_set_homepage.js --headless \
-  && ./mach test browser/components/preferences/in-content/tests/browser_search_subdialogs_within_preferences_1.js --headless
+  && ./mach test browser/components/newtab/test/browser \
+  browser/components/newtab/test/xpcshell \
```
Mardak: Oh, and I guess just run from the parent `browser/components/newtab` to include both "test/browser" and "test/xpcshell" now that no headless

mochitest.sh
```diff
@@ -15,9 +15,17 @@ cd /activity-stream && npm install . && npm run buildmc
...
+  browser/components/preferences/in-content/tests/browser_hometab_restore_defaults.js \
+  browser/components/preferences/in-content/tests/browser_newtab_menu.js \
+  browser/components/enterprisepolicies/tests/browser/browser_policy_set_homepage.js \
+  browser/components/preferences/in-content/tests/browser_search_subdialogs_within_preferences_1.js --log-tbpl test_run_log
```
Mardak: nit: move the `--log-tbpl` to the first line `./mach test --log-tbpl` and we can chain the individual test/paths below it to not hide this log option

mochitest.sh
Mardak: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1483003

If it gets fixed, we could get rid of the grepping and just keep all the `./mach test … … …` in a single command, as that prints the summary together as well as running all the tests.

PR #4338, opened 2018-08-14
Fix Bug 1476684 - Onboarding shouldn't be affected by snippets experiment prefs

sarracini:
feedback needed before i fix all the tests... @k88hudson is this the right approach for this?

### DIFF
--- a/content-src/components/Base/Base.jsx
+++ b/content-src/components/Base/Base.jsx
@@ -114,7 +114,7 @@ export class _Base extends React.PureComponent {
     const {initialized} = App;
 
     const prefs = props.Prefs.values;
-    if ((prefs.asrouterExperimentEnabled || prefs.asrouterOnboardingCohort > 0) && window.location.hash === "#asrouter") {
+    if (prefs.asrouterExperimentEnabled && window.location.hash === "#asrouter") {
       return (<ASRouterAdmin />);
     }
 
@@ -124,7 +124,12 @@ export class _Base extends React.PureComponent {
 
     // Until we can delete the existing onboarding tour, just hide the onboarding button when users are in
     // the new simplified onboarding experiment. CSS hacks ftw
-    if (prefs.asrouterOnboardingCohort > 0) {
+    let isOnboardingEnabled = false;
+    try {
+      isOnboardingEnabled = JSON.parse(prefs["asrouter.messageProviders"]).find(i => i.id === "onboarding").enabled;
+    } catch (e) {}
+
+    if (isOnboardingEnabled) {
       global.document.body.classList.add("hide-onboarding");
     }
 
--- a/content-src/lib/asroutercontent.js
+++ b/content-src/lib/asroutercontent.js
@@ -1,0 +1,13 @@
+export function enableASRouterContent(store, asrouterContent) {
+  // Enable asrouter content
+  store.subscribe(() => {
+    const state = store.getState();
+    if (state.Prefs.values.asrouterExperimentEnabled && !asrouterContent.initialized) {
+      asrouterContent.init();
+    } else if (!state.Prefs.values.asrouterExperimentEnabled && asrouterContent.initialized) {
+      asrouterContent.uninit();
+    }
+  });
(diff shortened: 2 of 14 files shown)


content-src/components/Base/Base.jsx
```diff
@@ -114,17 +116,29 @@ export class _Base extends React.PureComponent {
...
     if (!props.isPrerendered && !initialized) {
       return null;
     }
-
     // Until we can delete the existing onboarding tour, just hide the onboarding button when users are in
     // the new simplified onboarding experiment. CSS hacks ftw
-    if (prefs.asrouterOnboardingCohort > 0) {
+    let isOnboardingEnabled = false;
+    try {
+      isOnboardingEnabled = JSON.parse(prefs["asrouter.messageProviders"]).find(i => i.id === "onboarding").enabled;
```
k88hudson: could we avoid all this parsing by converting this pref to a `getPref` function?

k88hudson: Also this could presumably false if onboarding is not defiled at all, no?

content-src/components/Base/Base.jsx
```diff
@@ -114,17 +116,29 @@ export class _Base extends React.PureComponent {
...
 
     const prefs = props.Prefs.values;
-    if ((prefs.asrouterExperimentEnabled || prefs.asrouterOnboardingCohort > 0) && window.location.hash === "#asrouter") {
+
+    // Enable asrouter content
+    if (prefs.asrouterExperimentEnabled && !asrouterContent.initialized) {
+      asrouterContent.init();
+    } else if (!prefs.asrouterExperimentEnabled && asrouterContent.initialized) {
+      asrouterContent.uninit();
+    }
```
k88hudson: Personally I am not a big fan of side effects happening in `render` functions; would I would probably do here is move this to `activity-stream.jsx` and do this logic in `store.subscribe`
```

**Comment by k88hudson** [comment]
Yep, this looks good overall! I just had some minor comments about where to put stuff and the way the pref works

File: content-src/components/Base/Base.jsx (hunk 1 above)
Comment (sarracini) [reply content-src/components/Base/Base.jsx]: `getPref` i think can only read booleans, integers, and strings, but we'd still have to parse the JSON value to get the `enabled` field, no?
And yes, it would stay false if it's not defined, which is i think is fine, because we'd want to fall back on the old onboarding experience in that case

File: lib/ASRouterTargeting.jsm [hunk 3]
```diff
@@ -153,7 +153,14 @@ const TargetingGetters = {
 
   // Temporary targeting function for the purposes of running the simplified onboarding experience
   get isInExperimentCohort() {
-    return Services.prefs.getIntPref(ONBOARDING_EXPERIMENT_PREF, 0);
+    const allProviders = Services.prefs.getStringPref(ONBOARDING_MESSAGE_PROVDIER_EXPERIMENT_PREF, "");
+    try {
+      const {cohort} = JSON.parse(allProviders).find(i => i.id === "onboarding");
+      return (typeof cohort === "number" ? cohort : 0);
```
sarracini: we don't _have_ to check that it's a number here, since "1" could still technically be valid, but if we want to keep consistent return types we should do this check. up to you @k88hudson

test/unit/asrouter/ASRouterTargeting.test.js
```diff
@@ -98,3 +98,26 @@ describe("ASRouterTargeting#isBelowFrequencyCap", () => {
     });
   });
 });
+
+describe("ASRouterTargeting#isInExperimentCohort", () => {
```
sarracini: we also don't really need this test, since i wrote a mochitest too that does basically this... but we can never have too many tests!!!

content-src/lib/asroutercontent.js
```diff
@@ -0,0 +1,13 @@
+export function enableASRouterContent(store, asrouterContent) {
```
sarracini: was easier to make this it's own js file for testing

PR #4339, opened 2018-08-14
bug 1478755 - Ensure spocs load without a flash of rec via a slight r…

ScottDowne:
Steps to test:

1. Open the browser with the Profile manager.
2. Create a new Profile, but DO NOT open it.
3. Navigate to the Profile's location on the computer.
(Win - C:\Users\<username>\AppData\Roaming\Mozilla\Firefox\Profiles\
Mac - Users/<username>/Library/Application Support/Firefox/Profiles/
Linux - /home/<username>/.mozilla/firefox)
Create a file named user.js and open it with a text editor.
Write :
```
user_pref("browser.newtabpage.activity-stream.feeds.section.topstories.options", '{"api_key_pref":"extensions.pocket.oAuthConsumerKey","hidden":false,"provider_icon":"pocket","provider_name":"Pocket","read_more_endpoint":"https://getpocket.com/explore/trending?src=fx_new_tab","stories_endpoint":"file:///home/scott/spoc.json","stories_referrer":"https://getpocket.com/recommendations","topics_endpoint":"https://getpocket.cdn.mozilla.net/v3/firefox/trending-topics?version=2&consumer_key=$apiKey&locale_lang=en-US","show_spocs":true,"personalized":true}');
```
4. Start the browser with the newly created profile.

That's the steps for the original bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1478755

What's actually causing the flash is the fact that `NEW_TAB_REHYDRATED` is happening after render of sections.

I'm not a fan of this solution, tbh, but it was approved by Nick Chapman, and much much easier than a refactor (which I believe is doable with a bit more work)

Thoughts?

### DIFF
--- a/content-src/components/Sections/Sections.jsx
+++ b/content-src/components/Sections/Sections.jsx
@@ -18,6 +18,11 @@ function getFormattedMessage(message) {
 }
 
 export class Section extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {delayThirdCard: props.id === "topstories"};
+  }
+
   get numRows() {
     const {rowsPref, maxRows, Prefs} = this.props;
     return rowsPref ? Prefs.values[rowsPref] : maxRows;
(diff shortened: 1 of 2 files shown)


ScottDowne:
Closing this in favour of this https://github.com/mozilla/activity-stream/pull/4356

PR #4290, opened 2018-08-07
Bug 1480513: Add telemetry for search shortcut

ncloudioj:
This documents the telemetry part for [search shortcuts](https://bugzilla.mozilla.org/show_bug.cgi?id=1479806) in Top Sites of Activity Stream.

@fmarier Could you take a look at this please? The data review request was attached at [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1480513#c2)

PR #4340, opened 2018-08-14
Fix Bug 1483297 - Add search shortcut count to session stats

ncloudioj:


ncloudioj:
@fmarier This is a follow up metric of #4290, which allows us to compute the overall click through rate and block rate for the search short feature.

The data review form was attached [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1483297#c2)

ncloudioj:
datareview+ at https://bugzilla.mozilla.org/show_bug.cgi?id=1483297#c3

PR #4342, opened 2018-08-14
Fix Bug 1483255: don't allow parallel TopSites refreshes

AdamHillier:
This turned out to be a pretty ugly race involving both the pinned cache and parallel calls to `refresh` in TopSitesFeed. Not allowing calls to `refresh` to execute in parallel fixes the problem and should also reduce the number of computationally intensive refresh calls we run.

### DIFF
--- a/lib/TopSitesFeed.jsm
+++ b/lib/TopSitesFeed.jsm
@@ -68,6 +68,8 @@ this.TopSitesFeed = class TopSitesFeed {
     this.pinnedCache = new LinksCache(NewTabUtils.pinnedLinks, "links",
       [...CACHED_LINK_PROPS_TO_MIGRATE, ...PINNED_FAVICON_PROPS_TO_MIGRATE]);
     PageThumbs.addExpirationFilter(this);
+    this._refreshRequestQueue = [];
+    this._refreshInProgress = false;
   }
 
   init() {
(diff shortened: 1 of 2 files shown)


Mardak:
How is the race causing problems? There's multiple unpin actions that all expire the pinned cache and request a refresh. But the last unpin should still expire then request the refresh with all the searches unpinned... ?

AdamHillier:
You can verify that this fixes things: without the patch if you have pinned e.g. just Twitter and Wikipedia, then unpin them in one go, it looks like Wikipedia doesn't get unpinned. In fact it did get unpinned - restart the browser and it goes away. With the patch it goes away immediately, as you'd expect it to.

Mardak:
Ah, I do see the latter UNPIN's refresh finish first then get overwritten by the former's UNPIN. We should probably fix this generally, but it seems a bit dangerous to make this type of change late in the beta cycle.

@rlr, could we make PIN/UNPIN actions take a single object or an array of objects to pin/unpin?

rlr:
> @rlr, could we make PIN/UNPIN actions take a single object or an array of objects to pin/unpin? (Or always an array, potentially single item.)

Yeah. I this only affects the "Add Search Engine" modal, I would suggest we just create a new action type and send the checked search engines to the feed and move all the pin/unpin logic from the UI component to the feed.

Mardak:
As in from `onSaveButtonClick`
https://github.com/mozilla/activity-stream/blob/b2d7f56e615ac05a037176e07abed13299f6632e/content-src/components/TopSites/SearchShortcutsForm.jsx#L66-L73
```
dispatch({
  type: UPDATE_SEARCH_ENGINES
  data: [
    {keyword: "…", selected: true/false},
    …
  ]
})
```

??

rlr:
yeah, I probably should've done that instead of trying to re-use the existing pin/unpin actions.

rlr:
I guess we dont need this... For now at least....

PR #4343, opened 2018-08-15
Bug 1483562 - Use object for button_action in ASR

k88hudson:
To test this, just open the onboarding panels via `about:newtab#asrouter` and make sure the buttons still work.

Changes `button_action` and `button_params` to be a single `button_action`


k88hudson:
@piatra doing it as a follow-up (or just a separate PR) seems more logical to me as they are separate issues, however if it's logistically easier to put them together just go for it.

PR #4353, opened 2018-08-21
Fix Bug 1471391: Create CFR doorhanger

AdamHillier:


lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,241 @@
...
+   */
+  _handleClick(event) {
+    const browser = this.window.gBrowser.selectedBrowser;
+    if (!RecommendationMap.has(browser)) {
+      // There's no recommendation for this browser, so the user shouldn't have
+      // been able to click
+      this.hide();
+      return;
+    }
+    const {message, name, icon, action} = RecommendationMap.get(browser);
```
k88hudson: I think it would be good to pass this in as a single `messageContent` object that follows the [schema](https://github.com/mozilla/activity-stream/blob/master/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json), so you're accessing `messageContent.description` etc.

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,241 @@
...
+    // A hacky way of setting the popup anchor outside the usual url bar icon box
+    // See https://searchfox.org/mozilla-central/rev/847b64cc28b74b44c379f9bff4f415b97da1c6d7/toolkit/modules/PopupNotifications.jsm#42
+    browser.cfrpopupnotificationanchor = this.hbox;
+
+    let mainAction;
+    if (action.type === ACTION_TYPE_INSTALL_ADDON) {
+      mainAction = {
+        label: "Add to Firefox",
+        accessKey: "_", // FIXME
+        callback: () => this._popupInstallAddon(action.data)
```
k88hudson: I don't think it's necessary to implement different handlers here, but rather a general function that dispatches an action back to activity stream (and activity stream can handle the logic). i.e.

```js
mainAction = {
  label: messageContent.buttons.primary.label
  callback: () => this._dispatchToActivityStream(messageContent.buttons.primary.action)
}
```

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,241 @@
...
+      }
+    }
+  }
+
+  // This is called when the popup closes as a result of clicking "No thanks"
+  _popupDismissed() {
+    this._clearScheduledStateChanges();
+    if (this.hbox.getAttribute("state") === "expanded") {
+      this.hbox.setAttribute("state", "collapsed");
+    }
```
k88hudson: you could pass in `messageContent.buttons.secondary.action` to this function and then call `this._dispatchToActivityStream` on it

k88hudson: Also, `_popupStateChange` and `_popupDismissed` should have some shared code

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,241 @@
...
+    // The recommendation should remain either collapsed or expanded while the
+    // doorhanger is showing
+    this._clearScheduledStateChanges();
+
+    // A hacky way of setting the popup anchor outside the usual url bar icon box
+    // See https://searchfox.org/mozilla-central/rev/847b64cc28b74b44c379f9bff4f415b97da1c6d7/toolkit/modules/PopupNotifications.jsm#42
+    browser.cfrpopupnotificationanchor = this.hbox;
+
+    let mainAction;
+    if (action.type === ACTION_TYPE_INSTALL_ADDON) {
```
k88hudson: No need to check this if you use a generic dispatch handler (see commend below at line 150).

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,241 @@
...
+ */
+const PageActionMap = new WeakMap();
+
+/**
+ * We need one PageAction for each window
+ */
+class PageAction {
+  constructor(win) {
+    this.window = win;
+    this.hbox = win.document.getElementById("contextual-feature-recommendation");
```
k88hudson: Could you give this a name that reflects its specific function rather than just `hbox`?

k88hudson:
For testing, I was thinking something like this: 
```diff
diff --git a/lib/ASRouter.jsm b/lib/ASRouter.jsm
index 8a360dda..b8a34409 100644
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -520,6 +520,11 @@ class _ASRouter {
     return impressions;
   }
 
+  _testDoorhanger({message, target}) {
+    // TODO
+    // create a CFRPageAction here given the current target 
+  }
+
   /**
    * getLongestPeriod
    *
@@ -610,7 +615,11 @@ class _ASRouter {
     await this.setState({lastMessageId: id});
     const newMessage = this.getMessageById(id);
 
-    await this._sendMessageToTarget(newMessage, target, force, action.data);
+    if (newMessage && newMessage.template === "doorhanger") {
+      this._testDoorhanger({message: newMessage, target});
+    } else {
+      await this._sendMessageToTarget(newMessage, target, force, action.data);
+    }
   }
 
   async blockMessageById(idOrIds) {

```

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,225 @@
...
+  // This is called when the popup closes as a result of interaction _outside_
+  // the popup, e.g. by hitting <esc>
+  _popupStateChange(state) {
+    if (["dismissed", "removed"].includes(state)) {
+      this._collapse();
+    }
+  }
+
+  _dispatchToActivityStream(action) {
+    // TODO
```
k88hudson: This is actually going to be `dispatchToASRouter` now that I think about it. You could actually pass this in as a function in the constructor

AdamHillier:
You should be able to test with
```js
ChromeUtils.import("resource://activity-stream/lib/CFRPageActions.jsm");
CFRPageActions.addRecommendation(gBrowser.selectedBrowser, "www.amazon.com", {notification_text: "Recommendation", addon: {icon: "https://addons.cdn.mozilla.net/user-media/addon_icons/337/337359-64.png?modified=mcrushed"}, text: "Amazon Assistant helps you make better shopping decisions by showing product comparisons at thousands of retail sites.", buttons: {primary: {label: "Add", accessKey: "a"}, secondary: {label: "No thanks", accessKey: "n"}}});
```

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * A WeakMap from browsers to {host, recommendation} pairs. Recommendations are
+ * defined in the ExtensionDoorhanger.schema.json.
+ *
+ * A recommendation is specific to a browser and host and is active until the
+ * given browser is closed or the user navigates (within that browser) away from
+ * the host.
+ */
+const RecommendationMap = new WeakMap();
```
k88hudson: it might be worth making these properties of `CFRPageActions` for future testing purposes

lib/CFRPageActions.jsm
AdamHillier: I've renamed it but haven't passed it into the constructor, though I can do if you'd prefer?

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,232 @@
...
+    // A hacky way of setting the popup anchor outside the usual url bar icon box
+    // See https://searchfox.org/mozilla-central/rev/847b64cc28b74b44c379f9bff4f415b97da1c6d7/toolkit/modules/PopupNotifications.jsm#42
+    browser.cfrpopupnotificationanchor = this.container;
+
+    const {primary, secondary} = recommendation.buttons;
+
+    const mainAction = {
+      label: primary.label,
+      accessKey: primary.accessKey,
+      callback: () => this._dispatchToActivityStream(primary.action)
```
k88hudson: This should be` _dispatchToASRouter`

AdamHillier: Oh oops thanks

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,232 @@
...
+
+      // After one second, expand
+      this._expand(DELAY_BEFORE_EXPAND_MS);
+
+      // Five seconds later, collapse again
+      this._collapse(DELAY_BEFORE_EXPAND_MS + DURATION_OF_EXPAND_MS);
+    }
+  }
+
+  hide() {
```
k88hudson: It might be good to make this `async` for testing purposes, i.e. resolve `this_collapse` when the timeout has completed

lib/CFRPageActions.jsm
AdamHillier: I haven't actually moved the definitions otherwise there are circular no-use-before-defined issues but I added references to them to `CFRPageActions`

lib/CFRPageActions.jsm
```diff
@@ -0,0 +1,229 @@
...
+
+    // The recommendation should remain either collapsed or expanded while the
+    // doorhanger is showing
+    this._clearScheduledStateChanges();
+
+    // A hacky way of setting the popup anchor outside the usual url bar icon box
+    // See https://searchfox.org/mozilla-central/rev/847b64cc28b74b44c379f9bff4f415b97da1c6d7/toolkit/modules/PopupNotifications.jsm#42
+    browser.cfrpopupnotificationanchor = this.container;
+
+    const {primary, secondary} = recommendation.buttons;
```
k88hudson: all of these properties actually properties of `recommendation.content`, not `recommendation` itself

PR #4355, opened 2018-08-21
bug 1483655 - Add a new explainer to the Recommended by Pocket section

ScottDowne:
Comps: https://mozilla.invisionapp.com/share/AMNI4YWRJK6#/screens/314216120_pocket-Section-Disclaimer

To test:

1. Go to about:config and change browser.newtabpage.activity-stream.pocketCta to this: `{"cta_button":"","cta_text":"","cta_url":"","use_cta":true}` by making those strings empty, it by default uses the l10n'd version. If we run a shield study on the copy here, we would need to target a certain locale with it but that to my knowledge is fine.

1. Go to about:home and you should see:
loggedout

2. now log into pocket, and you should see the same thing, but topics replacing the get pocket cta.

You can also test the copy in the cta using the pref "pocketCta"


lib/ActivityStream.jsm
```diff
@@ -77,6 +77,15 @@ const PREFS_CONFIG = new Map([
     title: "Show sponsored cards in spoc experiment (show_spocs in topstories.options has to be set to true as well)",
     value: true
   }],
+  ["pocketCta", {
```
piatra: I don't see where this value gets set?

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -145,12 +112,14 @@ export class _CollapsibleSection extends React.PureComponent {
     this.setState({showContextMenu});
   }
 
+  onCtaClick(e) {
+    e.stopPropagation();
```
k88hudson: why the stop propagation?

locales/en-US/strings.properties
```diff
@@ -143,6 +143,10 @@ pocket_read_more=Popular Topics:
 # LOCALIZATION NOTE (pocket_read_even_more): This is shown as a link at the
 # end of the list of popular topic links.
 pocket_read_even_more=View More Stories
+pocket_more_reccommendations=More Recommendations
+pocket_learn_more=Learn More
+pocket_cta_button=Get Pocket
+pocket_cta_text=Save the stories you love in Pocket, and fuel your mind with fascinating reads.
```
ScottDowne: @Mardak I'm told I need to reach out to you about making these strings marked for marketing. I'm told some sort of comment or annotation needs to be added?

content-src/components/CollapsibleSection/CollapsibleSection.jsx
ScottDowne: It's to ensure when the link is clicked it doesn't also collapse the section.

I'm happy to fix this in a way where it doesn't use stop propagation though.

content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
```diff
@@ -0,0 +1,25 @@
...
+    border-radius: 4px;
+    cursor: pointer;
+    max-width: $max-button-width;
+    min-height: $min-button-height;
+    padding: 0 6px;
+    display: inline-flex;
+    justify-content: center;
+    align-items: center;
+    font-size: 11px;
+    margin-right: 10px;
```
k88hudson: you want to use `margin-inline-end` here so it works on RTL

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -159,10 +123,17 @@ export class _CollapsibleSection extends React.PureComponent {
...
         <div className="section-top-bar">
           <h3 className="section-title">
-            <span className="click-target" onClick={this.onHeaderClick}>
-              {this.renderIcon()}
-              {getFormattedMessage(title)}
-              {isCollapsible && <span className={`collapsible-arrow icon ${collapsed ? "icon-arrowhead-forward-small" : "icon-arrowhead-down-small"}`} />}
+            <span className="click-target-container">
+              <span className="click-target" onClick={this.onHeaderClick}>
+                {this.renderIcon()}
+                {getFormattedMessage(title)}
```
k88hudson: Why not just use `<FormattedMessage id={title} />` here?

locales/en-US/strings.properties
Mardak: I'm not really sure what specially needs to be done for marketing? Unless you're asking if these strings are finalized for use in product. @aaronrbenson has worked with people to get approval on specific wording of strings.

content-src/components/CollapsibleSection/CollapsibleSection.jsx
ScottDowne: Agreed, I can update it. The why is simply I moved it and didn't think much of it ;)

locales/en-US/strings.properties
nchapman: @Mardak The marketing team would like to provide the translations for this string, but we aren't sure what the right process is. For example we have the string we would like to use in German but don't exactly know how to manage putting it in the right place.

Mardak: Ah... @flodolo ? How do we get specific german marketing strings into pontoon?

ScottDowne: @flodolo thanks, super helpful. The uplift you mentioned is just uplifting the changes to the .properties file, right?

Also, aside from filing a bug/setting up a call, is there something that needs to be added to the code here and now? I was under the impression there is something I would still need to do even on top of a bug and call?

flodolo: Nothing to add here, because it would show for 100 languages, and confuse them (I assume this string will be only used for German, but it's still exposed to all language we ship Firefox in).

As for uplift: it would require a bit more, given how Activity Stream is built. They'd need to build an update version, possibly only for beta or release, and uplift that code.

content-src/components/CollapsibleSection/CollapsibleSection.jsx
ScottDowne: I think `getFormattedMessage` is intended. It looks like it does checks to ensure if title is a string, it just uses the string, if it's an object, it uses `FormattedMessage` and I'm not yet sure about why title can sometimes be a string and sometimes be an object, but I'm guessing we should leave it as is.

ScottDowne:
@k88hudson fixed the ltr rtl issues. I also did some work to make it display a bit better on the smallest media query.

k88hudson:
I'm really sorry, could you fix this last alignment issue? After that this is good to go

image

common/Reducers.jsm
```diff
@@ -386,6 +390,18 @@ function Pocket(prevState = INITIAL_STATE.Pocket, action) {
   switch (action.type) {
     case at.POCKET_WAITING_FOR_SPOC:
       return {...prevState, waitingForSpoc: action.data};
+    case at.POCKET_LOGGED_IN:
+      return {...prevState, isUserLoggedIn: !!action.data};
+    case at.POCKET_CTA:
+      return {
+        ...prevState,
+        pocketCta: {
+          ctaButton: action.data.cta_button,
```
k88hudson: why use camel case v.s. underscores differently?

ScottDowne: Initially, was just trying to match the case in the object keys on browser.newtabpage.activity-stream.feeds.section.topstories.options.

Once I started to build it out I decided I wanted to put all the values onto the state object `ctaButton` anyway, and at this point it made sense to keep everything in camel case.

So I guess just trying to match case without going too far out of my way.

k88hudson: 👍

PR #4357, opened 2018-08-22
Bug 1485345 - Add timestamp targeting

piatra:
This adds a `currentTimestamp` attribute for targeting. Allows ASRouter messages to filter by time, also allows for more complicated JEXL expressions like "profile age".

I also updated some of the docs `publish_start|end` is not something we support.

And I've moved our remote snippet endpoint to fetch from the snippets cdn. We can now validate the integration of the two services. Also the previous endpoint was using a schema we no longer support (link are now only rendered as part of Fluent).

### DIFF
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -67,6 +67,9 @@ const TargetingGetters = {
       update: settings.update
     };
   },
+  get timestamp() {
+    return new Date(new Date().getTime());
+  },
   get profileAgeCreated() {
     return new ProfileAge(null, null).created;
   },
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -67,7 +67,7 @@ const TargetingGetters = {
       update: settings.update
     };
   },
-  get currentTimestamp() {
+  get currentDate() {
     return new Date();
   },
   get profileAgeCreated() {
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -67,7 +67,7 @@ const TargetingGetters = {
       update: settings.update
     };
   },
-  get timestamp() {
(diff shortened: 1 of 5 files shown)


lib/ASRouterTargeting.jsm
```diff
@@ -67,6 +67,9 @@ const TargetingGetters = {
       update: settings.update
     };
   },
+  get currentTimestamp() {
+    return new Date(new Date().getTime());
```
sarracini: Why do we need to nest these? It looks like just doing `new Date()` returns the same as this, and when i subbed it out in the jexl expression it seems to still return the same result, and it passes the tests too

piatra: I wanted to return a date object not a timestamp.

Mardak:
```js
new Date()
Date 2018-08-25T23:02:00.134Z
```

test/browser/browser_asrouter_targeting.js
```diff
@@ -110,6 +110,16 @@ add_task(async function checkProfileAgeReset() {
     "should select correct item by profile age reset");
 });
 
+add_task(async function checkCurrentTimestamp() {
+  let message = {id: "foo", targeting: `currentTimestamp < '${new Date(new Date().getTime() + 1000)}'|date`};
```
sarracini: you should be able to do `new Date(Date.now() + 1000)` here instead of nesting the new Date's again

lib/ASRouterTargeting.jsm
```diff
@@ -67,6 +67,9 @@ const TargetingGetters = {
       update: settings.update
     };
   },
+  get currentTimestamp() {
```
sarracini: Since this is not really a timestamp, maybe we should rename this to `currentDate` instead. We might want an actual `currentTimestamp` for different targeting in the future

PR #4358, opened 2018-08-22
Fix Bug 1482134 - Add client_id to onboarding telemetry in Activity Stream router

ncloudioj:


ncloudioj:
Thanks @sarracini 

Data review r+'d at https://bugzilla.mozilla.org/show_bug.cgi?id=1482134#c5

PR #4360, opened 2018-08-22
Fix Bug 1484031 - Add template schema for doorhanger

piatra:

PR #4366, opened 2018-08-23
[WIP] Fix Bug 1471389 - Create CFR message provider with experiment recipes

sarracini:
😱 some CFR recipes with targeting + triggers! 
Based off of latest schema patch in #4360 
To test, make sure the provider is enabled, and one of the `open_urls` from CFRMessageProvider.jsm is in your top sites (I used www.audible.com for easiness). Then visit a url from `open_urls` and verify that there was a match for the amazon assistant message (I just set a breakpoint in ASRouterTargeting,jsm when `match = candidate` and verified it got hit with the correct message). Then you can install this add-on: https://addons.mozilla.org/en-US/firefox/addon/the-camelizer-price-history-ch/ and try visiting a url from `open_urls` again and see that we *don't* match anymore!
Note that you may need to restart your browser after having new top sites, since the results are cached for 6 hours (see in ASTargeting.jsm)
Another thing to note is that this doesn't test the frequency lifetime cap stuff, since we don't *show* it, we don't get an impression, so we'll keep matching even after 1 time.
Also, i fixed a small bug with the trigger handler not passing in the {trigger} data via the message handler caller. It was being undefined when unpacking from `onMessage`

Some initial feedback on the message format here would be great @k88hudson @AdamHillier


lib/ASRouterTargeting.jsm
```diff
@@ -105,7 +105,6 @@ const TargetingGetters = {
         return {addons: info, isFullData: fullData};
       });
   },
-
```
sarracini: just tidying up the newline chars here.... the inconsistency was driving me 🍌s

lib/CFRMessageProvider.jsm
```diff
@@ -0,0 +1,95 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+const BASE_ADDONS_DOWNLOAD_URL = "https://addons.mozilla.org/firefox/downloads/file";
+const AMAZON_ASSISTANT_PARAMS = {
+  existing_addons: ["abb@amazon.com", "{75c7fe97-5a90-4b54-9052-3534235eaf41}", "{ef34596e-1e43-4e84-b2ff-1e58e287e08d}", "{ea280feb-155a-492e-8016-ac96dd995f2c}", "izer@camelcamelcamel.com", "amptra@keepa.com", "pricealarm@icopron.ch", "{774f76c7-6807-481e-bf64-f9b7d5cda602}"],
+  open_urls: ["smile.amazon.com", "www.audible.com", "www.amazon.com"]
```
AdamHillier: Is this going to be targeting US only?

lib/CFRMessageProvider.jsm
```diff
@@ -0,0 +1,95 @@
...
+
+const CFR_MESSAGES = [
+  {
+    id: "AMAZON_ASSISTANT_1",
+    template: "cfr",
+    notification_text: "Recommendation",
+    heading_text: "Recommended Extension",
+    info_icon: {
+      label: "why_seeing_this",
+      url: "https://support.mozilla.org"
```
AdamHillier: Out of interest will we have a specific CFR help page?

lib/ActivityStream.jsm
```diff
@@ -212,6 +212,12 @@ const PREFS_CONFIG = new Map([
       url: "https://activity-stream-icons.services.mozilla.com/v1/messages.json.br",
       updateCycleInMs: ONE_HOUR_IN_MS * 4,
       enabled: false
+    }, {
+      id: "cfr",
+      type: "local",
+      localProvider: "CFRMessageProvider",
+      enabled: false,
+      cohort: 0
```
AdamHillier: Will we have different CFR cohorts?

lib/CFRMessageProvider.jsm
sarracini: en-US locales only, but i think we can do that via shield

lib/CFRMessageProvider.jsm
sarracini: great question, i have no idea.... @k88hudson ?

lib/ActivityStream.jsm
sarracini: i _think_ so... this comment seems to suggest we will: https://bugzilla.mozilla.org/show_bug.cgi?id=1471389#c1
but it's still unclear how we're going to run this experiment with the other recommendations at the same time so for now i just left this empty, but we can do something basically like the onboarding cohort if we want

lib/CFRMessageProvider.jsm
```diff
@@ -0,0 +1,95 @@
...
+      },
+      secondary: {
+        label: "No Thanks",
+        action: {type: "CANCEL"}
+      }
+    },
+    frequency: {lifetime: 1},
+    targeting: `
+      (${JSON.stringify(AMAZON_ASSISTANT_PARAMS.existing_addons)} intersect addonsInfo.addons|keys)|length == 0 &&
+      (${JSON.stringify(AMAZON_ASSISTANT_PARAMS.open_urls)} intersect topFrecentSites|mapToProperty('host'))|length > 0`,
```
sarracini: @AdamHillier thank you for adding `mapToProperty` transform.... it's a dream

lib/CFRMessageProvider.jsm
k88hudson: I'm assuming yes, we'll need to request this from shell

AdamHillier:
Looking at the current Amazon Assistant listing on [AMO](https://addons.mozilla.org/en-GB/firefox/addon/amazon-browser-bar/), the description quite prominently includes "By installing you agree to the Conditions of Use at amazon.com/gp/BIT/TOU". This message doesn't appear anywhere in the CFR install flow. Just wanted to double check that that's okay?

lib/CFRMessageProvider.jsm
k88hudson: 1. Extension recommendation (To Google Translate, Enhancer for YouTube, Wikipedia Context Menu Search, Reddit Enhancement Suite, Facebook Container): https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/extensionrecommendations
2. Extension promotion (Amazon Assistant):  https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/extensionpromotions

k88hudson: So... apparently we'll need to have dynamic values :'(

lib/ActivityStream.jsm
```diff
@@ -212,6 +212,12 @@ const PREFS_CONFIG = new Map([
       url: "https://activity-stream-icons.services.mozilla.com/v1/messages.json.br",
       updateCycleInMs: ONE_HOUR_IN_MS * 4,
       enabled: false
+    }, {
+      id: "cfr",
+      type: "local",
+      localProvider: "CFRMessageProvider",
+      enabled: false,
```
k88hudson: can we do default on for beta at least

lib/CFRMessageProvider.jsm
k88hudson: let's make this `sumo_path` and the first part of the URL can be handled by the template. Can you also update the schema when you do that?

content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
```diff
@@ -30,10 +30,10 @@
             {"description": "Text for button tooltip used to provider information about the doorhanger."}
           ]
         },
-        "url": {
+        "sumo_path": {
           "allOf": [
-            {"$ref": "#/definitions/linkUrl"},
-            {"description": "Link to support page with the information about the doorhanger."}
+            {"$ref": "#/definitions/plainText"},
+            {"description": "Last part of the path in the URL to the support page with the information about the doorhanger."}
```
k88hudson: can you add an example?

content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
```diff
@@ -30,10 +30,10 @@
             {"description": "Text for button tooltip used to provider information about the doorhanger."}
           ]
         },
-        "url": {
+        "sumo_path": {
           "allOf": [
-            {"$ref": "#/definitions/linkUrl"},
-            {"description": "Link to support page with the information about the doorhanger."}
+            {"$ref": "#/definitions/plainText"},
```
k88hudson: I think this should be `string`, plainText is kind of for human-readable text

lib/CFRMessageProvider.jsm
```diff
@@ -0,0 +1,97 @@
...
+const FACEBOOK_CONTAINER_PARAMS = {
+  existing_addons: ["@contain-facebook", "{bb1b80be-e6b3-40a1-9b6e-9d4073343f0b}", "{a50d61ca-d27b-437a-8b52-5fd801a0a88b}"],
+  open_urls: ["www.facebook.com"],
+  sumo_path: "extensionrecommendations"
+};
+
+const CFR_MESSAGES = [
+  {
+    id: "AMAZON_ASSISTANT_1",
+    template: "cfr",
```
k88hudson: can we do `cfr_doorhanger` maybe?

PR #4372, opened 2018-08-27
Bug 1485528 - Ensure proper cache load when browser shutsdown and res…

ScottDowne:
To test:

1. Open about:config
2. Find config browser.newtabpage.activity-stream.feeds.section.topstories.spoc.impressions
3. Set it to
```
{"api_key_pref":"extensions.pocket.oAuthConsumerKey","hidden":false,"provider_icon":"pocket","provider_name":"Pocket","read_more_endpoint":"https://getpocket.com/explore/trending?src=fx_new_tab","stories_endpoint":"https://gist.githubusercontent.com/ScottDowne/164995d9535b4203846048bdee29d169/raw/9531da1c81fa4d15f8048567245141ef6948583f/spoc.json","stories_referrer":"https://getpocket.com/recommendations","topics_endpoint":"https://getpocket.cdn.mozilla.net/v3/firefox/trending-topics?version=2&consumer_key=$apiKey&locale_lang=en-US","show_spocs":true,"personalized":true}
```
4. Load about:home
5. Ensure the third rec is from theringer.com
6. close and open or restart the browser.
7. Expected: theringer.com still in third spot.

### DIFF
--- a/lib/TopStoriesFeed.jsm
+++ b/lib/TopStoriesFeed.jsm
@@ -86,10 +86,14 @@ this.TopStoriesFeed = class TopStoriesFeed {
     }
   }
 
+  async clearCache() {
+    await this.cache.set("stories", {});
+    await this.cache.set("topics", {});
+    await this.cache.set("spocs", {});
+  }
+
   uninit() {
     this.storiesLoaded = false;
-    this.cache.set("stories", {});
-    this.cache.set("topics", {});
     Services.obs.removeObserver(this, "idle-daily");
     SectionsManager.disableSection(SECTION_ID);
   }
@@ -126,6 +130,8 @@ this.TopStoriesFeed = class TopStoriesFeed {
         this.spocCampaignMap = new Map(body.spocs.map(s => [s.id, `${s.campaign_id}`]));
         this.spocs = this.transform(body.spocs).filter(s => s.score >= s.min_score);
         this.cleanUpCampaignImpressionPref();
+        // Spocs won't exist without stories, so no need to worry about last updated.
+        this.cache.set("spocs", this.spocs);
       }
       this.storiesLastUpdated = Date.now();
       body._timestamp = this.storiesLastUpdated;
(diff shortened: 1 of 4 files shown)


lib/TopStoriesFeed.jsm
```diff
@@ -86,10 +86,14 @@ this.TopStoriesFeed = class TopStoriesFeed {
     }
   }
 
-  uninit() {
+  async clearCache() {
     this.storiesLoaded = false;
```
k88hudson: you probably still want to have `.storiesLoaded` be false in `uninit`

lib/TopStoriesFeed.jsm
```diff
@@ -147,9 +155,11 @@ this.TopStoriesFeed = class TopStoriesFeed {
     }
     if (stories && stories.length > 0 && this.storiesLastUpdated === 0) {
       this.updateSettings(data.stories.settings);
-      const rows = this.transform(stories);
-      this.stories = rows;
+      this.stories = this.rotate(this.transform(stories));
```
k88hudson: Can you explain adding `rotate` here?

ScottDowne: Rotate checks if a story can be removed from the list and the next one moved forward, in most cases that's because a story has expired.

I added the rotate here to try to figure out why the spoc kept going into new positions, thinking it had to do with rotate and wondering why cache load didn't do a rotate and fetch did. Wondered if this was my bug. It didn't fix it, but still thought it was weird that cache didn't do it and fetch did. So I asked Christian about it.

If you use it when nothing is ready to expire, it shouldn't do anything.

I left it here because if we're loading the same stories, chances are better that something has expired, it felt harmless, and It also has the added benefit of if the user is offline, their stories can now refresh. If there is potential that this could cause an issue, or you want it tackled in anther patch, happy to revert it, as it shouldn't have any effect on the rest of the patch.

lib/TopStoriesFeed.jsm
ScottDowne: Agreed, and it also allowed me to find a test I probably should have had but didn't.

PR #4389, opened 2018-08-30
Fix Bug 1487156 - Track impressions for frequency capping in CFR

sarracini:
Adds frequency capping for when the chiclet is shown via `shouldExpand` only 
To test this, either see the impressions go up in about:newtab#asrouter, or in real life, see that the recommendation only appears once.

### DIFF
--- a/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -54,7 +54,7 @@ export class ASRouterAdmin extends React.PureComponent {
   renderMessageItem(msg) {
     const isCurrent = msg.id === this.state.lastMessageId;
     const isBlocked = this.state.messageBlockList.includes(msg.id);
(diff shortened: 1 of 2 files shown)


content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -54,7 +54,7 @@ export class ASRouterAdmin extends React.PureComponent {
   renderMessageItem(msg) {
     const isCurrent = msg.id === this.state.lastMessageId;
     const isBlocked = this.state.messageBlockList.includes(msg.id);
-    const impressions = this.state.messageImpressions[msg.id] ? this.state.impressions[msg.id].length : 0;
+    const impressions = this.state.messageImpressions[msg.id] ? this.state.messageImpressions[msg.id].length : 0;
```
sarracini: This was a bug, it wasn't actually displaying the frequency caps in about:newtab#asrouter

PR #4395, opened 2018-08-31
Port Bug 1485418 - Land CFR UI Strings r=flod,k88hudson

Mardak:
r?@piatra https://hg.mozilla.org/mozilla-central/rev/c14b6edd8863

Fixed various merge conflicts and updated messages to have string_id. Schema test fails.


Mardak:
Just to document, we ended up changing `getStrings` API to take 2 parameters with one optional of the attribute to request and always returns a "string"-ish value/object (where originally it was 1 object parameter with 2 possible string or hasAttributes properties and returns a string or object)

PR #4396, opened 2018-08-31
Fix Bug 1487772 - show logo and wordmark when only search enabled

rlr:
Before:
screen shot 2018-08-31 at 1 05 03 pm

After:
screen shot 2018-08-31 at 1 04 09 pm
screen shot 2018-08-31 at 1 03 56 pm


k88hudson:
We had some design concerns with this, including:
- the colour of the wordmark
- the size of the wordmark
- RTL orientation
- the overall balance of the page
- whether the wordmark should be different for different channels (e.g. nightly etc.)

bbell needs to loop back with shorlander on some of this for design feedback.

Mardak:
Just noting that bug 1409054 removed the "unofficial" about-logo@2x.png file, so if we need it, we could add it back.

https://hg.mozilla.org/mozilla-central/rev/81367b23ee5f

https://hg.mozilla.org/mozilla-central/log/81367b23ee5f/browser/branding/unofficial/content/about-logo%402x.png
https://hg.mozilla.org/mozilla-central/raw-file/0de748c25f49/browser/branding/unofficial/content/about-logo%402x.png

PR #4400, opened 2018-09-03
Fix Bug 1486909 - Remove init/uninit assignments to _currentSearchHostname, delete lazy…

piatra:
… getter before setting value

I didn't understand why we we're initializing it with null and setting it to null on uninit. It doesn't seem necessary. 
Deleting the lazy getter before setting the pref in the observer seems to be [the way to do it](https://searchfox.org/mozilla-central/rev/721842eed881c7fcdccb9ec0fe79e4e6d4e46604/browser/base/content/browser-thumbnails.js#99) if you want to reassign something to a property you previously had a lazy getter on.

In order to test this run our mochitests before and after the patch is applied. There will be warnings in the console triggered from the Topsites section uninit function.

### DIFF
--- a/lib/TopSitesFeed.jsm
+++ b/lib/TopSitesFeed.jsm
@@ -60,7 +60,6 @@ function getShortURLForCurrentSearch() {
 this.TopSitesFeed = class TopSitesFeed {
   constructor() {
     this._tippyTopProvider = new TippyTopProvider();
-    this._currentSearchHostname = null;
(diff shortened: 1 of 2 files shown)


k88hudson:
In general I think it's just easier to see what the properties of the class are supposed to be if you initialize them in the constructor with null or whatever rather than having to read through individual functions.

Actually I think it would be fine to add the lazy getter in the constructor, since it's lazy – did you want to make that change too?

PR #4401, opened 2018-09-04
Fix Bug 1488846 - Pocket cta for logged out users flickers a bit before displaying the logged in version (topics) and shifting

ScottDowne:
@k88hudson 

Two small design related issues noticed in https://bugzilla.mozilla.org/show_bug.cgi?id=1483655, which this fixes.

1. Flicker of logged in/out state while the page loads. Adding a gif below.

2. The height of the cta was actually a few pixels larger than the topics, which potentially causes a slight shift in content as it loads. I made sure on max screen sizes the content doesn't move.

### DIFF
--- a/common/Reducers.jsm
+++ b/common/Reducers.jsm
@@ -44,7 +44,7 @@ const INITIAL_STATE = {
   },
   Sections: [],
   Pocket: {
-    isUserLoggedIn: false,
+    isUserLoggedIn: null,
     pocketCta: {},
     waitingForSpoc: true
   }
--- a/content-src/components/Sections/Sections.jsx
+++ b/content-src/components/Sections/Sections.jsx
@@ -155,14 +155,21 @@ export class Section extends React.PureComponent {
     const maxCards = maxCardsPerRow * numRows;
     const maxCardsOnNarrow = CARDS_PER_ROW_DEFAULT * numRows;
 
-    const shouldShowPocketCta = (id === "topstories" &&
-      Pocket.pocketCta.useCta && !Pocket.isUserLoggedIn);
+    const {pocketCta, isUserLoggedIn} = Pocket || {};
+    const {useCta} = pocketCta || {};
 
-    // Show topics only for top stories and if it's not initialized yet (so
-    // content doesn't shift when it is loaded) or has loaded with topics
+    // Don't display anything until we have a definitve result from Pocket,
+    // to avoid a flash of logged out state while we render.
+    const isPocketLoggedInDefined = (isUserLoggedIn === true || isUserLoggedIn === false);
+
+    const shouldShowPocketCta = (id === "topstories" &&
+      useCta && isUserLoggedIn === false);
+
+    // Show topics only for top stories and if it has loaded with topics.
+    // The classs .top-stories-bottom-container ensures content doesn't shift as things load.
     const shouldShowTopics = (id === "topstories" &&
-      (!topics || topics.length > 0) &&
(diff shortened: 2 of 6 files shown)


common/Reducers.jsm
```diff
@@ -39,7 +39,6 @@ const INITIAL_STATE = {
   },
   Sections: [],
   Pocket: {
-    isUserLoggedIn: false,
```
ScottDowne: We initially start out without a true or false state, we just don't know yet.

content-src/components/Sections/Sections.jsx
```diff
@@ -137,14 +137,21 @@ export class Section extends React.PureComponent {
...
     const shouldShowPocketCta = (id === "topstories" &&
-      Pocket.pocketCta.useCta && !Pocket.isUserLoggedIn);
+      useCta && isUserLoggedIn === false);
 
-    // Show topics only for top stories and if it's not initialized yet (so
-    // content doesn't shift when it is loaded) or has loaded with topics
+    // Show topics only for top stories and if it has loaded with topics.
+    // The classs .top-stories-bottom-container ensures content doesn't shift as things load.
     const shouldShowTopics = (id === "topstories" &&
-      (!topics || topics.length > 0) &&
```
ScottDowne: I noticed the previous code would display empty topics as a placeholder of sorts to ensure we don't have shifting content while the page loads. I added a wrapper `top-stories-bottom-container` and moved the "more recs" bit out of topics so we no longer need to rely on topics for this, thus we can remove the `!topics` bit from the `!topics || topics.length > 0` check.

content-src/components/Sections/Sections.jsx
```diff
@@ -137,14 +137,21 @@ export class Section extends React.PureComponent {
...
-      !shouldShowPocketCta);
+      (topics && topics.length > 0) &&
+      ((useCta && isUserLoggedIn === true) || (!useCta && isPocketLoggedInDefined)));
```
ScottDowne: I'm essentially trying to sort out this logic so we don't display anything until we know enough of what to display.

Like I mentioned above, the content has a placecholder, and the "more recs" link is still displayed. I think this should be fast as it's only waiting for the pocket jsm to return if the user is logged in or not.

content-src/components/Sections/_Sections.scss
```diff
@@ -83,6 +83,7 @@
 
   @media (min-width: $break-point-large) {
     line-height: 16px;
+    height: 16px;
```
ScottDowne: This just ensures the placeholder is always the same height no matter what content it has so it doesn't shift the page.
without this the page shifts by two pixels when the cta is added.

ScottDowne:
Example of the flicker between logged in and out state.
opt

ScottDowne:
@k88hudson let me know if you have any questions or I missed anything. Or if I should be asking someone else for review.

Thanks!

k88hudson:
@thecount can you file a new bug for this?

k88hudson:
I noticed this also happens now on the prerendered page, which isn't ideal – do you think you could fix this too?
image

common/Reducers.jsm
k88hudson: actually could you make this `null` instead of false, just so we can see the shape of the state?

k88hudson:
Good to go, just add that one change to `Reducers`

Mardak:
The commit referenced bug 1483655 instead of bug 1488846.

PR #4404, opened 2018-09-05
Fix Bug 1482534 - remove -moz-osx-font-smoothing: grayscale;

rlr:


rlr:
@AdamHillier I am still waiting on design feedback. But I almost want to wontfix this from our end and assign bug to the appropriate platform component that is causing this?

PR #4405, opened 2018-09-06
Fix Bug 1483977 - Add targeting getter for total bookmarks count

piatra:


lib/ASRouterTargeting.jsm
```diff
@@ -168,6 +168,9 @@ const TargetingGetters = {
       Cu.reportError("Problem parsing JSON message provider pref for ASRouter");
     }
     return cohorts;
+  },
+  get totalBookmarksCount() {
+    return asProvider.getTotalBookmarksCount();
```
k88hudson: Would you mind adding some caching to this similar to how `TopFrecentSitesCache` works?

piatra:
@k88hudson updated

lib/ASRouterTargeting.jsm
```diff
@@ -23,31 +23,43 @@ const FRECENT_SITES_IGNORE_BLOCKED = true;
...
-        this._topFrecentSites = await asProvider.getTopFrecentSites({
-          ignoreBlocked: FRECENT_SITES_IGNORE_BLOCKED,
-          numItems: FRECENT_SITES_NUM_ITEMS,
-          topsiteFrecency: FRECENT_SITES_MIN_FRECENCY,
-          onePerDomain: true,
-          includeFavicon: false
-        });
-        this._lastUpdated = now;
+      if (now - targetingGetter._lastUpdated >= updateInterval) {
+        targetingGetter._value = await asProvider[property](options);
```
k88hudson: This could use some error handling (i.e. reject if this fails)

k88hudson: Also some tests for the cache invalidation behaviour, although a bit hard to do with mochi tests (it would be better with sinon)

PR #4406, opened 2018-09-06
Fix Bug 1484206 - Add previousSessionEnd to targeting context

piatra:
This PR has unit tests instead of our usual mochitests for targeting because targeting parameter was added to `ASRouter` and that's more complicated to set up for mochitesting than our (stateless) `ASRouterTargeting.jsm`

Fetches and updates the timestamp in the router, passes it down to targeting through the context parameter. Targeting will merge the provided context and default one. `ASRouterFeed` calls `Router.uninit` on shutdown and that's when we update the timestamp.


piatra:
Checking to see what's failing.

PR #4407, opened 2018-09-06
Fix Bug 1486617: Add unit tests for CFRPageActions.jsm

AdamHillier:


test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,634 @@
...
+        CFRPageActions.PageActionMap.set(fakeBrowser.ownerGlobal, pageAction);
+        await CFRPageActions.addRecommendation(fakeBrowser, fakeHost, fakeRecommendation, dispatchStub);
+        // Reset the call history for all spies and stubs so that we needn't consider the effects of the above
+        sandbox.resetHistory();
+        getStringsStub.callsFake(async a => a)
+          .withArgs({string_id: "primary_button_id"})
+          .resolves({value: "Primary Button", attributes: {accesskey: "p"}})
+          .withArgs({string_id: "secondary_button_id"})
+          .resolves({value: "Secondary Button", attributes: {accesskey: "s"}})
+          .withArgs(sinon.match({string_id: "cfr-doorhanger-extension-learn-more-link"}))
```
piatra: I don't understand why the call to `sinon.match` in the setup step.

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,634 @@
...
+      it("should do nothing if a pageAction doesn't exist for the window", () => {
+        const win = fakeBrowser.ownerGlobal;
+        CFRPageActions.PageActionMap.delete(win);
+        CFRPageActions.updatePageActions(fakeBrowser);
+        assertDidNothing();
+      });
+      it("should do nothing if the browser is not the `selectedBrowser`", () => {
+        const someOtherFakeBrowser = {};
+        CFRPageActions.updatePageActions(someOtherFakeBrowser);
+        assertDidNothing();
```
piatra: nit: I would keep the assertions in-line since the function is only used twice.

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,634 @@
...
+        assert.isUndefined(pageActions[1]);
+        assert.notCalled(pageActions[2].hide);
+      });
+      it("should clear the PageActionMap and the RecommendationMap", () => {
+        CFRPageActions.clearRecommendations();
+
+        // Both are WeakMaps and so are not iterable, cannot be cleared, and
+        // cannot have their length queried directly, so we have to check
+        // whether previous elements still exist
+        for (const win of windows) {
```
piatra: This also needs some assertion related to the number of `windows` or `browsers` we expect to find otherwise (if 0) the test doesn't assert anything.

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,634 @@
...
+        pageAction._expand();
+        assert.calledOnce(pageAction._clearScheduledStateChanges);
+        assert.equal(pageAction.urlbar.getAttribute("cfr-recommendation-state"), "expanded");
+      });
+      it("with a delay, should set the expanded state after the correct amount of time", () => {
+        const delay = 1234;
+        pageAction._expand(delay);
+        assert.lengthOf(pageAction.stateTransitionTimeoutIDs, 1);
+        assert.notOk(pageAction.urlbar.getAttribute("cfr-recommendation-state"));
+        clock.tick(delay - 1);
```
piatra: I think these assertions are too specific: testing the correct behaviour of setTimeout. I would instead only test that the correct attribute is set. It could be grouped with the test for the `#show` method and not have an explicit call for `_expand`.
I don't see any use for the `delay = 0` branch. Can that be removed?

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,634 @@
...
+      const elem = document.createElement("div");
+      elem.setAttribute("id", id);
+      containerElem.appendChild(elem);
+      elements[id] = elem;
+    }
+  });
+
+  afterEach(() => {
+    CFRPageActions.clearRecommendations();
+    containerElem.remove();
```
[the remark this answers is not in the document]
AdamHillier: All the other elements are children of the container so doesn't removing the container also remove them?

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,634 @@
...
+    let pageAction;
+    let getStringsStub;
+
+    beforeEach(() => {
+      pageAction = new PageAction(window, dispatchStub);
+      getStringsStub = sandbox.stub(pageAction, "getStrings").resolves("");
+    });
+
+    afterEach(() => {
+      pageAction._clearScheduledStateChanges();
```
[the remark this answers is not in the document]
AdamHillier: Turns out this cleanup isn't even needed

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,641 @@
...
+        const fakeNotification = {};
+
+        pageAction.currentNotification = fakeNotification;
+        pageAction._popupStateChange("dismissed");
+        assert.equal(pageAction.urlbar.getAttribute("cfr-recommendation-state"), "collapsed");
+        assert.calledOnce(global.PopupNotifications.remove);
+        assert.calledWith(global.PopupNotifications.remove, fakeNotification);
+
+        pageAction.currentNotification = fakeNotification;
+        sandbox.resetHistory();
```
piatra: can you split this into two tests instead of the reset? Should make it easier to debug in the future.

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,641 @@
...
+            bucket_id: fakeRecommendation.content.bucket_id,
+            event: "INSTALL"
+          }
+        });
+        // Should remove the recommendation
+        assert.isFalse(CFRPageActions.RecommendationMap.has(fakeBrowser));
+
+        // Is the secondaryAction correct?
+        assert.deepEqual(secondaryAction.label, {value: "Secondary Button", attributes: {accesskey: "s"}});
+        sandbox.resetHistory(); // .hide was called above so reset the call count
```
piatra: I would also split this up and use a `beforeEach` block if setup is too complicated/long

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,641 @@
...
+        assert.equal(await pageAction.getStrings({string_id: "hello_world"}, "second_attr"), "some string");
+      });
+    });
+
+    describe("#_handleClick", () => {
+      beforeEach(async () => {
+        CFRPageActions.PageActionMap.set(fakeBrowser.ownerGlobal, pageAction);
+        await CFRPageActions.addRecommendation(fakeBrowser, fakeHost, fakeRecommendation, dispatchStub);
+        // Reset the call history for all spies and stubs so that we needn't consider the effects of the above
+        sandbox.resetHistory();
```
piatra: should this be in a `afterEach` block => `sandbox.restore()` instead?

piatra: Is there any way we can work around this? It seems like a weird hack, at most I would isolate this call to the specific tests (`it` blocks) that require it otherwise it's easy to overlook when adding new tests.

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -0,0 +1,641 @@
...
+            action: {id: "primary_action"}
+          },
+          secondary: {
+            label: {string_id: "secondary_button_id"},
+            action: {id: "secondary_action"}
+          }
+        }
+      }
+    };
+    fakeHost = "www.mozilla.org";
```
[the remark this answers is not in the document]
AdamHillier: This does contain `www.`. Do you mean they should _not_ contain `www.`?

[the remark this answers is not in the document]
piatra: _not_

test/unit/asrouter/CFRPageActions.test.js
piatra: @AdamHillier I meant https://github.com/mozilla/activity-stream/blob/f0443a23070f3b68b13546c40c5e881b3e2921d5/lib/CFRPageActions.jsm#L124-L128

test/unit/asrouter/CFRPageActions.test.js
AdamHillier: You're right, it is a bit weird, luckily it seems to not actually be necessary so I've removed it

PR #4408, opened 2018-09-07
Fix bug 1489619 - Loading spocs from cache doesn't properly reset cam…

ScottDowne:
…paign impression map


ScottDowne:
@k88hudson ^^

lib/TopStoriesFeed.jsm
```diff
@@ -169,8 +166,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
       this.updateSettings(data.stories.settings);
       this.stories = this.rotate(this.transform(stories));
       this.storiesLastUpdated = data.stories._timestamp;
-      if (spocs && spocs.length) {
-        this.spocs = spocs;
+      if (data.stories.spocs && data.stories.spocs.length) {
```
ScottDowne: So instead of looking at post processed spocs that we cached, we look at the un processed spocs cached in body (which gets renamed to stories). Then just do the same processing to them that the fetch does.

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -195,22 +195,23 @@ describe("Top Stories Feed", () => {
...
+          spocs: [{"id": "spoc1"}]
+        }
       });
       instance.storiesLastUpdated = 0;
       globals.set("NewTabUtils", {blockedLinks: {isBlocked: () => {}}});
 
       await instance.loadCachedData();
-
-      assert.equal(instance.spocs[0].id, "spoc1");
+      assert.equal(instance.spocs[0].guid, "spoc1");
```
ScottDowne: The key here is if after loading cache, we have a spoc with guid and not id, which means the proper processing has happened.

ScottDowne:
@k88hudson had to do a rebase, should be good now, but not sure if I remember what updates you wanted on this.

PR #4411, opened 2018-09-11
Fix Bug 1484207 - Add geo and firefox version targeting getters

piatra:
Some more insights: This is needed because previously Snippets could execute code in the browser and it was done to get [user country](https://github.com/mozmeao/snippets-service/blob/2f84e39574bd496921a3defc496888c8ff8c66d0/snippets/base/templates/base/includes/snippet_as.js#L484) and [useragent](https://github.com/mozmeao/snippets-service/blob/2f84e39574bd496921a3defc496888c8ff8c66d0/snippets/base/templates/base/includes/snippet_as.js#L234). Now we need to provide that information.

Using `browser.search.region` is equivalent because Search services calls the same endpoint to get country code [0] [1].

I also updated the remote snippets url, previously it was poiting to a static bundle.

[0] https://searchfox.org/mozilla-central/rev/9e7995b3c384945740985107a4de601b27904718/toolkit/components/search/nsSearchService.js#561
[1] https://searchfox.org/mozilla-central/rev/9e7995b3c384945740985107a4de601b27904718/modules/libpref/init/all.js#5664

### DIFF
--- a/content-src/asrouter/schemas/message-format.md
+++ b/content-src/asrouter/schemas/message-format.md
@@ -100,6 +100,8 @@ Name | Type | Example value | Description
 `providerCohorts` | `Object` | `{onboarding: "hello"}` | Cohorts defined for all providers
 `previousSessionEnd` | `Number` | `1536325802800` | Timestamp in milliseconds of previously closed session
 `totalBookmarksCount` | `Number` | `8` | Total number of bookmarks
+`firefoxVersion` | `Number` | `64` | Firefox version number rounded down
+`geoCountry` | `String` | `US` | Country code retrieved from `location.services.mozilla.com` can be `""` if request did not finish, encountered an error
 #### addonsInfo Example
 
 ```javascript
--- a/content-src/asrouter/schemas/message-format.md
+++ b/content-src/asrouter/schemas/message-format.md
@@ -100,8 +100,8 @@ Name | Type | Example value | Description
 `providerCohorts` | `Object` | `{onboarding: "hello"}` | Cohorts defined for all providers
 `previousSessionEnd` | `Number` | `1536325802800` | Timestamp in milliseconds of previously closed session
 `totalBookmarksCount` | `Number` | `8` | Total number of bookmarks
-`firefoxVersion` | `Number` | `64` | Firefox version number rounded down
-`geoCountry` | `String` | `US` | Country code retrieved from `location.services.mozilla.com` can be `""` if request did not finish, encountered an error
+`firefoxVersion` | `Number` | `64` | The major Firefox version of the browser 
+`region` | `String` | `US` | Country code retrieved from `location.services.mozilla.com` can be `""` if request did not finish, encountered an error
 #### addonsInfo Example
 
 ```javascript
--- a/content-src/asrouter/schemas/provider-response.schema.json
+++ b/content-src/asrouter/schemas/provider-response.schema.json
@@ -2,7 +2,7 @@
   "title": "ProviderResponse",
   "description": "A response object for remote providers of AS Router",
   "type": "object",
-  "version": "6",
+  "version": "6.1.0",
   "properties": {
     "messages": {
       "type": "array",
--- a/content-src/asrouter/schemas/provider-response.schema.json
+++ b/content-src/asrouter/schemas/provider-response.schema.json
@@ -2,7 +2,7 @@
   "title": "ProviderResponse",
   "description": "A response object for remote providers of AS Router",
   "type": "object",
-  "version": "0.1.0",
+  "version": "6",
   "properties": {
     "messages": {
       "type": "array",
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -181,9 +181,9 @@ const TargetingGetters = {
     return TotalBookmarksCountCache.getTotalBookmarksCount;
   },
   get firefoxVersion() {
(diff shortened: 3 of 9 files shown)


test/browser/browser_asrouter_targeting.js
```diff
@@ -285,6 +285,18 @@ add_task(async function checkFrecentSites() {
   TopFrecentSitesCache.expire();
 });
 
+add_task(async function check_firefox_version() {
+  const message = {id: "foo", targeting: "firefoxVersion > 0"};
```
piatra: What I considered is `undefined > 0 // false` if there is an issue and the getter does not return a value. Not sure how to make the test more strict while still working across firefox versions.

test/browser/browser_asrouter_targeting.js
```diff
@@ -285,6 +285,18 @@ add_task(async function checkFrecentSites() {
...
+  is(await ASRouterTargeting.findMatchingMessage({messages: [message]}), message,
+    "should select correct item when filtering by firefox version");
+});
+
+add_task(async function check_geoCountry() {
+  const message = {id: "foo", targeting: "geoCountry in ['US']"};
```
piatra: I will do a try run before to confirm this expectation holds in our testing infrastructure but local builds seem to be `US`.

lib/ASRouterTargeting.jsm
```diff
@@ -168,6 +171,12 @@ const TargetingGetters = {
       Cu.reportError("Problem parsing JSON message provider pref for ASRouter");
     }
     return cohorts;
+  },
+  get firefoxVersion() {
+    return parseInt(AppConstants.MOZ_APP_VERSION, 10);
+  },
+  get geoCountry() {
+    return Services.prefs.getStringPref(SEARCH_REGION_PREF);
```
k88hudson: `Services.prefs.getStringPref(SEARCH_REGION_PREF, "")` maybe?

test/browser/browser_asrouter_targeting.js
k88hudson: Why not use `SpecialPowers.pushPrefEnv` to test this more robustly (probably with something other than `US`)

test/browser/browser_asrouter_targeting.js
k88hudson: I think this is fine, what you really want to test here is probably that we're getting an integer anyway (AppConstants.MOZ_APP_VERSION is presumably pretty safe)

content-src/asrouter/schemas/provider-response.schema.json
```diff
@@ -2,7 +2,7 @@
   "title": "ProviderResponse",
   "description": "A response object for remote providers of AS Router",
   "type": "object",
-  "version": "0.1.0",
+  "version": "6",
```
k88hudson: Is this to line up with the services endpoint?

piatra: Yes, we were using that version number to format the URL pref

test/unit/unit-entry.js
```diff
@@ -200,7 +200,7 @@ const TEST_GLOBAL = {
   },
   EventEmitter,
   ShellService: {isDefaultBrowser: () => true},
-  FilterExpressions: {eval() { return Promise.resolve(true); }},
+  FilterExpressions: {eval() { return Promise.resolve(false); }},
```
piatra: The fix from yesterday landed before I realised the mistake. The test was incorrect, we should default to failing in our mocks.

k88hudson: good idea

lib/ASRouterTargeting.jsm
```diff
@@ -176,6 +179,12 @@ const TargetingGetters = {
   },
   get totalBookmarksCount() {
     return TotalBookmarksCountCache.getTotalBookmarksCount;
+  },
+  get firefoxVersion() {
+    return parseInt(AppConstants.MOZ_APP_VERSION, 10);
```
k88hudson: how about `return parseInt(AppConstants.MOZ_APP_VERSION.match(/\d+/), 10);` to make this more robust

lib/ASRouterTargeting.jsm
```diff
@@ -176,6 +179,12 @@ const TargetingGetters = {
...
+  },
+  get geoCountry() {
```
k88hudson: maybe we should call this `region` to be more consistent with serverside filtering

content-src/asrouter/schemas/message-format.md
```diff
@@ -100,6 +100,8 @@ Name | Type | Example value | Description
 `providerCohorts` | `Object` | `{onboarding: "hello"}` | Cohorts defined for all providers
 `previousSessionEnd` | `Number` | `1536325802800` | Timestamp in milliseconds of previously closed session
 `totalBookmarksCount` | `Number` | `8` | Total number of bookmarks
+`firefoxVersion` | `Number` | `64` | Firefox version number rounded down
```
k88hudson: Better description here would be `The major Firefox version of the browser`

lib/ASRouterTargeting.jsm
```diff
@@ -176,6 +179,12 @@ const TargetingGetters = {
...
+    return Services.prefs.getStringPref(SEARCH_REGION_PREF, "");
```
k88hudson: I wonder if we should do `Services.prefs.getStringPref(SEARCH_REGION_PREF, "") || "ZZ"` when nothing is found like the server side filter does?

PR #4413, opened 2018-09-11
Fix Bug 1433243 - Implement ASR persistent caching for remote endpoints

rlr:
I also added the ETag header so the CDN can return 304 (Not Modified) if the file hasn't changed since the last fetch. I manually tested by adding a bunch of console statements and it all seems to be working for me locally.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -41,6 +41,9 @@ const LOCAL_MESSAGE_PROVIDERS = {OnboardingMessageProvider, CFRMessageProvider};
 const STARTPAGE_VERSION = "0.1.0";
 
 const MessageLoaderUtils = {
+  STARTPAGE_VERSION,
+  REMOTE_LOADER_CACHE_KEY: "RemoteLoaderCache",
+
   /**
    * _localLoader - Loads messages for a local provider (i.e. one that lives in mozilla central)
    *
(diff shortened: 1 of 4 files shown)


lib/ASRouter.jsm
```diff
@@ -55,21 +55,45 @@ const MessageLoaderUtils = {
...
    * @param {obj} provider An AS router provider
    * @param {string} provider.url An endpoint that returns an array of messages as JSON
+   * @param {obj} storage A storage object with get() and set() methods for caching.
    * @returns {Promise} resolves with an array of messages, or an empty array if none could be fetched
    */
-  async _remoteLoader(provider) {
+  async _remoteLoader(provider, storage) {
     let remoteMessages = [];
     if (provider.url) {
+      const cacheKey = `RemoteLoaderCache-${provider.url}`;
```
k88hudson: I think it would be better to store everything under one cache key, just because clearing it will be easier; also, let's just cache by `provider.id` (since we want to clear outdated responses anyway).

Also, I think it would be good to store the `STARTPAGE_VERSION` (which is really just the version of ASRouter, it's very badly named) so we can invalidate if the major version changes.

I ended up [writing some code](https://github.com/rlr/activity-streams/compare/bug1433243/asr-cache...k88hudson:suggestions-for-caching) to think about this one if it's helpful

lib/ASRouter.jsm
```diff
@@ -55,21 +55,45 @@ const MessageLoaderUtils = {
    *
    * @param {obj} provider An AS router provider
    * @param {string} provider.url An endpoint that returns an array of messages as JSON
+   * @param {obj} storage A storage object with get() and set() methods for caching.
    * @returns {Promise} resolves with an array of messages, or an empty array if none could be fetched
    */
-  async _remoteLoader(provider) {
+  async _remoteLoader(provider, storage) {
```
k88hudson: You need to add a special case never to cache preview URLs somehow – maybe check if `updateCycleInMs` is 0?

lib/ASRouter.jsm
rlr: that makes sense, awesome! thanks for the patch 🕺

lib/ASRouter.jsm
```diff
@@ -380,6 +404,7 @@ class _ASRouter {
     await this.setState({messageBlockList, providerBlockList, messageImpressions, providerImpressions, previousSessionEnd});
     this._updateMessageProviders();
     await this.loadMessagesFromAllProviders();
+    await MessageLoaderUtils.cleanupCache(this.state.providers, storage);
```
rlr: @k88hudson I added the cache cleanup at the end of `init()`. Thoughts?

k88hudson: that seems reasonable yep!

lib/ASRouter.jsm
k88hudson: Awesome, works great!

k88hudson:
Awesome!

PR #4417, opened 2018-09-12
Fixes bug 1490734 - Fresh profile can sometimes display no card in th…

ScottDowne:
…e third card position

### DIFF
--- a/content-src/components/Sections/Sections.jsx
+++ b/content-src/components/Sections/Sections.jsx
@@ -1,6 +1,6 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 import {Card, PlaceholderCard} from "content-src/components/Card/Card";
 import {FormattedMessage, injectIntl} from "react-intl";
-import {actionCreators as ac} from "common/Actions.jsm";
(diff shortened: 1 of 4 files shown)


ScottDowne:
More info in bugzilla around the problem https://bugzilla.mozilla.org/show_bug.cgi?id=1490734

Solution is to move the `NEW_TAB_REHYDRATED` event from app.initialized check to section.initialized check. This I *think* has the added bonus of no longer needing to check for `hasTopStoriesSectionChanged` to fix this event because if the section changes, it should just created a new Section component. I think. I came to that conclusions because from my testing, that's what it's doing.

content-src/components/Sections/Sections.jsx
```diff
@@ -103,6 +107,10 @@ export class Section extends React.PureComponent {
     }
   }
 
+  componentWillUpdate(nextProps) {
+    this.sendNewTabRehydrated(nextProps.initialized);
```
k88hudson: Don't you still need to check for relevant updates here?

k88hudson: Ah I see, you removed that part

PR #4420, opened 2018-09-13
Fix Bug 1490241 - CFR fluent and string overrides should both have attributes

piatra:
Turned the usage examples from https://github.com/mozilla/activity-stream/commit/21268904be632c4d74f1fa981e852412a94794ef#r30406394 into unit tests.

### DIFF
--- a/lib/CFRPageActions.jsm
+++ b/lib/CFRPageActions.jsm
@@ -185,10 +185,15 @@ class PageAction {
   async getStrings(string, subAttribute = "") {
     if (!string.string_id) {
       if (subAttribute) {
-        return string.attributes[subAttribute];
+        if (string.attributes) {
+          return string.attributes[subAttribute];
+        }
+
+        Cu.reportError(`String ${string.value} does not contain any attributes`);
+        return subAttribute;
       }
 
-      if (string.attributes) {
+      if (typeof string.value === "string") {
         const stringWithAttributes = new String(string.value); // eslint-disable-line no-new-wrappers
         stringWithAttributes.attributes = string.attributes;
         return stringWithAttributes;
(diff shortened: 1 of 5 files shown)


lib/CFRPageActions.jsm
```diff
@@ -184,6 +184,16 @@ class PageAction {
    */
   async getStrings(string, subAttribute = "") {
     if (!string.string_id) {
+      if (subAttribute) {
+        return string.attributes[subAttribute];
+      }
+
+      if (string.attributes) {
```
Mardak: This check might want to be something with `string.value` maybe typeof string? Unless we're saying only set `.value` if specifying attributes because otherwise just use a plain string instead of an object.

Although I suppose additionally accepting an object with a single `value` property makes it simpler to remove attributes from a string… (but would that even happen anyway?)

lib/CFRPageActions.jsm
```diff
@@ -184,6 +184,16 @@ class PageAction {
    */
   async getStrings(string, subAttribute = "") {
     if (!string.string_id) {
+      if (subAttribute) {
+        return string.attributes[subAttribute];
```
Mardak: We have a similar issue of a potential exception where a localized string doesn't have any attributes and here the `string` parameter might not have `.attributes`. It seems more likely that remote provided messages/strings could be improperly structured, so we might want to fall back somehow (ignoring the fact that popupnotification has its own stricter requirements for accesskey).

lib/CFRPageActions.jsm
piatra: We would always specify an object, that can be `{string_id...}` or `{value...}` regardless of attributes.

piatra: > only set .value if specifying attributes because otherwise just use a plain string instead of an object.

I think it would be: set .value or .string_id if specifying attributes otherwise use a plain string.

PR #4422, opened 2018-09-13
Fix Bug 1484996: add weights for ASRouter messages

AdamHillier:
Derived from Algorithm A [here](http://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf)

The paper doesn't give proofs of the right mathematical properties; I have provided some justification [here in the Bugzilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1484996#c2).

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -202,7 +202,11 @@ const MessageLoaderUtils = {
       Cu.reportError(new Error(`Tried to load messages for ${provider.id} but the result was not an Array.`));
     }
     const lastUpdated = Date.now();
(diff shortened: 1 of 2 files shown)


AdamHillier:
I've rebased but I can't see why TaskCluster is failing - it won't show me any logs and I can't retrigger without LDAP login

piatra:
@AdamHillier The error is fixed in my PR #4437 
TC seems available to me in private browsing mode.

AdamHillier:
@piatra Oh sorry yes I see them now, thanks

PR #4428, opened 2018-09-17
Fix Bug 1485626 - Tab_Index reaches elements inside collapsed menus in activity stream

rlr:
To verify:
* Make sure collapsing and expanding still works
* Make sure you can't tab through top sites, for example, while they are collapsed

### DIFF
--- a/content-src/components/CollapsibleSection/CollapsibleSection.jsx
+++ b/content-src/components/CollapsibleSection/CollapsibleSection.jsx
@@ -69,7 +69,7 @@ export class _CollapsibleSection extends React.PureComponent {
     // Get the current height of the body so max-height transitions can work
     this.setState({
       isAnimating: true,
-      maxHeight: `${this.sectionBody.scrollHeight}px`,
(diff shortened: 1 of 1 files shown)


content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -79,6 +79,18 @@ export class _CollapsibleSection extends React.PureComponent {
     }));
   }
 
+  _getSectionBodyHeight() {
+    const div = this.sectionBody;
+    if (div.style.display === "none") {
+      // If the div isn't displayed, we can't get it's height. So we display it
+      // with a 0px height to get the height. We don't undo this because we are about
+      // to expand the section anyway;
+      div.style.maxHeight = "0px";
```
rlr: @piatra found we can remove this line because we have this already set in the default css -> https://github.com/mozilla/activity-stream/blob/4104d533b8a46a85ab6d1d5756e368a4dfbcbc2f/content-src/components/CollapsibleSection/_CollapsibleSection.scss#L132

PR #4429, opened 2018-09-18
Fix Bug 1487793 - Use UpdateUtils.getUpdateChannel() instead of AppConstants.MOZ_UPDATE_CHANNEL

sarracini:
I'm assuming we want to include the partner bits, as that's one of the differences between AppConstants.MOZ_UPDATE_CHANNEL and UpdateUtils.getUpdateChannel(). However all the use cases in tree currently do not include partner bits, so maybe for consistency we don't want to either? @Mardak  what do you think?


lib/ActivityStream.jsm
```diff
@@ -166,7 +168,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["improvesearch.topSiteSearchShortcuts", {
     title: "Experiment to show special top sites that perform keyword searches",
-    value: AppConstants.MOZ_UPDATE_CHANNEL !== "release",
+    value: UpdateUtils.getUpdateChannel() !== "release",
```
k88hudson: Looks like update channel might contain some partner information (unless we pass in `false` here), is it intended that we wouldn't match those? https://searchfox.org/mozilla-central/source/toolkit/modules/UpdateUtils.jsm#

sarracini: Yeah I'm not sure about that either (see PR comment). I asked @Mardak if he had some insight on this

ncloudioj:
From the telemetry's perspective, we might not want to include the partner info in the release_channel for the backward compatibility.

lib/ActivityStream.jsm
Mardak: Let's pass in `true` in the empty / default calls just to be explicit just in case someone thinks about adding another "release" check for something else and comes across the empty one before noticing the `false` one.

PR #4431, opened 2018-09-18
Fix Bug 1482113 - Don't collect metrics when previewing Snippets

rlr:


content-src/asrouter/components/SnippetBase/SnippetBase.jsx
```diff
@@ -7,6 +7,10 @@ export class SnippetBase extends React.PureComponent {
   }
 
   onBlockClicked() {
+    if (this.props.provider === "preview") {
```
piatra: We don't want to prevent the dismiss action otherwise it might look like a bug to QA.

PR #4433, opened 2018-09-19
(mochitests): Fix CFR mochitest intermittent

piatra:


test/browser/browser_asrouter_cfr.js
```diff
@@ -42,9 +42,11 @@ function trigger_cfr_panel(browser, trigger, cb) {
 
 add_task(async function test_cfr_notification_show() {
   // addRecommendation checks that scheme starts with http and host matches
-  let tab = await BrowserTestUtils.openNewForegroundTab(gBrowser, "http://example.com", false);
```
piatra: The issue was this 3rd parameter of `openNewForgroundTab`, `waitForLoad`. I switched to a different method of loading a URL that seemed reliable in debug verify tests.

piatra: We use [`openNewForegroundTab`](https://github.com/mozilla/activity-stream/blob/143e4f6ca5d905bdf05097f87d53d945af801198/test/browser/head.js#L135) with `false`. The default value for that parameter is `true`

Mardak: We skip the automatic load check for `test_newtab` because we explicitly wait for load ourselves to handle both preloaded and unpreloaded about:newtab cases. Regular pages should be just loaded "normally"

PR #4436, opened 2018-09-19
first pass l10n

sarracini:
Ok... assuming there's an onboarding.ftl in browser/locales/en-US/browser/newtab/ with all these strings IDs, this would be the approach we could take to localize onboarding until bug 1488973 lands.
The gross part is the "Welcome to Firefox" header and the "Start Browsing" button that exist in the modal UI, because they're not part of any message, and they're not really part of any given "bundle" but they are part of onboarding. So I'm returning a 'header' and a 'button_label' as part of the onboarding provider.
This leads to ugliness....
@k88hudson @Mardak @piatra what are your thoughts


lib/ASRouter.jsm
```diff
@@ -591,7 +591,30 @@ class _ASRouter {
...
+    // handle finding these strings on its own. See bug 1488973
+    const {header, button_label} = await this._getExtraAttributes(originalMessage);
+
+    return {bundle: this._orderBundle(result), ...(header && {header}), ...(button_label && {button_label}), provider: originalMessage.provider, template: originalMessage.template};
+  }
+
+  async _getExtraAttributes(originalMessage) {
+    let localProvider = this._findProvider(originalMessage.provider);
+    let header = null;
+    if (localProvider && localProvider.getHeader) {
```
k88hudson: Can you combine these into something like `localProvider.getExtraTemplateStrings` or something like that? Seems like it would be easier to make a change if we need something added

k88hudson: Also can you add a comment about why this is necessary/ a reference to the bug that needs to be fixed to remove this?

lib/ASRouter.jsm
```diff
@@ -591,7 +591,30 @@ class _ASRouter {
       return null;
     }
 
-    return {bundle: this._orderBundle(result), provider: originalMessage.provider, template: originalMessage.template};
+    // The bundle may have some extra attributes, like a header, or a dismiss button, so attempt to get those strings now
+    // This is a temporary solution until we can use Fluent strings in the content process, in which case the content can
+    // handle finding these strings on its own. See bug 1488973
+    const {header, button_label} = await this._getExtraAttributes(originalMessage);
+
+    return {bundle: this._orderBundle(result), ...(header && {header}), ...(button_label && {button_label}), provider: originalMessage.provider, template: originalMessage.template};
```
k88hudson: maybe `extraTemplateStrings & extraTemplateStrings` or something

lib/ASRouter.jsm
sarracini: There's a reference to bug 1488973 in the comment above with a brief explanation. Want me to move it into this function?

PR #4441, opened 2018-09-21
Fix Bug 1492454: Normalise ASRouter frequency caps

AdamHillier:
This bug was caused by [this line](https://github.com/mozilla/activity-stream/blob/d2c948f0fddd54335fca6fcfab3c64436727b3d2/lib/ASRouter.jsm#L717) in the impression cleanup. We filter out all the provider impressions older than the longest cap period (from `getLongestPeriod`), and the provider cap period was given as `"daily"` i.e. not a number, so all the provider impressions were getting removed.

I think the most straightforward solution to this is to replace all aliases like `"daily"` with their appropriate numerical values when we load the providers/messages so that we don't have to consider these special cases each time we use the frequency caps.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -297,6 +297,18 @@ class _ASRouter {
     this.dispatchToAS(ac.BroadcastToContent({type: at.AS_ROUTER_PREF_CHANGED, data: ASRouterPreferences.specialConditions}));
   }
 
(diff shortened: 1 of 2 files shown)


AdamHillier:
Uh apparently I now can't add labels/reviewers to PRs - I guess anyone can review though

AdamHillier:
@sarracini I also can't merge things now, could you?

PR #4446, opened 2018-09-24
Backport Bug 1491394 - Update callsites to use new mozILocaleService API. r=jfkthame

sarracini:
Backport of bug https://bugzilla.mozilla.org/show_bug.cgi?id=1491394 landed

### DIFF
--- a/test/unit/lib/ActivityStream.test.js
+++ b/test/unit/lib/ActivityStream.test.js
@@ -255,7 +255,7 @@ describe("ActivityStream", () => {
     it("should be false with expected geo and unexpected locale", () => {
       sandbox.stub(global.Services.prefs, "prefHasUserValue").returns(true);
       sandbox.stub(global.Services.prefs, "getStringPref").returns("US");
-      sandbox.stub(global.Services.locale, "getAppLocaleAsLangTag").returns("no-LOCALE");
+      sandbox.stub(global.Services.locale, "appLocaleAsLangTag").get(() => "no-LOCALE");
 
       as._updateDynamicPrefs();
 
@@ -264,7 +264,7 @@ describe("ActivityStream", () => {
     it("should be true with expected geo and locale", () => {
       sandbox.stub(global.Services.prefs, "prefHasUserValue").returns(true);
       sandbox.stub(global.Services.prefs, "getStringPref").returns("US");
-      sandbox.stub(global.Services.locale, "getAppLocaleAsLangTag").returns("en-US");
+      sandbox.stub(global.Services.locale, "appLocaleAsLangTag").get(() => "en-US");
 
       as._updateDynamicPrefs();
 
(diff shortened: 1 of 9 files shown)


test/unit/lib/ActivityStream.test.js
```diff
@@ -255,7 +255,7 @@ describe("ActivityStream", () => {
     it("should be false with expected geo and unexpected locale", () => {
       sandbox.stub(global.Services.prefs, "prefHasUserValue").returns(true);
       sandbox.stub(global.Services.prefs, "getStringPref").returns("US");
-      sandbox.stub(global.Services.locale, "getAppLocaleAsLangTag").returns("no-LOCALE");
+      sandbox.stub(global.Services.locale, "appLocaleAsLangTag").get(() => "no-LOCALE");
```
sarracini: slightly different than the backport, since these are getters now the stubs needed updating in order for our tests to pass

test/unit/ping-centre/PingCentre.test.js
```diff
@@ -45,7 +45,7 @@ describe("PingCentre", () => {
 
     sandbox.stub(global.Services.prefs, "getBranch")
         .returns(new FakePrefs({initHook: prefInitHook}));
-    sandbox.stub(global.Services.locale, "getAppLocalesAsLangTags")
+    sandbox.stub(global.Services.locale, "appLocaleAsLangTag")
```
sarracini: For some reason this PingCentre test isn't in central

test/unit/unit-entry.js
```diff
@@ -107,8 +107,7 @@ const TEST_GLOBAL = {
   DownloadsViewUI: {DownloadElementShell},
   Services: {
     locale: {
-      getAppLocaleAsLangTag() { return "en-US"; },
-      getAppLocalesAsLangTags() {},
+      get appLocaleAsLangTag() { return "en-US"; },
```
Mardak: Hmm.. I guess as a getter, a missing `appLocalesAsLangTags` returns `undefined` which is what the original method was returning (but the method needed to be defined as trying to call `undefined` as a function would throw).. I guess this is fine 😉

test/unit/ping-centre/PingCentre.test.js
```diff
@@ -45,8 +45,8 @@ describe("PingCentre", () => {
 
     sandbox.stub(global.Services.prefs, "getBranch")
         .returns(new FakePrefs({initHook: prefInitHook}));
-    sandbox.stub(global.Services.locale, "getAppLocalesAsLangTags")
-        .returns([FAKE_LOCALE]);
+    sandbox.stub(global.Services.locale, "appLocaleAsLangTag")
+        .get(() => FAKE_LOCALE);
```
Mardak: Ha, I guess we only export activity stream stuff to newtab directory. Maybe we should just export this test as part of newtab especially once the tests are actually run from m-c.

sarracini: It would appear as though we're ignoring them on purpose? https://github.com/mozilla/activity-stream/blob/master/.mcignore#L17-L18

PR #4398, opened 2018-09-02
feat(experiments): add recipe executor

jonathankoren:
### Update

* Changed name of file
* Added comments describing how RecipeExecutor works

### Description
This code is part of the Pocket Personalization v2 experiment which uses content analysis to locally build interest profiles.

`RecipeExecutor` is essentially a library of operations to do machine learning feature engineering in the browser. These functions can be applied one after another to manipulate a Javascript object in order to build a user's interest vector, or to manipulate another object to make it comparable to an interest vector. The order the steps is controlled by a JSON file that is not part of this PR. This PR rather is simply to test the elementary functions, and to check the functional correctness of evaluating  steps.

Integrating this into the rest of Activity Stream will follow under a separate PR.

This code is dark.

### Testing
Unit tests
This code has no current consumers.

### Related
https://github.com/mozilla/activity-stream/pull/4294

### See Also
https://docs.google.com/document/d/12OtUZywivIvBnO3hmMNjptmIQ8cQLFOIOzJO4bRLqdg/edit

PR #4447, opened 2018-09-24
Fix Bug 1497616 - feat(experiments): Pocket Personalization V2

jonathankoren:
https://bugzilla.mozilla.org/show_bug.cgi?id=1497616

## Description

This is the third (and hopefully final) major PR for Pocket Personalization v2. This code provides that code to fetch tagging models and personalization recipes, feed them through the various taggers and RecipeExecutor. It builds and caches a personal interest vector from the user's browse history on the daily-idle thread, and then uses this to rescore recommended items provided by Pocket.

This code is controlled by an about:config preference that allows us to switch between no personalization, the preëxisting v1 personalization, and this new v2 personalization.

## Testing

All code is unit tested. Additionally, we tested this code locally using data pushed to dev remote settings and by reconfiguring about:config to activate v2. Using console logs, we saw that a reasonable interest vector was built, and items received from pocket were appearing in different orders when comparing off with v2. 

It's important to note, that you need to have a browse history of sufficient depth, in order to build a working interest vector. In testing, this was achieved by simply copying over the places.sqlite from our main profile to a test profile.

To activate v2, set the `browser.newtabpage.activity-stream.feeds.section.topstories.options` about:config preference to 
`{"api_key_pref":"extensions.pocket.oAuthConsumerKey","hidden":false,"provider_icon":"pocket","provider_name":"Pocket","read_more_endpoint":"https://getpocket.com/explore/trending?src=fx_new_tab","stories_endpoint":"https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?count=27&version=3&consumer_key=$apiKey&locale_lang=en-US&feed_variant=default_spocs_on","stories_referrer":"https://getpocket.com/recommendations","topics_endpoint":"https://getpocket.cdn.mozilla.net/v3/firefox/trending-topics?version=2&consumer_key=$apiKey&locale_lang=en-US","model_keys":["nmf_model_animals","nmf_model_business","nmf_model_career","nmf_model_datascience","nmf_model_design","nmf_model_education","nmf_model_entertainment","nmf_model_environment","nmf_model_fashion","nmf_model_finance","nmf_model_food","nmf_model_health","nmf_model_home","nmf_model_life","nmf_model_marketing","nmf_model_politics","nmf_model_programming","nmf_model_science","nmf_model_shopping","nmf_model_sports","nmf_model_tech","nmf_model_travel","nb_model_animals","nb_model_books","nb_model_business","nb_model_career","nb_model_datascience","nb_model_design","nb_model_economics","nb_model_education","nb_model_entertainment","nb_model_environment","nb_model_fashion","nb_model_finance","nb_model_food","nb_model_game","nb_model_health","nb_model_history","nb_model_home","nb_model_life","nb_model_marketing","nb_model_military","nb_model_philosophy","nb_model_photography","nb_model_politics","nb_model_productivity","nb_model_programming","nb_model_psychology","nb_model_science","nb_model_shopping","nb_model_society","nb_model_space","nb_model_sports","nb_model_tech","nb_model_travel","nb_model_writing"],"show_spocs":true,"personalized":true,"version":2}`

## Open Questions

We need some help checking the performance of this PR on various machines.

## Related
https://github.com/mozilla/activity-stream/pull/4294
https://github.com/mozilla/activity-stream/pull/4398


lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+    if (!this.interestVector ||
+      (Date.now() - this.interestVector.lastUpdate) >= STORE_UPDATE_TIME) {
+      this.interestVector = await this.createInterestVector();
+      this.interestVector.lastUpdate = Date.now();
+      this.store.set("interest-vector", this.interestVector);
+    }
+    this.initialized = true;
+  }
+
+  getRemoteSettings(name) {
```
ncloudioj: nit: the function name sounds more like a getter for the RemoteSettings property, but it's actually a lookup function on RemoteSettings. So let's rename it (e.g. `getFromRemoteSettings`) to avoid ambiguity.

ncloudioj: Another question about the RemoteSettings, are you going to use multiple Remote Settings buckets for this feature? Because the `name` parameter specifies the bucket name in `RemoteSettings(name).get()`. I believe the idiomatic way of using Remote Settings is that creating multiple records in a single bucket for this feature, and use `filters` to look up specific records. For instance,

Bucket name: "Pocket-Personalization"
```js
{
  "data": [
     {"name": "personality-provider-models", ...},
     {"name": "other-models", ...},
  ]
}
``` 

Then you can query it with 
```js
RemoteSettings("Pocket-Personalization").get({filters: {name: "personality-provider-models"}});
```

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+    return RemoteSettings(name).get();
+  }
+
+  getRecipeExecutor(nbTaggers, nmfTaggers) {
```
ncloudioj: nit: this function creates a new executor instead of returning an existing one. Better rename it to `generateRecipeExecutor()` or `createRecipeExecutor()`. Same applies to `getNaiveBayesTextTagger()` and `getNmfTextTagger()`.

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+      (Date.now() - this.interestVector.lastUpdate) >= STORE_UPDATE_TIME) {
+      this.interestVector = await this.createInterestVector();
+      this.interestVector.lastUpdate = Date.now();
+      this.store.set("interest-vector", this.interestVector);
+    }
+    this.initialized = true;
+  }
+
+  getRemoteSettings(name) {
+    return RemoteSettings(name).get();
```
ncloudioj: Although `RemoteSettings.get()` won't throw, it might return an empty array in certain cases (such as can't sync with the upstream, or IndexedDB failures etc), some care is needed here. Will comment separately at those places.

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Returns a Recipe from remote settings to be consumed by a RecipeExecutor.
+   * A Recipe is a set of instructions on how to processes a RecipeExecutor.
+   */
+  async getRecipe() {
+    if (!this.recipe) {
+      this.recipe = await this.getRemoteSettings("personality-provider-recipe");
+    }
+    return this.recipe[0];
```
ncloudioj: Note: `this.recipe[0]` could be `undefined`, add some special handling for that?

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Returns a Recipe Executor.
+   * A Recipe Executor is a set of actions that can be consumed by a Recipe.
+   * The Recipe determines the order and specifics of which the actions are called.
+   */
+  async generateRecipeExecutor() {
+    let nbTaggers = [];
+    let nmfTaggers = {};
+    const models = await this.getRemoteSettings("personality-provider-models");
```
ncloudioj: Again, `models` could be an empty array. Shall we short-circuit, or create a `RecipeExecutor` regardless?

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+      }
+    }
+    return this.getRecipeExecutor(nbTaggers, nmfTaggers);
+  }
+
+  /**
+   * Grabs a slice of browse history for building a interest vector
+   */
+  async fetchHistory(columns, beginTimeSecs, endTimeSecs) {
+    let sql = `SELECT *
```
ncloudioj: A couple of questions for this query:
* Perhaps you only want to fetch certain columns other than `SELECT *` here
* Columns like `description` and `title` could also be `NULL` other than `""`
* Do you want to set a limit on the return set? It's always a good idea to do so to avoid the unexpected behavior
* You can leverage the `params` of executePlaceQuery function for goodies such as specify the column names and set the limit etc, here is an [example](https://searchfox.org/mozilla-central/source/toolkit/modules/NewTabUtils.jsm#1092)

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+
+  getNaiveBayesTextTagger(model) {
+    return new NaiveBayesTextTagger(model);
+  }
+
+  getNmfTextTagger(model) {
+    return new NmfTextTagger(model);
+  }
+
+  getNewTabUtils() {
```
ncloudioj: Why wrapping NewTabUtils here? You can simply just do 'await NewTabUtils.activityStreamProvider.executePlacesQuery'

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,197 @@
...
+    WHERE last_visit_date >= ${beginTimeSecs * 1000000}
+    AND last_visit_date < ${endTimeSecs * 1000000}`;
+    columns.forEach(requiredColumn => {
+      sql += ` AND ${requiredColumn} <> ""`;
+    });
+
+    const {activityStreamProvider} = this.getNewTabUtils();
+    const history = await activityStreamProvider.executePlacesQuery(sql, {
+      columns,
+      params: {},
```
ncloudioj: See the comment above about the usage `params`.

lib/TopStoriesFeed.jsm
```diff
@@ -511,6 +526,23 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.init();
   }
 
+  processAffinityProividerVersion(data) {
```
ncloudioj: This function wants some comments for the better readability.

lib/TopStoriesFeed.jsm
```diff
@@ -578,6 +610,18 @@ this.TopStoriesFeed = class TopStoriesFeed {
         if (action.data.name === "pocketCta") {
           this.dispatchPocketCta(action.data.value, true);
         }
+        if (action.data.name === OPTIONS_PREF) {
+          try {
+            const options = JSON.parse(action.data.value);
+            if (this.processAffinityProividerVersion(options)) {
+              await this.clearCache();
+              this.uninit();
```
ncloudioj: What's the purpose of uninit then re-init upon the version pref change? If it's merely for resetting some states, this sounds a bit overkill to me. Another question is do we really need to support two versions of personalization at the same time? Even if the answer is yes, that could be done separately instead of complicating this experiment.

lib/TopStoriesFeed.jsm
```diff
@@ -85,10 +82,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
     SectionsManager.onceInitialized(this.onInit.bind(this));
   }
 
-  observe(subject, topic, data) {
+  async observe(subject, topic, data) {
     switch (topic) {
       case "idle-daily":
-        this.updateDomainAffinityScores();
+        await this.updateDomainAffinityScores();
```
ncloudioj: With the typical parameter settings, have you done any benchmarks for this feature to get some basic idea about the typical required running time? Even better to do this on the reference machine. It's hard for me to review this part without those inputs.

The V1 personalization is currently being computed in the `idle-daily` handler since we are aware of its typical calculation duration is within hundreds of ms. Assuming V2 needs longer, computing it in daily-idle perhaps is not optimal anymore due to the fact that daily-idle is already crowded with other players such as daily maintenance for Places and IndexedDB quota manager.

lib/PersonalityProvider.jsm
ScottDowne: I got into the habit of doing this for globals for test purposes because a lot of things (example remote-settings) are globals I cannot set as something else in unit tests. While I only needed to wrap remote settings too make stubbing possible, the rest was just for convenience to make it easier to stub.

lib/TopStoriesFeed.jsm
jonathankoren: We do need this 3 way comparison. Off is our baseline for organic recommendations, but v1 is the baseline for sponsored content. This allows us to check the behavior across both. If we reach parity with v1 we plan to kill it completely in Q1.

ScottDowne: Shield doesn't change prefs right away, so if this pref changes, it means there is already a moment where v1 or a non personalized version has started to be generated. It is my understanding (based on other areas of the code) that the only way to regenerate everything with new pref is an uninit/init.

lib/PersonalityProvider.jsm
ScottDowne: short-circuit?

ncloudioj: In case of `getRemoteSettings` returns an empty array (we've seen this happened in prod), shall we just bail out instead of generating a malformed recipe executor?

lib/PersonalityProvider.jsm
ncloudioj: That's a fair point. `NewtabUtils.activityStreamProvider` already has a default stub defined in `test/unit-entry.js`, is it possible to reuse that here?

lib/PersonalityProvider.jsm
ScottDowne: I currently have two collections, 1 is "personality-provider-recipe" which there is really only 1 record for. The other is personality-provider-models which has many records.

The only reason I kept them separate is it looked like collections were expected to have similar data structures. personality-provider-recipe was something different than personality-provider-models, so it didn't look correct to add personality-provider-recipe and personality-provider-models to the same collection. I'm a bit confused though.

lib/PersonalityProvider.jsm
ScottDowne: These were just here to make testing easier and consistent with the only way I could stub remote settings. If it's causing issues, it's probably best if I just remove all of these except for `getFromRemoteSettings` and just stub them in as globals int he tests.

lib/PersonalityProvider.jsm
ScottDowne: @jonathankoren thoughts on this one? I think you're more familiar with this function currently.

lib/PersonalityProvider.jsm
ScottDowne: Ah, I'm not sure. If we bail, I think things are going to be pretty wonky. I think the safest thing to do is return empty arrays, which if it doesn't already, I can ensure causes everything to do nothing.

It would end up generating something as if it was a fresh profile with no history.

lib/PersonalityProvider.jsm
ScottDowne: Are we sure it'll only ever return an empty array or data? Part of me wants to do something like this just in case.

`return RemoteSettings(name).get() || [];` just to be sure.

ScottDowne: Oh wait, this is a promise that above wont work quite like that.

lib/PersonalityProvider.jsm
ncloudioj: Okay, looks like having two separate buckets is expected here.

lib/PersonalityProvider.jsm
ncloudioj: Yep, see here https://searchfox.org/mozilla-central/source/services/settings/remote-settings.js#345

lib/PersonalityProvider.jsm
ncloudioj: Yep, I was wondering that if we could just abort this round completely until the Remote Settings is ready. @jonathankoren what do you think?

This also reminds me that shall we introduce some mechanism to trigger this whole personalization. Like if RS is not ready or Places doesn't return enough records, then just wait for next cycle.

ScottDowne: Hm, what does it mean when you say "RS is not ready"? Does it go through a state change where, it's not ready, and using it is slow, and it is ready and using it is fast, or something?

lib/PersonalityProvider.jsm
ScottDowne: Yeah, I'm removing all these get functions cept for the needed one, which was renamed. I'm using stubs in the tests instead.

lib/PersonalityProvider.jsm
ncloudioj: In short, `RS.get()` might run into some trouble in the wild. For example, this is what we've observed in production "NetworkError when attempting to fetch resource". It printed this error message and returned an empty array, perhaps would retry it later.

lib/PersonalityProvider.jsm
jonathankoren: Fixing the query to explicitly avoid NULL is definable something we should do. 

I wouldn't put a LIMIT on it, because we have no idea what the upper limit of pages should be. Internal polling on our how histories at Pocket showed that the `WHERE description <> ""` has the habit of reducing the total number of URLs in the history somewhere between 75% and 80%.

The reason why we run SELECT * is that we just wanted to make all the columns available. We ran into a problem before we went down the JS route, where not all the columns (specifically `description`) were not available on the browse history API available to extensions. This future proofs us.

ncloudioj: Yes, settings a very high LIMIT is definitely better than nothing. I believe you can find a good tradeoff from the model's minimal requirement and the computing limitation. 

> The reason why we run SELECT * is that we just wanted to make all the columns available.

Do you really need to make all the columns available? Explicitly specifying the needed columns has at least two benefits:
* It significantly reduces the IO for the big query results, there is no point to load columns without using them
* It makes the code self-described about what Places columns are in use here

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,198 @@
...
+      }
+      this.interestVector.lastUpdate = Date.now();
+      this.store.set("interest-vector", this.interestVector);
+    }
+    this.initialized = true;
+  }
+
+  async getFromRemoteSettings(name) {
+    const result = await RemoteSettings(name).get();
+    return result || [];
```
ncloudioj: `|| []` is unnecessary here. `RemoteSettings().get()` _always_ returns an array, even an empty one upon errors.

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Returns a Recipe from remote settings to be consumed by a RecipeExecutor.
+   * A Recipe is a set of instructions on how to processes a RecipeExecutor.
+   */
+  async getRecipe() {
+    if (!this.recipe) {
```
ncloudioj: Hmm, looks like there is a corner case here, assume that the first call of `this.recipe = await this.getFromRemoteSettings("personality-provider-recipe");` fails somehow, hence `this.recipe` will be an empty array. Then that user switches to the old version of personalization temporarily, then switches it back to V2 again. That will call the `init` function, which in turn calls `getRecipe()`. At this point, `this.recipe` is still an empty array, so it will skip the if statement without retrying the `RemoteSettings.get()`, which might be ready to query now.

Although this is rare, perhaps change it to `if (this.recipe === undefined || !this.recipe.length)`?

ScottDowne: How about `if (!this.recipe || !this.recipe.length)`?

Or was the `=== undefined` there to catch something specific?

ScottDowne: but this is a good catch. The if statement there was actually an early quick addition. Should probably be a bit more sophisticated, I just lost it in the shuffle.

ncloudioj:
The code looks good to me now. Need to figure out a way to try it out locally. Meanwhile, I'd like to share a few thoughts on this feature:

* I believe that we want to add some telemetry probes to measure the key metrics, such as calculation duration, number of records fetched from Places, success/fail calculation rates, etc. You can reference the implementation of personalization V1 for that
* We can't export this code until the bucket(s) get created in production Remote Settings 
* We might also want to add some basic mechanism to avoid unnecessary calculation, for instance, if Places returns the same (or almost the same) history records, we can simply skip it. Off the top of my head, we can compare the total number of records as well as the last visit timestamp with that of last run
* I am still concerned about the performance impact of this calculation. Worst case scenario, we manage to fetch all the 30000 records from places, then crunch them one by one, and finally persistent the result. That whole process will inevitably introduce a huge spike both to CPU and I/O usage. Telemetry would provide us with more details on that, though we should be aware of that impact. If that's the common case, as I mentioned earlier, we will have to split this task into multiple sub-tasks and complete them in a series of idleDispatchHandlers. That would also force us to implement this algorithm in an incremental fashion. Again, we don't need to do that now, just for your reference, telemetry will tell us how it performs in the wild
* We need to think about how to respond to user's delete history request. Again, not necessarily for this experiment, but we need to properly cleanup their interest vector upon their request of history deletion in release.

ScottDowne:
@ncloudioj

I'll add some instructions to show how to test this locally in the next comment.

I'm in the process of getting this on prod/staging remote settings. You know roughly how long that'll take?

+1 to telemetry and unnecessary calculation. Do you think that's something  we want for this test, or more for later?

Yeah we've talked about how to optimize this performance going further.

For me when profiling this, the largest spike seems to be fetching things from remote setting or cache, and hasn't been from calculating things. I would be interested to see what others find.

I agree we need an incremental solution for this. I also wonder if it's worth exploring using an incomplete picture during these incremental builds, as still being better than nothing.

So far the biggest gains we've seen from performance is doing it all async and notifying things when it's done using `doContentUpdate` and caching taggers and rehydrating the recipe executor with those taggers allowing us to avoid re fetching the models.

ScottDowne:
To test:

1. Copy a places.sqlite from an existing profile with some history to a new profile. (this one might be optional depending on how much history you want to test with)
2. You might need to remove a places.sqlite-wal if it exists.
3. Add these to about:config (You'll probably need to create services.settings.verify_signature):
```
    services.settings.server : https://kinto.dev.mozaws.net/v1
    services.settings.verify_signature : false
```
4. Run this in the debugger to sync up with remote settings (I've already got the dev server updated and keep it daily until staging is good to go (should be done with staging today)):
```
const { RemoteSettings } = ChromeUtils.import("resource://services-settings/remote-settings.js", {});
await RemoteSettings.pollChanges();
```
5. You might need to force an idle-daily to update stuff, if you think you need to, this is how, although might not be needed:
```
Services.obs.notifyObservers(null, "idle-daily");
```
6. Set version to 2 in the pref feeds.section.topstories.options
7. Run the browser with the above created profile. Open a new tab. Initially the recs shouldn't be personalized, because it's still running.
8. After a few seconds (for me) if you refresh or open a new tab, the order or recs should have changed.

ncloudioj:
@ScottDowne 

Setting up Remote Settings usually takes several days, they often ask us to test in stage first (you can create new bucket on your own in stage), then file the deploy bug for prod.

Yes, I think it makes sense to start collecting telemetry in this experiment, it will allow us to understand how the new algorithm performs as well as the key metrics of the calculation. I'd recommend to add various `performance` pings to record:
* duration of the calculation
* state of the calculation (succeeded or failed for whichever reason)
* # of records fetched from Places
* interest vector related information

jonathankoren:
I would not really worry about the 30000 records. It took me a year to get many records. Also, since we're getting like 20%, You'd have to visit 150,000 pages in 30 days. That's 5000 pages a day. That's 1 every 17 seconds, 24/7. Yes, everything happens on the Internet, but that's not human behavior.

But your bigger point about telemetry is very good.

ncloudioj:
Managed to test it locally, it worked as expected. A few more observations that we should pay attention:

* The bucket size for `personality-provider-models` is substantially bigger than other buckets in Remote Settings, and it took a while for RS to download it. we need to make sure that non-experiment users won't download this bucket once we deploy it to production. @leplatrem can you confirm this? This feature is behind a pref and will be run as an shield experiment, will RS sync this bucket even for those non-experiment users?  
* It generated a 20MB result file in the persistent cache based on my profile, this is a bit concerning. Remember that Activity Stream will load this file in memory and parse the JSON file during the initialization, there was a noticable jank when I restarted browser. Let's see if we can find a way to reduce the size of this output

ScottDowne:
@ncloudioj and @leplatrem I looked at both attachments and target filters, to try and find a way to make this only downloaded for users in the test. "non-experiment users won't download this bucket once we deploy it to production"

It looks like target filters is our best option. https://remote-settings.readthedocs.io/en/latest/target-filters.html

I'm not totally sure if attachments would do this or not, but it also brings another layer of complexity that concerns me because of the time until code freeze.

leplatrem:
> we need to make sure that non-experiment users won't download this bucket once we deploy it to production. 

I don't fully understand: why is that a problem that non-experiment users download the collection of records? Does it have many records? Does it have a few huge records?

When the synchronization happens, the content of the remote collection is downloaded entirely by the client if:
- a remote settings client was instantiated for this collection (a call to `RemoteSettings("cid")` was made)
- some data exist in the local database (chrome indexeddb)
- a local JSON dump exists in `services/settings/dumps/`

If what worries you is the time to download the dataset, note that it can be lightened by shipping a JSON dump of the PROD collection in `services/settings/dumps/`. But every user will have it.

Also, the target filters will only affect the list returned by `.get()`, it won't affect the amount of downloaded data or local database size. Note that the whole list will be iterated on `.get()` in order to execute the JEXL filter on each record (it is not cached on purpose) before returning the filtered list.

The most obvious ways that come in my mind:
1. Offload the data into attachments. Every users will download the metadata, only activated users will download the heavy stuff.
2. Use a pref flip, OFF by default, and activated using Normandy for example. If the RS client is never instantiated with pref OFF, users will never have a local dump nor a local database and therefore will not download the "unknown" collection.
3. Use two collections: a light one that everyone downloads. A big one that only elected users will synchronize.

Does it make sense?

ScottDowne:
@leplatrem Option 2 looks like the fastest path forward. Kinda ideal actually. If I understand you, if the pref is off, and a get on the collection is never made for that user, they never get the records downloaded?

and yeah, the collections are kinda big. There are 57 records, which totals around 20MB. It sounds like we're hoping this is ok for the experiment (a third of a percent), but just need to be sure it'll only download 20ish MB for the third of a percent, and not for 100 percent.

I'll give this a test today, and if I don't see the records we should be good?

ScottDowne:
@leplatrem and @ncloudioj and @jonathankoren I tested the above 2. solution, and it seems to be working.

As long as a call to get on our collection is never made, it'll never start downloading the records. And given all our get calls are behind the pref check, we should be good here. :)

jonathankoren:
@ScottDowne  I'm cool with this if @ncloudioj  is.

leplatrem:
> If I understand you, if the pref is off, and a get on the collection is never made for that user, they never get the records downloaded?

Indeed. If the collection does not have any client registered, no local dump and nothing in local database, then it's not downloaded.

You can double check but we already have collections for Focus and Rocket that already leverage that behavior.

> There are 57 records, which totals around 20MB. 

That's a bit extreme, but apart the memory impact when calling `.get()`, there should not be any major problem. IndexedDB should be able to handle that.

leplatrem:
> That's a bit extreme, but apart the memory impact when calling .get(), there should not be any major problem. IndexedDB should be able to handle that.

I forgot to mention that the signature verification is probably going to be laborious on old machines, since we will build a canonical JSON of the whole dataset and compute a cryptographic signature of it.

ncloudioj:
> Indeed. If the collection does not have any client registered, no local dump and nothing in local database, then it's not downloaded.

@leplatrem Thanks for confirming this!

@ScottDowne @jonathankoren Could you take one more look at the size of the persistent cache? Is it possible to reduce its size?

jonathankoren:
We’re not going to be getting these files smaller before code freeze.

--
jonathankoren
jonathan@jonathankoren.com

Mardak:
At least previous values I've seen have values like `-0.6901305870205201` and `-5.994137996346771`. Are they all negative and all need that much precision? E.g., potentially transforming those to `.69013058` and `5.9941379`

Mardak:
Just a quick sanity check on NB_space.json:
```
   24375 orig.json
    8419 orig.json.gz
   16720 small.json
    5705 small.json.gz
```
Where small removed the trailing 8 digits: `s/[0-9]{8}([,\]])/\1/g`. Roughly 33% smaller for both text and gz. No clue about accuracy / behavior changes 😉

jonathankoren:
@mardak I already did the 30% reduction by dropping digits. Those files are the ones we had uploaded to dev RS. ;)

jonathankoren:
https://bugzilla.mozilla.org/show_bug.cgi?id=1497616

Mardak:
Is there cleaning up of Remote Settings when a user finishes the experiment (I'm guessing version changes from 2 to 1)? Will there be continued syncing even when not using v2 anymore?

ScottDowne:
@Mardak that would probably be wise, but not sure how to do that.

ncloudioj:
Data review granted https://bugzilla.mozilla.org/show_bug.cgi?id=1497616#c10

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,249 @@
...
+    }
+    this.interestVector = this.interestVector || await this.createInterestVector();
+    if (!this.interestVector) {
+      this.dispatch(ac.PerfEvent({event: "PERSONALIZATION_V2_CREATE_INTEREST_VECTOR_ERROR"}));
+      return;
+    }
+
+    this.dispatch(ac.PerfEvent({
+      event: "PERSONALIZATION_V2_TOTAL_DURATION",
+      value: Math.round(perfService.absNow() - this.perfStart),
```
ncloudioj: Why initialize `this.prefStart` in the constructor? It'll be incorrect if another `init` call gets triggered. I believe a local `perfStart` should be good enough?

lib/PersonalityProvider.jsm
```diff
@@ -0,0 +1,249 @@
...
+      }
+    }
+
+    const finalResult = this.recipeExecutor.executeRecipe(
+      interestVector,
+      this.interestConfig.interest_finalizer);
+
+    this.dispatch(ac.PerfEvent({
+      event: "PERSONALIZATION_V2_CREATE_INTEREST_VECTOR_DURATION",
+      value: Math.round(perfService.absNow() - start),
```
ncloudioj: You start the timer before `fetchHistory`, is that intended? If not, let's move it down a bit. Otherwise, ignore this comment.

ScottDowne: Yeah it was my initial intent to track the time the whole function takes, but I'm open to tracking the duration of just the calculation. Really I didn't have a strong opinion on it, so either or is fine with me.

lib/PersonalityProvider.jsm
ScottDowne: Yeah this can be updated. I see no issues with this.

PR #4451, opened 2018-09-26
Fix Bug 1494275 - CFR can wrongly recommend outdated versions of the addons

sarracini:
Fetch the correct url for the latest xpi for an addon given it's GUID.

### DIFF
--- a/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
+++ b/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
@@ -100,6 +100,12 @@
     "addon": {
       "type": "object",
       "properties": {
+        "id": {
+          "allOf": [
+            {"$ref": "#/definitions/plainText"},
+            {"description": "Unique addon ID"}
+          ]
+        },
         "title": {
           "allOf": [
             {"$ref": "#/definitions/plainText"},
@@ -212,10 +218,9 @@
                 "data": {
                   "properties": {
                     "url": {
-                      "allOf": [
-                        {"$ref": "#/definitions/linkUrl"},
-                        {"description": "URL used in combination with the primary action dispatched."}
-                      ]
+                      "type": "null",
+                      "$comment": "This is dynamically generated from the addon.id. See CFRPageActions.jsm",
+                      "description": "URL used in combination with the primary action dispatched."
                     }
                   }
                 }
--- a/lib/CFRPageActions.jsm
+++ b/lib/CFRPageActions.jsm
@@ -3,14 +3,17 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
+ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
 ChromeUtils.import("resource://gre/modules/Localization.jsm");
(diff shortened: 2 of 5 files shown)


content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
```diff
@@ -212,10 +218,8 @@
                 "data": {
                   "properties": {
                     "url": {
-                      "allOf": [
-                        {"$ref": "#/definitions/linkUrl"},
-                        {"description": "URL used in combination with the primary action dispatched."}
-                      ]
+                      "type": "null",
```
k88hudson: shouldn't a link URL still be allowed?

content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
```diff
@@ -212,10 +218,8 @@
...
+                      "description": "URL used in combination with the primary action dispatched."
```
k88hudson: You could add a note about how this can be dynamically generated from the addon.id, see (filename)

lib/CFRPageActions.jsm
```diff
@@ -431,7 +460,18 @@ const CFRPageActions = {
...
+
+    // If this is a CFR for an add-on fetch the url for the latest xpi
+    if (template === "cfr_doorhanger") {
+      const url = await this._fetchLatestAddonVersion(content.addon);
+      // If we failed to get a url to the latest xpi, do not show a recommendation at all
+      if (!url) {
+        return false;
+      }
+      // Update the action's data with the url to the latest xpi
+      content.buttons.primary.action.data.url = url;
```
k88hudson: modifying this object directly might cause side effects so is generally not a great idea, although updating a property this deep is gonna be a LOT of `...`

lib/CFRPageActions.jsm
```diff
@@ -396,16 +399,69 @@ const CFRPageActions = {
     }
   },
 
+  /**
+   * Fetch the URL to the latest add-on xpi so the recommendation can download it.
+   * @param id             The add-on ID provided by the CFRMessageProvider
```
k88hudson: Right now this is wrong, the first param is an object

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -640,6 +640,23 @@ describe("CFRPageActions", () => {
...
       });
+      it("should add the right url if we fetched and addon install URL", async () => {
+        sinon.stub(CFRPageActions, "_fetchLatestAddonVersion").returns(Promise.resolve("latest-addon.xpi"));
+        fakeRecommendation.template = "cfr_doorhanger";
+        await CFRPageActions.addRecommendation(fakeBrowser, fakeHost, fakeRecommendation, dispatchStub);
+        const recommendation = CFRPageActions.RecommendationMap.get(fakeBrowser);
+        assert.equal(recommendation.content.buttons.primary.action.data.url, "latest-addon.xpi");
+
+        // sanity check - just go through some of the rest of the attributes to make sure they were untouched
+        assert.equal(recommendation.id, fakeRecommendation.id);
```
k88hudson: you could maybe just do a `deepEquals` here to check the whole thing

PR #4456, opened 2018-09-28
chore(l10n): Update from l10n-central Fri Sep 28 2018 13:46:11 GMT-0400 (EDT)

k88hudson:


k88hudson:
Hm, looks like there's an issue with the `is` locale failing a test. Maybe lets hold off on this until we figure out what the issue is.

PR #4459, opened 2018-10-01
Fix Bug 1488462 - Big slowdown happening every x minutes

Mardak:
r?@sarracini Just process locally by grabbing the first result and adding type.


Mardak:
Doh! That's what I get for jumping between firefox-63 branch and master without updating!

PR #4460, opened 2018-10-02
Fix Bug 1495546 - Localization fails in OnboardingMessageProvider if we've already localized strings before

sarracini:
I was shallow copying `msg` before and so it was changing `ONBOARDING_MESSAGES` and then the next time we looked up the strings and passed them into `L10N` they no longer had `string_id` and the function would throw, giving us an error.
I'm deep copying now - any more elegant solutions for deep copying the entire nested object @k88hudson ?

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -8,7 +8,7 @@ const L10N = new Localization([
   "browser/newtab/onboarding.ftl",
 ]);
 
-const ONBOARDING_MESSAGES = [
+const ONBOARDING_MESSAGES = () => ([
   {
     id: "ONBOARDING_1",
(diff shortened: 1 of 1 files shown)


lib/OnboardingMessageProvider.jsm
```diff
@@ -93,7 +93,8 @@ const OnboardingMessageProvider = {
   async translateMessages(messages) {
     let translatedMessages = [];
     for (const msg of messages) {
-      let translatedMessage = msg;
+      // deep copy 'msg' to avoid changing original ONBOARDING_MESSAGES
+      let translatedMessage = JSON.parse(JSON.stringify(msg));
```
k88hudson: As we discussed let's change the const to a function instead

PR #4462, opened 2018-10-02
Fix Bug 1477633 - Add tooltip text to dismiss button

piatra:
Added `block_button_text` as alt text to the dismiss button. The translation and english fallback should come from snippets endpoint.

@glogiotatidis r?

### DIFF
--- a/content-src/asrouter/components/SnippetBase/SnippetBase.jsx
+++ b/content-src/asrouter/components/SnippetBase/SnippetBase.jsx
@@ -19,14 +19,14 @@ export class SnippetBase extends React.PureComponent {
       return (
         <div className="footer">
           <div className="footer-content">
-            <button className="ASRouterButton secondary" onClick={this.props.onDismiss}>{this.props.content.dismiss_button_label}</button>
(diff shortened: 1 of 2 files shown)


content-src/asrouter/components/SnippetBase/SnippetBase.jsx
```diff
@@ -23,7 +23,7 @@ export class SnippetBase extends React.PureComponent {
       <div className="innerWrapper">
         {props.children}
       </div>
-      <button className="blockButton" onClick={this.onBlockClicked} />
+      <button className="blockButton" title={this.props.content.block_button_text} onClick={this.onBlockClicked} />
```
glogiotatidis: Can we make this to default to `Remove this`?

piatra: We didn't want to add fallbacks in the templates because right now we don't have an option to provide updates to them. I think it would be best if snippets would provide them until we get https://bugzilla.mozilla.org/show_bug.cgi?id=1433218

PR #4466, opened 2018-10-04
Fix Bug 1496357 - Allow links to trigger actions

piatra:
I've also updated the schema and the way actions are triggered.
Until now they depended on the `button_action` payload to be an object with `{type, target}` structure. I've changed that so that we now construct that object ourselves and added `target` and `button_action_target` attributes for anchors and buttons.
Now they both work in the same way.


content-src/asrouter/asrouter-content.jsx
```diff
@@ -104,7 +104,15 @@ const ALLOWED_TAGS = {
...
     return Object.keys(links).reduce((acc, linkTag) => {
-      acc[linkTag] = <a href={safeURI(links[linkTag].url)} data-metric={links[linkTag].metric} onClick={sendClick} />;
+      const {action} = links[linkTag];
+      // Setting the value to false will not include the attribute in the anchor
+      const url = action ? false : safeURI(links[linkTag].url);
+
+      acc[linkTag] = (<a href={url}
+        data-metric={links[linkTag].metric}
+        data-action={action}
+        data-target={links[linkTag].target}
```
sarracini: is it possible to not call it `target` ? it's a bit confusing because `UITour.showMenu` also takes a `target.browser.ownerGlobal` as the first parameter, and then `action.data.target` as the second, and it's a bit confusing that the two targets mean different things here

PR #4467, opened 2018-10-04
Fix Bug 1496167 - Add installation source to addons installed from activitystream

rpl:
This PR adds an additional `telemetryInfo` parameter (introduced in [Bug 1433334](https://bugzilla.mozilla.org/1433334)) to the `AddonManager.getInstallForURL` call (and add an additional assertion in the existing unit test that cover that ActivityStream feature).

This additional `telemetryInfo` parameter ensures that we can store the original "installation source" and be able to include it into the addonsManager telemetry events related to the addons installed from ActivityStream (e.g. when the addon has been disable/enabled/uninstalled by the users, even across Firefox restarts).

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -215,7 +215,12 @@ const MessageLoaderUtils = {
     try {
       const aUri = Services.io.newURI(url);
       const systemPrincipal = Services.scriptSecurityManager.getSystemPrincipal();
-      const install = await AddonManager.getInstallForURL(aUri.spec, "application/x-xpinstall");
+
+      // AddonManager installation source associated to the addons installed from activitystream
+      // (See Bug 1496167 for a rationale).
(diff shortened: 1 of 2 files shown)


rpl:
@k88hudson this is the PR that we have discussing about over IRC.

k88hudson:
@rpl this looks good, is there a manual test step I can try to verify this?

rpl:
@k88hudson sure, here is an STR to manually verify it:

- trigger the addon installation from ActivityStream (and allow the addon to be installed)
- open `about:telemetry#events-tab` in a tab
- look for the telemetry events with `addonsManager` category
- check that the telemetry events has the expected value set on the `source` in the `extra` column.
 
(as a additional verification you may also enable, disable and/or uninstall the addon and then check that the related telemetry events collected also have the expected value on the `source` extra var).

PR #4470, opened 2018-10-04
bug 1496246 - Update Pocket CTA implementation with feedback from Leg…

ScottDowne:
…al and UX

### DIFF
--- a/content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
+++ b/content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
@@ -5,8 +5,8 @@
   margin-inline-end: 20px;
   display: inline-block;
 
-  a {
-    background: $pocket-red;
+  .pocket-cta-button {
+    background: $blue-60;
     letter-spacing: -0.34px;
     color: $white;
     border-radius: 4px;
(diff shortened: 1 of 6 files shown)


locales/en-US/strings.properties
```diff
@@ -145,6 +145,7 @@ pocket_read_more=Popular Topics:
 pocket_read_even_more=View More Stories
 pocket_more_reccommendations=More Recommendations
 pocket_learn_more=Learn More
```
k88hudson: Is this being used anywhere else? If not, please remove it

ScottDowne: Removed

k88hudson:
Personally I find this colour combination a little weird, is that what's intended?
image

content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
```diff
@@ -5,8 +5,8 @@
   margin-inline-end: 20px;
   display: inline-block;
 
-  a {
-    background: $pocket-red;
+  .pocket-cta-button {
```
k88hudson: Looking at the invision spec, it looks like padding of `0 9px` is closer

ScottDowne:
Updating. I double checked with Aaron, and he said the desired padding was 8px so went with that. Super close to 9 though :)

PR #4474, opened 2018-10-05
Merge central into master

k88hudson:
I re-wrote the history of `central` to deal with the divergence, what do you think?


Mardak:
Oh, and closed PR because we turned off merge commits as normal PRs want rebasing/squashing. So.. we'll just push central <-> master merges directly.

PR #4475, opened 2018-10-05
Bug 1488846 - Ensure pocket cta doesn't flash between logged in and o…

ScottDowne:
…ut while loading, and ensure content doesn't shift when adding cta.

### DIFF
--- a/common/Reducers.jsm
+++ b/common/Reducers.jsm
@@ -44,7 +44,7 @@ const INITIAL_STATE = {
   },
   Sections: [],
   Pocket: {
-    isUserLoggedIn: false,
+    isUserLoggedIn: null,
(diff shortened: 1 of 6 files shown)


ScottDowne:
I'm having a hard time charry-picking this from activity-stream firefox-63 to mozilla-central bookmarks/beta.

A BUNCH of changes I'm not responsible for come with it, and not sure what commit is causing those changes. I've tried to do the cherry-pick from various commits on both activity stream and mozilla cental, and it's doing some weird stuff. I think something has changed from when I did my last uplift, and not sure what yet.

Mardak:
https://hg.mozilla.org/releases/mozilla-beta/rev/b505048f7843

PR #4478, opened 2018-10-08
Fix Bug 1492085 - Use the Newsletter template for FxA signup

piatra:
* adds a `form_method` field to the schema (FxA template uses `GET`, newsletter template uses `POST`)
* prevents the form from being submitted twice in a row

Snippet to test `https://gist.githubusercontent.com/piatra/d193ca7e0f513cc19fc6a1d396c214f7/raw/ec505c9153f3ae8a7e661502fb166d3a61802bbb/newsletter_snippet.json`


content-src/asrouter/templates/NewsletterSnippet/NewsletterSnippet.schema.json
```diff
@@ -43,6 +43,11 @@
       "type": "string",
       "description": "Endpoint to submit form data."
     },
+    "form_method": {
```
k88hudson: Is it really necessary to make this user configurable? Instead, why not just do something like this:

```js
// Where NewsletterSnippet is renamed to the generic SubmitFormSnippet:
const NewsletterSnippet = props => <SubmitFormSnippet {...props} form_method="POST" />;
const FXASnippet = props => <SubmitFormSnippet {...props} form_method="GET" />;
...
```

k88hudson:
Can you make sure this matches the fields in https://github.com/mozmeao/snippets/blob/master/activity-stream/newsletter-subscribe.html?

PR #4479, opened 2018-10-08
Fix Bug 1494101 - Telemetry for custom url category

rlr:


docs/v2-system-addon/data_events.md
```diff
@@ -31,6 +31,30 @@ where the "value" is encoded as:
...
 
+## Page takeover ping
+
+This is a ping indicating whether Activity Stream is currently being used on about:home and about:newtab. If either about:home or about:newtab is set to a custom URL, the category of the URL is included. It's submitted once upon Activity Stream initialization.
+
+```js
+{
+  "event": "PAGE_TAKEOVER_DATA",
+  "value": {
+    "home_pref": ["default", "about-blank", "web-extension", "custom-url"],
```
ncloudioj: Since "default", "about:blank", and "web-extension" are already covered by the heartbeat ping, perhaps we are only interested in the "custom-url" for about:home and about:newtab here? That said, this ping is only sent when either about:home or about:newtab was replaced by a custom URL. It will simplify the implementation a bit below, also save a ping for those users who stick with the default setting.

What do you think?

docs/v2-system-addon/data_events.md
```diff
@@ -31,6 +31,30 @@ where the "value" is encoded as:
...
+    "home_url_category": [undefined, "search-engine", "search-engine-mozilla-tag", "search-engine-other-tag", "news-portal", "ecommerce", "social-media", "other"],
```
ncloudioj: Any difference between `undefined` and `other`? If it's intended to keep them separately, we have to change `undefined` to a string, like `unset` or `unknown`, because our database doesn't support the `undefined` type.

ncloudioj: nit: let's use `|` instead of `,` in those arrays.

lib/SiteClassifier.jsm
```diff
@@ -0,0 +1,62 @@
...
+      break;
+    } catch (e) {}
+  }
+
+  if (parsedURL) {
+    // If we parsed successfully, find a match.
+    const hostname = parsedURL.hostname.replace(/^www\./i, "");
+    const params = parsedURL.searchParams;
+    const siteTypes = await RS("sites-classification").get();
+    const sortedSiteTypes = siteTypes.sort((x, y) => y.weight - x.weight);
```
ncloudioj: If we check the type of URL in the ascending order in weight, and return immediately once we find a match. Will we miss the chance to get the more specific matches with higher weight? For example, say we have two types defined as:

```js
{
  "type": "host-match",
  "critiria": {
     "hostname": "foobar.com"
   }
  "weight": 100,
}

{
  "type": "host-and-params-must-match",
  "critiria": {
     "hostname": "foobar.com",
     "params": ...
   }
  "weight": 200,
}
```

Given the `foobar.com/must/match`, it'll return type as `host-match`. Shall we check the type in descending order then?

lib/SiteClassifier.jsm
```diff
@@ -0,0 +1,62 @@
...
+      parsedURL = new URL(_url);
+      break;
+    } catch (e) {}
+  }
+
+  if (parsedURL) {
+    // If we parsed successfully, find a match.
+    const hostname = parsedURL.hostname.replace(/^www\./i, "");
+    const params = parsedURL.searchParams;
+    const siteTypes = await RS("sites-classification").get();
```
ncloudioj: Let's add a comment here to mention that a local copy will be added to m-c as a fallback for this bucket, so we don't need to handle the case where `RS.get()` returns an empty array.

rlr:
datareview+ @ https://bugzilla.mozilla.org/show_bug.cgi?id=1494101#c10

PR #4482, opened 2018-10-09
Add CFR addon install test

piatra:


piatra:
This requires https://github.com/mozilla/activity-stream/pull/4477 to land first.

piatra:
I've included ASRouter's dispatch method to actually trigger the install action (protect against events like #4477). I also did a try run with debug builds https://treeherder.mozilla.org/#/jobs?repo=try&revision=e2af327ac0893940b95033da0c8b0e8f68ea108b&selectedJob=204577273

test/browser/browser_asrouter_cfr.js
```diff
@@ -1,7 +1,9 @@
 const {CFRPageActions} =
   ChromeUtils.import("resource://activity-stream/lib/CFRPageActions.jsm", {});
+const {ASRouter} =
+  ChromeUtils.import("resource://activity-stream/lib/ASRouter.jsm", {});
 
-function trigger_cfr_panel(browser, trigger, cb) {
+function trigger_cfr_panel(browser, trigger, action) {
```
Mardak: We can have the default "FOO" fake action to avoid callers needing to.

test/browser/browser_asrouter_cfr.js
```diff
@@ -67,3 +75,38 @@ add_task(async function test_cfr_notification_show() {
   Assert.equal(PopupNotifications._currentNotifications.length, 0,
     "Should have removed the notification");
 });
+
+add_task(async function test_cfr_addon_install() {
+  // Prevent fetching the real addon url and making a network request
+  CFRPageActions._maybeAddAddonInstallURL = x => x;
```
Mardak: We'll probably want this at the top level of this file and restore original functionality if none of the tests here would trigger install to avoid breaking other tests that might rely more on parts of the actual implementation.

PR #4483, opened 2018-10-09
Backport - Replace nsINavBookmarksService::onItemAdded with 'page-bookmarked'

sarracini:
Blocked on Bug 1426245 landing in central first. https://bugzilla.mozilla.org/show_bug.cgi?id=1426245
This backports the changes from PlacesFeed.jsm, but also fixes the tests, and adds 2 trailing commas to PlacesFeed.jsm, just as heads up for next export.

I'm assuming similar changes are to come with `onItemRemoved` so we might want to refactor PlacesFeed.jsm at some point when BookmarksObserver won't be valid anymore.

### DIFF
--- a/lib/PlacesFeed.jsm
+++ b/lib/PlacesFeed.jsm
@@ -80,46 +80,6 @@ class BookmarksObserver extends Observer {
   }
 
   /**
-   * onItemAdded - Called when a bookmark is added
-   *
-   * @param  {str} id
-   * @param  {str} folderId
-   * @param  {int} index
-   * @param  {int} type       Indicates if the bookmark is an actual bookmark,
-   *                          a folder, or a separator.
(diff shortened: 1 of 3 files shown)


Mardak:
I believe we want port PRs to be against the central branch now. @k88hudson ?

sarracini:
I was assuming we do both here and then also merge it into the central branch?

Mardak:
Instead of two PRs for each branch, after it lands on central branch, we can merge central to master directly without PR. The merge can happen sooner if it's needed for master to work / pass tests; otherwise the central to master merge could happen just before export.

Mardak:
Turns out there were test changes landed in a separate commit that fixed things up differently than here https://hg.mozilla.org/mozilla-central/rev/85a806b69f15

PR #4486, opened 2018-10-10
Fix Bug 1496721 - Flip the pref to turn simplified onboarding on

sarracini:
THE FINAL PIECE OF THE PUZZLE

this needs to wait for bug 1462415 to land in central

well, it doesn't *need* to... it just... should


sarracini:
https://hg.mozilla.org/mozilla-central/rev/c452c4911daf landed!

PR #4489, opened 2018-10-11
Fix Bug 1492853 - Add isUpdated targeting parameter

piatra:
This is the relevant idl file. https://searchfox.org/mozilla-central/source/toolkit/mozapps/update/nsIUpdateService.idl#246
To test I just used the `isMatch` method from ASRouterTargeting. Not sure if this can be automated since it makes a network request.

edit: need to update the docs.


piatra:
Mochitest is trivial but I thought it's a good sanity check and help preserve the correct semantics if we ever change things around.

PR #4490, opened 2018-10-11
Fix Bug 1485633 - Context menu is displayed wrong for the last Highlight from a row

sarracini:
When compact cards were added and we fit 4 cards on a row, this needed to be adjusted for a 4th child
Before:
screen shot 2018-10-11 at 10 54 23 am

After:
screen shot 2018-10-11 at 10 53 07 am

### DIFF
--- a/content-src/components/Sections/_Sections.scss
+++ b/content-src/components/Sections/_Sections.scss
@@ -22,7 +22,9 @@
     }
 
     @media (min-width: $break-point-widest) and (max-width: $break-point-widest + 2 * $card-width) {
-      :nth-child(3n) {
+      // 3n for normal cards, 4n for compact cards
+      :nth-child(3n),
+      :nth-child(4n) {
         @include context-menu-open-left;
       }
     }


content-src/components/Sections/_Sections.scss
```diff
@@ -16,13 +16,17 @@
     }
 
     @media (min-width: $break-point-large) and (max-width: $break-point-large + 2 * $card-width) {
-      :nth-child(3n) {
+      // 3n for normal cards, 4n for compact cards
+      :nth-child(3n),
+      :nth-child(4n) {
```
rlr: I think we  don't need this one? Or maybe it's the one below... Not sure:

screen shot 2018-10-11 at 11 15 50 am

sarracini: removing this one seems to still work, i wasn't sure if it was for something else so i just left it in there for safety. I can take it out though

rlr: From what I can tell, this `@media {}` block is for when the browser is wide enough to show 3 compact columns but not 4. Sort of that in between state. So only the `3n` should be needed.

rlr:
^ testing this out quick...

PR #4496, opened 2018-10-15
Fix Bug 1499068 - Adjust modal button to have variable width

sarracini:
oops my bad
Before:
screen shot 2018-10-15 at 10 58 09 am
After:
screen shot 2018-10-15 at 10 57 50 am


sarracini:
Um I don't know why task cluster would have a problem with some CSS margins but I'm assuming it's not me......

PR #4499, opened 2018-10-15
Fix Bug 1483027 - Add title to three doted menu in home

arshadkazmi42:
Add title to TopSite, Card, CollabsibleSection, 3 dotted menu.
Add `Menu` in `strings.properties` for localization

Fixes [Bugzilla#1483027](https://bugzilla.mozilla.org/show_bug.cgi?id=1483027)

### DIFF
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -206,5 +206,6 @@ firstrun_privacy_notice=Privacy Notice
 firstrun_continue_to_login=Continue
 firstrun_skip_login=Skip this step
 
-# Tooltip for context menu
+# LOCALIZATION NOTE (context_menu_title):
+# context_menu_title is shown as tooltip
 context_menu_title=Menu
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -208,4 +208,4 @@ firstrun_skip_login=Skip this step
 
 # LOCALIZATION NOTE (context_menu_title):
 # context_menu_title is shown as tooltip
-context_menu_title=Menu
+context_menu_title=Open menu
--- a/locales/en-US/strings.properties
+++ b/locales/en-US/strings.properties
@@ -206,6 +206,5 @@ firstrun_privacy_notice=Privacy Notice
 firstrun_continue_to_login=Continue
 firstrun_skip_login=Skip this step
 
-# LOCALIZATION NOTE (context_menu_title):
-# context_menu_title is shown as tooltip
+# LOCALIZATION NOTE (context_menu_title): Action tooltip to open a context menu
 context_menu_title=Open menu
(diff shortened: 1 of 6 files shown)


locales/en-US/strings.properties
```diff
@@ -205,3 +205,6 @@ firstrun_privacy_notice=Privacy Notice
 
 firstrun_continue_to_login=Continue
 firstrun_skip_login=Skip this step
+
+# Tooltip for context menu
```
sarracini: This comment should have the format `#LOCALIZATION NOTE (context_menu_title):` so that the localisers know this comment is for them

test/unit/content-src/components/TopSites.test.jsx
```diff
@@ -548,53 +548,53 @@ describe("<TopSite>", () => {
   });
 
   it("should render a TopSite", () => {
-    const wrapper = shallow(<TopSite link={link} />);
+    const wrapper = shallow(<TopSite link={link} intl={DEFAULT_PROPS.intl} />);
```
sarracini: you should be able to use `shallowWithIntl(<TopSite link={link} />)` here instead of adding the `intl` prop to each wrapper. similarly with all the ones below too

locales/en-US/strings.properties
```diff
@@ -205,3 +205,6 @@ firstrun_privacy_notice=Privacy Notice
...
+context_menu_title=Menu
```
Mardak: I believe typically these are "actions," e.g,. "Open Context Menu" I'll check with UX in the bug.

Also, I'm not sure if we want the same title for both cards and the section menu, but I suppose if so or different, the section menu title could be fixed in a separate PR.

arshadkazmi42:
Thanks @sarracini and @Mardak for the review 👍 
@sarracini I have done the changes as per your feedback.
@Mardak Let me know if there are any changes required for that.

locales/en-US/strings.properties
sarracini: I asked Aaron and he said "Menu" was good for all

sarracini:
@arshadkazmi42 looks like we want "Open menu" instead of just "Menu" for the string

arshadkazmi42:
@sarracini Just checked in Bugzilla. 
Making that change 👍

locales/en-US/strings.properties
```diff
@@ -205,3 +205,7 @@ firstrun_privacy_notice=Privacy Notice
 
 firstrun_continue_to_login=Continue
 firstrun_skip_login=Skip this step
+
+# LOCALIZATION NOTE (context_menu_title):
+# context_menu_title is shown as tooltip
```
Mardak: Repeating the string id is redundant as it's already in the localization note, and the comment can start on the previous line. The comment should also mention that this is an action (as "open" could be confused as an adjective describing that the menu is open). So maybe just:

`# LOCALIZATION NOTE (context_menu_title): Action tooltip to open a context menu`

arshadkazmi42: Thanks @Mardak for this feedback. Will make the change right away 👍

PR #4500, opened 2018-10-15
Port Bug 1497960: Make ProfileAge return the same instances for the same profile directory and preload times.json to avoid data races. r=janerik

Mardak:
r?@piatra Took a while to finally track down why tests weren't passing and turns out `await onAction` wasn't actually waiting because it wasn't returning a Promise. I also simplified getting a custom test value by optionally stashing a `fakeProfileAge.created`.

https://hg.mozilla.org/mozilla-central/rev/3edf43a3f1a4

### DIFF
--- a/lib/ManualMigration.jsm
+++ b/lib/ManualMigration.jsm
@@ -39,7 +39,7 @@ this.ManualMigration = class ManualMigration {
   }
 
   async isMigrationMessageExpired() {
-    let profileAge = new ProfileAge();
+    let profileAge = await ProfileAge();
     let profileCreationDate = await profileAge.created;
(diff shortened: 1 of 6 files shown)


lib/ManualMigration.jsm
```diff
@@ -93,10 +93,10 @@ this.ManualMigration = class ManualMigration {
     this.store.dispatch({type: at.MIGRATION_COMPLETED});
   }
 
-  onAction(action) {
+  async onAction(action) {
     switch (action.type) {
       case at.PREFS_INITIAL_VALUES:
-        this.expireIfNecessary(action.data.migrationExpired);
+        await this.expireIfNecessary(action.data.migrationExpired);
```
piatra: Why was this change required?

PR #4503, opened 2018-10-17
Fix Bug 1499358 - Add button_url field to schema as fallback for OPEN_URL

piatra:


content-src/asrouter/templates/SimpleSnippet/SimpleSnippet.jsx
```diff
@@ -16,9 +16,12 @@ export class SimpleSnippet extends React.PureComponent {
...
       this.props.sendUserActionTelemetry({event: "CLICK_BUTTON", id: this.props.UISurface});
     }
+    const {button_url} = this.props.content;
+    // If button_url is defined handle it as OPEN_URL action
+    const type = this.props.content.button_action || (button_url && "OPEN_URL");
     this.props.onAction({
-      type: this.props.content.button_action,
-      data: {args: this.props.content.button_action_args},
+      type,
+      data: {args: this.props.content.button_action_args || button_url},
```
sarracini: just to make sure i understand correctly, can it ever be the case when the message contains a `button_url` but also `button_action_args` but no `button_action`?
So it would end up like: `{type: "OPEN_URL", data: {args: this.props.content.button_action_args}}`, which i'm assuming wouldn't work to actually open the url since it wouldn't have gotten the `button_url`. I'm assuming we're saying that will never be the case? Am I making sense?

PR #4506, opened 2018-10-17
 Fix Bug 1497565 - Rename locales to locales-src to avoid accidentally exporting l10n-exposed strings

Mardak:
r?@k88hudson Separated out one commit for just renaming and other for updating references, so I'll squash when landing. Refactored out various references to "locales" and "en-US" into package.json configs.


flodolo:
Side note: looks like the CODEOWNERS notifications are working now.

PR #4508, opened 2018-10-17
Add enable/disable/reset devtools for ASRouter

k88hudson:
### IMPORTANT NOTE

A big part of this change is indentation so I would recommending viewing this with https://github.com/mozilla/activity-stream/pull/4508/files?w=1

### Summary
This patch adds some additional utilities to `about:newtab#asrouter`, including
* A way to enable/disable providers via the UI;
* A way to filter the messages shown by provider;
* A way to reset the preference via the UI

### Testing instructions
* Turn on devtools by setting `browser.newtabpage.activity-stream.asrouter.devtoolsEnabled` to `true`
* Visit `about:newtab#asrouter`
* Try enabling/disabling different providers; see if their messages show up/disappear on the page
* Try the messages filter dropdown under the "Messages" section
* After changing some settings, try the "Restore defaults" button and ensure `browser.newtabpage.activity-stream.asrouter.messageProviders` is the default value in `about:config`

### DIFF
--- a/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -5,8 +5,10 @@ export class ASRouterAdmin extends React.PureComponent {
   constructor(props) {
     super(props);
     this.onMessage = this.onMessage.bind(this);
+    this.handleEnabledToggle = this.handleEnabledToggle.bind(this);
+    this.onChangeMessageFilter = this.onChangeMessageFilter.bind(this);
     this.findOtherBundledMessagesOfSameTemplate = this.findOtherBundledMessagesOfSameTemplate.bind(this);
-    this.state = {};
+    this.state = {messageFilter: "all"};
   }
 
   onMessage({data: action}) {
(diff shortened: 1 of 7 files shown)


content-src/components/Base/Base.jsx
```diff
@@ -82,8 +82,11 @@ export class _Base extends React.PureComponent {
     const {initialized} = App;
 
     const prefs = props.Prefs.values;
-    if (prefs["asrouter.devtoolsEnabled"] && window.location.hash === "#asrouter") {
-      return (<ASRouterAdmin />);
+    if (prefs["asrouter.devtoolsEnabled"]) {
+      if (window.location.hash === "#asrouter") {
+        return (<ASRouterAdmin />);
+      }
+      console.log("ASRouter devtools enabled. To access visit %cabout:newtab#asrouter", "font-weight: bold"); // eslint-disable-line no-console
```
piatra: Maybe in an `else` so we don't show it if you actually are on the admin page.

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -81,34 +87,60 @@ export class ASRouterAdmin extends React.PureComponent {
...
         <td>id</td>
+        <td>enabled</td>
         <td>source</td>
         <td>last updated</td>
       </tr>
     </thead>);
   }
 
+  handleEnabledToggle(event) {
+    const action = {type: event.target.checked ? "ENABLE_PROVIDER" : "DISABLE_PROVIDER", data: event.target.name};
```
piatra: I added 
> this.setState({messageFilter: "all"});
To fix that issue

content-src/components/Base/Base.jsx
k88hudson: I agree that would be easier to read but eslint doesn't allow it because of the early return :P

PR #4514, opened 2018-10-19
TC trigger listeners failure

piatra:
Try doesn't complain https://treeherder.mozilla.org/#/jobs?repo=try&revision=5cfa3e4007e1fcab55dc9ee1859dd07829406616&selectedJob=206662390

### DIFF
--- a/mochitest.sh
+++ b/mochitest.sh
@@ -14,9 +14,10 @@ cd /activity-stream && npm install . && npm run buildmc
 # Build latest m-c with Activity Stream changes
 cd /mozilla-central && ./mach build \
   && ./mach lint -l codespell browser/components/newtab \
+  && ./mach test browser/components/newtab/test/browser --headless \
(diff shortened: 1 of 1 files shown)


mochitest.sh
```diff
@@ -14,9 +14,10 @@ cd /activity-stream && npm install . && npm run buildmc
 # Build latest m-c with Activity Stream changes
 cd /mozilla-central && ./mach build \
   && ./mach lint -l codespell browser/components/newtab \
+  && ./mach test browser/components/newtab/test/browser --headless \
+  && ./mach test browser/components/newtab/test/xpcshell \
```
Mardak: This looks like it would regress the `--log-tbpl test_run_log` check for actually failing. Unless `mach test` correctly fails now ?

Mardak: Actually, looks like failing test should cause the command to have non-zero exit code:

https://bugzilla.mozilla.org/show_bug.cgi?id=1473392

PR #4516, opened 2018-10-19
(nobug) Add more support for setting user preferences in ASR devtools

k88hudson:
Adds support for setting user preferences via UI. Test by visiting `about:newtab#asrouter` with `browser.newtabpage.activity-stream.asrouter.devtoolsEnabled` on.

### DIFF
--- a/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -6,6 +6,7 @@ export class ASRouterAdmin extends React.PureComponent {
     super(props);
     this.onMessage = this.onMessage.bind(this);
     this.handleEnabledToggle = this.handleEnabledToggle.bind(this);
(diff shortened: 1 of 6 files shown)


content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -110,36 +111,49 @@ export class ASRouterAdmin extends React.PureComponent {
...
   renderProviders() {
     const providersConfig = this.state.providerPrefs;
     const providerInfo = this.state.providers;
+    const userPrefInfo = this.state.userPrefs;
+
     return (<table>{this.renderTableHead()}<tbody>
       {providersConfig.map((provider, i) => {
         const isTestProvider = provider.id === "snippets_local_testing";
         const info = providerInfo.find(p => p.id === provider.id) || {};
+        const isUserEnabled = provider.id in userPrefInfo ? userPrefInfo[provider.id] : true;
```
piatra: Can we just use `userPrefInfo[provider.id]` the `in + if` syntax is a bit weird to read.

PR #4517, opened 2018-10-19
Folder path tweak for docs

gvn:
From inspecting the commands in the manifest it seems like the compiled AS folder is being generated in a different location now.


gvn:
CC: @ScottDowne @k88hudson

PR #4521, opened 2018-10-22
about:library build changes

sarracini:
@k88hudson / @Mardak  how would y'all feel about landing this so we can have some set up for about:library? these changes make it look exactly like what's currently in central, just now it's moved over to our process. Thoughts?


k88hudson:
Seems good to me!

yamscripts.yml
```diff
@@ -27,6 +27,12 @@ scripts:
     stringsExport: cp $npm_package_config_locales_dir/$npm_package_config_default_locale/strings.properties $npm_package_config_mc_dir/browser/locales/$npm_package_config_default_locale/chrome/browser/activity-stream/newtab.properties
     copyPingCentre: cpx "ping-centre/PingCentre.jsm" $npm_package_config_mc_dir/browser/modules
 
+# buildlibrary: Export about:library code to mozilla-central
```
k88hudson: You could maybe not that this is intentionally not being called by the general build task right now

PR #4523, opened 2018-10-24
 Fixes bug 1507973 - v2 with attachments for personality provider in activity stream.

ScottDowne:
To test:
1. Make a new profile, but don't log into it.
2. Go into that profile's dir, and paste in a places.sqlite from an existing profile to get a good chunk of history to work with.
3. Create a user.js file with this in it:
```
user_pref("services.settings.server", "https://kinto.dev.mozaws.net/v1");
user_pref("services.settings.verify_signature", false);
```
4. Now run the profile.
5. In the browser debugger, paste in `Services.obs.notifyObservers(null, "idle-daily");`
6. Make an insignificant change to the pref browser.newtabpage.activity-stream.feeds.section.topstories.options to trigger a fresh cache.

Expected: you should now be seeing personalized stories on about:home.

### DIFF
--- a/lib/PersonalityProvider.jsm
+++ b/lib/PersonalityProvider.jsm
@@ -96,7 +96,9 @@ this.PersonalityProvider = class PersonalityProvider {
     const {attachment: {filename}} = record;
     await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
     const path = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
-    return OS.File.remove(path, {ignoreAbsent: true});
+
+    await OS.File.remove(path, {ignoreAbsent: true});
+    return OS.File.removeEmptyDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME), { ignoreAbsent: true });
   }
 
   async getAttachment(record) {
--- a/lib/PersonalityProvider.jsm
+++ b/lib/PersonalityProvider.jsm
@@ -37,6 +37,23 @@ const PERSONALITY_PROVIDER_DIR_NAME = "personality-provider";
 const RECIPE_NAME = "personality-provider-recipe-attachment";
 const MODELS_NAME = "personality-provider-models-attachment";
 
+function getHash(aStr) {
+  // return the two-digit hexadecimal code for a byte
+  let toHexString = charCode => ("0" + charCode.toString(16)).slice(-2);
+
+  let hasher = Cc["@mozilla.org/security/hash;1"].
+               createInstance(Ci.nsICryptoHash);
+  hasher.init(Ci.nsICryptoHash.MD5);
+  let stringStream = Cc["@mozilla.org/io/string-input-stream;1"].
+                     createInstance(Ci.nsIStringInputStream);
+  stringStream.data = aStr;
+  hasher.updateFromStream(stringStream, -1);
+
+  // convert the binary hash data to a hex string.
+  let binary = hasher.finish(false);
+  return Array.from(binary, (c, i) => toHexString(binary.charCodeAt(i))).join("").toLowerCase();
+}
+
 /**
  * V2 provider builds and ranks an interest profile (also called an “interest vector”) off the browse history.
  * This allows Firefox to classify pages into topics, by examining the text found on the page.
@@ -84,9 +101,15 @@ this.PersonalityProvider = class PersonalityProvider {
     const {attachment: {location, filename}} = record;
     const headers = new Headers();
     headers.set("Accept-Encoding", "gzip");
-    const resp = await fetch((await baseAttachmentsURL) + location, {headers});
+    const filepath = (await baseAttachmentsURL) + location;
+    const resp = await fetch(filepath, {headers});
+    if (!resp.ok) {
+      Cu.reportError(`Failed to fetch ${filepath}: ${resp.status}`);
+      return;
+    }
     const buffer = await resp.arrayBuffer();
     const bytes = new Uint8Array(buffer);
+    console.log(bytes);
     await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
     const path = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
     return OS.File.writeAtomic(path, bytes, {tmpPath: `${path}.tmp`});
@@ -102,9 +125,42 @@ this.PersonalityProvider = class PersonalityProvider {
   }
 
   async getAttachment(record) {
-    const {attachment: {filename}} = record;
-    await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
+/*
+      // Figure out hash + md5sum check.
+      // const {attachment: {filename, size}} = record;
+      // Figure out hash + md5sum check.
+      const {attachment: {filename, hash, size}} = record;
+
+      if (!await OS.File.exists(filepath) || await OS.File.stat(filepath).size !== size) {
+        await this.downloadAttachment(record);
+      } else {
+        attachment = await this.getAttachment(record);
+        if () {
+          await this.downloadAttachment(record);
+          attachment = await this.getAttachment(record);
+        }
+      }
+      attachment = attachment || await this.getAttachment(record);
+      console.log(attachment);
+      console.log(getHash(JSON.stringify(attachment)), hash);
+      console.log(getHash(JSON.stringify(attachment)) === hash);
+      //if (!await ) {
+      //  await this.downloadAttachment(record);
+      //}
+*/
+    const {attachment: {filename, hash, size}} = record;
     const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+
+    if (!await OS.File.exists(filepath) || await OS.File.stat(filepath).size !== size) {
+      await this.downloadAttachment(record);
+    }
+
+
+
+
+
+    //const {attachment: {filename}} = record;
+    await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
     try {
       const fileExists = await OS.File.exists(filepath);
       if (fileExists) {
@@ -150,15 +206,9 @@ this.PersonalityProvider = class PersonalityProvider {
   async getFromRemoteSettings(name) {
     const result = await RemoteSettings(name).get();
     return Promise.all(result.map(async record => {
-      const {attachment: {filename, size}} = record;
-      // Figure out hash + md5sum check.
-      // const {attachment: {filename, hash, size}} = record;
+      const {attachment: {filename}} = record;
       await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
       const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
-      // Figure out hash + md5sum check.
-      if (!await OS.File.exists(filepath) || await OS.File.stat(filepath).size !== size) {
-        await this.downloadAttachment(record);
-      }
       return {...await this.getAttachment(record), recordKey: record.key};
     }));
   }
--- a/lib/PersonalityProvider.jsm
+++ b/lib/PersonalityProvider.jsm
@@ -22,10 +22,6 @@ XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
 
 XPCOMUtils.defineLazyGetter(this, "gTextDecoder", () => new TextDecoder());
 
-// This doesn't work currently, the fetch wasn't returning the expected
-// data structure from https://kinto.dev.mozaws.net/v1/, failing on
-// capabilities not being defined.
-
 XPCOMUtils.defineLazyGetter(this, "baseAttachmentsURL", async () => {
   const server = Services.prefs.getCharPref("services.settings.server");
   const serverInfo = await (await fetch(`${server}/`)).json();
@@ -33,9 +29,9 @@ XPCOMUtils.defineLazyGetter(this, "baseAttachmentsURL", async () => {
   return base_url;
 });
 
-const PERSONALITY_PROVIDER_DIR_NAME = "personality-provider";
-const RECIPE_NAME = "personality-provider-recipe-attachment";
-const MODELS_NAME = "personality-provider-models-attachment";
+const PERSONALITY_PROVIDER_DIR = OS.Path.join(OS.Constants.Path.localProfileDir, "personality-provider");
+const RECIPE_NAME = "personality-provider-recipe";
+const MODELS_NAME = "personality-provider-models";
 
 function getHash(aStr) {
   // return the two-digit hexadecimal code for a byte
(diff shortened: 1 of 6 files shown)


ScottDowne:
@leplatrem So this seems to be working for me, thoughts?

Also @jonathankoren @ncloudioj @Mardak in case you're curious.

ScottDowne:
Oh right, I had some issues with the `baseAttachmentsURL` but I left comments in the code explaining it, and hard coded a value to become un blocked.

ScottDowne:
I'll do more serious testing on this tomorrow.

lib/PersonalityProvider.jsm
```diff
@@ -38,6 +56,46 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      const toDownload = created.concat(updated.map(u => u.new));
+      await Promise.all(
+        toDownload.map(async record => {
+          const {attachment: {location, filename}} = record;
+
+          // This should be using this line:
+          // const resp = await fetch(`${baseAttachmentsURL}${location}`);
+          // but see above issue in baseAttachmentsURL
+          // for now I'm hard conding in dev url to test
+          const resp = await fetch(`https://kinto.dev.mozaws.net/attachments/${location}`);
```
leplatrem: Maybe it's already the case by default, but we should have `Accept-Encoding: gzip` here to make sure we leverage Cloudfront gzipping

lib/PersonalityProvider.jsm
```diff
@@ -15,6 +15,24 @@ const {RecipeExecutor} = ChromeUtils.import("resource://activity-stream/lib/Reci
...
+// ChromeUtils.import("resource://gre/modules/Services.jsm");
+ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
+ChromeUtils.defineModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm");
+XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
+
+XPCOMUtils.defineLazyGetter(this, "gTextDecoder", () => new TextDecoder());
+
+// This doesn't work currently, the fetch wasn't returning the expected
+// data structure from https://kinto.dev.mozaws.net/v1/, failing on
+// capabilities not being defined.
```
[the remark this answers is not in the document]
ScottDowne: Yeah it worked for me everywhere cept the fetch.

lib/PersonalityProvider.jsm
```diff
@@ -38,6 +56,46 @@ this.PersonalityProvider = class PersonalityProvider {
     this.scores = scores || {};
     this.interestConfig = this.scores.interestConfig;
     this.interestVector = this.scores.interestVector;
+    this.syncAttachment("personality-provider-recipe-attachment");
+    this.syncAttachment("personality-provider-models-attachment");
```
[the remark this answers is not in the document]
ScottDowne: Works for me.

lib/PersonalityProvider.jsm
```diff
@@ -38,6 +56,46 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      // Download every new/updated attachment.
+      const toDownload = created.concat(updated.map(u => u.new));
+      await Promise.all(
+        toDownload.map(async record => {
+          const {attachment: {location, filename}} = record;
+
+          // This should be using this line:
+          // const resp = await fetch(`${baseAttachmentsURL}${location}`);
+          // but see above issue in baseAttachmentsURL
+          // for now I'm hard conding in dev url to test
```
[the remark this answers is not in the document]
ScottDowne: :+1:

lib/PersonalityProvider.jsm
ScottDowne: OK, I think I figured this out.

lib/PersonalityProvider.jsm
```diff
@@ -38,6 +56,46 @@ this.PersonalityProvider = class PersonalityProvider {
...
+          const resp = await fetch(`https://kinto.dev.mozaws.net/attachments/${location}`);
+          const buffer = await resp.arrayBuffer();
+          const bytes = new Uint8Array(buffer);
+
+          const path = OS.Path.join(OS.Constants.Path.localProfileDir, filename);
```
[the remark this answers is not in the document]
ScottDowne: I think given we only have 1 non model data set, and in the future it's not going to be a lot, I wouldn't mind just making 1 dir for models and the recipe called "personality-provider/" or something?

lib/PersonalityProvider.jsm
```diff
@@ -81,13 +155,16 @@ this.PersonalityProvider = class PersonalityProvider {
   async getRecipe() {
     if (!this.recipe || !this.recipe.length) {
       const start = perfService.absNow();
-      this.recipe = await this.getFromRemoteSettings("personality-provider-recipe");
+      this.recipe = await this.getFromRemoteSettings("personality-provider-recipe-attachment");
```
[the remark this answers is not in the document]
ScottDowne: Just for the tests, but not currently sure the best way to handle this.

lib/PersonalityProvider.jsm
```diff
@@ -100,13 +177,17 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      const models = await this.getFromRemoteSettings("personality-provider-models-attachment");
 
       if (models.length === 0) {
         return null;
       }
 
       for (let model of models) {
+        if (!model) {
+          continue;
+        }
```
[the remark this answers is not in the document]
ScottDowne: I agree with the first point, I just did that to be safe and wasn't sure what could or couldn't be returned. I certainly didn't expect a falsy item to be returned.

I considered returning the attachment directly, but the current implementation currently needs the key param I add to the meta data of the record. So the attachment alone isn't enough, but I could still abstract it and return both the attachment and the metadata.

lib/PersonalityProvider.jsm
```diff
@@ -69,6 +127,22 @@ this.PersonalityProvider = class PersonalityProvider {
...
+        return JSON.parse(json);
+      }
+    } catch (error) {
+      Cu.reportError(`Failed to load ${filepath}: ${error.message}`);
+    }
+    return null;
+  }
+
   async getFromRemoteSettings(name) {
     const result = await RemoteSettings(name).get();
```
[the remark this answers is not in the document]
ScottDowne: Is it possible this downloadAttachment can become a race condition with the download in the sync, or do they just write over each other and it's harmless?

ScottDowne:
@leplatrem updated, still a few issues, but pretty close now it seems.

The bulk of the attachment stuff is now in like, a few functions in one spot, and pretty generic now, can probably move it to a jsm somewhere.

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
leplatrem: I would say that `OS.File.writeAtomic()` does the job, but would have to be confirmed

lib/PersonalityProvider.jsm
```diff
@@ -38,11 +66,67 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      const toDownload = created.concat(updated.map(u => u.new));
+      await Promise.all(toDownload.map(record => this.downloadAttachment(record)));
+    });
+  }
+
+  async downloadAttachment(record) {
+    const {attachment: {location, filename}} = record;
+    const headers = new Headers();
+    headers.set("Accept-Encoding", "gzip");
+    // Still not quite working. Something up with promises.
```
leplatrem: Yeah, maybe `defineLazyGetter`does not work with promises... 

You may want to replace with something like:

```js
let _gBaseAttachmentURL;
async function baseAttachmentURL() {
  if (!_gBaseAttachmentURL) {
    const server = Services.prefs.getCharPref("services.settings.server");
    const serverInfo = await fetch(`${server}/`);
    const {capabilities: {attachments: {base_url}}} = serverInfo;
    _gBaseAttachmentURL = base_url;
  }
  return return _gBaseAttachmentURL;
}
```

and then: 
```js
const resp = await fetch((await baseAttachmentURL()) + location);
```

lib/PersonalityProvider.jsm
```diff
@@ -71,7 +155,21 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      // const {attachment: {filename, hash, size}} = record;
+      // const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+      const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, filename);
+      // Figure out hash + md5sum check.
+      if (!await OS.File.exists(filepath) || await OS.File.stat(filepath).size !== size) {
+        await this.downloadAttachment(record);
+      }
+      return {
+        attachment: await this.getAttachment(record),
+        metaData: record,
```
leplatrem: Here apparently you only use the `key` attribute. Maybe `modelKey` would be enough?

Or why not `return {...attachment, modelKey: record.key}` ?

lib/PersonalityProvider.jsm
```diff
@@ -38,11 +66,67 @@ this.PersonalityProvider = class PersonalityProvider {
...
+    // const path = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+    const path = OS.Path.join(OS.Constants.Path.localProfileDir, filename);
+    return OS.File.writeAtomic(path, bytes, {tmpPath: `${path}.tmp`});
+  }
+
+  deleteAttachment(record) {
+    const {attachment: {filename}} = record;
+    // const path = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+    const path = OS.Path.join(OS.Constants.Path.localProfileDir, filename);
+    return OS.File.remove(path, {ignoreAbsent: true});
```
leplatrem: `OS.File.removeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME), { ignoreAbsent: true });`

lib/PersonalityProvider.jsm
```diff
@@ -38,11 +66,67 @@ this.PersonalityProvider = class PersonalityProvider {
...
+  }
+
+  async getAttachment(record) {
+    const {attachment: {filename}} = record;
+    // const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
```
leplatrem: `await OS.File.makeDir(PERSONALITY_PROVIDER_DIR_NAME, {from: OS.Constants.Path.localProfileDir});`

lib/PersonalityProvider.jsm
```diff
@@ -71,7 +155,21 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      // Figure out hash + md5sum check.
+      // const {attachment: {filename, hash, size}} = record;
+      // const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+      const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, filename);
+      // Figure out hash + md5sum check.
+      if (!await OS.File.exists(filepath) || await OS.File.stat(filepath).size !== size) {
+        await this.downloadAttachment(record);
+      }
+      return {
+        attachment: await this.getAttachment(record),
```
leplatrem: There might be better a name than `attachment` here... features? classifiers? vectors? training data?

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
ScottDowne: From my testing it looks like yeah, `writeAtomic` does the job.

lib/PersonalityProvider.jsm
ScottDowne: yup that works.

lib/PersonalityProvider.jsm
ScottDowne: Looks like we can still use `defineLazyGetter` async we just need to do this when we call it `const resp = await fetch((await baseAttachmentsURL) + location, {headers});`

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
ScottDowne: If it helps, this is what it is returning:
```
bodyUsed: false
headers: Headers { guard: "response" }
ok: true
redirected: false
status: 200
statusText: "OK"
type: "basic"
url: "https://kinto.dev.mozaws.net/v1/"
```

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
ScottDowne: The only thing I cannot seem to figure out  is the fileMd5sum stuff. I tried looking around and found gCryptoHash and tried playing around with it, but couldn't make sense of it.

Do you have any advice or examples in FF where this sort of comparison happens?

lib/PersonalityProvider.jsm
ScottDowne: I'm not sure if we need to remove the dir in general. Only if we're removing all the attachments do we need to remove the dir.

Are you thinking if we remove the dir but it has a file in it, does it just fail/catch or something?

Given that, the last attempt on the last file would work, fully cleaning up the dir?

ScottDowne:
@leplatrem Updates.

Just two more things unresolved. Something about md5sum and the removeDir stuff which I left comments about.

lib/PersonalityProvider.jsm
leplatrem: Oh, yeah that would be neat indeed :) 

TBH I just wanted to give you the snippet to delete the dir, I hadn't thought too much about the corner cases ;)

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
leplatrem: > Do you have any advice or examples in FF where this sort of comparison happens?

Unfortunately I don't...
Maybe with this V1 we can just conform to file size comparison.

@mozmark do you guyz have some helpers in the crypto libs to verify a md5 of a file efficiently?

lib/PersonalityProvider.jsm
```diff
@@ -81,7 +168,7 @@ this.PersonalityProvider = class PersonalityProvider {
   async getRecipe() {
     if (!this.recipe || !this.recipe.length) {
       const start = perfService.absNow();
-      this.recipe = await this.getFromRemoteSettings("personality-provider-recipe");
+      this.recipe = await this.getFromRemoteSettings(RECIPE_NAME);
```
leplatrem: Should be plural no? `this.recipes`

lib/PersonalityProvider.jsm
```diff
@@ -71,7 +147,18 @@ this.PersonalityProvider = class PersonalityProvider {
...
+      const {attachment: {filename, size}} = record;
+      // Figure out hash + md5sum check.
+      // const {attachment: {filename, hash, size}} = record;
+      await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
+      const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+      // Figure out hash + md5sum check.
+      if (!await OS.File.exists(filepath) || await OS.File.stat(filepath).size !== size) {
+        await this.downloadAttachment(record);
+      }
+      return {...await this.getAttachment(record), recordKey: record.key};
```
leplatrem: Maybe here we'd want another safety check (redundant I know), because `this.getAttachment()` CAN return `null`

```suggestion
      const attachment = await this.getAttachment(record);
      if (!attachment) {
          throw new Error("Model data is missing and could not be downloaded");
      }
      return {...attachment, recordKey: record.key};
```

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
mozmark: Take a look at: https://searchfox.org/mozilla-central/source/browser/components/migration/360seProfileMigrator.js#76 

Question, though; what's the md5 hash used for?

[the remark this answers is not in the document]
leplatrem: > Question, though; what's the md5 hash used for?

Just a way to make sure that the attachment is not altered/corrupted (the server provides a `hash` value in the record)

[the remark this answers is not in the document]
mozmark: OK; if we're protecting against malicious modification, something stronger than MD5 might be a good idea. Ci.nsICryptoHash.SHA256 perhaps?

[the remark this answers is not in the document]
ScottDowne: To be clear, we're doing this in a followup?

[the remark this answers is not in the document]
leplatrem: > OK; if we're protecting against malicious modification, something stronger than MD5 might be a good idea. Ci.nsICryptoHash.SHA256 perhaps?

I don't mind adding another field with the SHA256 hash, but currently the md5 was here just as a download checksum.

> To be clear, we're doing crypto stuff in a followup?

As you want, based on what Mark showed us, it does not seem super complicated to check the md5. 
But if it's too heavy and requires some offloading into a worker or something, then maybe we should postpone it indeed...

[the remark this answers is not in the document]
ScottDowne: OK, I'll give it a go in this patch.

lib/PersonalityProvider.jsm
ScottDowne: Good catch, instead of a safety check, how about just making it so `getAttachment` returns `{}` instead of `null` in the previous `null` cases?

ScottDowne: This code seems to have been changed since you commented, so I think while what you said still applies about the null return being an issue, it's an issue in slightly different code now.

ScottDowne: The suggested changes feature threw me off for a sec, I thought I was looking at my diff.

lib/PersonalityProvider.jsm
ScottDowne: Looks like this won't work as hoped, because from here: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/OSFile.jsm/OS.File_for_the_main_thread#OS.File.removeDir()

"Remove an existing directory and its contents."

ScottDowne: Here we go though: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/OSFile.jsm/OS.File_for_the_main_thread#OS.File.removeEmptyDir()

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
ScottDowne: Having issues getting the hash's to match.

When I run this function on the file contents:
```
function getHash(aStr) {
  // return the two-digit hexadecimal code for a byte
  let toHexString = charCode => ("0" + charCode.toString(16)).slice(-2);

  let hasher = Cc["@mozilla.org/security/hash;1"].
               createInstance(Ci.nsICryptoHash);
  hasher.init(Ci.nsICryptoHash.MD5);
  let stringStream = Cc["@mozilla.org/io/string-input-stream;1"].
                     createInstance(Ci.nsIStringInputStream);
  stringStream.data = aStr;
  hasher.updateFromStream(stringStream, -1);

  // convert the binary hash data to a hex string.
  let binary = hasher.finish(false);
  return Array.from(binary, (c, i) => toHexString(binary.charCodeAt(i))).join("").toLowerCase();
}
```
I'm not getting the same thing as the hash the server is returning. Not sure yet if it's something in the function or something in my file, or what, but for the moment I'm guessing at what might be wrong.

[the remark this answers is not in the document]
ScottDowne: An  example of hash differences:
Server: `04e7382507921a1a6d0d35da7c2ebd8592bd3937c5dac9b619c96ea890be4bb8`
getHash function: `04b50b53251e08c8f590c55a5b3db4a9`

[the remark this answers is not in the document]
ScottDowne: I updated the `getHash` function above to instead of `MD5` use `SHA256` and it matches the server returned value now.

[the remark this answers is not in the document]
leplatrem: That's my fault, I'm truely sorry :(
https://github.com/Kinto/kinto-attachment/blob/c905b57deeb6918c29e4952f7579e2d85eaf2f1d/kinto_attachment/utils.py#L138

lib/PersonalityProvider.jsm
```diff
@@ -15,6 +15,42 @@ const {RecipeExecutor} = ChromeUtils.import("resource://activity-stream/lib/Reci
...
+ChromeUtils.import("resource://gre/modules/Services.jsm");
+ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
+ChromeUtils.defineModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm");
+XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
+
+XPCOMUtils.defineLazyGetter(this, "gTextDecoder", () => new TextDecoder());
+
+// This doesn't work currently, the fetch wasn't returning the expected
+// data structure from https://kinto.dev.mozaws.net/v1/, failing on
+// capabilities not being defined.
```
leplatrem: This isn't true anymore, right?

lib/PersonalityProvider.jsm
```diff
@@ -15,6 +15,42 @@ const {RecipeExecutor} = ChromeUtils.import("resource://activity-stream/lib/Reci
...
+
+XPCOMUtils.defineLazyGetter(this, "baseAttachmentsURL", async () => {
+  const server = Services.prefs.getCharPref("services.settings.server");
+  const serverInfo = await (await fetch(`${server}/`)).json();
+  const {capabilities: {attachments: {base_url}}} = serverInfo;
+  return base_url;
+});
+
+const PERSONALITY_PROVIDER_DIR_NAME = "personality-provider";
```
leplatrem: you could do this I think? 
`const PERSONALITY_PROVIDER_DIR = OS.Path.join(OS.Constants.Path.localProfileDir, "personality-provider")`

lib/PersonalityProvider.jsm
```diff
@@ -38,6 +74,76 @@ this.PersonalityProvider = class PersonalityProvider {
...
+    const {attachment: {filename, hash, size}} = record;
+    await OS.File.makeDir(OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME));
+    const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, PERSONALITY_PROVIDER_DIR_NAME, filename);
+    let jsonStr = "{}";
+
+    try {
+      if (!await OS.File.exists(filepath) ||
+          await OS.File.stat(filepath).size !== size ||
+          getHash(await this._getFileStr(filepath)) !== hash) {
+        await this.downloadAttachment(record);
```
leplatrem: Shouldn't this be in a `while (retry < 3) {}` ?

lib/PersonalityProvider.jsm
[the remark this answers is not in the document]
ScottDowne: np, was super rewarding to figure it out myself :) :muscle:

lib/PersonalityProvider.jsm
ScottDowne: Good catch.

lib/PersonalityProvider.jsm
ScottDowne: Good catch.

lib/PersonalityProvider.jsm
ScottDowne: Agreed, do I need a similar or same retry on the sync download too?

ScottDowne: I'm considering moving the retry and hash check logic into downloadAttachment.

ScottDowne: I moved some stuff around to get the retry stuff to fit in better.

ScottDowne:
@leplatrem Some updates. I *think* I just need to start adding some tests if everything else looks good to you.

ScottDowne:
hey @Mardak or @ncloudioj not sure which of you is ideal to take a look at this, but the attachments implementation is working on my end, probably time to get someone else running it locally and do a more activity stream focused review. I'm working on tests now, but I don't think that would stop either of you from starting to give it a spin. :)

ScottDowne:
@ncloudioj Finally finished the tests for this. I think it's ready for a complete review now.

ScottDowne:
OK, finally fixed up. @ncloudioj ready again I think for reals this time.

PR #4525, opened 2018-10-25
Fix Bug 1501656 - Prevent onLocationChange from being called twice be…

piatra:
…cause of redirect events


k88hudson:
👍

piatra:
I'll follow up with a mochitest. There's some navigation tests that seem to do this.

PR #4526, opened 2018-10-25
Fix Bug 1501817 - Create FxA onboarding card

sarracini:
<strike>Bug 1502140 with the strings needs to land before this</strike>
To test this, paste the following code in the browser console in a new profile:
```js
ChromeUtils.import("resource:///modules/AttributionCode.jsm");

let appPath = Services.dirsvc.get("GreD", Ci.nsIFile).parent.parent.path;
let attributionSvc = Cc["@mozilla.org/mac-attribution;1"]
                           .getService(Ci.nsIMacAttributionService);

let referrer = `https://allizom.org/anything/?utm_campaign=non-fx-button&utm_source=addons.mozilla.org`;
attributionSvc.setReferrerUrl(appPath, referrer, true);
  
let env = Cc["@mozilla.org/process/environment;1"].getService(Ci.nsIEnvironment);
env.set("XPCSHELL_TEST_PROFILE_DIR", "testing");
AttributionCode._clearCache();
AttributionCode.getAttrDataAsync();
```
Then visit `about:welcome`. Observe that the first card should be Private Browsing, followed by Screenshots, followed by the new Sync card. View that it matches the spec: https://mozilla.invisionapp.com/share/XFOQD3H23YE#/screens/327208697_Card_Variants

<strike> Also note that this is missing the correct sync svg. Aaron is working on getting me that now. I'm also working on a mochitest for this </strike>


sarracini:
I think we can review this and then i can wait for the svg before merging. Make sure you're on latest central so you have the strings!!

lib/OnboardingMessageProvider.jsm
```diff
@@ -69,12 +71,30 @@ const ONBOARDING_MESSAGES = () => ([
...
+    bundled: 3,
+    order: 4,
+    content: {
+      title: {string_id: "onboarding-fxa-title"},
+      text: {string_id: "onboarding-fxa-text"},
+      icon: "sync",
+      button_label: {string_id: "onboarding-button-label-get-started"},
+      button_action: {
+        type: "OPEN_URL",
+        data: {args: "https://accounts.firefox.com/"},
```
sarracini: Note: this URL is actually not the correct url that it should point to, but I need to talk to @vladikoff about what the proper url will be... we can do it as a follow up since this won't be live for a bit

sarracini: Yup I'll add a comment, and I'll file a bug 👍

sarracini: Bug filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1503912

sarracini: I'm going to close bug 1503912 because i just added the correct url as part of this pr

lib/ASRouterTargeting.jsm
```diff
@@ -163,10 +163,13 @@ const TargetingGetters = {
   get browserSettings() {
     const {settings} = TelemetryEnvironment.currentEnvironment;
     return {
-      attribution: settings.attribution,
```
k88hudson: Please leave this in, in order to support backwards compatibility

PR #4529, opened 2018-10-26
Typescript with jsdocs working

k88hudson:
This patch adds typescript to a single file as well as some configuration.

To test this, I'd recommend checking out the code and making a change that would cause a failure, then running `npm run lint`.


k88hudson:
Unfortunately it doesn't seem like typescript is able to recognize `.jsm` files;
Zombie worked on a possible solution for this here: https://github.com/zombie/pilot

content-src/asrouter/components/Button/Button.jsx
```diff
@@ -1,16 +1,31 @@
+// @ts-check
+
+/**
+ * @typedef {import("./Button").ButtonProps} ButtonProps
```
k88hudson: There might be a better way to do this...

PR #4531, opened 2018-10-26
Fix Bug 1496241 - Activity Stream doesn't honor default preference set via Autoconfig (and policy)

Mardak:
r?@k88hudson This removes some test-only methods on `Prefs` and `DefaultPrefs` while making the latter extend `Preferences` like the former. This was to reuse the `get` method that knows how to get a preference by type because turns out just plain `getPrefType() == INVALID` is insufficient to correctly determine if there's a default preference value.


k88hudson:
Looks good to me!

PR #4534, opened 2018-10-29
Fix Bug 1502333 - Add copy paste functionality to targeting parameters

piatra:
The [clipboard](https://developer.mozilla.org/en-US/docs/Web/Events/copy) event API is only available when an actual copy action is triggered by the user.

Unfortunately we can only trigger this for copy events, ["paste" execCommand](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) is not supported.

This means that copy will happen on button click but to paste in the parameters you still need to trigger the action manually. The `paste` event listener is also hidden behind a button so that random paste events don't get captured and treated as trigger parameter values.


piatra:
PR adds these 2 new buttons.
grafik

Clicking the copy button will copy all the data in your clipboard automatically
grafik

Clicking the second button will enable you to paste data as targeting parameters (the CMD/CTRL+V action is still needed, limitation described above)
grafik

PR #4535, opened 2018-10-29
Fix Bug 1499901 - Add builduplift task that does not export strings

k88hudson:
This task exports activity-stream to MC without exporting strings to browser/locales.

To test run `npm run builduplift`


yamscripts.yml
```diff
@@ -27,6 +27,12 @@ scripts:
     stringsExport: cp $npm_package_config_locales_dir/$npm_package_config_default_locale/strings.properties $npm_package_config_mc_dir/browser/locales/$npm_package_config_default_locale/chrome/browser/activity-stream/newtab.properties
     copyPingCentre: cpx "ping-centre/PingCentre.jsm" $npm_package_config_mc_dir/browser/modules
 
+# builduplift: Build and export to mozilla central for uplifts without exporting strings to browser/locales
+  builduplift:
+    pre: rimraf $npm_package_config_mc_dir/browser/components/newtab/
```
piatra: Isn't this just `buildmc:pre` ?

Mardak: package.json doesn't actually know of a `buildmc:pre` but a `prebuildmc` instead. So yeah, this line could be `pre: => prebuildmc`

PR #4539, opened 2018-10-31
Bug 1497507 - aligning cta w. value proposition

gvn:
I can't assign a reviewer, so @k88hudson or @ScottDowne R?

### DIFF
--- a/content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
+++ b/content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
@@ -27,7 +27,7 @@
   .cta-text {
     font-weight: normal;
     font-size: 13px;
-    line-height: 1.230769231; // (16 / 13) – 16px computed
+    line-height: 1.230769231; // (16 / 13) –> 16px computed
   }
 
   .pocket-cta-button,
--- a/content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
+++ b/content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
@@ -27,6 +27,7 @@
   .cta-text {
     font-weight: normal;
     font-size: 13px;
+    line-height: 1.230769231; // (16 / 13) – 16px computed
   }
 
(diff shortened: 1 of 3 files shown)


k88hudson:
Taskcluster failures are unrelated, disregard

content-src/components/PocketLoggedInCta/_PocketLoggedInCta.scss
```diff
@@ -27,6 +27,7 @@
   .cta-text {
     font-weight: normal;
     font-size: 13px;
+    line-height: 1.230769231; // (16 / 13) – 16px computed
```
ScottDowne: "more recommendations" is no longer aligned: 
align
I think it just needs a matching line-height at the least.

content-src/components/Sections/_Sections.scss
```diff
@@ -84,11 +84,6 @@
       font-size: 13px;
     }
   }
-
-  @media (min-width: $break-point-large) {
-    line-height: 16px;
-    height: 16px;
```
ScottDowne: This does change `top-stories-bottom-container`'s height to 20.8px from 16px; is that intended?

gvn: Well, we really shouldn't be setting fixed heights on anything containing live text unless there's a good reason to, so...yes?

gvn: Why was height set in the first place?

ScottDowne: I *think* it was something to do with the height transition for collapsing the panel, it was causing some jitter and movement for the highlights container. Something I currently cannot reproduce.

ScottDowne: Other than that, I cannot really see what it's doing.

gvn: Ok, I think it's probably safe to remove then.

gvn:
@ScottDowne - Made that line height fixup...I think it's good now?

PR #4541, opened 2018-10-31
Fix Bug 1503685 - Add "extension controlled" info to PAGE_TAKEOVER_DATA telemetry ping

rlr:


rlr:
@ncloudioj  after chatting with francois and mconnor, we decided to just send the web extension IDs of controlling extensions.

You can test this out with https://addons.mozilla.org/en-US/firefox/addon/binghomepage/ for the home page and https://addons.mozilla.org/en-US/firefox/addon/search-and-new-tab-by-yahoo/ for the newtab page.

rlr:
data-review+: https://bugzilla.mozilla.org/show_bug.cgi?id=1503685#c6

PR #4545, opened 2018-11-02
Fix Bug 1501304 - Add telemetry for CFR opt-out controls

rlr:


rlr:
data-review+: https://bugzilla.mozilla.org/show_bug.cgi?id=1501304#c3

PR #4548, opened 2018-11-06
Fix Bug 1502332 - Add message filtering based on targeting expression and parameters

piatra:
This will go through all the messages and filter out those that have a targeting expression that does not match the current values of the targeting parameters. Updating those parameters allows to check which messages match.
grafik

Some improvements to the inputs
grafik

It adds some improvements to the error states
grafik


piatra:
Closing this PR for now, follow along with the status in bug 1502332.

PR #4549, opened 2018-11-06
Bug 1476458 - Prevent expired sponsored stories from showing to users

gvn:
To repro:

- Edit config `browser.newtabpage.activity-stream.feeds.section.topstories.options` and change `stories_endpoint` to a local copy of the service JSON that you can edit.
- See what spoc shows up in the new tab
- Change spoc's `expiration_timestamp` to `1` (or anything below `Date.now() / 1000`) in your local JSON
- Tweak `browser.newtabpage.activity-stream.feeds.section.topstories.options` to force a reload of JSON
- Observe that the spoc no longer appears


gvn:
@ScottDowne Tests are passing again! Ready for review!

gvn:
Yeah that sounds like a good idea.

lib/TopStoriesFeed.jsm
```diff
@@ -401,29 +410,43 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
-      }
-      if (Math.random() > this.spocsPerNewTabs) {
-        this.dispatchSpocDone(target);
-        return false;
-      }
-      if (!this.spocs || !this.spocs.length) {
-        // We have stories but no spocs so there's nothing to do and this update can be
-        // removed from the queue.
-        this.dispatchSpocDone(target);
+  filterSpocs() {
```
gvn: I broke this out of `maybeAddSpoc` to a separate method to make it single purpose and therefore easier to isolate for unit testing. Now return values can be directly analyzed instead of having to dig deeper into other parts of the app's data flow. Also `maybeAddSpoc` was just feeling overloaded...

gvn:
I took a slightly different approach, which I think is cleaner and more self-contained.

R?

gvn:
@k88hudson It would be nice to hear your thoughts as well since this is my first test for Activity Stream! 🎉

ScottDowne:
@gvn I have no issues with this, I'll give it a spin and take a closer look at the tests.

ScottDowne:
Actually I'm seeing something in my console logs I didn't see before, figuring out if it's related to this patch or now.

ScottDowne:
ok, my review stands, its working fine and what I was seeing in the console is unrelated.

gvn:
@k88hudson @ScottDowne – Can we land this?

@k88hudson – Any comments/suggestions? 👀

lib/TopStoriesFeed.jsm
```diff
@@ -401,29 +410,43 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+    }
+
+    // Filter spocs based on frequency caps
+    const impressions = this.readImpressionsPref(SPOC_IMPRESSION_TRACKING_PREF);
+    let spocs = this.spocs.filter(s => this.isBelowFrequencyCap(impressions, s));
+
+    // Filter out expired spocs based on `expiration_timestamp`
+    spocs = spocs.filter(spoc => {
+      // If cached data is so old it doesn't contain this property, assume the spoc is ok to show
+      if (typeof spoc.expiration_timestamp === `undefined`) {
```
k88hudson: would `"expiration_timestamp" in spoc === false` work here?

gvn: Yeah it would. Could also do `!("expiration_timestamp" in spoc)`

PR #4558, opened 2018-11-09
Fix Bug 1502501 - Conditionally show "Return to AMO" instead of StartupOverlay for FxA sign-in on about:welcome

sarracini:
This adds some logic so that StartupOveraly for FXA sign-in becomes a message which has targeting, which is the complement of the targeting that we're going to use for Return to AMO. This doesn't actually show the Return to AMO UI conditionally because that UI doesn't exist yet, but this is the set up for that logic.
Best way to test this is to make sure it's working exactly as before. Then try the attribution steps in PR #4526, visit about:welcome, you'll see you don't get the FxA StartupOverlay anymore, and you do get the new sync card in onboarding instead

The next PR that will introduce the UI for Return to AMO will most likely change this to have some state in asrouter-content which knows whether or not to show an overlay in general, as a stand in solution for what we hope to do with assigning "ui spaces" for messages in the future. For simplicity, i just looked at the message template and bundle template to get this working in its original form.

Note that this technically only works because onboarding is a bundle and fxa is a message so they don't conflict and they render separately. That's why i didn't need to add a state for showing onboarding and showing an overlay just yet. Does that make sense?


content-src/asrouter/asrouter-content.jsx
```diff
@@ -181,11 +186,23 @@ export class ASRouterUISurface extends React.PureComponent {
     }
   }
 
+  componentDidMount() {
+    if (this.props.document.location.href === "about:welcome") {
+      // Trigger the onboarding overlay slightly later than the startup overlay is mounted,
```
piatra: This should just be `...after the startup overlay is mounted`

PR #4562, opened 2018-11-13
Fix Bug 1506229 - Run check for duplicate files in TC

piatra:
This is the check that is failing. Do we want to have this as part of TC?
https://searchfox.org/mozilla-central/rev/7f7c353e969e61a6a85201cc8ad3c3de12ac30d8/toolkit/mozapps/installer/packager.mk#38

### DIFF
--- a/mochitest.sh
+++ b/mochitest.sh
@@ -13,6 +13,7 @@ cd /activity-stream && npm install . && npm run buildmc
 
 # Build latest m-c with Activity Stream changes
 cd /mozilla-central && ./mach build \
+  && MOZ_AUTOMATION=1 ./mach package \
   && ./mach lint -l codespell browser/components/newtab \
   && ./mach test browser/components/newtab/test/browser --headless \
   && ./mach test browser/components/newtab/test/xpcshell \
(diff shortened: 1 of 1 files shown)


piatra:
Local vs TC differences in the environment variables. Will investigate.

piatra:
So adding `MOZ_AUTOMATION` runs the additional dedupe test mentioned in comment0 that failed on the export. It also triggered [an extra step](https://searchfox.org/mozilla-central/rev/d850d799a0009f851b5535580e0a8b4bb2c591d7/toolkit/mozapps/installer/packager.mk#97) that failed when searching for enviroment variables.
I just added those variables when running the command because the alternative (running just that tests) seems not as easy, [a lot of arguments are passed in mostly generated](https://treeherder.mozilla.org/logviewer.html#?job_id=210681511&repo=try&lineNumber=2378)
f? @Mardak

Mardak:
I chatted with mshal and suggested a separate variable to trigger just the find-dupes test:
https://bugzilla.mozilla.org/show_bug.cgi?id=1507610

If that lands, then we don't need to worry about the other MOZ_AUTOMATION and other variables that happen to be additionally required.

piatra:
I've updated to use `RUN_FIND_DUPES`.

PR #4563, opened 2018-11-13
Fix Bug 1506594 - Open snippet links in same window

piatra:
I kept privacy and learn more links as `target=_blank` since it isn't the main action of the message.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -928,7 +928,7 @@ class _ASRouter {
         target.browser.ownerGlobal.OpenBrowserWindow({private: true});
         break;
       case ra.OPEN_URL:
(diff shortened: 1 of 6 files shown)


lib/ASRouter.jsm
```diff
@@ -928,7 +928,7 @@ class _ASRouter {
         target.browser.ownerGlobal.OpenBrowserWindow({private: true});
         break;
       case ra.OPEN_URL:
-        target.browser.ownerGlobal.openLinkIn(action.data.args, "tabshifted", {
+        target.browser.ownerGlobal.openLinkIn(action.data.args, "current", {
```
sarracini: This is going to open the onboarding links in the current tab too, which is unwanted behaviour

PR #4565, opened 2018-11-14
Fix Bug 1506838 - Update Send To Device snippet to better match old template

piatra:


content-src/asrouter/rich-text-strings.js
```diff
@@ -7,6 +7,7 @@ import {MessageContext} from "fluent";
  */
 const RICH_TEXT_CONFIG = {
   "text": ["text", "scene1_text"],
+  "scene2_text": "scene2_text",
```
piatra: So this was a weird one. This way of expressing multiple possibilities with `Array` doesn't work if 2 or more options will be present. If you have both `scene1` and `scene2` when it looks for `scene1` text, because in the code we do `.pop()` it will pick from the end `scene2` and offer that translation option. Essentially showing `scene2` text twice.

k88hudson:
cc @glogiotatidis

piatra:
Let me follow up with some specific screenshots of the changes.

piatra:
Fixed the width of the container to 786
grafik

Increased the font size for the button to match the input size.
Increased the legal notice text to match the previous snippet implementation.
Set `scene2_icon` margin to match previous implementation.
screen shot 2018-11-15 at 17 29 06

* `scene2_text` is now rich text
* I didn't do anything about `scene2_title` (which does show up) newline positioning. Still waiting on a response here https://bugzilla.mozilla.org/show_bug.cgi?id=1506600
* added all the defaults

I didn't change any of the colors (buttons or background colors) because the current ones are set to match our color scheme: i.e. dark or light Firefox theme will also update snippets.

content-src/asrouter/templates/SendToDeviceSnippet/SendToDeviceSnippet.jsx
```diff
@@ -27,11 +27,31 @@ function processFormData(input, message) {
...
+      scene1_button_label: "Learn More",
+      scene2_dismiss_button_text: "Dismiss",
+      scene2_button_label: "Send",
+      scene2_input_placeholder: "YOUR EMAIL HERE",
+      locale: "en-US",
+      country: "us",
+      message_id_email: "",
+      ...props.content,
+      // Endpoint can send a reply that can start with an uppercase letter
+      include_sms: (props.content.include_sms || "false").toLowerCase(),
```
glogiotatidis: I doubled checked and the service correctly returns boolean `true` or `false`. Apologies for asking for this change, it seems that this [line](https://github.com/mozmeao/snippets/blob/master/activity-stream/send-to-device.html#L385) confused me.

content-src/asrouter/rich-text-strings.js
k88hudson: O_o

content-src/asrouter/components/SnippetBase/_SnippetBase.scss
```diff
@@ -36,7 +36,7 @@
 
     max-width: $wrapper-max-width-large;
     @media (min-width: $break-point-widest) {
-      max-width: $wrapper-max-width-widest;
```
k88hudson: This change is incorrect, we want the width to change according to the inner content so it can be centred.

Also, `wrapper-max-width-widest` is correct for the spec according to aaron (legacy snippets did not have breakpoints/widths that aligned with new tab content)

content-src/asrouter/templates/SubmitFormSnippet/SubmitFormSnippet.jsx
```diff
@@ -150,7 +150,10 @@ export class SubmitFormSnippet extends React.PureComponent {
     return (<SnippetBase {...this.props} className={containerClass} footerDismiss={true}>
         {content.scene2_icon ? <div className="scene2Icon"><img src={content.scene2_icon} /></div> : null}
         <div className="message">
-          <p>{content.scene2_title ? <h3 className="scene2Title">{content.scene2_title}</h3> : null} {content.scene2_text}</p>
+          <p>
+            {content.scene2_title ? <h3 className="scene2Title">{content.scene2_title}</h3> : null}
+            <RichText scene2_text={content.scene2_text} localization_id="scene2_text" />
```
k88hudson: Putting this on a new line removes the space between the title and body text which is not desired; perhaps you could add space with `{' '}` to avoid this problem in the future.

content-src/asrouter/components/SnippetBase/_SnippetBase.scss
```diff
@@ -5,7 +5,7 @@
   right: 0;
   background-color: var(--newtab-snippets-background-color);
   color: var(--newtab-text-primary-color);
-  font-size: 12px;
```
k88hudson: disclaimer text and privacy notice should still be `12px` after this change

content-src/asrouter/components/SnippetBase/_SnippetBase.scss
```diff
@@ -5,7 +5,7 @@
...
+  font-size: 14px;
   line-height: 16px;
```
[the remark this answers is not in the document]
piatra: I didn't see this in the design spec

content-src/asrouter/components/SnippetBase/_SnippetBase.scss
piatra: Ok I think we want `width` here instead so that they take up the required amount and still adjust on smaller screen.

With `width`
grafik

With `max-width`
grafik

content-src/asrouter/components/SnippetBase/_SnippetBase.scss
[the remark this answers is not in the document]
k88hudson: I'll ask Aaron to add it, but it's visually indicated

content-src/asrouter/components/SnippetBase/_SnippetBase.scss
k88hudson: I'm not seeing that problem with max-width, how can I reproduce that?

This is on nightly, the snippet seems to be aligned appropriately:

image

k88hudson:
Ok looks good, thanks!

PR #4569, opened 2018-11-20
Fix Bug 1507306 - Unit tests running on Firefox 54

piatra:
From travis system info

```
Installed Firefox version
firefox 56.0.2
```

Issue seems to be that the script downloads the "latest" firefox from http://archive.mozilla.org/pub/firefox/tinderbox-builds/mozilla-release-linux64-add-on-devel/ but the newest version in that folder is 54. 
Checking to see if we even have to install firefox.


piatra:
lgtm?

.travis.yml
```diff
@@ -22,10 +22,6 @@ install:
   - npm config set spin false
   - npm install
 
-before_script:
-  - bash bin/download-firefox-travis.sh release-linux64-add-on-devel
```
rlr: probably can remove `bin/download-firefox-travis.sh` now or is it used elsewhere?

piatra: yeah I was just trying to see what happens. Wasn't sure if `npm <cmd>` would pick up the system firefox since our previous script specifically exported it.

PR #4572, opened 2018-11-21
Fix Bug 1508667 - update eslint and plugins to latest versions

rlr:


rlr:
looks like travis is happy 😄

.eslintrc.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   // When adding items to this file please check for effects on sub-directories.
+  "parser": "babel-eslint",
   "parserOptions": {
+    "ecmaVersion": 2018,
```
Mardak: This is already set to `9` (same thing) via mozilla, but maybe explicit is better here https://searchfox.org/mozilla-central/rev/d35199ef15ffa57d190886e20d5df6471faf0870/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#75

PR #4577, opened 2018-11-26
Fix Bug 1508431 - Update new tab breakpoints for better content layouts

gvn:


content-src/components/Sections/Sections.jsx
```diff
@@ -248,11 +248,11 @@ export class Section extends React.PureComponent {
           </div>}
         {id === "topstories" &&
           <div className="top-stories-bottom-container">
-            <div>
+            <div className={`${shouldShowTopics ? `wrapper-topics` : `wrapper-cta`}`}>
```
gvn: This is causing a linting error because it's pushing the complexity of `render` up too high.

`Method 'render' has a complexity of 35  complexity`

@k88hudson @ScottDowne – How do you normally deal with this? Do you set a linting exception per file? I could externalize this logic, but that feels more confusing and harder to maintain.

ScottDowne: I hit this issue in this exact file not long ago,I filed it with something like the first one below. (was the easiest at the time)

I bet if if you had a `const sensibleClassName = shouldShowTopics ? 'wrapper-topics' : 'wrapper-cta'` up top, and `className={sensibleClassName}` in the div, it would fix it.

Another solution could probably something like this (they both shouldn't ever be able to be rendered):
```
          <div className="top-stories-bottom-container">
            {shouldShowTopics &&
              <div className="wrapper-topics">
                <Topics topics={this.props.topics} />
              </div>
            }
            {shouldShowPocketCta &&
              <div className="wrapper-cta">
                <PocketLoggedInCta />
              </div>
            }
            <div className="wrapper-more-recommendations">
              {shouldShowReadMore &&
                <MoreRecommendations read_more_endpoint={read_more_endpoint} />}
            </div>
          </div>
```

but maybe we just need a `{this.renderBottomContainer()}`

gvn:
@ScottDowne This should be good now.

@k88hudson Can we ignore the Taskcluster failure?

Mardak:
Looks like taskcluster failure is needing a port of Bug 1453264 - Dropping nsIBrowserSearchService::currentEngine in favour of defaultEngine

Mardak:
I merged central to master, so if you rebase, taskcluster should be happy https://github.com/mozilla/activity-stream/commit/bea035f1836519f3b8a30d24892ba24c8174c9d8

gvn:
Thanks @Mardak !

ScottDowne:
@gvn interesting solution to the 1px breakpoint range, I never thought of that when I dealt with this, I just removed the uses of max-width.

I think this is fine. The only issue I can see is if you zoom in using ctrl +, the missing breakpoint range somehow manages to come back, probably something about a half pixel range? Dunno if we care, super fringe, but it is pretty interesting. We would need to refactor to remove max-width or min-width. I found some words on the issue here: http://damienclarke.me/code/posts/those-1px-gaps-between-media-queries-can-be-a-problem

Anyway, unless @gvn you want to do any more on this, or unless @k88hudson has any strong opinions here, I say we land it.

gvn:
I'm fine landing it with that edge case since it's really edgy (zoomed *and* within 1px specific range) and considering our 2019 plans.

k88hudson:
Yep this seems fine to me. Merge away!

PR #4578, opened 2018-11-26
Fix Bug 1475354 - Create first-run template and content for Return To AMO experience

sarracini:
To test this, open `about:newtab#asrouter`, scroll all the way to the bottom, click on the button labeled "Force Attribution" and open `about:welcome` and observe that you see the return to amo overlay. Click through to add the extension, make sure it successfully adds it (or doesn't if you cancel), and then it should move on to the simplified onboarding overlay. Clicking the 'Get Started with Firefox' button should dismiss the return to amo overlay, and show the simplified onboarding overlay as well. Then, go back to `about:newtab#asrouter`, and change the `source` and `campaign` fields to be something else, then visit `about:welcome`. Notice now you'll see the original FxA overlay, followed by the simplified onboarding overlay (the original flow).
Spec for reference : https://mozilla.invisionapp.com/share/Y5I1DJYT8KU#/screens/332388691_FirstRun-Specs


content-src/asrouter/templates/ReturnToAMO/ReturnToAMO.jsx
```diff
@@ -0,0 +1,54 @@
...
+    document.body.classList.remove("welcome", "hide-main", "amo");
+  }
+
+  onBlockButton() {
+    this.props.onBlock();
+    document.body.classList.remove("welcome", "hide-main", "amo");
+  }
+
+  renderText() {
+    const customElement = <img src={this.props.content.addon_icon} width="20px" height="20px" />;
```
sarracini: This allows us to place the icon before the name of the addon in any language

content-src/asrouter/templates/ReturnToAMO/_ReturnToAMO.scss
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  p {
+    color: $grey-90-50;
+    font-size: 14px;
+    line-height: 18px;
+    margin-bottom: 4px;
+  }
+
+  .puffy {
```
sarracini: ok `.puffy`,`.blue`, `.default` and `.grey` and button styles for these buttons according to the spec and https://design.firefox.com/photon/components/buttons.html. They are every so slightly different than the regular buttons in activity stream or as router so they needed their own styles

lib/ASRouter.jsm
```diff
@@ -921,6 +956,27 @@ class _ASRouter {
     }
   }
 
+  async forceAttribution(data) {
```
sarracini: This is for `about:newtab#asrouter`

lib/ASRouter.jsm
```diff
@@ -994,6 +1050,9 @@ class _ASRouter {
         await this.blockProviderById(action.data.id);
         this.messageChannel.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {type: "CLEAR_PROVIDER", data: {id: action.data.id}});
         break;
+      case "DISMISS_BUNDLE":
```
sarracini: I'm leaving `BLOCK_BUNDLE` action in here because tests were complaining about coverage

content-src/asrouter/templates/ReturnToAMO/_ReturnToAMO.scss
```diff
@@ -0,0 +1,131 @@
+.ReturnToAMOOverlay,
+.amo + body.hide-main {  // sass-lint:disable-line no-qualifying-elements
+  background: $white;
```
k88hudson: Not sure what colour exactly because there's not a specification in the invision, but this is supposed to be light grey (I'm assuming the same grey as the new tab page)

content-src/asrouter/templates/ReturnToAMO/_ReturnToAMO.scss
```diff
@@ -0,0 +1,131 @@
...
+  position: fixed;
+  top: 0;
+  width: 100%;
+  display: block;
+  z-index: 2100;
+
+  .ReturnToAMOOuterContainer {
+    width: 835px;
+    height: 400px;
+    position: fixed;
```
k88hudson: Any reason you didn't want to use `flex`/ center this? Is the spec supposed to be fixed to the top?

lib/ASRouter.jsm
k88hudson: can you add a comment that says that basically?

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -352,6 +359,47 @@ export class ASRouterAdmin extends React.PureComponent {
       </tbody></table>);
   }
 
+  onChangeAttributionParameters(event) {
+    const {name} = event.target;
```
k88hudson: These can be one line `{name, value} = event.target`

lib/OnboardingMessageProvider.jsm
```diff
@@ -125,20 +173,31 @@ const OnboardingMessageProvider = {
...
-      let translatedMessage = msg;
       // If the message has no content, do not attempt to translate it
       if (!msg.content) {
         translatedMessages.push(msg);
         continue;
       }
-      const [button_string, title_string, text_string] = await L10N.formatMessages([
-        {id: msg.content.button_label.string_id},
+
+      let translatedMessage = msg;
```
k88hudson: Any chance you'd be modifying a source object here? do you need to make a copy with `{...msg}`?

sarracini: I'll change it just in case.... you can never be too sure with javascript

lib/ASRouter.jsm
```diff
@@ -43,6 +44,8 @@ const MAX_MESSAGE_LIFETIME_CAP = 100;
 const LOCAL_MESSAGE_PROVIDERS = {OnboardingMessageProvider, CFRMessageProvider, SnippetsTestMessageProvider};
 const STARTPAGE_VERSION = "6";
 
+const ADDONS_API_URL = "https://services.addons.mozilla.org/api/v3/addons/addon";
```
mixedpuppy: Retrieving the data should have used AddonRepository.jsm so there would be no need to hard code a url here.  At the very least, there should be a followup to remove this and use the pref.  Same with CFRPageActions.jsm.

piatra: @mixedpuppy the code was refactored and we're no longer making this call.
https://github.com/mozilla/activity-stream/blob/master/lib/OnboardingMessageProvider.jsm

PR #4579, opened 2018-11-26
Fix Bug 1509810 - Add missing defaults for snippet templates

piatra:
This addresses 1509810 and 1509851.

I also updated the tests to create assertions out of every schema property with a `default` field so that we don't miss them anymore. Meanwhile I found some minor inconsistencies which I also updated.

Regarding `donation_form_url` it looks like setting `form_action` to a URL with GET parameters will drop those parameters on form submit. Workaround was to convert those parameters to `input[type=hidden]`. This needs an update on the snippets side: the url needs to contain a `?` so we know where the paramters begin in the url. This doesn't seem to be an issue and support in the legacy implementation as well https://github.com/mozmeao/snippets/blob/94b18ab63703ad339b24db206e065061aaf0b30f/activity-stream/mofo-eoy-2017.html#L293 cc @glogiotatidis

### DIFF
--- a/content-src/asrouter/templates/SendToDeviceSnippet/SendToDeviceSnippet.schema.json
+++ b/content-src/asrouter/templates/SendToDeviceSnippet/SendToDeviceSnippet.schema.json
@@ -65,11 +65,6 @@
       "type": "string",
       "description": "Small icon that shows up before the title / text. 16x16px. SVG or PNG preferred. Grayscale."
     },
-    "scene2_email_placeholder_text": {
-      "type": "string",
-      "description": "Value to show while input is empty.",
-      "default": "Your email here"
-    },
     "scene2_button_label": {
       "type": "string",
       "description": "Label for form submit button",
--- a/test/unit/asrouter/templates/EOYSnippet.test.jsx
+++ b/test/unit/asrouter/templates/EOYSnippet.test.jsx
@@ -32,8 +32,10 @@ describe("EOYSnippet", () => {
       onAction: sandbox.stub(),
       onBlock: sandbox.stub(),
     };
-    assert.jsonSchema(props.content, schema);
-    return mount(<EOYSnippet {...props} />);
+    const comp = mount(<EOYSnippet {...props} />);
+    // Check schema with the final props the component receives (including defaults)
+    assert.jsonSchema(comp.children().get(0).props.content, schema);
+    return comp;
   }
 
   beforeEach(() => {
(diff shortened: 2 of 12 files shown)


content-src/asrouter/templates/SendToDeviceSnippet/SendToDeviceSnippet.schema.json
```diff
@@ -65,11 +65,6 @@
       "type": "string",
       "description": "Small icon that shows up before the title / text. 16x16px. SVG or PNG preferred. Grayscale."
     },
-    "scene2_email_placeholder_text": {
```
k88hudson: Ah, I guess this is not needed because we have `scene2_input_placeholder`?

k88hudson:
This looks good to me as long as giorgos confirms that the strategy for parsing params is OK.

I didn't see a fix for the checkbox issue on linux, did you want to include that fix?

test/unit/asrouter/templates/EOYSnippet.test.jsx
```diff
@@ -84,6 +97,17 @@ describe("EOYSnippet", () => {
     assert.notCalled(onBlockStub);
   });
 
+  it("it should preserve URL GET params as hidden inputs", () => {
+    wrapper = mountAndCheckProps({donation_form_url: "https://donate.mozilla.org/pl/?utm_source=desktop-snippet&amp;utm_medium=snippet&amp;utm_campaign=donate&amp;utm_term=7556"});
+
+    const hiddenInputs = wrapper.find("input[type='hidden']");
+
+    assert.propertyVal(hiddenInputs.find("[name='utm_source']").props(), "value", "desktop-snippet");
```
k88hudson: should we include a test for URL-encoded values here if they are allowed?

content-src/asrouter/templates/SendToDeviceSnippet/SendToDeviceSnippet.schema.json
piatra: Yes, the other field (the one removed) is not present in the template

https://github.com/mozmeao/snippets/blob/master/activity-stream/send-to-device.html#L25

glogiotatidis:
> This needs an update on the snippets side: the url needs to contain a `?` so we know where the paramters begin in the url.

I don't understand the ask here. If we're passing parameters we url will contain a `?`. Otherwise I guess it's safe to append `?` in the url while you're doing the arg parsing dance?

piatra:
@glogiotatidis `donation_form_url` should have a `?` before the query parameters like `<DOMAIN>/?foo=foo`. 

I came across this kind of url

```
https://donate.mozilla.org/de/?utm_source=desktop-snippet&utm_medium=snippet&utm_campaign=FundSeptV1R2&utm_term=9422&utm_content=REL_ESR&sample_rate=0.001&snippet_name=9422&amount=undefined&currency=usd&presets=50,25,10,3&frequency=single
```

Which is ok, but also 

```
https://donate.mozilla.org/pl/utm_source=desktop-snippet&amp;utm_medium=snippet&amp;utm_campaign=donate&amp;utm_term=7556
```

Without the `?` which would fail given how we currently implemented handling the URL parameters.
Reading the "legacy" EOY snippet implementation both seem to be handled but we wanted to make sure only the first case will be used from now on.

glogiotatidis:
Agreed. This is not something I can enforce on the server level though. This is an incorrectly inputted string by the Snippet Editors. I will migrate what we have now to ensure that we use `?` when utm params are passed and communicate with them the correct use.

k88hudson:
Is this obsolete due to  #4581?

PR #4580, opened 2018-11-27
Fix Bug 1509755 - Wait for browser initialization to finish before adding CFR trigger listeners

piatra:
This should help us ensure we're not trying to add that listener too early in the startup process. 

I have a try run https://treeherder.mozilla.org/#/jobs?repo=try&revision=942859ad6573d7f6c7e203a8004685de685f0f8c&selectedJob=213978989 that shows a failure linking to the same issue but I think it's missclassified.
Checking the failure logs shows
```
TEST-UNEXPECTED-FAIL | browser/base/content/test/plugins/browser_CTP_drag_drop.js | leaked 1 window(s) until shutdown [url = http://127.0.0.1:8888/browser/browser/base/content/test/plugins/plugin_test.html]
```
This does not correspond to the failures associated with this bug(s) it should specifically mention `win.gBrowser is undefined;`

### DIFF
--- a/lib/ASRouterTriggerListeners.jsm
+++ b/lib/ASRouterTriggerListeners.jsm
@@ -24,6 +24,10 @@ this.ASRouterTriggerListeners = new Map([
     _triggerHandler: null,
     _hosts: null,
 
+    /**
+     * Wait for browser startup to finish to avoid accessing uninitialized
+     * properties
+     */
     async _checkStartupFinished(win) {
       if (!win.gBrowserInit.delayedStartupFinished) {
         await new Promise(resolve => {
--- a/lib/ASRouterTriggerListeners.jsm
+++ b/lib/ASRouterTriggerListeners.jsm
@@ -24,11 +24,24 @@ this.ASRouterTriggerListeners = new Map([
     _triggerHandler: null,
     _hosts: null,
 
+    async _checkStartupFinished(win) {
+      if (!win.gBrowserInit.delayedStartupFinished) {
+        await new Promise(resolve => {
+          let delayedStartupObserver = () => {
+            Services.obs.removeObserver(delayedStartupObserver, "browser-delayed-startup-finished");
+            resolve();
+          };
+
+          Services.obs.addObserver(delayedStartupObserver, "browser-delayed-startup-finished");
+        });
+      }
+    },
+
     /*
      * If the listener is already initialised, `init` will replace the trigger
      * handler and add any new hosts to `this._hosts`.
      */
-    init(triggerHandler, hosts = []) {
+    async init(triggerHandler, hosts = []) {
       if (!this._initialized) {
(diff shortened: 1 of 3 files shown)


k88hudson:
Is this obsolete due to  #4581?

lib/ASRouterTriggerListeners.jsm
```diff
@@ -24,11 +24,28 @@ this.ASRouterTriggerListeners = new Map([
...
 
+    /**
+     * Wait for browser startup to finish to avoid accessing uninitialized
+     * properties
+     */
+    async _checkStartupFinished(win) {
+      if (!win.gBrowserInit.delayedStartupFinished) {
+        await new Promise(resolve => {
+          let delayedStartupObserver = () => {
+            Services.obs.removeObserver(delayedStartupObserver, "browser-delayed-startup-finished");
```
Mardak: I see similar code in browser.js explicitly checking the message and target… maybe it's just being unnecessarily careful? https://searchfox.org/mozilla-central/rev/b4ebbe90ae4d0468fe6232bb4ce90699738c8125/browser/base/content/browser.js#1166-1171

lib/ASRouterTriggerListeners.jsm
```diff
@@ -24,11 +24,28 @@ this.ASRouterTriggerListeners = new Map([
     _triggerHandler: null,
     _hosts: null,
 
+    /**
+     * Wait for browser startup to finish to avoid accessing uninitialized
+     * properties
+     */
+    async _checkStartupFinished(win) {
+      if (!win.gBrowserInit.delayedStartupFinished) {
+        await new Promise(resolve => {
```
Mardak: Some reason making the method async to await a single inline Promise without doing anything after doesn't seem quite right… but I guess it works ;)

lib/ASRouterTriggerListeners.jsm
```diff
@@ -24,11 +24,28 @@ this.ASRouterTriggerListeners = new Map([
...
+      }
+    },
+
     /*
      * If the listener is already initialised, `init` will replace the trigger
      * handler and add any new hosts to `this._hosts`.
      */
-    init(triggerHandler, hosts = []) {
+    async init(triggerHandler, hosts = []) {
       if (!this._initialized) {
```
Mardak: Do we need to worry about init being called multiple times? Potentially multiple concurrent `init`s are waiting for startup to finish because `_initialized` is set at the end.

I suppose ASRouter's `await ASRouterTriggerListeners.get(trigger.id).init(…)` should guard this even if the trigger id repeats as it'll wait.

lib/ASRouterTriggerListeners.jsm
piatra: Given the source of the bug I think it's a good idea. Added the check.

PR #4581, opened 2018-11-27
Fix Bug 1500563 - Use one schema for every snippet template and other ASR snippet fixes

piatra:
Fixes bugs 1500563, 1509810 and 1509851.

Combine https://github.com/mozilla/activity-stream/commit/afc948a958ef90094e8b6b8d28c7ddef1a8faeca and https://github.com/mozilla/activity-stream/pull/4579/commits

Also drop checkbox input style because of platform inconsistencies.


k88hudson:
Looks like checkboxes are still not looking right on linux due to width/height, can you either get rid of that or add a specific rule for linux?

PR #4587, opened 2018-12-11
Fix Bug 1511156 - Windows attribution data is double encoded

piatra:


lib/OnboardingMessageProvider.jsm
```diff
@@ -9,10 +9,13 @@ ChromeUtils.import("resource://gre/modules/addons/AddonRepository.jsm");
 
 async function getAddonName() {
   try {
-    const {content} = await AttributionCode.getAttrDataAsync();
+    let {content} = await AttributionCode.getAttrDataAsync();
     if (!content) {
       return null;
     }
+    while (content.includes("%")) {
+      content = decodeURIComponent(content);
```
Mardak: An actual `%` without an invalid encoding will result in `URIError: malformed URI sequence` e.g.,

`decodeURIComponent("%")`

piatra: I was just working on fixing that. 😁

lib/OnboardingMessageProvider.jsm
```diff
@@ -9,10 +9,22 @@ ChromeUtils.import("resource://gre/modules/addons/AddonRepository.jsm");
 
 async function getAddonName() {
   try {
-    const {content} = await AttributionCode.getAttrDataAsync();
+    let {content} = await AttributionCode.getAttrDataAsync();
     if (!content) {
       return null;
     }
+    // Attribution data can be double encoded
+    while (content.includes("%")) {
```
Mardak: This code assumes the content can be encoded any number of times and that it should be decoded all the way. Are we sure that we won't need to pass along any encoded content? Even if we only try decoding twice, non-windows platforms without the double encoding would then incorrectly be decoded. In particular, will we never want to use a "%40" without decoding to "@" (e.g., if windows gives us "%252540" for a triple-encoded "@" attempting to convey "%40" but we incorrectly decode to "@"?)

Mardak:
Notably, the unit tests for attribution seem to sometimes want encoded content: https://searchfox.org/mozilla-central/rev/3160ddc1f0ab55d230c595366662c62950e5c785/browser/components/attribution/test/xpcshell/test_AttributionCode.js#11-28

lib/OnboardingMessageProvider.jsm
mixedpuppy: I think in this particular use case it should be decoded all the way.  content here is specifically and extension id.  Seems like source should also be verified before doing anything with content.

PR #4589, opened 2018-12-12
Fix Bug 1511165 - about:welcome addon card on windows says "addon name"

piatra:
In order to test this go to `about:newtab#asrouter` and use the force attribution button at the button of the page. After that navigate to `about:welcome` and you should get the onboarding message that prompts you to install the addon. You can change it to test with a different addon: you can use the following value for the `content` field (again on the `about:newtab#asrouter` page) `jid1-MnnxcxisBPnSXQ@jetpack` and this should force Privacy Badger as a message.

Requires #4587 to land first.

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -26,12 +26,16 @@ async function getAddonInfo() {
       }
     }
     const [addon] = await AddonRepository.getAddonsByIDs([content]);
+    if (addon.sourceURI.scheme !== "https") {
+      return null;
+    }
     return {
       name: addon.name,
       url: addon.sourceURI.spec,
       iconURL: addon.icons["64"] || addon.icons["32"],
     };
   } catch (e) {
+    Cu.reportError("Failed to get the latest add-on version for Return to AMO");
     return null;
   }
 }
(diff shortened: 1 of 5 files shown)


lib/OnboardingMessageProvider.jsm
```diff
@@ -25,8 +25,12 @@ async function getAddonName() {
         break;
       }
     }
-    const addons = await AddonRepository.getAddonsByIDs([content]);
-    return addons[0].name;
+    const [addon] = await AddonRepository.getAddonsByIDs([content]);
+    return {
+      name: addon.name,
+      url: addon.sourceURI.spec,
```
Mardak: I realize this is code moved from ASRouter.jsm, but could we check that sourceURI.scheme === "https"

piatra:
I think the issue is caused by the fact that we had 2 different APIs to fetch the information.
`name` came from the AddonsRepository API and that failed because of the double encoding while the addon download url and icon was retrieved using fetch `fetch(`${ADDONS_API_URL}/${content}`);` and that might have worked around the enconding issue.
grafik


Also Fluent fails with the string name when null/undefined is passed in.
grafik

Now we simply abort if any information is missing (including name) which also helps with the "Emergency off plan for RTAMO" bug https://bugzilla.mozilla.org/show_bug.cgi?id=1513317#c1

PR #4593, opened 2018-12-14
Fix Bug 1514337 - Activity Stream layout pref and layout data for rapid experimentation

ScottDowne:


lib/TopStoriesFeed.jsm
```diff
@@ -195,6 +204,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
     const data = await this.cache.get();
     let stories = data.stories && data.stories.recommendations;
     let topics = data.topics && data.topics.topics;
+    this.dispatchLayoutUpdate(data.layout);
```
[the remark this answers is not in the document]
ScottDowne: Yup, this looks like a bug. Good catch.

lib/ActivityStream.jsm
```diff
@@ -73,6 +73,7 @@ const PREFS_CONFIG = new Map([
...
       show_spocs: showSpocs(args),
       personalized: true,
+      use_layout: false,
```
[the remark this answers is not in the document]
ScottDowne: I dunno, no huge reason, but I guess it's explicit and more reliable (layout/feed can be stuck behind a cache)

But tbh I'm not sure if layout being cached is actually an issue, and I don't know if it being more explicit is helpful at all.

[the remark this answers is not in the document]
ScottDowne: It's also follow a pattern already used for spocs.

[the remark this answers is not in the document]
ScottDowne: Oh, I guess it also means if people want to opt out, we can just get this pref flipped?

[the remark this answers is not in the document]
Mardak: If we expose UI for toggling this part of the JSON pref, it will prevent any future default values as firefox sees the whole pref as a custom value.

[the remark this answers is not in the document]
ScottDowne: Ah I see, is that why we have a showSponsored pref on top of the show_spocs pref inside the json?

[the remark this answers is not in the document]
dmose: @Mardak that's true, but we do need to support the opt-out somehow, and that would be controlled separately from the stuff that the feeds send down, wouldn't it?

If so, this could end up being the opt-out pref, I'd think.

CC @piatra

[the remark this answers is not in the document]
dmose: (And the JSON pref that normandy changes would be different)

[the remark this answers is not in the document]
dmose: (Assuming we go with a JSON pref, which does seem like a reasonable plan.)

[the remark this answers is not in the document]
ScottDowne: I'm fine with dropping the use_layout pref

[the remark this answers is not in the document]
dmose: @ScottDowne so I'm assuming that your concern with this code (outside of the pref bit) is simply getting something stood up as a basis for further work.  If that's true, we can certainly sort out the pref store (opt-out vs JSON, etc) separately.  If you want to go that way, just update the patch here, and I'll review it this evening.

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -1460,4 +1460,38 @@ describe("Top Stories Feed", () => {
     assert.calledOnce(instance.uninit);
     assert.calledOnce(instance.init);
   });
+  describe("#layout", () => {
+    it("should call dispatchLayoutUpdate from loadCachedData and fetchStories", async () => {
```
dmose: For a single test, this is pretty complex.  I'd suggest splitting it in two.

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -1460,4 +1460,38 @@ describe("Top Stories Feed", () => {
...
+        "layout": [1, 2],
+      };
+      globals.set("fetch", fetchStub);
+      fetchStub.resolves({ok: true, status: 200, json: () => Promise.resolve(response)});
+      sinon.spy(instance, "dispatchLayoutUpdate");
+      instance.cache.get = () => ({stories: {layout: [2, 3]}});
+
+      await instance.loadCachedData();
+      assert.calledOnce(instance.dispatchLayoutUpdate);
+      const firstCallArgs = instance.dispatchLayoutUpdate.firstCall.args;
```
dmose: The firstCall, secondCall argument grovelling makes the test fairly difficult to skim.  If, instead, it were switched to spy.withArgs(), creating a spy that only fired when certain arguments were passed, it would likely be notably more readable.

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -1460,4 +1460,38 @@ describe("Top Stories Feed", () => {
...
+      assert.equal(firstCallArgs[0][0], 2);
+      assert.equal(firstCallArgs[0][1], 3);
+
+      await instance.fetchStories();
+      assert.calledTwice(instance.dispatchLayoutUpdate);
+      const secondCallArgs = instance.dispatchLayoutUpdate.secondCall.args;
+      assert.equal(secondCallArgs[0][0], 1);
+      assert.equal(secondCallArgs[0][1], 2);
+    });
+    it("should call dispatch from dispatchLayoutUpdate if preffed on and with available data", () => {
```
dmose: "if preffed on and" is probably no longer necessary here.  :-)

dmose: Do we want another test here that makes sure that if the data is unavailable, that nothing gets dispatched?

test/unit/lib/TopStoriesFeed.test.js
dmose: Assuming the above is true, I'd also switch the other newly added tests to this mechanism.

test/unit/lib/TopStoriesFeed.test.js
ScottDowne: k

test/unit/lib/TopStoriesFeed.test.js
ScottDowne: Because of splitting it in two, I could easily just run `calledWith` to easily check the values. I couldn't as easily do this before because I was testing a first and second call.

Does that work?

test/unit/lib/TopStoriesFeed.test.js
ScottDowne: done

k88hudson:
Looking great!

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -1460,4 +1460,41 @@ describe("Top Stories Feed", () => {
...
+      assert.calledOnce(instance.dispatchLayoutUpdate);
+      assert.calledWith(instance.dispatchLayoutUpdate, [2, 3]);
+    });
+    it("should call dispatch from dispatchLayoutUpdate with available data", () => {
+      instance.dispatchLayoutUpdate([1, 2]);
+      assert.calledOnce(instance.store.dispatch);
+      const [action] = instance.store.dispatch.firstCall.args;
+      assert.equal(action.type, "CONTENT_LAYOUT");
+      assert.equal(action.data[0], 1);
+      assert.equal(action.data[1], 2);
```
dmose: These last two lines would be more readable if written as:

```
assert.equal(action.data, [1, 2])
```

lib/TopStoriesFeed.jsm
```diff
@@ -112,6 +112,13 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.store.dispatch(shouldBroadcast ? ac.BroadcastToContent(action) : ac.AlsoToPreloaded(action));
   }
 
+  dispatchLayoutUpdate(data, shouldBroadcast) {
```
dmose: For improved readability of calling code, this I'd suggest calling this `maybeDispatchLayoutUpdate`, since it may or may not call dispatch under the hood.

test/unit/common/Reducers.test.js
```diff
@@ -655,4 +655,13 @@ describe("Reducers", () => {
       assert.equal(state.pocketCta.useCta, data.use_cta);
     });
   });
+  describe("Layout", () => {
+    it("should return INITIAL_STATE by default", () => {
+      assert.equal(Layout(undefined, {type: "some_action"}), INITIAL_STATE.Layout);
+    });
+    it("should set layout with CONTENT_LAYOUT", () => {
```
dmose: Maybe "it should set layout.type to CONTENT_LAYOUT"?

ScottDowne: How about "should set layout data with layout.type CONTENT_LAYOUT"?

test/unit/lib/TopStoriesFeed.test.js
ScottDowne: This didn't actually work, I think I need some sort of deep equal? I'll try that.

test/unit/common/Reducers.test.js
dmose: Sure

test/unit/lib/TopStoriesFeed.test.js
dmose: Thanks.

test/unit/lib/TopStoriesFeed.test.js
dmose: Totally.

test/unit/lib/TopStoriesFeed.test.js
dmose: Thanks.

test/unit/lib/TopStoriesFeed.test.js
dmose: Ah, gotta love complex JS equality semantics.  :-)

PR #4597, opened 2018-12-18
Fix Bug 1514890 - Change installation source for addons installed from ActivityStream

rpl:
This PR is a follow up related to the changes previously applied in #4467, per Bug 1514890 we would like to change the installation source associated to addons installed from ActivityStream.


ddurst:
@k88hudson Can we get this in for 66? It's just a value change.

k88hudson:
Oops sorry, this fell off my radar – yes we can get this in.

PR #4599, opened 2018-12-19
Fix Bug 1508388 - Search hand-off

rlr:
This adds a new pref "improvesearch.handoffToAwesomebar".
* When the pref is enabled, we replace the current search box+button with a button that is styled to be just like the current search.
* When the button is clicked (mousedown), we transfer focus to the awesomebar but suppress the awesomebar's focus styles.
* When the button is clicked (keydown), we transfer focus normally to the awesomebar.
* Note that I still had to include a hidden `<input type="text".../>` so that we could load ContentSearchUIController because that has code that sets `--newtab-search-icon` for us. I initially tried to implement that with a `SearchFeed` but it is way more involved than I thought. There is a bunch of work that needs to happen to convert the icon data into a blob. We can figure that out later on if/when we ditch our current search implementation.


lib/ActivityStream.jsm
```diff
@@ -192,6 +192,10 @@ const PREFS_CONFIG = new Map([
     title: "A comma-delimited list of search shortcuts that have previously been pinned",
     value: "",
   }],
+  ["improvesearch.handoffToAwesomebar", {
+    title: "Should the search box handoff to the Awesomebar?",
+    value: true,
```
rlr: Should we land this enabled on Nightly?

rlr: I brought it up in my 1:1, Tim thinks we should ship it enable on Nightly and Beta. Then we'll do a study in Release before rolling out. If something goes wrong in Nightly it's easy enough to disable it with a one liner patch.

k88hudson: Ok, sounds good to me 👍

common/Reducers.jsm
```diff
@@ -439,10 +445,23 @@ function Layout(prevState = INITIAL_STATE.Layout, action) {
   }
 }
 
+function Search(prevState = INITIAL_STATE.Search, action) {
+  switch (action.type) {
+    case at.HIDE_SEARCH:
+      return Object.assign({}, prevState, {hide: true});
```
k88hudson: Just use destructuring for this, i.e. `{...prevState, hide: true}`

common/Reducers.jsm
```diff
@@ -439,10 +445,23 @@ function Layout(prevState = INITIAL_STATE.Layout, action) {
...
+      return prevState;
+  }
+}
+
 this.INITIAL_STATE = INITIAL_STATE;
 this.TOP_SITES_DEFAULT_ROWS = TOP_SITES_DEFAULT_ROWS;
 this.TOP_SITES_MAX_SITES_PER_ROW = TOP_SITES_MAX_SITES_PER_ROW;
 
-this.reducers = {TopSites, App, ASRouter, Snippets, Prefs, Dialog, Sections, Pocket, Layout};
+this.reducers = {TopSites, App, ASRouter, Snippets, Prefs, Dialog, Sections, Pocket, Layout, Search};
```
k88hudson: Would you mind splitting these onto different lines now that there's a lot?

lib/ActivityStream.jsm
k88hudson: Let's go with on for now and talk about our plan for testing it (I think we should be careful about the possible impact to our search numbers this could have)

If you're going to land this with just `true`, would you mind filing another issue to turn it off for release etc. so we dont forget about that work?

k88hudson:
Also: kind of an edge-case, but if you "focus" on the input and then scroll the new page, it doesn't become sticky. Probably doesn't matter too much though? Might be worth checking in with aaron

rlr:
@k88hudson OK. All fixed ^

I have notes to:

- [ ] add detailed comment for QA
- [ ] file bug to turn off pref before Release
- [ ] request a11y review
- [ ] request UX to review

rlr:
@k88hudson oh! is there anything special I need to do to the pref so that it can be rolled out? I think I heard you mention something about firefox js prefs in some other cases.

k88hudson:
@rlr in order to make it roll-out-able you would need to put it in firefox.js https://searchfox.org/mozilla-central/source/browser/app/profile/firefox.js#1282, you could always do that as a follow-up though

k88hudson:
YAY! This is great!!!

One remaining issue is that when you press a key other than an actual letter (such as shift) it hides the element, that might not be desired? That could be filed as a follow-up though (and we should maybe ask design).

Thanks!

rlr:
My TODO list:

- [ ] add detailed comment in bug for QA
- [ ] file bug to turn off pref before Release
- [ ] request a11y review
- [ ] request UX to review (including what to do with modifier keys (shift, alt, ctrl)?)
- [ ] file bug to put pref in firefox.js

PR #4622, opened 2018-12-20
Fix Bug 1514027 - Create a topics header section component for Activi…

ScottDowne:
…ty Stream experiment functionality

### DIFF
--- a/content-src/components/Sections/Sections.jsx
+++ b/content-src/components/Sections/Sections.jsx
@@ -1,13 +1,13 @@
 import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 import {Card, PlaceholderCard} from "content-src/components/Card/Card";
 import {FormattedMessage, injectIntl} from "react-intl";
+import {Topics, TopicsHeader} from "content-src/components/Topics/Topics";
 import {CollapsibleSection} from "content-src/components/CollapsibleSection/CollapsibleSection";
 import {ComponentPerfTimer} from "content-src/components/ComponentPerfTimer/ComponentPerfTimer";
 import {connect} from "react-redux";
 import {MoreRecommendations} from "content-src/components/MoreRecommendations/MoreRecommendations";
 import {PocketLoggedInCta} from "content-src/components/PocketLoggedInCta/PocketLoggedInCta";
 import React from "react";
-import {Topics} from "content-src/components/Topics/Topics";
 import {TopSites} from "content-src/components/TopSites/TopSites";
 
 const VISIBLE = "visible";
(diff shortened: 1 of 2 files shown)


ScottDowne:
I'm currently testing this with this data: https://gist.github.com/ScottDowne/164995d9535b4203846048bdee29d169

It gives you a good idea as to how we can consume props from data and plug it into a component.

The component is not styled.

content-src/components/Sections/Sections.jsx
```diff
@@ -306,14 +306,32 @@ export class _Sections extends React.PureComponent {
     return sections;
   }
 
+  renderComponent(component) {
+    const {props} = component;
+    if (component.type === "TopicsHeader") {
```
ScottDowne: I imagine we would expand the if here in some way to handle more components, dunno if we want to key the component names, seems logical, but for now I'm not making that assumption.

content-src/components/Sections/Sections.jsx
```diff
@@ -306,14 +306,32 @@ export class _Sections extends React.PureComponent {
...
+      const {topics} = this.props.Sections.find(section => section.id === "topstories") || {};
+      return (
+        <TopicsHeader
+          topics={topics}
```
ScottDowne: I'm using the old topics returned from the old topics endpoint, I imagine we want to lock this down to be new test data specific to our tests.

content-src/components/Sections/Sections.jsx
```diff
@@ -306,14 +306,32 @@ export class _Sections extends React.PureComponent {
...
+          headerText={props.headerText}
+          moreRecommendationsText={props.moreRecommendationsText}
+          moreRecommendationsLink={props.moreRecommendationsLink}
+          learnMoreText={props.learnMoreText}
+          learnMoreLink={props.learnMoreLink}
+          learnMoreMessage={props.learnMoreMessage} />
+      );
+    }
+    return (<div>{component.type}</div>);
```
ScottDowne: For now, we just return a stub for other component types.

content-src/components/Topics/Topics.jsx
```diff
@@ -19,3 +19,30 @@ export class Topics extends React.PureComponent {
     );
   }
 }
+
+export class TopicsHeader extends React.PureComponent {
```
ScottDowne: The previous topics component was build to work inside a cards section, so I felt forking here was logical.

content-src/components/Sections/Sections.jsx
dmose: Having to have a separate dispatcher just to manage the prop-passing seems a bit surprising, I wonder if there's a way to make this simpler.  Maybe it's something we can get redux to do for us?  That said, an XXX comment about this might be the right thing for now.

ScottDowne: Hm, what do you mean by dispatcher here?

Mardak: I'm assuming dmose was referring to the duplication of reading out each property to pass it on. You could probably do something like:
```js
const validComponents = {
  TopicsHeader
};
const extraProperties = {};
switch (type) {
  case "TopicsHeader":
    extraProperties.topics = …
    break;
}
return (<validComponents[type] {...properties} {...extraProperties} />);
```

ScottDowne:
This helped me think through this patch: https://github.com/mozilla/activity-stream/pull/4630

which landed, so I am going to close this one.

PR #4627, opened 2018-12-21
Fix Bug 1515411 - Disable Return to AMO for 65 Beta and Release

piatra:
I've added a field named `exclude` in the `OnboardingMessagesProvider` preferences that will populate the blocked messages state. This is used to disable a particular message within the local provider.

I also removed the targeting on the FxA message so we fallback to it when RTAMO is blocked.

When selecting the next message to send `sortMessagesByTargeting` will have targeted messages in the first positions (like RTMO) and ones without targeting at the end making FxA a fallback.

I've removed the `triggerOnboarding` from `asrouter-content` because the `TRIGGER` message was being sent twice: once when `asrouter-content` loaded (from it's `componentWillMount` method) and then again when either FxA or RTAMO messages loaded.

### DIFF
--- a/content-src/asrouter/asrouter-content.jsx
+++ b/content-src/asrouter/asrouter-content.jsx
@@ -144,10 +144,6 @@ export class ASRouterUISurface extends React.PureComponent {
     return () => ASRouterUtils.dismissBundle(bundle);
   }
 
-  triggerOnboarding() {
-    ASRouterUtils.sendMessage({type: "TRIGGER", data: {trigger: {id: "showOnboarding"}}});
-  }
-
   onMessageFromParent({data: action}) {
     switch (action.type) {
       case "SET_MESSAGE":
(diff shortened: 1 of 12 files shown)


test/unit/asrouter/ASRouter.test.js
```diff
@@ -130,6 +130,14 @@ describe("ASRouter", () => {
 
       assert.deepEqual(Router.state.messageBlockList, ["foo"]);
     });
+    it("should set state.messageBlockList to the block list in ASRouterPreferences", async () => {
```
k88hudson: Is there another test around sorting via targeting you could add?

content-src/asrouter/asrouter-content.jsx
```diff
@@ -144,10 +144,6 @@ export class ASRouterUISurface extends React.PureComponent {
     return () => ASRouterUtils.dismissBundle(bundle);
   }
 
-  triggerOnboarding() {
```
k88hudson: I'm not seeing the regular onboarding experience show up with both AMO/FXA. I don't think that's intended?

lib/ActivityStream.jsm
```diff
@@ -207,6 +207,7 @@ const PREFS_CONFIG = new Map([
       type: "local",
       localProvider: "OnboardingMessageProvider",
       enabled: true,
+      exclude: ["RETURN_TO_AMO_1"],
```
k88hudson: I like this, let's document this or add some tests somewhere so this is easy to understand/use.

k88hudson:
Ok, did a final round of testing and everything is working nicely. Thanks!

PR #4624, opened 2018-12-20
Layout and stories feed

dmose:
The initial hacking for https://docs.google.com/document/d/1vt1PzDCFL3GTuZuxHzBgcWpZVOHleRplo1pyzitarPs/edit

PR #4629, opened 2018-12-21
Bug 1516008 - Add pocket layout code behind its own separate pref

k88hudson:
Builds on scott's initial patch, but separates stuff out. Not completely done but some good progress. Uses nick's suggested naming convention `DiscoveryStream`.  Based on some of dmose's work in #4624.

### Testing instructions:
1. Ensure the new tab is the regular version by default.
2. In `about:config`, change  `browser.newtabpage.activity-stream.discoverystream.config` to be `{"enabled":true,"layout_endpoint":"https://gist.githubusercontent.com/ScottDowne/164995d9535b4203846048bdee29d169/raw/05804e9e1a9a209b3676509d9752fc8596a7d30c/spoc.json"}`
3. Reset the pref, ensure everything is back to normal.


content-src/components/Base/Base.jsx
```diff
@@ -163,7 +165,7 @@ export class BaseContent extends React.PureComponent {
                   <ManualMigration />
                 </div>
                 }
-              <Sections />
+              {isDiscoveryStream ? <DiscoveryStreamBase /> : <Sections />}
```
Mardak: wolasi was asking about the ManualMigration section, and nchapman suggested everything including search box and migration should be controllable but if it simplifies MVP, it can be removed. We should probably note somewhere… in trello? that we don't or have a followup bug to allow controlling.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -0,0 +1,24 @@
+import {connect} from "react-redux";
+import React from "react";
+
+export class _DiscoveryStreamBase extends React.PureComponent {
+  render() {
+    return (
+      <div className="discovery-stream">
+        <div className="layout">
+          {this.props.DiscoveryStream.layout.map((section, sectionIndex) => (
```
Mardak: Just to make sure, the old behavior rendered old layout for `[]` but here it'll end up rendering empty div.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -0,0 +1,148 @@
...
+
+  get config() {
+    if (this._prefCache.config) {
+      return this._prefCache.config;
+    }
+    try {
+      this._prefCache.config = JSON.parse(Services.prefs.getStringPref(CONFIG_PREF_NAME, ""));
+    } catch (e) {
+      // istanbul ignore next
+      Cu.reportError(`Could not parse preference. Try resetting ${CONFIG_PREF_NAME} in about:config.`);
```
Mardak: Should this at least set `_prefCache.config = {}`? Otherwise `.config` could be undefined and result in brokeness elsewhere `.config.enabled`

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -0,0 +1,24 @@
+import {connect} from "react-redux";
+import React from "react";
+
+export class _DiscoveryStreamBase extends React.PureComponent {
+  render() {
+    return (
+      <div className="discovery-stream">
+        <div className="layout">
```
Mardak: Are we expecting non-layout things inside discovery-stream? I suppose it might be useful to have an extra wrapper ?

content-src/components/Base/Base.jsx
k88hudson: sounds like a good follow-up to me

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
k88hudson: I could combine these

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
k88hudson: I will file a ticket to handle the empty state for this better

k88hudson: Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1516034

PR #4630, opened 2018-12-31
no bug - Stubbing out Discovery Stream components.

ScottDowne:
Updated the gist with a better set of data, closer matching the current API.

I also started to stub out the components we probably need. (without proper design, this is all I was willing to get into)

~~TopSites is a whole thing that I copied in, I left a githube comment in the pr explaining some of my reasoning, but it's not intended to be a proper solution, I might just rip it out and go with a placeholder instead.~~ Removed this, and went with a simple stub.

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -11,17 +11,17 @@ export class _DiscoveryStreamBase extends React.PureComponent {
   renderComponent(component) {
     switch (component.type) {
       case "TopSites":
-        return (<TopSites {...component.properties} />);
+        return (<TopSites />);
       case "SectionTitle":
-        return (<SectionTitle {...component.properties} />);
+        return (<SectionTitle />);
       case "CardGrid":
-        return (<CardGrid {...component.properties} />);
+        return (<CardGrid />);
       case "Hero":
-        return (<Hero {...component.properties} />);
+        return (<Hero />);
       case "HorizontalRule":
-        return (<HorizontalRule {...component.properties} />);
+        return (<HorizontalRule />);
       case "List":
-        return (<List {...component.properties} />);
+        return (<List />);
       default:
         return (<div>{component.type}</div>);
     }
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -1,15 +1,41 @@
+import {CardGrid} from "content-src/components/DiscoveryStreamComponents/CardGrid/CardGrid";
 import {connect} from "react-redux";
+import {Hero} from "content-src/components/DiscoveryStreamComponents/Hero/Hero";
+import {HorizontalRule} from "content-src/components/DiscoveryStreamComponents/HorizontalRule/HorizontalRule";
+import {List} from "content-src/components/DiscoveryStreamComponents/List/List";
 import React from "react";
+import {SectionTitle} from "content-src/components/DiscoveryStreamComponents/SectionTitle/SectionTitle";
+import {TopSites} from "content-src/components/DiscoveryStreamComponents/TopSites/TopSites";
 
 export class _DiscoveryStreamBase extends React.PureComponent {
+  renderComponent(component) {
+    switch (component.type) {
+      case "TopSites":
+        return (<TopSites {...component.properties} />);
+      case "SectionTitle":
+        return (<SectionTitle {...component.properties} />);
+      case "CardGrid":
+        return (<CardGrid {...component.properties} />);
+      case "Hero":
+        return (<Hero {...component.properties} />);
+      case "HorizontalRule":
+        return (<HorizontalRule {...component.properties} />);
+      case "List":
+        return (<List {...component.properties} />);
+      default:
+        return (<div>{component.type}</div>);
+    }
+  }
+
   render() {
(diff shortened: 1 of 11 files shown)


content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -1,15 +1,41 @@
...
+        return (<div>{component.type}</div>);
+    }
+  }
+
   render() {
+    const {layout} = this.props.DiscoveryStream;
     return (
       <div className="discovery-stream layout">
-        {this.props.DiscoveryStream.layout.map((section, sectionIndex) => (
-          <div key={`section-${sectionIndex}`} className={`column column-${section.width}`}>
```
ScottDowne: Seemed weird calling this section still.

content-src/components/DiscoveryStreamComponents/TopSites/SearchShortcutsForm.jsx
```diff
@@ -0,0 +1,141 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
+import {FormattedMessage} from "react-intl";
+import React from "react";
+import {TOP_SITES_SOURCE} from "./TopSitesConstants";
```
ScottDowne: I kinda just copied all of topsites over, I wouldn't expect this to be the final solution, but it did kinda just work.

I looked at the code a bit, and we'll probably need to decide on what to do with some of the features in the topsite components, like drag and drop, and what's the value/harm in keeping it. (the whole topsites section was more complex than I would have guessed)

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -1,15 +1,41 @@
...
+  renderComponent(component) {
+    switch (component.type) {
+      case "TopSites":
+        return (<TopSites {...component.properties} />);
+      case "SectionTitle":
+        return (<SectionTitle {...component.properties} />);
+      case "CardGrid":
+        return (<CardGrid {...component.properties} />);
+      case "Hero":
+        return (<Hero {...component.properties} />);
```
Mardak: I suppose this should be fine for now especially lacking designs. If we end up needing to copy/paste a bunch or there isn't that much difference in the react components, we probably could minimize bugs in similar-but-not-quite code with something like `<PocketCards type={component.type} {...components.properties} />`

content-src/components/DiscoveryStreamComponents/CardGrid/CardGrid.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export class CardGrid extends React.PureComponent {
+  render() {
+    return (
+      <div>
```
gvn: As we're scaffolding could we add class names to these components now too?

gvn:
This is working well! I also plugged in the [live endpoint](https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=dev-test-1) and it worked!

Just one suggestion, but this seems good to land to me.

lib/ActivityStream.jsm
```diff
@@ -219,7 +219,7 @@ const PREFS_CONFIG = new Map([
     title: "Configuration for the new pocket new tab",
     value: JSON.stringify({
       enabled: false,
-      // Set this to https://gist.githubusercontent.com/ScottDowne/164995d9535b4203846048bdee29d169/raw/0cf538411e6ee898eb116208d70842c62c8d52f1/spoc.json to test
+      // Set this to https://gist.githubusercontent.com/ScottDowne/3dc0a631a442e223aef9cede8b7f9d86/raw/833d0d44fd77b38c81cd00b36d866c576f45e734/layout.json to test
       layout_endpoint: "",
```
k88hudson: Do you want to maybe just land the actual endpoint here since this is off by default anyway?

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
k88hudson: slightly less code if you want:

```js
const ComponentTypes = {TopSites, CardGrid, Hero};
...
// in renderComponent
const ItemComponent = ComponentTypes[component.type];
return (<ItemComponent {...component.properties} />
```

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -1,15 +1,41 @@
...
       <div className="discovery-stream layout">
-        {this.props.DiscoveryStream.layout.map((section, sectionIndex) => (
-          <div key={`section-${sectionIndex}`} className={`column column-${section.width}`}>
-            {section.components.map((component, componentIndex) => (
+        {layout.map((row, rowIndex) => (
+          <div key={`row-${rowIndex}`} className={`column column-${row.width}`}>
+            {row.components.map((component, componentIndex) => (
               <div key={`component-${componentIndex}`}>
-                <div>{component.type}</div>
+                {this.renderComponent(component)}
```
k88hudson: This is a security vulnerability if you don't whitelist these props

lib/ActivityStream.jsm
gvn: Yeah, it's safe to put in `https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=dev-test-1` instead.

ScottDowne: Sure

PR #4631, opened 2019-01-02
Fix Bug 1517327 - Search telemetry failures not caught by taskcluster

Mardak:
Directly fixing test failures (turning off `handoffToAwesomebar` pref in `browser.ini`s) as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1516272 export, so probably don't want to merge this yet…


Mardak:
Rebased on TASKCLUSTER_ROOT_URL and now that central has the browser.ini changes. All looks green!

PR #4633, opened 2019-01-04
Fix Bug 1517342 - Taskcluster fails because of missing env var

piatra:


piatra:
I assume the bug that will have the workers set this automatically for us is being tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1508383 I am following the bug and will try to revert this PR at a later time.

PR #4636, opened 2019-01-04
Bug 1517596 - Add basic discoverystream devtools

k88hudson:
This adds basic devtools for discoverystream. Visit about:newtab#devtools-ds

Please ensure cache timing is accurate. I should also probably add tests for that time formatting function


content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -433,9 +492,14 @@ export class ASRouterAdminInner extends React.PureComponent {
           <h2>Pocket</h2>
           {this.renderPocketStories()}
         </React.Fragment>);
+      case "ds":
+        return (<React.Fragment>
+          <h2>Discovery Stream</h2>
+          <DiscoveryStreamAdmin state={this.props.DiscoveryStream} dispatch={this.props.dispatch} />
+          </React.Fragment>);
```
ScottDowne: Should this be indented down by two spaces?

content-src/components/Base/Base.jsx
```diff
@@ -86,8 +88,10 @@ export class _Base extends React.PureComponent {
       if (window.location.hash.startsWith("#asrouter") ||
           window.location.hash.startsWith("#devtools")) {
         return (<ASRouterAdmin />);
+      } else if (!didLogDevtoolsHelpText) {
```
ScottDowne: I don't know if you need this to be a `else if`, just a new `if`?

ScottDowne: I guess all the `else if` means is if you enable it while already on the devtools url, you don't get a log? That probably doesn't matter much, if I'm reading this right?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -107,7 +112,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   async disable() {
     await this.clearCache();
     // Reset reducer
-    this.store.dispatch(ac.BroadcastToContent({type: at.DISCOVERY_STREAM_LAYOUT_UPDATE, data: []}));
+    this.store.dispatch(ac.BroadcastToContent({type: at.DISCOVERY_STREAM_LAYOUT_RESET}));
```
ScottDowne: This is cool, because it can set it to whatever is in initial_state instead of making assumptions here as to what is considered initial.

content-src/components/Base/Base.jsx
ScottDowne: I get this now, makes sense. If you're one one of the above urls, logging how to find those urls would be pointless.

lib/DiscoveryStreamFeed.jsm
k88hudson: yeah I was thinking this would be a bit safer in case we change things

PR #4637, opened 2019-01-04
Fixes bug 1518171 - Adding feeds.

ScottDowne:


test/unit/lib/DiscoveryStreamFeed.test.js
```diff
@@ -89,6 +89,77 @@ describe("DiscoveryStreamFeed", () => {
...
+      sandbox.stub(feed.cache, "get").returns(Promise.resolve(fakeCache));
+      clock.tick(THIRTY_MINUTES + 1);
+
+      const feedResp = await feed.getComponentFeed("foo.com");
+
+      assert.equal(feedResp.data, "data");
+    });
+  });
+
   describe("#clearCache", () => {
```
k88hudson: You actually need to add a line to `clearCache` to acccount for this new property (`"feeds"`), please add a test as well

lib/DiscoveryStreamFeed.jsm
```diff
@@ -104,8 +105,65 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
+  async getComponentFeeds() {
```
k88hudson: can you add a comment about what this does

k88hudson: also probably rename to make it clearer what's going on

PR #4642, opened 2019-01-08
Bug 1515078 - Add opt out UX in about:preferences for Pocket Newtab

piatra:
In order to test this 
* set `browser.newtabpage.activity-stream.discoverystream.config` to have `enable` set to true
* navigate to `about:preferences#home`

### DIFF
--- a/lib/AboutPreferences.jsm
+++ b/lib/AboutPreferences.jsm
@@ -100,7 +100,7 @@ this.AboutPreferences = class AboutPreferences {
 
   async observe(window) {
     this.renderPreferences(window, await this.strings, [...PREFS_BEFORE_SECTIONS,
-      ...this.store.getState().Sections, ...PREFS_AFTER_SECTIONS], this.store.getState().Prefs);
+      ...this.store.getState().Sections, ...PREFS_AFTER_SECTIONS], this.store.getState().DiscoveryStream.config.enabled);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/AboutPreferences.jsm
```diff
@@ -257,6 +260,36 @@ this.AboutPreferences = class AboutPreferences {
       });
     });
 
+    try {
+      const discoveryStreamPref = JSON.parse(ASPreferences.values["discoverystream.config"]);
+      if (discoveryStreamPref.enabled) {
```
Mardak: Seems like we should probably avoid parsing the pref here. @ScottDowne would it be safe to check `state.DiscoveryStream.layout` to be non-0 length? Or even DiscoveryStream.config.enabled would avoid re-parsing json. ??

lib/AboutPreferences.jsm
```diff
@@ -257,6 +260,36 @@ this.AboutPreferences = class AboutPreferences {
...
+        // If Discovery Stream is enabled hide Home Content options
+        contentsGroup.style.visibility = "hidden";
+
+        const discoveryGroup = homeGroup.insertAdjacentElement("afterend", homeGroup.cloneNode());
+        discoveryGroup.id = "discoveryContentsGroup";
+        discoveryGroup.setAttribute("data-subcategory", "contents");
```
Mardak: Probably make the subcategory "discovery". As this is now, it prevents `about:preferences#home-contents` from working when it's not enabled (as it tries to highlight a hidden section).

lib/AboutPreferences.jsm
ScottDowne: `DiscoveryStream.config.enabled` would probably be ideal. I dunno if checking the length might be weird.

piatra:
@Mardak this will require a follow up m-c patch to handle the Restore Defaults button. We still want to restore about:home/about:newtab changes but not other preferences (for Sections) that might have changed. We want the button to be hidden if only the hidden preferences have changes.

This is what I had in mind https://gist.github.com/piatra/d290a031c7aae836dfe9b1e532f79721
What do you think?

PR #4643, opened 2019-01-08
Fix Bug 1517737 - handle keyboard events on fake searchbox when focused

rlr:
To test this, tab your way to the in-content search box and start typing.

This approach of switching focus on the keydown event is working so well (at least for me locally), that I wonder if we really need to do the fake focus trick. But I'd like some QA before changing any of that.

r?


test/unit/content-src/components/Search.test.jsx
```diff
@@ -102,5 +102,33 @@ describe("<Search>", () => {
       });
       assert.calledWith(dispatch, {type: "FOCUS_SEARCH"});
     });
+    it("should hand-off search when user types", () => {
+      const dispatch = sinon.spy();
+      const wrapper = shallowWithIntl(<Search {...DEFAULT_PROPS} handoffEnabled={true} dispatch={dispatch} />);
+      wrapper.instance().onSearchHandoffKeyDown({key: "f"});
```
piatra: Could you use [`simulate("keydown", "f")`](https://airbnb.io/enzyme/docs/api/ShallowWrapper/simulate.html) so it's closer to the real user action.

rlr: ah, good point 👍

PR #4644, opened 2019-01-08
Hero Component (WIP)

gvn:


content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -17,7 +17,10 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       case "CardGrid":
         return (<CardGrid />);
       case "Hero":
-        return (<Hero />);
+        return (<Hero
+          feed={component.feed.url}
+          style={component.properties.style}
```
dmose: Is this intended to be on every component as the way to override the local style from the server?  Or is it something else?

gvn: I'm basing this off [this specification](https://docs.google.com/document/d/1RCVxPGY2EbOr3tG19MIpfAaA59ZOApt303XUI0-RMoM/edit#), which is what the [endpoint schema](https://app.swaggerhub.com/apis/PocketNewTab/FirefoxHome/0.2) is derived from. 

In this case it's just controlling whether or not the component has a border.

gvn: PS: There's a lot more coming in this patch. Not ready for review yet, but questions are totally fine! :)

content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -2,14 +2,37 @@ import {connect} from "react-redux";
 import React from "react";
 
 export class _Hero extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    const feed = this.props.DiscoveryStream.feeds[this.props.feed.url];
+
+    this.state = {
+      recommendations: feed.data.recommendations
```
ScottDowne: I dunno what happens if you put something stored in redux into a component's state. Example, if the redux store changes, I don't know what happens to the data in the component's state. Might work, dunno.

gvn: What's the alternative?

Mardak: Typically just read the data from `props`

gvn:
@ScottDowne you can go ahead and review so we can get a rough version landed and do a follow-on ticket for the remaining work.

gvn:
@ScottDowne [This is the faked endpoint](https://gist.githubusercontent.com/gvn/76af39c92085c247dd5dd7ccc2412618/raw/0366a32853b35244807d9b90b1becb61157ccef9/disco.json) I used for testing the 3 layout widths.

gvn:
Stuff remaining to do:

- fix image aspect ratios
- wire up links
- generally tighten up styling
- set count of cards based on remote layout definition

content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -1,15 +1,51 @@
 import {connect} from "react-redux";
+import {DSCard} from "../DSCard/DSCard.jsx";
```
ScottDowne: It looks like most components only use relative links if it's `./` and it also looks like the .jsx is always omitted.

I don't personally have an opinion, just noticing this.

PR #4646, opened 2019-01-08
Bug 1518197 - handle paste events in search hand-off

rlr:
This depends on pull #4643. That should be reviewed first, so this can be rebased to not include that commit.

Before:
screen shot 2019-01-08 at 5 52 18 pm

After:
screen shot 2019-01-08 at 5 51 24 pm

Also, tabbing to the in-content search and pasting via keyboard works now.


rlr:
This is ready for r?

Whoever gets to it first wins!

piatra:
I'm not able to get tabbing + paste to work (click and paste definitely works).
There is also some inconsistent behavior between click and tabbing into the element.
* if you clicked into the element: typing the CMD key (without any character) will dismiss the in content search and focus the address bar
* if you tab into the element: typing the CMD key won't do anything.

content-src/components/Search/Search.jsx
```diff
@@ -96,6 +131,7 @@ export class _Search extends React.PureComponent {
       this.props.hide && "search-hidden",
       this.props.focus && "search-active",
     ].filter(v => v).join(" ");
+    const onPaste = this.props.handoffEnabled ? this.onSearchHandoffPaste : null;
```
rlr: dont need this anymore. removing

content-src/components/Search/Search.jsx
```diff
@@ -53,7 +56,34 @@ export class _Search extends React.PureComponent {
...
+    // We need to listen to paste events that bubble up from the Search Hand-off
+    // button. Adding the paste listener to the button itself or it's parent
+    // doesn't work consistently until the page is clicked on.
+    if (this._searchHandoffButton === global.document.activeElement) {
+      this.onSearchHandoffPaste(event);
+    }
+  }
+
+  componentWillMount() {
+    if (global.document) {
```
rlr: this is so prerender doesnt blow up

PR #4649, opened 2019-01-10
Fixes bug 1518921 - Discovery Stream Spocs

ScottDowne:


common/Reducers.jsm
```diff
@@ -457,6 +462,26 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
       return {...prevState, lastUpdated: INITIAL_STATE.DiscoveryStream.lastUpdated, layout: INITIAL_STATE.DiscoveryStream.layout};
     case at.DISCOVERY_STREAM_FEEDS_UPDATE:
       return {...prevState, feeds: action.data || prevState.feeds};
+    case at.DISCOVERY_STREAM_SPOCS_ENDPOINT:
+      return {
+        ...prevState,
+        spocs: {
+          spocs_endpoint: action.data || prevState.spocs.spocs_endpoint,
```
k88hudson: I guess there is intentionally no `lastUpdated` because you want to force the cache to clear?

common/Reducers.jsm
```diff
@@ -457,6 +462,26 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
...
+          data: prevState.spocs.data,
+        },
+      };
+    case at.DISCOVERY_STREAM_SPOCS_UPDATE:
+      if (action.data) {
+        return {
+          ...prevState,
+          spocs: {
```
k88hudson: why not just 
```js
spocs: {
  ...prevState.spocs,
  lastUpdated: action.data.lastUpdated,
  data: action.data.spocs,
}
```

common/Reducers.jsm
```diff
@@ -457,6 +462,26 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
       return {...prevState, lastUpdated: INITIAL_STATE.DiscoveryStream.lastUpdated, layout: INITIAL_STATE.DiscoveryStream.layout};
     case at.DISCOVERY_STREAM_FEEDS_UPDATE:
       return {...prevState, feeds: action.data || prevState.feeds};
+    case at.DISCOVERY_STREAM_SPOCS_ENDPOINT:
+      return {
+        ...prevState,
+        spocs: {
+          spocs_endpoint: action.data || prevState.spocs.spocs_endpoint,
+          data: prevState.spocs.data,
```
k88hudson: Also why would you leave the data for the previous spocs if the endpoint changed?

k88hudson: Can you describe the behaviour for this in a comment actually since it's kinda complex

common/Reducers.jsm
k88hudson: Also why use the previous if none is defined? what if there is intentionally no spocs

PR #4650, opened 2019-01-11
Bug 1520449 - Do the search hand-off on keydown instead of mouse click

rlr:
@Mardak @piatra @k88hudson Thoughts on this simplification? 😄 

Since we have to handing off on key stroke for users navigating the page with the keyboard, may as well do that for mouse users too instead of doing the hidden focus trickery?

There is more fresh code that can be removed from the m-c side too (`gUrlBar.hiddenFocus()` and related styles).

f? Should I file a bug and land this?


content-src/components/Search/Search.jsx
```diff
@@ -31,25 +31,28 @@ export class _Search extends React.PureComponent {
...
+      this._searchHandoffButton.focus();
+    }
   }
 
   onSearchHandoffKeyDown(event) {
     if (event.key.length === 1 && !event.altKey && !event.ctrlKey && !event.metaKey) {
       // We only care about key strokes that will produce a character.
       const text = event.key;
       this.props.dispatch(ac.OnlyToMain({type: at.HANDOFF_SEARCH_TO_AWESOMEBAR, data: {text}}));
+      this.props.dispatch({type: at.HIDE_SEARCH});
```
piatra: We need to do this for paste events as well.

rlr:
The more I think and look at this, the more I like it. I think it's the way to go, but I want to make sure I'm not missing some reason we shouldn't do it. @k88hudson ?

rlr:
Filed Bugzilla bug for this https://bugzilla.mozilla.org/show_bug.cgi?id=1520449

k88hudson:
This seems cool! Let me try it out manually

k88hudson:
Seems good to me. Thanks!

PR #4652, opened 2019-01-11
bug 1519484 - Working on css overrides on topsites.

ScottDowne:


ScottDowne:
This did not change how the existing layout works, example, I didn't add css grid or percentages for widths, I kept it as close to the old topsites layout to reduce changes.

ScottDowne:
To test I have been using this feed, which puts a topsites in each column: https://gist.githubusercontent.com/ScottDowne/e90dd44a9f7bddaef8b4f2895221f58d/raw/523b05a06d57ad1ca726ec60596af64c8eb6188a/test

ScottDowne:
Yeah, so css grid is calculating where to put the gap differently than the spec did. I have a fiddle demonstrating this in a reduced case. If you inspect the grid in chrome it really shows the break down.

https://jsfiddle.net/2ent4vu0/

The widths of 592 and 296 in the comps should actually be 608 and 280.

There are 12 columns, with a 48 pixel gap. That creates a  single row of 936px with 11 potential gaps of 48. This is a total space used for gaps alone of 528px. This leaves us with 408px of actual non gap space. 408 / 12 is 34 which is the space of a single column with. So a 4 wide column is 4 34 and 3 48, or (34 * 4) + (48 * 3) = 280 and (34 * 8) + (48 * 7) = 608.

content-src/components/DiscoveryStreamComponents/TopSites/_TopSites.scss
```diff
@@ -10,5 +10,77 @@
...
+    .top-site-outer {
+      padding: 0 12px;
+    }
+
+    .top-sites-list {
+      margin: 0 -12px;
+    }
+  }
+
+  // Only show 6 cards for 2/3 and 1/3
```
dmose: Maybe include an XXX comment here that someday the old TopSites should get an explicit prop allowing .hide-for-narrow to be passed in rather than having it render stuff and then make it display: none.

content-src/components/DiscoveryStreamComponents/TopSites/_TopSites.scss
```diff
@@ -10,5 +10,77 @@
...
+      padding: 0 10px;
+    }
+
+    .top-sites-list {
+      margin: 0 -10px;
+    }
+
+    .top-site-inner {
+      .tile {
+        width: 84.67px;
```
dmose: These numbers are unusual enough that it probably warrants a comment as to where they come from, in case anyone needs to change it in the future.

content-src/components/DiscoveryStreamComponents/TopSites/_TopSites.scss
```diff
@@ -10,5 +10,77 @@
...
+  }
+
+  // Only show 6 cards for 2/3 and 1/3
+  .hide-for-narrow {
+    display: none;
+  }
+}
+
+// Only show 8 cards for the full row.
+.ds-column-12 {
```
gvn: Can you add the other nearby column widths for future compatibility?

EG:

```css
.ds-column-9 &,
.ds-column-10 &,
.ds-column-11 &,
.ds-column-12 & {
```

piatra:
TC is complaining

image

ScottDowne:
@gvn @piatra should be all fixed up now.

ScottDowne:
Quick chat with @dmose on this and he said his r+ still stands and we can land this now.

PR #4654, opened 2019-01-14
Bug 1519766 - standup list component for DiscoveryStream

dmose:
This stands up a basic list component for DiscoveryStream.

How to test:
* go to about:config
* search for stream.config
* set that preference to {"enabled":true,"layout_endpoint":"https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=control"}
* open a new tab

The list is the "Recommended by Pocket" is the thing to look at.  

Known bugs are already captured as dependents of the ds-list-component bug: https://bugzilla.mozilla.org/showdependencytree.cgi?id=1514020&hide_resolved=0

@ScottDowne A review would be great.  I'll wait for you to land bug 1519484 before I land this so that TopSites doesn't have a weird intermediate state due to me adding the 936px width constraint here.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/List/List.jsx
+++ b/content-src/components/DiscoveryStreamComponents/List/List.jsx
@@ -1,15 +1,60 @@
 import {connect} from "react-redux";
 import React from "react";
 
-export class _List extends React.PureComponent {
-  render() {
-    // const feed = this.props.DiscoveryStream.feeds[this.props.feed.url];
-    return (
-      <div className="ds-list">
-        List
-      </div>
-    );
-  }
+/**
+ * @note exported for testing only
+ */
+export function ListItem(props) {
+  // TODO performance: get feeds to send appropriately sized images rather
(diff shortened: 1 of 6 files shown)


content-src/components/DiscoveryStreamComponents/List/List.jsx
```diff
@@ -1,15 +1,65 @@
...
+    // XXX
+    /* eslint-disable-next-line no-console */
+    console.log(feed);
   }
+
+  if (!recs) {
+    return null;
+  }
+
+  // XXX does rec.id meet necessary constraints for `key`?
```
ScottDowne: Should be fine. If it doesn't it probably means we have dupes in the data which would be more bad than a duplicate id.

ScottDowne: If you're worried, sometimes I grab the index as well and do `${id}-${index}`

PR #4656, opened 2019-01-15
Bug 1519396 - 'Force Attribution' from ASRouter for Return to AMO doe…

piatra:
…s not work on Windows

Inspired by a [windows only xpcshell](https://searchfox.org/mozilla-central/rev/b29663c6c9c61b0bf29e8add490cbd6bad293a67/browser/components/attribution/test/xpcshell/test_AttributionCode.js#61-70)

Landing this PR would unblock QA on windows for RTAMO which was problematic so far including the [patch that landed in order to disable the feature in beta and release](https://bugzilla.mozilla.org/show_bug.cgi?id=1515411#c15)


lib/ASRouter.jsm
```diff
@@ -948,6 +950,20 @@ class _ASRouter {
     }
   }
 
+  // Windows specific calls to write attribution data
```
k88hudson: Can you add more comments about how this is only used for QA/testing basically

piatra:
> Looks good, thanks! Any manual test plan needed here?

This would have been needed for bug 1515411. Same test steps apply as currently listed in the [User Story](https://bugzilla.mozilla.org/show_bug.cgi?id=1515411)

PR #4658, opened 2019-01-15
Starting a spoc rendering function.

ScottDowne:


lib/DiscoveryStreamFeed.jsm
```diff
@@ -160,15 +160,14 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
-        await this.cache.set("spocs", spocs);
-      } else {
+      if (!spocsResponse) {
         Cu.reportError("No response for spocs_endpoint prop");
       }
+      spocs = {
+        lastUpdated: Date.now(),
+        data: spocsResponse || [],
+      };
+      await this.cache.set("spocs", spocs);
```
ScottDowne: The changes above are what ensure a loaded state always gets added in the case of an error.

k88hudson: seems good to me

PR #4661, opened 2019-01-15
Bug 1514022 – Card Grid component (WIP)

gvn:
Endpoint for testing:

https://gist.githubusercontent.com/gvn/76af39c92085c247dd5dd7ccc2412618/raw/843f2d53f03993fa7a1ea267140619ecd96b5b66/disco.json


gvn:
There's a fair amount of tweaks left, but would be good to land this to get a basic version available and also get the line count mixin in the codebase.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -8,9 +8,9 @@ export class DSCard extends React.PureComponent {
           <div className="img" style={{backgroundImage: `url(${this.props.image_src}`}} />
         </div>
         <div className="meta">
-          <header>{this.props.title}</header>
-          <p>{this.props.excerpt}</p>
-          <p>{this.props.source}</p>
+          <header className="title">{this.props.title}</header>
```
ScottDowne: This is probably fine, but there is another component that uses the classname title (topsites) but I think in both cases and css styling has proper specificity in the css as to not leak. In your case it's `.ds-card .meta .title`

PR #4665, opened 2019-01-16
NO BUG – using max-width instead of fixed width for primary content area

gvn:


gvn:
screen shot 2019-01-16 at 09 33 25-fullpage

We'll need to refactor TopSites a bit to use percentages instead of fixed pixel values so it looks good under 936px but this feels a lot better.

gvn:
Q from @k88hudson : Does this affect drag and drop?

k88hudson:
I don't see any major issues – I think writing things with having to support responsive design in mind makes sense, however, I think we we should make a decision with on whether we're going to fix the container size or not for the experiment so we don't get half-broken / half-supported states.

ScottDowne:
I think the issue with topsites is that I don't know if it's trivial or not. It might be trivial though.

I think once the components are updated, we can take this patch, and from there enforce new components to be responsive.

gvn:
Ok, let's pause on this then. I'll leave it here for now.

PR #4672, opened 2019-01-16
Bug 1516034 - Use non-prerendered version of HTML for DiscoveryStream New Tab

punamdahiya:


punamdahiya:
@k88hudson  Patch updates PrerenderData to handle prefs with value as data in json format. If enabled property
differs from initialPrefs enabled value of browser.newtabpage.activity-stream.discoverystream.config, prerendering is not used. Please review. Thanks!

test/unit/common/PrerenderData.test.js
```diff
@@ -94,6 +94,33 @@ describe("_PrerenderData", () => {
...
+      assert.isFalse(instance.arePrefsValid(getPrefs));
+    });
+    it("should return true if all of jsonPrefs group enabled value matches", () => {
+      FAKE_PREFS = {foo: true, bar: {"enabled": true}, baz: {"enabled": false}};
+      const instance = new _PrerenderData({
+        initialPrefs: {foo: true, bar: {"enabled": true}, baz: {"enabled": false}},
+        validation: ["foo", "bar", {jsonPrefs: ["baz", "bar"]}],
+      });
+
+      assert.isFalse(instance.arePrefsValid(getPrefs));
```
punamdahiya: just realised, check here is wrong and should fix the broken test

k88hudson:
@punamdahiya also could we rename this bug to reflect the actual work being done here, i.e. always use non-prerendered version of HTML for DiscoveryStream new tab

punamdahiya:
> This looks good to me! The only thing I noticed is that it's kind of hard to figure out if the prerendered version is being rendered or not. What do you think about adding something (like a comment even) to the HTML `render-activity-stream-html` so QA could verify this more easily?


We can add a comment "“prerendered html not shown”) in web console  for QA to verify when discovery stream experience is enabled. We can log it at either of below two places

https://github.com/mozilla/activity-stream/blob/master/content-src/components/Base/Base.jsx#L98
or
https://github.com/mozilla/activity-stream/blob/master/content-src/activity-stream.jsx#L23
 Should that suffice?

content-src/activity-stream.jsx
```diff
@@ -19,8 +19,10 @@ new DetectUserSessionStart(store).sendEventOrAddListener();
 // to request state rehydration (see Base.jsx). If we are NOT in a prerendered state,
 // we can request it immedately.
 if (!global.gActivityStreamPrerenderedState) {
+  console.log("Prerendered html not used for Discovery Stream New Tab");
```
punamdahiya: Added comment to verify in console for non-prerendered version for discovery stream new tab

test/unit/common/PrerenderData.test.js
punamdahiya: check condition updated

bin/render-activity-stream-html.js
```diff
@@ -93,6 +93,8 @@ for (const src of ${JSON.stringify(scripts, null, 2)}) {
   script.src = src;
 }
     </script>`;
+  // Hidden tag to check if prerendered version of html is used. See Bug 1516034
+  const isPrerenderedStatusTag = "<span hidden>Prerendered version html not shown</span>";
```
k88hudson: Would you mind just making this a comment? I think it would be just as easy to inspect and no extra element

punamdahiya: +1, done

bin/render-activity-stream-html.js
```diff
@@ -108,6 +110,7 @@ for (const src of ${JSON.stringify(scripts, null, 2)}) {
     <div id="snippets-container">
       <div id="snippets"></div>
     </div>${options.noscripts ? "" : scriptTag}
+    ${isPrerendered ? "" : isPrerenderedStatus}
```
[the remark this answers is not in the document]
punamdahiya: @Mardak Agree reusing check in root tag is much better, my initial attempt to reuse that space didn’t work probably due to some other issue because of which ended up putting comment in its own check. 

I have updated code https://github.com/mozilla/activity-stream/pull/4691 and verified the comment shows for discovery stream when we are not using pre-rendered version. Thanks

PR #4673, opened 2019-01-16
Bug 1520662: Ensure undeterministic load order is not causing crashes

k88hudson:
To test, ensure the page doesn't crash on multiple refreshes.

### DIFF
--- a/content-src/lib/selectLayoutRender.js
+++ b/content-src/lib/selectLayoutRender.js
@@ -4,7 +4,7 @@ function calculateSpocs(component, spocs) {
   let spocIndex = 0;
   return component.spocs.positions.map(position => {
     const rickRoll = Math.random();
-    if (rickRoll <= component.spocs.probability) {
+    if (spocs.data.spocs[spocIndex] && rickRoll <= component.spocs.probability) {
       return {
         ...position,
         result: spocs.data.spocs[spocIndex++],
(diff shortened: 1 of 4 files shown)


content-src/lib/selectLayoutRender.js
```diff
@@ -4,7 +4,7 @@ function calculateSpocs(component, spocs) {
   let spocIndex = 0;
   return component.spocs.positions.map(position => {
     const rickRoll = Math.random();
-    if (rickRoll <= component.spocs.probability) {
+    if (spocIndex < spocs.data.spocs.length && rickRoll <= component.spocs.probability) {
```
ScottDowne: `spocs.data.spocs[spocIndex]` maybe?

content-src/lib/selectLayoutRender.js
```diff
@@ -37,7 +37,7 @@ export const selectLayoutRender = createSelector(
         }
 
         // Calculate if we should display a spoc or not.
-        if (component.spocs) {
+        if (component.spocs && spocs.data && spocs.data.spocs && spocs.data.spocs.length) {
```
ScottDowne: I *think* we can remove the `spocs.data` part in the checks as that comes from reduxand *should* always be defaulted.

Ending up with just `if (component.spocs && spocs.data.spocs && spocs.data.spocs.length) {`

PR #4677, opened 2019-01-17
Bug 1515078 - Use PreferenceExperiments API to unenroll from Pocket Newtab experiment

piatra:
[`"individual-opt-out"`](https://searchfox.org/mozilla-central/rev/bee8cf15c901b9f4b0c074c9977da4bbebc506e3/toolkit/components/normandy/content/about-studies/about-studies.js#189) is the telemetry event name used when unenrolling from an experiment on the `about:studies` page. We could change that to something specific to `about:preferences` if we're interested in the data.

I was not familiar with how Normandy chooses users and toggles preferences. Specifically what happens if we roll out from 1% to 5% but before the second rollout some users voluntarily leave the experiment.
> because of the sampling characteristics of Normandy, if you increase from 1% to 5%, the additional 4% are guaranteed to be distinct from the original 1%

### DIFF
--- a/lib/AboutPreferences.jsm
+++ b/lib/AboutPreferences.jsm
@@ -286,14 +286,15 @@ this.AboutPreferences = class AboutPreferences {
         .addEventListener("click", async () => {
           const activeExperiments = await PreferenceExperiments.getAllActive();
           const experiment = activeExperiments.find(exp => exp.preferenceName === DISCOVERY_STREAM_CONFIG_PREF_NAME);
+          // Unconditionally update the UI for a fast user response and in
+          // order to help with testing
+          discoveryGroup.style.display = "none";
+          contentsGroup.style.visibility = "visible";
           if (experiment) {
             await PreferenceExperiments.stop(experiment.name, {
               resetValue: true,
               reason: "individual-opt-out",
             });
-
-            discoveryGroup.style.display = "none";
-            contentsGroup.style.visibility = "visible";
           }
         }, {once: true});
     }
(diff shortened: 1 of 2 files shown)


lib/AboutPreferences.jsm
```diff
@@ -279,10 +283,18 @@ this.AboutPreferences = class AboutPreferences {
...
       createAppend("hbox", discoveryGroup)
         .appendChild(contentDiscoveryButton)
-        .addEventListener("click", () => {
-          discoveryGroup.style.display = "none";
-          contentsGroup.style.visibility = "visible";
-          Services.prefs.clearUserPref(DISCOVERY_STREAM_CONFIG_PREF_NAME);
+        .addEventListener("click", async () => {
+          const activeExperiments = await PreferenceExperiments.getAllActive();
+          const experiment = activeExperiments.find(exp => exp.preferenceName === DISCOVERY_STREAM_CONFIG_PREF_NAME);
+          if (experiment) {
```
Mardak: Hmm.... this has an odd behavior of discovery stream can be enabled according to `DiscoveryStream.config.enabled` but the button only does something if the experiment is active.

If we condition showing opt-out based on an active experiment, it makes it hard to test by just flipping the pref. But depending on how long these async calls take, waiting on that to determine what to show might be problematic too.

Maybe we can just conditionally stop the experiment and unconditionally (and immediately before `await`ing) hide. If the tester manually changed the pref without being in the experiment, the pref will be unchanged, but at least can confirm the before/after UI states.

PR #4678, opened 2019-01-17
Bug 1520852 – Changing property name to match new API schema

gvn:


gvn:
To accomodate this change: https://www.diffchecker.com/AinjnWUu

PR #4680, opened 2019-01-17
bug 1520893 - Display spocs

ScottDowne:
To test:

1. set your browser.newtabpage.activity-stream.discoverystream.config pref to `{"enabled":true,"layout_endpoint":"https://gist.githubusercontent.com/ScottDowne/1be633daea7d1ce65a8ca903a3be9ae9/raw/3d51663327126879f6fa1c6b3e7730e7805c9e1e/spocs"}`
2. Load a new tab, take note of the 3rd card in both the card grid and hero grid.
3. Refresh a couple times, noting the 3rd card in each refresh.

Expected: The third position on each hero and card component has a 50% chance to be a spoc.

There are 4 combinations based on the rolls you get.

1. where both rolls fail, and in this case you see the same card in both positions in both components because they share the same feed.

2. where the card component rolls a spoc and the hero does not.

3. where hero has a spoc and card doesn't.

4. And finally where they both have spocs.

We need to ensure when they both display a spoc, it's not the same spoc.

Dunno the best way to determine which card is what because I'm using live feed data. At the moment I'm writing this the card is from "theatlantic.com" and the two available spocs are from "thepointsguy.com" and have different images.

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -39,6 +39,27 @@ export function isAllowedCSS(property, value) {
     url.slice(5).startsWith(prefix)));
 }
 
+function maybeInjectSpocs(data, spocs) {
+  if (!data || !spocs.positions || !spocs.positions.length) {
+    return data;
+  }
+
+  const recommendations = [...data.recommendations];
+
+  for (let position of spocs.positions) {
+    const {result} = position;
+    if (result) {
+      // Insert spoc into the desired index.
+      recommendations.splice(position.index, 0, result);
+    }
+  }
+
+  return {
+    ...data,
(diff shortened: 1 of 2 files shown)


content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -34,7 +34,21 @@ export class Hero extends React.PureComponent {
...
+    if (spocs.positions && spocs.positions.length) {
+      for (let position of spocs.positions) {
+        const {result} = position;
+        if (result) {
+          // Insert spoc into the desired index and pop the last item.
+          recs.splice(position.index, 0, result);
+          recs.pop();
+        }
+      }
+    }
```
gvn: Let's move this block up into `DiscoveryStreamBase.jsx` instead of putting copies of it in both components.

PR #4685, opened 2019-01-18
Bug 1519775 - Make one and two-column lists work

dmose:
Looking for a quick-turnaround review so I can get this (layout-breaking) bug in before the impending export and tidy up individual bits of the lists later.  Whoever wants this, feel free to take it!

How to test:

 * go to about:config
 * search for stream.config
 * set that preference to {"enabled":true,"layout_endpoint":"https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=dev-test-all"}
 * open a new tab

The various trending stories should all be appropriate widths.

Things to note:

* some horizontal rules above or below individual stories may be incorrectly there or incorrectly missing.  https://bugzilla.mozilla.org/show_bug.cgi?id=1521119 has been filed.

Known bugs are already captured as dependents of the ds-list-component bug: https://bugzilla.mozilla.org/showdependencytree.cgi?id=1514020&hide_resolved=0

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -161,11 +161,13 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       case "HorizontalRule":
         return (<HorizontalRule />);
       case "List":
-        rows = this.extractRows(component, MAX_ROWS_LIST);
+        rows = this.extractRows(component,
+          Math.min(component.properties.items, MAX_ROWS_LIST));
         return (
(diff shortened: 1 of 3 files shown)


piatra:
Is this correct? It looks like the extra 16px (592+16) from the 2 column right hand side `Trending Stories` should go to the 1 column `Trending stories` (296-16)
image
image
image

dmose:
I think you're quite possibly right, but I think it's most important to get the fix for this layout bustage into m-c, so I'd propose fixing that in a separate bug.

dmose:
https://bugzilla.mozilla.org/show_bug.cgi?id=1521128 filed on the space issue.

PR #4688, opened 2019-01-18
Bug 1519243 - Split title component into two, the new one is navigation

ScottDowne:
I've been testing with this: https://gist.githubusercontent.com/ScottDowne/84457bf09597afe67926cca558f20d22/raw/e42431c06c21867388468e8ab734b394c32f9a77/title-nav

changing out things as I need.

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -4,6 +4,7 @@ import {Hero} from "content-src/components/DiscoveryStreamComponents/Hero/Hero";
 import {HorizontalRule} from "content-src/components/DiscoveryStreamComponents/HorizontalRule/HorizontalRule";
 import {ImpressionStats} from "content-src/components/DiscoveryStreamImpressionStats/ImpressionStats";
 import {List} from "content-src/components/DiscoveryStreamComponents/List/List";
+import {Navigation} from "content-src/components/DiscoveryStreamComponents/Navigation/Navigation";
 import React from "react";
 import {SectionTitle} from "content-src/components/DiscoveryStreamComponents/SectionTitle/SectionTitle";
 import {selectLayoutRender} from "content-src/lib/selectLayoutRender";
@@ -130,7 +131,9 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       case "TopSites":
         return (<TopSites />);
       case "SectionTitle":
-        return (<SectionTitle />);
+        return (<SectionTitle header={component.header} />);
+      case "Navigation":
+        return (<Navigation links={component.properties.links} alignment={component.properties.alignment} />);
       case "CardGrid":
         rows = this.extractRows(component, MAX_ROWS_CARDGRID);
         return (
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -131,9 +131,16 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       case "TopSites":
         return (<TopSites />);
       case "SectionTitle":
-        return (<SectionTitle header={component.header} />);
+        return (
+          <SectionTitle
+            header={component.header} />
+        );
       case "Navigation":
-        return (<Navigation links={component.properties.links} alignment={component.properties.alignment} />);
+        return (
+          <Navigation
+            links={component.properties.links}
+            alignment={component.properties.alignment} />
+        );
       case "CardGrid":
         rows = this.extractRows(component, MAX_ROWS_CARDGRID);
         return (
--- a/content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
+++ b/content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
@@ -11,12 +11,10 @@ export class Navigation extends React.PureComponent {
   render() {
     const {links} = this.props || [];
     const {alignment} = this.props || "centered";
-    const moreRecs = links.pop();
     return (
       <span className={`ds-navigation ds-navigation-${alignment}`}>
         <ul>
           {links && links.map(t => <Topic key={t.name} url={t.url} name={t.name} />)}
-          {moreRecs ? <li><a href={moreRecs.url} className="ds-more-recommendations">{moreRecs.name}</a></li> : null }
         </ul>
       </span>
     );
--- a/content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
+++ b/content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
@@ -1,0 +1,24 @@
+import React from "react";
+
+export class Topic extends React.PureComponent {
+  render() {
+    const {url, name} = this.props;
+    return (<li><a key={name} href={url}>{name}</a></li>);
+  }
+}
+
+export class Navigation extends React.PureComponent {
+  render() {
+    const {links} = this.props || [];
+    const {alignment} = this.props || "centered";
+    const moreRecs = links.pop();
+    return (
+      <span className={`ds-navigation ds-navigation-${alignment}`}>
+        <ul>
+          {links && links.map(t => <Topic key={t.name} url={t.url} name={t.name} />)}
+          {moreRecs ? <li><a href={moreRecs.url} className="ds-more-recommendations">{moreRecs.name}</a></li> : null }
+        </ul>
+      </span>
+    );
+  }
+}
--- a/content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
+++ b/content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
@@ -39,23 +39,7 @@
       &:active,
       &:visited {
         color: $blue-70;
-
-        &.ds-more-recommendations::after {
-          fill: $blue-70;
-        }
       }
     }
-  }
-
-  .ds-more-recommendations::after {
-    background: url('../data/content/assets/topic-show-more-12.svg') no-repeat center center;
-    content: '';
-    -moz-context-properties: fill;
-    fill: var(--newtab-link-primary-color);
-    display: inline-block;
-    height: 16px;
-    margin-inline-start: 5px;
-    vertical-align: top;
-    width: 12px;
   }
 }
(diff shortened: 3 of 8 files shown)


content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
```diff
@@ -0,0 +1,44 @@
...
+      content: none;
+    }
+
+    a {
+      color: $blue-60;
+    }
+  }
+
+  .ds-more-recommendations::after {
+    background: url('../data/content/assets/topic-show-more-12.svg') no-repeat center center;
```
Mardak: The designs (at least for the 2/3 and 1/3 columns) don't show the arrow. Is this intended to be split out from this PR?

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
```diff
@@ -0,0 +1,44 @@
...
+      content: '•';
+      padding: 8px;
+    }
+
+    &:last-child::after {
+      content: none;
+    }
+
+    a {
+      color: $blue-60;
```
Mardak: The designs have hover and active styling that don't seem to be here. Also, the rest of the page had been using another level of indirection to avoid copying this `$blue-60` to each link-ish thing and instead referencing some "link-primary-color". I guess maybe we aren't doing that now but definitely helps for supporting multiple themes.

content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,24 @@
...
+export class Navigation extends React.PureComponent {
+  render() {
+    const {links} = this.props || [];
+    const {alignment} = this.props || "centered";
+    const moreRecs = links.pop();
+    return (
+      <span className={`ds-navigation ds-navigation-${alignment}`}>
+        <ul>
+          {links && links.map(t => <Topic key={t.name} url={t.url} name={t.name} />)}
+          {moreRecs ? <li><a href={moreRecs.url} className="ds-more-recommendations">{moreRecs.name}</a></li> : null }
```
Mardak: Seems like the last item is still a `Topic` so maybe have that component handle "more" or "last" prop.

content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+}
+
+export class Navigation extends React.PureComponent {
+  render() {
+    const {links} = this.props || [];
+    const {alignment} = this.props || "centered";
+    const moreRecs = links.pop();
+    return (
+      <span className={`ds-navigation ds-navigation-${alignment}`}>
```
Mardak: Any concerns here about concatenating server provided string?

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  .ds-more-recommendations::after {
```
Mardak: Is it expected to use the `&::after` related declarations for this too?

content-src/components/DiscoveryStreamComponents/SectionTitle/_SectionTitle.scss
```diff
@@ -1,30 +1,15 @@
...
-      padding: 8px;
-    }
+  text-align: center;
 
-    &:last-child::after {
-      content: none;
-    }
+  .title {
+    line-height: 48px;
+    font-size: 36px;
```
Mardak: Do these need different sizes for other column sizes?

content-src/components/DiscoveryStreamComponents/SectionTitle/SectionTitle.jsx
```diff
@@ -1,22 +1,13 @@
...
     return (
-      <span className="ds-section-title">
-        <ul>
-          {topics && topics.map(t => <Topic key={t.name} url={t.url} name={t.name} />)}
-          <li><a className="ds-more-recommendations">More Recommendations</a></li>
-        </ul>
-      </span>
+      <div className="ds-section-title">
+        <div className="title">{title}</div>
+        {subtitle ? <div className="subtitle">{subtitle}</div> : null}
```
Mardak: Is it expected that section headers reuse this component or each will have their own styling? A quick glance I see Hero using ds-header and List using ds-list-title and neither support subtitles… The server seems to have a Header schema anyway… ?

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
ScottDowne: Yeah I didn't notice that.

To me it actually looks like the more recommendation link isn't displayed in the 2/3 and 1/3 views at all, which the arrow is implicitly tied to and only displayed in the last item. So no more recs link, no more arrows. The more recs link is also not explicitly defined, it's just the last item in the array that the server returns. I treat all of these as links the same and I don't have a real sense as to which is which. So I think the best I can do is always hide the last item in the 1/3 and 2/3 cases, and get assume that the server always has a more recs link as the last item. If for some reason the server stops sending back a more recs link as the last item, we'll start unintentionally hiding the last item in 1/2 and 2/3.

ScottDowne: Other options are, the server explicitly returns a prop for more recs link instead of having it in the end of the array, or we somehow return the array in the string.

content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
ScottDowne: Last item is more recs, but handled like a topic as implemented by the server.

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
ScottDowne: Not sure I'm totally following.

content-src/components/DiscoveryStreamComponents/SectionTitle/_SectionTitle.scss
ScottDowne: Technically the comps only have this defined with 1 width. I don't think it's their intent to use the title in the 1/3 and 2/3 views, but if they decide to, yeah, adding a size update might be worth it, otherwise, it'll still do something not terrible.

content-src/components/DiscoveryStreamComponents/SectionTitle/SectionTitle.jsx
ScottDowne: Not sure, they don't seem similar enough to me visually, but I cannot comment on list vs hero.

However, if there are slight differences, and keeping the components completely de coupled seems to make some sense to me as long as they are simple in complexity.

I think the only real value in reusability at the moment is in the ds-card for card grid and hero.

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
ScottDowne: I think for now I'll just hide the last item in the 1/3 and 2/3 while we discus this with the server. Pending there reaction, we may or may not open a new ticket to make this more explicit and less prone to error.

content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
ScottDowne: Not that I'm aware of. I don't *think* a server defined variable can escape the back quote.

All it can do is define some pretty funky css classnames, right?

For what it's worth we're doing this in other places too, so if it is an issue we should probably flag it for multiple spots in other components too.

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
ScottDowne: I'm actually seeing in The example comps there are 1/3 and 2/3 navigation elements with 1 item and an arrow, so I think we shouldn't do anything here.

ScottDowne:
@Mardak updates and comments.

Mardak:
Server seems to be adding the arrow into the text, so there shouldn't need to be any special handling of the last item anymore. ??

screen shot 2019-01-22 at 9 03 01 am

Mardak:
Just documenting from wolasi: `yes. using text for the arrows is fine. using text gives us more flexibility`

ScottDowne:
Updated

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
Mardak: Multiple selectors would have applied styles to the "more" li::after, but we're getting rid of this, so nothing special anymore.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -130,7 +131,9 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       case "TopSites":
         return (<TopSites />);
       case "SectionTitle":
-        return (<SectionTitle />);
+        return (<SectionTitle header={component.header} />);
+      case "Navigation":
+        return (<Navigation links={component.properties.links} alignment={component.properties.alignment} />);
```
Mardak: nit: other components pass in props on separate lines

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
```diff
@@ -0,0 +1,45 @@
+.ds-navigation {
+
+  &.ds-navigation-centered {
+    text-align: center;
+  }
+
+  &.ds-navigation-right-aligned {
+    text-align: right;
```
Mardak: Hmmmmmm............ I wonder if the API should be specified as start/end instead of right/left…

content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    a {
+      &:hover {
+        // text-decoration: underline; didn't quite match comps.
+        border-bottom: 1px solid var(--newtab-link-primary-color);
+
+        &:active,
+        &:visited {
+          border-bottom: 1px solid $blue-70;
```
Mardak: I wonder if we should introduce a `var(--newtab-link-active-color)` as we'll want to use the blue-70 elsewhere too

PR #4693, opened 2019-01-21
Bug 1521615 - Dry out header css for reusage usage with top sites.

ScottDowne:
To test:

1. Ensure the new tab is the regular version by default.
2. In `about:config`, change  `browser.newtabpage.activity-stream.discoverystream.config` to be `{"enabled":true,"layout_endpoint":"https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=dev-test-all"}`
3. Ensure new tab is rendering components with readable headers for List, Hero, and top sites.
4. Reset the pref, ensure everything is back to normal.

### DIFF
--- a/content-src/components/DiscoveryStreamBase/_DiscoveryStreamBase.scss
+++ b/content-src/components/DiscoveryStreamBase/_DiscoveryStreamBase.scss
@@ -27,4 +27,11 @@
     display: grid;
     grid-row-gap: var(--gridRowGap);
   }
+
+  .ds-header {
+    font-size: 17px;
+    line-height: 24px;
+    color: $grey-90;
+    margin: 16px 0;
+  }
 }
--- a/content-src/components/DiscoveryStreamComponents/List/List.jsx
+++ b/content-src/components/DiscoveryStreamComponents/List/List.jsx
@@ -68,7 +68,7 @@ export function _List(props) {
 
   return (
     <div>
-      <h3 className="ds-list-title">{props.header && props.header.title}</h3>
+      {props.header && props.header.title ? <div className="ds-header">{props.header.title}</div> : null }
       <hr className="ds-list-border" />
       <ul className="ds-list">{recMarkup}</ul>
     </div>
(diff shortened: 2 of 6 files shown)


content-src/components/DiscoveryStreamBase/_DiscoveryStreamBase.scss
```diff
@@ -27,4 +27,11 @@
     display: grid;
     grid-row-gap: var(--gridRowGap);
   }
+
+  .ds-header {
```
ScottDowne: I moved this into layout generic CSS, as it's not reusable. (We could make it into a component if it becomes more than just a className)

content-src/components/DiscoveryStreamComponents/List/List.jsx
```diff
@@ -68,7 +68,7 @@ export function _List(props) {
 
   return (
     <div>
-      <h3 className="ds-list-title">{props.header && props.header.title}</h3>
+      {props.header && props.header.title ? <div className="ds-header">{props.header.title}</div> : null }
```
ScottDowne: If we were to make this a component (new bug is fine with me), I can see value in being able to do validation for internal required props so it doesn't choke in one place.

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -4,8 +4,15 @@ import React from "react";
 
 export class _TopSites extends React.PureComponent {
   render() {
+    const header = this.props.header || {};
     return (
       <div className="ds-top-sites">
+        {header.title ? (
```
ScottDowne: Again, more mildly annoying validation.

content-src/components/DiscoveryStreamComponents/List/List.jsx
gvn: Yeah it could be a very tiny component if you'd like to add that to this PR. There is some value in that for sure.

ScottDowne: I would rather keep focus on spocs if possible, so would rather punt this.

PR #4701, opened 2019-01-22
Bug 1520888 - Ensure spocs don't load over a rec.

ScottDowne:


common/Reducers.jsm
```diff
@@ -60,6 +60,7 @@ const INITIAL_STATE = {
       spocs_endpoint: "",
       lastUpdated: null,
       data: {}, // {spocs: []}
+      loaded: false,
```
k88hudson: Does this prop need to be set to `loaded: false` when a layout reset happens?

k88hudson: This is covered by `INITIAL_STATE` in `DISCOVERY_STREAM_SPOCS_ENDPOINT`

PR #4702, opened 2019-01-23
Bug 1521891 - portrait images in list items should correctly cover 72x72 box

Mardak:
r?@piatra or @dmose 

Before
screen shot 2019-01-22 at 8 00 53 pm

After
screen shot 2019-01-22 at 8 00 43 pm


Mardak:
Actually, from the meeting I thought wolasi was asking why the images didn't fill in the 72x72 space. I see now that dmose filed the bug as adding whitespace…

Mardak:
I think the intention is to fill the 72x72 with as much image and not whitespace. dmose said wolasi asked for `object-fit: cover` so adding whitespace would go against that.

PR #4706, opened 2019-01-23
Bug 1521886 - update thicknesss/color of ds-list horizontal rules

dmose:
How to test:

* go to about:config
* search for stream.config
* set that preference to {"enabled":true,"layout_endpoint":"https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=dev-test-all"}
* open a new tab

Inspect the long horizontal rule at the top of the one of the list.  Also inspect the shorter line (which is implemented as a border) above one of the .ds-items.  Both should be 1 pixel thick, and the color of the lines should be #D7D7DB (which is the same as rgb(215, 215, 219)).

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/List/_List.scss
+++ b/content-src/components/DiscoveryStreamComponents/List/_List.scss
@@ -3,7 +3,12 @@ $item-font-size: 13;
 $item-line-height: 20;
 
 .ds-list-border {
-  border: 1px solid $grey-40-36;
+  border: 1px solid $grey-30;
+  border-left: 0;
+  border-right: 0;
+  border-bottom: 0;
+
+  padding-top: 1px;
 
   // Instead of using margin, we need to use these to override stuff that comes
   // by default from <hr>.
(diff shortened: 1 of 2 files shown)


Mardak:
Are you fixing the incorrect padding from the horizontal rules here?

Mardak:
I guess you can fix the padding issues as part of bug 1521119 instead

content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -3,7 +3,12 @@ $item-font-size: 13;
 $item-line-height: 20;
 
 .ds-list-border {
-  border: 1px solid $grey-40-36;
+  border: 1px solid $grey-30;
+  border-left: 0;
+  border-right: 0;
+  border-bottom: 0;
```
Mardak: No need to reinvent the styling here with `$border-secondary` existing. Also you can reset all with just `border: 0`, so
```css
border: 0;
border-top: $border-secondary;
```

content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -88,8 +93,8 @@ $item-line-height: 20;
 // Unfortunately the border needs to go _above_ the row gap as currently
 // set up, which means that some refactoring will be required to do this.
 .ds-list-item:nth-child(-n+3) { // all but the last three items
-  border-bottom: 2px solid $grey-40-36;
-  margin-bottom: -2px;  // cancel out the 2 pixels we used for the border
+  border-bottom: 1px solid $grey-30;
```
Mardak: Same thing here: `border-bottom: $border-secondary;` Also grey-40-36 is no longer used and shouldn't exist, so remove it.

dmose: Fixed.

content-src/components/DiscoveryStreamComponents/List/_List.scss
dmose: good points; fixed!

PR #4713, opened 2019-01-23
(nobug) – Reverting #4707

gvn:
Figma was inconsistent. `$grey-90` is what's actually intended.


Mardak:
Handled in https://github.com/mozilla/activity-stream/commit/85f5016ab88e22207e92dac6aa493cdc47188c98

PR #4714, opened 2019-01-23
Bug 1519773 - implement image placeholder style for list items

Mardak:
The duplication was hurting my eyes ;) r? @k88hudson Not sure if it's better as a mixin or a css class that each div additionally sets.. Also, turns out img with no image has some default border that can't be overridden??

Before
screen shot 2019-01-23 at 1 30 27 pm

After
screen shot 2019-01-23 at 1 31 15 pm


Mardak:
Ah the border gets darker because the placeholder background is darker than the usual background.

Mardak:
wolasi: "it’s an optical illusion that makes it look darker :wink:"

PR #4715, opened 2019-01-23
Bug 1520853 - update default search shortcuts config for Fx66

rlr:
This enables @amazon in only 6 countries: DE, FR, GB, IT, JP, US

And it skips pinning a shortcut if it matches your current default search engine.

To verify, I fired up a local clean build and it only pinned Amazon instead of both Google and Amazon.

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -176,14 +176,8 @@ const PREFS_CONFIG = new Map([
         return "";
       }
       const searchShortcuts = [];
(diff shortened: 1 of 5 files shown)


mconnormoz:
This definitely isn't the right way to do it, or the correct final state.  Updating the bug to be more precise about how we should do this.

k88hudson:
Code looks good here, works as expected – just waiting on some feedback from Product to get this merged.

PR #4716, opened 2019-01-23
Bug 1520849 - Add error boundary for DiscoveryStream

k88hudson:
Adds a simple error boundary for DiscoveryStream.

To test this, you will need to apply this patch:
```diff
diff --git a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
index 6d946618..cc13d287 100644
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -193,6 +193,7 @@ export class _DiscoveryStreamBase extends React.PureComponent {
   }
 
   render() {
+    throw new Error("Oops!");
     const {layoutRender} = this.props.DiscoveryStream;
     const styles = [];
     return (
```

Ensure you can still use search with this error being thrown.

### DIFF
--- a/content-src/components/Base/Base.jsx
+++ b/content-src/components/Base/Base.jsx
@@ -171,7 +171,10 @@ export class BaseContent extends React.PureComponent {
                   <ManualMigration />
                 </div>
                 }
-              {isDiscoveryStream ? <DiscoveryStreamBase /> : <Sections />}
+              {isDiscoveryStream ? (
+                <ErrorBoundary className="borderless-error">
+                  <DiscoveryStreamBase />
+                </ErrorBoundary>) : <Sections />}
               <PrefsButton onClick={this.openPreferences} />
             </div>
             <ConfirmDialog />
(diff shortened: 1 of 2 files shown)


content-src/components/Base/_Base.scss
```diff
@@ -54,6 +54,11 @@ main {
   height: 100vh;
 }
 
+
+.as-err.ds-error-fallback.ds-error-fallback {
```
Mardak: Probably want just `.ds-error-fallback` it renders as `class="ds-error-fallback as-error-fallback"`

k88hudson: Unfortunately due the way the css files are ordered the specificity isn't high enough without this, I could alternatively move this class to the error css file

k88hudson: oh sorry, this is supposed to be `.as-error-fallback.ds-error-fallback`

PR #4718, opened 2019-01-23
Bug 1520889 - Pref for toggling spocs on and off.

ScottDowne:


lib/ActivityStream.jsm
```diff
@@ -215,8 +215,9 @@ const PREFS_CONFIG = new Map([
   // See browser/app/profile/firefox.js for other ASR preferences. They must be defined there to enable roll-outs.
   ["discoverystream.config", {
     title: "Configuration for the new pocket new tab",
-    value: JSON.stringify({
+    getValue: args => JSON.stringify({
       enabled: false,
+      show_spocs: showSpocs(args),
```
k88hudson: Let's leave this as `true` for now, I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1522301 as a follow up. Alternatively, we could allow passing a list of GEOs through the pref

PR #4719, opened 2019-01-24
Bug 1521119 - make list horizontal rules show up in right places

dmose:


content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -14,7 +14,22 @@ $item-line-height: 20;
...
+  .ds-list-item:nth-child(-n+#{$columns}) {
+    border-bottom: $border-secondary;
+    margin-bottom: -1px;  // cancel out the pixel we used for the border
+
+    padding-bottom: 2px;
+  }
+}
+
 .ds-list {
+  @include except-last-grid-row(2); // XXX why do we need 2 rather than 1 here?
```
Mardak: This isn't doing what you think it's doing. The reason why you need to give it 2 is that your selector is selecting the first 2 items. Similarly your `expect(3)` is selecting the first 3. It just happens that the 3-column wide lists all have 6 items, so it looks like it's working.

You probably want `.ds-list-item:not(:nth-last-child(-n+3))` to select all but the last 3 items. Also, if you do this, you'll want to scope `expect-last-grid-row(1)` to `.ds-column[1-4]` instead of to any `.ds-list`

content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -14,6 +14,19 @@ $item-line-height: 20;
   margin-block-end: 8px;
 }
 
+// XXX this is gross, and attaches the bottom-border to the item above.
+// Ideally, we'd attach the top-border to the item that needs it.
+// Unfortunately the border needs to go _above_ the row gap as currently
+// set up, which means that some refactoring will be required to do this.
+@mixin bottom-border-except-last-grid-row($columns) {
```
Mardak: This would be more accurately named something along of "except last item in each column" as opposed to "last row"

Mardak:
wolasi says the correct behavior is to have a line on every item except the last row, which is what your named your mixin. So the behavior here is not quite right for jagged columns, but I believe we'll only use full rows anyway.

content-src/components/DiscoveryStreamComponents/List/_List.scss
dmose: Right now, the patch I have up does exactly what the name suggests.  Since it sounds like we're going to want to file another bug and change that behavior, I suggest we change the name then.

dmose: Actually, the above comment is wrong.  But I vote we leave things alone and I'll file that bug where we can address the behavioral problem.

PR #4720, opened 2019-01-24
Bug 1519764 - Disable Dark theme for Discovery Stream

punamdahiya:


punamdahiya:
@piatra Opened PR using 'unthemed' css selector for helping disable dark theme for discovery stream. I have modified all CSS where we are using [lwt-newtab-brighttext] to check for .unthemed

This has advantage of keeping experiment specific changes in Activity Stream itself and main reason I chose it instead of alternate solution of modifying [contenttheme.js](https://searchfox.org/mozilla-central/source/browser/base/content/contentTheme.js#14) to check for 'unthemed' css selector.
I have to move check in Base.jsx instead of DiscoveryStreamBase.jsx due to click on search on new tab was causing Base.jsx to rerender unsetting '.unthemed' selector on body tag. 

@k88hudson , @nt1m including you both as I discussed the approaches with you today. Please feel free to comment if you see anything amiss. Thanks!

content-src/asrouter/components/Button/_Button.scss
```diff
@@ -41,7 +41,7 @@
   }
 }
 
-[lwt-newtab-brighttext] {
+[lwt-newtab-brighttext]:not(.unthemed) {
```
piatra: I would change the name to something more descriptive like `force-light-theme` or something similar.

PR #4722, opened 2019-01-24
Bug 1521915 - replace DiscoveryStream three-period ellipses with a single unicode ellipsis character

Mardak:
r?@piatra or @gvn Moved `truncateText` to its own file with tests. Made it slightly smarter about whitespace. Oh and switched to `…`


Mardak:
Hmm.. after playing around with this for list excerpts, I think manually truncating at some fixed cap is the wrong approach as it will often be wrong depending on the letters in the string leading to a lot of wasted whitespace:
screen shot 2019-01-23 at 10 47 47 pm

k88hudson:
Agreed, manual truncation seems like it's pretty error-prone (especially with different fonts on Windows/Mac)

PR #4725, opened 2019-01-24
Bug 1522590 – Hero Component Refinements

gvn:


content-src/components/DiscoveryStreamComponents/List/List.jsx
```diff
@@ -79,6 +79,7 @@ export function _List(props) {
 }
 
 _List.defaultProps = {
+  recStartingPoint: 0,
```
k88hudson: can you add a comment about what this is?

content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -53,6 +54,16 @@ export class Hero extends React.PureComponent {
         source={truncateText(rec.domain, 22)} />
     ));
 
+    let list = (
+      <List
```
k88hudson: Can you just confirm this won't be weird in terms of the telemetry? (i.e. will it think we have another list component or something?)

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -157,6 +157,8 @@ export class _DiscoveryStreamBase extends React.PureComponent {
         return (
           <ImpressionStats rows={rows} dispatch={this.props.dispatch} source={component.type}>
             <Hero
+              subComponentType={embedWidth >= 9 ? `cards` : `list`}
```
k88hudson: personally I would pass in row width and add this logic to `Hero` if we intend to keep it consistent across usages

gvn: Yeah, I can see it both ways I guess. I'm trying to keep the components fairly dumb and light on logic.

gvn: This is also the only way I can foresee Hero being embedded.

gvn:
All feedback should be addressed! 🎉

PR #4726, opened 2019-01-24
Bug 1519878 - implement full width ds-list items

Mardak:
r?@dmose or @piatra Slightly refactoring of using :not to avoid applying a style then overriding with more specificity.

screen shot 2019-01-24 at 4 34 21 pm

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/List/_List.scss
+++ b/content-src/components/DiscoveryStreamComponents/List/_List.scss
@@ -1,5 +1,6 @@
 // Type sizes
 $item-font-size: 13;
+$item-image-size: 72px;
 $item-line-height: 20;
 
 // XXX this is gross, and attaches the bottom-border to the item above.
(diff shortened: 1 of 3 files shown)


content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -24,21 +50,35 @@ $item-line-height: 20;
...
+    .ds-column-9 &,
+    .ds-column-10 &,
+    .ds-column-11 &,
+    .ds-column-12 & {
+      grid-template-columns: repeat(3, 1fr);
+      grid-row-gap: 18px;
+    }
+
+    .ds-list-item-excerpt {
+      display: none;
```
piatra: Could we not render this instead, since `fullWidth` is a prop?

Mardak: There had been requests to avoid conditionally rendering content to allow remote css to unhide if necessary.

PR #4727, opened 2019-01-24
Bug 1522677 – Card Grid refinements

gvn:


gvn:
If this looks good feel free to land it while I'm out.

content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
```diff
@@ -1,11 +1,23 @@
+$col4-header-line-height: 20;
+$col4-header-font-size: 13;
```
ScottDowne: I don't see this actually being set to the font size, just added to the limit-visibile-lines mixin.

ScottDowne:
One nit, and I'm also not sure what's going on with "Vertical gap between rows should be 24px" how do I test that, or what should I be looking for?

ScottDowne:
Given Gavin is out for the day, I think merging this and doing a followup for the font size that might be broken when Gavin's back is fine.

Mardak:
The hover state of borderless cards should have no border. Also seems like dev-test-all 2/3 column "Card" vs "Borderless Card" both are actually set as border: no-border !

content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
gvn: That mixin now takes the font size for some internal calculations and emits it as well.

PR #4729, opened 2019-01-25
Bug 1521860 - Ensure we don't use object spread syntax with data from server

piatra:
This is the only occurrence I could still find. For future work how should we handle this? Do we prevent rendering or should we provide default props with an appropriate shape that allows destructuring?

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/List/List.jsx
+++ b/content-src/components/DiscoveryStreamComponents/List/List.jsx
@@ -60,8 +60,16 @@ export function _List(props) {
 
   const recs = feed.data.recommendations;
 
-  let recMarkup = recs.slice(props.recStartingPoint, props.items).map((rec, index) => (
-    <ListItem {...rec} key={`ds-list-item-${index}`} index={index} type={props.type} dispatch={props.dispatch} />)
+  let recMarkup = recs.slice(0, props.items).map((rec, index) => (
+    <ListItem key={`ds-list-item-${index}`}
+      index={index}
+      type={props.type}
+      id={rec.id}
+      title={rec.title}
+      image_src={rec.image_src}
+      url={rec.url}
+      domain={rec.domain}
+      dispatch={props.dispatch} />)
   );
 
   const listStyles = [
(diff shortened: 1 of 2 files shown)


Mardak:
I believe @ScottDowne intended to file the bug as not using "object spread" instead of "object destructuring" ??

ScottDowne:
Ugh, yeah, @Mardak is correct.

ScottDowne:
To be specific, this is the line that got my attention: https://github.com/mozilla/activity-stream/blob/master/content-src/components/DiscoveryStreamComponents/List/List.jsx#L66

content-src/components/DiscoveryStreamComponents/List/List.jsx
```diff
@@ -60,8 +60,16 @@ export function _List(props) {
...
+  let recMarkup = recs.slice(0, props.items).map((rec, index) => (
+    <ListItem key={`ds-list-item-${index}`}
+      index={index}
+      type={props.type}
+      id={rec.id}
+      title={rec.title}
+      image_src={rec.image_src}
+      url={rec.url}
+      domain={rec.domain}
+      dispatch={props.dispatch} />)
```
Mardak: I think we'll need `rec.excerpt` for wide lists

Mardak: Keys on the rec seem to include ` "id", "url", "domain", "title", "excerpt", "image_src", "published_timestamp", "parameter_set", "domain_affinities", "item_score" `

Mardak:
I rebased and added excerpt to https://github.com/mozilla/activity-stream/commit/ad01e12992b5b2227fd93f99ab8e43dcb3eec715

content-src/components/DiscoveryStreamComponents/List/List.jsx
```diff
@@ -60,8 +60,16 @@ export function _List(props) {
 
   const recs = feed.data.recommendations;
 
-  let recMarkup = recs.slice(props.recStartingPoint, props.items).map((rec, index) => (
-    <ListItem {...rec} key={`ds-list-item-${index}`} index={index} type={props.type} dispatch={props.dispatch} />)
+  let recMarkup = recs.slice(0, props.items).map((rec, index) => (
```
Mardak: I guess we all missed the slice change to 0 instead of `recStartingPoint` 😊 Fixed in #4740

PR #4730, opened 2019-01-25
Bug 1518261

k88hudson:
This patch adds a periodic check after every system tick (i.e. 5 minutes) and refactors some of the fetching into a single function. If the cache has expired, a refresh will occur.

I think this one is going to be hard to manually QA, but I'll try to think of something

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -67,44 +67,21 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
-  async fetchLayout() {
-    const endpoint = this.config.layout_endpoint;
+  async fetchFromEndpoint(endpoint) {
     if (!endpoint) {
-      Cu.reportError("No endpoint configured for pocket, so could not fetch layout");
+      Cu.reportError("Tried to fetch endpoint but none was configured.");
       return null;
     }
     try {
       const response = await fetch(endpoint, {credentials: "omit"});
       if (!response.ok) {
         // istanbul ignore next
-        throw new Error(`Layout endpoint returned unexpected status: ${response.status}`);
+        throw new Error(`${endpoint} returned unexpected status: ${response.status}`);
       }
       return response.json();
     } catch (error) {
       // istanbul ignore next
-      Cu.reportError(`Failed to fetch layout: ${error.message}`);
-    }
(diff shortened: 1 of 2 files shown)


lib/DiscoveryStreamFeed.jsm
```diff
@@ -67,54 +67,31 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
-  async fetchLayout() {
-    const endpoint = this.config.layout_endpoint;
+  async fetchFromEndpoint(endpoint) {
     if (!endpoint) {
-      Cu.reportError("No endpoint configured for pocket, so could not fetch layout");
+      Cu.reportError("Tried to fetch endpoint but none was configured.");
```
ScottDowne: How do errors get reported? Is it obvious where the error is coming from, or would it be helpful to add a bit of context to the error, example "Tried to fetch Discovery Stream endpoint but none was configured."

lib/DiscoveryStreamFeed.jsm
```diff
@@ -279,6 +250,12 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           await this.enable();
         }
         break;
+      case at.SYSTEM_TICK:
+        // Only refresh if we loaded once in .enable()
+        if (this.config.enabled && this.loaded) {
+          await this.refreshAll({updateOpenTabs: false});
```
ScottDowne: I think this'll end up firing update events on every system tick.

If I'm reading it correctly, it goes into a layout/component/spoc load function, gets cache. If it needs to get fresh data, it does. Then it sends the fresh data or cache to the reducer.

The old code was slightly less dry, but also only fired system tick updates if cache has expired.

lib/DiscoveryStreamFeed.jsm
k88hudson: You do get a line number / stack trace, so that should be enough (although I could improve this by maybe adding a reference to the endpoint?

k88hudson:
Ok, I have added some more logic that covers checking cache expiry – I think ideally it would be nicer if we added that right to the cache API itself.

lib/DiscoveryStreamFeed.jsm
ScottDowne: Line numbers is fine, let's not worry about this too much then.

PR #4731, opened 2019-01-25
Bug 1520258 - Adding spoc frequency caps based on visibility.

ScottDowne:


ncloudioj:
@thecount - I've also implemented the card-level impression stats in #4743 with the same `INTERSECTION_RATIO`, would it be possible to reuse that here for your case?

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -1,12 +1,69 @@
...
+  setupIntersectionObserver() {
+    const options = {threshold: INTERSECTION_RATIO};
+    this._intersectionObserver = new IntersectionObserver(this.onIntersectionObserve, options);
+    this._intersectionObserver.observe(this.cardElement);
+  }
+
+  onIntersectionObserve(entries) {
+    for (let entry of entries) {
+      if (entry.isIntersecting && entry.intersectionRatio >= INTERSECTION_RATIO) {
+        this.isIntersecting();
```
ncloudioj: You might want to break this for loop once the expected intersection gets handled.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -1,12 +1,69 @@
...
+      }
+    }
+  }
+
+  isIntersecting() {
```
ncloudioj: Consider renaming this function. It does not check if the card is intersecting with the viewport,  instead, it just does some specific handling for SPOCS.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -1,12 +1,69 @@
...
     this.onLinkClick = this.onLinkClick.bind(this);
   }
 
+  componentDidMount() {
+    if (this.props.document.visibilityState === VISIBLE) {
+      this.setupIntersectionObserver();
+    } else {
+      this._onVisibilityChange = () => {
+        if (this.props.document.visibilityState === VISIBLE) {
+          this.setupIntersectionObserver();
```
ncloudioj: nit: `this.setupIntersectionObserver.bind(this)` is not needed here since the fat arrow function automatically binds this for you.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -25,7 +82,7 @@ export class DSCard extends React.PureComponent {
 
   render() {
     return (
-      <a href={this.props.url} className="ds-card" onClick={this.onLinkClick}>
+      <a href={this.props.url} className="ds-card" onClick={this.onLinkClick} ref={this.cardElementRef}>
```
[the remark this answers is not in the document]
ScottDowne: I thought react was deprecating string based refs?

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
ScottDowne: I'm not longer using the arrow for this anymore.

ncloudioj: `this._onVisibilityChange = () => {...`, and you bound `this` here https://github.com/mozilla/activity-stream/pull/4731/files#diff-cb93e859d1f25eeace50eb38edb98cd7R12

lib/DiscoveryStreamFeed.jsm
```diff
@@ -203,11 +207,57 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+    return data;
+  }
+
+  // Frequency caps are based on campaigns, which may include multiple spocs.
+  // We currently support two types of frequency caps:
+  // - lifetime: Indicates how many times spocs from a campaign can be shown in total
+  // - period: Indicates how many times spocs from a campaign can be shown within a period
+  //
+  // So, for example, the feed configuration below defines that for campaign 1 no more
+  // than 5 spocs can be show in total, and no more than 2 per hour.
```
ncloudioj: nit: s/show/shown/

lib/DiscoveryStreamFeed.jsm
```diff
@@ -203,11 +207,57 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+  //    "period": 3600
+  //  }
+  // }
+  isBelowFrequencyCap(impressions, spoc) {
+    const campaignImpressions = impressions[spoc.campaign_id];
+    if (!campaignImpressions) {
+      return true;
+    }
+
+    const lifeTimeCap = Math.min(spoc.caps && spoc.caps.lifetime, MAX_LIFETIME_CAP);
```
ncloudioj: nit: let's add a guard here in case of `spoc.caps.lifetime` is unset, otherwise, `Math.min(undefined, ...)` would return NAN.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -277,6 +327,50 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
+  recordCampaignImpression(campaignId) {
+    let impressions = this.readImpressionsPref(SPOC_IMPRESSION_TRACKING_PREF);
+
+    const timeStamps = impressions[campaignId] || [];
+    timeStamps.push(Date.now());
+    impressions = Object.assign(impressions, {[campaignId]: timeStamps});
```
ncloudioj: Let's do `{...impressions, [campaignID]: timeStamps}` this instead.

PR #4733, opened 2019-01-25
Bug 1522680 - Source for spocs

ScottDowne:


k88hudson:
Spacing looks off between the sponsored link and URL according to https://bug1522680.bmoattachments.org/attachment.cgi?id=9038936

PR #4734, opened 2019-01-25
Bug-1522909 - Add optional header to navigation and update header fon…

punamdahiya:
…t weight


punamdahiya:
Navigation with header screenshot 

navigation-header

PR #4743, opened 2019-01-28
Bug 1522832 - Fine-grained visibility on impression stats

ncloudioj:
This intends to improves the accuracy of the Pocket impression stats by switching from page level visibility (`dom.visitbilityState`) to section level [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).


ncloudioj:
There is an edge case for the 100% visibility detection, which is that when a component has multiple columns (or been zoomed) that can't fit in a single page without scrolling, this implementation would never report impression due to the 100% intersection ratio. Need to re-consider it, perhaps we have to use the per-article visibility detection other than at the component level.

ncloudioj:
According to the discussion in this [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1522832#c4), we made a few changes to the impression wrapper:

* Use Intersection Observer API and page visibility together to detect the impression
* Push the current impression wrapping from the section level down to the card level for the max accuracy
* Impression ping for each card will be sent separately without batching, we might want to optimize this if the overhead turns out to be significant

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -2,24 +2,13 @@ import {CardGrid} from "content-src/components/DiscoveryStreamComponents/CardGri
...
 
-// According to the Pocket API endpoint specs, `component.properties.items` is a required property with following values:
-//   - List 1-12 items
-//   - Hero 1-5 items
-//   - CardGrid 1-16 items
-// To enforce that, we define various maximium items for individual components as an extra check.
-// Note that these values are subject to the future changes of the specs.
-const MAX_ROWS_HERO = 5;
-const MAX_ROWS_LIST = 12;
-const MAX_ROWS_CARDGRID = 16;
```
ScottDowne: Is this something we're not enforcing anymore? I don't see it in the patch anymore.

ScottDowne: Ah, I see, because we're now wrapping the contents inside the components, and not around the outside of the components, we have access to the content already that these consts were being used to get.

ncloudioj:
This needs a rebase against the existing intersection observer stuff in master.

ncloudioj:
Added two more commits to:

* Support batch impression reporting instead of sending one ping for each impression separately, which should save lots of HTTP requests. The difference from the old implementation is now that we will report the impression stats right before the session ends, therefore a latency should be expected.
* Integrate `SpocsIntersectionObserver` into the `ImpressionStats` because they use the exact same logic with `IntersectionObserver`.

@ScottDowne Can you take another look at this? To make it easier to review, I've broken it down into separate smaller commits, and I believe that you had already reviewed the first one last week.

Thanks!

k88hudson:
@ScottDowne @ncloudioj what's the status on this?

ncloudioj:
@k88hudson This requires another review from @ScottDowne, see the comment above.

ScottDowne:
Also somewhere related: https://github.com/mozilla/activity-stream/pull/4795

content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -74,16 +74,19 @@ export class Hero extends React.PureComponent {
               <div className="img" style={{backgroundImage: `url(${heroRec.image_src})`}} />
             </div>
             <div className="meta">
-              <div className="header-and-excerpt">
-                <header>{heroRec.title}</header>
-                <p className="excerpt">{heroRec.excerpt}</p>
-              </div>
+              <header>{heroRec.title}</header>
```
Mardak: I'll fix this loss of `<div className="header-and-excerpt">` in https://github.com/mozilla/activity-stream/pull/4798

PR #4745, opened 2019-01-29
chore(tests): Add basic hero and list tests to ensure correct item offset rendering

Mardak:
r?@punamdahiya or @gvn Adds tests to cover the regressions fixed in #4739 and #4740.

### DIFF
--- a/test/unit/content-src/components/DiscoveryStreamComponents/List.test.jsx
+++ b/test/unit/content-src/components/DiscoveryStreamComponents/List.test.jsx
@@ -4,7 +4,7 @@ import React from "react";
 import {shallow} from "enzyme";
 
 describe("<List> presentation component", () => {
-  const ValidRecommendations = [{a: 1}, {a: 2}];
+  const ValidRecommendations = [{url: 1}, {url: 2}, {url: 3}];
   const ValidListProps = {
     DiscoveryStream: {
       feeds: {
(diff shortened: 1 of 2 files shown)


test/unit/content-src/components/DiscoveryStreamComponents/List.test.jsx
```diff
@@ -48,6 +48,27 @@ describe("<List> presentation component", () => {
...
+    const wrapper = shallow(<List {...ValidListProps} recStartingPoint={1} />);
+
+    const listItem = wrapper.find(ListItem);
+    assert.lengthOf(listItem, ValidRecommendations.length - 1);
+  });
+
+  it("should return expected ListItem when offset", () => {
+    const wrapper = shallow(<List {...ValidListProps} items={1} recStartingPoint={2} />);
+
+    const listItem = wrapper.find(ListItem);
```
punamdahiya: nit: can add check assert.lengthOf(listItem, ValidRecommendations.length - 2); to test that recommendations are offset by 2 and will show 2 less items from feed.

Mardak: The .prop below implicitly requires length to be 1 as it only works for a single item. I'll add a lengthOf to be explicit

punamdahiya: For this case extra assert checking is what I had in my mind

Mardak: The previous 2 tests already test 1) item limiting and 2) offsetting, so it shouldn't be necessary to repeat the length check as theoretically that's been tested already so this test can focus on a different aspect of making sure the contents and not just the length are expected. I'll add the lengthOf == 1 check to be explicit (as length -2 isn't quite right as we're saying there should be `items={1}`)

Mardak: Actually, decided to try a different combination not tested elsewhere. items=2 start=1:

```js
const wrapper = shallow(<List {...ValidListProps} items={2} recStartingPoint={1} />);
const listItemUrls = wrapper.find(ListItem).map(i => i.prop("url"));
assert.sameOrderedMembers(listItemUrls, …);
```

PR #4746, opened 2019-01-29
Port Bug 1514594: Part 3 - Change ChromeUtils.import API. *** Bug 1514594: Part 3a - Change ChromeUtils.import to return an exports object; not pollute global. r=mccr8

Mardak:
r? @k88hudson Second commit has the local fixes to get our automation passing. I'll squash together when landing.


Mardak:
Updated approach to use updated plugins where statements are removed. Fixes up other eslint and istanbul issues for common/PerfService

common/PerfService.jsm
```diff
@@ -1,17 +1,18 @@
-/* globals Services */
 "use strict";
 
 /* istanbul ignore if */
 if (typeof ChromeUtils !== "undefined") {
-  ChromeUtils.import("resource://gre/modules/Services.jsm");
+  // eslint-disable-next-line no-var, vars-on-top
+  var {Services} = ChromeUtils.import("resource://gre/modules/Services.jsm");
```
k88hudson: Maybe this instead?
```js
let Services;
if (typeof ChromeUtils !== "undefined") {
  ChromeUtils.import("resource://gre/modules/Services.jsm");
  Services = ChromeUtils.import("resource://gre/modules/Services").Services;
```

k88hudson:
We might want to communicate this change more broadly since it's likely to impact uplifts. Unless of course that bug gets uplifted to 66 (which would be a lot easier!)

common/PerfService.jsm
Mardak: That's what I had originally 😉 , but that just still has the local scoped Services blocking the global Services problem.

k88hudson: Ah ok I see the problem. Leave it as is (might be worth commenting)

Mardak: Actually, I guess that was only a problem when I was doing double Proxy ?? I'll check again with `let Services`…

Mardak: I ended up adding a comment as to why `var` inside instead of `let` then also making import return the current `global`

PR #4747, opened 2019-01-29
Bug 1523408 - Ensure we don't fire impressions stat events in topsite…

ScottDowne:
…s via discovery stream


ScottDowne:
@ncloudioj What we talked about yesterday. I'm not sure if we want to do this sort of check in other places or not, but for now this fixes the bug and steps to reproduce as mentioned in the bugzilla bug.

ScottDowne:
@ncloudioj Updated. Take a quick look?

ncloudioj:
@ScottDowne Looks great. Thanks!

PR #4749, opened 2019-01-31
Bug 1518521 - Record a scalar for how long it takes to present about:home after start-up

ncloudioj:


rlr:
@ncloudioj code looks good. I'm trying to test it locally but I can't find the value in `about:telemetry#scalars-tab`. Any tips?

ncloudioj:
@rlr We need to wait for @mikeconley to land the telemetry change first in M-C, then we can test it. That work is tracked by https://phabricator.services.mozilla.com/D16112

ncloudioj:
This new metric is verified in my local build.

PR #4750, opened 2019-01-31
Bug 1524021 - Dark Mode not supported messaging

punamdahiya:


locales-src/en-US/strings.properties
```diff
@@ -213,3 +213,7 @@ firstrun_skip_login=Skip this step
 
 # LOCALIZATION NOTE (context_menu_title): Action tooltip to open a context menu
 context_menu_title=Open menu
+
+# This message is shown to nightly en-US users viewing New Tab experience in dark mode.
+ds_message_content=This version of New Tab doesn not support dark mode yet.
```
flodolo: does not, doesn't? I don't see mention of copy in the bug.

Also, are we showing it really only to en-US users?

Mardak: We'll show to all users but because this is temporary, I believe it should be fine to not bother localizing. Hopefully we get dark mode in before the end of 67 nightly.

flodolo: > We'll show to all users but because this is temporary, I believe it should be fine to not bother localizing. Hopefully we get dark mode in before the end of 67 nightly.

If that's the plan it should be hardcoded. If you put it in the .properties file it will get translated.

content-src/components/Base/Base.jsx
```diff
@@ -186,7 +186,7 @@ export class BaseContent extends React.PureComponent {
                 }
               {isDiscoveryStream ? (
                 <ErrorBoundary className="borderless-error">
-                  <DiscoveryStreamBase />
+                  <DiscoveryStreamBase isDarkModeMessage={prefs.darkModeMessage} />
```
punamdahiya: @k88hudson will this work or is there a better way to get prefs.darkModeMessage inside DiscoveryStreamBase.jsx to avoid passing it as props?

lib/ActivityStream.jsm
```diff
@@ -222,6 +222,10 @@ const PREFS_CONFIG = new Map([
       layout_endpoint: "https://getpocket.com/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=basic",
     }),
   }],
+  ["darkModeMessage", {
+    title: "Boolean flag that decides whether to show the dark Mode message or not.",
+    value: true,
+  }],
```
punamdahiya: @k88hudson this pref is true for nightly en-US locale only and stay false or undefined for rest of the channels, is there another place where we set that? thanks!

content-src/components/DiscoveryStreamComponents/DarkModeMessage/DarkModeMessage.jsx
```diff
@@ -0,0 +1,40 @@
...
+export class _DarkModeMessage extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.handleSwitch = this.handleSwitch.bind(this);
+    this.handleCancel = this.handleCancel.bind(this);
+  }
+
+  handleSwitch() {
+    // TBD: Call switch to older version
+    // this.props.dispatch(ac.UserEvent({event: at.SWITCH_DEFAULT}));
```
Mardak: This will want to be `this.store.dispatch(ac.BroadcastToContent({type: at.DISCOVERY_STREAM_OPT_OUT}));`

punamdahiya: perfect, thanks!

punamdahiya:
@Mardak or @k88hudson requesting review. Switch to older version flow has dependency on https://github.com/mozilla/activity-stream/pull/4754

Open to feedback on improving passing darkModeMessage pref from base.jsx and ensuring darkModeMessage pref is true for nightly channel only. Thanks!

punamdahiya:
Screenshot: Dark Mode Message
dark-mode-message

lib/ActivityStream.jsm
k88hudson: You could basically set this to `IS_NIGHTLY_OR_UNBRANDED_BUILD`, so that it would only be true if the build is nightly (and default `false` otherwise)

Mardak:
Merged #4754 for `DISCOVERY_STREAM_OPT_OUT` but it conflicts with the added new pref in ActivityStream.jsm

Mardak:
Rebasing and merging as tests should be green on latest master…

PR #4754, opened 2019-02-01
Bug 1521084 - Make opt-out button work about:preferences if users are not enrolled in an experiment

Mardak:
r?@k88hudson or @ScottDowne or @piatra This changes the enabled checks to an active check where the feed combines the opt out pref with config enabled. Other consumers are updated to check a new `DiscoveryStream.active` property on the redux state.

Notably, I did initially try just exposing `DiscoveryStream.config.enabled` as the combined optout+config value, but other places are expecting config to be exactly what the pref is, e.g., devtools triggering DISCOVERY_STREAM_CONFIG_SET_VALUE.

@punamdahiya There's an example of broadcasting the opt out that you'll want for the button in #4750. The reducer updates `active` immediately so it should cause content pages to respond without needing to wait for the actual pref to change.


punamdahiya:
> @punamdahiya There's an example of broadcasting the opt out that you'll want for the button in #4750. The reducer updates `active` immediately so it should cause content pages to respond without needing to wait for the actual pref to change.

@Mardak does [this call](https://github.com/mozilla/activity-stream/pull/4750/files#diff-c3e2fa0fdef5277f80112c27c30e6777R14) look correct? I tried cherry-picking my commit on top of your changes. Using BroadcastToContent cause page to respond but fails to change optOut pref to true.

Updating call to this.props.dispatch(ac.AlsoToMain({type: at.DISCOVERY_STREAM_OPT_OUT})); changes pref value but not sure if we need an additional call to dispatch to all content processes

Mardak:
Ah, I guess Broadcast is intended from main to content so it sends it to itself and all content. I guess typically we don't send from one content to other content, so yeah AlsoToMain is fine. It'll update the current page and main where the feed will set the pref and cause other content to update too.

common/Reducers.jsm
```diff
@@ -49,6 +49,7 @@ const INITIAL_STATE = {
   },
   // This is the new pocket configurable layout state.
   DiscoveryStream: {
+    active: false,
```
k88hudson: Might be worth adding a comment about what this means / what it is?

Mardak:
@k88hudson I updated the approach as you suggested to expose config.enabled as the combined value with opt out. This ended up needing to change some other parts (in particular `DISCOVERY_STREAM_CONFIG_SET_VALUE`) to not assume `.config` has a clean value.

I also ended up getting rid of `Services.prefs` usage for PrefsFeed/getState().Prefs stuff.

piatra:
This looks good. TC is failing because it's not rebased on top of latest master.

Mardak:
Oh right, rebased and repushed on top of latest master

k88hudson:
Looks great, thanks!

k88hudson:
Thanks for fixing :+1:

Mardak:
Rebased and fixed conflicts from #4731

PR #4757, opened 2019-02-05
Bug 1494232 - addonsInfo.addons does not include userDisabled addons

piatra:


piatra:
Closing this for now, follow along with the status in bug 1494232.

PR #4759, opened 2019-02-05
Bug 1524669 - Add protocol checking to link urls for pocket new tab

punamdahiya:


content-src/components/DiscoveryStreamComponents/SafeAnchor/SafeAnchor.jsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+export class SafeAnchor extends React.PureComponent {
+  safeURI(url) {
+    if (!url) {
+      return "";
+    }
+    const {protocol} = new URL(url);
```
piatra: If this doesn't contain `http`/`https` it will throw an error. We need to wrap it in a try catch.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -78,26 +79,28 @@ export class DSCard extends React.PureComponent {
...
     return (
-      <a href={this.props.url} className="ds-card" onClick={this.onLinkClick} ref={this.cardElementRef}>
-        <div className="img-wrapper">
-          <div className="img" style={{backgroundImage: `url(${this.props.image_src}`}} />
-        </div>
-        <div className="meta">
-          <div className="info-wrap">
-            <header className="title">{this.props.title}</header>
-            {this.props.excerpt && <p className="excerpt">{this.props.excerpt}</p>}
+      <div ref={this.cardElementRef}>
```
punamdahiya: keeping the ref prop out of SafeAnchor as we are overriding this logic in bug 1519879 where we will pull out logic that handles intersection observer in DSCard as a separate component

punamdahiya:
@piatra Updated PR with tests and fixes to handle merge conflicts with latest master. thanks!

content-src/components/DiscoveryStreamComponents/SafeAnchor/SafeAnchor.jsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+export class SafeAnchor extends React.PureComponent {
+  safeURI(url) {
+    try {
+      const {protocol} = new URL(url);
```
piatra: nit: only wrap the call the `new URL` in a try catch. imo it's best to keep the try catch block as small as possible so that it doesn't accidentally catch things we are not expecting.

punamdahiya: @piatra done, thanks!

Mardak:
The bug was filed for also checking images. Is that happening as a separate commit/pr/bug?

punamdahiya:
> The bug was filed for also checking images. Is that happening as a separate commit/pr/bug?

Thanks Ed for calling that out, will update patch with allowing only http[s] protocol for background images url.

Mardak:
I just realized our CSP of `img-src https: data: blob:` should prevent non-https CSS `background-image` and `` so we might not need to additionally filter. Would be good to confirm and potentially we might want to restrict images to `https://img-getpocket.cdn.mozilla.net` instead of just being `https` -- checking with security.

punamdahiya:
> I just realized our CSP of `img-src https: data: blob:` should prevent non-https CSS `background-image` and `` so we might not need to additionally filter.

Good to know,  so we don't need  http[s] protocol check for images in this PR.

 > Would be good to confirm and potentially we might want to restrict images to `https://img-getpocket.cdn.mozilla.net` instead of just being `https` -- checking with security.

Isn't restricting domain with whitelist pref scoped here - https://bugzilla.mozilla.org/show_bug.cgi?id=1521203#c6

punamdahiya:
As per comment https://bugzilla.mozilla.org/show_bug.cgi?id=1524669#c3, keeping scope for protocol check to links.

PR #4760, opened 2019-02-05
Bug 1523703 - Pref on Pocket Newtab experience in Nightly for EN and CA

k88hudson:
The result of this patch should be that pocket is on by default for US/CA nightly users (not other channels). SPOCs should only be on by default in the US


piatra:
@k88hudson Is merging the backport into this PR ok? Do you want it done separately?

PR #4761, opened 2019-02-05
Bug 1525413 - Perf work on loading discovery stream

ScottDowne:


ScottDowne:
First attempt, seems to work and might be just enough. Thoughts?

content-src/components/Base/Base.jsx
```diff
@@ -162,6 +162,7 @@ export class BaseContent extends React.PureComponent {
 
     const outerClassName = [
       "outer-wrapper",
+      isDiscoveryStream && "discovery-stream-outer-wrapper",
```
ScottDowne: I had to add this because otherwise the search bar starts in the middle of the page then jumps to the top. Dunno what's worse.

content-src/components/Base/_Base.scss
```diff
@@ -54,6 +54,12 @@ main {
 
 }
 
+.discovery-stream-outer-wrapper {
```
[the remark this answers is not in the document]
ScottDowne: Would you also accept a comment above the `margin: 0 auto;` as to why I'm doing that?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -296,8 +296,10 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       this.store.dispatch;
 
     await this.loadLayout(dispatch);
-    await this.loadComponentFeeds(dispatch);
-    await this.loadSpocs(dispatch);
+    await Promise.all([
+      this.loadComponentFeeds(dispatch),
+      this.loadSpocs(dispatch),
+    ]);
```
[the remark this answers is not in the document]
ScottDowne: We should chat over vidyo when you get in around how best to catch these.

content-src/components/Base/_Base.scss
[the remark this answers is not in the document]
dmose: I see three advantages to doing this by classname:

* in addition to making reading the SCSS source clearer, it makes using the devtools inspector and reading the JSX clearer as well (no need to refer back to the source to figure out what's going on, enabling smoothing code reading and debugging)
* it tends to be more robust against refactoring
* encourages separation of concerns by making it clear that this class is specifically about a certain functionality, and other functionality should live in other classes

common/Reducers.jsm
```diff
@@ -465,7 +468,14 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
     case at.DISCOVERY_STREAM_LAYOUT_RESET:
       return {...prevState, lastUpdated: INITIAL_STATE.DiscoveryStream.lastUpdated, layout: INITIAL_STATE.DiscoveryStream.layout};
     case at.DISCOVERY_STREAM_FEEDS_UPDATE:
-      return {...prevState, feeds: action.data || prevState.feeds};
+      return {
+        ...prevState,
+        feeds: {
+          ...prevState.feeds,
```
dmose: It's not obvious to me from reading this code why ...prevState.feeds is necessary.  If it is truly necessary, I think a comment in the code explaining why would be helpful.

ScottDowne: I'm just being careful that if other data ends up being placed on that in a future patch, I won't end up clobbering it.

It's not actually a requirement to my patch.

test/unit/common/Reducers.test.js
```diff
@@ -668,6 +668,16 @@ describe("Reducers", () => {
       const state = DiscoveryStream(undefined, {type: at.DISCOVERY_STREAM_CONFIG_CHANGE, data: {enabled: true}});
       assert.deepEqual(state.config, {enabled: true});
     });
+    it("should load feeds with DISCOVERY_STREAM_FEEDS_UPDATE", () => {
+      const data = {
```
dmose: Please split this into setup/execute/verify blocks separated by whitespace for readability.

test/unit/content-src/components/DiscoveryStreamComponents/DiscoveryStreamBase.test.jsx
```diff
@@ -0,0 +1,37 @@
+import {_DiscoveryStreamBase as DiscoveryStreamBase} from "content-src/components/DiscoveryStreamBase/DiscoveryStreamBase";
+import React from "react";
+import {shallow} from "enzyme";
+
+describe("<DiscoveryStreamBase>", () => {
```
dmose: Thanks for adding the new test file!

dmose:
Be sure to add a "how to QA" section in the bug itself.

test/unit/content-src/components/DiscoveryStreamComponents/DiscoveryStreamBase.test.jsx
ScottDowne: It's what the code wanted.

PR #4765, opened 2019-02-06
Bug 1525581 - Load cached content on startup and update in the backgr…

piatra:
…ound

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -295,8 +295,12 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     return feed;
   }
 
+  /**
+   * Called at startup to update cached data in the background.
+   */
   async _maybeUpdateCachedData() {
     const expirationPerComponent = await this._checkExpirationPerComponent();
+    // Pass in `store.dispatch` to send the updates only to main
     if (expirationPerComponent.layout) {
       await this.loadLayout(this.store.dispatch);
     }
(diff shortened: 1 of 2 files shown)


lib/DiscoveryStreamFeed.jsm
```diff
@@ -286,18 +319,22 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
     await this.loadSpocs(dispatch);
+    await this._maybeUpdateLayout(isStartup, layout);
+
     this.loaded = true;
   }
 
-  async enable() {
-    await this.refreshAll({updateOpenTabs: true});
+  async enable(options = {}) {
+    const {isStartup} = options;
```
ScottDowne: I don't think you need this. I think we can always assume the `refreshAll` inside `enable` is always functionally a startup.

ScottDowne: The other case of a pref change can be treated as a startup too.

I'm only aware of three cases where we call enable. Init, pref change, system tick.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -286,18 +319,22 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
-    const dispatch = options.updateOpenTabs ?
+    const {updateOpenTabs, isStartup} = options;
+    const dispatch = updateOpenTabs ?
       action => this.store.dispatch(ac.BroadcastToContent(action)) :
       this.store.dispatch;
 
-    await this.loadLayout(dispatch);
+    const layout = await this.loadLayout(dispatch, isStartup);
     await this.loadComponentFeeds(dispatch);
     await this.loadSpocs(dispatch);
```
ScottDowne: I would confirm with Nick, but I think layout, feeds, and spocs (so the above three functions) all have caches that need to rely on a week long cache for startups.

PR #4772, opened 2019-02-07
Bug 1525494 - Parallel component feed loads

ScottDowne:


content-src/lib/selectLayoutRender.js
```diff
@@ -45,7 +45,16 @@ export const selectLayoutRender = createSelector(
...
+        let {data} = feeds.data[component.feed.url];
+
+        if (component && component.properties && component.properties.offset) {
+          data = {
+            ...data,
+            recommendations: data.recommendations.slice(component.properties.offset),
+          };
+        }
+
+        return {...component, data: maybeInjectSpocs(data, component.spocs)};
```
ScottDowne: These changes are further optimizations to lower the number of feed requests.

If the server wants to combine feeds, they can now. Because I've added:

1. Duplicate feeds with the same url don't get re fetched.
2. If two components use the same feed, they can be told to use a different segment from the same data.

It's still on the server to ensure the offset param doesn't allow dupes. So we can essentially combine smaller feeds into one, the components can use the same url, and the data would be the same, so the offset would cut out unique slices from the data.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -145,28 +145,48 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
+  setupFeedPromise(url, newFeeds) {
+    const feedPromise = this.getComponentFeed(url);
+
+    feedPromise.then(data => {
+      newFeeds[url] = data;
```
ScottDowne: This is needed because the format we expected the finished data to be in wasn't what the promise.all would return, so we map it here as it finishes.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -145,28 +145,48 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+    }
+
+    for (let row of DiscoveryStream.layout) {
+      if (!row || !row.components) {
+        continue;
+      }
+      for (let component of row.components) {
+        if (component && component.feed) {
+          const {url} = component.feed;
+          if (!newFeeds[url]) {
```
ScottDowne: This is the line that stops duplicate feeds from being re fetched needlessly.

lib/DiscoveryStreamFeed.jsm
dmose: Makes sense.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -145,25 +145,48 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+            newFeedsPromises.push(this.setupFeedPromise(url, newFeeds));
           }
         }
       }
-
-      await this.cache.set("feeds", newFeeds);
-      sendUpdate({type: at.DISCOVERY_STREAM_FEEDS_UPDATE, data: newFeeds});
     }
+
+    await Promise.all(newFeedsPromises);
```
dmose: It's probably OK that you don't have a catch here, since all the subpromises do have catches.  Can you add a comment explaining that to the code?  It's at least possible that the m-c eslint rules will barf on this.

content-src/lib/selectLayoutRender.js
dmose: Truly hawt -- nice work!

lib/DiscoveryStreamFeed.jsm
```diff
@@ -163,25 +163,51 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+
+          newFeedsPromises.push(feedPromise);
+        }
+      });
+
+      return {newFeedsPromises, newFeeds};
+    }, initialData);
+  }
+
   async loadComponentFeeds(sendUpdate, isStartup) {
```
dmose: This function looks much cleaner now, which is fantastic!

lib/DiscoveryStreamFeed.jsm
```diff
@@ -163,25 +163,51 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
 
+  buildFeedPromises(layout, isStartup) {
+    const initialData = {
+      newFeedsPromises: [],
+      newFeeds: {},
+    };
+
+    return layout.filter(row => row && row.components).reduce(({newFeedsPromises, newFeeds}, row) => {
+      row.components.filter(component => component && component.feed).forEach(component => {
+        const {url} = component.feed;
```
dmose: I find buildFeedPromises very difficult to read and reason about, and I think in large part it's because five layers of abstraction are introduced in the above three lines (return, filter, reduce, filter, forEach).  This might want to broken up into multiple real functions, but at the very least, I'd suggest not doing nearly so much with manually in-lined arrow functions.

dmose: https://docs.codeclimate.com/docs/cognitive-complexity has a some discussion around this.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -163,25 +163,84 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
-            const {url} = component.feed;
-            newFeeds[url] = await this.getComponentFeed(url, isStartup);
-          }
-        }
-      }
 
-      await this.cache.set("feeds", newFeeds);
-      sendUpdate({type: at.DISCOVERY_STREAM_FEEDS_UPDATE, data: newFeeds});
+    if (!DiscoveryStream || !DiscoveryStream.layout) {
+      // istanbul ignore next
```
piatra: Could we add a test to cover this?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -163,25 +163,84 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
+  /**
+   * buildFeedPromise - Creates a promise for a single unique feed
+   *                    and populates newFeeds with the result.
+   * @param {Object} Has both newFeedsPromises (Array) and newFeeds (Object)
+   * @param {Boolean} isStartup We have different cache handling for startup.
+   * @returns {Function} A for each function, we return it so we can
```
piatra: I would instead mention what kind of data it transforms: `A function that works over components with valid 'feed' fields` or something like that.
Similar for the reduce cb `reduceFeedComponents`.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -163,25 +163,84 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
   }
 
+  /**
+   * buildFeedPromise - Creates a promise for a single unique feed
+   *                    and populates newFeeds with the result.
```
dmose: Maybe s/populates newsFeeds with the result/pushes the result onto the newsFeeds array/ for clarity...

dmose:
One suggested change nit, but with or without that, it looks good to me.

lib/DiscoveryStreamFeed.jsm
ScottDowne: That's better than what's currently there, as I was writing it a tweaked it even more, how does this look?

```
   * buildFeedPromise - Pushes a promise to newsFeedsPromises and
   *                    populates newFeeds with the promise result.
```

ScottDowne: Reversed the order too:
```
   * buildFeedPromise - Populates newFeeds with the promise result and
   *                    pushes a promise to newsFeedsPromises.
```

PR #4775, opened 2019-02-07
(nobug) - Refactor loadLayout so that _timestamp is now lastUpdated

piatra:
A little cleanup.

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -102,7 +102,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     const EXPIRATION_TIME = isStartup ? STARTUP_CACHE_EXPIRE_TIME : updateTimePerComponent[key];
     switch (key) {
       case "layout":
-        return (!layout || !(Date.now() - layout._timestamp < EXPIRATION_TIME));
+        return (!layout || !(Date.now() - layout.lastUpdated < EXPIRATION_TIME));
       case "spocs":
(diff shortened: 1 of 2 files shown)


piatra:
While refactoring I noticed that only for layout we read the updated timestamp from the endpoint response (so update time will be `_timestamp` that the payload sends) and for spocs and feeds we set it to `Date.now`. Is there a reason for this? Do we want to change this behavior? 
cc @k88hudson @ScottDowne

lib/DiscoveryStreamFeed.jsm
```diff
@@ -135,30 +135,31 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
     const cachedData = await this.cache.get() || {};
-    let {layout: layoutResponse} = cachedData;
+    let {layout} = cachedData;
     if (this.isExpired({cachedData, key: "layout", isStartup})) {
-      layoutResponse = await this.fetchFromEndpoint(this.config.layout_endpoint);
+      const layoutResponse = await this.fetchFromEndpoint(this.config.layout_endpoint);
       if (layoutResponse && layoutResponse.layout) {
-        layoutResponse._timestamp = Date.now();
-        await this.cache.set("layout", layoutResponse);
+        layout = {
```
ScottDowne: Would it be better to do:
```
layout = {
  ...layoutResponse,
  lastUpdated: Date.now(),
};
```

k88hudson:
This is fine, I would just consider the possible bug vectors this might introduce for future uplifts

PR #4778, opened 2019-02-08
Bug 1500061 - Tighen CSP for new tab

k88hudson:
Note that the rest of snippets will be removed in a future patch.

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -97,7 +97,7 @@ for (const src of ${JSON.stringify(scripts, null, 2)}) {
 <html lang="${options.locale}" dir="${options.direction}">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; object-src 'none'; script-src resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
     <title>${options.strings.newtab_page_title}</title>
     <link rel="icon" type="image/png" href="chrome://branding/content/icon32.png"/>
     <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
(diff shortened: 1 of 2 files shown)


k88hudson:
@freddyb would you mind taking a look at this? So far I've only removed `unsafe-inline` for scripts

bin/render-activity-stream-html.js
```diff
@@ -97,7 +97,7 @@ for (const src of ${JSON.stringify(scripts, null, 2)}) {
 <html lang="${options.locale}" dir="${options.direction}">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
```
mozfreddyb: Looks good!  :+1: 

The adjacent `connect-src` surprises me a bit. Do you actually use fetch, XHR, EventSource, sendBeacon,   `<a ping>` or Websockets?
To clarify, I found it just surprising. Not concerning :)

Mardak:
Oh, and there was a comment from ckerschb about adding `object-src 'none'` https://bugzilla.mozilla.org/show_bug.cgi?id=1500061#c3

bin/render-activity-stream-html.js
k88hudson: Ah good catch – we used to use `fetch` from old snippets, but it's no longer needed. I supposed we can remove that

k88hudson: Nevermind, still needed for one of the form submit actions in a template.

Mardak:
Reverted this https://github.com/mozilla/activity-stream/commit/a3fe11f0f32415c98fa7a31b996735b4d580bfb1 for test failures when trying to export:
```
Console message: [JavaScript Error: "Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”)." {file: "about:newtab" line: 16}]
```
https://treeherder.mozilla.org/logviewer.html#/jobs?job_id=228060887&repo=try&lineNumber=1980-1987

Automation uses artifact builds. Try was failing on asan.

PR #4779, opened 2019-02-08
Bug 1526387 - CFR Addon Recommendations call remote API before clicking "Install"

rlr:


k88hudson:
Just got out of a meeting, taking a look at this now

lib/CFRPageActions.jsm
```diff
@@ -176,6 +176,25 @@ class PageAction {
...
 
+  /**
+   * Fetch the URL to the latest add-on xpi so the recommendation can download it.
+   * @param id          The add-on ID
+   * @return            A string for the URL that was fetched
+   */
+  async _fetchLatestAddonVersion(id) {
+    let url = null;
+    try {
+      const response = await fetch(`${ADDONS_API_URL}/${id}/`);
```
k88hudson: Can you add `{credentials: omit}` just in case?

k88hudson: Also, we need to file a follow-up to handle the case of this failing a little better than just showing nothing. For now , I think this is fine.

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -468,12 +471,12 @@ describe("CFRPageActions", () => {
         });
       });
       it("should set the main action correctly", async () => {
+        sinon.stub(pageAction, "_fetchLatestAddonVersion").returns(Promise.resolve("latest-addon.xpi"));
```
k88hudson: you can actually do `sinon.stub(pageAction, "_fetchLatestAddonVersion").resolves("latest-addon.xpi")` here

PR #4781, opened 2019-02-11
Bug 1525366 - Show user preference for SPOCs if Discoverystream is on

piatra:
The issue I ran into while implementing this: 

tldr 2 checkboxes linking to the same preference (`showSponsored`) doesn't work.

Currently in `about:preferences` we hide the default prefs (via CSS) and append the experiment preferences. In this bug we add a checkbox (for the experiment prefs) to toggle spocs. This checkbox links to the same pref as TopStories section (`.showSponsored`) essentially having 2 checkboxes linking to the same preference, [which is not allowed](https://searchfox.org/mozilla-central/rev/5e3bffe964110b8d1885b4236b8abd5c94d8f609/toolkit/content/preferencesBindings.js#35).

And there doesn't seem to be a way to unlink a binding between an element and a pref so my solution was to just reload the page if the user leaves the experiment. 
I could also reach into `Preferences._all` and remove it and the listener but at that point I would be implementing the missing `Preferences.remove` method.
I'm open to suggestions to improve this.

### DIFF
--- a/lib/AboutPreferences.jsm
+++ b/lib/AboutPreferences.jsm
@@ -286,8 +286,19 @@ this.AboutPreferences = class AboutPreferences {
       createAppend("label", discoveryGroup)
         .appendChild(document.createElementNS(HTML_NS, "h2"))
         .textContent = formatString("prefs_content_discovery_header");
-      createAppend("description", discoveryGroup)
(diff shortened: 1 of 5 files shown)


piatra:
QA steps in https://bugzilla.mozilla.org/show_bug.cgi?id=1525366 user story.

PR #4782, opened 2019-02-11
Fixing li

ScottDowne:


Mardak:
This PR is missing a bug number so wasn't automatically linked to https://bugzilla.mozilla.org/show_bug.cgi?id=1526861 and similarly the commit message was missing the bug number so it didn't show up when merged.

PR #4783, opened 2019-02-11
Bug 1526452 - Stop exporting sourcemaps to MC, only support during startmc

k88hudson:
This patch removes external sourcemaps and retains them only during the startmc, as inline sourcemaps. To test this, try using devtools with `npm run startmc` (you should see line numbers, modules) and `npm run buildmc` (you should not see line numbers, modules).

cc @Mardak 

Notes:
* We don't currently have a clean up task that would remove existing files, we could consider adding that
* An additional patch will be needed to remove the existing files on the MC side


Mardak:
> We don't currently have a clean up task that would remove existing files, we could consider adding that

Different from `npm run prebuildmc` ?

Mardak:
Oh, I see. Existing files from activity-stream side.

Mardak:
> An additional patch will be needed to remove the existing files on the MC side

This should just happen "automatically" as part of the normal export process with `buildmc`:
```
mozilla-central$ git status
	modified:   browser/components/newtab/css/activity-stream-linux.css
	deleted:    browser/components/newtab/css/activity-stream-linux.css.map
	modified:   browser/components/newtab/css/activity-stream-mac.css
	deleted:    browser/components/newtab/css/activity-stream-mac.css.map
	modified:   browser/components/newtab/css/activity-stream-windows.css
	deleted:    browser/components/newtab/css/activity-stream-windows.css.map
	modified:   browser/components/newtab/data/content/activity-stream.bundle.js
	deleted:    browser/components/newtab/data/content/activity-stream.bundle.js.map
```

dmose:
FWIW, I would expect inlining the source maps will change the memory footprint (and maybe some perf characteristics) of the dev build.  I suspect that doesn't matter, but it might.

dmose:
> > We don't currently have a clean up task that would remove existing files, we could consider adding that
> 
> Different from `npm run prebuildmc` ?
> 
> Edit: Oh, I see. Existing files from activity-stream side to prevent them from being copied over with buildmc.

If we don't do this, I think we're going to at least want to document in the export checklist that everyone who ever exports probably needs to do this once to inadvertently landing old maps.

k88hudson:
@dmose Good to note, I'll add that to the comments above the `startmc` task definition; presumably we can just use mozilla-central directly or `buildmc` if we're doing any perf/profiling work.

dmose:
Maybe we could use a husky hook to prevent the inadvertent export.

dmose:
@k88hudson As far as the perf thing, I think this probably wants to also be documented somewhere in our markdown docs about developing.

k88hudson:
@dmose Updated with comment documentation

webpack.prerender.config.js
```diff
@@ -9,7 +9,7 @@ NOTE: This file is generated by webpack from ${srcPath}
 using the buildmc:html npm task.
 `;
 
-module.exports = Object.assign({}, webpackConfig, {
+module.exports = Object.assign({}, webpackConfig(), {
```
dmose: `webpack.aboutlibrary.config.js` will need this change too.

webpack.system-addon.config.js
```diff
@@ -5,14 +5,14 @@ const absolute = relPath => path.join(__dirname, relPath);
...
-module.exports = {
+module.exports = (env = {}) => ({
   mode: "none",
   entry: absolute("content-src/activity-stream.jsx"),
   output: {
     path: absolute("data/content"),
     filename: "activity-stream.bundle.js",
   },
-  devtool: "source-map",
+  devtool: env.development ? "inline-source-map" : false,
```
dmose: https://webpack.js.org/configuration/devtool/ seems to suggest using eval-source-map for development rather than inline-source-map.  What motivated the choice of inline-source-map?  If we do want to keep inline-source-map, I'd suggest adding a comment explaining the motivation.

k88hudson: I am pretty sure `eval` is prevented by our CSP

dmose: I'm mostly suggesting this because from that chart, it appears to slow down build times substantially, and our build times are already a bit slow.

We could use a different CSP for development.

That said, I don't have a crazy strong opinion here.  If it truly slows development enough, we can always switch it in the future.  So it's really up to you.

k88hudson: Our existing implementation does regular production source maps, which are also slow (we never used eval) so I would consider that an enhancement

k88hudson:
I have updated the documentation for exports with a notice (see https://docs.google.com/document/d/1olEyxOfoF1C9qi-x8jHCZgynY4czXtmJCx2eKm8Wayc/edit#heading=h.wzgy79ypf8tl) and will sent out a quick note via email / slack when this lands.

As we discussed, I will not implement automation to check for source maps accidentally being included and a bug will be filed only if we see an error show up in a Phabricator patch.

webpack.system-addon.config.js
dmose: OK, if you want to do that later, how about at least leaving `XXX switch to eval-source-map for faster builds` in the file, both as a reminder, and to give the reader a clue that there's no intrinsic reason we need to leave this as inline-source-map.

Mardak: The reason is that jar.mn doesn't have an entry to build/package the map file anymore. I suppose that should be in the comment for why it's inline…

dmose: @Mardak that would also be true of eval-source-map, so I'm not sure I understand the point you're making.

Mardak:
From my quick testing, it doesn't seem like much of a startmc:webpack regression with this PR:

```
PR/inline
Time: 1226ms
Time: 1265ms
Time: 1362ms
Time: 1396ms
Time: 1479ms

before/separate
Time: 1137ms
Time: 1180ms
Time: 1332ms
Time: 1344ms
Time: 1444ms

eval
Time: 1285ms
Time: 1290ms
Time: 1350ms
Time: 1382ms
Time: 1440ms
```

Compared to before, maybe 30-100ms regression, so 2%-7% regression. Although maybe some people switch between their editor and open firefox builds faster than me. ;)

PR #4784, opened 2019-02-12
Bug 1518258 - Use pref based consumer key in discovery stream layout …

punamdahiya:
…endpoint

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -4,6 +4,7 @@
 "use strict";
 
 const {XPCOMUtils} = ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
+const {Services} = ChromeUtils.import("resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
 ChromeUtils.defineModuleGetter(this, "perfService", "resource://activity-stream/common/PerfService.jsm");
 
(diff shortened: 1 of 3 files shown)


lib/DiscoveryStreamFeed.jsm
```diff
@@ -32,12 +33,25 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
   get config() {
     if (this._prefCache.config) {
       return this._prefCache.config;
     }
     try {
       this._prefCache.config = JSON.parse(this.store.getState().Prefs.values[PREF_CONFIG]);
+      const layoutUrl = this._prefCache.config.layout_endpoint;
+      const apiKeyPref = this._prefCache.config.api_key_pref;
+      if (layoutUrl && apiKeyPref) {
+        const apiKey = Services.prefs.getCharPref(apiKeyPref);
```
k88hudson: Can you pass a second param like `""` to this just in case the pref is not defined? I think we don't want to throw in this case.

punamdahiya: If we are not not throwing error for empty apiKey , is it ok to return layout_endpoint url with consumer_key as ""?

punamdahiya: Updated with second optional param as "" to avoid throwing an exception while trying to retrieve pref

k88hudson: If the url contains `$apiKey` your `finalLayoutEndpoint` function wills till throw if the result is `""`, I think that will be a more descriptive error than just the pref not existing

lib/DiscoveryStreamFeed.jsm
```diff
@@ -32,12 +33,25 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     this._prefCache = {};
   }
 
+  finalLayoutEndpoint(url, apiKey) {
+    if (url.includes("$apiKey") && !apiKey) {
+      throw new Error(`Layout Endpoint - An API key was specified but none configured: ${url}`);
```
k88hudson: I just realized this is going to be caught by the outer try/catch... maybe make sure to log it or change the way the flow works so this will show up?

punamdahiya: Good catch, updated to log error in catch block

punamdahiya:
> Looks good, having some test steps to verify the expected error states would be good to have as well.

Done, updated to cover some more edge cases, let me know if there are more that will be good to capture in tests. Thanks!

PR #4787, opened 2019-02-12
(no bug) - Fix taskcluster by actually using packaged dist files for referenced test

Mardak:
Checking taskcluster first… before r? @piatra or @k88hudson


Mardak:
Most likely caused by 
Bug 1517510 Update browser_all_files_referenced.js to check all shipped fluent files are referenced
https://hg.mozilla.org/mozilla-central/rev/777d4b26e0b1
https://bugzilla.mozilla.org/show_bug.cgi?id=1517510

Still not entirely sure why lots of files fail even with `mach package` but in the past we just grepped for the ones we cared about…

Mardak:
Hold on. Reading the bug shows we should be running ./mach with `--appname=dist`.. looks good locally. Repushing to taskcluster…

PR #4790, opened 2019-02-13
Bug 1527702 - Rotate based on impressions for recs.

ScottDowne:


ScottDowne:
To test.

Basic tests:
1. Use dev-test-all, view a bunch of stories, check the new pref for timestamps, should have a bunch of stuff in it.
2. Change your system clock + 90 minutes. Look at new tab again, stories should be a bit different.

Check that it clears test:
1. Use dev-test-feeds.
2. View everything on the page (scroll from top to bottom)
3. Check pref for rec impressions, should be big.
4. Now load basic. (shouldn't have to view the page, pull in the feed should be enough)
5. You should see the impression pref be reduced quite a bit. It should only retain any impressions that you've seen that both layouts share in a returned feed.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -412,6 +415,26 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+  // impressions i.e. have been displayed for longer than recsExpireTime.
+  rotate(feedResponse) {
+    const {recommendations} = feedResponse;
+
+    const maxImpressionAge = Math.max(feedResponse.settings.recsExpireTime * 1000 || DEFAULT_RECS_EXPIRE_TIME, DEFAULT_RECS_EXPIRE_TIME);
+    const impressions = this.readImpressionsPref(PREF_REC_IMPRESSIONS);
+    const expired = [];
+    const active = [];
+    for (const item of recommendations) {
+      if (impressions[item.id] && Date.now() - impressions[item.id] >= maxImpressionAge) {
```
Mardak: What's the expected behavior? If a story was most recently seen more than 1 hour ago, expire it? Or should it be story was first seen 1 hour ago, expire it?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -537,13 +560,30 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     this.writeImpressionsPref(PREF_SPOC_IMPRESSIONS, impressions);
   }
 
+  recordTopRecImpressions(recId) {
+    let impressions = this.readImpressionsPref(PREF_REC_IMPRESSIONS);
+    impressions = {...impressions, [recId]: Date.now()};
```
Mardak: If we want to expire based on first seen time, we should check
```js
if (!impressions[recId]) {
  impressions[recId] = Date.now();
  writeImpressions()
}
```
Not sure if we really need to spread for a new object in this situation, but I see that's how spocs work.

lib/DiscoveryStreamFeed.jsm
ScottDowne: Checking, you may have found a bug.

lib/DiscoveryStreamFeed.jsm
ScottDowne: I think you're right but something's up with github making it hard to confirm, give me a sec.

lib/DiscoveryStreamFeed.jsm
ScottDowne: It should be first seen more than x time again, expire it.

lib/DiscoveryStreamFeed.jsm
ScottDowne: you're right, this is a bug, it should filter out items already in the impressions and not overwrite them.

k88hudson:
You probably want to be clearing `browser.newtabpage.activity-stream.discoverystream.rec.impressions` when feed endpoints change – I'm not seeing that happen right now

k88hudson:
One major issue I am seeing is that right now, updates to `discoverystream.rec.impressions` pref are being pushed to all open tabs (since it is under the activity-stream branch) and causing extra messages and re-renders. For spocs, this isn't too big of a deal, but now that we are sending impressions for every item, it seems pretty wasteful (especially since this list isn't needed in content at all);

I would recommend possibly: 
* Using `Services.prefs.setStringPref` to set the pref directly;
* Using a prefname not under the activity stream branch (maybe discoverystream.impressions.recs` or something?) so our pref service doesn't watch it.

Any other ideas @Mardak?

lib/ActivityStream.jsm
```diff
@@ -237,6 +237,10 @@ const PREFS_CONFIG = new Map([
     title: "Track spoc impressions",
     value: "{}",
   }],
+  ["discoverystream.rec.impressions", {
+    title: "Track rec impressions",
+    value: "{}",
```
Mardak: #4743 says it doesn't do batching yet, so we could try to implement batching here although it's not entirely clear how to debounce appropriately.

Alternatively, we could add an attribute here along the lines `updateOnlyMain: true` and additionally check that before broadcasting or just plain action:
https://github.com/mozilla/activity-stream/blob/8f09010f3034beb5f988c804719e2b19760bc83f/lib/PrefsFeed.jsm#L49-L50

Mardak: Longer term we might want to flip the condition to `shouldBroadcastUpdates` only for prefs that content cares about as most prefs we don't care but probably not the right time to audit… @k88hudson ?

Mardak: Quick hack seems to avoid content reducer from triggering:
```diff
diff --git a/lib/PrefsFeed.jsm b/lib/PrefsFeed.jsm
   onPrefChanged(name, value) {
     if (this._prefMap.has(name)) {
-      this.store.dispatch(ac.BroadcastToContent({type: at.PREF_CHANGED, data: {name, value}}));
+      this.store.dispatch(ac[name == "discoverystream.rec.impressions" ? "OnlyToMain" : "BroadcastToContent"]({type: at.PREF_CHANGED, data: {name, value}}));
```

ScottDowne: Is this above hack more desirable than using `Services.prefs.setStringPref`?

k88hudson: That seems fine to me, although possibly more risky for an uplift (i'd probably prefer to see it as a non-66 refactor)

lib/DiscoveryStreamFeed.jsm
```diff
@@ -20,8 +20,8 @@ const MAX_LIFETIME_CAP = 500; // Guard against misconfiguration on the server
 const PREF_CONFIG = "discoverystream.config";
 const PREF_OPT_OUT = "discoverystream.optOut.0";
 const PREF_SHOW_SPONSORED = "showSponsored";
-const PREF_SPOC_IMPRESSIONS = "discoverystream.spoc.impressions";
-const PREF_REC_IMPRESSIONS = "discoverystream.rec.impressions";
+const PREF_SPOC_IMPRESSIONS = "browser.discoverystream.spoc.impressions";
+const PREF_REC_IMPRESSIONS = "browser.discoverystream.spoc.impressions";
```
Mardak: should be .rec. instead of .spoc. for the latter

lib/DiscoveryStreamFeed.jsm
```diff
@@ -537,20 +566,42 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+    const activeStories = Object.keys(newFeeds).reduce((accumulator, currentValue) => {
+      const {recommendations} = newFeeds[currentValue].data;
+      return accumulator.concat(recommendations.map(i => `${i.id}`));
+    }, []);
+    this.cleanUpImpressionPref(id => !activeStories.includes(id), PREF_REC_IMPRESSIONS);
+  }
+
   writeImpressionsPref(pref, impressions) {
-    this.store.dispatch(ac.SetPref(pref, JSON.stringify(impressions)));
+    Services.prefs.setStringPref(pref, JSON.stringify(impressions));
```
Mardak: Need to define Services

Mardak: Actually, rebasing onto latest master has Services already

lib/ActivityStream.jsm
Mardak: If we do end up refactoring after 66, then having a different pref name outside of browser.newtabpage.activity-stream will be problematic.. or at least needs some migration.

Mardak: If we keep the pref under the existing branch and want to avoid the broadcasts, we can switch to Services.prefs AND remove the entries from ActivityStream.jsm

Mardak:
Here's a patch to maintain backwards/forwards compatibility with the pref if we refactor outside of 66:
```diff
diff --git a/lib/ActivityStream.jsm b/lib/ActivityStream.jsm
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -234,14 +234,6 @@ const PREFS_CONFIG = new Map([
     title: "Opt out of new layout v0",
     value: false,
   }],
-  ["discoverystream.spoc.impressions", {
-    title: "Track spoc impressions",
-    value: "{}",
-  }],
-  ["discoverystream.rec.impressions", {
-    title: "Track rec impressions",
-    value: "{}",
-  }],
   ["darkModeMessage", {
     title: "Boolean flag that decides whether to show the dark Mode message or not.",
     value: IS_NIGHTLY_OR_UNBRANDED_BUILD,
diff --git a/lib/DiscoveryStreamFeed.jsm b/lib/DiscoveryStreamFeed.jsm
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -21,8 +21,9 @@ const MAX_LIFETIME_CAP = 500; // Guard against misconfiguration on the server
 const PREF_CONFIG = "discoverystream.config";
 const PREF_OPT_OUT = "discoverystream.optOut.0";
 const PREF_SHOW_SPONSORED = "showSponsored";
-const PREF_SPOC_IMPRESSIONS = "browser.discoverystream.spoc.impressions";
-const PREF_REC_IMPRESSIONS = "browser.discoverystream.spoc.impressions";
+// For now use full pref with Services.prefs to avoid broadcasting to content
+const PREF_SPOC_IMPRESSIONS = "browser.newtabpage.activity-stream.discoverystream.spoc.impressions";
+const PREF_REC_IMPRESSIONS = "browser.newtabpage.activity-stream.discoverystream.rec.impressions";
 
 this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   constructor() {
```

ScottDowne:
Forgot to remove from ActivityStream.jsm, doing that now.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -622,6 +670,11 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       case at.DISCOVERY_STREAM_OPT_OUT:
         this.store.dispatch(ac.SetPref(PREF_OPT_OUT, true));
         break;
+      case at.DISCOVERY_STREAM_IMPRESSION_STATS:
+        if (action.data.tiles && action.data.tiles[0] && action.data.tiles[0].id) {
+          this.recordTopRecImpressions(action.data.tiles[0].id);
```
Mardak: Forgot to update tests based on this new API. Fixing with https://github.com/mozilla/activity-stream/pull/4799

lib/PrefsFeed.jsm
```diff
@@ -46,8 +46,9 @@ this.PrefsFeed = class PrefsFeed {
   }
 
   onPrefChanged(name, value) {
-    if (this._prefMap.has(name)) {
-      this.store.dispatch(ac.BroadcastToContent({type: at.PREF_CHANGED, data: {name, value}}));
+    const prefItem = this._prefMap.get(name);
+    if (prefItem) {
```
rlr: this broke pref changes to prefs that were set to falsy values. Fix in pull #4803

PR #4792, opened 2019-02-14
Bug 1527980 - Add devtools for resetting and picking a layout variant

k88hudson:
screen shot 2019-02-14 at 1 46 03 pm

Testing enabled / opt-out:
1. Turn on new tab devtools: Set `browser.newtabpage.activity-stream.asrouter.devtoolsEnabled` to `true`.
2. Navigate to about:newtab#devtools-ds
3. Make sure the enabled checkbox is checked. Ensure discovery stream is visible on another tab.
4. Go to about:preferences#home and click the "Turn off discovery stream" button.
5. On about:newtab#devtools-ds, ensure that the enabled checkbox is now unchecked and you see the message "(Note: User has opted-out. Check this box to reset)" next to the checkbox.
6. Check the enabled box. Ensure that discovery stream now shows up again on a new tab page and the opt-out message has disappeared next to the checkbox.

Testing layout variants.
1. On about:newtab#devtools-ds, ensure discovery stream is enabled.
2. Check the "basic" radio box. Ensure you see a 3-card layout on new tab.
3. Check the "dev-test-all" radio box. Ensure you see a long layout with many elements.
4. On about:config, try changing the `browser.newtabpage.activity-stream.discoverystream.config` pref to include `layout_variant=basic` in the endpoint.
5. Check that on about:newtab#devtools-ds the basic radio checkbox is now checked.

### DIFF
--- a/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -25,10 +25,21 @@ function relativeTime(timestamp) {
   return new Date(timestamp).toLocaleString();
 }
 
+const OPT_OUT_PREF = "discoverystream.optOut.0";
+const LAYOUT_VARIANTS = {
+  "basic": "Basic default layout (on by default in nightly)",
+  "dev-test-all": "A little bit of everything. Good layout for testing all components",
+  "dev-test-feeds": "Stress testing for slow feeds",
+};
 class DiscoveryStreamAdmin extends React.PureComponent {
   constructor(props) {
     super(props);
     this.onEnableToggle = this.onEnableToggle.bind(this);
+    this.changeEndpointVariant = this.changeEndpointVariant.bind(this);
+  }
+
+  get isOptedOut() {
+    return this.props.otherPrefs[OPT_OUT_PREF];
   }
 
   setConfigValue(name, value) {
@@ -37,6 +48,13 @@ class DiscoveryStreamAdmin extends React.PureComponent {
 
   onEnableToggle(event) {
     this.setConfigValue("enabled", event.target.checked);
+  }
+
+  changeEndpointVariant(event) {
+    const endpoint = this.props.state.config.layout_endpoint;
+    if (endpoint) {
+      this.setConfigValue("layout_endpoint", endpoint.replace(/layout_variant=.+/, `layout_variant=${event.target.value}`));
+    }
   }
 
   renderComponent(width, component) {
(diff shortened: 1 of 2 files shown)


content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -75,13 +98,27 @@ class DiscoveryStreamAdmin extends React.PureComponent {
...
+
+      <h3>Endpoint variant</h3>
+      <p>You can also change this manually by changing this pref: <code>browser.newtabpage.activity-stream.discoverystream.config</code></p>
+      <table style={config.enabled ? null : {opacity: 0.5}}><tbody>
+        {Object.keys(LAYOUT_VARIANTS).map(id => (<Row key={id}>
+          <td className="min"><input type="radio" value={id} checked={this.isCurrentVariant(id)} onChange={this.changeEndpointVariant} /></td>
+          <td className="min">{id}</td>
+          <td>{LAYOUT_VARIANTS[id]}</td>
+        </Row>))}
+      </tbody></table>
```
Mardak: Might be useful to have a "full url" textbox to set just the endpoint directly. Currently if it's none of the explicit `LAYOUT_VARIANTS`, nothing is checked. And if the url doesn't match `layout_variant=`, clicking the radio buttons do nothing, e.g., if it's set to `https://gist.githubusercontent.com/punamdahiya/44f78e2f338cbf95f5c28d31faed4639/raw/1c61974ca8ebd02431512d9425031e3a9c8e04f4/test-message-feed.json`

Having the textbox is slightly clearer what's going on with the gist endpoint. But we don't need to spend too much (/ any?) effort here as there's always about:config…

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -25,10 +25,20 @@ function relativeTime(timestamp) {
   return new Date(timestamp).toLocaleString();
 }
 
+const OPT_OUT_PREF = "discoverystream.optOut.0";
+const LAYOUT_VARIANTS = {
+  "basic": "Basic default layout (on by default in nightly)",
+  "dev-test-all": "A little bit of everything. Good layout for testing all components",
```
Mardak: Not sure if we should add some more here, as it would need to get cleaned up if we don't want something anymore. `dev-test-feeds` might be a good candidate for now. The other ones like `survey-a` through `survey-d` seem less likely to include.

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
k88hudson: I was thinking it might be a good idea to ask security before doing that, but if it's just the end of the URL it's probably fine. I won't do it right now just cause i dont have much time

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
k88hudson: Sure, I can add dev-test-feeds

PR #4794, opened 2019-02-14
Bug 1527347 - Fixing breakpoints for search too

ScottDowne:


content-src/components/Base/_Base.scss
```diff
@@ -69,6 +69,10 @@ main {
     // and Activity Stream breakpoints do some wonky things.
     width: 1042px;
   }
+
+  .search-wrapper .search-inner-wrapper {
+    width: 736px;
```
k88hudson: * Please use `$searchbar-width-large`
* This should probably be in `_Search.scss`, right?
* This doesn't make sense for `fixed-search` (i.e. when you scroll), we still want break points there

PR #4795, opened 2019-02-14
Bug 1527437 [Cards][Lists] - Align spocs items to bottom of container

punamdahiya:


punamdahiya:
@Mardak @gvn please review. Thanks!

Mardak:
This should have already been fixed for cards with #4737. @ScottDowne did something change to remove that functionality but we want it back again?

Mardak:
Ah, bisecting the regression points to #4774 which probably has the same underlying issue as the fixes for #4782 and #4785. Although the latter one #4785 changes the parent order again and didn't seem to fix…

Mardak:
We probably don't even need to do this as a `React.Fragment` if we make users of `<original stuff><SpocIntersectionObserver />` instead of `<SpocIntersectionObserver><original stuff></>` although that would touch a lot of lines and probably should be avoided for uplifts. @k88hudson ?

Mardak:
Oh… thecount pointed out that @ncloudioj is removing `SpocIntersectionObserver` with #4743 so this PR probably should look for fixes that don't depend on stuff in there.

punamdahiya:
> Oh… thecount pointed out that @ncloudioj is removing `SpocIntersectionObserver` with #4743 so this PR probably should look for fixes that don't depend on stuff in there.

I agree if we are planning to remove SpocIntersectionObserver, we should not have a className dependency in there. I believe  rendering as below is a better approach. Will update PR. Thanks!
```
  <li className="ds-list-item">
    <original stuff>
    <SpocIntersectionObserver />
  </li>
```

gvn:
@punamdahiya Can you add QA steps?

punamdahiya:
@Mardak Updated PR with review feedback, thanks!

@gvn To view spoc item in list and hero components use layout_endpoint below "layout_endpoint":"https://gist.githubusercontent.com/punamdahiya/4852948f6b24e1a82f4cc2f107ddc234/raw/7e54aa261eeeb718b94a55965b1da28348df93dc/feed.json"

To verify Card layout switch endpoint from basic to dev-test-all

Thanks!

content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -162,6 +162,7 @@ $item-line-height: 20;
 
     display: flex;
     justify-content: space-between;
+    height: 100%;
```
gvn: Is this necessary?

punamdahiya: yes, without this link doesn't take full height from parent and fails to bottom align source.

PR #4803, opened 2019-02-18
Bug 1528494 - Fix pref check in PrefsFeed.onPrefChanged()

rlr:


lib/PrefsFeed.jsm
```diff
@@ -46,9 +46,8 @@ this.PrefsFeed = class PrefsFeed {
   }
 
   onPrefChanged(name, value) {
-    const prefItem = this._prefMap.get(name);
-    if (prefItem) {
-      this.store.dispatch(ac[prefItem.skipBroadcast ? "OnlyToMain" : "BroadcastToContent"]({type: at.PREF_CHANGED, data: {name, value}}));
+    if (this._prefMap.has(name)) {
```
Mardak: How was this breaking? The PREFS_CONFIG we pass in should be a mapping from strings to objects, so even if a pref wanted a false value, the map get would result in `{ title: "foo", value: false }` ???

Is this something we need to uplift too?

rlr: oh hmm. interesting... but the tests fail without this change and pass when i apply it... 😕

Mardak: Oh.. I see. Specially for prefs defined in firefox.js, the prefMap gets a plain value instead of an object.

https://github.com/mozilla/activity-stream/blob/b4ec95c93d493ffd40d8c0c6778fe1c934b0b0e1/lib/PrefsFeed.jsm#L89-L101

rlr:
try run of this branch simulating beta: https://treeherder.mozilla.org/#/jobs?repo=try&revision=408f10750e0683be6763af8b468d1c3f8942e5d3

Mardak:
FYI, in this case looks like the test fails without needing to simulate all of being beta ;) Just setting the default pref value in the NIGHTLY case to false gets it to fail without your PR.

PR #4805, opened 2019-02-19
 Bug 1508744 - Instrument pin tab event 

piatra:
A little different from our other events this is triggered from the parent process and is not associated with a `session_id` is that an issue?

### DIFF
--- a/docs/v2-system-addon/data_events.md
+++ b/docs/v2-system-addon/data_events.md
@@ -362,6 +362,23 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as 
 }
 ```
 
+#### Pinning a tab
+
+```js
+{
+  "event": "TabPinned",
+  "source": "TAB_CONTEXT_MENU",
+
+  // Basic metadata
+  "action": "activity_stream_user_event",
+  "client_id": "aabaace5-35f4-7345-a28e-5502147dc93c",
+  "version": "67.0a1",
+  "addon_version": "20190218094427",
+  "locale": "en-US",
+  "user_prefs": 59,
+}
+```
+
 #### Adding or editing a new TopSite
 
 ```js
--- a/docs/v2-system-addon/data_events.md
+++ b/docs/v2-system-addon/data_events.md
@@ -368,6 +368,7 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as 
 {
   "event": "TABPINNED",
   "source": "TAB_CONTEXT_MENU",
+  "value": "{\"max_concurrent_pinned_tabs\":2}",
 
   // Basic metadata
   "action": "activity_stream_user_event",
--- a/docs/v2-system-addon/data_events.md
+++ b/docs/v2-system-addon/data_events.md
@@ -366,7 +366,7 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as 
 
 ```js
 {
-  "event": "TabPinned",
+  "event": "TABPINNED",
   "source": "TAB_CONTEXT_MENU",
 
   // Basic metadata
--- a/docs/v2-system-addon/data_events.md
+++ b/docs/v2-system-addon/data_events.md
@@ -368,7 +368,6 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as 
 {
   "event": "TABPINNED",
   "source": "TAB_CONTEXT_MENU",
-  "value": "{\"max_concurrent_pinned_tabs\":2}",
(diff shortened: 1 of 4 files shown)


ncloudioj:
> A little different from our other events this is triggered from the parent process and is not associated with a `session_id` is that an issue?

`session_id` is a required field in the event ping, though you can set it to `n/a` if it's not applicable in this case.

lib/TelemetryFeed.jsm
```diff
@@ -63,10 +68,49 @@ this.TelemetryFeed = class TelemetryFeed {
...
+    win.removeEventListener(TAB_PINNED_EVENT, this.handleEvent);
+  }
+
+  _addWindowListeners(win) {
+    win.addEventListener(DOMWINDOW_UNLOAD_TOPIC, this.handleEvent);
+    win.addEventListener(TAB_PINNED_EVENT, this.handleEvent);
+  }
+
+  countPinnedTab(source = "TAB_CONTEXT_MENU") {
+    this.onAction(ac.UserEvent({
```
ncloudioj: This will create a session implicitly. You can reference the `sendPageTakeoverData()` in this module to see how to send an event ping via `createPing()` with your own fields.

docs/v2-system-addon/data_events.md
```diff
@@ -362,6 +362,25 @@ A user event ping includes some basic metadata (tab id, addon version, etc.) as
 }
 ```
 
+#### Pinning a tab
+
+```js
+{
+  "event": "TABPINNED",
+  "source": "TAB_CONTEXT_MENU",
```
ncloudioj: Let's also add the `value` here?

lib/TelemetryFeed.jsm
```diff
@@ -63,10 +69,73 @@ this.TelemetryFeed = class TelemetryFeed {
...
+    this.handleEvent = this.handleEvent.bind(this);
   }
 
   init() {
     Services.obs.addObserver(this.browserOpenNewtabStart, "browser-open-newtab-start");
+    // Add pin tab event listeners on future windows
+    Services.obs.addObserver(this._addWindowListeners, DOMWINDOW_OPENED_TOPIC);
+    // Listen for pin tab events on all open windows
+    for (let win of Services.wm.getEnumerator("navigator:browser")) {
+      if (win.closed || PrivateBrowsingUtils.isWindowPrivate(win)) {
```
ncloudioj: Hmm, this ping was still sent when I pinned tabs in the private window.

lib/TelemetryFeed.jsm
```diff
@@ -63,10 +69,73 @@ this.TelemetryFeed = class TelemetryFeed {
...
+        break;
+    }
+  }
+
+  _removeWindowListeners(win) {
+    win.removeEventListener(DOMWINDOW_UNLOAD_TOPIC, this.handleEvent);
+    win.removeEventListener(TAB_PINNED_EVENT, this.handleEvent);
+  }
+
+  _addWindowListeners(win) {
```
ncloudioj: Perhaps move the private window check here? And only call `removeEventListener` if the window is private in the `_removeWindowListeners()`.

ncloudioj: That being said, it'll be nice to have a unit test for this.

rlr:
I wonder if we should track "TabUnpinned" too?

lib/TelemetryFeed.jsm
piatra: It's because I was checking for private windows only on currently open windows and then

```
Services.obs.addObserver(this._addWindowListeners, DOMWINDOW_OPENED_TOPIC);
```
I was adding PinTab event listeners for all new windows (private or not).

lib/TelemetryFeed.jsm
piatra: I ended up adding the events for all windows because of an issue I noticed while testing where `isWindowPrivate` returns a value based on the type of window you previously had focused.
STR
Have 2 browser windows open: one private, one regular.
Set a breakpoint in this function.
Use the keyboard shortcut or menu option to create a new _non-private_ window.
Result:
* when focus is set on the non-private window `isWindowPrivate` returns `false` for the new window as well
* when focus is set on the private window `isWindowPrivate` returns `true` for the new window as well

I assume the new window inherits attributes from the previous when it is initialized and those get changed later.
Even more interesting: there is also CMD+Shift+N to create a new window and it does not have this issue. But the shortcut shown in the browser menu is CMD+N.

PR #4806, opened 2019-02-19
Bug 1529045 – Use CDN for DiscoveryStream Layouts

gvn:


gvn:
Just needs 1 reviewer IMO, so whoever has time!

PR #4807, opened 2019-02-20
Bug 1529007 - React 16.8.3 upgrade

dmose:
This patch adds a script to make React upgrades less error-prone, upgrades React to 16.8.3, and adds in the eslint plugin for React Hooks rules recommended in the React 16.8 blog post.

Basic testing suggests that things seem to be working, and all tests are passing locally.

There are some other packages that will take advantage of or work slightly better with React 16.8.3 after they get upgraded (eslint-plugin-react, enzyme, react-redux), but I think those can be upgraded separately. 

I'll add a new attachment to the bug for review of update the license.html in toolkit/.


dmose:
For landing, my plan is to squash the two "Update" commits together and keep the other commits intact.

dmose:
Heh, React 16.8.3 just got released, so I'm gonna update a few things.

.eslintrc.js
```diff
@@ -15,7 +15,8 @@ module.exports = {
     "import", // require("eslint-plugin-import")
     "json", // require("eslint-plugin-json")
     "promise", // require("eslint-plugin-promise")
-    "react" // require("eslint-plugin-react")
+    "react", // require("eslint-plugin-react")
+    "react-hooks"
```
Mardak: Probably good to add the fake require comment as the previous lines have to make dumb simple grepping work.

bin/vendor-react.js
```diff
@@ -0,0 +1,26 @@
...
+  "react/umd/react.production.min.js": "react.js",
+  "react/umd/react.development.js": "react-dev.js",
+  "react-dom/umd/react-dom.production.min.js": "react-dom.js",
+  "react-dom/umd/react-dom.development.js": "react-dom-dev.js",
+};
+
+set("-v"); // Echo all the copy commands so the user can see what's going on
+for (let srcPath of Object.keys(filesToVendor)) {
+  cp(path.join("node_modules", srcPath),
+    path.join("vendor", filesToVendor[srcPath]));
```
Mardak: Could probably have just shoved these into a single npm run vendor-react without a separate `bin/.js` but I guess this is more readable. 😉 

Still could be useful to expose a run script.

.eslintrc.js
dmose: Sure.  Out of curiousity, in what circumstances does one grep for this?  Just
when manually checking to see if this module is still used?

bin/vendor-react.js
dmose: What can I say, I'm a readability nerd. :-)  Good idea, though; I'll add it to
yamscripts.yaml for discoverability.

PR #4808, opened 2019-02-21
Bug 1528059 - Fixing dismissed Pocket stories come back until stories…

ScottDowne:
… update

### DIFF
--- a/lib/TopStoriesFeed.jsm
+++ b/lib/TopStoriesFeed.jsm
@@ -113,6 +113,23 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.store.dispatch(shouldBroadcast ? ac.BroadcastToContent(action) : ac.AlsoToPreloaded(action));
   }
 
+  /**
+   * doContentUpdate - Updates topics and stories in the topstories section.
+   *
+   *                   Sections have one update action for the whole section.
+   *                   Redux creates a state race condition if you call the same action,
+   *                   twice, concurrently. Because of this, doContentUpdate is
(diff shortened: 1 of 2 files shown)


ScottDowne:
@Mardak I have test work to do still, but the general structure is there.

lib/TopStoriesFeed.jsm
```diff
@@ -113,13 +113,18 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
 
-  doContentUpdate(shouldBroadcast) {
+  doContentUpdate({stories, topics}, shouldBroadcast) {
     let updateProps = {};
-    if (this.stories) {
-      updateProps.rows = this.stories;
+    if (stories) {
+      updateProps.rows = stories;
+    } else {
+      const {Sections} = this.store.getState();
```
Mardak: Add a comment on why we need to get the previous `rows` especially because we don't need to for `topics` below.

lib/TopStoriesFeed.jsm
```diff
@@ -113,13 +113,18 @@ this.TopStoriesFeed = class TopStoriesFeed {
     this.store.dispatch(shouldBroadcast ? ac.BroadcastToContent(action) : ac.AlsoToPreloaded(action));
   }
 
-  doContentUpdate(shouldBroadcast) {
+  doContentUpdate({stories, topics}, shouldBroadcast) {
```
Mardak: Probably good to document the method to be clear why it exists vs the previous behavior -- namely a single place to update both stories and topics instead of trying to update them individually.

PR #4810, opened 2019-02-21
Bug 1529334 - Blocking/unblocking CFR doesn't clear impressions

piatra:
I don't think this ever worked:
When we're turning CFR off we call `cleanupImpressions` that calls `_cleanupImpressionsForItems` and updates the store with the new impression numbers.
When we block we never called `store.set` for impressions.


k88hudson:
This *definitely* worked before for me since it's how I used to do testing all the time... I'm kind of curious what broke it

PR #4814, opened 2019-02-22
Bug 1517849 - Mirror impression telemetry to Structured Ingestion dat…

ncloudioj:
…a pipeline

This mirrors all the impression stats metrics (including the old Top Stories and the new Discovery Stream) to the new Firefox data pipeline (aka. Structured Ingestion) in beta and nightly channel.

We're planning on performing the face-to-face comparison between the new data pipeline and the tiles one in the pre-release channel.


ncloudioj:
It has been verified in stage with data team.

Kick off code review as well as data review.

lib/TelemetryFeed.jsm
```diff
@@ -62,11 +64,15 @@ this.TelemetryFeed = class TelemetryFeed {
     this._impressionId = this.getOrCreateImpressionId();
     this.telemetryEnabled = this._prefs.get(TELEMETRY_PREF);
     this.eventTelemetryEnabled = this._prefs.get(EVENTS_TELEMETRY_PREF);
+    this.structuredIngestionTelemetryEnabled = this._prefs.get(STRUCTURED_INGESTION_TELEMETRY_PREF);
+    this._structuredIngestionEndpointBase = this._prefs.get(STRUCTURED_INGESTION_ENDPOINT_PREF);
```
rlr: Instead of keeping a copy of these prefs, you can just get them when you are about to use them below. That way you don't need to add the observer to listen for changes. Currently, you aren't listening for changes to the endpoint pref so that value would get out of sync if you change the pref.

Or is there a reason we keep a copy of all these prefs locally instead of just getting them when they are about to be used? Changing it would simplify the code a bit and require less tests.

lib/TelemetryFeed.jsm
```diff
@@ -556,6 +568,32 @@ this.TelemetryFeed = class TelemetryFeed {
...
+   * https://github.com/mozilla/gcp-ingestion/blob/master/docs/edge.md#postput-request
+   *
+   * @param {String} pingType  Type of the ping, such as "impression-stats".
+   * @param {String} version   Endpoint version for this ping type.
+   */
+  _generateStructuredIngestionEndpoint(pingType, version) {
+    const uuid = gUUIDGenerator.generateUUID().toString();
+    // Structured Ingestion does not support the UUID generated by gUUIDGenerator,
+    // because it contains leading and trailing braces. Need to trim them first.
+    const docID = uuid.substring(1, uuid.length - 1);
```
rlr: You can also do `uuid.slice(1, -1)` but I guess that's pretty much the same minus checking the length.

lib/TelemetryFeed.jsm
ncloudioj: Confirmed with the original author for the rationale about this, looks like the intention was to avoid load those prefs for each use. Assuming `Preferences.get()` should be fast, let's try reading directly from it every time, which will largely simplify the code and its tests.

lib/TelemetryFeed.jsm
ncloudioj: Good idea, let's stick to slice.

lib/TelemetryFeed.jsm
```diff
@@ -54,24 +54,36 @@ const USER_PREFS_ENCODING = {
...
     this.sessions = new Map();
     this._prefs = new Prefs();
     this._impressionId = this.getOrCreateImpressionId();
-    this.telemetryEnabled = this._prefs.get(TELEMETRY_PREF);
-    this.eventTelemetryEnabled = this._prefs.get(EVENTS_TELEMETRY_PREF);
     this._aboutHomeSeen = false;
-    this._onTelemetryPrefChange = this._onTelemetryPrefChange.bind(this);
-    this._prefs.observe(TELEMETRY_PREF, this._onTelemetryPrefChange);
-    this._onEventsTelemetryPrefChange = this._onEventsTelemetryPrefChange.bind(this);
-    this._prefs.observe(EVENTS_TELEMETRY_PREF, this._onEventsTelemetryPrefChange);
```
rlr: nice cleanup 😄

rlr:
r+ @ncloudioj :shipit:

PR #4816, opened 2019-02-22
Bug 1529035 - State and pref reset

ScottDowne:
Adding steps to test now...

### DIFF
--- a/common/Reducers.jsm
+++ b/common/Reducers.jsm
@@ -466,7 +466,13 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
     case at.DISCOVERY_STREAM_LAYOUT_UPDATE:
       return {...prevState, lastUpdated: action.data.lastUpdated || null, layout: action.data.layout || []};
     case at.DISCOVERY_STREAM_LAYOUT_RESET:
-      return {...prevState, lastUpdated: INITIAL_STATE.DiscoveryStream.lastUpdated, layout: INITIAL_STATE.DiscoveryStream.layout};
+      return {
+        ...prevState,
+        lastUpdated: INITIAL_STATE.DiscoveryStream.lastUpdated,
+        layout: INITIAL_STATE.DiscoveryStream.layout,
+        feeds: INITIAL_STATE.DiscoveryStream.feeds,
+        spocs: INITIAL_STATE.DiscoveryStream.spocs,
+      };
     case at.DISCOVERY_STREAM_FEEDS_UPDATE:
       return {
         ...prevState,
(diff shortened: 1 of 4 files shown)


common/Reducers.jsm
```diff
@@ -466,7 +466,13 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
     case at.DISCOVERY_STREAM_LAYOUT_UPDATE:
       return {...prevState, lastUpdated: action.data.lastUpdated || null, layout: action.data.layout || []};
     case at.DISCOVERY_STREAM_LAYOUT_RESET:
-      return {...prevState, lastUpdated: INITIAL_STATE.DiscoveryStream.lastUpdated, layout: INITIAL_STATE.DiscoveryStream.layout};
+      return {
+        ...prevState,
```
k88hudson: I wonder if it actually might be better to do:
```js
return {...INITIAL_STATE, config: prevState.config}
``
Since the intention is for everything to be reset except the config. What do you think?

File: lib/DiscoveryStreamFeed.jsm [hunk 2]
```diff
@@ -534,6 +534,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   }
 
   async disable() {
```
k88hudson: if you're going to use this for both the disable and change case,  think it might be good to actually rename this function and add a comment saying it's used in both cases in case someone accidentally adds something that is disable-specific without realizing it's used when the pref is changed. So maybe `resetState` or something?

ScottDowne: I like `reset`, does that work?

ScottDowne: OK, I'm maybe we go with `resetState`, pull `clearCache` and `clearImpressionPrefs` out, rename them `resetCache` and `resetImpressionsPrefs`, and add a `reset` that calls those three.

```
  async reset() {
    this.resetImpressionPrefs();
    await this.resetCache();
    this.resetState();
  }

  async resetCache() {
    await this.cache.set("layout", {});
    await this.cache.set("feeds", {});
    await this.cache.set("spocs", {});
  }

  resetImpressionPrefs() {
    this.writeImpressionsPref(PREF_SPOC_IMPRESSIONS, {});
    this.writeImpressionsPref(PREF_REC_IMPRESSIONS, {});
  }

  resetState() {
    // Reset reducer
    this.store.dispatch(ac.BroadcastToContent({type: at.DISCOVERY_STREAM_LAYOUT_RESET}));
    this.loaded = false;
    this.layoutRequestTime = undefined;
    this.spocsRequestTime = undefined;
    this.componentFeedRequestTime = undefined;
    this.totalRequestTime = undefined;
  }

  async onPrefChange() {
    // We always want to clear the cache/state if the pref has changed
    await this.reset();
    if (this.config.enabled) {
      // Load data from all endpoints
      await this.enable();
    }
  }
```

common/Reducers.jsm
ScottDowne: Works for me. Should do exactly the same thing.

ScottDowne:
Updated, also some good steps to test in the bug in the last comment.

punamdahiya:
@k88hudson Looks like you need to submit review approving the requested changes before pull request can be merged

PR #4818, opened 2019-02-26
Bug 1527642 - Pocket New Tab context menu UI

punamdahiya:


content-src/components/DiscoveryStreamComponents/DSLinkMenu/_DSLinkMenu.scss
```diff
@@ -0,0 +1,21 @@
+.ds-hero-item,
+.ds-list-item,
+.ds-card {
+  @include context-menu-button;
+  // For max-width  1378px, with $break-point-widest 1042 px
+  // and $card-width 224px
+  @media (max-width: $break-point-widest + $card-width * 1.5) {
+    &.last-item {
```
punamdahiya: @Mardak This PR handles last item show in card grid and lists to show context-menu to the left. However this doesn't work for complex layout where we display card and list items in 2/3 and 1/3 width. This issue exists for Top Sites also when shown in 2/3 and 1/3 width when seen on dev-test-all

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -4,14 +4,22 @@ export class ContextMenu extends React.PureComponent {
...
   }
 
+  onShow() {
+    if (this.props.onShow) {
+      this.props.onShow();
+    }
+  }
+
   componentDidMount() {
+    this.onShow();
```
punamdahiya: Added onShow that notifies components rendering ContextMenu, it has mounted. This is used by DSLinkMenu  to check if rendering of ContextMenu has caused any horizontal scroll.

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
```diff
@@ -0,0 +1,84 @@
...
+    });
+  }
+
+  onMenuUpdate(showContextMenu) {
+    this.setState({showContextMenu});
+  }
+
+  onMenuShow() {
+    if (window.scrollMaxX > 0) {
+      this.contextMenuButtonRef.current.parentElement.classList.add("last-item");
```
punamdahiya: @Mardak Updated logic to add last-item className by checking if rendering of context menu is causing any horizontal scrolls, if yes we should open context menu to left.

punamdahiya:
@ScottDowne @gvn @Mardak including to review. This has all UI and styles changes needed for context menu support in List, Card and Hero components. It renders link menu options 'Open in new window' and 'Open in new private window' that comes for free with no added back-end support.

Rest of the three options Dismiss, Bookmark and Save to Pocket needs server work and has their own respective bugs opened. See https://bugzilla.mozilla.org/show_bug.cgi?id=1527642#c6

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
ScottDowne: Can you help me understand why this is causing horizontal scroll in parts of this patch?

punamdahiya: By default context-menu open to the right when clicked. For last item in a row, this default open behavior causes horizontal scroll and need to be handled by fliping context menu direction to the left.

ScottDowne: Ah, interesting. For what it's worth this is an issue in topsites too.

ScottDowne: If I open a menu in topsites with discovery turned off, it somehow also manages to flip it to the left.

ScottDowne: I would be curious to see how this gets solved in the old activity stream, and why it's no longer working in discovery stream, including topsites.

punamdahiya: https://searchfox.org/mozilla-central/search?q=context-menu-open-left&path=

we are handling explicit breakpoints in css. See context-menu-open-left in _sections.scss for recommended top stories and _topsites.scss for top sites.

Considering the dynamic layout of discovery stream experience, IMO its hard to handle different edge cases in css

punamdahiya:
PR updated with conflicts resolved with latest master

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -28,31 +29,33 @@ export class DSCard extends React.PureComponent {
...
     return (
-      <SafeAnchor url={this.props.url} className="ds-card" onLinkClick={this.onLinkClick}>
-        <div className="img-wrapper">
-          <div className="img" style={{backgroundImage: `url(${this.props.image_src}`}} />
-        </div>
-        <div className="meta">
-          <div className="info-wrap">
-            <header className="title">{this.props.title}</header>
-            {this.props.excerpt && <p className="excerpt">{this.props.excerpt}</p>}
+      <DSLinkMenu containerClassName="ds-card" {...this.props}>
```
ScottDowne: Can you help me understand what we are solving with wrapping the rest of the component?

ScottDowne: I bring it up because of the other bugs caused by wrapping things previously, just making sure we've considered that and ensure it's not breaking things.

punamdahiya: [CSS on context-menu-button](https://github.com/mozilla/activity-stream/blob/master/content-src/styles/_variables.scss#L151) expects [context menu button](https://github.com/mozilla/activity-stream/pull/4818/files#diff-dc7ba3119134890bf2373097e54ada9fR44) to be at same level as anchored item (ds-card-link) and the reason for wrapping anchor here

ScottDowne: Ok, it doesn't seem to regress anything. So I think should be fine. Def was a rd flag though given how similar changes have caused bugs in the past.

ScottDowne: Same level as `ds-card-link` or parent to `ds-card-link`?

ScottDowne: Oh, I see, the button is returned as a sibling to `{this.props.children}`

Mardak: Couldn't you just put the DSLinkMenu as a sibling of the existing code? It definitely seems dangerous to introduce additional div nesting given the regressions from before

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
ScottDowne: I'm seeing last-item applied to the middle or second element in 3 or 4 card grids.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
Mardak: Oh, I suppose a direct sibling might cause issues with the parent assuming individual children for grid reasons ?

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
ScottDowne: Oh, I see, it moves it if there is a horizontal scroll bar, doesn't seem to matter if it caused it or not.

ScottDowne: I can make the menu appear partially off screen on the left side. (getting a screen shot)

punamdahiya: may be rename last-item to last-screen-item or something better to explain it's last item closer to edge on the screen, so if a layout has middle or second element on the edge it will have last-screen-item applied

punamdahiya: > I can make the menu appear partially off screen on the left side. (getting a screen shot)

With responsive support, horizontal scroll appearing due to window resize should be resolved

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
ScottDowne: Yeah it makes me nervous but I'm not actually able to break it. I think Punam has covered all the cases. maybe we're fine, what do you think  @Mardak ?

punamdahiya: > Oh, I suppose a direct sibling might cause issues with the parent assuming individual children for grid reasons ?

yes tried to avoid nesting but  'ds-card' classname has to be at the individual item container level

Mardak: It would seem to make more sense if the components were structured as:
```jsx
card render =>
  <div className="ds-card">
    <SafeAnchor>
      … existing stuff
      … existing <ImpressionStats/>
    </SafeAnchor>
    <DSLinkMenu />
  </div>
```

So for List, DSLinkMenu doesn't need to care about `<div>` vs `<li>`. Hopefully it also makes it a little bit more clear that these visual components are creating a single container for their parents to grid align and the sibling relationship of anchor and menu.

Mardak: Actually.. there's the active state that needs to modify the "ds-card" container class list. And generally common callback behavior of the menu being needed for all the components. Probably not the time to refactor the visual components to avoid needing to reimplement in each place…

punamdahiya: Either that or we can use contextMenuButtonRef to access containing visual component and add active class to it, we need to explicitly remove active class when context menu hides

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -28,31 +29,34 @@ export class DSCard extends React.PureComponent {
...
+              <span className="source">{this.props.source}</span>
+            </p>
+          </div>
+          <ImpressionStats
+            campaignId={this.props.campaignId}
+            rows={[{id: this.props.id, pos: this.props.pos}]}
+            dispatch={this.props.dispatch}
+            source={this.props.type} />
+        </SafeAnchor>
+        <DSLinkMenu {...this.props} />
```
ScottDowne: :+1:

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
```diff
@@ -0,0 +1,72 @@
...
+    if (!showContextMenu) {
+      const dsLinkMenuHostDiv = this.contextMenuButtonRef.current.parentElement;
+      dsLinkMenuHostDiv.parentElement.classList.remove("active", "last-item");
+    }
+    this.setState({showContextMenu});
+  }
+
+  onMenuShow() {
+    const dsLinkMenuHostDiv = this.contextMenuButtonRef.current.parentElement;
+    if (window.scrollMaxX > 0) {
```
ScottDowne: I'm still seeing scrollMax, is that still needed with the opacity check?

ScottDowne: I think I imagined rendering an opacity 0 menu, grabbing the width, and checking that the width + starting position is less than the window viewport.

punamdahiya: Opacity check is helping with hiding the possible jank due to flip from default right direction to left as context menu renders with opacity 0 and force scroll and thats why scrollMaxX check.

punamdahiya:
TC failures due to dark mode fixes tracked in https://github.com/mozilla/activity-stream/pull/4829

PR #4819, opened 2019-02-27
Bug 1528953 - Add pref to opt out of recommended features

piatra:
https://phabricator.services.mozilla.com/D21408

### DIFF
--- a/lib/ASRouterPreferences.jsm
+++ b/lib/ASRouterPreferences.jsm
@@ -63,6 +63,10 @@ class _ASRouterPreferences {
       if (!Services.prefs.prefHasUserValue(oldPref)) {
         continue;
       }
+      if (Services.prefs.prefHasUserValue(newPref)) {
+        Services.prefs.clearUserPref(oldPref);
+        continue;
+      }
       // If the pref was user modified we assume it was set to false
       const oldValue = Services.prefs.getBoolPref(oldPref, false);
       Services.prefs.clearUserPref(oldPref);
--- a/lib/ASRouterPreferences.jsm
+++ b/lib/ASRouterPreferences.jsm
@@ -56,6 +56,8 @@ class _ASRouterPreferences {
     }, []);
   }
 
+  // XXX Bug 1531734
+  // Required for 67 when the pref change will happen
   _migratePrefs() {
     for (let [oldPref, newPref] of MIGRATE_PREFS) {
       if (!Services.prefs.prefHasUserValue(oldPref)) {
@@ -63,6 +65,7 @@ class _ASRouterPreferences {
       }
       // If the pref was user modified we assume it was set to false
       const oldValue = Services.prefs.getBoolPref(oldPref, false);
+      Services.prefs.clearUserPref(oldPref);
       Services.prefs.setBoolPref(newPref, oldValue);
     }
   }
--- a/lib/ASRouterPreferences.jsm
+++ b/lib/ASRouterPreferences.jsm
@@ -16,9 +16,15 @@ const DEFAULT_STATE = {
   _devtoolsPref: DEVTOOLS_PREF,
 };
 
+const MIGRATE_PREFS = [
+  // Old pref, New pref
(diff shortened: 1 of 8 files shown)


lib/CFRMessageProvider.jsm
```diff
@@ -284,7 +289,7 @@ const CFR_MESSAGES = [
           label: {string_id: "cfr-doorhanger-extension-manage-settings-button"},
           action: {
             type: "OPEN_PREFERENCES_PAGE",
-            data: {category: "general-cfr", origin: "CFR"},
```
k88hudson: you probably want to check with @ncloudioj to make sure this change won't affect any of our data / is OK

lib/ASRouterPreferences.jsm
```diff
@@ -50,6 +56,17 @@ class _ASRouterPreferences {
     }, []);
   }
 
+  _migratePrefs() {
```
k88hudson: Can you add a comment about what version this is happening and when it should be removed if ever (maybe we should file a ticket for that also)

lib/ASRouterPreferences.jsm
```diff
@@ -50,6 +56,17 @@ class _ASRouterPreferences {
...
+    for (let [oldPref, newPref] of MIGRATE_PREFS) {
+      if (!Services.prefs.prefHasUserValue(oldPref)) {
+        continue;
+      }
+      // If the pref was user modified we assume it was set to false
+      const oldValue = Services.prefs.getBoolPref(oldPref, false);
+      Services.prefs.setBoolPref(newPref, oldValue);
```
k88hudson: I think you also need to clear the user value of the old pref to prevent this from happening each time

lib/CFRMessageProvider.jsm
piatra: I've only changed the `category` field used to target which section and option in `about:preferences` to open to and highlight. The `origin` field used for telemetry is the same.

https://searchfox.org/mozilla-central/rev/00f3836a87b844b5e4bc82f698c559b9966e4be2/browser/base/content/utilityOverlay.js#856

lib/ASRouterPreferences.jsm
piatra: https://bugzilla.mozilla.org/show_bug.cgi?id=1531734

lib/CFRMessageProvider.jsm
ncloudioj: This change won't affect the telemetry.

@piatra This patch needs a data review, because it also added two events (`MANAGE` and 'PIN_CURRENT_TAB') to the telemetry.

test/unit/asrouter/ASRouter.test.js
```diff
@@ -382,6 +382,44 @@ describe("ASRouter", () => {
...
+      sandbox.stub(ASRouterPreferences, "providers").value(providers);
+      sandbox.stub(ASRouterPreferences, "getUserPreference")
+        .withArgs("cfrFeatures").returns(false)
+        .withArgs("cfrAddons")
+        .returns(true);
+      Router._updateMessageProviders();
+      assert.equal(Router.state.providers.length, 1);
+      assert.equal(Router.state.providers[0].id, providers[0].id);
+    });
+    it("should only add the providers that are enabled", () => {
```
k88hudson: These tests could use more descriptive names

k88hudson:
LGTM with requested changes, feel free to separate out the the telemetry changes to a separate patch (if you can) if you don't want to be blocked on merging this

lib/ASRouterPreferences.jsm
k88hudson: Actually, one more thing I am wondering here – if a user preference is already set for the new pref, do we still want to override it with the old one? Probably not right?

piatra: Wouldn't the migration code run before the user can do any changes to the prefs. I don't understand how we could get in that scenario.

k88hudson: Maybe syncing an old profile for example – it's definitely an edge case though

PR #4820, opened 2019-02-28
Bug 1521203 - Ensure appropriate resource requests for endpoints (layout, feed, spoc)

Mardak:
r?@ScottDowne This adds a prefix check based on a string pref. I was thinking maybe just a plain boolean but that loses out on some flexibility if we ever needed to change the pref to something else for Release users…

Also, I guess we'll need to let people know to set the pref to something like "http" if using custom endpoints.

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -239,6 +239,10 @@ const PREFS_CONFIG = new Map([
       });
     },
   }],
+  ["discoverystream.endpoints", {
+    title: "Endpoint prefixes (comma-separated) that are allowed to be requested",
+    value: "https://getpocket.cdn.mozilla.net/",
+  }],
   ["discoverystream.optOut.0", {
     title: "Opt out of new layout v0",
     value: false,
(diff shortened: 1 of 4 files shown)


lib/ActivityStream.jsm
```diff
@@ -239,6 +239,10 @@ const PREFS_CONFIG = new Map([
       });
     },
   }],
+  ["discoverystream.endpoints", {
```
ScottDowne: This is a separate pref on purpose to make it impossible to change via shield? Or do we want to consider allowing moving this to our primary json blob pref so shield can modify it while doing an experiment?

Again, this might be intended as something we explicitly don't want to allow, and if so, this makes sense.

lib/ActivityStream.jsm
```diff
@@ -239,6 +239,10 @@ const PREFS_CONFIG = new Map([
...
+    title: "Endpoint prefixes (comma-separated) that are allowed to be requested",
+    value: "https://getpocket.cdn.mozilla.net/",
```
ScottDowne: Has this been run by mathijs to make sure he's not expecting other pocket urls to be allowed?

ScottDowne:
@Mardak A few questions so far, that I don't think are quite a finished review yet.

lib/ActivityStream.jsm
Mardak: He provided a list of production and dev values https://bugzilla.mozilla.org/show_bug.cgi?id=1521203#c1 and we're only handling the feeds and not images here, so we end up with just the cdn.

lib/ActivityStream.jsm
Mardak: Nod. I believe security wanted a standalone pref easy to check if it's been modified. For our own normandy pref flip studies, do we expect to ever point large amount of users not at the cdn?

PR #4822, opened 2019-03-01
Bug 1525442 – Enable Dark Mode

gvn:
*Has had a design review from Wolasi.

Repro:

- Turn on Dark theme in preferences

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
@@ -10,6 +10,10 @@ $excerpt-line-height: 20;
   position: relative;
 
   &:hover {
+    @include dark-theme-only {
+      background: $grey-70;
+    }
+
     header {
       color: $blue-60;
     }
@@ -54,18 +58,30 @@ $excerpt-line-height: 20;
 
     .context,
     .source {
+      @include dark-theme-only {
+        color: $teal-60;
+      }
+
       font-size: 13px;
       color: $teal-80;
     }
   }
 
   header {
+    @include dark-theme-only {
+      color: $blue-40;
+    }
+
     line-height: $header-line-height * 1px;
     font-size: $header-font-size * 1px;
     color: $grey-90;
   }
 
   p {
+    @include dark-theme-only {
+      color: $grey-30;
+    }
+
     font-size: $excerpt-font-size * 1px;
     line-height: $excerpt-line-height * 1px;
     color: $grey-50;
--- a/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
@@ -10,11 +10,11 @@ $excerpt-line-height: 20;
   position: relative;
 
   &:hover {
-    @include dark-theme-only {
-      background: $grey-70;
-    }
-
     header {
+      @include dark-theme-only {
+        color: $blue-40;
+      }
+
       color: $blue-60;
     }
   }
@@ -59,7 +59,7 @@ $excerpt-line-height: 20;
     .context,
     .source {
       @include dark-theme-only {
-        color: $teal-60;
+        color: $teal-10;
       }
 
       font-size: 13px;
@@ -69,7 +69,7 @@ $excerpt-line-height: 20;
 
   header {
     @include dark-theme-only {
-      color: $blue-40;
+      color: $grey-10;
     }
 
     line-height: $header-line-height * 1px;
--- a/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
@@ -63,7 +63,6 @@ $excerpt-line-height: 20;
     line-height: $header-line-height * 1px;
     font-size: $header-font-size * 1px;
     color: $grey-90;
-
   }
 
   p {
--- a/content-src/styles/_mixins.scss
+++ b/content-src/styles/_mixins.scss
@@ -15,3 +15,9 @@
   max-height: 1em * $line-count * $line-height / $font-size;
   overflow: hidden;
 }
+
+@mixin dark-theme-only {
+  [lwt-newtab-brighttext]:not(.force-light-theme) & {
+    @content;
+  }
+}
--- a/content-src/styles/_mixins.scss
+++ b/content-src/styles/_mixins.scss
@@ -21,3 +21,19 @@
     @content;
   }
 }
+
+@mixin ds-border-top {
+  @include dark-theme-only {
+    border-top: 1px solid $grey-60;
+  }
+
+  border-top: 1px solid $grey-30;
+}
+
+@mixin ds-border-bottom {
+  @include dark-theme-only {
+    border-bottom: 1px solid $grey-60;
+  }
+
+  border-bottom: 1px solid $grey-30;
+}
--- a/content-src/styles/_theme.scss
+++ b/content-src/styles/_theme.scss
@@ -82,7 +82,7 @@ body {
     // General styles
     --newtab-background-color: #{$grey-80};
     --newtab-border-primary-color: #{$grey-10-80};
-    --newtab-border-secondary-color: #{$grey-10-10};
+    --newtab-border-secondary-color: #{$grey-60};
     --newtab-button-primary-color: #{$blue-60};
     --newtab-button-secondary-color: #{$grey-70};
     --newtab-element-active-color: #{$grey-10-20};
--- a/content-src/styles/_theme.scss
+++ b/content-src/styles/_theme.scss
@@ -82,7 +82,7 @@ body {
     // General styles
     --newtab-background-color: #{$grey-80};
     --newtab-border-primary-color: #{$grey-10-80};
-    --newtab-border-secondary-color: #{$grey-60};
+    --newtab-border-secondary-color: #{$grey-10-10};
     --newtab-button-primary-color: #{$blue-60};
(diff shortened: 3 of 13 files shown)


content-src/styles/_mixins.scss
```diff
@@ -15,3 +15,9 @@
   max-height: 1em * $line-count * $line-height / $font-size;
   overflow: hidden;
 }
+
+@mixin dark-theme-only {
+  [lwt-newtab-brighttext]:not(.force-light-theme) & {
+    @content;
```
Mardak: Why not set the css variables to the appropriate values instead of overriding with an additional selector for each use?

https://github.com/mozilla/activity-stream/blob/master/content-src/styles/_theme.scss

gvn: I tried both approaches and found this declarative one *far* easier to reason about and more consistent with the approach currently used for DS components.

Adding variables to every declaration that in turn point to other variables becomes incredibly cumbersome and convoluted. This approach feels far cleaner and more straightforward for maintenance.

content-src/styles/_theme.scss
```diff
@@ -82,7 +82,7 @@ body {
     // General styles
     --newtab-background-color: #{$grey-80};
     --newtab-border-primary-color: #{$grey-10-80};
-    --newtab-border-secondary-color: #{$grey-10-10};
+    --newtab-border-secondary-color: #{$grey-60};
```
ScottDowne: Is this something that would change non discovery stream dark mode?

content-src/styles/_mixins.scss
```diff
@@ -15,3 +15,9 @@
...
+  }
+}
```
ScottDowne: Can you explain what's going on here?

ScottDowne: Ah I think I get it, nvm.

gvn: Sure, it's a mixin that can wrap CSS declarations to make them only apply to the selector they're nested under in dark mode.

content-src/styles/_theme.scss
gvn: Good call. I can tweak that so it's separate.

content-src/styles/_mixins.scss
```diff
@@ -15,3 +15,9 @@
   max-height: 1em * $line-count * $line-height / $font-size;
   overflow: hidden;
 }
+
+@mixin dark-theme-only {
+  [lwt-newtab-brighttext]:not(.force-light-theme) & {
```
ScottDowne: Can you explain this bit? I get what it does based on the name of the mixin, I just don't really see how this selector does that.

gvn: It's based on an attribute that gets added to the body tag for dark mode.

content-src/styles/_mixins.scss
gvn: It basically allows dark mode code to be declared as overrides adjacent to the light mode defaults.

content-src/styles/_mixins.scss
Mardak: The variables represent a higher level concept that reflects what design is thinking of anyway. Design isn't deciding that this card should be grey-30 and that hero should also be grey-30 -- it's because dark theme text should be grey-30. With this PR, there's now two different ways to handle dark theme on the same page, but if it's preferable to maintain both, I guess that's fine.

ScottDowne: I'm leaning towards gvn a bit here currently.

Mardak: The css variables are used for web extension themes. See https://bugzilla.mozilla.org/show_bug.cgi?id=1347204 for more details

ScottDowne: I think I missed what Mardak was saying, but not sure. Maybe there are two requests here.

Are e talking about moving all of these:
```
  @include dark-theme-only {
    color: $grey-30;
  }
```

Or are we talking about updating all of these:
```
  @include dark-theme-only {
    color: $grey-30;
  }
```
to be:
```
  @include dark-theme-only {
    color: var(--newtab-text-primary-color);
  }
```

I think the way gavin has it makes sense only if this is a specific case where design is specifically asking for that element to be that colour in dark theme, or has already specifically asked for that element to not be a shared colour.

In general though, I see value in both. Gavin is probably seeing this as "I don't want to maintain ds components css in two plaecs" and Mardak is saying "We shouldn't be maintaining dark theme in two places?"

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -11,6 +11,10 @@ $excerpt-line-height: 20;
 
   &:hover {
     header {
+      @include dark-theme-only {
+        color: $blue-40;
```
ScottDowne: I think I lean towards something Mardak might be saying here, just took me a bit to put it all together.

For this component's colour, we can just use `--newtab-link-primary-color` because in dark mode it's set to `$blue-40` and in light mode, it's at `$blue-60`

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -54,19 +58,30 @@ $excerpt-line-height: 20;
 
     .context,
     .source {
+      @include dark-theme-only {
+        color: $teal-10;
```
ScottDowne: @Mardak This styling isn't in _theme.scss so it does feel like a specific request from design to have a colour not already defined. Is the idea here we add something to theme.scss for this text to swap from teal-80/teal-10?

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -54,19 +58,30 @@ $excerpt-line-height: 20;
...
+      }
+
       font-size: 13px;
       color: $teal-80;
     }
   }
 
   header {
+    @include dark-theme-only {
+      color: $grey-10;
```
ScottDowne: This could be `--newtab-text-primary-color` without the @include

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -54,19 +58,30 @@ $excerpt-line-height: 20;
...
+
     line-height: $header-line-height * 1px;
     font-size: $header-font-size * 1px;
     color: $grey-90;
-
   }
 
   p {
+    @include dark-theme-only {
+      color: $grey-30;
```
ScottDowne: @Mardak This one doesn't quite fit with what's in theme.scss

content-src/styles/_mixins.scss
gvn: I believe the request is to make all color values variables which are then pointed to the actually desired color variables within the theme CSS.

I see the benefit of that if there are many themes, but it feels overly abstracted and less declarative. I don't see the ROI for 2 themes and the approach I'm taking feels more consistent with how CSS is typically architected (eg: responsive layouts and overrides for specific instances being declarative and grouped).

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
gvn: I think we should decouple DS and AS as much as possible to prevent them from regressing each other if only one has changes.

gvn: If we do use the current approach we should create a namespaced set of variables IMO (eg: `--ds-link-primary-color`)

content-src/styles/_mixins.scss
ScottDowne: I hope I described this well enough, but @Mardak here is a bug to start the processes of talking about theme extensions, feel free to add comments :) https://bugzilla.mozilla.org/show_bug.cgi?id=1532353

ScottDowne:
@gvn given today's meeting, and my r+ from before, I still think this is good to land.

gvn:
Cool, I just tested again with latest master and everything looks good so I'm going to merge.

Mardak:
Taskcluster was red before this merged https://tools.taskcluster.net/groups/dKQuEOXPQ7WxQTgF1tNPCg/tasks/dKQuEOXPQ7WxQTgF1tNPCg/runs/0/logs/public%2Flogs%2Flive.log

Also failing on master https://tools.taskcluster.net/groups/dKQuEOXPQ7WxQTgF1tNPCg

gvn:
@Mardak Investigating now.

Mardak:
Anyone remember if `DarkModeMessage` was explicitly not removed when dark mode was enabled? @punamdahiya this and related should be removable now?

https://github.com/mozilla/activity-stream/blob/8b20ec0902e5109303c2fe30a32fda7cf33d74e8/content-src/components/Base/Base.jsx#L179

punamdahiya:
> Anyone remember if `DarkModeMessage` was explicitly not removed when dark mode was enabled? @punamdahiya this and related should be removable now?
> 
> [activity-stream/content-src/components/Base/Base.jsx](https://github.com/mozilla/activity-stream/blob/8b20ec0902e5109303c2fe30a32fda7cf33d74e8/content-src/components/Base/Base.jsx#L179)
> 
> Line 179 in [8b20ec0](/mozilla/activity-stream/commit/8b20ec0902e5109303c2fe30a32fda7cf33d74e8)
>  {prefs.darkModeMessage && <DarkModeMessage />}

DarkModeMessage removal wasn't part of enable dark mode, yes it's safe to remove, will file a bug for it to track.

Mardak:
Ah, I see the PR here removed the `disableDarkTheme` method which added `.force-light-theme` but didn't clean up the usage of the class. Cleaning up the leftovers in #4894

PR #4823, opened 2019-03-01
Bug 1531604 - Fixing impression pos

ScottDowne:


ScottDowne:
@ncloudioj yeah, so there are a few things that can cause position changes.

1. Spoc injection, which happens in `selectLayoutRender` in `maybeInjectSpocs` which happens before our stuff.
2. Rotation happens in the feed before `selectLayoutRender`.
3. Offset which happens before spoc injection in `selectLayoutRender`
4. frequency capping happens in the feed so also before `selectLayoutRender`.

So I also think we're good here.

PR #4824, opened 2019-03-04
Bug 1528966

piatra:
Phab changes https://phabricator.services.mozilla.com/D22084

This adds the new schema field `descriptionDetails`. 
Updated `_renderPopup` to handle the two different code paths for addons and features CFRs.
It will show the panel with the message content but not the image.

image

* Currently the Pin tab message still has `exclude: true`.

### DIFF
--- a/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
+++ b/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
@@ -166,6 +166,26 @@
         }
       ]
     },
+    "descriptionDetails": {
+      "description": "Additional information and steps on how to use",
+      "type": "object",
+      "properties": {
+        "steps": {
+          "description": "Array of messages or string_ids",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "string_id": {
+                "type": "string",
+                "description": "Id of string to localized addon description"
+              }
+            }
+          }
+        }
+      },
+      "required": ["steps"]
+    },
     "buttons": {
       "description": "The label and functionality for the buttons in the pop-over.",
       "type": "object",
--- a/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
+++ b/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
@@ -15,6 +15,10 @@
     }
   },
   "properties": {
+    "category": {
+      "type": "string",
+      "description": "Attribute used for different groups of messages from the same provider"
+    },
     "bucket_id": {
       "type": "string",
       "description": "A bucket identifier for the addon. This is used in order to anonymize telemetry for history-sensitive targeting."
@@ -319,5 +323,5 @@
     }
   },
   "additionalProperties": false,
-  "required": ["bucket_id", "notification_text", "heading_text", "text", "buttons"]
+  "required": ["category", "bucket_id", "notification_text", "heading_text", "text", "buttons"]
 }
--- a/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
+++ b/content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
@@ -184,7 +184,8 @@
                 "type": "string",
                 "description": "Id of string to localized addon description"
               }
-            }
+            },
+            "required": ["string_id"]
           }
         }
       },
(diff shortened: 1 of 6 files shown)


content-src/asrouter/templates/CFR/templates/ExtensionDoorhanger.schema.json
```diff
@@ -166,6 +166,26 @@
...
+    "descriptionDetails": {
+      "description": "Additional information and steps on how to use",
+      "type": "object",
+      "properties": {
+        "steps": {
+          "description": "Array of messages or string_ids",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
```
k88hudson: Can you add  `"required": ["string_id"]`

lib/CFRMessageProvider.jsm
```diff
@@ -306,7 +306,6 @@ const CFR_MESSAGES = [
     id: "PIN_TAB",
     template: "cfr_doorhanger",
     category: "cfrFeatures",
-    exclude: true,
```
k88hudson: Do we want to wait for the animation and bullet point formatting to actually turn enable this?

lib/CFRMessageProvider.jsm
```diff
@@ -315,7 +314,14 @@ const CFR_MESSAGES = [
         label: {string_id: "cfr-doorhanger-extension-sumo-link"},
         sumo_path: REDDIT_ENHANCEMENT_PARAMS.sumo_path,
       },
-      text: "Get easy access to your most-used sites. Keep sites open in a tab (even when you restart).",
+      text: {string_id: "cfr-doorhanger-pintab-description"},
+      descriptionDetails: {
+        steps: [
+          {"string_id": "cfr-doorhanger-pintab-step1"},
```
k88hudson: Cool, this works 👍

lib/CFRPageActions.jsm
```diff
@@ -288,12 +273,82 @@ class PageAction {
...
+      panelTitle = await this.getStrings(content.heading_text);
+      let stepsContainer = this.window.document.getElementById(stepsContainerId);
+
+      // Container for bullet point list
+      if (!stepsContainer) {
+        stepsContainer = this.window.document.createElement("vbox");
+        stepsContainer.setAttribute("id", stepsContainerId);
+        for (let step of content.descriptionDetails.steps) {
+          const li = this.window.document.createElement("li");
+          li.setAttribute("data-l10n-id", step.string_id);
```
k88hudson: can you use `this._l10n.setAttributes(li, step.string_id)` instead?

lib/CFRPageActions.jsm
```diff
@@ -341,22 +389,22 @@ class PageAction {
...
-      await this.getStrings(content.addon.title),
+      panelTitle,
       "cfr",
       mainAction,
       secondaryActions,
       options
     );
+
+    // XXX Find a better way to do this
+    // This piece of content is added dinamically when PopupNotifications.show is called
```
k88hudson: `dynamically`

lib/CFRPageActions.jsm
```diff
@@ -288,12 +273,82 @@ class PageAction {
...
+        this.dispatchUserAction(primary.action);
+        this.hideAddressBarNotifier();
+        this._sendTelemetry({message_id: id, bucket_id: content.bucket_id, event: "PIN"});
+        RecommendationMap.delete(browser);
+      };
+      panelTitle = await this.getStrings(content.heading_text);
+      let stepsContainer = this.window.document.getElementById(stepsContainerId);
+
+      // Container for bullet point list
+      if (!stepsContainer) {
```
k88hudson: I think you probably intended to only have `stepsContainer = this.window.document.createElement("vbox")` under this if statement? Right now the behaviour here will be that the text never gets changed if a second CFR shows up

piatra:
@k88hudson I took the idea from the `update urlbar icon` review and applied it here to show hide elements. I needed to move `category` inside the `message.content` because we only pass around `content`.
I keep removing/adding `exclude: true` from the message definition so it actually shows up when you try to run the PR. Do we want to move that to the provider pref so it's easier to turn on?

k88hudson:
I merged this into `mozilla/cfr-pinned-tabs` so ricky's patch can land against it (and we can merge both togther once data review is done)

k88hudson:
Merged to `cfr-pinned-tabs`

PR #4828, opened 2019-03-05
Bug 1532777 - Track cfr user preferences in AS pings

ncloudioj:


ncloudioj:
Merge it as the data review+ was granted in bugzilla.

PR #4831, opened 2019-03-06
Bug 1533089 - HTTP Referer not set when clicking articles

Mardak:
r?@k88hudson or @punamdahiya Tested with https://gist.githubusercontent.com/punamdahiya/44f78e2f338cbf95f5c28d31faed4639/raw/1c61974ca8ebd02431512d9425031e3a9c8e04f4/test-message-feed.json which has a "bill gates" article with javascript uri. Still sends telemetry that a click was attempted but doesn't open the link. Other links open fine with referrer:
screen shot 2019-03-06 at 11 42 35 am

And for spoc:
screen shot 2019-03-06 at 11 43 27 am

Also checked that links for navigation don't send referrer:
screen shot 2019-03-06 at 11 44 18 am

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/SafeAnchor/SafeAnchor.jsx
+++ b/content-src/components/DiscoveryStreamComponents/SafeAnchor/SafeAnchor.jsx
@@ -1,6 +1,34 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 import React from "react";
 
 export class SafeAnchor extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.onClick = this.onClick.bind(this);
+  }
+
+  onClick(event) {
+    // Use dispatch instead of normal link click behavior to include referrer
+    if (this.props.dispatch) {
+      event.preventDefault();
+      const {altKey, button, ctrlKey, metaKey, shiftKey} = event;
+      this.props.dispatch(ac.OnlyToMain({
+        type: at.OPEN_LINK,
+        data: {
+          event: {altKey, button, ctrlKey, metaKey, shiftKey},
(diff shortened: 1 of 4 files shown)


content-src/components/DiscoveryStreamComponents/SafeAnchor/SafeAnchor.jsx
```diff
@@ -1,6 +1,33 @@
...
+    // Use dispatch instead of normal link click behavior to include referrer
+    if (this.props.dispatch) {
+      event.preventDefault();
+      const {altKey, button, ctrlKey, metaKey, shiftKey} = event;
+      this.props.dispatch(ac.OnlyToMain({
+        type: at.OPEN_LINK,
+        data: {
+          event: {altKey, button, ctrlKey, metaKey, shiftKey},
+          referrer: "https://getpocket.com/recommendations",
+          url: event.currentTarget.href,
```
k88hudson: Any reason you don't want to use `props.url` here?

Mardak: SafeAnchor can clean up the url to be different (i.e., "") so we don't want to accidentally open an unsafe url via chrome. I'll add a comment

k88hudson:
Tested with your `gist` and with `dev-test-all`, verified headers and being sent and telemetry is still being sent as expected.

content-src/components/DiscoveryStreamComponents/SafeAnchor/SafeAnchor.jsx
k88hudson: ah, makes sense

PR #4827, opened 2019-03-05
Bug 1529340 - CFR Pinned tabs popup animation

rlr:
@piatra this is built on top of the branch from your PR #4824.

The commit includes a patch file for m-c. I'll open the phab review once we think it's stable.

Light theme animating:
screen shot 2019-03-05 at 3 57 07 pm

Light theme paused or with cosmetic animations pref disabled:
screen shot 2019-03-05 at 3 58 08 pm

Dark theme animating:
screen shot 2019-03-05 at 3 58 55 pm

Dark theme paused or with cosmetic animations pref disabled:
screen shot 2019-03-05 at 3 59 30 pm

PR #4833, opened 2019-03-07
Bug 1529340 - CFR Pinned tabs popup animation

rlr:
This replaces pull #4827. This one is against the `cfr-pinned-tabs` branch.

The m-c changes that go with this are here: https://phabricator.services.mozilla.com/D22398


lib/CFRMessageProvider.jsm
```diff
@@ -305,7 +305,6 @@ const CFR_MESSAGES = [
   {
     id: "PIN_TAB",
     template: "cfr_doorhanger",
-    exclude: true,
```
rlr: do we want to get this "unexcluded" with this patch or should I undo this change?

lib/CFRPageActions.jsm
```diff
@@ -275,6 +276,52 @@ class PageAction {
...
+    const footer = this.window.document.getElementById("cfr-notification-footer");
+    const ANIMATION_CONTAINER_ID = "cfr-notification-footer-pintab-animation-container";
+    let animationsEnabled = Services.prefs.getBoolPref(ANIMATIONS_ENABLED_PREF, true);
+    let animationContainer = this.window.document.getElementById(ANIMATION_CONTAINER_ID);
+    if (animationContainer) {
+      animationContainer.toggleAttribute("animate", animationsEnabled);
+      animationContainer.removeAttribute("paused");
+    } else {
+      animationContainer = this.window.document.createElement("vbox");
+      animationContainer.setAttribute("id", ANIMATION_CONTAINER_ID);
```
rlr: I see a pattern of `.createElement(); .setAttribute("id",....)` that I guess could be factored out?

rlr: along with the `.appendChild()` perhaps. 🙄

rlr: something like this ?

```javascript
  _createElement(type, parent, id) {
    let element = this.window.document.createElement(type);
    if (id) {
      element.setAttribute("id", id);
    }
    parent.appendChild(element);
    return element;
  }

  async _renderPinTabAnimation() {
    const footer = this.window.document.getElementById("cfr-notification-footer");
    const ANIMATION_CONTAINER_ID = "cfr-notification-footer-pintab-animation-container";
    let animationsEnabled = Services.prefs.getBoolPref(ANIMATIONS_ENABLED_PREF, true);
    let animationContainer = this.window.document.getElementById(ANIMATION_CONTAINER_ID);
    if (animationContainer) {
      animationContainer.toggleAttribute("animate", animationsEnabled);
      animationContainer.removeAttribute("paused");
    } else {
      animationContainer = this._createElement("vbox", footer, ANIMATION_CONTAINER_ID);
      animationContainer.toggleAttribute("animate", animationsEnabled);

      // spacer
      this._createElement("vbox", animationContainer).setAttribute("flex", 1);

      let controlsContainer = this._createElement("hbox", animationContainer, "cfr-notification-footer-animation-controls");

      // spacer
      this._createElement("vbox", controlsContainer).setAttribute("flex", 1);

      let pauseButton = this._createElement("hbox", controlsContainer, "cfr-notification-footer-pause-button");
      pauseButton.addEventListener("click", () => {
        animationContainer.setAttribute("paused", true);
      });

      let pauseLabel = this._createElement("label", pauseButton, "cfr-notification-footer-pause-label");
      pauseLabel.textContent = await this.getStrings({"string_id": "cfr-doorhanger-pintab-animation-pause"});

      // pause icon
      this._createElement("image", pauseButton, "cfr-notification-footer-pause-icon");
    }
  }
```

rlr:
The task-cluster failures are due to unreferenced images. That should fix itself once https://phabricator.services.mozilla.com/D22398 makes its way to m-c.

rlr:
I ended up pushing the images in the m-c patch to avoid test failures on both sides.

lib/CFRPageActions.jsm
piatra: I like it. Writing unit tests that mess with the DOM is weird, this should make it nicer.

lib/CFRMessageProvider.jsm
piatra: undo it, we'll turn it on with a separate PR.

lib/CFRPageActions.jsm
```diff
@@ -275,6 +276,52 @@ class PageAction {
...
+
+      let spacer = this.window.document.createElement("vbox");
+      spacer.setAttribute("flex", 1);
+      animationContainer.appendChild(spacer);
+
+      let controlsContainer = this.window.document.createElement("hbox");
+      controlsContainer.setAttribute("id", "cfr-notification-footer-animation-controls");
+
+      spacer = this.window.document.createElement("vbox");
+      spacer.setAttribute("flex", 1);
```
piatra: Could this be replaced by `margins` in the m-c CSS?

lib/CFRPageActions.jsm
piatra: I would group `type` and `id` into an options object and have something like
```
_createElementAndAppend({id, type}, parent)
```

lib/CFRPageActions.jsm
```diff
@@ -275,6 +276,52 @@ class PageAction {
...
+      controlsContainer.appendChild(spacer);
+
+      let pauseButton = this.window.document.createElement("hbox");
+      pauseButton.setAttribute("id", "cfr-notification-footer-pause-button");
+      pauseButton.addEventListener("click", () => {
```
piatra: We need to remove this event listener.

piatra: Probably in `_popupStateChange`

lib/CFRPageActions.jsm
```diff
@@ -275,6 +276,52 @@ class PageAction {
...
+      });
+
+      let pauseLabel = this.window.document.createElement("label");
+      pauseLabel.setAttribute("id", "cfr-notification-footer-pause-label");
+      pauseLabel.textContent = await this.getStrings({"string_id": "cfr-doorhanger-pintab-animation-pause"});
+      pauseButton.appendChild(pauseLabel);
+
+      let pauseIcon = this.window.document.createElement("image");
+      pauseIcon.setAttribute("id", "cfr-notification-footer-pause-icon");
+      pauseButton.appendChild(pauseIcon);
```
piatra: Could we add the icon via CSS as background so we don't create so much markup?

lib/CFRPageActions.jsm
rlr: possibly? Would it auto-grow? But the m-c change is already in autoland so it would have to be a followup.

lib/CFRPageActions.jsm
rlr: it is CSS but I guess maybe it didnt need it's own element. also, a followup since m-c change already shipped.

rlr:
The images also shipped with the m-c CSS changes so tests would pass (unreferenced file failures). So I guess I have a list of things I can clean up next week. I can file a followup.

lib/CFRPageActions.jsm
rlr: I just added it on `window.unload`.

lib/CFRPageActions.jsm
rlr: Oh I see. You mean not using a spacer and just setting padding-top or margin-top on the container. Makes sense 👍

rlr:
I'll file followup bugs to:
* simplify the markup:
** remove some unnecessary elements for spacing and the icon and update the CSS accordingly
** get and use 2X res images
** make the images configurable (this would be for Firefox 68+, we don't need it in 67)

Anything else?

rlr:
Taskcluster should be green once https://hg.mozilla.org/integration/autoland/rev/6a3b22130b996d2ea931525093f3e6103ed13d20 makes its way to m-c

lib/CFRPageActions.jsm
k88hudson: I don't think this is strictly necessary as-is considering the event listener gets added only on element creation and presumably the element gets destroyed on `window.unload` anyway, although I wonder if it would be more appropriate to attach/remove the event listener on `_expand`/`_collapse` so that it's not long-lived throughout the whole life of the window. What do you think?

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -110,6 +113,8 @@ describe("CFRPageActions", () => {
     for (const id of elementIDs) {
       const elem = document.createElement("div");
       elem.setAttribute("id", id);
+      // TODO: Remove this once travis is on Firefox 63+
+      elem.toggleAttribute = () => {};
```
k88hudson: lol

lib/CFRPageActions.jsm
```diff
@@ -275,6 +284,47 @@ class PageAction {
...
+
+      // pause icon
+      this._createElement({type: "image", id: "cfr-notification-footer-pause-icon"}, pauseButton);
+    }
+
+    animationContainer.toggleAttribute("animate", Services.prefs.getBoolPref(ANIMATIONS_ENABLED_PREF, true));
+    animationContainer.removeAttribute("paused");
+
+    let pauseButton = this.window.document.getElementById(PAUSE_BUTTON_ID);
+    this.onPauseClick = () => { animationContainer.setAttribute("paused", true); };
```
k88hudson: you could also put these lines under `if !(this.onPauseClick) { }` just in case

rlr: yep, that's a good idea 👍

k88hudson:
Ok, I'm merging this to `mozilla/cfr-pinned-tabs` and rebasing on the export.

PR #4834, opened 2019-03-07
Backport Bug 1528955 - Tab successfully pinned notification

rlr:
This is against master.

The changes to `lib/ASRouter.jsm` are already in central.


rlr:
errr. I guess the test fixes should be in a PR to master. Fixing.

piatra:
Backports should be against `central` branch and we then merge that into master.

rlr:
The backport itself is already on the central branch. But it needed an update for tests to pass. So I included the merge from central and the update commit in this PR against master. I *think* that's the process from what I understood in the docs?

k88hudson:
Looks good, I'll land this for you

PR #4837, opened 2019-03-08
Bug 1531585 - Dismiss as link menu option

punamdahiya:


common/Reducers.jsm
```diff
@@ -484,6 +484,39 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
           spocs_endpoint: action.data || INITIAL_STATE.DiscoveryStream.spocs.spocs_endpoint,
         },
       };
+    case at.PLACES_LINK_BLOCKED:
+      if (!action.data || !prevState.spocs.loaded || !prevState.feeds.loaded) {
```
k88hudson: You might want to add a comment here about why you're returning early

punamdahiya:
@ScottDowne @k88hudson submitting dismiss link menu option for review. Thanks!

common/Reducers.jsm
```diff
@@ -484,6 +484,37 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
           spocs_endpoint: action.data || INITIAL_STATE.DiscoveryStream.spocs.spocs_endpoint,
         },
       };
+    case at.PLACES_LINK_BLOCKED:
+      // Return if action data is empty, or spocs or feeds data is not loaded
+      if (!action.data || !prevState.spocs.loaded || !prevState.feeds.loaded) {
+        return prevState;
+      }
+      // Return if spocs is undefined in prevState.
+      if (!(prevState.spocs.data && prevState.spocs.data.spocs)) {
```
punamdahiya: This check impacts if feeds are present, updating PR to check for empty spocs data inside return

punamdahiya:
@ncloudioj Including you to help validate that discovery stream is sending similar telemetry around dismiss action as is done in default activity stream dismiss flow. Thanks!

punamdahiya:
> @ncloudioj Including you to help validate that discovery stream is sending similar telemetry around dismiss action as is done in default activity stream dismiss flow. Thanks!

Created https://bugzilla.mozilla.org/show_bug.cgi?id=1534845 for pocket new tab link menu telemetry parity review

ncloudioj:
@punamdahiya Sure, I can take care of the telemetry validation for this :)

common/Reducers.jsm
```diff
@@ -484,6 +484,33 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
...
+        return prevState;
+      }
+
+      return {
+        ...prevState,
+        spocs: {
+          ...prevState.spocs,
+          data: prevState.spocs.data.spocs ? {
+            spocs: prevState.spocs.data.spocs.filter(s => s.url !== action.data.url),
+          } : {},
```
k88hudson: It seems like weird behaviour that that we'd go from `spocs.data.spocs` being falsey to `spocs.data.spocs` being `{}`

common/Reducers.jsm
```diff
@@ -484,6 +484,33 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
...
+          ...prevState.feeds,
+          data: Object.keys(prevState.feeds.data).reduce((accumulator, feed_url) => {
+            accumulator[feed_url] = {
+              data: {
+                ...prevState.feeds.data[feed_url].data,
+                recommendations: prevState.feeds.data[feed_url].data.recommendations.filter(r => r.url !== action.data.url),
+              },
+            };
+            return accumulator;
+          }, {}),
```
k88hudson: Does blocking an item cause the spocs to be recalculated in `selectLayoutRender.js` and if so, is that the expected behaviour? In other words should this count as a "refresh" the data?

If the expectation is that we *don't* cause a recalculation, you could consider using a flag that tracks whether any changes were made during `filter`.

ncloudioj:
Hi @punamdahiya, there are a few tweaks needed for `DSLinkMenu.jsx` in order to make the telemetry work with `link-menu-options.js`. Those changes are not in this PR, so I am writing them down here instead.  

* We can't use the constant `TOP_STORIES_SOURCE` in `DSLinkMenu.jsx`, because we now have various sources. i.e. "HERO", "LIST", and "CARDGRID". I believe you can get that from `this.props.type`, let's also make it uppercase by convention.
* Add `guid` to the `site` property in `LinkMenu.render()`. This is a required property for `link-menu-options.js`. You can simply pass the article ID, i.e. `rec.id`, when you render a DSLinkMenu, and then add it to the `site` as `guid: this.props.id`.
* For the `index` property of DSLinkMenu, let's use the `pos` from each article, i.e. `rec.pos` or `heroRec.pos`, which is the actual action position other tha `this.props.index`.

Let me know if you have any further questions. Also, feel free to tag me for another scan once those changes get fixed.

Thanks!

common/Reducers.jsm
punamdahiya: Goal here is to reset spocs.data back to initial state of {} ( check prevState.spocs.data.spocs is at spocs.data level, I should indent to make it readable)

common/Reducers.jsm
```diff
@@ -484,6 +484,33 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
           spocs_endpoint: action.data || INITIAL_STATE.DiscoveryStream.spocs.spocs_endpoint,
         },
       };
+    case at.PLACES_LINK_BLOCKED:
+      // Return if action data is empty, or spocs or feeds data is not loaded
+      if (!action.data || !prevState.spocs.loaded || !prevState.feeds.loaded) {
+        return prevState;
+      }
+
+      return {
```
gvn: I feel like this block could use a comment to explain what is going on since it's fairly dense.

gvn:
Are there repro steps or is this not going to QA?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -218,6 +217,13 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     };
   }
 
+  filterRecommendations(feed) {
+    if (feed && feed.data && feed.data.recommendations && feed.data.recommendations.length) {
+      return {data: this.filterBlocked(feed.data, "recommendations")};
+    }
+    return feed;
```
gvn: Maybe wrap this in an `else` for clarity?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -313,11 +320,22 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
 
+  filterBlocked(data, type) {
+    if (data && data[type] && data[type].length) {
+      const filteredItems = data[type].filter(item => !NewTabUtils.blockedLinks.isBlocked({"url": item.url}));
+      return {
+        ...data,
+        [type]: filteredItems,
+      };
+    }
+    return data;
```
gvn: Maybe wrap this in an else for clarity?

punamdahiya:
> Are there repro steps or is this not going to QA?

Updated with QA steps https://bugzilla.mozilla.org/show_bug.cgi?id=1531585#c3

lib/DiscoveryStreamFeed.jsm
punamdahiya: unfortunately doing so fails eslint(no-else-return)

common/Reducers.jsm
punamdahiya: As discussed, any change in state including adding filter flag will cause selectLayoutRender to run. Considering the ask is to remove dismissed item from all components rendered on new tab, it makes sense to update the state but we should fix the [issue raised here](https://github.com/mozilla/activity-stream/pull/4837#pullrequestreview-214090555) in a separate bug

punamdahiya:
@ncloudioj Thanks for the feedback. Updated PR with suggested telemetry changes. Please review

common/Reducers.jsm
punamdahiya: Done

punamdahiya:
@k88hudson @gvn Updated PR with review feedback and discussed changes. Thanks!

lib/DiscoveryStreamFeed.jsm
gvn: Ah...well I guess it's fine in that case.

gvn:
This works well for me locally, so as long as everyone else's concerns are addressed then it's R+ from me! Thanks!

ncloudioj:
@punamdahiya Feel free to do those [telemetry tweaks](https://github.com/mozilla/activity-stream/pull/4837#issuecomment-472437381) in a follow up patch :)

ncloudioj:
The telemetry bit looks good too, thanks! r=nanj!

punamdahiya:
Landing as TCI failures unrelated to the patch

punamdahiya:
> I noticed this causes a bug that's not triggered right now.
> 
> Right no spocs roll based on a percentage. If that percentage is always 100%, we always display a spoc.
> 
> This change causes a case where blocking a non spoc, causes a reroll of 100% so that it always ends up displaying the same thing.
> 
> If the roll isn't always 100% (server changes in the future) This would have a bug where blocking a nonn spoc would cause existing spocs to appear and disappear.
> 
> I think we can file that, land this, and run this decision by Nick that we have this potential bug.

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1535401

PR #4839, opened 2019-03-08
CFR pinned tabs feature branch

k88hudson:


Mardak:
A quick check in #l20n seems like Kwan ran into something similar when working on https://phabricator.services.mozilla.com/D14905 I'll see if a bug was filed, but seems like we should be okay with this…

PR #4843, opened 2019-03-12
Bug 1508669 - Update dev dependencies

piatra:
Major version changes:
* @octokit/rest https://github.com/octokit/rest.js/releases/tag/v16.0.1
* eslint-watch https://github.com/rizowski/eslint-watch/releases/tag/v5.0.1
* karma https://github.com/karma-runner/karma/releases/tag/v4.0.0
* mocha https://github.com/mochajs/mocha/releases/tag/v6.0.0
* raw-loader https://github.com/webpack-contrib/raw-loader/releases/tag/v1.0.0

TODO:
- [x] Create bug for future update. https://bugzilla.mozilla.org/show_bug.cgi?id=1535328


piatra:
I'll do the `babel-*` packages separately. The new version requires an update for plugins as well and it doesn't look like a drop in replacement.

dmose:
Shouldn't there be accompanying package-lock.json changes?

piatra:
Rebased on top of latest backport to fix TC.

k88hudson:
(also please remember to file the follow-up bug)

piatra:
Thanks. Will merge this with https://github.com/mozilla/activity-stream/pull/4845 and then land.
Followup bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1535328

dmose:
>     * that people update npm (6.9.0) via `npm install -g npm@latest` if they haven't recently

Can you unpack this?  Ed and I recently spent some time going around on this in #4807 when I landed the React updates recently, and we ended up sticking with 6.4.1 to land those changes precisely because they added gunk into package-lock.json that wouldn't go away, even if node_modules was removed.  Now the primary reason there, I think, was to make the export cleaner and easier for relman folks, and I'm not sure that's entirely necessary at this point in time.

That said, I'd like to be sure we're all on the same page here, and at this point the motivation behind 6.9.0 is not yet clear to me.

Yarn doesn't have this problem, as it actually allows for specification of which version of yarn is required, but c'est la vie.  :-)

dmose:
As it turns out, once we decide on what we want to do here, we can actually enforce it: 

https://gist.github.com/dmose/e9ea296c42836cfee8dca580bbb547aa

k88hudson:
@dmose We can stick to `6.4.1` if it makes things simpler, I just didn't notice a different in `package-lock` so I thought it wouldn't be an issue.

dmose:
Let's stick to 6.4.1 for now and list it in package-json so that at least we can figure out what version we're all supposed to be agreeing on in the future.  Please also file a bug on pinning it (i.e. adding the .npmrc), and mention in your email that we're looking at hard-pinning the versions of node and npm in bug X, but haven't yet decided exactly how.  This way, other devs get some warning and a chance to weigh in rather than having this sprung on them as a surprise, since it will likely be disruptive for some.

PR #4845, opened 2019-03-13
Bug 1536115 - Update babel dev-dependencies

piatra:
Follow up to #4843 to update babel-* dependencies and plugins.


piatra:
Changing this from draft to PR in order to trigger Travis and make sure automation is ✅

.travis.yml
```diff
@@ -7,6 +7,9 @@ node_js:
 python:
   - "2.7"
 
+addons:
+  firefox: "latest-nightly"
+
```
dmose: I'm guessing that we're currently testing against firefox release, meaning that passing our tests suggests that whatever code we've got is likely to be upliftable to release.   I'm also guessing that this switches us from release to nightly, meaning we lose early feedback about landing stuff that's not upliftable, but we gain confidence that this will work right on nightly.  Is that correct?

Assuming it is, it's not obvious to me which of these choices is the right tradeoff (in some ideal world, we'd test against both, and maybe beta too, but that's probably for another day...).

What motivated this change here?

Whatever we choose, I think we want a comment explaining the tradeoffs and the choice made in the yml file.
That might be the right tradefo

karma.mc.config.js
```diff
@@ -89,12 +89,11 @@ module.exports = function(config) {
             use: [{
               loader: "babel-loader", // require("babel-core")
               options: {
+                presets: [["@babel/preset-env", {"targets": {"firefox": "60"}}]],
```
dmose: What's the motivation for choosing version 60 here?  That probably wants a comment too.

I notice that the presets seem to be duplicated in a several places, perhaps this wants to be hoisted to a single const that's reused in various places?

package.json
```diff
@@ -17,40 +17,39 @@
     "reselect": "4.0.0"
   },
   "devDependencies": {
-    "@octokit/rest": "15.17.0",
-    "acorn": "6.0.4",
-    "babel-core": "6.26.3",
+    "@babel/core": "^7.3.4",
+    "@babel/preset-env": "^7.3.4",
+    "@babel/preset-react": "^7.0.0",
```
dmose: These semvers don't seem to follow our standard of pinning to a specific single version, so they probably want to be changed.

On a related note, it appears that this patch includes a bunch of changes from the other dev-dependencies patch.  Is there an intended landing order here?

package.json
```diff
@@ -59,23 +58,23 @@
...
+    "sass": "1.17.2",
     "sass-lint": "1.12.1",
-    "shelljs": "0.8.2",
+    "shelljs": "0.8.3",
     "simple-git": "1.107.0",
-    "sinon": "7.1.1",
-    "webpack": "4.25.1",
-    "webpack-cli": "3.1.2",
+    "sinon": "7.2.7",
+    "webpack": "^4.29.6",
```
dmose: Here's another not-so-precise version.

.travis.yml
piatra: Travis uses Firefox 56 by default which is missing some Javascript API features (we've hit this before like in [this comment](https://github.com/mozilla/activity-stream/blob/ac76f59c25e8561030217df099636887a39cd690/test/unit/asrouter/CFRPageActions.test.js#L117-L118)).
While removing some of the babel plugins I ran into the same issue.
I thought to set this to `latest-nightly` because that's where our code will land.

piatra: Just to clarify: `latest-nightly` is not needed, it could just as well be `stable`, it was just the fact that we land in Nightly.

package.json
piatra: Yes this is supposed to land after #4843.

package.json
piatra: I'll update these to a fixed version.

piatra:
TC failing on a backport that needs to happen other that it's ok to re-review.

.travis.yml
dmose: @k88hudson do you have an opinion on what version we should use here?  I'm not actually sure how much it matters what we choose here, are you aware of us having hit any problems that would have been caught earlier had we been using a newer version on travis?  If not, maybe this isn't worth overthinking.

k88hudson: Personally I think it makes sense to use `latest-nightly` since we want to be in line with what firefox tests would run against (and presumably if we run tests in tree, that's what they would run against)

dmose: That sounds reasonable to me.  @piatra let's go with that, and add a comment explaining the reasoning please.

dmose:
I'd prefer not squashing this into the first dev-dependencies patch.  I'd like to look at this again after the main thing lands, then I'll be able to be sure that I'm looking at just the babel-related changes.

karma.mc.config.js
```diff
@@ -132,6 +117,9 @@ module.exports = function(config) {
             enforce: "post",
             test: /\.jsm?$/,
             loader: "istanbul-instrumenter-loader",
+            query: {
+              esModules: true,
+            },
```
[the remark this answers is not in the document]
dmose: FWIW, https://webpack.js.org/concepts/loaders/#loader-features suggests that `query` has been deprecated in favor of using `options`.  I don't think it's super important that you fix this before landing, but fixing it would make the configuration more readable.

test/unit/common/Reducers.test.js
```diff
@@ -819,4 +819,12 @@ describe("Reducers", () => {
       assert.propertyVal(nextState, "hide", false);
     });
   });
+  describe("ASRouter", () => {
+    it("should listen for pref changes on AS_ROUTER_PREF_CHANGED", async () => {
+      const action = {data: {foo: "bar"}, type: at.AS_ROUTER_PREF_CHANGED};
+      const result = ASRouter({}, action);
```
dmose: Ideally, there would be a newline between the `action` and `result` lines, to follow the semi-convention of splitting tests into

* setup
* execute
* verify

chunks

karma.mc.config.js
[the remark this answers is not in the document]
piatra: Updated to use `options`. 

https://github.com/webpack-contrib/istanbul-instrumenter-loader#with-babel

piatra:
> Looks good to me. The only changes that I've flagged here are nits.
> 
> An interesting side-effect of this change is that ASRouter had been spamming the browser console with a useless error message, and now this error message actually has a partial stack:
> 
> 15:13:53.568 (intermediate value).messages is undefined ASRouter.jsm:116
> _remoteLoader resource://activity-stream/lib/ASRouter.jsm:116
> InterpretGeneratorResume self-hosted:1284
> AsyncFunctionNext self-hosted:839

This is a known issue with the snippets backend (it's because there are no messages for nightly builds and the payload is empty).

https://bugzilla.mozilla.org/show_bug.cgi?id=1523924

PR #4848, opened 2019-03-13
Bug 1517303 - Add tests to verify CFR & targeting

rlr:
@k88hudson r? was this what you had in mind or was there more? These tests assume that the trigger is setting up the context properly.

### DIFF
--- a/test/browser/browser_asrouter_targeting.js
+++ b/test/browser/browser_asrouter_targeting.js
@@ -2,6 +2,7 @@ const {ASRouterTargeting, QueryCache} =
   ChromeUtils.import("resource://activity-stream/lib/ASRouterTargeting.jsm");
 const {AddonTestUtils} =
   ChromeUtils.import("resource://testing-common/AddonTestUtils.jsm");
+const {CFRMessageProvider} = ChromeUtils.import("resource://activity-stream/lib/CFRMessageProvider.jsm");
 ChromeUtils.defineModuleGetter(this, "ProfileAge",
   "resource://gre/modules/ProfileAge.jsm");
 ChromeUtils.defineModuleGetter(this, "AddonManager",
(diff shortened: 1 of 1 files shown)


test/browser/browser_asrouter_targeting.js
```diff
@@ -434,3 +435,41 @@ add_task(async function check_pinned_tabs() {
...
+
+  trigger.context.recentVisits.push({timestamp: timeMinutesAgo(59)});
+  is(await ASRouterTargeting.findMatchingMessage({messages, trigger}), messages.find(m => m.id === "PIN_TAB"),
+    "should select PIN_TAB mesage");
+
+  await BrowserTestUtils.withNewTab({gBrowser, url: "about:blank"}, async browser => {
+    let tab = gBrowser.getTabForBrowser(browser);
+    gBrowser.pinTab(tab);
+    is(await ASRouterTargeting.findMatchingMessage({messages, trigger}), undefined,
+      "should not select PIN_TAB mesage with only 2 visits in past hour");
```
piatra: I think the message description here should be updated to mention the pinned tab?

test/browser/browser_asrouter_targeting.js
```diff
@@ -434,3 +435,41 @@ add_task(async function check_pinned_tabs() {
...
+  });
+
+  trigger.param = "foo.bar";
+  is(await ASRouterTargeting.findMatchingMessage({messages, trigger}), undefined,
+    "should not select PIN_TAB mesage with a param/host not in our list");
```
piatra: Also here the trigger is not related to the hostlist.

test/browser/browser_asrouter_targeting.js
```diff
@@ -434,3 +435,41 @@ add_task(async function check_pinned_tabs() {
...
+
+  // Cleanup
+  await clearHistoryAndBookmarks();
```
piatra: Is this required? I think you're only adding visits in the local `trigger` object.

test/browser/browser_asrouter_targeting.js
```diff
@@ -434,3 +435,41 @@ add_task(async function check_pinned_tabs() {
...
+  is(await ASRouterTargeting.findMatchingMessage({messages, trigger}), undefined,
+    "should not select PIN_TAB mesage with only 2 visits in past hour");
+
+  trigger.context.recentVisits.push({timestamp: timeMinutesAgo(59)});
+  is(await ASRouterTargeting.findMatchingMessage({messages, trigger}), messages.find(m => m.id === "PIN_TAB"),
+    "should select PIN_TAB mesage");
+
+  await BrowserTestUtils.withNewTab({gBrowser, url: "about:blank"}, async browser => {
+    let tab = gBrowser.getTabForBrowser(browser);
+    gBrowser.pinTab(tab);
```
piatra: You should probably unpin at the end of the test or do it a `registerCleanupFunction`.

test/browser/browser_asrouter_targeting.js
rlr: doh. lots of copy/paste fails here. fixed.

PR #4852, opened 2019-03-15
Bug 1534934 - Investigate multiple CFR triggers being called at the s…

piatra:
…ame time

I've made it so that it rejects a second `addRecommendation` call for the same browser. That map entry [is removed](https://github.com/mozilla/activity-stream/blob/348fd7ec68f44311d5da64af29a2fbb147c417a7/lib/CFRPageActions.jsm#L530) when the user navigates away.
I've kept the `_DUMMY` messages in case it might useful to test this in the review process but I didn't end up using them for mochitests so I want to remove before landing.

### DIFF
--- a/test/unit/asrouter/CFRPageActions.test.js
+++ b/test/unit/asrouter/CFRPageActions.test.js
@@ -1,4 +1,5 @@
 import {CFRPageActions, PageAction} from "lib/CFRPageActions.jsm";
+import {FAKE_RECOMMENDATION} from "./constants";
 import {GlobalOverrider} from "test/unit/utils";
 
 describe("CFRPageActions", () => {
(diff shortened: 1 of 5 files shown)


lib/CFRMessageProvider.jsm
```diff
@@ -354,6 +354,99 @@ const CFR_MESSAGES = [
     frequency: {lifetime: 3},
     trigger: {id: "frequentVisits", params: PINNED_TABS_TARGET_SITES},
   },
+  {
```
k88hudson: You could actually maybe add this to the `SnippetsTestProvider` (although perhaps renaming it to make it more generic would make sense)

k88hudson: Although I guess we wouldn't want this to always show up during regular debugging...

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -790,6 +790,40 @@ describe("CFRPageActions", () => {
...
+          // Adding failed
+          assert.isFalse(messageAdded);
+          // First message is still there
+          assert.deepInclude(CFRPageActions.RecommendationMap.get(fakeBrowser), {
+            id: fakeRecommendation.id,
+            host: fakeHost,
+            content: fakeRecommendation.content,
+          });
+        });
+        it("should send impressions just for the first message", async () => {
```
k88hudson: I think it would be more definitive to check the actual state of the last provider impression / to run the frequency cap function than just inspecting the messages that get sent, although that would probably be a lot more difficult.

PR #4853, opened 2019-03-15
Bug 1534278 –  Remove Extra borders on Hero cards on Dark theme

gvn:
Repro: https://bugzilla.mozilla.org/show_bug.cgi?id=1534278#c0

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/Hero/_Hero.scss
+++ b/content-src/components/DiscoveryStreamComponents/Hero/_Hero.scss
@@ -72,7 +72,7 @@ $card-header-in-hero-line-height: 20;
     padding: 16px 0;
     height: 100%;
 
-    @at-root .ds-hero-no-border .wrapper {
+    @at-root .ds-hero-no-border .ds-hero-item .wrapper {
       border-top: 0;
       border-bottom: 0;
       padding: 0 0 8px;
--- a/content-src/components/DiscoveryStreamComponents/Hero/_Hero.scss
+++ b/content-src/components/DiscoveryStreamComponents/Hero/_Hero.scss
@@ -22,11 +22,6 @@ $card-header-in-hero-line-height: 20;
     margin: 4px 0 8px;
   }
 
-  .ds-list {
-    border-top: 0;
-    padding-top: 0;
-  }
-
   .ds-card {
     border: 0;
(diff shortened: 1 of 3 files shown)


content-src/styles/_mixins.scss
```diff
@@ -17,7 +17,7 @@
 }
 
 @mixin dark-theme-only {
-  [lwt-newtab-brighttext]:not(.force-light-theme) & {
```
gvn: `force-light-theme` appears to be deprecated.

punamdahiya: yes with [Enable Dark Mode](https://github.com/mozilla/activity-stream/commit/2225e93564508e4ace4e69f2cd2b735fb3c5ee59#diff-3f95337718772ca7ed0b5acae1358dd0) , we are no longer using force-light-theme and should remove it from rest of the places where as its used.

https://searchfox.org/mozilla-central/search?q=force-light-theme&path=
https://github.com/mozilla/activity-stream/commit/8f3f2f6a4660d8bf557e9823ec2d7e78415775b4

punamdahiya: @gvn as this is follow up of enabling dark mode,  can we do this cleanup as part of this PR?

content-src/components/DiscoveryStreamComponents/Hero/_Hero.scss
```diff
@@ -61,18 +56,17 @@ $card-header-in-hero-line-height: 20;
...
-    @include ds-border-bottom;
     @include dark-theme-only {
       color: $grey-30;
     }
 
     color: $grey-50;
     display: block;
     margin: 12px 0 16px;
-    padding: 16px 0;
+    padding: 16px 0 0;
```
punamdahiya: nit: we can switch to padding-top here

content-src/styles/_mixins.scss
gvn: I would prefer to do that as a separate ticket, but I will file it!

gvn:
My guess is we can probably get rid of the dark mode message component as well as part of the force-light-theme cleanup.

content-src/styles/_mixins.scss
punamdahiya: sounds good, thanks!

PR #4857, opened 2019-03-18
Bug 1535464 - Update CFR addon Learn More links

piatra:
When we fetch addon information [we only replace the install url](https://github.com/mozilla/activity-stream/blob/7b98ad991489a79e31c20378ea49eefe39737917/lib/OnboardingMessageProvider.jsm#L216). `amo_url` is used for the [`Learn More`](https://github.com/mozilla/activity-stream/blob/7b98ad991489a79e31c20378ea49eefe39737917/lib/CFRPageActions.jsm#L373) link in the dropdown panel.
AMO takes care of redirecting the user to the correct locale for the page based on the browser language settings. We just need to drop `en-US` from the URL.

### DIFF
--- a/lib/CFRMessageProvider.jsm
+++ b/lib/CFRMessageProvider.jsm
@@ -66,7 +66,7 @@ const CFR_MESSAGES = [
         rating: 4.6,
         users: 299019,
         author: "Mozilla",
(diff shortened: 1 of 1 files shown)


k88hudson:
I totally forgot about the learn more link connection, thanks.

This would be worth flagging QA to manually check a couple of other locales.

PR #4858, opened 2019-03-19
Bug 1535401 - Fix spocs display based off probability

punamdahiya:


content-src/lib/selectLayoutRender.js
```diff
@@ -1,79 +1,70 @@
...
-  // any of these values change.
-  [
-    state => state.DiscoveryStream.layout,
-    state => state.DiscoveryStream.feeds,
-    state => state.DiscoveryStream.spocs,
-  ],
-
-  // Adds data to each component from feeds. This function only re-runs if one of the inputs change.
-  // TODO: calculate spocs
-  function layoutRender(layout, feeds, spocs) {
```
punamdahiya: Only chnage here is removing selector specific code. layoutRender function stay as is and renamed to selectLayoutRender with arguments as DiscoveryStream redux state and array of cached random values

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -158,8 +159,15 @@ export class _DiscoveryStreamBase extends React.PureComponent {
     return (<style key={json} data-styles={json} ref={this.onStyleMount} />);
   }
 
+  componentWillReceiveProps(oldProps) {
+    if (this.props.DiscoveryStream.layout !== oldProps.DiscoveryStream.layout) {
+      rickRollCache = [];
```
punamdahiya: Resetting cached random values on layout refresh

content-src/lib/selectLayoutRender.js
```diff
@@ -1,79 +1,70 @@
...
-          }
+  function maybeInjectSpocs(data, spocsConfig) {
+    if (data &&
+        spocsConfig && spocsConfig.positions && spocsConfig.positions.length &&
+        spocs.data.spocs && spocs.data.spocs.length) {
+      const recommendations = [...data.recommendations];
+      for (let position of spocsConfig.positions) {
+        // Cache random number for a position
+        if (!rickRollCache[position.index]) {
+          rickRollCache[position.index] = Math.random();
```
punamdahiya: Cache random value for a spoc position, so that it can be used on next render caused by a user interaction such as dismiss. We are resetting this cache on page refresh.

punamdahiya:
@ScottDowne @k88hudson Submitting patch to resolve spocs item not honoring probability values other than 0 and 1. Patch caches random values for a spoc position and uses it in subsequent render. Thanks!

content-src/lib/selectLayoutRender.js
```diff
@@ -1,79 +1,70 @@
...
-            recommendations.splice(position.index, 0, spocs.data.spocs[spocIndex++]);
-          }
+  function maybeInjectSpocs(data, spocsConfig) {
+    if (data &&
+        spocsConfig && spocsConfig.positions && spocsConfig.positions.length &&
+        spocs.data.spocs && spocs.data.spocs.length) {
+      const recommendations = [...data.recommendations];
+      for (let position of spocsConfig.positions) {
+        // Cache random number for a position
+        if (!rickRollCache[position.index]) {
```
ScottDowne: Is position.index unique? (I forget)

ScottDowne: I think a new component could in theory have a new set of positions with the same index as before.

punamdahiya: yes for a specific component picked from layout , it should be unique.
```
{
          "type": "CardGrid",
          "properties": {
            "items": 3
          },
          "header": {
            "title": ""
          },
          "feed": {
            "embed_reference": null,
            "url": "https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?..."
          },
          "spocs": {
            "probability": 1,
            "positions": [
              {
                "index": 2
              }
            ]
          }
        },


```

punamdahiya: > I think a new component could in theory have a new set of positions with the same index as before.
yes, in case of  a layout with multiple components say comp 1 and comp 2., comp 2 can have spoc position with same index as comp 1 and will be using the cached random value .

ScottDowne: This is probably not ideal though?

punamdahiya: My understanding spoc item should hide and display based off probability passed in the feed. 
For layout with multiple components, spoc items in different components displayed at the same position will show/hide based off spocs config probability. 

The only caveat is if spoc items in different components has same config probability and same config position , they will always show and hide at the same time (which can happen in random cases even with existing implementation in master). 

For  Hero, List and Card Grid with different number of items, whats the chance for spoc items showing up at same position and with same probability. May be  we should run it by Product and get their take on it. Is it worth supporting this use case by saving rick rolls by component and spoc positions.

ScottDowne: I don't think I follow what you mean by same config probability?

punamdahiya: > I don't think I follow what you mean by same config probability?

It's spocs -> probability received in layout for a respective component https://github.com/mozilla/activity-stream/pull/4858#discussion_r267532491

content-src/lib/selectLayoutRender.js
```diff
@@ -2,18 +2,29 @@ export const selectLayoutRender = (state, rickRollCache) => {
...
       for (let position of spocsConfig.positions) {
         // Cache random number for a position
-        if (!rickRollCache[position.index]) {
-          rickRollCache[position.index] = Math.random();
+        let rickRoll;
+        if (isFirstRun) {
+          rickRoll = Math.random();
+          rickRollCache.push(rickRoll);
+        } else {
+          rickRoll = rickRollCache.pop();
```
ScottDowne: I think if you used `array.shift();` here instead of pop, you wouldn't need the `rickRollCache.reverse();` above. I *think*

content-src/lib/selectLayoutRender.js
```diff
@@ -2,18 +2,29 @@ export const selectLayoutRender = (state, rickRollCache) => {
   const {layout, feeds, spocs} = state;
   let spocIndex = 0;
 
+  // rickRollCache stores random probability values for each spoc position. This cache is empty
+  // on page refresh and gets filled with random values on first render inside maybeInjectSpocs.
+  const isFirstRun = !rickRollCache.length;
```
ScottDowne: :+1:

ScottDowne: I think the buffer array could be stored in this local scope here, and not in base.

content-src/lib/selectLayoutRender.js
ScottDowne: After grabbing a reference to the roll here, I think you should also push it to the buffer array. Then when you finish your final maybeInjectSpocs you just reset the cache array (which should now be empty again) with the buffer that's in this local scope (which should now be full)

content-src/lib/selectLayoutRender.js
punamdahiya: done

punamdahiya:
> Some thoughts. I think we're close.

@ScottDowne thanks for review feedback, it looks much neater.

content-src/lib/selectLayoutRender.js
```diff
@@ -1,79 +1,83 @@
...
+        // Cache random number for a position
+        let rickRoll;
+        if (isFirstRun) {
+          rickRoll = Math.random();
+          rickRollCache.push(rickRoll);
+        } else {
+          rickRoll = rickRollCache.shift();
+          bufferRollCache.push(rickRoll);
+          if (!rickRollCache.length) {
+            rickRollCache.push(...bufferRollCache);
```
ScottDowne: Possible to do the above check and push right after the `maybeInjectSpocs` call on line 66?

PR #4859, opened 2019-03-19
Bug 1530359 - Handle URL redirects for CFR triggers

piatra:
The way I'm handling redirects is to additionally check the original request host against the hosts defined in the message.
For example if `de.quora.com` fails the initial check but we initially navigated to `quora.com` then the check will now pass.
This doesn't mean that `bar.mozilla.com` will match for `mozilla.com` because we check that `(originalLocation !== location)` before trying the second lookup so it's only for redirects when this check would return true.

### DIFF
--- a/lib/ASRouterTriggerListeners.jsm
+++ b/lib/ASRouterTriggerListeners.jsm
@@ -152,7 +152,12 @@ this.ASRouterTriggerListeners = new Map([
     },
 
     onLocationChange(aBrowser, aWebProgress, aRequest, aLocationURI, aFlags) {
-      let host = aLocationURI ? aLocationURI.host : "";
+      let host;
+      try {
+        host = aLocationURI ? aLocationURI.host : "";
+      } catch (e) { // about: pages will throw errors
+        return;
+      }
       // Some websites trigger redirect events after they finish loading even
       // though the location remains the same. This results in onLocationChange
       // events to be fired twice.
@@ -273,7 +278,12 @@ this.ASRouterTriggerListeners = new Map([
     },
 
     onLocationChange(aBrowser, aWebProgress, aRequest, aLocationURI, aFlags) {
-      let host = aLocationURI ? aLocationURI.host : "";
+      let host;
+      try {
+        host = aLocationURI ? aLocationURI.host : "";
+      } catch (e) { // about: pages will throw errors
+        return;
+      }
       // Some websites trigger redirect events after they finish loading even
       // though the location remains the same. This results in onLocationChange
       // events to be fired twice.
--- a/lib/ASRouterTriggerListeners.jsm
+++ b/lib/ASRouterTriggerListeners.jsm
@@ -38,19 +38,19 @@ function isPrivateWindow(win) {
  * Additionally verify for redirects and check original request URL against
  * the whitelist.
  *
- * @returns {string} - the host of the URL that matched
+ * @returns {string} - the host that matched the whitelist
  */
-function checkHost(location, hosts, aRequest) {
+function checkHost(aLocationURI, hosts, aRequest) {
   // Check current location against whitelisted hosts
-  if (hosts.has(location.host)) {
-    return true;
+  if (hosts.has(aLocationURI.host)) {
+    return aLocationURI.host;
   }
 
   // The original URL at the start of the request
   const originalLocation = aRequest.QueryInterface(Ci.nsIChannel).originalURI;
   // We have been redirected
-  if (originalLocation.spec !== location.spec) {
-    return hosts.has(originalLocation.host);
+  if (originalLocation.spec !== aLocationURI.spec) {
+    return hosts.has(originalLocation.host) && originalLocation.host;
   }
 
   return false;
@@ -152,13 +152,14 @@ this.ASRouterTriggerListeners = new Map([
     },
 
     onLocationChange(aBrowser, aWebProgress, aRequest, aLocationURI, aFlags) {
-      const host = aLocationURI ? aLocationURI.host : "";
+      let host = aLocationURI ? aLocationURI.host : "";
       // Some websites trigger redirect events after they finish loading even
       // though the location remains the same. This results in onLocationChange
       // events to be fired twice.
       const isSameDocument = !!(aFlags & Ci.nsIWebProgressListener.LOCATION_CHANGE_SAME_DOCUMENT);
       if (host && aWebProgress.isTopLevel && !isSameDocument) {
-        if (checkHost(aLocationURI, this._hosts, aRequest)) {
+        host = checkHost(aLocationURI, this._hosts, aRequest);
+        if (host) {
           this.triggerHandler(aBrowser, host);
         }
       }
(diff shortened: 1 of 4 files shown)


k88hudson:
Another thing you could consider is adding a ` onStateChange` listener and checking `aStateFlags & Ci.nsIWebProgressListener.STATE_REDIRECTING` / `aRequest.originalURL`, although I'm not sure what's better v.s. just looking at `aRequest.QueryInterface(Ci.nsIChannel).originalURI` every time. I'd go with whatever is the least expensive if there's a meaningful difference.

Also this is kinda unrelated but I happened to notice – is there a reason we need to create a new `URL` object instead of just using `aLocationURI.host` (and `originalLocation.host` for that matter)?

piatra:
I added a `onStateChange` handler and attempted the same thing. Some things I noticed:
* `onStateChange` is fired after `onLocationChange`
  * this was surprising, I expected the other way around and planned on using `onStateChange` to set some state that we could later use in `onLocationChange`
* `onStateChange` also fires for domains that redirect to and from the `www.` version of the website
  * `foo.com` redirects to `www.foo.com`
  * we would risk calling the `triggerHandler` twice in this scenario
* `aStateFlags & Ci.nsIWebProgressListener.STATE_REDIRECTING` was never `true` for `quora.com` even though the request got redirected
  * `aRequest.originalURL` was different from `aRequest.URI` and this is what I used to confirm the redirect attempt

The current implementation that checks for original vs current URL doesn't seem to have a significant cost to it. I also updated `checkHost` to not create the extra URL objects and reuse the `aLocation` arguments (the profiler screenshot was taken before)
image

lib/ASRouterTriggerListeners.jsm
```diff
@@ -33,6 +33,29 @@ function isPrivateWindow(win) {
   return !(win instanceof Ci.nsIDOMWindow) || win.closed || PrivateBrowsingUtils.isWindowPrivate(win);
 }
 
+/**
+ * Check current location against the list of whitelisted hosts
+ * Additionally verify for redirects and check original request URL against
+ * the whitelist.
+ *
+ * @returns {string} - the host of the URL that matched
```
k88hudson: This should be `@returns {boolean}`

lib/ASRouterTriggerListeners.jsm
```diff
@@ -33,6 +33,29 @@ function isPrivateWindow(win) {
...
+ */
+function checkHost(location, hosts, aRequest) {
```
k88hudson: maybe name this `aLocationURI` for consistency?

lib/ASRouterTriggerListeners.jsm
```diff
@@ -252,18 +272,15 @@ this.ASRouterTriggerListeners = new Map([
...
       // events to be fired twice.
       const isSameDocument = !!(aFlags & Ci.nsIWebProgressListener.LOCATION_CHANGE_SAME_DOCUMENT);
-      if (location && aWebProgress.isTopLevel && !isSameDocument) {
-        try {
-          const host = (new URL(location)).hostname;
-          if (this._hosts.has(host)) {
-            this._triggerHandler(aBrowser, {id: "openURL", param: host});
-          }
-        } catch (e) {} // Couldn't parse location URL
+      if (host && aWebProgress.isTopLevel && !isSameDocument) {
```
k88hudson: You could combine these if statements if you want

lib/ASRouterTriggerListeners.jsm
```diff
@@ -33,6 +33,29 @@ function isPrivateWindow(win) {
   return !(win instanceof Ci.nsIDOMWindow) || win.closed || PrivateBrowsingUtils.isWindowPrivate(win);
 }
 
+/**
+ * Check current location against the list of whitelisted hosts
+ * Additionally verify for redirects and check original request URL against
+ * the whitelist.
+ *
+ * @returns {string} - the host that matched the whitelist
```
piatra: I went back on this and now returning the `host` that matched the whitelist because we have an additional check 
https://github.com/mozilla/activity-stream/blob/b89bae21fd28625f3d70cadb4f1a404340aeb129/lib/ASRouterTargeting.jsm#L341
Where ensure that the selected message matches the trigger. If we would have passed down `aLocationURI.host` that could be `redirect.foo.com` and not match `foo.com` even if `checkHost` initially returned `true`.

lib/CFRMessageProvider.jsm
```diff
@@ -38,12 +38,10 @@ const REDDIT_ENHANCEMENT_PARAMS = {
   min_frecency: 10000,
 };
 const PINNED_TABS_TARGET_SITES = [
-  "docs.google.com", "www.docs.google.com", "calendar.google.com",
-  "messenger.com", "www.messenger.com", "web.whatsapp.com", "mail.google.com",
-  "outlook.live.com", "facebook.com", "www.facebook.com", "twitter.com", "www.twitter.com",
-  "reddit.com", "www.reddit.com", "github.com", "www.github.com", "youtube.com", "www.youtube.com",
-  "feedly.com", "www.feedly.com", "drive.google.com", "amazon.com", "www.amazon.com",
-  "messages.android.com",
+  "docs.google.com", "calendar.google.com",
```
piatra: Because we're only comparing hosts now I don't think we need the `www.` version anymore. Any thoughts on this?

k88hudson: `www.google.com` is a different host than `google.com` so I think you still need both (or, functionally speaking with this patch, whatever the redirection ends up at)

test/unit/asrouter/ASRouterTriggerListeners.test.js
```diff
@@ -178,10 +178,59 @@ describe("ASRouterTriggerListeners", () => {
...
+        const webProgress = {isTopLevel: true};
+        const aLocationURI = {host: "subdomain.mozilla.org", spec: "subdomain.mozilla.org"};
+        const aRequest = {
+          QueryInterface: sandbox.stub().returns({
+            originalURI: {spec: "subdomain.mozilla.org", host: "subdomain.mozilla.org"},
+          }),
+        };
+        openURLListener.onLocationChange(browser, webProgress, aRequest, aLocationURI);
+        assert.calledOnce(aRequest.QueryInterface);
+        assert.notCalled(newTriggerHandler);
```
k88hudson: Might be worth adding a `www` test specifically since someone else might forget that's a subdomain?

k88hudson: or maybe just add a comment near the actual list

piatra:
I added a test to make sure we don't try to remove it again.

test/unit/asrouter/CFRMessageProvider.test.js
```diff
@@ -41,4 +41,9 @@ describe("CFRMessageProvider", () => {
       }
     }
   });
+  it("should contain `www.` version of the hosts", () => {
```
k88hudson: :+1:

lib/ASRouterTriggerListeners.jsm
```diff
@@ -152,7 +152,12 @@ this.ASRouterTriggerListeners = new Map([
     },
 
     onLocationChange(aBrowser, aWebProgress, aRequest, aLocationURI, aFlags) {
-      let host = aLocationURI ? aLocationURI.host : "";
+      let host;
+      try {
+        host = aLocationURI ? aLocationURI.host : "";
+      } catch (e) { // about: pages will throw errors
```
k88hudson: Ah, good call.

PR #4863, opened 2019-03-21
Bug 1527500 - Save to pocket and bookmark menu options

punamdahiya:


common/Reducers.jsm
```diff
@@ -457,6 +457,182 @@ function Pocket(prevState = INITIAL_STATE.Pocket, action) {
...
+        ...state.feeds,
+        data: Object.keys(state.feeds.data).reduce((accumulator, feed_url) => {
+          accumulator[feed_url] = {
+            data: {
+              ...state.feeds.data[feed_url].data,
+              recommendations: state.feeds.data[feed_url].data.recommendations.map(item => {
+                if (item.url === action.data.url) {
+                  const {bookmarkGuid, bookmarkTitle, dateAdded} = action.data;
+                  return Object.assign({}, item, {
+                    bookmarkGuid,
```
Mardak: Just a quick glance.. this is the only place bookmarkGuid is set, i.e., when the bookmark is added? This won't persist after restarting firefox as that would probably want DiscoveryStreamFeed changes to query every single url, but this loss-on-restart behavior might be acceptable for the initial implementation and have a bug for later/backlog ?

common/Reducers.jsm
```diff
@@ -457,6 +457,182 @@ function Pocket(prevState = INITIAL_STATE.Pocket, action) {
...
+function DiscoveryStreamFeeds(state = {}, action) {
+  switch (action.type) {
+    case at.PLACES_LINK_BLOCKED:
+      return {
+        ...state.feeds,
+        data: Object.keys(state.feeds.data).reduce((accumulator, feed_url) => {
+          accumulator[feed_url] = {
+            data: {
+              ...state.feeds.data[feed_url].data,
+              recommendations: state.feeds.data[feed_url].data.recommendations.filter(r => r.url !== action.data.url),
```
Mardak: I didn't look that closely at all the others, but it seems like a higher order function that takes a function returning new recommendations might clean this up:
```js
case BLOCKED:
  return HOF(recs => recs.filter(r => …));
case SAVED:
  return HOF(recs => recs.map(item => …));
```

common/Reducers.jsm
punamdahiya: we have https://bugzilla.mozilla.org/show_bug.cgi?id=1531596 for items losing bookmarked state on restart. Default Activity Stream top stories are keeping bookmarked state on page refresh, will look and weigh in effort needed to port that support

lib/PlacesFeed.jsm
```diff
@@ -348,6 +349,32 @@ class PlacesFeed {
...
+      // console.log("On Rebookmark", data);
+      this.store.dispatch({type: at.PLACES_LINKS_CHANGED});
+      this.store.dispatch({
+        type: at.PLACES_BOOKMARK_ADDED,
+        data: {
+          bookmarkGuid: data.guid,
+          bookmarkTitle: data.title,
+          url,
+        },
+      });
```
punamdahiya: @mardak Took a shot at checking first if bookmark exists before calling addBookmark on NewtabUtils. As per places team, this check shouldn't be expensive call. On rebookmark, story shows 'Remove bookmark' state on page refresh, trying to figure why?

common/Reducers.jsm
Mardak: Here's what I was thinking from our chat earlier (totally untested! mostly just copy pasted stuff 😉 ):
```js
function DiscoveryStream(prevState, action) {
  const updateLinks = onLinks => {
      // Return if action data is empty, or spocs or feeds data is not loaded
      return !action.data || !prevState.spocs.loaded || !prevState.feeds.loaded ? prevState : {
        ...prevState,
        spocs: {
          ...prevState.spocs,
          data: prevState.spocs.data.spocs ? {spocs: onLinks(prevState.spocs.data.spocs)} : {},
        },
        feeds: {
          ...prevState.feeds,
          data: Object.keys(prevState.feeds.data).reduce((accumulator, feed_url) => {
            accumulator[feed_url] = {
              data: {
                ...prevState.feeds.data[feed_url].data,
                recommendations: onLinks(prevState.feeds.data[feed_url].data.recommendations),
              },
            };
            return accumulator;
          }, {});
        },
      };
  };

  switch(action.type) {
    case at.PLACES_LINK_BLOCKED:
      return updateLinks(recs => recs.filter(r => r.url !== action.data.url));
  }
}
```
This is assuming the same onLinks callback can be used for feed links and spocs.

lib/PlacesFeed.jsm
Mardak: Just noting from our chat, it doesn't update existing pages as we need to Broadcast the action. However, if we end up in the situation where we incorrectly show "Bookmark" action we'll handle that in https://bugzilla.mozilla.org/show_bug.cgi?id=1531596 so probably good to update that bug title and add your findings.

punamdahiya: Updated https://bugzilla.mozilla.org/show_bug.cgi?id=1531596#c1

punamdahiya:
@Mardak looks like we should add link to place utils history when bookmark is added https://searchfox.org/mozilla-central/source/browser/components/newtab/lib/SectionsManager.jsm#445

punamdahiya:
@ncloudioj Including you to help review telemetry part of the PR essentially helping validate required telemetry from  save to pocket and bookmark menu options in pocket new tab is working similar to activity stream. Thanks!

punamdahiya:
> @Mardak looks like we should add link to place utils history when bookmark is added https://searchfox.org/mozilla-central/source/browser/components/newtab/lib/SectionsManager.jsm#445

Please ignore, verified link is getting added to history using the handler in SectionsManager.jsm

punamdahiya:
@ScottDowne @Mardak PR looks good to review, the only issue open is https://bugzilla.mozilla.org/show_bug.cgi?id=1539904 and to be fixed in its own PR. Thanks

ncloudioj:
@punamdahiya The telemetry bit looks good, verified in the local build.

r=nanj 🚢

punamdahiya:
Updated with delete and archive icons received from Wolasi

Screenshot 2019-03-28 18 16 57

gvn:
Just flagging that Save to Pocket works, but I see no feedback in UI after I click it.

Is that expected?

punamdahiya:
> Just flagging that Save to Pocket works, but I see no feedback in UI after I click it.
> 
> Is that expected?
> 
Yes that's out of scope and expected
> Looks like current experience is that "Saved to Pocket" appears in the footer of the card (pretty subtle IMO):
>

gvn:
Ok, aside from that looks good and works well!

If we don't have a ticket to add UI feedback on save we probably should file one in the interest of keeping parity between DS and AS.

punamdahiya:
> add UI feedback on save

Done https://bugzilla.mozilla.org/show_bug.cgi?id=1540842

data/content/assets/glyph-pocket-archive-16.svg
```diff
@@ -0,0 +1 @@
+<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg"><path d="M1 14.667V4h14v10.667c0 .736-.597 1.333-1.333 1.333H2.333A1.333 1.333 0 0 1 1 14.667zM1 0h14a1 1 0 0 1 0 2H1a1 1 0 1 1 0-2zm9.341 7.247l-3.295 2.884-.839-.838a1 1 0 1 0-1.414 1.414l1.5 1.5a1 1 0 0 0 1.366.046l4-3.5a1 1 0 0 0-1.318-1.506z" fill="#000" fill-rule="evenodd"/></svg>
```
Mardak: The rest of the svgs use context fill instead of hardcoded values, so this will lead to unexpected behaviors:
Screen Shot 2019-04-03 at 11 56 03 AM

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1541543

punamdahiya: Good catch! Submitted PR with the fix https://github.com/mozilla/activity-stream/pull/4882

PR #4867, opened 2019-03-25
Bug 1530714 - Report ASRouter errors in ASRouterAdmin

piatra:
This adds a new section in ASRouterAdmin where you can see all logged errors.

image

I've broken down the `_remoteLoader` to report missing messages, response failures or failures to parse json. Also all calls to `Cu.reportError` now go through the same function that captures them to be displayed in the admin panel.

### DIFF
--- a/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -152,7 +152,6 @@ export class ASRouterAdminInner extends React.PureComponent {
     this.onCopyTargetingParams = this.onCopyTargetingParams.bind(this);
     this.onPasteTargetingParams = this.onPasteTargetingParams.bind(this);
     this.onNewTargetingParams = this.onNewTargetingParams.bind(this);
-    this.renderErrorMessage = this.renderErrorMessage.bind(this);
     this.state = {
       messageFilter: "all",
       evaluationStatus: {},
@@ -563,12 +562,8 @@ export class ASRouterAdminInner extends React.PureComponent {
       </div>);
   }
 
-  reportErrorToConsole(error) {
-    return () => console.log(error); // eslint-disable-line no-console
-  }
-
   renderErrorMessage({timestamp, error}) {
-    return (<tr onClick={this.reportErrorToConsole(error)}>
+    return (<tr>
       <td>{error.message}</td>
       <td>{relativeTime(timestamp)}</td>
     </tr>);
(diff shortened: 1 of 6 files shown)


lib/ASRouter.jsm
```diff
@@ -49,6 +49,16 @@ const STARTPAGE_VERSION = "6";
...
   STARTPAGE_VERSION,
   REMOTE_LOADER_CACHE_KEY: "RemoteLoaderCache",
+  _errors: [],
+
+  reportError(e) {
+    Cu.reportError(e);
+    this._errors = [{timestamp: new Date(), error: e}, ...this._errors];
+  },
+
+  get errors() {
```
k88hudson: I like the `reportError` strategy, but personally I would return the error through `loadMessagesForProvider` so the state can be centralized in `ASRouter._state`, or maybe even as a property inside provider in `ASRouter._state.providers` itself?

Keeping it here makes cleanup more complicated IMO

lib/ASRouter.jsm
```diff
@@ -104,17 +114,29 @@ const MessageLoaderUtils = {
...
-          (response.ok || response.status === 302)
-        ) {
-          remoteMessages = (await response.json())
-            .messages
+        response = await fetch(provider.url, {headers});
+      } catch (e) {
+        MessageLoaderUtils.reportError(e);
+      }
+      if (
+        response &&
```
k88hudson: This might be a little too strict, personally I usually go with `status >= 200 && status < 400`

piatra:
Updated the PR
* We now only report `remote` provider errors
* `errors` are part of the `provider` state
* reverted some of the `Cu.reportError` changes so we store provider errors and everything else is reported to the console
* reporting `provider` id now with the errors
image

lib/ASRouter.jsm
```diff
@@ -104,17 +116,26 @@ const MessageLoaderUtils = {
...
-          remoteMessages = (await response.json())
-            .messages
+        response = await fetch(provider.url, {headers, credentials: "omit"});
+      } catch (e) {
+        MessageLoaderUtils.reportError(e);
+      }
+      if (
+        response &&
+        (response.status >= 200 && response.status < 400) &&
+        (response.ok || response.status === 302)
```
k88hudson: You don't need `=== 302` if you have `<400`

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -562,6 +562,37 @@ export class ASRouterAdminInner extends React.PureComponent {
       </div>);
   }
 
+  renderErrorMessage({id, errors}) {
+    const providerId = <td rowSpan={errors.length}>{id}</td>;
+    // .reverse() so that the last error (most recent) is first
+    return errors.reverse().map(({error, timestamp}, cellKey) => (<tr key={cellKey}>
```
k88hudson: can you `.reverse` after `.map` (don't mutate input)

PR #4868, opened 2019-03-26
Bug 1530740 - Add loaded content ping for Discovery Stream

ncloudioj:
This adds a new ping for content loaded event in Discovery Stream. Similar to impression pings, it relies on the page visibility, but not the individual visibility.

NB:
* This needs a data review
* Requires a schema change in both old&new data pipeline.

### DIFF
--- a/content-src/components/DiscoveryStreamImpressionStats/ImpressionStats.jsx
+++ b/content-src/components/DiscoveryStreamImpressionStats/ImpressionStats.jsx
@@ -60,6 +60,35 @@ export class ImpressionStats extends React.PureComponent {
     }
   }
 
+  // This checks if the given cards are the same as those in the last loaded content ping.
+  // If so, it should not send the same loaded content ping again.
+  _needsLoadedContent(cards) {
+    if (!this.loadedContentGuids || (this.loadedContentGuids.length !== cards.length)) {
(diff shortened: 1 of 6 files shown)


content-src/components/DiscoveryStreamImpressionStats/ImpressionStats.jsx
```diff
@@ -60,6 +60,35 @@ export class ImpressionStats extends React.PureComponent {
     }
   }
 
+  // This checks if the given cards are the same as those in the last loaded content ping.
+  // If so, it should not send the same loaded content ping again.
+  _needsLoadedContent(cards) {
```
ScottDowne: Interesting, what's the idea behind this, what is it solving

ScottDowne: Like, are there multiple events being fired in one tab, or is this for opening multiple tabs and only tracking the first one?

ncloudioj:
Data review granted https://bugzilla.mozilla.org/show_bug.cgi?id=1530740#c5! 

Will merge after the data pipeline changes get deployed.

ncloudioj:
Pipeline changes are deployed. Merge it now.

PR #4869, opened 2019-03-27
Bug 1533359 - Targeting support for any subdomain

piatra:
Work on top of #4859, ~~which needs to land first.~~

One issue I ran into here: 
* ASRouterTargeting checks that the selected message has a `trigger param` that matches the `host`
  * https://github.com/mozilla/activity-stream/compare/master...piatra:bug1533359?expand=1#diff-60f0c109d5c368ec6a6b11e2f311b3efR340
* For patterns this check will fail (we would compare `*.foo.com` with `bar.foo.com`) ~~because `MatchPattern` is not available to content~~
* This needs to be fixed when we want to land two different messages that use the same `frequentVisits` trigger
  * - [x] create follow up bug -- https://bugzilla.mozilla.org/show_bug.cgi?id=1540182

### DIFF
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -337,6 +337,8 @@ this.ASRouterTargeting = {
       return false;
     } else if (!candidateMessageTrigger.params) {
       return true;
+    } else if (candidateMessageTrigger.patterns) {
+      return true;
     }
     return candidateMessageTrigger.params.includes(trigger.param);
   },
--- a/lib/ASRouterTriggerListeners.jsm
+++ b/lib/ASRouterTriggerListeners.jsm
@@ -82,26 +82,32 @@ this.ASRouterTriggerListeners = new Map([
     _visits: null,
 
     async init(triggerHandler, hosts = [], patterns) {
-      if (this._initialized) {
-        return;
-      }
-      this.onTabSwitch = this.onTabSwitch.bind(this);
+      if (!this._initialized) {
+        this.onTabSwitch = this.onTabSwitch.bind(this);
 
-      // Add listeners to all existing browser windows
-      for (let win of Services.wm.getEnumerator("navigator:browser")) {
-        if (isPrivateWindow(win)) {
-          continue;
+        // Add listeners to all existing browser windows
+        for (let win of Services.wm.getEnumerator("navigator:browser")) {
+          if (isPrivateWindow(win)) {
+            continue;
+          }
+          await checkStartupFinished(win);
+          win.addEventListener("TabSelect", this.onTabSwitch);
+          win.gBrowser.addTabsProgressListener(this);
         }
-        await checkStartupFinished(win);
-        win.addEventListener("TabSelect", this.onTabSwitch);
-        win.gBrowser.addTabsProgressListener(this);
-      }
 
-      this._initialized = true;
+        this._initialized = true;
+        this._visits = new Map();
+      }
       this._triggerHandler = triggerHandler;
-      this._visits = new Map();
       if (patterns) {
-        this._matchPatternSet = new MatchPatternSet(patterns, MATCH_PATTERN_OPTIONS);
+        if (this._matchPatternSet) {
+          this._matchPatternSet = new MatchPatternSet([
+            ...this._matchPatternSet.patterns,
+            ...patterns,
+          ]);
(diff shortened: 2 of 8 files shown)


lib/ASRouterTriggerListeners.jsm
```diff
@@ -65,27 +78,37 @@ this.ASRouterTriggerListeners = new Map([
...
+        this._initialized = true;
+        this._visits = new Map();
+      }
       this._triggerHandler = triggerHandler;
-      this._visits = new Map();
+      if (patterns) {
+        if (this._matchPatternSet) {
+          this._matchPatternSet = new MatchPatternSet([
+            ...this._matchPatternSet.patterns,
+            ...patterns,
```
k88hudson: It might be worth removing duplicates here (for example, if we have a bunch of messages that target the same site, we would add in multiple copies)

test/browser/browser_asrouter_cfr.js
```diff
@@ -372,3 +372,15 @@ add_task(async function test_onLocationChange_cb() {
 
   Assert.equal(count, 2, "We moved to a new document");
 });
+
+add_task(async function test_matchPattern() {
+  let count = 0;
+  const triggerHandler = () => ++count;
+  ASRouterTriggerListeners.get("frequentVisits").init(triggerHandler, [], ["*://*.example.com/"]);
```
k88hudson: I don't think you need to re-test the whole set of functionality of `PatternMatch` here, but I'd like to see a couple more types of patterns

lib/ASRouterTriggerListeners.jsm
```diff
@@ -40,12 +41,24 @@ function isPrivateWindow(win) {
  *
  * @returns {string} - the host that matched the whitelist
  */
-function checkHost(aLocationURI, hosts, aRequest) {
+function checkHost(aLocationURI, {hosts, matchPatternSet}, aRequest) {
```
k88hudson: This is kind of weirdly named now... It's more like `checkURLMatch`

lib/ASRouterTriggerListeners.jsm
```diff
@@ -40,12 +41,24 @@ function isPrivateWindow(win) {
...
   // Check current location against whitelisted hosts
   if (hosts.has(aLocationURI.host)) {
     return aLocationURI.host;
   }
 
+  if (matchPatternSet) {
+    const uri = Services.io.newURI(aLocationURI.spec);
```
k88hudson: I guess you can't use `aLocationURI` directly?

piatra: Hm actually I should be able, I see matches has 2 implementations and one of them takes `strings`.

lib/ASRouterTargeting.jsm
```diff
@@ -337,6 +337,8 @@ this.ASRouterTargeting = {
       return false;
     } else if (!candidateMessageTrigger.params) {
       return true;
+    } else if (candidateMessageTrigger.patterns) {
```
k88hudson: Shouldn't the logic here be

```js
else if (!candidateMessageTrigger.params && !candidateMessageTrigger.patterns) {
  return true;
}
```

piatra: This is abusing the fact that we have only 1 message with patterns. I needed a condition to exit early and with `true` because for the next check we would essentially have to duplicate the `MatchPatternSet` checks from `TriggerListener.jsm`
I mentioned in the first comment that I would rather fix this as a follow up. Or come up with a better idea that creating a new `MatchPatternSet`.

k88hudson: Got it, I think this might break pretty quickly if we start landing test messages / fallbacks; that said, happy if this just has a comment and a follow-up is filed with `P1`

piatra:
Here is the follow up bug https://bugzilla.mozilla.org/show_bug.cgi?id=1540182

piatra:
I've added [documentation](https://github.com/mozilla/activity-stream/blob/ff07488a0759bad3e4f733e92ea262223ecd354c/content-src/asrouter/schemas/message-format.md#trigger-params) in message-format.md and we could link to it from docs.

Regarding QA we currently don't have any patterns. The initial spec for Pinned Tabs included one but it was removed since we didn't support it at the time. I can't find it anymore.

PR #4870, opened 2019-03-28
(nobug) - Ensure fetch calls don't include credentials

piatra:
There were 2 occurrences in `PingCentre` and 1 in `PersonalityProvider` that were an actual issue because that code runs in the main process. The content process occurrence is a false positive (the request object includes `credentials: omit`) and the others are just built time usages.
I've looked into `no-restricted-syntax` but it seems to completely forbid usage of certain features regardless of how they are used/their options.


k88hudson:
Also this is really useful and might be interesting elsewhere in Mozilla Central, maybe let Standard8 know about it?

piatra:
I've removed the dependency on `requireindex` as well.

PR #4871, opened 2019-03-29
Bug 1539904 - Handle Delete and Archive from pocket link menus when h…

punamdahiya:
…ighlights not shown

### DIFF
--- a/lib/HighlightsFeed.jsm
+++ b/lib/HighlightsFeed.jsm
@@ -228,34 +228,6 @@ this.HighlightsFeed = class HighlightsFeed {
     });
   }
 
-  /**
-   * Deletes an item from a user's saved to Pocket feed and then refreshes highlights
-   * @param {int} itemID
-   *  The unique ID given by Pocket for that item; used to look the item up when deleting
-   */
-  async deleteFromPocket(itemID) {
-    try {
-      await NewTabUtils.activityStreamLinks.deletePocketEntry(itemID);
-      this.fetchHighlights({broadcast: true});
-    } catch (err) {
-      Cu.reportError(err);
-    }
-  }
-
-  /**
-   * Archives an item from a user's saved to Pocket feed and then refreshes highlights
-   * @param {int} itemID
-   *  The unique ID given by Pocket for that item; used to look the item up when archiving
-   */
-  async archiveFromPocket(itemID) {
-    try {
-      await NewTabUtils.activityStreamLinks.archivePocketEntry(itemID);
-      this.fetchHighlights({broadcast: true});
-    } catch (err) {
-      Cu.reportError(err);
-    }
-  }
-
   onAction(action) {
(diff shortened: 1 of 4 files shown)


punamdahiya:
@Mardak @ScottDowne @gvn whoever has bandwidth, please review. Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1539904#c0 and pocket new tab delete and archive link menus  https://github.com/mozilla/activity-stream/pull/4863 .  Thanks!

lib/HighlightsFeed.jsm
```diff
@@ -228,34 +228,6 @@ this.HighlightsFeed = class HighlightsFeed {
...
 
-  /**
-   * Deletes an item from a user's saved to Pocket feed and then refreshes highlights
-   * @param {int} itemID
-   *  The unique ID given by Pocket for that item; used to look the item up when deleting
-   */
-  async deleteFromPocket(itemID) {
-    try {
-      await NewTabUtils.activityStreamLinks.deletePocketEntry(itemID);
-      this.fetchHighlights({broadcast: true});
```
Mardak: The old code would wait for the entry to be deleted before refreshing. But perhaps that ordering doesn't matter right now?

lib/HighlightsFeed.jsm
```diff
@@ -280,10 +252,10 @@ this.HighlightsFeed = class HighlightsFeed {
         this.fetchHighlights({broadcast: true});
         break;
       case at.DELETE_FROM_POCKET:
-        this.deleteFromPocket(action.data.pocket_id);
+        this.fetchHighlights({broadcast: true});
```
ScottDowne: I think it's confusing why a `DELETE_FROM_POCKET` case fires off a `fetchHIghlights`

lib/HighlightsFeed.jsm
```diff
@@ -280,10 +252,10 @@ this.HighlightsFeed = class HighlightsFeed {
...
         break;
       case at.ARCHIVE_FROM_POCKET:
-        this.archiveFromPocket(action.data.pocket_id);
+        this.fetchHighlights({broadcast: true});
```
ScottDowne: I think it's confusing why a `ARCHIVE_FROM_POCKET` case fires off a `fetchHIghlights`

lib/HighlightsFeed.jsm
punamdahiya: This doesn't impact DS but in Activity stream both delete and archive from pocket actions should trigger refresh of highlight content so that we don't show deleted and archived pocket links in highlights.

ScottDowne: Ah yes, I see these lines inside the deleted highlights code.

One thought though is maybe a comment that you're just forcing a refresh after something has changed.

k88hudson:
Looks good, I'm going to put up a follow-up patch to handle the ordering better as @Mardak mentioned

PR #4877, opened 2019-04-01
Bug 1529983 - Add placeholder when there are no stories to show

punamdahiya:


punamdahiya:
@k88hudson @gvn @ScottDowne Please review. Thanks!

content-src/components/DiscoveryStreamComponents/DSEmptyState/DSEmptyState.jsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+export class DSEmptyState extends React.PureComponent {
+  render() {
+    return (
+      <div className="section-empty-state">
+        <p className="empty-state-message">
```
wolasi: This isn't a blocker, but as we begin to address accessibility, i think we should use semantic markup whenever possible. With that in mind I recommend we change this to…

```
<div className="section-empty-state">
  <h2>You're caught up!</h2>
  <p>Check back later for more stories.</p>
</div>
```

content-src/components/DiscoveryStreamComponents/DSEmptyState/DSEmptyState.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+export class DSEmptyState extends React.PureComponent {
+  render() {
+    return (
+      <div className="section-empty-state">
+        <div className="empty-state-message">
+          <h2>You are caught up!</h2>
+          <div>Check back later for more stories.</div>
```
wolasi: Let's use `<p>` tag here…
```
<p>Check back later for more stories.</p>
```

punamdahiya: > Let's use `<p>` tag here…

sure, is paragraph tag also accessibility friendly?

wolasi: I believe so

punamdahiya: done

content-src/components/DiscoveryStreamComponents/DSEmptyState/_DSEmptyState.scss
```diff
@@ -0,0 +1,26 @@
+.section-empty-state {
+  border: $border-secondary;
+  border-radius: 4px;
+  display: flex;
+  height: 96px;
```
k88hudson: Why do you need the fixed height here? Is that to line it up with other sections/elements? 

If so I would try to use a variable (like is that `$grid-unit` maybe?) or at least a comment

punamdahiya: yeah specs call for setting our empty state section height to 96px, will change to use $grid-unit

https://www.figma.com/file/0xZkmT86rSXbGRB4eIvrn0ts/New-Tab-Spec-for-Fx-v68?node-id=91%3A12

content-src/components/DiscoveryStreamComponents/CardGrid/CardGrid.jsx
```diff
@@ -12,23 +13,40 @@ export class CardGrid extends React.PureComponent {
...
-        url={rec.url}
-        id={rec.id}
-        type={this.props.type}
-        context={rec.context}
-        dispatch={this.props.dispatch}
-        source={rec.domain}
-        pocket_id={rec.pocket_id}
-        bookmarkGuid={rec.bookmarkGuid} />
-    ));
+    if (!data.recommendations.length) {
```
k88hudson: I think rendering the `ds-header` in both the empty and non-empty case is probably not the best, maybe you could do something like this?
```js
render() {
    const {data} = this.props;

    // Handle a render before feed has been fetched by displaying nothing
    if (!data) {
      return null;
    }

    // Handle the case where a user has dismissed all recommendations
    const isEmpty = data.recommendations.length === 0;

    return (<div>
      <div className="ds-header">{this.props.title}</div>
      {isEmpty ?
        <div className="ds-card-grid empty"><DSEmptyState /></div> :
        this.renderCards()
      }
    </div>);
  }
}
```

content-src/components/DiscoveryStreamComponents/DSEmptyState/DSEmptyState.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+export class DSEmptyState extends React.PureComponent {
+  render() {
+    return (
+      <div className="section-empty-state">
+        <div className="empty-state-message">
+          <h2>You are caught up!</h2>
+          <p>Check back later for more stories.</p>
```
Mardak: Are we intentionally not exposing these strings to localization?

punamdahiya: Yes, for 68 keeping the strings to en locale only.  I believe 69+ we do need to localize.

punamdahiya:
@k88hudson Updated PR with review feedback, thanks!

PR #4879, opened 2019-04-02
Bug 1540765 - PersistentCache should do JSON parsing off of the main thread

rlr:
…

### DIFF
--- a/lib/PersistentCache.jsm
+++ b/lib/PersistentCache.jsm
@@ -6,7 +6,7 @@
 const {XPCOMUtils} = ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
 
 ChromeUtils.defineModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm");
-XPCOMUtils.defineLazyGetter(this, "gTextDecoder", () => new TextDecoder());
+XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
 
 /**
  * A file (disk) based persistent cache of a JSON serializable object.
@@ -59,9 +59,8 @@ this.PersistentCache = class PersistentCache {
         const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, this._filename);
         const fileExists = await OS.File.exists(filepath);
         if (fileExists) {
-          const binaryData = await OS.File.read(filepath);
-          const json = gTextDecoder.decode(binaryData);
-          data = JSON.parse(json);
+          const file = await fetch(`file://${filepath}`);
+          data = await file.json();
         }
       } catch (error) {
         Cu.reportError(`Failed to load ${this._filename}: ${error.message}`);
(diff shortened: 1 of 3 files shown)


lib/PersistentCache.jsm
```diff
@@ -59,9 +59,8 @@ this.PersistentCache = class PersistentCache {
         const filepath = OS.Path.join(OS.Constants.Path.localProfileDir, this._filename);
         const fileExists = await OS.File.exists(filepath);
```
mikeconley: Out of curiosity, is it still necessary to do the stat here? Can we do the fetch, and handle the failure status in the event that the file does not exist?

rlr: good point, will fix 👍

lib/PersistentCache.jsm
```diff
@@ -59,9 +59,8 @@ this.PersistentCache = class PersistentCache {
...
         if (fileExists) {
-          const binaryData = await OS.File.read(filepath);
-          const json = gTextDecoder.decode(binaryData);
-          data = JSON.parse(json);
+          const file = await fetch(`file://${filepath}`);
+          data = await file.json();
```
piatra: I think these two calls can be joined together
```javascript
await (await fetch(...)).json();
```

test/unit/lib/DiscoveryStreamFeed.test.js
```diff
@@ -385,7 +385,7 @@ describe("DiscoveryStreamFeed", () => {
 
       await feed.loadSpocs(feed.store.dispatch);
 
-      assert.notCalled(global.fetch);
+      assert.calledOnce(global.fetch); // Persistent cache attempts a fetch
```
piatra: This is kinda weird and might make test debugging confusing since the test description mentions `should not fetch`.
Can you instead stub out PersistentCache so that call doesn’t happen? It should prevent having to make all the other changes as well.

test/unit/lib/PersistentCache.test.js
```diff
@@ -36,31 +33,23 @@ describe("PersistentCache", () => {
...
-    it("doesnt try to read the file if it was already loaded", async () => {
-      fakeOS.File.read = sinon.spy();
+    it("doesnt try to fetch the file if it was already loaded", async () => {
       await cache._load();
-      fakeOS.File.read.resetHistory();
+      fakeFetch.resetHistory();
       await cache.get("foo");
-      assert.notCalled(fakeOS.File.read);
-    });
-    it("should catch and report errors", async () => {
```
piatra: I think it would be useful to keep the `Cu.reportError` Call if fetch or json parsing fails.

test/unit/lib/DiscoveryStreamFeed.test.js
rlr: yeah, good call

test/unit/lib/PersistentCache.test.js
rlr: the problem was `fetch()` throws if the cache file doesn't exist yet. So, it would be spammy to report that an an error since it is expected on a cold cache. I'll break up the `fetch` from the `json` and only report json parsing errors.

PR #4884, opened 2019-04-05
Bug 1540939 - Remove SnippetsFeed and related components

piatra:
I also turned back on `test/unit/lib/ActivityStream.test.js` which was commented out.
There's an m-c patch that also needs to land in order to remove `browser.aboutHomeSnippets.updateUrl`
 - [x] todo: m-c patch https://phabricator.services.mozilla.com/D27252


k88hudson:
This is mostly good, a couple of things:
* `browser.newtabpage.activity-stream.feeds.snippets` is the preference we depend on as a user preference so it needs to be initialized somewhere (it could be renamed/migrated, but it's probably fine to keep it as is)
* you can remove the `disableSnippets` pref definition since it is no long being used or necessary

Let's also make sure QA is checking for regressions after this lands (with user preferences for example)

piatra:
I've kept `feeds.snippets` pref and removed `disableSnippets`.
I [added QA steps](https://bugzilla.mozilla.org/show_bug.cgi?id=1540939#c2) in the bug and will request QA once this is in Nightly.

PR #4885, opened 2019-04-05
Bug 1531933 – Use proper image sizes

gvn:
Testing: 

- Try some layout variants (recommend: `66-beta-2-complex`, `basic`, `dev-test-all`).
- Ensure no images break (caveat: since there's a CDN and a image processing service middle tier you may see something slow to load if you're the first to hit it...)


gvn:
Going to do a little refactoring...

gvn:
Added lazy loading & tightened up the measurement implementation. 🕺

ScottDowne:
That went together fairly fast!

First thing I noticed (unless I missed it) you probably need something similar to this:

```
  componentWillUnmount() {
    if (this._handleIntersect && this.impressionObserver) {
      this.impressionObserver.unobserve(this.refs.impression);
    }
    if (this._onVisibilityChange) {
      this.props.document.removeEventListener(VISIBILITY_CHANGE_EVENT, this._onVisibilityChange);
    }
  }
```

content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
```diff
@@ -0,0 +1,95 @@
...
+    constructedURL = `https://pocket-image-cache.com/${width}x0/filters:format(jpeg):quality(60):no_upscale():strip_exif()/${(constructedURL)}`;
+
+    // Use Mozilla CDN:
+    return `https://img-getpocket.cdn.mozilla.net/direct?url=${encodeURIComponent(constructedURL)}`;
+  }
+
+  componentDidMount() {
+    let options = {
+      root: document.querySelector(`document`),
+      threshold: 1,
```
ScottDowne: I think you want this to be 0.

what this does is saying how much of the element we're observing needs to be displayed before it is considered displayed. A 1 in this value is 100%, so if I scroll down and 1% of an image is still not fully scrolled, I see mostly a blank image, at 0, it waits until even a tiny bit of the image is in view before it's displayed.

ScottDowne:
If I open a new tab using the `+` button, it doesn't actually render images. I think this has to do with either prerendered images, or `document.visibilityState === visible` or a combination of both?

I know when me and Nan were working on the Impressions stuff, we had to use a bit of care with the visibility stuff.

Locally I removed the `document.visibilityState === visible` check, and the `+` is now working as expected.

I wonder if we should sanity check this with Nick Chapman?

ScottDowne:
Finally, I think you need to go through the array in `element` in `onSeen` because sometimes the first item is an unseen intersection, and the last one is the seen one? Not exactly sure how that works, but an example can be seen here: https://github.com/mozilla/activity-stream/blob/master/content-src/components/DiscoveryStreamImpressionStats/ImpressionStats.jsx#L138

Notice the `entries.some`. I *think* it has to do if you scroll or page load, it doesn't fire the events off async as the browser updates, but does it on it's own clock, so if two events come through in that time, it sends back both entries, and you need to check for any. But not sure about that... From my testing, I am getting back cases where I have an array of two items, the first `element[0].isIntersecting` is false and the second `element[1].isIntersecting` is true.

content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
gvn: I dropped to 0.5, but we can adjust if that still feels like too much.

gvn:
Looks like I'll need to add some test coverage for `DSImage.jsx` now that our requirements are higher.

gvn:
> If I open a new tab using the + button, it doesn't actually render images. I think this has to do with either prerendered images, or document.visibilityState === visible or a combination of both?

Taking out that check fixed opening new new tabs. 👍

content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
ScottDowne: I was initially thinking we do this asap. With 0.5 we just need to be aware that if someone scrolls down and sees just under half an image, it appears blank still. Might be more noticeable in cases where the screen height is low enough and the resolution is low enough that the page loads with < than 50% of an image visible.

What thoughts do you have on the pros and cons between 0.5 and 0?

gvn: I just tried 0 and yeah, it's fine so I'll just change it! :D

content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
```diff
@@ -0,0 +1,107 @@
...
+    this.onOptimizedImageError = this.onOptimizedImageError.bind(this);
+
+    this.state = {
+      isSeen: false,
+      optimizedImageFailed: false,
+    };
+  }
+
+  onSeen(element) {
+    if (this.state && element[0].isIntersecting) {
```
ScottDowne: Yeah still seeing issues with the [0] This event can fire multiple intersections, and the first one might not be the most up to date.

I see about a 50/50 chance an image shows as is.

I fixed this by trying this:

```
  onSeen(entries) {
    if (this.state) {
      if (entries.some(entry => entry.isIntersecting)) {
        if (this.props.optimize) {
          this.setState({
            containerWidth: ReactDOM.findDOMNode(this).clientWidth,
          });
        }

        this.setState({
          isSeen: true,
        });

        // Stop observing since element has been seen
        this.observer.unobserve(ReactDOM.findDOMNode(this));
      }
    }
  }
```

It kinda mentions this here but tbh they don't do enough to stress this: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Intersection_change_callbacks

content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
```diff
@@ -0,0 +1,107 @@
...
+  reformatImageURL(url, width) {
+    // Change the image URL to request a size tailored for the parent container width
+    // Also: force JPEG, quality 60, no upscaling, no EXIF data
+    // Uses Thumbor: https://thumbor.readthedocs.io/en/latest/usage.html
+    return `https://img-getpocket.cdn.mozilla.net/${width}x0/filters:format(jpeg):quality(60):no_upscale():strip_exif()/${encodeURIComponent(url)}`;
+  }
+
+  componentDidMount() {
+    let options = {
+      root: document.querySelector(`document`),
```
ScottDowne: This might actually be the default used, I wonder if we remove this, if things still work.

content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
```diff
@@ -0,0 +1,107 @@
...
+      threshold: 0, // Load as soon as the first pixel crosses into view
```
ScottDowne: Heh, I think 0 is also the default here too, so we can remove this one. I'm sure about this one, not sure about root.

PR #4886, opened 2019-04-08
Bug 1533601 - About preferences settings for discovery stream

punamdahiya:


punamdahiya:
@Mardak PR is close but not ready yet as its failing mochitest, do you know why? Thanks

lib/AboutPreferences.jsm
```diff
@@ -269,77 +267,9 @@ this.AboutPreferences = class AboutPreferences {
...
-      // Add the Learn more link in the description
-      const topstoriesSection = prefStructure.find(s => s.id === "topstories");
-      const learnMoreURL = topstoriesSection && topstoriesSection.learnMore.link.href;
-      const link = createAppend("label", descriptionHbox);
-      link.classList.add("learn-sponsored");
-      link.classList.add("text-link");
-      link.setAttribute("href", learnMoreURL);
-      link.textContent = formatString("prefs_topstories_sponsored_learn_more");
-
-      if (discoveryStreamConfig.show_spocs && sponsoredStoriesCheckbox) {
```
punamdahiya: DiscoveryStreamFeed shows and hide spocs based off pref showSponsored.
https://searchfox.org/mozilla-central/source/browser/components/newtab/lib/DiscoveryStreamFeed.jsm#74

Do we need to explicitly check in about:preferences going forward for discoveryStreamConfig.show_spocs if discovery stream is enabled or handling via showSponsored pref suffice?

lib/AboutPreferences.jsm
```diff
@@ -269,77 +267,9 @@ this.AboutPreferences = class AboutPreferences {
...
-        }
       });
     });
 
-    if (discoveryStreamConfig.enabled) {
-      // If Discovery Stream is enabled hide Home Content options
-      contentsGroup.style.visibility = "collapse";
-
-      const discoveryGroup = homeGroup.insertAdjacentElement("afterend", homeGroup.cloneNode());
-      discoveryGroup.id = "discoveryContentsGroup";
```
Mardak: I believe this id becomes unused, so it can be removed from the styles above.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -152,49 +194,21 @@ export class _DiscoveryStreamBase extends React.PureComponent {
...
-            </div>
-          </div>
-        ))}
-        {this.renderStyles(styles)}
-      </div>
-    );
-  }
-}
+  return (
+  <div {...props} className={`ds-column ds-column-${props.row.width}`}>
```
Mardak: What prop are we trying to get from here? I believe we want to avoid spread on component properties.

punamdahiya:
> The approach in both `DiscoveryStreamBase` and `AboutPreferences` don't seem quite right. Basically both places seem to be doing special checks to change behavior, e.g., conditionally render TopSites or conditionally hide topstories rows dropdown. This can lead to odd behaviors such as reserving space for the dropdown when the text could have fit on one line:

@wolasi As per specs, we are wrapping text for top stories preference. https://www.figma.com/file/0xZkmT86rSXbGRB4eIvrn0ts/New-Tab-Spec-for-Fx-v68?node-id=69%3A0

Is it ok if the text doesn't wrap and takes available space and shows in one line. 

Another question, should it work if we show How it works link besides 'Recommended by Pocket' heading instead of moving it in top stories pref description for discovery stream

punamdahiya:
discovery-stream-preferences

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
punamdahiya: +1,  its redundant

wolasi:
Yes it's fine if the text doesn't wrap as shown in the spec

content-src/lib/selectLayoutRender.js
```diff
@@ -38,6 +38,34 @@ export const selectLayoutRender = (state, rickRollCache) => {
...
+
+    if (!prefs["feeds.section.topstories"]) {
+      filterArray.push(...DS_COMPONENTS);
+    }
+
+    for (const [rowIndex, row] of Object.entries(layout)) {
+      for (const [index, component] of Object.entries(row.components)) {
+        if (filterArray.includes(component.type)) {
+          if (row.components.length === 1) {
+            layout.splice(rowIndex, 1);
```
Mardak: I used a double for loop for #4890 because I wanted to return early. Might be cleaner/safer to use array methods, e.g., reduce/filter instead of modifying the array that you're iterating over as we need to visit every item anyway.

Mardak: E.g.,
```js
return layout.map(row => ({
    ...row,
      components: row.components.filter(component => !filterArray.includes(component.type))
  })).filter(row => row.components.length).map(row => ({
    ...row,
```
And no need to have a separate `filterComponentsByPref` inline function as it only gets called once.

Mardak: Actually, looking at this again, the existing `layout.map` looks pretty much like what we want to do anyway. So perhaps more like

```js
return layout.map(row => ({
  ...row,
  // Loops through desired components and adds a .data property
  components: row.components.filter(c => !filterArray.includes(c.type)).map(component => …),
})).filter(row => row.components.length);

So just before adding .data property, first filter out the components. Then if there's no more components in a row, filter out empty rows.

File: lib/AboutPreferences.jsm [hunk 5]
```diff
@@ -104,9 +94,32 @@ this.AboutPreferences = class AboutPreferences {
     }
   }
 
+  handleDiscoverySettings(sections) {
+    // Deep copy object to not modify original Sections state in store
+    let sectionsCopy = JSON.parse(JSON.stringify(sections));
+    sectionsCopy.forEach(obj => {
+      if (obj.id === "highlights") {
+        obj.shouldHidePref = true;
```
Mardak: Ah neat. This cleaned up nicely.

content-src/lib/selectLayoutRender.js
```diff
@@ -38,6 +38,34 @@ export const selectLayoutRender = (state, rickRollCache) => {
...
+    const DS_COMPONENTS = ["Message", "SectionTitle", "Navigation",
+      "CardGrid", "Hero", "HorizontalRule", "List"];
+
+    const filterArray = [];
+
+    if (!prefs["feeds.topsites"]) {
+      filterArray.push("TopSites");
+    }
+
+    if (!prefs["feeds.section.topstories"]) {
```
Mardak: Mochitest is probably failing because this pref is false causing nothing to be rendered.

test/unit/content-src/components/DiscoveryStreamComponents/DiscoveryStreamBase.test.jsx
```diff
@@ -9,8 +9,9 @@ describe("<DiscoveryStreamBase>", () => {
       feeds: {loaded: true},
       layout: [],
     };
+    const prefsProps = {values: {}};
 
-    const wrapper = shallow(<DiscoveryStreamBase DiscoveryStream={discoveryStreamProps} />);
+    const wrapper = shallow(<DiscoveryStreamBase DiscoveryStream={discoveryStreamProps} Prefs={prefsProps} />);
```
Mardak: We can delete this file instead of fixing it up. There's a more comprehensive test at `test/unit/content-src/components/DiscoveryStreamBase.test.jsx` (not under `DiscoveryStreamComponents`). I've pushed a separate PR #4892 to clean things up and if you rebase on it when it's on master, your failing test should pass.

content-src/lib/selectLayoutRender.js
Mardak: This should also make coverage happy as there aren't separate conditional branches to test.

lib/AboutPreferences.jsm
Mardak: Just to document, the nested sponsored prefs only show up when spocs are enabled:
https://github.com/mozilla/activity-stream/blob/8b20ec0902e5109303c2fe30a32fda7cf33d74e8/lib/SectionsManager.jsm#L24-L28

punamdahiya:
> Overall approach seems fine. Just need to get the automation passing (mochitest and coverage -- #4892 should fix `this.props.Prefs is undefined` failure)

Thanks @Mardak Updated PR with feedback, waiting for tests to pass before merge.

Mardak:
The mochitest is failing now because topstories is trying to make network requests. We can probably just enable topstories but make it do nothing for all tests via browser.ini:
https://github.com/mozilla/activity-stream/blob/7b5039dd0005779c65115ea6a443cf97eb6d755c/test/browser/browser.ini#L6-L8
```ini
  browser.newtabpage.activity-stream.feeds.section.topstories=true
  browser.newtabpage.activity-stream.feeds.section.topstories.options={}
```

lib/AboutPreferences.jsm
```diff
@@ -269,77 +278,9 @@ this.AboutPreferences = class AboutPreferences {
...
-        // If there is no element to reuse create one
-        const discoveryDetails = createAppend("vbox", discoveryGroup);
-        const subcheck = createAppend("checkbox", discoveryDetails);
-        subcheck.setAttribute("label", formatString("prefs_topstories_options_sponsored_label"));
-        linkPref(subcheck, PREF_SHOW_SPONSORED, "bool");
-      }
-
-      const contentDiscoveryButton = document.createElementNS(HTML_NS, "button");
-      contentDiscoveryButton.classList.add("contentDiscoveryButton");
-      contentDiscoveryButton.textContent = formatString("prefs_content_discovery_button");
```
Mardak: Oops we forgot to remove this string:
https://github.com/mozilla/activity-stream/blob/da0171e0ed567c660f624a09ceb4ba6dfc71ccc6/locales-src/en-US/strings.properties#L97

Mardak: Oh and `formatString("prefs_content_discovery_header");` too ?

punamdahiya: ah, good catch

PR #4890, opened 2019-04-08
Bug 1535773 - Allow collapsing of top sites and the rest of Discovery Stream content

Mardak:
r? @punamdahiya This is on top of #4892 and should probably land after your #4886 to actually get top sites or pocket section to hide. But if it does land ahead of the preferences PR, it will still mostly work.

`basic` with collapsed top sites and showing menu for pocket
Screen Shot 2019-04-09 at 10 10 11 PM

`66-release-2-above-2-below-1` with top sites menu open
Screen Shot 2019-04-09 at 10 14 48 PM

---

Original:

@punamdahiya @k88hudson Looks like this approach of extracting the first `TopSites` and first `Message` works for various layout json. Still need to fix up styling and the section menu options.

66-release-2-above-2-below-1 (individual topsite row followed by individual message row)
66-release-2-above-2-below-1

dev-test-all (custom Message title, many "row"s with components scattered)
dev-test-all

basic (single 12-width "row" with many components)
basic

66-beta-2-complex (individual topsite row followed by individual message row)
66-beta-2-complex

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -223,6 +223,7 @@ export class _DiscoveryStreamBase extends React.PureComponent {
           dispatch={this.props.dispatch}
           icon={topStories.icon}
           id={topStories.id}
+          isFixed={true}
           learnMore={{
             link: {
               href: message.header.link_url,
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -218,6 +218,7 @@ export class _DiscoveryStreamBase extends React.PureComponent {
           components: [topSites],
         }])}
         {layoutRender.length > 0 && <CollapsibleSection
+          className="ds-layout"
           collapsed={topStories.pref.collapsed}
           dispatch={this.props.dispatch}
           icon={topStories.icon}
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -1,4 +1,5 @@
 import {CardGrid} from "content-src/components/DiscoveryStreamComponents/CardGrid/CardGrid";
+import {CollapsibleSection} from "content-src/components/CollapsibleSection/CollapsibleSection";
 import {connect} from "react-redux";
 import {DSMessage} from "content-src/components/DiscoveryStreamComponents/DSMessage/DSMessage";
 import {Hero} from "content-src/components/DiscoveryStreamComponents/Hero/Hero";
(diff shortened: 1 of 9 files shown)


content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -168,13 +169,63 @@ export class _DiscoveryStreamBase extends React.PureComponent {
...
+            return component;
+          }
+        }
+      }
+      return null;
+    };
+
+    // Extract TopSites to render before the rest and Message to use for header
+    const topsites = extractComponent("TopSites");
+    const message = extractComponent("Message");
```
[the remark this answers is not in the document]
ScottDowne: Eh, probably not worth it. We're likely to find the component we're looking for fairly quickly and ca return early from the function anyway.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -168,13 +169,63 @@ export class _DiscoveryStreamBase extends React.PureComponent {
...
+    // Render a DS-style TopSites then the rest in a section that reuses the
+    // "topstories" Section state
+    const topStories = this.props.Sections.find(s => s.id === "topstories");
+    return (
+      <React.Fragment>
+        {topsites && this.renderLayout([{
+          width: 12,
+          components: [topsites],
+        }])}
+        <CollapsibleSection
```
ScottDowne: I think Tim mentioned this feature being something we could turn on and off via a pref, not sure if we still want to go that route though.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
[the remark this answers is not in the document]
punamdahiya: @mardak with this change  it's still possible to  show topsites under recommended by pocket section, is that correct?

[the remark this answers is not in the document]
punamdahiya: Trying to figure from about:preferences if a user hides topsites, should it hide only the first topsites component or everywhere on new tab. I believe its later.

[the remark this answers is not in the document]
Mardak: It seems unlikely that a layout would contain multiple TopSites, but the current behavior with dev-test-all does move the first TopSites and leaves the others untouched. So in the common case of a single TopSites component, it actually **wouldn't** be possible to show TopSites under a Recommended by Pocket section.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
Mardak: Tim suggests that it'll be good to have something as part of `.config` pref to turn off if we need to. Probably a good idea as without `collapsible: true` some layouts wouldn't be possible.

Mardak:
Should be ready to review now rebased on top of #4892 and edited original comment with latest screenshots.

Mardak:
I split out the styling to bug https://bugzilla.mozilla.org/show_bug.cgi?id=1543442 so this PR just reuses the existing styling.

content-src/components/SectionMenu/SectionMenu.jsx
```diff
@@ -12,6 +12,10 @@ export class _SectionMenu extends React.PureComponent {
     const {props} = this;
 
     const propOptions = props.isWebExtension ? [...WEBEXT_SECTION_MENU_OPTIONS] : [...DEFAULT_SECTION_MENU_OPTIONS];
```
punamdahiya: For discovery stream we should hide 'Remove Section' link menu option from ...DEFAULT_SECTION_MENU_OPTIONS

Mardak: Confirmed that we want to keep all the menu options including remove for 68.

Mardak:
Rebased on latest master with #4886 fixing simple conflicts to DSBase.jsx. This gets remove option working correctly with the restored about:preferences functionality.

lib/ActivityStream.jsm
```diff
@@ -236,6 +236,7 @@ const PREFS_CONFIG = new Map([
       const isEnabled = IS_NIGHTLY_OR_UNBRANDED_BUILD && locales && locales.includes(locale);
       return JSON.stringify({
         api_key_pref: "extensions.pocket.oAuthConsumerKey",
+        collapsible: true,
```
punamdahiya: Making collapsible false will still keep TopSites collapsible and will only remove section menu options and collapsible arrow from discovery stream top stories

Mardak: Yup. And probably same thing for highlights when that gets restored for parity.

PR #4891, opened 2019-04-09
Bug 1538804 - Update targeting for CFR Pinned tabs to include more locales and TLDs

piatra:


piatra:
TC failure due to https://bugzilla.mozilla.org/show_bug.cgi?id=1543007
2nd run is green.

PR #4895, opened 2019-04-10
Use Photon edit icon

nt1m:


nt1m:
It's from https://design.firefox.com/icons/viewer/#edit

Mardak:
@aaronrbenson I guess we added edit before the photon icon was finalized?

Before:
Screen Shot 2019-04-10 at 10 09 40 PM

After:
Screen Shot 2019-04-10 at 10 08 02 PM

nt1m:
For more context, shorlander created the icon here: https://github.com/FirefoxUX/photon-icons/pull/48

Mardak:
Oh. Hrmm.. Looks like there's a customize icon that is closer to the existing "edit":

https://design.firefox.com/icons/viewer/#custom

aaronrbenson:
@Mardak the icon in your "After" shot is correct. I think you're right, we didn't have that icon available to us at the time. The brush is used for Customize Mode.

PR #4897, opened 2019-04-11
Message schema for CFR FxA Message

piatra:


piatra:
- [ ] This needs to be updated with a `Close` button that dismisses the message.

PR #4898, opened 2019-04-11
(nobug) - Remove unused strings and exclude them from export

Mardak:
r? @piatra or @k88hudson We forgot to remove a couple of strings in https://github.com/mozilla/activity-stream/pull/4886#discussion_r274426369 Then I realized the export still contained strings for other locales that we no longer used, so fixed that too. A second `Object.assign({}, ...spread)` to combine arbitrary number of objects! 😉

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -40,7 +40,11 @@ function getStrings(locale, allStrings) {
   const localeFallbacks = [DEFAULT_LOCALE, ...similarLocales, locale];
 
   // Get strings from each locale replacing with those from more desired ones
-  return Object.assign({}, ...localeFallbacks.map(l => allStrings[l]));
+  const desired = Object.assign({}, ...localeFallbacks.map(l => allStrings[l]));
+
+  // Only include strings that are currently used (defined by default locale)
+  return Object.assign({}, ...Object.keys(allStrings[DEFAULT_LOCALE]).map(
+    key => ({[key]: desired[key]})));
 }
 
 /**
--- a/locales-src/en-US/strings.properties
+++ b/locales-src/en-US/strings.properties
@@ -92,9 +92,7 @@ section_disclaimer_topstories_buttontext=Okay, got it
 prefs_home_header=Firefox Home Content
 prefs_home_description=Choose what content you want on your Firefox Home screen.
 
-prefs_content_discovery_header=Firefox Home
 prefs_content_discovery_description=Content Discovery in Firefox Home allows you to discover high-quality, relevant articles from across the web.
-prefs_content_discovery_button=Turn Off Content Discovery
 
 # LOCALIZATION NOTE (prefs_section_rows_option): This is a semi-colon list of
 # plural forms used in a drop down of multiple row options (1 row, 2 rows).


bin/render-activity-stream-html.js
```diff
@@ -40,7 +40,11 @@ function getStrings(locale, allStrings) {
   const localeFallbacks = [DEFAULT_LOCALE, ...similarLocales, locale];
 
   // Get strings from each locale replacing with those from more desired ones
-  return Object.assign({}, ...localeFallbacks.map(l => allStrings[l]));
+  const desired = Object.assign({}, ...localeFallbacks.map(l => allStrings[l]));
+
+  // Only include strings that are currently used (defined by default locale)
```
piatra: Could we throw an error here if the number of strings is different? Catch unused strings?

piatra: I guess not because of fallbacks.

Mardak: Probably not super useful anyway because the localized strings we import is a combination of mozilla-central/beta/release, so naturally there will be unused-on-nightly strings. Although I was surprised to see a few locales with many removed as part of a test export, e.g., ga-IE:
```diff
diff --git a/browser/components/newtab/prerendered/locales/ga-IE/activity-stream-strings.js b/browser/components/newtab/prerendered/locales/ga-IE/activity-stream-strings.js
--- a/browser/components/newtab/prerendered/locales/ga-IE/activity-stream-strings.js
+++ b/browser/components/newtab/prerendered/locales/ga-IE/activity-stream-strings.js
@@ -35,19 +35,17 @@ window.gActivityStreamStrings = {
   "search_button": "Cuardach",
   "search_header": "Cuardach {search_engine_name}",
   "search_web_placeholder": "Cuardaigh an Gréasán",
   "section_disclaimer_topstories": "The most interesting stories on the web, selected based on what you read. From Pocket, now part of Mozilla.",
   "section_disclaimer_topstories_linktext": "Learn how it works.",
   "section_disclaimer_topstories_buttontext": "Okay, got it",
   "prefs_home_header": "Firefox Home Content",
   "prefs_home_description": "Choose what content you want on your Firefox Home screen.",
-  "prefs_content_discovery_header": "Firefox Home",
   "prefs_content_discovery_description": "Content Discovery in Firefox Home allows you to discover high-quality, relevant articles from across the web.",
-  "prefs_content_discovery_button": "Turn Off Content Discovery",
   "prefs_section_rows_option": "{num} row;{num} rows",
   "prefs_search_header": "Web Search",
   "prefs_topsites_description": "The sites you visit most",
   "prefs_topstories_description2": "Great content from around the web, personalized for you",
   "prefs_topstories_options_sponsored_label": "Sponsored Stories",
   "prefs_topstories_sponsored_learn_more": "Learn more",
   "prefs_highlights_description": "A selection of sites that you’ve saved or visited",
   "prefs_highlights_options_visited_label": "Visited Pages",
@@ -105,54 +103,10 @@ window.gActivityStreamStrings = {
   "firstrun_form_sub_header": "to continue to Firefox Sync",
   "firstrun_email_input_placeholder": "Email",
   "firstrun_invalid_input": "Valid email required",
   "firstrun_extra_legal_links": "By proceeding, you agree to the {terms} and {privacy}.",
   "firstrun_terms_of_service": "Terms of Service",
   "firstrun_privacy_notice": "Privacy Notice",
   "firstrun_continue_to_login": "Continue",
   "firstrun_skip_login": "Skip this step",
-  "context_menu_title": "Open menu",
-  "default_label_loading": "Á Lódáil…",
-  "header_stories": "Barrscéalta",
-  "header_visit_again": "Cuairt Arís",
-  "header_bookmarks": "Leabharmharcanna Le Déanaí",
-  "header_bookmarks_placeholder": "Níl aon leabharmharcanna agat.",
-  "header_stories_from": "ó",
-  "type_label_synced": "Sioncronaithe ó ghléas eile",
-  "type_label_open": "Oscailte",
-  "type_label_topic": "Ábhar",
-  "menu_action_copy_address": "Cóipeáil an Seoladh",
-  "menu_action_email_link": "Seol an Nasc trí Ríomhphost…",
-  "search_for_something_with": "Déan cuardach ar {search_term} le:",
-  "search_settings": "Socruithe Cuardaigh",
-  "welcome_title": "Fáilte go dtí cluaisín nua",
-  "welcome_body": "Úsáidfidh Firefox an spás seo chun na leabharmharcanna, ailt, físeáin, agus leathanaigh is tábhachtaí a thaispeáint duit, ionas go mbeidh tú in ann filleadh orthu gan stró.",
-  "welcome_label": "Buaicphointí á lorg",
-  "time_label_less_than_minute": "< 1 n",
-  "time_label_minute": "{number}n",
-  "time_label_hour": "{number}u",
-  "time_label_day": "{number}l",
-  "settings_pane_header": "Sainroghanna do Chluaisín Nua",
-  "settings_pane_body": "Roghnaigh na rudaí a fheicfidh tú nuair a osclóidh tú cluaisín nua.",
-  "settings_pane_search_header": "Cuardach",
-  "settings_pane_search_body": "Cuardaigh an Gréasán go díreach ón gcluaisín nua.",
-  "settings_pane_topsites_body": "Na suímh Ghréasáin a dtugann tú cuairt orthu is minice.",
-  "settings_pane_topsites_options_showmore": "Taispeáin dhá shraith",
-  "settings_pane_bookmarks_header": "Leabharmharcanna Le Déanaí",
-  "settings_pane_bookmarks_body": "Do chuid leabharmharcanna nua in áit amháin.",
-  "settings_pane_visit_again_header": "Cuairt Arís",
-  "settings_pane_visit_again_body": "Taispeánann Firefox nascanna ó do stair bhrabhsála a mbeadh suim agat filleadh orthu amach anseo.",
-  "settings_pane_pocketstories_header": "Barrscéalta",
-  "settings_pane_pocketstories_body": "Le Pocket, ball de theaghlach Mozilla, beidh tú ábalta teacht ar ábhar den chéad scoth go héasca.",
-  "settings_pane_done_button": "Déanta",
-  "edit_topsites_button_label": "Saincheap na Barrshuímh",
-  "edit_topsites_showmore_button": "Taispeáin níos mó",
-  "edit_topsites_showless_button": "Taispeáin níos lú",
-  "edit_topsites_done_button": "Déanta",
-  "edit_topsites_pin_button": "Greamaigh an suíomh seo",
-  "edit_topsites_unpin_button": "Díghreamaigh an suíomh seo",
-  "edit_topsites_dismiss_button": "Ruaig an suíomh seo",
-  "edit_topsites_add_button": "Cuir leis",
-  "pocket_feedback_header": "Ábhar den chéad scoth ón Ghréasán, le níos mó ná 25 milliún duine i mbun coimeádaíochta.",
-  "pocket_feedback_body": "Le Pocket, ball de theaghlach Mozilla, beidh tú ábalta teacht ar ábhar den chéad scoth go héasca.",
-  "pocket_send_feedback": "Tabhair Aiseolas Dúinn"
+  "context_menu_title": "Open menu"
 };
```
(Localized old/unused strings are automatically removed when the localizer updates the file, so in this case, it hasn't been touched for a long time.)

bin/render-activity-stream-html.js
```diff
@@ -40,7 +40,11 @@ function getStrings(locale, allStrings) {
   const localeFallbacks = [DEFAULT_LOCALE, ...similarLocales, locale];
 
   // Get strings from each locale replacing with those from more desired ones
-  return Object.assign({}, ...localeFallbacks.map(l => allStrings[l]));
+  const desired = Object.assign({}, ...localeFallbacks.map(l => allStrings[l]));
```
rlr: Is `Object.assign` needed? Won’t it do the same thing as `{...localeFallbacks.map()}`?

Mardak: This is using the spread syntax for passing in multiple object arguments to Object.assign. The spread syntax within an object is expecting a single object. So in this case, we're combining multiple objects with multiple string-key/string-value pairs.

rlr:
oh right, map returns an array. carry on ;)

bin/render-activity-stream-html.js
Mardak: `{...array}` does actually work but probably not what you expect ;) The array indices are added to the object as keys with matching values. Kinda like object-ifying an array.

PR #4901, opened 2019-04-11
Bug 1543436 - Updated usesFirefoxSync to use UIState

dannycoates:
Testing the water with this one. @vbudhram @piatra f?

### DIFF
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -17,6 +17,8 @@ ChromeUtils.defineModuleGetter(this, "AppConstants",
   "resource://gre/modules/AppConstants.jsm");
 ChromeUtils.defineModuleGetter(this, "AttributionCode",
   "resource:///modules/AttributionCode.jsm");
+ChromeUtils.defineModuleGetter(this, "UIState",
+  "resource://services-sync/UIState.jsm");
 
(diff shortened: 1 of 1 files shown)


piatra:
Looks good to me. What happens on startup? You mentioned some delay before `state` is "accurate". If a user is signed in but `UIState` is not up to date yet then the getter retuns `false` and they see a FxA snippet?

dannycoates:
The pref is good enough at startup. I've updated the code to use that until UIState is ready.

piatra:
Then this looks good to me. You might need to update the mochitests as well 
https://github.com/mozilla/activity-stream/blob/48d1f4c662c7f086ee9691272c9a0e79203b3c0d/test/browser/browser_asrouter_targeting.js#L142-L150
Right now it's setting the pref.

piatra:
We should add a new tests that checks the `STATUS_LOGIN_FAILED` scenario (`usesFirefoxSync` is `false` and no messages are returned).
I think we can simply overwrite `UIState`:
```javascript
UIState.get = () => ({status: UIState.STATUS_LOGIN_FAILED});
```
As long as we restore it after the tests ends. If you switch your PR to `Ready for review` you can have the automated tests run on your PR without having to build it yourself locally.

piatra:
The existing approach is fine for the use case so we're not going to update this.

PR #4904, opened 2019-04-15
bug 1543800 - local layout

ScottDowne:
To test:

1. reset `browser.newtabpage.activity-stream.discoverystream.config`
2. Load a new tab, and ensure it is displaying basiclayout.
3. Updates `browser.newtabpage.activity-stream.discoverystream.config` so hardcoded_layout is true
4. Load a new tab, and ensure it is displaying the hardcoded layout.

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -239,6 +239,7 @@ const PREFS_CONFIG = new Map([
         collapsible: true,
         enabled: isEnabled,
         show_spocs: showSpocs({geo}),
+        hardcoded_layout: false,
         personalized: false,
         // This is currently an exmple layout used for dev purposes.
         layout_endpoint: "https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=basic",
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -29,251 +29,7 @@ const PREF_SHOW_SPONSORED = "showSponsored";
 const PREF_SPOC_IMPRESSIONS = "discoverystream.spoc.impressions";
 const PREF_REC_IMPRESSIONS = "discoverystream.rec.impressions";
 
-const DEFAULT_LAYOUT_RESP = {
-  "spocs": {
-    "url": "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs?consumer_key=$apiKey",
-    "spocs_per_domain": 1,
-  },
-  "layout": [
-    {
-      "width": 12,
-      "components": [
-        {
-          "type": "TopSites",
-          "header": {
-            "title": "Top Sites",
-          },
-          "properties": {},
-        },
-      ],
-    },
-    {
-      "width": 12,
-      "components": [
-        {
-          "type": "Message",
-          "header": {
-            "title": "Recommended by Pocket",
-            "subtitle": "",
-            "link_text": "How it works",
-            "link_url": "https://getpocket.com/firefox/new_tab_learn_more",
-            "icon": "resource://activity-stream/data/content/assets/glyph-pocket-16.svg",
-          },
-          "properties": {},
-          "styles": {
-            ".ds-message": "margin-bottom: -20px",
-          },
-        },
-      ],
-    },
-    {
-      "width": 8,
-      "components": [
-        {
-          "type": "List",
-          "header": {
-            "title": "",
-          },
-          "feed": {
-            "embed_reference": null,
-            "url": "https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=en-US",
-          },
-          "properties": {
-            "items": 8,
-            "has_numbers": false,
-            "has_images": true,
-            "border": "no-border",
-          },
-          "styles": {
-            ".ds-list": "margin-right: -12px",
-          },
-        },
-      ],
-    },
-    {
-      "width": 4,
-      "components": [
-        {
-          "type": "CardGrid",
-          "properties": {
-            "items": 1,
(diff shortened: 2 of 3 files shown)


lib/DiscoveryStreamFeed.jsm
```diff
@@ -40,11 +286,35 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
     if (url.includes("$apiKey") && !apiKey) {
       throw new Error(`Layout Endpoint - An API key was specified but none configured: ${url}`);
     }
-    return url.replace("$apiKey", apiKey);
+    return this.replaceEndpointApiKey(url, apiKey);
+  }
+
+  replaceDefaultLayoutApiKey(apiKey) {
+    if (!apiKey) {
+      throw new Error(`Default Layout - An API key was specified but none configured`);
```
Mardak: Not sure if we need to conditionally inject api key or just always try. Layout, component, spocs all go through `async fetchFromEndpoint` anyway, so unconditionally replace any $apiKey in the url with the pref's value.

lib/ActivityStream.jsm
```diff
@@ -239,6 +239,7 @@ const PREFS_CONFIG = new Map([
         collapsible: true,
         enabled: isEnabled,
         show_spocs: showSpocs({geo}),
+        hardcoded_layout: false,
         personalized: false,
         // This is currently an exmple layout used for dev purposes.
         layout_endpoint: "https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=basic",
```
Mardak: Do we want to keep both values (layout_endpoint and hardcoded_layout) vs just having layout_endpoint be a special value. Or perhaps even if there's no layout_endpoint defined, assume it's the default hardcoded layout object.

ScottDowne: Went through a few ideas on this, ended up picking this one because it makes it easier to toggle between them without having to remember or reset the url endpoint value constantly.

Mardak: Not a huge issue. Seems like it could introduce some testing confusion "I set the endpoint but nothing changed" If this is for developer convenience, there is about:newtab#devtools-ds to switch some variants.

lib/DiscoveryStreamFeed.jsm
ScottDowne: I tried to go through the pros and cons, and while I think there are some arguments for and against both ways, I couldn't come up with any strong reason one way or another, so I'm happy to make this change or not.

lib/ActivityStream.jsm
Mardak: Although changing the url to something special would then break that devtools behavior 😉

lib/DiscoveryStreamFeed.jsm
```diff
@@ -29,6 +29,252 @@ const PREF_SHOW_SPONSORED = "showSponsored";
 const PREF_SPOC_IMPRESSIONS = "discoverystream.spoc.impressions";
 const PREF_REC_IMPRESSIONS = "discoverystream.rec.impressions";
 
+const DEFAULT_LAYOUT_RESP = {
```
k88hudson: maybe put it at the bottom so you don't have to scroll as much? :P

lib/DiscoveryStreamFeed.jsm
ScottDowne: OK, thinking about this, I think I agree with this change.

ScottDowne: I think all I needed was a comment in the replace that clears up my main con for doing it in `fetchFromEndpoint`

ScottDowne: It's def a lot less code.

lib/ActivityStream.jsm
ScottDowne: Yeah I considered the test confusion too but this felt more explicit.

lib/DiscoveryStreamFeed.jsm
ScottDowne: Like, put the const itself at the bottom of the file?

ScottDowne: I don't know if I follow.

Mardak: Maybe like how we hide things at the end here:
https://github.com/mozilla/activity-stream/blob/27af57c7499e5735f2a780ef1ad1127cd7687c57/lib/FilterAdult.jsm#L9-L10

ScottDowne:
Some changes based on comments and some tests.

lib/DiscoveryStreamFeed.jsm

k88hudson:
Also, can you file a follow-up ticket to turn this on by default for Firefox 68?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -181,6 +193,16 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
         Cu.reportError("No response for response.layout prop");
       }
     }
+    return layout;
+  }
+
+  async loadLayout(sendUpdate, isStartup) {
+    let layout = {};
+    if (this.config.hardcoded_layout) {
+      layout = {lastUpdate: Date.now(), ...defaultLayoutResp};
```
Mardak: This bypasses the call to `fetchLayout` which adds a layout entry to the cache. Currently the cache is used to determine if things should be expired, so lacking the entry makes it seem like the layout is always expired:
https://github.com/mozilla/activity-stream/blob/48d1f4c662c7f086ee9691272c9a0e79203b3c0d/lib/DiscoveryStreamFeed.jsm#L157-L161

And because layout appears to be expired, it triggers `loadLayout` leading to the spocs endpoint being "updated" and sets loaded status to false and prevents DiscoveryStreamBase from rendering.

PR #4905, opened 2019-04-15
Bug 1544574 - Simplify SPOC impression handler of DiscoveryStreamFeed

ncloudioj:


lib/DiscoveryStreamFeed.jsm
```diff
@@ -462,8 +462,10 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           return isBelow;
         }),
       };
-      // send caps to redux
-      this.store.dispatch({type: at.DISCOVERY_STREAM_SPOCS_CAPS, data: caps});
+      // send caps to redux if any.
+      if (caps.length) {
```
ScottDowne: Did you notice this being possible? (I'm not surprised, but just curious)

ScottDowne: Oh I see.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -825,16 +827,19 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
-          this.store.dispatch(ac.AlsoToPreloaded({
-            type: at.DISCOVERY_STREAM_SPOCS_UPDATE,
-            data: {
-              lastUpdated: spocs.lastUpdated,
-              spocs: this.transform(this.frequencyCapSpocs(spocs.data)),
-            },
-          }));
+          // Apply frequency capping to SPOCs in the redux store, only update the
+          // store if the SPOCs are changed.
+          const {spocs} = this.store.getState().DiscoveryStream;
```
ScottDowne: What happens if spocs is an empty array, and can that even happen?

ScottDowne: Nope, the reducer defaults it to a proper empty state. :+1:

ncloudioj:
@ScottDowne Nice catch.

The current reducer for `DISCOVERY_STREAM_SPOCS_CAPS` does the replacement for `frequency_caps` because it re-computes everything from the cache. I've changed it to chain the new frequency_caps to the existing array so that it can track the frequency capped items for the whole cycle until the next `loadSpocs`.

PR #4906, opened 2019-04-16
Bug 1544627 - render ASRouterUISurface from Base.jsx

rlr:


content-src/components/Base/Base.jsx
```diff
@@ -87,7 +88,10 @@ export class _Base extends React.PureComponent {
     if (prefs["asrouter.devtoolsEnabled"]) {
       if (window.location.hash.startsWith("#asrouter") ||
           window.location.hash.startsWith("#devtools")) {
-        return (<ASRouterAdmin />);
+        return (<div>
+          <ASRouterAdmin />
+          <ASRouterUISurface dispatch={this.props.dispatch} />
```
k88hudson: Ideally it would be great not to render this in two places, although I think for now this is probably your best option

PR #4909, opened 2019-04-16
Bug 1548329 - Load discovery stream feeds sequentially without blocking

ScottDowne:
To test, turn DS on and off, and watch it load.

If you watch it closely, you can see it loading from top to bottom as feeds are ready.

I ran this on a reference machine, and got these medians:

cached:
control:7.85/patch:6.61
no cached:
control:7.66/patch:7.3

We see the biggest improvement with cached, not totally sure why, but these show improvements in both.

Look for other regressions.

The reason why this works is the ref machine considers ready when above the fold is loaded with images. This makes it so the higher feeds load in potentially sooner than the lower ones, which means above the fold can get priority.

Making a bug for this now...


lib/DiscoveryStreamFeed.jsm
```diff
@@ -537,8 +554,16 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   async getComponentFeed(feedUrl, isStartup) {
     const cachedData = await this.cache.get() || {};
     const {feeds} = cachedData;
+    if (!this.timer) {
```
ScottDowne: This is all just for debugging purposes. It'll be gone once I've finished it up.

common/Reducers.jsm
```diff
@@ -500,10 +500,22 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
         ...prevState,
         feeds: {
           ...prevState.feeds,
-          data: action.data || prevState.feeds.data,
           loaded: true,
         },
       };
+    case at.DISCOVERY_STREAM_FEED_UPDATE:
+      const newData = {};
+      newData[action.data.url] = action.data.feed;
```
punamdahiya: is it possible to have action.data empty here, if yes we should have a safe check

punamdahiya:
@ScottDowne It will be good to resolve conflicts in DiscoveryStreamBase and selectLayoutRender for testing the patch.  Thanks!

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -169,10 +169,7 @@ export class _DiscoveryStreamBase extends React.PureComponent {
   render() {
     // Select layout render data by adding spocs and position to recommendations
     const layoutRender = selectLayoutRender(this.props.DiscoveryStream, this.props.Prefs.values, rickRollCache);
-    const {config, feeds, spocs} = this.props.DiscoveryStream;
-    if (!spocs.loaded || !feeds.loaded) {
-      return null;
-    }
```
punamdahiya: With this removal we are no longer waiting for both spocs and feeds to finish loading before rendering. Are we checking later for a component we have both spocs and feeds before rendering?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -240,8 +240,16 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+
         feedPromise.then(feed => {
           newFeeds[url] = this.filterRecommendations(feed);
+          sendUpdate({
+            type: at.DISCOVERY_STREAM_FEED_UPDATE,
+            data: {
+              feed: newFeeds[url],
+              url,
+            },
+          });
```
punamdahiya: That's where we are hoping big win by updating state right when we are ready with feed data. It's sequential that we are invoking buildFeedPromise in sequential manner, though its possible for a higher (lower on the page) feedPromise to resolve early and sendUpdate , is that correct?

test/unit/common/Reducers.test.js
```diff
@@ -684,18 +684,6 @@ describe("Reducers", () => {
       const state = DiscoveryStream(undefined, {type: at.DISCOVERY_STREAM_CONFIG_CHANGE, data: {enabled: true}});
       assert.deepEqual(state.config, {enabled: true});
     });
-    it("should load feeds with DISCOVERY_STREAM_FEEDS_UPDATE", () => {
```
punamdahiya: Should have a test for updated DISCOVERY_STREAM_FEEDS_UPDATE

test/unit/common/Reducers.test.js
```diff
@@ -723,11 +711,6 @@ describe("Reducers", () => {
       const newState = DiscoveryStream(undefined, {type: at.PLACES_LINK_BLOCKED});
       assert.equal(newState, INITIAL_STATE.DiscoveryStream);
     });
-    it("should not update state if feeds are not loaded", () => {
-      const deleteAction = {type: at.PLACES_LINK_BLOCKED, data: {url: "foo.com"}};
```
punamdahiya: this is changed with spoc_fills patch

test/unit/content-src/components/DiscoveryStreamBase.test.jsx
```diff
@@ -87,18 +87,6 @@ describe("<DiscoveryStreamBase>", () => {
...
-
-    assert.equal(wrapper.type(), null);
-  });
-
-  it("should render null if feeds not loaded", () => {
-    wrapper = mountComponent({feeds: {loaded: false}});
-
-    assert.equal(wrapper.type(), null);
-  });
-
```
punamdahiya: Does checking for notNull as we continuing minimal rendering even when spocs and feeds are not loaded help?

test/unit/content-src/lib/selectLayoutRender.test.js
```diff
@@ -36,21 +37,11 @@ describe("selectLayoutRender", () => {
...
-    store.dispatch({type: at.DISCOVERY_STREAM_LAYOUT_UPDATE, data: {layout: FAKE_LAYOUT}});
-    store.dispatch({type: at.DISCOVERY_STREAM_FEEDS_UPDATE, data: {}});
-
-    const result = selectLayoutRender(store.getState().DiscoveryStream, {}, []);
-
-    assert.lengthOf(result, 1);
-    assert.propertyVal(result[0], "width", 3);
-    assert.deepEqual(result[0].components[0], FAKE_LAYOUT[0].components[0]);
-  });
-
```
punamdahiya: Should we check for just store.dispatch({type: at.DISCOVERY_STREAM_FEEDS_UPDATE}); above.

lib/DiscoveryStreamFeed.jsm
punamdahiya: > That's where we are hoping big win by updating state right when we are ready with feed data. It's sequential that we are invoking buildFeedPromise in sequential manner, though its possible for a higher (lower on the page) feedPromise to resolve early and sendUpdate , is that correct?
Tests below "should stop rendering feeds if we hit one that's not ready" help explain component will be rendered on the page sequentially

common/Reducers.jsm
ScottDowne: Not currently possible, the place we call it is explicitly creating the data object when it calls dispatch, it's also not called if url doesn't connect with a feed. Feed itself could be undefined if the server is misconfigured and doesn't return anything originally, but I imagine that's actually fine here because we don't directly reference something on feed, we just pass it along. If the server starts doing that, we probably got much bigger problems :D

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
ScottDowne: Yup, this block right here in selectLayoutRender is the key:

```
      for (const component of row.components.filter(c => !filterArray.includes(c.type))) {
        if (component.feed) {
          const spocsConfig = component.spocs;
          // Are we still waiting on a feed/spocs, render what we have, and bail out early.
          if (!feeds.data[component.feed.url] ||
            (spocsConfig && spocsConfig.positions && spocsConfig.positions.length && !spocs.loaded)) {
            return renderedLayoutArray;
          }
          components.push(handleComponent(component));
        } else {
          components.push(component);
        }
      }
```

First it's if it's not a feed based component, if it expects a feed, then it's checking if the feed exists yet, if it doesn't it bails and returns what it has up to this point. So if feed 1 is done, and feed 2 is loading still, we can render feed 1 and keep waiting on feed 2. It the checks to see if this feed expects a spoc and if spocs are loaded, if not, it considers the whole thing not ready too, and returns what we have, and keeps waiting.

ScottDowne: So it's less of a "check to make sure we have everything before rendering" and more of a "filter out the stuff we cannot render before rendering"

lib/DiscoveryStreamFeed.jsm
ScottDowne: Yup, it can be ready sooner, and when we go to render, we wouldn't attempt to render it unless everything before it is also ready.

So think of the rendering as sequential and not the fetching.

I tested a version where the fetches were also sequential, where the `feedPromise.then(feed => {` would chain, but it was 1. Way more complex, 2. Didn't impress me as being much faster, and 3. Could actually be slower.

The way I have it, if the last feed to load is blocking others, once it's done, instead of now having to fire off a bunch more `DISCOVERY_STREAM_FEED_UPDATE` we can just fire off the last 1 and the rest of the data should just be there and things should render completely in one go.

test/unit/common/Reducers.test.js
ScottDowne: That patch changes a lot of things for me :D

ScottDowne: Not sure why I deleted this test, it looks unrelated and seems to pass just fine. Possibly a mistake of some sort. Adding it back.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -170,14 +170,11 @@ export class _DiscoveryStreamBase extends React.PureComponent {
...
+    const {config, spocs, feeds} = this.props.DiscoveryStream;
 
     // Send SPOCS Fill if any. Note that it should not send it again if the same
     // page gets re-rendered by state changes.
-    if (spocsFill.length && !this._spocsFillSent) {
+    if (spocs.loaded && feeds.loaded && spocsFill.length && !this._spocsFillSent) {
```
ncloudioj: When the subsequent feed loads complete and trigger the state changes, which in turn might generate some new spocs fill pings. Is it possible that those pings would never get sent if it has already sent one earlier?

ScottDowne: Hm, not sure I follow, but the intended solution here is to not send a ping until everything is loaded, it would then go ahead and fire the ping, and set `_spocsFillSent` to true, then next time it comes in here, it would again see it as loaded, but this time `_spocsFillSent` is true, so it won't send again.

ncloudioj: Ah, I see. That's why you've added those `loaded` checks here. I was wondering if we can do the same check inside of `selectLayoutRender` [here](https://github.com/mozilla/activity-stream/blob/master/content-src/lib/selectLayoutRender.js#L115). I.e. if either SPOCS or feeds has not been fully loaded, then don't bother calculating the SPOCS Fill at all.

ScottDowne:
Back in review!

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
punamdahiya: Since spocs.loaded is a needed for rendering if a feed expects a spoc , should we have loadSpocs before loadComponentFeeds in refreshAll https://github.com/mozilla/activity-stream/pull/4909/files#diff-8b1b03c4d5dec5c8e4f7de50e05c0c4fR679

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -208,6 +205,10 @@ export class _DiscoveryStreamBase extends React.PureComponent {
     // Get "topstories" Section state for default values
     const topStories = this.props.Sections.find(s => s.id === "topstories");
 
+    if (!topStories) {
+      return null;
+    }
+
     // Extract TopSites to render before the rest and Message to use for header
     const topSites = extractComponent("TopSites");
     const message = extractComponent("Message") || {
```
punamdahiya: For non-cached, first load we can avoid extractComponent multiple calls if topSites is not null and message & message.type is Message

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
ScottDowne: I'll give that a try. Should work. :+1:

lib/DiscoveryStreamFeed.jsm
```diff
@@ -606,8 +615,10 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   async getComponentFeed(feedUrl, isStartup) {
     const cachedData = await this.cache.get() || {};
     const {feeds} = cachedData;
+
     let feed = feeds ? feeds[feedUrl] : null;
-    if (this.isExpired({cachedData, key: "feed", url: feedUrl, isStartup})) {
+    const isExpired = this.isExpired({cachedData, key: "feed", url: feedUrl, isStartup});
+    if (isExpired) {
```
punamdahiya: do we need const isExpired or is it for code readability?

ScottDowne: I think I did this because I was logging stuff while debugging, and needed to explicitly log that value, I jut left it because it seemed harmless. I can revert it.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
punamdahiya: As discussed over slack , optimizing these calls inside selectLayoutRender can be a refactoring in a separate bug

PR #4912, opened 2019-04-17
Bug 1543686 - Add ability to place a snippet below the Activity Stream search box

rlr:
@k88hudson thoughts on the portal approach?

### DIFF
--- a/content-src/asrouter/asrouter-content.jsx
+++ b/content-src/asrouter/asrouter-content.jsx
@@ -7,12 +7,14 @@ import {LocalizationProvider} from "fluent-react";
 import {NEWTAB_DARK_THEME} from "content-src/lib/constants";
 import {OnboardingMessage} from "./templates/OnboardingMessage/OnboardingMessage";
 import React from "react";
+import ReactDOM from "react-dom";
 import {ReturnToAMO} from "./templates/ReturnToAMO/ReturnToAMO";
 import {SnippetsTemplates} from "./templates/template-manifest";
 import {StartupOverlay} from "./templates/StartupOverlay/StartupOverlay";
 
 const INCOMING_MESSAGE_NAME = "ASRouter:parent-to-child";
 const OUTGOING_MESSAGE_NAME = "ASRouter:child-to-parent";
+const TEMPLATES_BELOW_SEARCH = ["simple_below_search_snippet"];
 
 export const ASRouterUtils = {
   addListener(listener) {
@@ -93,6 +95,9 @@ export class ASRouterUISurface extends React.PureComponent {
     this.sendImpression = this.sendImpression.bind(this);
     this.sendUserActionTelemetry = this.sendUserActionTelemetry.bind(this);
     this.state = {message: {}, bundle: {}};
+    if (props.document) {
+      this.portalContainer = props.document.getElementById("footer-snippets-container");
+    }
   }
 
   sendUserActionTelemetry(extraProps = {}) {
@@ -299,14 +304,22 @@ export class ASRouterUISurface extends React.PureComponent {
   render() {
     const {message, bundle} = this.state;
     if (!message.id && !bundle.template) { return null; }
-    return (
-      <React.Fragment>
-        {this.renderPreviewBanner()}
-        {this.renderFirstRunOverlay()}
-        {this.renderOnboarding()}
-        {this.renderSnippets()}
-      </React.Fragment>
-    );
+    const shouldRenderBelowSearch = TEMPLATES_BELOW_SEARCH.includes(message.template);
+
+    return shouldRenderBelowSearch ?
+      // Render special below search snippets in place;
+      <div className="below-search-snippet">{this.renderSnippets()}</div> :
+      // For onboarding, regular snippets etc. we should render
+      // everything in our footer container.
+      ReactDOM.createPortal(
+        <>
+          {this.renderPreviewBanner()}
+          {this.renderFirstRunOverlay()}
+          {this.renderOnboarding()}
+          {this.renderSnippets()}
+        </>,
+        this.portalContainer
+      );
   }
 }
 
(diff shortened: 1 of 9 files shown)


content-src/asrouter/asrouter-content.jsx
```diff
@@ -299,14 +312,22 @@ export class ASRouterUISurface extends React.PureComponent {
...
+
+    if (!TEMPLATES_BELOW_SEARCH.includes(message.template)) {
+      return ReactDOM.createPortal(
+        <React.Fragment>
+          {this.renderPreviewBanner()}
+          {this.renderFirstRunOverlay()}
+          {this.renderOnboarding()}
+          {this.renderSnippets()}
+        </React.Fragment>,
+        this.footerPortal
```
k88hudson: What's the  behaviour of this if `footerPortal` is `undefined`? (like in tests for whatever

rlr: I had to put in a fake portal container for tests.

content-src/asrouter/asrouter-content.jsx
```diff
@@ -299,14 +312,22 @@ export class ASRouterUISurface extends React.PureComponent {
...
-    return (
-      <React.Fragment>
-        {this.renderPreviewBanner()}
-        {this.renderFirstRunOverlay()}
-        {this.renderOnboarding()}
-        {this.renderSnippets()}
-      </React.Fragment>
-    );
+
+    if (!TEMPLATES_BELOW_SEARCH.includes(message.template)) {
```
k88hudson: Personally i'd use an affirmative variable for this like 

```js
render() {
  const {message, bundle} = this.state;
  if (!message.id && !bundle.template) { return null; }
  const shouldRenderBelowSearch = TEMPLATES_BELOW_SEARCH.includes(message.template);

  return shouldRenderBelowSearch ?
    // Render special below search snippets in place;
    this.renderSnippets() :
    // For onboarding, regular snippets etc. we should render
    // everything in our footer container.
    ReactDOM.createPortal(
      <>
        {this.renderPreviewBanner()}
        {this.renderFirstRunOverlay()}
        {this.renderOnboarding()}
        {this.renderSnippets()}
      </>,
      this.footerPortal
    );
}
```

content-src/components/Base/Base.jsx
```diff
@@ -159,6 +159,9 @@ export class BaseContent extends React.PureComponent {
                 </ErrorBoundary>
               </div>
             }
+            <div className="below-search-snippet">
```
k88hudson: If you want to leave this here, make sure the empty state doesn't have any extra padding. Do you need the container div?

content-src/asrouter/asrouter-content.jsx
k88hudson: ok cool

content-src/components/Base/Base.jsx
rlr: So... yeah. ick. I am setting a `min-height` on it to get the spacing between search and top sites right in both cases: with and without snippet. I'll try and see if I can come up with something cleaner.

content-src/asrouter/asrouter-content.jsx
rlr: much better 👍

content-src/components/Base/Base.jsx
rlr: ok. was able to fix that 💃

content-src/asrouter/asrouter-content.jsx
```diff
@@ -93,6 +95,15 @@ export class ASRouterUISurface extends React.PureComponent {
     this.sendImpression = this.sendImpression.bind(this);
     this.sendUserActionTelemetry = this.sendUserActionTelemetry.bind(this);
     this.state = {message: {}, bundle: {}};
+
+    if (props.document) {
+      // Create the portal element.
+      const footerContainer = props.document.getElementById("footer-snippets-container");
+      this.footerPortal = document.createElement("div");
```
k88hudson: can you put this inside the `if` since you're not going to use it if `footerContainer` doesn't exist? also might be worth logging a warning here...

content-src/asrouter/asrouter-content.jsx
```diff
@@ -211,6 +222,8 @@ export class ASRouterUISurface extends React.PureComponent {
 
   componentWillUnmount() {
     ASRouterUtils.removeListener(this.onMessageFromParent);
+
+    this.footerPortal.parent.removeChild(this.footerPortal);
```
k88hudson: probably want to wrap this with an `if(this.footerPortal)`

content-src/asrouter/asrouter-content.jsx
```diff
@@ -299,14 +301,22 @@ export class ASRouterUISurface extends React.PureComponent {
...
+      // For onboarding, regular snippets etc. we should render
+      // everything in our footer container.
+      ReactDOM.createPortal(
+        <>
+          {this.renderPreviewBanner()}
+          {this.renderFirstRunOverlay()}
+          {this.renderOnboarding()}
+          {this.renderSnippets()}
+        </>,
+        this.props.document.getElementById("footer-snippets-container")
```
k88hudson: you could still cache this reference

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -0,0 +1,51 @@
...
+    border-radius: 4px;
+  }
+
+  .innerWrapper {
+    flex-direction: column;
+    padding: 15px;
+    text-align: center;
+    width: 100%;
+
+    @media (min-width: $break-point-medium) {
```
k88hudson: this looks pretty weird with DiscoveryStream enabled right now, I think we should limit this to non-ds right now; do you mind changing that and adding a comment?

content-src/components/Search/_Search.scss
```diff
@@ -137,6 +131,39 @@ $glyph-forward: url('chrome://browser/skin/forward.svg');
...
+  }
+
+  @media (min-width: $break-point-large) {
+    width: $searchbar-width-large;
+  }
+}
+
+.non-collapsible-section + .below-search-snippet {
+  // If search is enabled, we need to invade its large bottom padding.
+  margin-top: -44px;
```
k88hudson: Personally I think this looks right around `-50px` but might be good to check with wolasi (feel free to land and we can follow-up with final tweaks)

PR #4913, opened 2019-04-17
bug 1545194 - v1 personalization for 68

ScottDowne:


ScottDowne:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1546129 for figuring out what to do with existing profiles and data.

PR #4914, opened 2019-04-17
Bug 1545503 – Improve image quality

gvn:
Testing: 

- Try some layout variants (recommend: `66-beta-2-complex`, `basic`, `dev-test-all`).
- Ensure image quality looks good


gvn:
I realized that the strategy I was using was a bit too aggressive in some cases and have created a PR that should improve image quality. Some aspect ratios were getting pulled at too small of a resolution before.

This will pull a specific width *AND* height based on the container size.

ScottDowne:
Test failure fyi, but I'll start reading the code now.

ScottDowne:
Your test failure is probably something that needs to be updated in the test and not your code, is my guess.

Also this looks fine from a high level. I'm a bit curious about what changed that let us remove that cache file, was it just not working as expected? What was it's goal in the first place? etc.

gvn:
Cache is maybe not the right word for it. I was doing memoization so that if a larger version of an image got pulled already we wouldn't then pull another smaller version. However, with pulling width and height that becomes a lot more convoluted. Also, the likelihood of a story being repeated many times in a non dev layout is pretty low, so it's adding a lot of complexity for an edge case. Perf still seemed very good without it and with height specified the images are actually even smaller in some cases/aspect ratios.

PR #4916, opened 2019-04-18
Bug 1500061 - Switch to regular script tags, tighten CSP

k88hudson:
Finally figured this out while I was looking into pre-rendering stuff.

Test failures before were happening because our script injection was happening from an inline script that we render for the non-noscript version (which is currently on in release and asan builds but not nightly)

To manually test all variants, check:
* prerendered (turn ds off)
* noscripts on (default on nightly)
* noscripts off: set `browser.tabs.remote.separatePrivilegedContentProcess` to `false`
* `browser.newtabpage.activity-stream.debug` on

I'm a bit worried this could result in a perf hit since loading an external resource is likely to be a bit slower than inline, what do you think?

Tests: https://treeherder.mozilla.org/#/jobs?repo=try&revision=60d5046b3ccc0d8ea5a245ce8c82de691a45f582

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -93,7 +93,7 @@ function templateHTML(options, html) {
 <html lang="${options.locale}" dir="${options.direction}">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; object-src 'none'; script-src resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
     <title>${options.strings.newtab_page_title}</title>
     <link rel="icon" type="image/png" href="chrome://branding/content/icon32.png"/>
     <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -85,18 +85,10 @@ function templateHTML(options, html) {
   if (isPrerendered) {
     scripts.unshift(`${options.baseUrl}prerendered/static/activity-stream-initial-state.js`);
   }
-  const scriptTag = `
-    <script>
-// Don't directly load the following scripts as part of html to let the page
-// finish loading to render the content sooner.
-for (const src of ${JSON.stringify(scripts, null, 2)}) {
-  // These dynamically inserted scripts by default are async, but we need them
-  // to load in the desired order (i.e., bundle last).
-  const script = document.body.appendChild(document.createElement("script"));
-  script.async = false;
-  script.src = src;
-}
-    </script>`;
+
+  // Add spacing and script tags
+  const scriptRender = `\n${scripts.map(script => `    <script src="${script}"></script>`).join("\n")}`;
+
   return `<!doctype html>
 <html lang="${options.locale}" dir="${options.direction}">
   <head>
@@ -109,7 +101,7 @@ for (const src of ${JSON.stringify(scripts, null, 2)}) {
   </head>
   <body class="activity-stream">
     <div id="root">${isPrerendered ? html : "<!-- Regular React Rendering -->"}</div>
-    <div id="footer-snippets-container" />${options.noscripts ? "" : scriptTag}
+    <div id="footer-snippets-container" />${options.noscripts ? "" : scriptRender}
   </body>
 </html>
 `;


bin/render-activity-stream-html.js
```diff
@@ -156,6 +169,7 @@ function writeFiles(name, destPath, filesMap, {html, state}, options) {
 
 const STATIC_FILES = new Map([
   ["activity-stream-debug.html", ({options}) => templateHTML(options)],
+  ["activity-stream-load-scripts.js", ({options}) => templateLoadScriptJs(options)],
```
k88hudson: I think I need to add another render for the debug version of this technically, not sure if we care about that or just always use the `noscripts` version for debug?

bin/render-activity-stream-html.js
```diff
@@ -69,54 +85,51 @@ function getTextDirection(locale) {
...
-    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; object-src 'none'; script-src resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
     <title>${options.strings.newtab_page_title}</title>
     <link rel="icon" type="image/png" href="chrome://branding/content/icon32.png"/>
     <link rel="stylesheet" href="chrome://browser/content/contentSearchUI.css" />
     <link rel="stylesheet" href="${options.baseUrl}css/activity-stream.css" />
   </head>
   <body class="activity-stream">
-    <div id="root">${isPrerendered ? html : "<!-- Regular React Rendering -->"}</div>
+    <div id="root"></div>
```
Mardak: Nothing will get prerendered without the `html` (unless I'm missing where it gets prerendered elsewhere??)

k88hudson: Whoops sorry, the `html` should still be in there

bin/render-activity-stream-html.js
Mardak: Hmm.. the test failures are complaining that the react.js and react-dom.js versions are unreferenced, so that means activity-stream-load-scripts.js is rendering only with the debug=true ? We definitely want the main file to be the non-dev versions. Still thinking about the overall debug situation… 🤔

Mardak: Ah yup static assumes debug… … … was that intended? 🤔 
https://github.com/mozilla/activity-stream/blob/a3f538fd0cd8b8c121b8f0b21ce89775dbecd1e9/bin/render-activity-stream-html.js#L242-L243

k88hudson: yeah, I think I need to just switch it to render those separately

Mardak: I think we can work with a single load-scripts.js file that dynamically determines its debug context.

rendered debug html:
```html
<script src="load-scripts.js#debug"/>
```

rendered not debug html:
```html
<script src="load-scripts.js#not-debug"/>
```

load-scripts.js:
```js
const isDebug = document.currentScript.src…;
const scripts = [
   // NB: use the full resource to make referenced test happy
  isDebug ? "resource://…react-dev.js" : "resource://…react.js"
…
```

Alternatively just package both a load-scripts-debug.js and load-scripts.js 😉

bin/render-activity-stream-html.js
```diff
@@ -54,6 +54,22 @@ function getTextDirection(locale) {
...
+  return [
+    "chrome://browser/content/contentSearchUI.js",
+    "chrome://browser/content/contentTheme.js",
+    `${options.baseUrl}vendor/react${debugString}.js`,
+    `${options.baseUrl}vendor/react-dom${debugString}.js`,
+    `${options.baseUrl}vendor/prop-types.js`,
+    `${options.baseUrl}vendor/react-intl.js`,
+    `${options.baseUrl}vendor/redux.js`,
+    `${options.baseUrl}vendor/react-redux.js`,
+    `${options.baseUrl}prerendered/${options.locale}/activity-stream-strings.js`,
```
Mardak: oh wait. If we're loading the locale strings here, this can't be rendered "static" as it changes per locale.

bin/render-activity-stream-html.js
```diff
@@ -85,33 +71,48 @@ function templateHTML(options, html) {
...
+  ].filter(i => i);
   return `<!doctype html>
 <html lang="${options.locale}" dir="${options.direction}">
+  <!-- Render info: ${renderInfo.join(", ")} -->
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; object-src 'none'; script-src resource: chrome:; connect-src https:; img-src https: data: blob:; style-src 'unsafe-inline';">
     <title>${options.strings.newtab_page_title}</title>
+    <meta name="scripts" content="${getScripts(options, isPrerendered).join(",")}" />
```
Mardak: Interesting approach… seems a little scary having a script that loads more, but I guess it works. Test is complaining because it's only looking for quoted chrome/resource uris. I think if you stringify, you'll happen to trick the test to be happy. I.e., `content='["resource://activity-stream/…","resource:…"]'` instead of `content="resource://activity-stream/…,resource:…"`

https://searchfox.org/mozilla-central/rev/d302c3058330a57f238be4062fddea629311ce66/browser/base/content/test/static/browser_all_files_referenced.js#421-424

Mardak: Ack. The logic there only checks for resource uris only if there are no chrome uris found on the line. I suppose you can just render as html comments the scripts that are included in the meta ?

Mardak: This makes the test happy…
```js
    <!-- Dynamically load scripts:
${getScripts(options, isPrerendered).map(s => `"${s}"`).join("\n")}
    -->
```

PR #4918, opened 2019-04-19
Bug 1545487 - Remove browser migration (import bookmarks, history, etc) message from Activity Stream

rlr:


content-src/components/ManualMigration/ManualMigration.jsx
```diff
@@ -1,49 +0,0 @@
...
-  onCancelTour() {
-    this.props.dispatch(ac.AlsoToMain({type: at.MIGRATION_CANCEL}));
-    this.props.dispatch(ac.UserEvent({event: at.MIGRATION_CANCEL}));
-  }
-
-  render() {
-    return (<div className="manual-migration-container">
-        <p>
-          <span className="icon icon-import" />
-          <FormattedMessage id="manual_migration_explanation2" />
```
Mardak: The icon and strings will be unused, so we can probably remove those?

lib/ActivityStream.jsm
```diff
@@ -93,18 +92,6 @@ const PREFS_CONFIG = new Map([
     title: "Remove adult pages from sites, highlights, etc.",
     value: true,
   }],
-  ["migrationExpired", {
-    title: "Boolean flag that decides whether to show the migration message or not.",
-    value: false,
```
Mardak: These prefs most likely will have user set values, so they'll stick around even when we the default values. That could be useful to keep around if something else might depend on those values down the line, so I guess we shouldn't proactively clean them up yet…

content-src/components/ManualMigration/ManualMigration.jsx
rlr: Oh the strings. Do we need to wait a couple of releases before removing them?

Mardak: Nah, strings still exist on beta, and localizers see a unified view of all strings across channels.

rlr: ooh, it was system addon days where we had to wait I guess. good old days

rlr: and by system addon i guess i meant test pilot 😉

PR #4920, opened 2019-04-22
Bug 1543800 - Landing local layout 66-release-2-above-6-below-4 for D…

ScottDowne:
…iscovery Stream for 68

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -930,38 +930,12 @@ defaultLayoutResp = {
       ],
     },
     {
-      "width": 8,
-      "components": [
-        {
-          "type": "List",
-          "header": {
-            "title": "",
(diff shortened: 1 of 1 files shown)


Mardak:
The "turn on" bug https://bugzilla.mozilla.org/show_bug.cgi?id=1545122 was marked duplicate of the bug numbered here that previously landed #4904. Are we supposed to flip the config value for `hardcoded_layout` here?

ScottDowne:
OK I see what happened with those bugs. After I landed the initial patch, more comments happened, which for some reason made me think it was another bug similar to the one I filed, thought maybe someone already filed it so closed mine. Either way, my bug I marked as a dupe was badly titled anyway. Looks like we don't actually have an adequate bug for this.

ScottDowne:
Just went with a new pr.

PR #4925, opened 2019-04-23
Bug 1544574 - Add a followup fix to handle the special case

ncloudioj:


lib/DiscoveryStreamFeed.jsm
```diff
@@ -856,7 +856,8 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           // store if the SPOCs are changed.
           const {spocs} = this.store.getState().DiscoveryStream;
           const newSpocs = this.frequencyCapSpocs(spocs.data);
-          if (spocs.data.spocs.length !== newSpocs.spocs.length) {
+          if (spocs.data.spocs && newSpocs.spocs &&
+            spocs.data.spocs.length !== newSpocs.spocs.length) {
```
ScottDowne: What do you think about something like this:

```
const prevSpocs = spocs.data.spocs || [];
const newSpocs = newSpocs.spocs || [];
```

Now you can compare those values lengths directly?

PR #4927, opened 2019-04-23
(nobug) Make button for collapsing devtools less prominent

k88hudson:
At aaron's request, slight design change for the devtools button.

image

### DIFF
--- a/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
+++ b/content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
@@ -805,11 +805,15 @@ export class CollapseToggle extends React.PureComponent {
   render() {
     const {props} = this;
     const {renderAdmin} = this;
-    const action = (this.state.collapsed || !renderAdmin) ? "Expand" : "Collapse";
+    const isCollapsed = this.state.collapsed || !renderAdmin;
+    const label = `${isCollapsed ? "Expand" : "Collapse"} devtools`;
     return (<React.Fragment>
       <a href="#devtools"
-        className={`asrouter-toggle asrouter-${action.toLowerCase()}`}
-        onClick={this.renderAdmin ? this.onCollapseToggle : null}>{action} Devtools</a>
+        title={label}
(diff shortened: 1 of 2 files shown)


k88hudson:
Oops just saw e35dec51aa0162cfe82ab5bdbf0f6a24d7f53034, going to ask if aaron still wants this

k88hudson:
He does still want the change

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -805,11 +805,14 @@ export class CollapseToggle extends React.PureComponent {
...
     const {props} = this;
     const {renderAdmin} = this;
-    const action = (this.state.collapsed || !renderAdmin) ? "Expand" : "Collapse";
+    const label = `${(this.state.collapsed || !renderAdmin) ? "Expand" : "Collapse"} devtools`;
     return (<React.Fragment>
       <a href="#devtools"
-        className={`asrouter-toggle asrouter-${action.toLowerCase()}`}
-        onClick={this.renderAdmin ? this.onCollapseToggle : null}>{action} Devtools</a>
+        title={label}
+        className={`asrouter-toggle ${this.state.collapsed ? "collapsed" : "expanded"}`}
```
Mardak: Looks like state.collapsed is default false even on a plain about:newtab so it renders the className as expanded (resulting in background). Probably just needs the `|| !renderAdmin` as above.

k88hudson: ah, good point

rlr: I think you need `this.state.collapsed || !renderAdmin` here like above for the label. Right now it's showing the expanded background on the first render.

rlr: Or maybe that's intentional?

rlr: gah. Ed beat me.

PR #4928, opened 2019-04-24
Bug 1535717 - Add SPOCS Fill ping

ncloudioj:
This adds a new ping called "SPOCS Fill" for Discovery Stream.

It'll be reported in following cases:
* When the `loadSpocs` gets called in the `DiscoveryStreamFeed`, such as browser startup, scheduled feed update, pref changes etc.
* When `selectLayoutRender` gets called for the newtab page (including the preloaded ones).
* When a SPOC reaches its frequency cap.

To test it locally, you can flip following prefs in your local build:
* `browser.ping-centre.log`
* `browser.newtabpage.activity-stream.telemetry`
* `browser.newtabpage.activity-stream.telemetry.structuredIngestion`

Then open the browser console, add `spoc_fills` in the "filter output" section.

You might want to reset the Discovery Stream pref to reload the SPOCS from endpoint.


ncloudioj:
To test it locally, you can flip following prefs in your local build:
* `browser.ping-centre.log`
* `browser.newtabpage.activity-stream.telemetry`
* `browser.newtabpage.activity-stream.telemetry.structuredIngestion`

Then open the browser console, add `spoc_fills` in the "filter output" section.

You might want to reset the Discovery Stream pref to reload the SPOCS from endpoint.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -179,6 +180,11 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       return this.renderLayout(layoutRender);
     }
 
+    // Send SPOCS Fill if any.
+    if (spocsFill.length) {
+      this.props.dispatch(ac.DiscoveryStreamSpocsFill({spoc_fills: spocsFill}));
```
ScottDowne: Is this being after:
```
    if (!config.collapsible) {
      return this.renderLayout(layoutRender);
    }
```

An issue?

content-src/lib/selectLayoutRender.js
```diff
@@ -13,6 +16,11 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
         spocs.data.spocs && spocs.data.spocs.length) {
       const recommendations = [...data.recommendations];
       for (let position of spocsConfig.positions) {
+        const spoc = spocs.data.spocs[spocIndex];
+        if (!spoc) {
+          break;
```
ScottDowne: What's the purpose of this break?

ncloudioj: If there is no SPOCS left, I figured that we can break out from this for loop. Does it make sense?

ScottDowne: Ah ok I think I get it.

After walking through it, I think it's safe.

The spocIndex starts at 0 and just goes up. If at any point the spoc array has nothing from a 0 or greater, it's spent and there are no more spocs to even attempt to calculate.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
ScottDowne: Do you need to make sure this only gets called on initial render, or if something changes that causes the page to update and re render, does this get fired again?

example, if you block a spoc or rec from the story context menu, it causes the page to call render again.

content-src/lib/selectLayoutRender.js
```diff
@@ -94,4 +106,24 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
       return {...component, data};
     }),
   })).filter(row => row.components.length);
+
+  let spocsFill = [];
```
ScottDowne: What do we want to sent if spocs are not enaled? Right now I think you return `[]` if spocs are not enabled, does that make sense?

content-src/lib/selectLayoutRender.js
```diff
@@ -94,4 +106,24 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
...
+  if (spocs.data.spocs) {
```
ScottDowne: Is this the proper if statement?

I wonder if we want to instead check if spocs are enabled, and if the has already been sent? Because I *think* if there are no spocs, this could be because the server is out of spocs or maybe we had no available spocs left?

ScottDowne: I wonder if this has something to do with the `fullRecalc` stuff you have in the feed.jsm?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -442,7 +467,13 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
 
   filterBlocked(data, type) {
     if (data && data[type] && data[type].length) {
-      const filteredItems = data[type].filter(item => !NewTabUtils.blockedLinks.isBlocked({"url": item.url}));
+      const filteredItems = data[type].filter(item => {
```
ScottDowne: I believe filter can happen here for spocs or recs. Do you need to do anything to handle that too?

content-src/lib/selectLayoutRender.js
ncloudioj: In `DiscoveryStreamBase.js`, it'll only send this ping if `spocsFill` is not empty.

lib/DiscoveryStreamFeed.jsm
ncloudioj: Good point. Yep, we should handle this.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -420,7 +439,13 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   scoreItems(items) {
     return items.map(item => this.scoreItem(item))
       // Remove spocs that are scored too low.
-      .filter(s => s.score >= s.min_score)
+      .filter(s => {
+        if (s.score >= s.min_score) {
+          return true;
+        }
+        this._addSpocsFill(s.id, "below_min_score", 0, 1);
```
ScottDowne: This could also happen for recs.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -491,6 +529,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           const isBelow = this.isBelowFrequencyCap(impressions, s);
           if (!isBelow) {
             caps.push(s);
+            this._addSpocsFill(s.id, "frequency_cap", 0, fullRecalc ? 1 : 0);
```
ScottDowne: I wonder if there's a way to do all the "add calls" inside the transform function? This way it would all be in one place. It's also spoc specific so you wouldn't need to worry about "if it spoc or rec" as many of these functions for filter and score are generic. But I dunno how we would engineer this in that way.

Otherwise, we just have a bunch of ifs and spoc fill stuff happening all over.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -40,6 +40,24 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+    // For SPOCS Fill telemetry
+    this._spocsFill = [];
+  }
+
+  _addSpocsFill(id, reason, displayed, fullRecalc) {
+    this._spocsFill.push({
+      id,
+      reason,
+      displayed,
+      full_recalc: fullRecalc,
```
ScottDowne: is full recalc needed? I'm wondering, if it's not a full recalc, do we need to send any pings? Like, a spoc impression won't actually cause a page to render something, especially render something different. It's just updating the main store which isn't used for when people see. When someone goes to view a page, then we grab the data, but at that point the impressions stuff is already updated.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -878,6 +917,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
                 spocs: newSpocs,
               },
             }));
+            this._handleSpocsFill();
```
ScottDowne: Similar to above, do we need to send this update at all?

lib/DiscoveryStreamFeed.jsm
ncloudioj: Agreed. Let me see if I can find a way to do so.

lib/DiscoveryStreamFeed.jsm
ScottDowne: Ah I think I see how you're seeing the recalc thing.

You send the ping in four cases.

1. When the server returns fresh spocs.
2. When an impression updates the spocs.
3. When the page renders fresh.
4. When the page re renders and is not fresh.

Does that add up to you?

lib/DiscoveryStreamFeed.jsm
ncloudioj: According to [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1535717#c5), I think this field is still useful for certain cases to provide more context of this ping. For example, frequency capping can be performed in either the recalculation case, or the non-recalc case.

lib/DiscoveryStreamFeed.jsm
ncloudioj: Based on the specs of this ping, yes. Analytics wanted to capture all the filtered SPOCS by frequency capping. Over here, a `SPOC_IMPRESSION` calls the `frequencyCapSpocs` which may get a SPOC filtered.

lib/DiscoveryStreamFeed.jsm
ScottDowne: If you don't see an easy way, don't sweat it too much. I think the solution you used still works, so if anything other than what you did is a lot of work, we can probably keep your existing solution.

ncloudioj:
> A bunch of comments, I dunno how many would require changes or not, but I think take a look at them, and if you think it's still good to go, and you can just help me understand a few things, we can set this as review requested again

Thanks for the feedback!

I've tweaked this patch as follows:

* Refactored all the filters in `DiscoveryStreamFeed.jsm` so that they share the same return value as `{data: result, filtered: a_list_of_filtered_items}`. I think it is clearer than calling `addSpocFill` all over the file.
* Added more comment & document to clarify how the `full_recalc` is defined.

test/unit/content-src/lib/selectLayoutRender.test.js
```diff
@@ -87,12 +122,17 @@ describe("selectLayoutRender", () => {
...
-    assert.lengthOf(result, 1);
-    assert.deepEqual(result[0].components[0].data.recommendations[0], "foo");
-    assert.deepEqual(result[0].components[0].data.recommendations[1], "bar");
+    assert.lengthOf(layoutRender, 1);
+    assert.deepEqual(layoutRender[0].components[0].data.recommendations[0], "foo");
+    assert.deepEqual(layoutRender[0].components[0].data.recommendations[1], "bar");
+
+    assert.deepEqual(spocsFill, [
+      {id: undefined, reason: "probability_selection", displayed: 0, full_recalc: 0},
+      {id: undefined, reason: "out_of_position", displayed: 0, full_recalc: 0},
```
punamdahiya: Shouldn't reason here for not displayed be "probability_selection" as we have positions: [{index: 0}, {index: 1}] available for both spocs?

test/unit/content-src/lib/selectLayoutRender.test.js
```diff
@@ -125,12 +170,17 @@ describe("selectLayoutRender", () => {
...
```
punamdahiya: Same concern as above https://github.com/mozilla/activity-stream/pull/4928/files#r279586949

test/unit/content-src/lib/selectLayoutRender.test.js
```diff
@@ -143,13 +193,18 @@ describe("selectLayoutRender", () => {
...
-    assert.deepEqual(result[0].components[0].data.recommendations[1], "fooSpoc");
-    assert.deepEqual(result[0].components[0].data.recommendations[2], "bar");
+    assert.lengthOf(layoutRender, 1);
+    assert.deepEqual(layoutRender[0].components[0].data.recommendations[0], "foo");
+    assert.deepEqual(layoutRender[0].components[0].data.recommendations[1], "fooSpoc");
+    assert.deepEqual(layoutRender[0].components[0].data.recommendations[2], "bar");
+
+    assert.deepEqual(spocsFill, [
+      {id: undefined, reason: "n/a", displayed: 1, full_recalc: 0},
+      {id: undefined, reason: "out_of_position", displayed: 0, full_recalc: 0},
```
punamdahiya: Same here https://github.com/mozilla/activity-stream/pull/4928/files#r279586949

test/unit/content-src/lib/selectLayoutRender.test.js
punamdahiya: After walking through code, help clarify out_of_position as reason for not displaying spocs. In selectLayoutRender there can be only one unchosenspoc  with reason as probability_selection as it will use all the available positions. This is hard to understand by looking at code, may be a line or two explaining what's happening [here](https://github.com/mozilla/activity-stream/pull/4928/files#diff-91e34b3436d635509e2017c193c45ec1R114) and adding tests with multiple out_of_positions inside  spocsFill as reason will help.  
Here's gist with few tests that I used  to debug flow https://gist.github.com/punamdahiya/8daf45627380412ea48d9e174d2409c1

ncloudioj: The way `maybeInjectSpocs` works is that it iterates over all the SPOCS positions, and for each position, it checks the probability with a generated random number so that a SPOC will be injected if it meets the selection criterion; Otherwise, it will skip the SPOC injection for this position and move forward to the next position with the same SPOC.

In this case, for both two positions, it didn't meet the probability selection criterion for the first SPOC with the reason `probability_selection`. And the second SPOC didn't even get a chance to be selected due to the lack of positions. Therefore, its reason was set as `out_of_position`.

The same applies to the following cases as well.

NB: the current probability configuration would allow every check to pass, i.e. no `probability_selection` should arise. But it's subject to future changes that could be made on the server side.

ncloudioj: Good point, will add some comments to clarify it.

content-src/lib/selectLayoutRender.js
```diff
@@ -23,8 +31,12 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
...
         }
 
-        if (spocs.data.spocs[spocIndex] && rickRoll <= spocsConfig.probability) {
-          recommendations.splice(position.index, 0, spocs.data.spocs[spocIndex++]);
+        if (rickRoll <= spocsConfig.probability) {
+          spocIndex++;
+          recommendations.splice(position.index, 0, spoc);
+          chosenSpocs.add(spoc);
+        } else {
+          unchosenSpocs.add(spoc);
```
punamdahiya: Is it safe to assume unchosenSpocs set will always have size 1 as spocIndex will only increase if a spoc is chosen meaning a single spoc will use all positions if its not chosen. It might help to rename it to unchosenSpoc.

ncloudioj: Indeed, technically there should be only one item with the reason `probability_selection`. However, I think we do need a `Set` to record all the state changes for each SPOC item, as it might fail the selection for one position, then it succeeds for the next one. The same can occur to every single SPOC. We store all those changes here, then we will reconcile the final state for each SPOC at the end of this function. I've added some comments there to make this less confusing. Hopefully, that will help.

punamdahiya: Thanks that helps! [Wrote test ](https://gist.github.com/punamdahiya/db78f5f81f8ee1a9c6975330f4394548#file-spocsfill-unchosenspocs)to capture this case :)

content-src/lib/selectLayoutRender.js
punamdahiya: That's correct, for fullRecalc case we need to ensure spoc data is loaded on first fetch or a layout reset, so we either keep this check or even better check for spocs.loaded similar to https://github.com/mozilla/activity-stream/blob/master/common/Reducers.jsm#L464

content-src/lib/selectLayoutRender.js
ncloudioj: Nice test, will add it in.

content-src/lib/selectLayoutRender.js
ncloudioj: Hmm, this `if` has nothing to do with `fullRecalc`, rather it is for the `spocs.data.spocs.slice(spocIndex)` down below. Also, if spocs is not enabled, shouldn't `spocs.data` be an empty object `{}`? Correct me if I was wrong here, I might've made a wrong assumption.

ncloudioj:
Added a few more review fixes:

* A unit test to test a more complex case for non-displayed SPOCS. Thanks! @punamdahiya
* A guard to prevent `DiscoveryStreamBase` from sending multiple SPOCS Fill pings on the state changes (e.g. when a user dismisses an item). Confirmed with the analytics team that this is the desired behavior.

content-src/lib/selectLayoutRender.js
punamdahiya: That's correct, looking at loadSpocs in DiscoveryStreamFeed.jsm, spocs.data is loaded from cache or endpoint only if spocs are enabled else data is empty

content-src/lib/selectLayoutRender.js
```diff
@@ -94,4 +106,28 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
       return {...component, data};
     }),
   })).filter(row => row.components.length);
+
+  // Generate the payload for the SPOCS Fill ping. Note that a SPOC could be rejected
+  // by the `probability_selection` first, then gets chosen for the next position. For
+  // all other SPOCS that never went through the probabilistic selection, its reason will
+  // be "out_of_position".
+  let spocsFill = [];
+  if (spocs.data.spocs) {
```
punamdahiya: @ncloudioj  spocs.data.spocs will throw error if spocs.data is undefined, can we add a check for that edge case.  thanks!

ncloudioj:
@punamdahiya Thanks for the review!

I've confirmed with Kirill that we do want to send a SPOCS Fill ping if a SPOC gets blocked by the user. Nice catch! 👍

However, I had to further complicate this patch by intercepting the `PLACES_LINK_BLOCKED` action to check if the dismissed item was a SPOC, if so send the ping out. A new action `DISCOVERY_STREAM_LINK_BLOCKED` was added to replace the old one in the reducer.

As for the type of the impression caps, it's not of the interest at the moment, we are find with it.

content-src/lib/selectLayoutRender.js
ncloudioj: Fixed, thank! Did you see this error happen? The init state of `spocs.data` was set as `{}` in the reducer, I wonder how/when would it be `undefined`?

punamdahiya: yes, came across this error if a faulty layout_endpoint is used in discoverystream.config, which fails to fetch endpoint on the first load

ncloudioj: Ah, I see. We might want to change that so the fetch failures won't affect the state structure. Otherwise, we'd have to check this all over the place.

punamdahiya: That will be the correct fix, it's [this call ](https://github.com/mozilla/activity-stream/blob/master/lib/DiscoveryStreamFeed.jsm#L370) that sets spocs to undefined

punamdahiya:
@ncloudioj Looks good! Verified blocked_by_user ping is sent when user dismisses story, and dismiss flow works as expected. Thanks!

ncloudioj:
Thank y'all for the review!

Merge it now as the data review has been granted on Bugzilla.

content-src/lib/selectLayoutRender.js
punamdahiya: Filed bug https://bugzilla.mozilla.org/show_bug.cgi?id=1548929

PR #4930, opened 2019-04-24
Bug 1546590 - Hardcoded layout results in only one new tab page rende…

ScottDowne:
…red with discovery stream content


Mardak:
We might be able to better catch this type of failure with a mochitest while not having any network (disabling by setting .endpoints to `data:` otherwise network monitor would crash the test for trying to make a remote connection). Because the hardcoded layout exists, the page will try to render top sites and section header, e.g.,
Screen Shot 2019-04-24 at 10 50 50 AM

So the mochitest could make sure at least top sites is rendered on subsequent page loads.

PR #4940, opened 2019-04-26
Bug 1546683 - Use fluent to localize trailhead first run

Mardak:
r?@k88hudson or @rlr Dynamically adds `<link rel=localization>` to activate fluent-dom. Existing onboarding happens to not do anything with the data-l10n-id as nothing is watching for those attributes. Do we need/want to remove the links on unmount? I suppose dismissing with Start Browsing could just hide the modal. Although I suppose we don't want to unnecessarily keep re-adding the links?
Screen Shot 2019-04-26 at 2 27 47 PM

FYI, the strings appear at `resource:///localization/en-US/trailhead.ftl` (note extra /)

### DIFF
--- a/content-src/asrouter/templates/Trailhead/Trailhead.jsx
+++ b/content-src/asrouter/templates/Trailhead/Trailhead.jsx
@@ -1,7 +1,25 @@
 import {OnboardingCard} from "../OnboardingMessage/OnboardingMessage";
 import React from "react";
 
+const FLUENT_FILES = [
+  "branding/brand.ftl",
+  "browser/branding/sync-brand.ftl",
+  // These are finalized strings exposed to localizers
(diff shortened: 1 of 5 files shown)


content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -24,9 +42,17 @@ export class Trailhead extends React.PureComponent {
           </div>
         </div>
         <div className="trailheadContent">
-          <h2>{content.ctaHeader}</h2>
-          <p>{content.ctaText}</p>
+          <h2 data-l10n-id="onboarding-membership-form-header" />
```
rlr: I can't believe that's all you have to do. fluent-dom 💥

PR #4942, opened 2019-04-29
Bug 1547746 - Render trailhead cards in a new portal / on the top of the page

rlr:


content-src/asrouter/asrouter-content.jsx
```diff
@@ -319,6 +321,7 @@ export class ASRouterUISurface extends React.PureComponent {
     const {message, bundle} = this.state;
     if (!message.id && !bundle.template) { return null; }
     const shouldRenderBelowSearch = TEMPLATES_BELOW_SEARCH.includes(message.template);
+    const shouldRenderInHeader = TEMPLATES_ABOVE_PAGE.includes(message.template);
```
Mardak: Do we want to support showing the below-search snippet for about:welcome? I suppose generally, would we need to support showing multiple templates potentially in above page, below search, and in footer?

Mardak: aaron says about:welcome has its own special "snippet" anyway, so only one at a time. So I believe this approach should be okay where if trailhead is active, then nothing else

content-src/asrouter/asrouter-content.jsx
```diff
@@ -333,7 +336,7 @@ export class ASRouterUISurface extends React.PureComponent {
           {this.renderOnboarding()}
           {this.renderSnippets()}
         </>,
-        this.portalContainer
+        shouldRenderInHeader ? this.headerPortal : this.footerPortal
```
k88hudson: Is this going to work with having to render 2 separate things at once though? i.e. onboarding on top, then cards?

k88hudson: Like don't we need  2 portals at once?

rlr: I was thinking we would render everything from `<Trailhead/>` and the CSS would take care of making the modal vs rendering inline on top of page. What you think?

rlr: So you would have `Trailhead` render something like:

```
<>
  <ModalOverlayWrapper>.....</ModalOverlayWrapper>
  <OnboardingCardsHere />
</>
```

PR #4943, opened 2019-04-29
Bug 1547744 - Add modal overlay to trailhead

k88hudson:
Had to refactor a little bit to get it working without the bottom bar;
Also removed the card rendering.

cc @Mardak

### DIFF
--- a/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
+++ b/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
@@ -1,5 +1,12 @@
 import React from "react";
 
+export const ModalOverlayWrapper = props => (<>
+  <div className={`modalOverlayOuter ${props.active ? "active" : ""}`} />
+  <div className={`modalOverlayInner ${props.active ? "active" : ""} ${props.innerClassName || ""}`}>
+    {props.children}
+  </div>
+</>);
+
 export class ModalOverlay extends React.PureComponent {
   componentWillMount() {
     this.setState({active: true});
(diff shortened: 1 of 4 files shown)


content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -1,5 +1,13 @@
 import React from "react";
 
+export const ModalOverlayWrapper = props => (<>
+  <div className={`modalOverlayOuter ${props.active ? "active" : ""}`}>
+    <div className={`modalOverlayInner ${props.active ? "active" : ""} ${props.innerClassName || ""}`}>
```
Mardak: Before these were adjacent instead of nested. The outer seems to be the faded background and inner is the actual content. Now because it's nested, it causes everything to be slightly transparent.

https://github.com/mozilla/activity-stream/blob/a0c1c2aead29d9a3ee0c5039abc22769efe7e884/content-src/asrouter/components/ModalOverlay/_ModalOverlay.scss#L7-L9

rlr: doh! yeah i guess it really wasn't Inner and Outer before

Mardak: nested:
Screen Shot 2019-04-29 at 2 30 56 PM


sibling:
Screen Shot 2019-04-29 at 2 30 37 PM

Mardak: I suppose it meant to be more "outside" / "inside" vs "outer" / "inner" 😛

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -20,41 +26,33 @@ export class Trailhead extends React.PureComponent {
     });
   }
 
-  render() {
-    const {props} = this;
-    const {content} = props.message;
+  componentDidMount() {
+    // We need to remove hide-main since we should show it underneath everything that has rendered
+    global.document.body.classList.remove("hide-main");
```
Mardak: Looks like we also needed to remove "welcome" which is used to hide overflow.

Mardak: We can probably fix this up when we add cards to top and figure out what to do with fixed search.

PR #4944, opened 2019-04-29
Bug 1547748 - add includeBundle stuff to trailhead

k88hudson:
(needs test fixes and tests)

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -690,18 +690,34 @@ class _ASRouter {
   }
 
   async _getBundledMessages(originalMessage, target, trigger, force = false) {
-    let result = [{content: originalMessage.content, id: originalMessage.id, order: originalMessage.order || 0}];
+    let result = [];
+    let bundleLength;
+    let bundleTemplate;
+    let originalId;
+
+    if (originalMessage.includeBundle) {
+      // The original message is not part of the bundle, so don't include it
+      bundleLength = originalMessage.includeBundle.length;
+      bundleTemplate =  originalMessage.includeBundle.template;
+    } else {
+      // The original message is part of the bundle
+      bundleLength = originalMessage.bundled;
+      bundleTemplate =  originalMessage.template;
+      originalId = originalMessage.id;
+      // Add in a copy of the first message
+      result.push({content: originalMessage.content, id: originalMessage.id, order: originalMessage.order || 0});
+    }
 
     // First, find all messages of same template. These are potential matching targeting candidates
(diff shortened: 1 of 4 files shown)


lib/ASRouter.jsm
```diff
@@ -690,18 +690,34 @@ class _ASRouter {
   }
 
   async _getBundledMessages(originalMessage, target, trigger, force = false) {
```
rlr: Unrelated to your changes, but `target` isn't used inside this function.

rlr:
This looks good. There are conflicts with the modal changes and the tests need some updating.

lib/ASRouter.jsm
k88hudson: ah weird

lib/ASRouter.jsm
```diff
@@ -772,7 +793,9 @@ class _ASRouter {
       const bundledMessages = await this._getBundledMessages(message, target, trigger, force);
       const action = bundledMessages ? {type: "SET_BUNDLED_MESSAGES", data: bundledMessages} : {type: "CLEAR_ALL"};
       target.sendAsyncMessage(OUTGOING_MESSAGE_NAME, action);
-
+    } else if (message.includeBundle) {
+      const bundledMessages = await this._getBundledMessages(message, target, message.includeBundle.trigger, force);
+      target.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {type: "SET_MESSAGE", data: {...message, bundle: bundledMessages && bundledMessages.bundle}});
```
Mardak: @k88hudson There's a try/catch around a different `target.sendAsyncMessage` a few lines earlier. Is that to avoid async exceptions such as…

https://treeherder.mozilla.org/logviewer.html#/jobs?job_id=243831430&repo=try&lineNumber=12245
```
INFO - TEST-UNEXPECTED-FAIL | browser/components/sessionstore/test/browser_newtab_userTypedValue.js | A promise chain failed to handle a rejection: Message port has been destroyed - stack: sendAsyncMessage@resource://gre/modules/remotepagemanager/MessagePort.jsm:270:13
_sendMessageToTarget@resource://activity-stream/lib/ASRouter.jsm:795:14
async*sendNextMessage@resource://activity-stream/lib/ASRouter.jsm:924:16
```

The test seems to be testing on a bunch of pages including about:welcome and the `await getBundledMessages` might be delaying it enough that the tab has closed destroying the message port. (It's actually the ` else if (message.bundled) {` block's sendAsyncMessage that is throwing, but generally this whole method might want to try/catch ??)

PR #4946, opened 2019-04-29
(nobug) - Partial quick visual changes for trailhead first run

Mardak:
r?@rlr Lots of temporary stuff that I made up (colors, sizing, etc.) but basic layout is there. I also ended up copying over the strings from mozilla-central side and concatenating them together for now.

Screen Shot 2019-04-29 at 4 26 55 PM

### DIFF
--- a/data/trailhead.ftl
+++ b/data/trailhead.ftl
@@ -1,66 +1,134 @@
-onboarding-welcome-body = You’ve got the browser. Let’s explore the rest of Firefox.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-onboarding-membership-form-header = Set Up Your Firefox Membership
-onboarding-membership-form-email =
-    .placeholder = Your email …
-onboarding-membership-form-legal-links = By proceeding, you agree to the <a data-l10n-name="terms">Terms of Service</a> and <a data-l10n-name="privacy">Privacy Notice</a>.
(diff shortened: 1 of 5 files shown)


data/trailhead.ftl
```diff
@@ -1,66 +1,134 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-onboarding-subhead-productive-title = Be productive
-onboarding-subhead-productive-text = Get your everyday done with the browser, features, and apps.
+## UI strings for the simplified onboarding modal
 
-onboarding-subhead-smarter-title = Be a smarter browser
-onboarding-subhead-smarter-text = Get doable advice from world-class experts.
+onboarding-button-label-learn-more = Learn More
```
Mardak: I was thinking of removing the ones that already exist, as modifying the trailhead.ftl won't do anything as onboarding.ftl versions will be preferred... but then nahh I'll just plain concat the two brandings.ftl and onboarding.ftl files together 😉

lib/OnboardingMessageProvider.jsm
```diff
@@ -51,11 +51,11 @@ function getPlaceholderTrailheadCard(id) {
     id: "ONBOARDING_2",
     template: "onboarding",
     content: {
-      title: `onboarding-card-${id}-title`,
-      text: `onboarding-card-${id}-text`,
+      title: `onboarding-${id}-title`,
```
Mardak: The existing onboarding.ftl file had "card" -title/text without a "-card-" so the m-c patch I have follows that pattern.

PR #4947, opened 2019-04-29
(nobug) Trailhead basic card styles

k88hudson:
Still needs work on responsive sizing, but here's a start


content-src/asrouter/templates/Trailhead/Trailhead.scss
```diff
@@ -83,6 +83,24 @@
   }
 }
 
+$trailhead-cards-bg: #F0EFF4;
```
k88hudson: I just colour picked these off the screenshot so they're not really real

PR #4948, opened 2019-04-29
Bug 1547061 - Port FXA signup form to trailhead first run

rlr:


content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -15,26 +16,83 @@ const FLUENT_FILES = [
...
+      emailInput: "",
+      isModalOpen: true,
+      flowId: "",
+      flowBeginTime: 0,
+    };
+    this.didFetch = false;
   }
 
-  componentWillMount() {
+  async componentWillMount() {
```
rlr: is it ok to make this async? Is the rendering going to be delayed until the fetch happens below?

rlr: I originally had it in `componentDidMount` but you aren't supposed to call `setState` from there.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -70,15 +128,25 @@ export class Trailhead extends React.PureComponent {
...
-              href="https://accounts.firefox.com/legal/privacy" />
-          </p>
-          <button data-l10n-id="onboarding-join-form-continue" />
+          <form method="get" action={this.props.fxaEndpoint} target="_blank" rel="noopener noreferrer" onSubmit={this.onSubmit}>
+            <input name="service" type="hidden" value="sync" />
+            <input name="action" type="hidden" value="email" />
+            <input name="context" type="hidden" value="fx_desktop_v3" />
+            <input name="entrypoint" type="hidden" value="activity-stream-firstrun" />
+            <input name="utm_source" type="hidden" value="activity-stream" />
+            <input name="utm_campaign" type="hidden" value="firstrun" />
```
rlr: I wonder if the FXA team wants new values for any of these params.

rlr: Per the copy doc, I added utm_term=trailhead so the FxA folks can filter the data easily.

test/unit/asrouter/templates/Trailhead.test.jsx
```diff
@@ -0,0 +1,45 @@
+import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
+import {mountWithIntl} from "test/unit/utils";
```
k88hudson: try just using `mount`

PR #4949, opened 2019-04-30
(nobug) - Switch existing onboarding to use fluent-dom

Mardak:
r?@k88hudson This removes the `L10N.formatMessages` step for onboarding and duplicates the fluent loading for `OnboardingMessage`. Looking at the PR now, I suppose I could have left the messages untouched and grabbed string_id in the OnboardingMessage.jsx instead… Also, see related #4950.

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -227,17 +227,6 @@ const OnboardingMessageProvider = {
         }
       }
 
-      if (msg.template === "onboarding") {
-        const [primary_button_string, title_string, text_string] = await L10N.formatMessages([
-          {id: msg.content.primary_button.label.string_id},
-          {id: msg.content.title.string_id},
-          {id: msg.content.text.string_id, args: msg.content.text.args},
-        ]);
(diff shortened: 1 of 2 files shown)


Mardak:
I wonder if we want to keep the ability to pass in `{string_id}` vs `"string"` where the former could allow for already-translated (e.g., from remote settings) and fluent ids.

lib/OnboardingMessageProvider.jsm
```diff
@@ -53,11 +53,11 @@ const ONBOARDING_MESSAGES = async () => ([
     bundled: 3,
     order: 2,
     content: {
-      title: {string_id: "onboarding-private-browsing-title"},
```
k88hudson: Let's leave those as `string_id` to match other cards

Mardak: Reverted these changes

Mardak:
Fixed up test in https://github.com/mozilla/activity-stream/commit/3a2fb3b13aa49136de6a1d7fd676b5592138f81f

PR #4950, opened 2019-04-30
(nobug) - Update finalized strings for layout and cards

Mardak:
r?@k88hudson This updates to the finalized strings and adjusts the layout. This also adds the ONBOARDING_MESSAGES with at least the string portion. This could land without #4949 as it also removes the `L10N.formatMessages` step but it would break old onboarding.

Screen Shot 2019-04-30 at 12 15 44 AM

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -151,6 +151,215 @@ const ONBOARDING_MESSAGES = async () => ([
     content: {},
   },
   {
+    id: "TRAILHEAD_CARD_1",
+    template: "onboarding",
+    bundled: 3,
+    content: {
+      title: {string_id: "onboarding-tracking-protection-title"},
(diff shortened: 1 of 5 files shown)


lib/OnboardingMessageProvider.jsm
```diff
@@ -150,6 +150,215 @@ const ONBOARDING_MESSAGES = async () => ([
     includeBundle: {length: 3, template: "onboarding", trigger: {id: "showOnboarding"}},
     content: {},
   },
+  {
+    id: "TRAILHEAD_CARD_1",
+    template: "onboarding",
+    bundled: 3,
+    content: {
+      title: "onboarding-tracking-protection-title",
```
k88hudson: Let's use `string_id` for these as well

Mardak: Added `string_id`s

PR #4952, opened 2019-04-30
Bug 1546422 - FXA-related snippets should not snow up if Firefox Acco…

piatra:
…unts are disabled


content-src/asrouter/docs/targeting-attributes.md
```diff
@@ -434,6 +435,16 @@ Does the user use Firefox sync?
 declare const usesFirefoxSync: boolean;
 ```
 
+### `isFxAEnabled`
+
+Does the user have Firefox sync enabled? The service could potentially be turned off.
```
k88hudson: Can you mention this is something that is common for enterprise (disabling)

k88hudson:
Can you adjust this so this persistently bypasses targeting rather than relying on the bookmark panel being open in the current page

PR #4961, opened 2019-05-01
Bug 1548267 - Overlay accessibility

k88hudson:


content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -1,28 +1,45 @@
...
-    document.body.classList.remove("modal-open");
-    this.setState({active: false});
+    this.props.document.removeEventListener("keydown", this.onKeyDown);
+    this.props.document.body.classList.remove("modal-open");
+  }
+
+  render() {
+    const {props} = this;
+    return (<React.Fragment>
+      <div className={`modalOverlayOuter active`} onClick={props.onClose} />
```
Mardak: Setting `role="presentation"` on this `<div onClick…>` makes these a11y errors go away:
```
  28:7  error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
  28:7  error  Static HTML elements with event handlers require a role                                         jsx-a11y/no-static-element-interactions
```
As I guess it no longer considers it "visible"

PR #4583, opened 2018-11-29
Bug 1511152 - adding lint:jsx-a11y script to be able to run jsx a11y …

yzen:
…rules independently.

This is a proposal of how to not be too disruptive but also flag a11y issue found with this tool.

PR #4962, opened 2019-05-01
Bug 1548268 - Turn on jsx-a11y for the trailhead related files

Mardak:
r?@yzen and @k88hudson  This somewhat supersedes #4583 but only enables for the files used by trailhead for now. Looks like tabindex=1 wasn't necessary for the old or new about:welcome.


Mardak:
@punamdahiya The plan is we'll merge trailhead branch to master soon, and then we could expand which files get linted (and fix up existing errors).

Mardak:
I'll fix up more accessibility stuff as part of this same bug

PR #4971, opened 2019-05-02
Bug 1548353 - Timeout DS fetches after 45 seconds.

ScottDowne:
To test.

1. You need something to delay a server response longer than 45 seconds, I made this https://github.com/ScottDowne/fetch-delay after another remote service failed me too many times. You can use something else but for the purposes of these test steps I'll be using the above.
2. After cloning fetch delay, run npm install and node start.js
4. Set this pref `browser.newtabpage.activity-stream.discoverystream.endpoints` to `https,http`
3. In the browser, set this pref `browser.newtabpage.activity-stream.discoverystream.config` to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":false,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"http://localhost:3000/?delay=50&url=https%3A%2F%2Fgetpocket.cdn.mozilla.net%2Fv3%2Fnewtab%2Flayout%3Fversion%3D1%26consumer_key%3D40249-e88c401e1b1f2242d9e441c4%26layout_variant%3Dbasic" }`
5. load a new tab, you should just see search for now.
5. After 45 seconds, the fetch should time out, and it'll default to the local hard coded layout. (Has 7 feeds, starting with recommended by pocket, then Health & Fitness 💪, then tech, etc)
6. now set this pref `browser.newtabpage.activity-stream.discoverystream.config` to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":false,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"http://localhost:3000/?delay=10&url=https%3A%2F%2Fgetpocket.cdn.mozilla.net%2Fv3%2Fnewtab%2Flayout%3Fversion%3D1%26consumer_key%3D40249-e88c401e1b1f2242d9e441c4%26layout_variant%3Dbasic" }`
7. After 10 seconds it should succeed (not timeout) and load basic, which only has 1 feed.

I considered a test to test the abort controller, but after a quick start, I stopped because it felt like I was testing abort controller and not DS.

The only real change here other than the timeout, is if it times out, or fails otherwise, we instead load the hardcoded layout.

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -146,7 +146,8 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       const controller = new AbortController();
       const {signal} = controller;
       const fetchPromise = fetch(endpoint, {credentials: "omit", signal});
-      const timeoutId = setTimeout(controller.abort, FETCH_TIMEOUT);
+      // istanbul ignore next
+      const timeoutId = setTimeout(() => { controller.abort(); }, FETCH_TIMEOUT);
 
       const response = await fetchPromise;
       if (!response.ok) {
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -5,6 +5,7 @@
 
 const {XPCOMUtils} = ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
 const {NewTabUtils} = ChromeUtils.import("resource://gre/modules/NewTabUtils.jsm");
+const {setTimeout, clearTimeout} = ChromeUtils.import("resource://gre/modules/Timer.jsm");
 const {Services} = ChromeUtils.import("resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
 ChromeUtils.defineModuleGetter(this, "perfService", "resource://activity-stream/common/PerfService.jsm");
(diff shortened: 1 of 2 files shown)


test/unit/lib/DiscoveryStreamFeed.test.js
```diff
@@ -145,7 +145,7 @@ describe("DiscoveryStreamFeed", () => {
 
       await feed.fetchFromEndpoint("https://getpocket.cdn.mozilla.net/dummy?consumer_key=$apiKey");
 
-      assert.calledWith(fetchStub, "https://getpocket.cdn.mozilla.net/dummy?consumer_key=replaced", {credentials: "omit"});
+      assert.calledWith(fetchStub, "https://getpocket.cdn.mozilla.net/dummy?consumer_key=replaced");
```
Mardak: If you don't want an exact comparison, you should be able to use `calledWithMatch` to still check for an object that at least has the "credentials" key with "omit" value

ScottDowne: Thanks!

punamdahiya:
TaskCluster is failing with  unreferenced files in test "browser/base/content/test/static/browser_all_files_referenced.js" which doesn't look related to patch.

punamdahiya:
Testing with steps above "http://localhost:3000/?delay=50&url=https%3A%2F%2Fgetpocket.cdn.mozilla.net%2Fv3%2Fnewtab%2Flayout%3Fversion%3D1%26consumer_key%3D40249-e88c401e1b1f2242d9e441c4%26layout_variant%3Dbasic" works fine showing expected 50 sec delay.

In discovery stream
a) After 45 seconds seeing error in browser console TypeError: 'abort' called on an object that does not implement interface AbortController. and no hard coded layout

b) After 50 seconds, new tab render with basic layout

lib/DiscoveryStreamFeed.jsm
```diff
@@ -141,10 +143,17 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
         throw new Error(`Not one of allowed prefixes (${allowed})`);
       }
 
-      const response = await fetch(endpoint, {credentials: "omit"});
+      const controller = new AbortController();
+      const {signal} = controller;
+      const fetchPromise = fetch(endpoint, {credentials: "omit", signal});
+      // istanbul ignore next
```
punamdahiya: Without istanbul ignore code coverage complains test coverage for this line https://github.com/mozilla/activity-stream/pull/4971/files#diff-8b1b03c4d5dec5c8e4f7de50e05c0c4fR185

Do you know why istanbul ignore next help fix that, will be good to add as comment here

punamdahiya: It seems code coverage is wrongly pointing to isExpired, added test covers that edge case

PR #4972, opened 2019-05-02
Bug 1535717 - Update telemetry endpoint for Spocs Fill

ncloudioj:
This is to match the [name change](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/327/files) of the telemetry endpoint for SPOCS Fill.

### DIFF
--- a/lib/TelemetryFeed.jsm
+++ b/lib/TelemetryFeed.jsm
@@ -844,7 +844,7 @@ this.TelemetryFeed = class TelemetryFeed {
    */
   handleDiscoveryStreamSpocsFill(data) {
     const payload = this.createSpocsFillPing(data);
-    this.sendStructuredIngestionEvent(payload, "spocs-fills", "1");
+    this.sendStructuredIngestionEvent(payload, "spoc-fills", "1");
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/TelemetryFeed.jsm
```diff
@@ -844,7 +844,7 @@ this.TelemetryFeed = class TelemetryFeed {
    */
   handleDiscoveryStreamSpocsFill(data) {
     const payload = this.createSpocsFillPing(data);
-    this.sendStructuredIngestionEvent(payload, "spocs-fills", "1");
+    this.sendStructuredIngestionEvent(payload, "spoc-fills", "1");
```
punamdahiya: Should we change name here as well https://github.com/mozilla/activity-stream/blob/master/common/Actions.jsm#L299

ncloudioj: Aha, good catch. Yes, let me clean it up a bit. Naming stuff is really hard ;)

punamdahiya:
Looking at the code we are using Spocs Fill and Spoc Fills interchangeably. Sorry for not catching this earlier, but it will help to clarify if there is a pattern where we are using these respectively. We can also consider making it uniform by choosing one name across, I see Spoc Fills use in fewer files. 

Either way this is a nit and more for code readability and ok landing as is. Thanks

ncloudioj:
Thanks! @punamdahiya 

Agreed. This mess was, unfortunately, caused by the my earlier error in the [schema definition](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/327/files). Since we use SPOC and SPOCS interchangeably, hopefully folks would tolerate me with this one ;)

PR #4935, opened 2019-04-25
Bug 1546898 - Bookmark Panel QA test provider

piatra:
I've changed the local providers to be lazily loaded with `defineLazyGlobalGetters`. And they are added only if `devtoolsEnabled`.
I assumed we'll be adding more and each will have their own file.
I also found some more code related to snippets that should have been removed. `allowLegacySnippets` in the `Reducer` state and in `ASRouterPreferences`.

PR #4979, opened 2019-05-03
Bug1548520 - Add user telemetry for click events

piatra:
Adds user telemetry for impression, click and dimiss. Requires #4935 to land first.

### DIFF
--- a/lib/BookmarkPanelHub.jsm
+++ b/lib/BookmarkPanelHub.jsm
@@ -16,22 +16,26 @@ class _BookmarkPanelHub {
     this._trigger = {id: "bookmark-panel"};
     this._handleMessageRequest = null;
     this._addImpression = null;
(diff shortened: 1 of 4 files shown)


lib/BookmarkPanelHub.jsm
```diff
@@ -181,6 +189,17 @@ class _BookmarkPanelHub {
   sendImpression() {
     this._addImpression(this._response);
   }
+
+  sendUserEventTelemetry(event) {
+    this._sendTelemetry({message_id: this._response.id, event});
```
ncloudioj: According to CFR data collection policy, we will use `bucket_id` instead of `message_id` in the release channel. So, like other CFR metrics, let's add a `bucket_id` field in the payload. I think you can just set it as the same value as message_id in this case.

PR #4982, opened 2019-05-03
Bug 1548929 - Fix broken state after spoc endpoint fetch failure

punamdahiya:
STR https://bugzilla.mozilla.org/show_bug.cgi?id=1548929#c0


punamdahiya:
Follow up of  https://github.com/mozilla/activity-stream/pull/4928#discussion_r280122719 . Now that we have 45sec fetch time out on our endpoints it increase probability of encountering this bug.

@ncloudioj @gvn @ScottDowne whoever has bandwidth, please review. Thanks!

punamdahiya:
> @punamdahiya Thanks for fixing it up!
> 
> Do you think we can clean up all those lengthy checks `if (spocs && spocs.data && ...)` now with this enforcement?

Done. Thanks!

PR #4983, opened 2019-05-03
Bug 1545110 - Unexpected search shortcut in Topsites

rlr:


rlr:
hmm it's not working quite as i hoped. no top sites being shown!

rlr:
ah I can't rename `SEARCH_SHORTCUTS_EXPERIMENT` because it is being referenced from m-c.

PR #4985, opened 2019-05-06
Bug 1548503 - Update card button actions with finalized urls or actions

rlr:
OK, wasn't as simple as I first thought....

### DIFF
--- a/content-src/asrouter/templates/StartupOverlay/StartupOverlay.jsx
+++ b/content-src/asrouter/templates/StartupOverlay/StartupOverlay.jsx
@@ -13,6 +13,8 @@ export class _StartupOverlay extends React.PureComponent {
     this.removeOverlay = this.removeOverlay.bind(this);
     this.onInputInvalid = this.onInputInvalid.bind(this);
 
+    this.utmParams = "utm_source=activity-stream&utm_campaign=firstrun&utm_medium=referral&utm_term=trailhead-control";
+
     this.state = {
       emailInput: "",
       overlayRemoved: false,
@@ -26,8 +28,8 @@ export class _StartupOverlay extends React.PureComponent {
     if (this.props.fxa_endpoint && !this.didFetch) {
       try {
         this.didFetch = true;
-        const fxaParams = "entrypoint=activity-stream-firstrun&utm_source=activity-stream&utm_campaign=firstrun&form_type=email";
-        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow?${fxaParams}`, {credentials: "omit"});
+        const fxaParams = "entrypoint=activity-stream-firstrun&form_type=email";
+        const response = await fetch(`${this.props.fxa_endpoint}/metrics-flow?${fxaParams}&${this.utmParams}`, {credentials: "omit"});
         if (response.status === 200) {
           const {flowId, flowBeginTime} = await response.json();
           this.setState({flowId, flowBeginTime});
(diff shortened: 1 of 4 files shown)


content-src/asrouter/templates/StartupOverlay/StartupOverlay.jsx
```diff
@@ -128,6 +128,7 @@ export class _StartupOverlay extends React.PureComponent {
                 <input name="entrypoint" type="hidden" value="activity-stream-firstrun" />
                 <input name="utm_source" type="hidden" value="activity-stream" />
                 <input name="utm_campaign" type="hidden" value="firstrun" />
```
rlr: TODO: this is missing utm_medium

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -131,9 +132,33 @@ export class _Trailhead extends React.PureComponent {
     return str.value;
   }
 
+  onCardAction(action) {
+    let actionUpdates = {};
+
+    if (action.type === "OPEN_URL") {
+      let url = new URL(action.data.args);
+      let params = new URLSearchParams(url.search.slice(1));
```
Mardak: Do we need to create a new `Params` vs just directly using `url.searchParams` ? If we reuse, appending will result in `url` getting updated too.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -131,9 +132,33 @@ export class _Trailhead extends React.PureComponent {
...
+      params.append("utm_campaign", "firstrun");
+      params.append("utm_medium", "referral");
+      params.append("utm_term", this.props.message.utm_term);
+
+      if (action.addFlowParams) {
+        params.append("flow_id", this.state.flowId);
+        params.append("flow_begin_time", this.state.flowBeginTime);
+      }
+
+      url = `${url.href.split("?")[0]}?${params.toString()}`;
```
Mardak: In general, we shouldn't split or build URLs manually. As noted above, if we use `url.searchParams`, we shouldn't need this line (and the `{args: url}` below should just work as part of `URL.toString()`

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -131,9 +132,33 @@ export class _Trailhead extends React.PureComponent {
...
+    if (action.type === "OPEN_URL") {
+      let url = new URL(action.data.args);
+      let params = new URLSearchParams(url.search.slice(1));
+      params.append("utm_source", "activity-stream");
+      params.append("utm_campaign", "firstrun");
+      params.append("utm_medium", "referral");
+      params.append("utm_term", this.props.message.utm_term);
+
+      if (action.addFlowParams) {
+        params.append("flow_id", this.state.flowId);
```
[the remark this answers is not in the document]
rlr: yeah, that probably needs updating. nice catch.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
rlr: Doh. MDN says `searchParams` is read only: https://developer.mozilla.org/en-US/docs/Web/API/URL. But a quick test in the console confirms that works.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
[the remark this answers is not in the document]
rlr: I updated this but lets wait for @clouserw to confirm this change and whether we need to update the old flow too.

rlr:
Just realized this is against master and so is the a11y one. We'll have to remember to keep the trailhead branch of to date somehow?

rlr:
oh neat, you can change the base branch from the edit pull request form. fixord

content-src/asrouter/templates/Trailhead/Trailhead.jsx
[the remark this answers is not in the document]
irrationalagent: Yes, the all the utm_parameters passed in the `metrics-flow` request should match those attached to the FxA URL. If by "old flow" you mean what will be the control, then yes its best to update that as well so that it includes the same parameters.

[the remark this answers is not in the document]
clouserw: Thanks @irrationalagent !

rlr:
@Mardak ok, I think this is ready for another pass. I added the utm_* to the learn, privacy and terms links too.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -131,9 +133,39 @@ export class _Trailhead extends React.PureComponent {
     return str.value;
   }
 
+  addUtmParams(url, isCard = false) {
```
Mardak: Mmm.. A quick comment/jsdoc about how this modifies a URL passed in or returns a new URL would help.

Mardak:
Oh, and all the other actions are indeed final? Or maybe except for the send tabs blog post. File a followup bug for "final final final actions" ? 😉

PR #4986, opened 2019-05-07
Bug 1522166 - Switch between light and dark icon in snippet messages

piatra:


lib/SnippetsTestMessageProvider.jsm
```diff
@@ -7,14 +7,38 @@ const TEST_ICON = "chrome://branding/content/icon64.png";
...
+    "weight": 100,
+    "id": "SIMPLE_TEST_1_LightDark",
+    "template": "simple_snippet",
+    "campaign": "test_campaign_blocking",
+    "content": {
+      "icon": "chrome://branding/content/icon64.png",
+      "icon_dark_theme": "chrome://branding/content/icon64.png",
+      "title": "Firefox Account!",
+      "title_icon": "chrome://branding/content/icon16.png",
+      "title_icon_dark_theme": "chrome://branding/content/icon16.png",
```
k88hudson: This is kind of hard to test if the dark theme icon isn't actually different, can you maybe just use another image that's available?

Also I think it would be good for at least one of the test messages to *not* have a dark theme icon defined so that case can be tested too.

PR #4987, opened 2019-05-07
Bug 1548268 - Improve accessibility of trailhead flow

Mardak:
r?@rlr Needed to do some document passing to get testing happy. You can test accessibility from devtools to see what a screenreader would "see" so changes here make the stuff behind modal "hidden" and correctly identifies the modal as a "dialog" with header. PR also makes tabbing cycle through just the dialog instead of to cards/topsites/etc behind. I'll rebase on your changes in #4985 as there'll be test merge conflicts.

before (shows cards, search, etc):
Screen Shot 2019-05-07 at 2 44 22 AM

after (dialog only):
Screen Shot 2019-05-07 at 2 45 37 AM

### DIFF
--- a/bin/render-activity-stream-html.js
+++ b/bin/render-activity-stream-html.js
@@ -100,9 +100,9 @@ function templateHTML(options, html) {
     <link rel="stylesheet" href="${options.baseUrl}css/activity-stream.css" />
   </head>
   <body class="activity-stream">
-    <div id="header-asrouter-container"></div>
+    <div id="header-asrouter-container" role="presentation"></div>
     <div id="root">${isPrerendered ? html : "<!-- Regular React Rendering -->"}</div>
-    <div id="footer-asrouter-container"></div>${options.noscripts ? "" : scriptRender}
+    <div id="footer-asrouter-container" role="presentation"></div>${options.noscripts ? "" : scriptRender}
   </body>
 </html>
 `;
--- a/content-src/asrouter/asrouter-content.jsx
+++ b/content-src/asrouter/asrouter-content.jsx
@@ -296,6 +296,7 @@ export class ASRouterUISurface extends React.PureComponent {
     const {message} = this.state;
     if (message.template === "trailhead") {
       return (<Trailhead
+        document={this.props.document}
         message={message}
         onAction={ASRouterUtils.executeAction}
         onDoneButton={this.dismissBundle(this.state.bundle.bundle)}
--- a/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
+++ b/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
@@ -26,7 +26,10 @@ export class ModalOverlayWrapper extends React.PureComponent {
     const {props} = this;
     return (<React.Fragment>
       <div className="modalOverlayOuter active" onClick={props.onClose} role="presentation" />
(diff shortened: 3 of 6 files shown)


yzen:
@Mardak looks great! I was wondering if it's also possible to hide the sibling "section" element to the dialog (e.g. it's probably a container of what's currently being hidden). Note this is not essential, but would make the doc structure look perfect from the a11y standpoint.

rlr:
This is neat. Taskcluster log is showing a bunch of weird warnings and errors but probably unrelated?

yzen:
@Mardak , Also I believe you make the dialog itself focusable, where best practice is to set focus on the first focusable element inside it when it's opened. The same goes for cycling the focus state, if you are moving it to the beginning, we should focus on the first focusable element within the dialog again. This will also make setting tabindex=-1 unnecessary on the dialog itself.

piatra:
Looks like a very fresh patch that just landed https://hg.mozilla.org/mozilla-central/file/tip/browser/components/sessionstore/TabStateFlusher.jsm

Mardak:
@yzen The two `section ""` immediately inside `document "New Tab"` are the "portals" that could potentially show snippets or trailhead content. Is the correct behavior to set those to `role=presentation` ?

And for focusing the first focusable element, I was trying to be generic in case we add a focusable element earlier than the "learn more" link and forget to update the "first" -- is there a standard way to find the first focusable instead of hardcoding a specific selector to the current first, e.g., `querySelector(".trailheadLearn")`

yzen:
> @yzen The two `section ""` immediately inside `document "New Tab"` are the "portals" that could potentially show snippets or trailhead content. Is the correct behavior to set those to `role=presentation` ?

Correct. If it's a container with no semantics, presentation will do the trick.

> 
> And for focusing the first focusable element, I was trying to be generic in case we add a focusable element earlier than the "learn more" link and forget to update the "first" -- is there a standard way to find the first focusable instead of hardcoding a specific selector to the current first, e.g., `querySelector(".trailheadLearn")`

Yeah we do something like that in DevTools: you can use a selector like this: https://searchfox.org/mozilla-central/source/devtools/client/shared/focus.js that is used to find all focusable elements within a given container. You'd just need the first one (and maybe last one as another way of checking if you are exiting the dialog on blur).

yzen:
You can see that selector used here for example:

https://searchfox.org/mozilla-central/source/devtools/client/shared/components/VirtualizedTree.js#13,929-931

Mardak:
Oh neat. The file is indeed visible at `resource://devtools/client/shared/focus.js` although our content page is statically bundled... would be nice to just require/requse it but maybe I'll just simplify it to look for the first a, button, input in our case… hmmm……

Mardak:
> you can use a selector like this: https://searchfox.org/mozilla-central/source/devtools/client/shared/focus.js 

pbro noted that previously, devtools tried to not have "external" usage modules as devtools could be disabled entirely / not even built with firefox. For cleaner refactoring, the there could be a firefox module that both devtools and activity stream can reference, but for now especially with our webpack usage, we'll just copy/paste.

yzen:
Yeah I was too suggesting using the same selector by copying. It is already copied in debugger.html project.

Mardak:
Rebased and updated to focus email input and set role=presentation on portals.

Screen Shot 2019-05-07 at 1 55 26 PM

PR #4988, opened 2019-05-07
Slight spoc adjustment and placeholder concept for cards that are sti…

ScottDowne:
This simplifies spoc injection a bit, and stubs out cards with empty data for cards, so it can render placeholders while it waits for a feed.


ScottDowne:
Closing for https://github.com/mozilla/activity-stream/pull/5024

PR #4990, opened 2019-05-07
Bug 1549716 - The email input field border wrongly remains red after clearing an invalid email

rlr:
I just realized we dont have any `:focus` styles on the email field. I'm checking with Aaron to see if he wants to add that.

### DIFF
--- a/content-src/asrouter/templates/Trailhead/_Trailhead.scss
+++ b/content-src/asrouter/templates/Trailhead/_Trailhead.scss
@@ -183,18 +183,28 @@
 
     button,
     input {
-      border: 0;
       width: 100%;
     }
 
     input {
       background-color: $white;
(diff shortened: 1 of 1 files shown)


content-src/asrouter/templates/Trailhead/_Trailhead.scss
```diff
@@ -183,18 +183,38 @@
...
     input {
       background-color: $white;
+      border: 1px solid $grey-50;
+      box-shadow: none;
       color: $grey-70;
       font-size: 15px;
+      transition: border-color 150ms, box-shadow 150ms;
+
+      &:hover {
+        border-color: $grey-90;
```
Mardak: Wow this hover is quite subtle… I had to zoom in all the way 😛

PR #4991, opened 2019-05-07
Bug 1549714 - The new onboarding modal is not scrollable if the browser window is horizontally resized to less than 25%

rlr:
I think it's a good sign that these are the kinds of bugs coming in 😃

### DIFF
--- a/content-src/asrouter/templates/Trailhead/_Trailhead.scss
+++ b/content-src/asrouter/templates/Trailhead/_Trailhead.scss
@@ -10,11 +10,6 @@
   height: auto;
   top: 100px;
 
-  @media (max-height: 700px) {
-    position: absolute;
-    top: 20px;
-  }
-
   a {
     color: $white;
     text-decoration: underline;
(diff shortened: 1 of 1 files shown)


content-src/asrouter/templates/Trailhead/_Trailhead.scss
```diff
@@ -401,9 +396,14 @@
   }
 }
 
-// If the window is too short, we need to allow scrolling so user can get to Start Browsing button.
-@media (max-height: 700px) {
+// If the window is too short or narrow, we need to allow scrolling so user can get to Start Browsing button.
+@media (max-height: 700px), (max-width: 765px) {
```
Mardak: Where did we get this 765px?

rlr: I made my height just above 700px (so it doesn't trigger the first case) and reduced the width until the button goes off the bottom. I set it to a little over that so that hopefully German will fit too.

Mardak: Set `intl.l10n.pseudo` to "accented" to test longer strings !

Mardak:
Hmm... looks like accented results in a cut off button no matter the width when height is just over 700px…

Although that might mean we need a different cutoff for height instead of width for that case.

Screen Shot 2019-05-07 at 12 28 48 PM

content-src/asrouter/templates/Trailhead/_Trailhead.scss
rlr: oh wow... both these values need increasing !

PR #4992, opened 2019-05-07
Bug 1544578 - TopStoriesFeed early return for discovery stream

punamdahiya:


Mardak:
Adding some notes from what we discussed earlier:
- we do want to "fast path" the default setting behavior via constructor
- check PREFS_INITIAL_VALUES to see if the user pref differed and lazily start top stories if discovery was disabled (and leave top stories enabled if user actually wants discovery)
- also check PREF_CHANGED similar to above

Caveats:
- constructor has no access to state
- PrefsFeed dispatches PREFS_INITIAL_VALUES on its INIT.. which.. alphabetically happens before TopStoriesFeed ?? so by the time TopStoriesFeed INITs, it might be able to just check store's pref state as it might never get the action

punamdahiya:
> Adding some notes from what we discussed earlier:
> 
>     * we do want to "fast path" the default setting behavior via constructor
> 
>     * check PREFS_INITIAL_VALUES to see if the user pref differed and lazily start top stories if discovery was disabled (and leave top stories enabled if user actually wants discovery)
> 
>     * also check PREF_CHANGED similar to above
> 
> 
> Caveats:
> 
>     * constructor has no access to state
> 
>     * PrefsFeed dispatches PREFS_INITIAL_VALUES on its INIT.. which.. alphabetically happens before TopStoriesFeed ?? so by the time TopStoriesFeed INITs, it might be able to just check store's pref state as it might never get the action

Thanks Ed! Updated PR to lazy load activity stream  top stories if discovery stream is not enabled.

lib/TopStoriesFeed.jsm
```diff
@@ -30,17 +30,29 @@ const SPOC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.spoc.impressions
...
+const DISCOVERY_STREAM_PREF = "discoverystream.config";
 
 this.TopStoriesFeed = class TopStoriesFeed {
-  constructor() {
-    this.spocCampaignMap = new Map();
+  constructor(ds) {
+    this.discoveryStreamEnabled = ds && ds.value && JSON.parse(ds.value).enabled;
+    if (!this.discoveryStreamEnabled) {
+      this.initializeProperties();
+    }
```
punamdahiya: @Mardak With this check in lazyLoadingTopStories, do you think we need it in constructor?

punamdahiya:
@mardak @ScottDowne Updated PR to leverage PREF_INITIAL_VALUES which gets triggered inside PrefsFeed.jsm init(). Prefsfeed.jsm INIT gets invoked before TopStoriesFeed INIT making at.INIT handling redundant inside TopSToriesFeed. Please review. Thanks!

lib/TopStoriesFeed.jsm
punamdahiya: Thinking more we should keep default initialization in place for TopStoriesFeed API

Mardak: The general pattern for feeds is to initialize properties that might be used from entrypoints (e.g., action callbacks) while deferring any actual work to INIT or other actions. If we do move properties to a later time, are we confident no entrypoint will try to access them? E.g., on SYSTEM_TICK, does it try to check some property that no longer exists?

lib/TopStoriesFeed.jsm
```diff
@@ -643,10 +655,24 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
+    this.init();
+  }
+
   async onAction(action) {
     switch (action.type) {
-      case at.INIT:
-        this.init();
+      // Check for pref initial values to lazy load activity stream top stories
+      case at.PREFS_INITIAL_VALUES:
```
Mardak: Oh this is neat. Probably worth a comment explicitly why we don't watch the usual INIT. As kinda implied in the other comment, there could be other actions that come in between INIT and PREFS_INITIAL_VALUES triggering something. I'm not sure if there are any other actions right now but nothing preventing some other feed from dispatching from their INIT either.

lib/TopStoriesFeed.jsm
```diff
@@ -30,17 +30,29 @@ const SPOC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.spoc.impressions
...
     this.contentUpdateQueue = [];
+    this.spocCampaignMap = new Map();
     this.cache = new PersistentCache(SECTION_ID, true);
     this._prefs = new Prefs();
   }
 
   async onInit() {
     SectionsManager.enableSection(SECTION_ID);
+    if (this.discoveryStreamEnabled) {
+      return;
```
Mardak: We `Services.obs.addObserver()` further down which we skip. So we should check enabled before we try to remove in uninit:
```
```console.log: "action type" "UNINIT"
JavaScript error: resource://activity-stream/lib/TopStoriesFeed.jsm, line 114: NS_ERROR_ILLEGAL_VALUE: Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIObserverService.removeObserver]```

File: lib/TopStoriesFeed.jsm (hunk 2 above)
Comment (Mardak) [reply lib/TopStoriesFeed.jsm]: I did confirm that there were no current actions between INIT and PREFS_INITIAL_VALUES:
```
"INIT"
"PREFS_INITIAL_VALUES"
"DISCOVERY_STREAM_CONFIG_SETUP"
"NEW_TAB_INIT"
"NEW_TAB_LOAD"
"NEW_TAB_STATE_REQUEST"
"NEW_TAB_INITIAL_STATE"
"UPDATE_SEARCH_SHORTCUTS"
"TELEMETRY_PERFORMANCE_EVENT"
"DISCOVERY_STREAM_LAYOUT_UPDATE"
"DISCOVERY_STREAM_SPOCS_ENDPOINT"
"DISCOVERY_STREAM_SPOCS_CAPS"
"DISCOVERY_STREAM_SPOCS_UPDATE"
"DISCOVERY_STREAM_SPOCS_FILL"
"DISCOVERY_STREAM_FEED_UPDATE"
```

File: lib/TopStoriesFeed.jsm [hunk 4]
```diff
@@ -492,7 +504,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
 
     if (this.storiesLoaded) {
       updateContent();
-    } else {
+    } else if (this.contentUpdateQueue) {
```
Mardak: I think we should be safer if we just prevent any unintended actions directly from `onAction`. E.g.,

```js
onAction(action) {
  if (discoveryStreamEnabled) {
    switch (action.type) {
      case PREFS_INITIAL_VALUES: …
      case PREF_CHANGED: …
      case UNINIT: …
    }
    return;
  }

  // existing fully running feed stuff
}
```

Mardak: This should also help ensure we don't waste cpu doing work on every NEW_TAB_REHYDRATED or other action

lib/TopStoriesFeed.jsm
punamdahiya: Checking discovery stream enabled before invoking remove observer is not full proof. It leaves out edge case where user manually disable and enable discovery stream experience back.

I have updated to invoke removeObserver inside try catch so that it continue uninit after catching error. Not sure if there is a way to check if observer exists before invoking remove, that should definitely be a better handling.

lib/TopStoriesFeed.jsm
punamdahiya: Done

punamdahiya:
@Mardak Travis is failing due to code coverage, do you know what flow it wants to be covered? PR is updated with review feedback and an edge case to avoid re-initialize properties for ROW where we have DS disabled. Thanks!

Mardak:
If you run testmc locally, after it fails coverage, you can `npm run debugcoverage` to open a web page with links that you can dig in to to see the red lines.
```
13 05 2019 17:32:37.407:ERROR [reporter.coverage-istanbul]: Coverage for statements (97.7%) in file /home/travis/build/mozilla/activity-stream/lib/TopStoriesFeed.jsm does not meet per file threshold (100%)
13 05 2019 17:32:37.408:ERROR [reporter.coverage-istanbul]: Coverage for lines (97.6%) in file /home/travis/build/mozilla/activity-stream/lib/TopStoriesFeed.jsm does not meet per file threshold (100%)
13 05 2019 17:32:37.408:ERROR [reporter.coverage-istanbul]: Coverage for functions (98.55%) in file /home/travis/build/mozilla/activity-stream/lib/TopStoriesFeed.jsm does not meet per file threshold (100%)
```

punamdahiya:
> 13 05 2019 17:32:37.408:ERROR [reporter.coverage-istanbul]: Coverage for functions (98.55%) in file /home/travis/build/mozilla/activity-stream/lib/TopStoriesFeed.jsm does not meet per file threshold (100%)

Its complaining https://github.com/mozilla/activity-stream/pull/4992/files#diff-ab2f31cc847abe1fcf02369e79b4dd4dR779 which looks like false alert

Mardak:
The line numbers are very off :( Looking over the added tests and code, it seems like at least UNINIT path is not tested?

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -91,12 +91,86 @@ describe("Top Stories Feed", () => {
...
+      instance.onAction({type: at.PREFS_INITIAL_VALUES, data: {"discoverystream.config": JSON.stringify({enabled: false})}});
+      assert.notCalled(instance.handleDisabled);
+
+      instance.onAction({type: at.NEW_TAB_REHYDRATED, meta: {fromTarget: {}}});
+      assert.calledOnce(instance.getPocketState);
+    });
+
+    it("should fire init on PREF_CHANGED", () => {
+      instance.onInit = sinon.spy();
+      instance.onAction({type: at.PREF_CHANGED, data: {name: "discoverystream.config", value: {}}});
```
Mardak: This test isn't testing with `discoveryStreamEnabled = false`

Mardak: Or true I think we actually want. Perhaps we want to put `instance.discoveryStreamEnabled = true;` in a `beforeEach` and only explicitly set it to false in certain tests.

test/unit/lib/TopStoriesFeed.test.js
```diff
@@ -91,12 +91,86 @@ describe("Top Stories Feed", () => {
...
+      instance.onAction({type: at.PREFS_INITIAL_VALUES, data: {"discoverystream.config": JSON.stringify({enabled: true})}});
+      assert.calledOnce(instance.handleDisabled);
+
+      instance.onAction({type: at.NEW_TAB_REHYDRATED, meta: {fromTarget: {}}});
+      assert.notCalled(instance.getPocketState);
+    });
+
+    it("should handle NEW_TAB_REHYDRATED when discoverystream is disabled", async () => {
+      instance.discoveryStreamEnabled = false;
+      instance.handleDisabled = sinon.spy();
```
Mardak: This line is replacing `handleDisabled` preventing the original code from running. If you actually want to spy it (i.e., allow it to run instead of this behavior of effectively stubbing), you'll want to `sinon.spy(instance, "handleDisabled")`

test/unit/lib/TopStoriesFeed.test.js
punamdahiya: Done

punamdahiya:
@Mardak Thanks for helping debug code coverage, updated tests to clearly call out stubbed behavior and covering actions flow for both when DS is enabled and disabled. Please review

lib/TopStoriesFeed.jsm
```diff
@@ -30,17 +30,30 @@ const SPOC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.spoc.impressions
...
+      this.initializeProperties();
+    }
+  }
+
+  initializeProperties() {
     this.contentUpdateQueue = [];
+    this.spocCampaignMap = new Map();
     this.cache = new PersistentCache(SECTION_ID, true);
     this._prefs = new Prefs();
+    this.isInitialized = true;
```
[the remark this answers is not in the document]
punamdahiya: +1 for propertiesInitialized

lib/TopStoriesFeed.jsm
```diff
@@ -30,17 +30,30 @@ const SPOC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.spoc.impressions
 const REC_IMPRESSION_TRACKING_PREF = "feeds.section.topstories.rec.impressions";
 const OPTIONS_PREF = "feeds.section.topstories.options";
 const MAX_LIFETIME_CAP = 500; // Guard against misconfiguration on the server
+const DISCOVERY_STREAM_PREF = "discoverystream.config";
 
 this.TopStoriesFeed = class TopStoriesFeed {
-  constructor() {
-    this.spocCampaignMap = new Map();
+  constructor(ds) {
+    this.discoveryStreamEnabled = ds && ds.value && JSON.parse(ds.value).enabled;
```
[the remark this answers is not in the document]
punamdahiya: Done

punamdahiya:
@Mardak Updated PR fixing nits. Thanks for the review!

punamdahiya:
@Mardak In 66 experiments, we used browser.newtabpage.activity-stream.discoverystream.optOut.0 pref to allow users to opt-out of discovery stream experiment.

Is there a way to find out how many users opted-out or have this pref set to true, essentially weighing if we need to put check for this pref before disabling TopStoriesFeed initialization.

PR #4994, opened 2019-05-07
Port 1549906 - Use brandings.ftl to allow OnboardingMessageProvider.jsm to use new strings

Mardak:
This isn't needed for current export and it actually makes us do an additional change in the next export reverting  https://hg.mozilla.org/mozilla-central/rev/c688fc5b2c0b

We'll want this as part of the uplift to avoid needing the all_files_referenced change as part of the uplift.

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -42,6 +42,7 @@ async function getAddonInfo() {
 
 const L10N = new Localization([
   "branding/brand.ftl",
+  "browser/branding/brandings.ftl",
   "browser/branding/sync-brand.ftl",
   "browser/newtab/onboarding.ftl",
 ]);


Mardak:
Ah yes taskcluster is failing. Maybe we should fix in mozilla-central and port to central and trailhead (so we don't forget to include when uplifting trailhead branch).

Mardak:
I'll merge to central and cherrypick directly to trailhead as we don't want to miss it when uplifting.

Mardak:
I merged central to master to include this port and hopefully fix up people's PR automation

Mardak:
Cherrypicked to trailhead branch https://github.com/mozilla/activity-stream/commit/3c010a333b3de74141cf4c32e480f99fda4c6f1b

PR #4995, opened 2019-05-08
Bug 1549995 - The "SKIPPED_SIGNIN" telemetry ping is wrongly sent after the sign in flow from the "First Run Modal" is completed

rlr:


content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -123,13 +123,18 @@ export class _Trailhead extends React.PureComponent {
...
     global.removeEventListener("visibilitychange", this.closeModal);
     this.props.document.body.classList.remove("welcome");
     this.props.document.getElementById("root").removeAttribute("aria-hidden");
     this.setState({isModalOpen: false});
     this.revealCards();
-    this.props.dispatch(ac.UserEvent({event: "SKIPPED_SIGNIN", ...this._getFormInfo()}));
+
+    // If closeModal() was triggered by a visibilitychange event, the user actually
+    // submitted the email form so we don't send a SKIPPED_SIGNIN ping.
+    if (ev && ev.type !== "visibilitychange") {
```
rlr: is that too weird/obscure?

rlr: I guess `!ev || ev.type !== “visibilitychange”` would be better.

Mardak: Yeah !ev should also catch clicking outside / hitting escape. There's 3 callers of closeModal: onsubmit visibilitychange, start button, modaloverlay (keyboard and click)

Mardak: I suppose we could make sure onClose is called always with an event if we fix https://github.com/mozilla/activity-stream/blob/4f311caef9d9aba89126ba26cb1487a7378c5ae6/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx#L11 to pass in event

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -123,13 +123,18 @@ export class _Trailhead extends React.PureComponent {
...
     global.removeEventListener("visibilitychange", this.closeModal);
     this.props.document.body.classList.remove("welcome");
     this.props.document.getElementById("root").removeAttribute("aria-hidden");
     this.setState({isModalOpen: false});
     this.revealCards();
-    this.props.dispatch(ac.UserEvent({event: "SKIPPED_SIGNIN", ...this._getFormInfo()}));
+
+    // If closeModal() was triggered by a visibilitychange event, the user actually
+    // submitted the email form so we don't send a SKIPPED_SIGNIN ping.
+    if (!ev || ev.type !== "visibilitychange") {
```
piatra: @rlr if `!ev === true` then the second part of the `||` will be evaluated and it will try to access a property on something that's  null / undefined?

PR #4996, opened 2019-05-08
Backport d30176

piatra:


Mardak:
The test failure is fixed on latest central branch but we shouldn't land this until https://hg.mozilla.org/integration/autoland/rev/a371ed7348b6 reaches mozilla-central

PR #4999, opened 2019-05-08
Bug 1546733 - [Card] Place source above title and adjust vertical spacing

gvn:
QA: Run `dev-test-all` layout and compare to requests in [1546733](https://bugzilla.mozilla.org/show_bug.cgi?id=1546733)


ScottDowne:
Putting that link at the bottom had some flexbox usage to do right.

Do we still need it? I'm wondering if we can kill them. I see:

`.ds-card-link` with:
```
display: flex;
flex-direction: column;
justify-content: space-between;
```

and `.meta` with:
```
display: flex;
flex-direction: column;
flex-grow: 1;
```

The above was only there to make that soure text float on the bottom of the card, I think.

Also, `.info-wrap` has a `margin: 0 0 12px;` and `flex-grow: 1;` that I wonder if we still need.

PR #5001, opened 2019-05-08
Bug 1550138 - DS feed images loaded with idle

ScottDowne:


gvn:
Did we decide if this is something we want to do? I recall concern over forcing people to download stuff they might never see.

I do think this approach is better for perceptual performance since it's less likely you'll see a gray box as you scroll.

My vote is to do this, but I'm trying to remember who was worried about it... 🤔

ScottDowne:
Closing this for now. It was an experiment to see how hard it was, result, not hard at all.

there are some questions about if we want to do this at all, and I think we can keep that contained in 1550138

gvn:
Yeah, another approach might be to do predictive loading when you get within a certain distance of an image (say 3-500px away) instead of right at the top border of it. That would also cut down on some of the flashing.

PR #5004, opened 2019-05-08
Bug 1549494 - Update image and styles for the sync branch of first run modal

rlr:
I had to add that animation in manually. I'm still tweaking it because it appears as if it was getting clipped but not sure if it's an illusion or some sort of anti aliasing effect.

Do you have some SVG superpowers?


rlr:
NOTE this is against master and changing the base isn't doing what I want.

rlr:
hang tight... got a new svg to amimate.

rlr:
guess i'll just close this and make a new one against the right branch.

PR #5006, opened 2019-05-08
Bug 1549494 - Update image and styles for the sync branch of first run modal

rlr:
UX Approved: https://www.dropbox.com/s/h1703pflaxpdpty/secondtry.mov?dl=0


content-src/asrouter/templates/Trailhead/_Trailhead.scss
```diff
@@ -63,7 +63,7 @@
 
     .trailheadContent {
       .trailheadBenefits {
-        background: url('#{$image-path}sync-devices.svg');
+        background: url('#{$image-path}sync-devices-trailhead.svg');
```
Mardak: we do have a trailhead directory although not sure if it'll be confusing to have the same leaf file name…

PR #5016, opened 2019-05-10
Bug 1550642 - Improve accessibility of autofocused email

rlr:


rlr:
a11y panel picked it up:

Screen Shot 2019-05-10 at 8 24 21 AM

rlr:
hmm, looks like we can take out the required and it will still validate for a well formed email address. But if it's empty it submits. So we could handle the one empty case separately in the `onSubmit`. Worth it?

PR #5017, opened 2019-05-10
Bug 1550723 - An inactive horizontal scrollbar is displayed at the bottom of the about:welcome pages which contain a modal

rlr:


rlr:
oh I guess that's what QA does on Mac... yeah yeah ... hmm. but kind of yuck.. 😝

PR #5018, opened 2019-05-10
Bug 1550714 - Notify all tabs of messages that have become invalid of…

piatra:
… pref change

f? @rlr . Initially I wanted to check if `this.state.messages[lastMessageId]` still has valid targeting and send a message to all tabs notifying if the should clear the message. `lastMessageId` is set to whatever the preloaded tab requested.. most of the time.
Unfortunately you can also trigger a CFR notification on your focused tab (after the preloaded tab loaded) which invalidates `lastMessageId`. This is why I went through all unblocked messages. The idea is to add a listener in content that checks if they are currently displayed a message that has become invalid and to remove it if needed.


lib/ASRouterPreferences.jsm
```diff
@@ -7,6 +7,7 @@ const {Services} = ChromeUtils.import("resource://gre/modules/Services.jsm");
 
 const PROVIDER_PREF_BRANCH = "browser.newtabpage.activity-stream.asrouter.providers.";
 const DEVTOOLS_PREF = "browser.newtabpage.activity-stream.asrouter.devtoolsEnabled";
+const FXA_USERNAME_PREF = "services.sync.username";
```
rlr: oh, this is a clever way to detect user logged in. i like 👍

piatra:
In order to test this open a few new tabs, make sure the message is there. Then set the value of `services.sync.username` to anything. Ensure that the message is not present in existing tabs, when opening a new tab you should not see the snippet.

test/unit/asrouter/ASRouterPreferences.test.js
```diff
@@ -13,8 +13,9 @@ const CFR_USER_PREF_FEATURES = "browser.newtabpage.activity-stream.asrouter.user
  *  3. browser.newtabpage.activity-stream.feeds.snippets (user preference - snippets)
  *  4. browser.newtabpage.activity-stream.asrouter.userprefs.cfr.addons (user preference - cfr)
  *  4. browser.newtabpage.activity-stream.asrouter.userprefs.cfr.features (user preference - cfr)
+ *  5. services.sync.username
```
rlr: eh.. there are two 4's above.

Mardak:
I guess we'll merge this to master for now and separately cherry-pick to trailhead branch if we want to uplift.

PR #5020, opened 2019-05-10
Bug 1548917 - [Accessibility] Focus state refinements

gvn:
QA:

- Set focus to a card
- Compare to [design spec](https://www.figma.com/file/0xZkmT86rSXbGRB4eIvrn0ts/New-Tab-Spec-for-Fx-v68?node-id=383%3A1)

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
+++ b/content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
@@ -28,16 +28,6 @@ $col4-header-font-size: 14;
 
       box-shadow: 0 1px 4px 0 $grey-90-10;
 
-      &:hover {
-        @include dark-theme-only {
-          box-shadow: 0 0 0 5px $grey-60;
-        }
-
-        box-shadow: 0 0 0 5px $grey-30;
-        transition: box-shadow 150ms;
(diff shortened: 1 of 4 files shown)


gvn:
Whoever has time to review...thx!

gvn:
Fixed.

punamdahiya:
@wolasi [spec here](https://www.figma.com/file/0xZkmT86rSXbGRB4eIvrn0ts/New-Tab-Spec-for-Fx-v68?node-id=383%3A0) says domain color as #005A71, is that for focus and active state?

punamdahiya:
Sees below state (last story link stayed visited blue), if a user drags the link and change focus to another story. I believe its an edge case and should be ok.
blue text stayed after lose focus

gvn:
Thanks! Design was approved as implemented, but follow-on tickets can certainly be filed if more changes are desired. 👍

wolasi:
> @wolasi [spec here](https://www.figma.com/file/0xZkmT86rSXbGRB4eIvrn0ts/New-Tab-Spec-for-Fx-v68?node-id=383%3A0) says domain color as #005A71, is that for focus and active state?

@punamdahiya that was referencing the older domain color, this task predates another task in which domain color was changed to gray.

PR #4997, opened 2019-05-08
Bug 1550053 - Fix busted topsites and topstores configs enable/disable

ScottDowne:
To test:

Turn off either topsites, or topstores, by clicking the gear in the topright of about:home

It should disable either of those, but display the other one, and not look broken.

If you see "Oops, something went wrong loading this content.
Refresh page to try again." or a topstories message title with no topstories, it's broken.

PR #5024, opened 2019-05-13
Bug 1550107 - Insert placeholders on feed cards while they load and slight spoc adjustment.

ScottDowne:


gvn:
Are there QA steps for this?

punamdahiya:
> Are there QA steps for this?

1. After cloning fetch delay https://github.com/ScottDowne/fetch-delay , run npm install and node start.js. We need this to delay server feeds and spocs to simulate slow load.
2. Set pref browser.newtabpage.activity-stream.discoverystream.endpoints to https,http
3. Set pref browser.newtabpage.activity-stream.discoverystream.config to 
{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"https://gist.githubusercontent.com/punamdahiya/5cc3a4c8eb6923cbaa01a6e38c7b3054/raw/37b5ebc24c24ab6de008ca3bbcf49f59b2a3a381/delay-feed-multiple.json"}

4. Load a new tab, you should see first feed for component list loading with delay of 5 seconds. While it waits for loading it should show placeholders.
5.  After 5 seconds next feed for component card grid should show up after delay of 10 seconds 
6. Last Hero component should show up after 15 seconds

content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -53,23 +55,17 @@ export class Hero extends React.PureComponent {
...
-        items={this.props.items - 1}
-        type={`Hero`} />
-    );
+    let heroCard = null;
 
-    return (
-      <div className={`ds-hero ds-hero-${this.props.border}`}>
+    if (!heroRec || heroRec.placeholder) {
+      heroCard = (
+        <div className="ds-card placeholder" />
```
punamdahiya: I believe Hero placeholder should be given some default height, will be good to get wolasi take on it. Here are initial design specs:

https://bugzilla.mozilla.org/show_bug.cgi?id=1529983#c2

content-src/lib/selectLayoutRender.js
```diff
@@ -108,9 +127,11 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
       for (const component of row.components) {
```
punamdahiya: We should put the filter here
const component of row.components.filter(c => !filterArray.includes(c.type))) {

See a bug that exists in master and totally ok submitting fix as separate patch. This looks like fallout of DS Sequential load and followup fix https://github.com/mozilla/activity-stream/pull/4997/files

To replicate the issue, set hardcoded_layout to false and try hiding topsites. It will fail to filter TopSites. We are only filtering TopSites from components in multiple layouts ( e.g. hard coded layout) and not from multiple components inside a single layout ( e.g. basic variant).

test/unit/content-src/lib/selectLayoutRender.test.js
```diff
@@ -283,7 +282,7 @@ describe("selectLayoutRender", () => {
 
     assert.equal(layoutRender[0].components[0].type, "foo1");
     assert.equal(layoutRender[0].components[1].type, "foo2");
-    assert.equal(layoutRender[0].components[2], undefined);
+    assert.isTrue(layoutRender[0].components[2].data.recommendations[0].placeholder);
```
punamdahiya: Only one component (which causes early bail out)  has placeholder when feed isn't ready. Will be good to add assert showing layout render has only three components  and fourth component is undefined. 

assert.lengthOf(layoutRender[0].components, 3);
assert.isUndefined(layoutRender[0].components[3]);

punamdahiya:
LGTM with Hero placeholder and topsites issue addressed. Thanks!

content-src/lib/selectLayoutRender.js
ScottDowne: I'm not sure I'm following totally.

Also, I don't know if I did the steps to test right, as topsites filtered fine when I used a non hard coded layout.

ScottDowne: I see, I was trying by clicking the collapse button, but if I disable it in the config, I see the issue you see now.

Nice catch.

ScottDowne: If I add the change you suggested, we regress 1550053

I don't fully understand why hardcoded seems to make a difference.

It looks like with one fix, we break hardcoded false, and with the other, we only break hardcoded true.

ScottDowne: Ok, cool, looks like it's not an issue with either this patch, or with hardcoded or not layouts.

The issue seems to be between basic and the hardcoded layout.

On master, if you see the layout to not be hardcoded, and set it to basic, you see the issues you describe above, if you set it to not be hardcoded but set the layout to the same layout that is hardcoded, it works.

Here is the pref for non hardcoded layout that's the same as the hardcoded layout. :S
`{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":false,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=3-col-1-row-octr-4-col-6-row-categories"}`

ScottDowne: So something up with basic on master. Dunno if we want to regress 1550053 and if we do want that change, we probably want it in another ticket.

punamdahiya: Here's the difference I see in layout feeds:
For basic https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=basic

For hardcoded layout or a complex variant: https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=40249-e88c401e1b1f2242d9e441c4&layout_variant=66-release-2-above-3-below-4
May be its mathijs questions why basic variant has single layout component with multiple types inside it?

ScottDowne: I see. If I have the filter in both spots, things probably work.

ScottDowne: https://bugzilla.mozilla.org/show_bug.cgi?id=1552237

PR #5025, opened 2019-05-14
Bug 1551359 - Fill cards with longer excerpts for shorter titles

Mardak:
r?@gvn or @punamdahiya This doesn't change over all uses of the existing CSS mixin, e.g., hero, so I left that untouched as we'll need some JS to dynamically set clamping anyway.

Before card:
Screen Shot 2019-05-13 at 4 48 23 PM

Before list:
Screen Shot 2019-05-13 at 4 47 49 PM

After card:
Screen Shot 2019-05-13 at 4 48 54 PM

After list:
Screen Shot 2019-05-13 at 4 49 16 PM


Mardak:
Wolasi says to use all but one line for the title, so up to 5 lines of title and 1 for excerpt for the 6-line cards.

content-src/lib/clamp-title-excerpt.js
```diff
@@ -0,0 +1,30 @@
...
+ * how many lines the title can be and how many total lines with the excerpt.
+ */
+export function clampTitleExcerpt(title) {
+  if (title) {
+    // Make sure we have enough data to even clamp lines
+    const excerpt = title.nextSibling;
+    const {titleLines, totalLines} = title.dataset;
+    if (excerpt && titleLines && totalLines) {
+      // Set the appropriate clamp for excerpt if title uses fewer lines
+      const lineHeight = global.getComputedStyle(title).lineHeight.slice(0, -2);
```
gvn: Seems like `parseInt` would be more bulletproof and self-documenting than slicing characters?

content-src/lib/clamp-title-excerpt.js
```diff
@@ -0,0 +1,30 @@
...
+      const lines = title.clientHeight / lineHeight;
+      if (lines < titleLines) {
+        const clamp = totalLines - lines;
+        excerpt.style.webkitLineClamp = `${clamp}`;
+
+        // Override the CSS to allow larger-than-default clamp
+        excerpt.style.maxHeight = "unset";
+
+        return clamp;
```
gvn: Is this essentially just for automated test purposes at this point?

content-src/components/DiscoveryStreamComponents/List/List.jsx
```diff
@@ -44,7 +45,10 @@ export class ListItem extends React.PureComponent {
           url={this.props.url}>
           <div className="ds-list-item-text">
             <div>
-              <div className="ds-list-item-title">{this.props.title}</div>
+              <div className="ds-list-item-title"
+                data-title-lines="2"
+                data-total-lines="4"
+                ref={clampTitleExcerpt}>{this.props.title}</div>
```
gvn: It might be clearer to put the `ref` on a parent wrapper since you're implicitly operating on the next sibling as well, but not absolutely necessary.

content-src/lib/clamp-title-excerpt.js
Mardak: How did you want to pass the string in to parseInt? A number with "px" results in `NaN`

content-src/components/DiscoveryStreamComponents/List/List.jsx
Mardak: As in have the ref callback iterate through its children to find which ones to clamp?

content-src/lib/clamp-title-excerpt.js
gvn: That's odd. If I use the console and do something like `parseInt(getComputedStyle(document.querySelector('h1')).lineHeight)` it works.

Mardak: Hmm really odd. Pretty sure that's what I tried first but got `NaN` so then ended up with `slice`. I do see `parseInt("20px")` returning `20` so okay! I'll switch it

content-src/components/DiscoveryStreamComponents/List/List.jsx
gvn: Yeah, something like putting `data-total-lines` on the wrapper and then `data-lines` on the elements themselves perhaps. Just something more explicitly targeted...

content-src/lib/clamp-title-excerpt.js
gvn: I wonder if you might be hitting a race condition with the DOM. That was another concern I was going to raise separately because sometimes the DOM isn't fully ready when refs are linked, at least in regards to doing things like computed style measurements.

content-src/components/DiscoveryStreamComponents/List/List.jsx
Mardak: That seems reasonable. I suppose from the parent we can `querySelectorAll("[data-clamp]")`. Wolasi wants to have at least 1 line of excerpt which we can bake into the API to count the number of things that want clamping and do some math. And the last thing / excerpt doesn't need an explicit number as it's the thing getting excess lines.

content-src/lib/clamp-title-excerpt.js
Mardak: Yeah it was only added for testing, but I could probably just check for webkitLineClamp property instead.

Mardak:
Basically reimplemented everything to handle the desired "use up all lines for title except 1 for excerpt":
Screen Shot 2019-05-16 at 11 44 24 PM

Mardak:
Wolasi doesn't want it to fill up as many lines now.

Mardak:
Updated with wolasi's latest:
```
*Cards and Hero*
4 line max for title, 2 line minimum for excerpt (6 
Ines total)

*List with Excerpt*
3 line max fo excerpt (4 line total)
```
Screen Shot 2019-05-17 at 8 19 29 AM

PR #5026, opened 2019-05-14
Bug 1550149 - About preferences switch Recommended by Pocket descript…

punamdahiya:
…ion back to previous string

### DIFF
--- a/lib/AboutPreferences.jsm
+++ b/lib/AboutPreferences.jsm
@@ -106,7 +106,6 @@ this.AboutPreferences = class AboutPreferences {
 
       if (obj.id === "topstories") {
         obj.rowsPref = "";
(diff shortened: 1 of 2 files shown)


punamdahiya:
@wolasi @kanhema Attaching screenshot with updated Top Stories desc string

about-pref

punamdahiya:
As discussed, for now keeping 'prefs_content_discovery_description' string in case we need to reuse it in about:preferences.   https://github.com/mozilla/activity-stream/blob/master/locales-src/en-US/strings.properties#L95

I can open bug to revisit and  remove this string in 69/70.

punamdahiya:
Thanks! Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1551653 as follow up

PR #5027, opened 2019-05-14
Bug 1551248 Fixing dark vs light theme issue

ScottDowne:


ScottDowne:
Just a light vs dark theme issue.

To test:
1. use dev-test-all
2. Scroll to the bottom looking for list items.
3. Try in both light and dark theme.

Should show list body text in both cases.

content-src/components/DiscoveryStreamComponents/List/_List.scss
```diff
@@ -206,7 +206,10 @@ $item-line-height: 20;
 
   .ds-list-item-excerpt {
     @include limit-visibile-lines(2, $item-line-height, $item-font-size);
-    color: $grey-10-80;
+    @include dark-theme-only {
+      color: $grey-10-80;
+    }
+    color: var(--newtab-text-secondary-color);
```
gvn: I've been avoiding using browser vars for DS and using explicit vars from `content-src/styles/_variables.scss` instead for clarity (like you did for the `dark-theme-only` case above).

Can you do that so it's consistent with the rest of the DS styles?

ScottDowne: That's fine with me.

punamdahiya:
Seeing empty text inside list for default theme on master. 
list-text-missing

punamdahiya:
> Seeing empty text inside list for default theme on master.


Coming from cache, looks good

PR #5028, opened 2019-05-14
Bug 1551677 - Improve accessibility of trailhead dialog

Mardak:
r?@rlr Fixes both required and windows dialog focus bug. For the latter, I'm getting a windows try build for asa to test, so we shouldn't merge until we confirm it actually fixes the problem…

https://treeherder.mozilla.org/#/jobs?repo=try&revision=5533578d7712c703e3ff785526db6d577c37c91c

### DIFF
--- a/content-src/asrouter/templates/Trailhead/Trailhead.jsx
+++ b/content-src/asrouter/templates/Trailhead/Trailhead.jsx
@@ -119,7 +119,17 @@ export class _Trailhead extends React.PureComponent {
     }
   }
 
-  onSubmit() {
+  onSubmit(event) {
+    // Dynamically require the email on submission so screen readers don't read
+    // out it's always required because there's also ways to skip the modal
+    const {email} = event.target.elements;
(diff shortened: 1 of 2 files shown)


content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -119,7 +119,17 @@ export class _Trailhead extends React.PureComponent {
...
   }
 
-  onSubmit() {
+  onSubmit(event) {
+    // Dynamically require the email on submission so screen readers don't read
+    // out it's always required because there's also ways to skip the modal
+    const {email} = event.target.elements;
+    if (!email.value.length) {
+      email.required = true;
+      email.checkValidity();
```
rlr: ah, that's a nice way to do it. neat 👍

Mardak:
jteh: `Confirmed that both fixes are working nicely here. Thanks much.`

PR #5029, opened 2019-05-14
Bug 1551566 - Enable jsx-a11y eslint for ASRouterAdmin/asrouter directory and fix errors

AllegroFox:
Fixed accessibility linting errors for asrouter and ASRouterAdmin components.

### DIFF
--- a/content-src/asrouter/components/RichText/RichText.jsx
+++ b/content-src/asrouter/components/RichText/RichText.jsx
@@ -24,7 +24,8 @@ export function convertLinks(links, sendClick, doNotAutoBlock, openNewWindow = f
       // Setting the value to false will not include the attribute in the anchor
       const url = action ? false : safeURI(links[linkTag].url);
 
-      acc[linkTag] = (<a href={url}
+      acc[linkTag] = (<a href={url} // eslint-disable-line jsx-a11y/anchor-has-content
+        // eslint was getting a false positive caused by the dynamic injection of content.
         target={openNewWindow ? "_blank" : ""}
         data-metric={links[linkTag].metric}
         data-action={action}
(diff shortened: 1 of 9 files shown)


content-src/asrouter/components/RichText/RichText.jsx
```diff
@@ -24,7 +24,8 @@ export function convertLinks(links, sendClick, doNotAutoBlock, openNewWindow = f
       // Setting the value to false will not include the attribute in the anchor
       const url = action ? false : safeURI(links[linkTag].url);
 
-      acc[linkTag] = (<a href={url}
+      acc[linkTag] = (<a href={url} // eslint-disable-line jsx-a11y/anchor-has-content
+        // eslint was getting a false positive caused by the dynamic injection of content.
```
Mardak: You can add a general override for this file here:
https://github.com/mozilla/activity-stream/blob/23821cf7560b709145ffc5b05ba1dc3a73daeade/.eslintrc.jsx-a11y.js#L7-L14

Although perhaps this inline comment might be better as it's for this one place.

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -480,7 +480,7 @@ export class ASRouterAdminInner extends React.PureComponent {
     if (!this.state.providers) {
       return null;
     }
-    return (<p>Show messages from <select value={this.state.messageFilter} onChange={this.onChangeMessageFilter}>
+    return (<p>Show messages from <select value={this.state.messageFilter} onBlur={this.onChangeMessageFilter}>
```
Mardak: Travis automation has this test failing: https://github.com/mozilla/activity-stream/blob/5175b975c5121b561e1ea4945f10ba60855f05c6/test/unit/content-src/components/ASRouterAdmin.test.jsx#L167-L178

You can run all the tests locally with `npm test`

If you want to run just this one test linked above, you can edit the file temporarily adding a `.only` to become `it.only("should hide…")`

AllegroFox:
The issue seems to be with a linting change I made - The test is simulating the wrong event now. (it should be onBlur instead of onChange). I'll tweak that and check again.

PR #5030, opened 2019-05-14
Bug 1551719 - Remove discoverystream.optOut.0 pref

punamdahiya:


punamdahiya:
@Mardak As discussed, submitting patch that removes support for optOut pref. Please review. Thanks!

PR #5031, opened 2019-05-15
Bug 1552009 - Remove husky in order to support mozilla-central-first workflow

k88hudson:
In order to support a mozilla-central-first workflow, I'd like to propose removing husky as a dependency and adding an optional pre-push hook that can be manually installed by anyone who's interested. This addresses the issue where running `npm install` in a sub-directory causes weird behaviour since husky assumes it's in the root directory.

I'm open to using an alternative package that doesn't auto-install (or possibly creating a "setup" npm task) if we don't think this manual process is sufficient. I'll send out an email describing it.

What do you think @Mardak @dmose @ScottDowne?


Mardak:
Anything extra need to be done to clean up existing development environments? It looks like `npm install` without `husky` does automatically clean up `.git/hooks` already

k88hudson:
> Anything extra need to be done to clean up existing development environments? It looks like `npm install` without `husky` does automatically clean up `.git/hooks` already

I'll send something out to the dev list explaining the clean up process

PR #5035, opened 2019-05-15
Bug 1551613 - Enable jsx-a11y eslint for TopSites component and fix errors

AllegroFox:


Mardak:
Looks like you'll want to rebase on latest master which includes your #5029 and then `push -f`

content-src/components/TopSites/TopSite.jsx
```diff
@@ -183,7 +183,9 @@ export class TopSiteLink extends React.PureComponent {
     }
     return (<li className={topSiteOuterClassName} onDrop={this.onDragEvent} onDragOver={this.onDragEvent} onDragEnter={this.onDragEvent} onDragLeave={this.onDragEvent} {...draggableProps}>
       <div className="top-site-inner">
-         <a href={link.searchTopSite ? undefined : link.url} tabIndex="0" onKeyPress={this.onKeyPress} onClick={onClick} draggable={true}>
+        {/*  Enhance accessibility of drag-and-drop anchor tags - see bug 1552005 */}
```
k88hudson: I'd maybe word this as: "We don't yet support an accessible drag-and-drop implementation, see Bug 1552005"

AllegroFox: I'll adjust that!

content-src/components/TopSites/TopSites.jsx
```diff
@@ -139,7 +139,7 @@ export class _TopSites extends React.PureComponent {
         <div className="edit-topsites-wrapper">
           {editForm &&
             <div className="edit-topsites">
-              <div className="modal-overlay" onClick={this.onEditFormClose} />
+              <div className="modal-overlay" onClick={this.onEditFormClose} role="presentation" />
```
k88hudson: This is correct, although in addition to this I think it might be worth filing a bug (not under this metabug, just under http://as-bugzy.herokuapp.com/feature/1550082 with the `access` keyword) to support the escape button for closing all modals

content-src/components/TopSites/_TopSites.scss
```diff
@@ -9,6 +9,7 @@ $default-icon-size: 32px;
 $default-icon-offset: 6px;
 $half-base-gutter: $base-gutter / 2;
 $hover-transition-duration: 150ms;
+$default-link-color: #1A0DAB;
```
k88hudson: In order to support dark theme, you'll want to use `var(--newtab-link-primary-color)` here. You could also consider creating a generic link-like-button class you could use here if you want

content-src/components/TopSites/TopSiteFormInput.jsx
```diff
@@ -42,12 +52,14 @@ export class TopSiteFormInput extends React.PureComponent {
       <div className={`field ${typeUrl ? "url" : ""}${validationError ? " invalid" : ""}`}>
         {this.props.loading ?
           <div className="loading-container"><div className="loading-animation" /></div> :
-          showClearButton && <div className="icon icon-clear-input" onClick={this.props.onClear} />}
+          showClearButton && <div className="icon icon-clear-input" onClick={this.props.onClear} onKeyPress={this.onClearIconPress} role="button" tabIndex="0" />}
```
k88hudson: I think it's a bit weird that this gets tabbed to first before the input itself, could you maybe just switch this to a button element instead?

content-src/components/TopSites/_TopSites.scss
AllegroFox: I'll see how often it comes up, and create a class if it keeps happening 👍

content-src/components/TopSites/TopSiteFormInput.jsx
AllegroFox: sure!

content-src/components/TopSites/_TopSites.scss
k88hudson: There's also a bit of padding you need to deal with

AllegroFox: Our new schmancy component is integrated and seems to be working as expected :)

content-src/components/TopSites/_TopSites.scss
```diff
@@ -406,8 +406,6 @@ $hover-transition-duration: 150ms;
       font-size: 13px;
       margin-top: 4px;
       cursor: pointer;
-      border: 0;
-      color: var(--newtab-link-primary-color);
 
       &:hover {
         text-decoration: underline;
```
Mardak: Is this style still needed?

content-src/components/A11yLinkButton/A11yLinkButton.jsx
```diff
@@ -0,0 +1,11 @@
+import {FormattedMessage} from "react-intl";
+import React from "react";
+
+export function A11yLinkButton(props) {
+  return (
+    <button {...props} className="a11y-link-button">
+      {props.children}
+      <FormattedMessage id="topsites_form_use_image_link" />
```
k88hudson: Since this is specific to the top sites form, in this case you actually want this to be in `TopSiteForm.jsx`, i.e.
```jsx
<A11yLinkButton onClick={this.onEnableScreenshotUrlForm} className="enable-custom-image-input">
  <FormattedMessage id="topsites_form_use_image_link" />
</A11yLinkButton>
```

content-src/components/A11yLinkButton/_A11yLinkButton.scss
```diff
@@ -0,0 +1,11 @@
+
+.a11y-link-button {
+  border: 0;
+  padding: unset;
```
k88hudson: I don't think you get anything extra from `unset` here, `0` would be more appropriate

content-src/components/A11yLinkButton/_A11yLinkButton.scss
```diff
@@ -0,0 +1,11 @@
+
+.a11y-link-button {
```
k88hudson: add `cursor:pointer`

content-src/components/TopSites/_TopSites.scss
Mardak: Ah yup, looks like there's no more `_TopSites.scss` changes in the PR right now, but this `:hover` style should be redundant with the one in `_A11yLinkButton.scss`

content-src/components/A11yLinkButton/A11yLinkButton.jsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+export function A11yLinkButton(props) {
+  return (
+    <button {...props} className="a11y-link-button">
```
k88hudson: Can you put `type="button"` here as well, before the `{...props}` (so it's possible to override?) Actually it would be a good idea to put `className` before so you can override that too

content-src/components/TopSites/TopSiteFormInput.jsx
```diff
@@ -33,23 +43,35 @@ export class TopSiteFormInput extends React.PureComponent {
...
   }
 
-  render() {
+  renderLoadingOrCloseButton() {
     const showClearButton = this.props.value && this.props.onClear;
+
+    if (this.props.loading) {
+      return (<div className="loading-container"><div className="loading-animation" /></div>);
+    } else if (showClearButton) {
+      return (<button className="icon icon-clear-input icon-button-style"
```
k88hudson: add `type="button"` here to prevent it submitting the form

content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -1,4 +1,5 @@
 import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
```
k88hudson: Ideally this form should call `onPreviewButtonClick` or `onDoneButtonClick` on a `onSubmit`, which would require some refactoring

PR #5038, opened 2019-05-17
Bug 1549688 - Return early if BookmarkPanelHub is not initialized

piatra:


lib/BookmarkPanelHub.jsm
```diff
@@ -64,6 +64,10 @@ class _BookmarkPanelHub {
    * @returns {obj|null} response object or null if no messages matched
    */
   async messageRequest(target, win) {
+    if (!this._initalized) {
```
k88hudson: This is actually misspelled, can you change this (and other usages) to`_initialized`

PR #5039, opened 2019-05-17
Bug 1549471 - Optimize ds cache loads and saves

ScottDowne:


lib/DiscoveryStreamFeed.jsm
```diff
@@ -36,13 +38,25 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   constructor() {
     // Internal state for checking if we've intialized all our data
     this.loaded = false;
-
-    // Persistent cache for remote endpoint data.
-    this.cache = new PersistentCache(CACHE_KEY, true);
+    this._kvstore = KeyValueService.getOrCreate(OS.Constants.Path.localProfileDir, CACHE_KEY);
```
ncloudioj: Note that each rkv store supports multiple sub-stores, and the signature of `KeyValueService.getOrCreate(store_path, store_name)`, So you might want to add one more level to the `store_path`. For example, say we want to create a store for "discovery-stream" with a sub-store named "remote-content", 
```js
const path = OS.Path.join(
  OS.Constants.Path.localProfileDir,
  "discovery-stream"
);
this._kvstore = KeyValueService.getOrCreate(path, "remote-content");

// You can also create another sub-store, say "foo", in the same store
this._store = KeyValueService.getOrCreate(path, "foo");
```

See its API doc [here](https://searchfox.org/mozilla-central/source/toolkit/components/kvstore/nsIKeyValue.idl#45).

lib/DiscoveryStreamFeed.jsm
```diff
@@ -189,12 +203,13 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   }
 
   async _checkExpirationPerComponent() {
-    const cachedData = await this.cache.get() || {};
-    const {feeds} = cachedData;
+    const layout = await this.kvstoreGet("layout") || {};
```
ncloudioj: How many key/value pairs do we have in this store? If there are only three of them we can just load all of them in one call using the [enumerator](https://searchfox.org/mozilla-central/source/toolkit/components/kvstore/kvstore.jsm#173).

lib/DiscoveryStreamFeed.jsm
ScottDowne: Is there performance implications around loading something out of 1 store if it's got a lot of data, vs splitting the data into two smaller stores, and only loading it from the smaller store as needed?

I was under the impression that the size of the store wasn't significant.

ScottDowne:
I don't think there are performance gains to be had here.

I'm going to close it for now.

PR #5040, opened 2019-05-17
Bug 152289 - Fix search only for DS

ScottDowne:


Mardak:
Looks like this was intended for bug 1552289 (extra 5)

PR #5042, opened 2019-05-17
Followup 1551359 - Make sure we have integer lines

Mardak:
r?@punamdahiya I believe this should fix what you're seeing…


punamdahiya:
@mardak , still see the issue with the patch after 2-3 refresh

line-clamp-v1

Mardak:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1552596 and closing this for now as it seems to be an issue related to `separatePrivilegedContentProcess=true` which is only set that way on nightly with turning that on for more channels currently stalled -> not as urgent to fix this bug.

PR #5043, opened 2019-05-17
Bug 1550160 - Clean up unused card CSS

gvn:
Testing:

Compare `dev-test-all` between Nightly and this version to ensure layout hasn't regressed.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
@@ -39,9 +39,6 @@ $excerpt-line-height: 20;
   }
 
   .ds-card-link {
-    display: flex;
-    flex-direction: column;
-    justify-content: space-between;
     height: 100%;
 
     &:focus {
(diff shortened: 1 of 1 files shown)


punamdahiya:
For cards seen with hero component, removing flex from .meta is causing 16px upward shift of the info content 

With patch
patch-cleanup

On master
master

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -83,9 +80,6 @@ $excerpt-line-height: 20;
   }
 
   .meta {
-    display: flex;
```
punamdahiya: flex-grow inside .info-wrap needs this display:flex

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -83,10 +80,8 @@ $excerpt-line-height: 20;
...
-    flex-direction: column;
```
punamdahiya: we need flex-direction to align context 'sponsored by..'. Only CSS unused and safe to remove is flex-grow:1

gvn:
Should be good now...thanks!

PR #5054, opened 2019-05-21
Bug 1552179 - Turn on eslint-a11y for the rest of the content-src folder

AllegroFox:


content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -81,11 +81,11 @@ export class ContextMenuItem extends React.PureComponent {
   render() {
     const {option} = this.props;
     return (
-      <li role="menuitem" className="context-menu-item">
-        <a onClick={this.onClick} onKeyDown={this.onKeyDown} tabIndex="0" className={option.disabled ? "disabled" : ""}>
+      <li role="menuitem" className="context-menu-item" onClick={this.onClick} onKeyDown={this.onKeyDown} tabIndex="0" >
+        <span className={option.disabled ? "disabled" : ""}>
```
k88hudson: Now that I look at this - out of curiosity, what's your reasoning for using `<span>` rather than `<button>` here?

AllegroFox: the `<button>` creates a second tab stop that makes the functionality kind of irritating, because you have to tab twice to move past each menu item - makes it really easy to pick the wrong one.

PR #5058, opened 2019-05-22
Bug 1550115 - Include impression id to spoc request.

ScottDowne:
At the moment testing this is just a matter of turning on spocs and making sure they still render.

Testing that the actual pocket id works when passed in is hard given the server still hasn't be made to handle this. This part can have a larger round of QA as different pieces for this integration fall into place. That said, if you have reason to suspect I am not properly passing in the pocket id, you should def call it out.

Part of what we're doing with this is passing in the client id to pocket's spoc endpoint. (@ncloudioj do you know what kind of specialized reviews I may need for that?)

We needed to update this fetch to be post, and not get, because of the new id.

And on a personal level, I wanted to include in this is not need to rely on a pref. So, these changes should be backwards compatible. If the spoc url is setup to be adzerk, it should make use of the pocket id, if not, it should just be ignored and the old experience is business as usual.

Thoughts?

To test:

1. Set `browser.newtabpage.activity-stream.discoverystream.config` to default
2. In browser console, open up the network tab and probably clear it.
3. in about:config, change `browser.newtabpage.activity-stream.discoverystream.config` to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":true,"personalized":false,"layout_endpoint":"https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=basic"}`
4. In the network tab, you should now see a POST to unique-spocs and also check that requests params tab for something along these lines: `{"pocket_id":"{123-123-123-123-123}","consumer_key":"123345678"}`

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -237,6 +237,7 @@ const PREFS_CONFIG = new Map([
         show_spocs: showSpocs({geo}),
         hardcoded_layout: true,
         personalized: false,
+        adserver: false,
         // This is currently an exmple layout used for dev purposes.
         layout_endpoint: "https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=basic",
       });
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -237,7 +237,6 @@ const PREFS_CONFIG = new Map([
         show_spocs: showSpocs({geo}),
         hardcoded_layout: true,
         personalized: false,
-        adserver: false,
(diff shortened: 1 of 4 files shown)


ScottDowne:
before we land this we probably want some discussion around the security needed on passing the impression id.

ncloudioj:
> Part of what we're doing with this is passing in the client id to pocket's spoc endpoint. (@ncloudioj do you know what kind of specialized reviews I may need for that?)

I'd recommend to add this change to the legal review conversation along with the impression one. Other than that, I don't see anything else that blocks it.

PR #5060, opened 2019-05-22
 Bug 1552195 - Support the escape button for closing all modals

AllegroFox:
I used the ModalOverlay component to add keyboard support to TopSites, and tweaked the CSS to be a little more universal.

All the modals seem to be working properly with the exception of the ASRouterAdmin targeting modal, which I'm not sure was working properly in the first place. I could work on it here, or file another bug and leave it for someone who knows what it's meant to do :) It seems to have weird CSS rules applied to it somewhere and has no close() or done() functions...


k88hudson:
Looks like `ModalOverlay` needs to be adjusted to work with dark theme:
image

k88hudson:
Top corners look a bit weird for the search modal as well:
image

content-src/asrouter/components/ModalOverlay/_ModalOverlay.scss
```diff
@@ -20,7 +21,7 @@
 
 .modalOverlayInner {
   width: 960px;
-  height: 570px;
+  // height: 570px;
```
k88hudson: Removing the height makes the older onboarding kinda broken (see `ONBOARDING_1 ` message in about:newtab#devtools). We might be removing that though... let me ask about that

Also, if you are removing the fixed height you probably also want some kind of `top` value that isn't `top: calc(50% - 285px);` (i.e. based on a height of 570)

k88hudson: Yeah, so we are still keeping this around for now; do you think you could put it in a separate class? Steps to preview this UI: 
image

AllegroFox: Durrr I keep forgetting to check on dark mode. :P I'll tweak things, and make the `top` positioning more dynamic.

content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -25,8 +25,8 @@ export class ModalOverlayWrapper extends React.PureComponent {
   render() {
     const {props} = this;
     return (<React.Fragment>
-      <div className="modalOverlayOuter active" onClick={props.onClose} role="presentation" />
-      <div className={`modalOverlayInner active ${props.innerClassName || ""}`}
+      <div className="modalOverlayOuter active" onClick={props.onClose} onKeyDown={this.onKeyDown} role="presentation" />
+      <div className={`${props.innerClassName || "modalOverlayInner active"}`}
```
k88hudson: Why did you switch this? Like why not `modalOverlayInner active ${props.innerClassName || ""}`?

PR #5062, opened 2019-05-23
Bug 1544022 - Turn on prerendering for Discovery Stream

k88hudson:
This patch adjusts pre-rendering such that it works for discovery stream and only renders search


k88hudson:
I am closing this in favour of our new approach, which is to remove prerendering.

PR #5065, opened 2019-05-23
Bug 1553917 - Make eslint compatible with mozilla central

k88hudson:
In this patch, I have:
* removed `esw`, our eslint watch tool (that was not being used for watch anyway) in order to get our environment closer to mc and restore colours in the output on `npm run lint`
* temporarily disable `json`, `promise`, `react-hooks`, and `fetch-options` since they are not yet vendored into MC eslint
* remove linting for `json` (see discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1553919 for removal of json plugin)
*  removed some redundant `eslintignore` paths
* merged the a11y lint task and configuration into the general lint task

To test this, apply [D32373](https://phabricator.services.mozilla.com/D32373) and run `./mach lint browser/components/newtab`.


k88hudson:
@dmose @Mardak I filed additional tickets to remove the json plugin, and add fetch-options/react-hooks to mozilla central (see http://as-bugzy.herokuapp.com/feature/1552007). What do you think about `promise`? Worth keeping or no?

Mardak:
We mostly use async/await instead of plain promises anyway, so we can probably get rid of `eslint-plugin-promise`

Mardak:
Yeah looking at the history, 3 years ago https://github.com/mozilla/activity-stream/commit/89b201de565f185e7ae82a2fb6afa8b09f1db76c was used to make sure we don't silently lose exceptions, but `await Promise.reject()` will throw / be caught.

.eslintignore
```diff
@@ -9,3 +7,5 @@ data/
 bin/prerender.js
 bin/prerender.js.map
 aboutlibrary/content/
+node_modules/
+**/*.json
```
Mardak: Do we need to add these? The latter we just don't include in `--ext` and `node_modules` should automatically be ignored. Looks like we can clean up a bit more for a minimal list right now:

```
bin/prerender.js                                                                                       
data/                                                                                                  
logs/                                                                                                  
prerendered/                                                                                           
vendor/
```
(And prefixing this minimal list with `browser/components/newtab/` seems to be good for `./mach lint` too.)

.eslintrc.jsx-a11y.js
```diff
@@ -3,15 +3,28 @@ module.exports = {
     "jsx-a11y" // require("eslint-plugin-jsx-a11y")
```
Mardak: Oh. and can delete this eslintrc file.

k88hudson:
I merged changes into https://phabricator.services.mozilla.com/D32373 since I figured it would be easier to do all at once

k88hudson:
Closing since this needs to be against central

PR #5067, opened 2019-05-24
Bug 1552809 - Enable Bookmark panel messages to work without translations

piatra:


lib/PanelTestProvider.jsm
```diff
@@ -5,20 +5,39 @@
...
       },
       "close_button": {
-        "tooltiptext": "cfr-doorhanger-bookmark-fxa-close-btn-tooltip",
+        "tooltiptext": {"string_id": "cfr-doorhanger-bookmark-fxa-close-btn-tooltip"},
+      },
+    },
+    "trigger": {"id": "bookmark-panel"},
+  },
+  {
+    "id": "SIMPLE_FXA_BOOKMARK_TEST_NON_FLUENT",
```
k88hudson: Remind me, what makes the test bookmark provider content show up v.s. not show up?

k88hudson:
Code looks good, I'm just wondering what makes the test content show up v.s. not (I noticed the non-translated content show up by default when I ran the patch)

lib/PanelTestProvider.jsm
piatra: The provider [sets targeting to `true`](https://github.com/mozilla/activity-stream/blob/8d5dc9939f14bf819b5646f22565bd6bf760c688/lib/PanelTestProvider.jsm#L31) so that the message always shows up. This was needed because we didn't have any messages in the `cfr-fxa` bucket.
Last week we landed the message in the production bucket so I can turn this off to only show up if you click `Show` in the admin page. I should be able to attach the message to the panel of the currently selected window even if we don't have a way to trigger the panel to drop down.

piatra:
I disabled the `panel_test_provider` messages, now you can trigger them only from ASR Admin.
You still have to use the `Disable Popup Autohide` feature but this is something we can improve later.

I disabled the message from the `cfr-fxa` production bucket. I updated it to include a `string_id` field for when this PR lands.
There was a mistake on my part because without this PR `BookmarkPanelHub` doesn't look for `string_id` in the payload it uses the message field value directly as fluent id. We'll have to serve a different message for 69+ vs 68 but targeting should make that easy.

k88hudson:
Having to disable popup hide isn't ideal but I think it's the best option for now. Thanks!

PR #5073, opened 2019-05-28
Bug 1548697 - Upgrade prod dependencies

piatra:


content-src/asrouter/components/RichText/RichText.jsx
```diff
@@ -1,4 +1,4 @@
-import {Localized} from "fluent-react";
+import {Localized} from "fluent-react/compat";
```
k88hudson: Why is `compat` needed?

piatra:
```
//.../prerender.js:13676
for await (const bundle of iterable) {
        ^^^^^

SyntaxError: Unexpected reserved word
```
I would have expected [the exclude rule](https://github.com/mozilla/activity-stream/blob/685427ad0ef2801120f165850e211a5845cdc53a/webpack.system-addon.config.js#L22) to prevent evaluating/including fluent at build time since it's not really needed for snippets.

piatra: Line is probably(?) from one of [these two places](https://github.com/projectfluent/fluent.js/blob/ab37e1807116f777eb485aa2798ceae4cc17e9a0/fluent-sequence/src/map_async.js#L11-L21).

Mardak: Well if it's only prerender.js having issues, we're getting rid of it in #5081 ! A quick rebase of this on that and removing `/compat` seems to make `npm run bundle` happy

piatra: That's great!

PR #5075, opened 2019-05-28
Bug 1548043 - Changing to personal computer emoji for wider support

gvn:
QA:

- Open default layout
- Confirm that you see a non broken emoji after the "Tech" section header:

Screen Shot 2019-05-28 at 11 01 12 AM


gvn:
Scott since this is a Linux issue can you R?

PR #5078, opened 2019-05-29
Bug 1555390 - Remove prerendering in AboutNewTabService

punamdahiya:


AboutNewTabService.jsm
```diff
@@ -266,11 +253,7 @@ AboutNewTabService.prototype = {
    * allow prerendering.
    */
```
k88hudson: The comment here needs to be updated not to reference prerendering

PR #5080, opened 2019-05-29
Bug 1555391 - Remove PrerenderData.jsm and usages

k88hudson:
This removes `PrerenderData` and its usages, I also modified the prerender jsx to return an empty div since it's going to be removed soon (and to simplify this patch)


k88hudson:
Ok, I've fixed the failure by removing the `global.gActivityStreamPrerenderedState` logic

PR #5082, opened 2019-05-30
Bug 1553375 - Remove react prerendering of placeholders (props, actions, telemetry)

Mardak:
r?@k88hudson and @ncloudioj  Should be okay to just remove the telemetry? Or would it be preferable to send `is_prerendered: false` ?

Reverts parts of https://github.com/mozilla/activity-stream/commit/c43276114e1bec7723f10b4882fb6304b2715d59 and https://github.com/mozilla/activity-stream/commit/5e613f4060564cb206cd28e5e2ecbeda7b1950c0


content-src/activity-stream.jsx
```diff
@@ -11,15 +11,11 @@ const store = initStore(reducers);
 
 new DetectUserSessionStart(store).sendEventOrAddListener();
 
-// If we are starting in a prerendered state, we must wait until the first render
-// to request state rehydration (see Base.jsx). If we are NOT in a prerendered state,
-// we can request it immedately.
 store.dispatch(ac.AlsoToMain({type: at.NEW_TAB_STATE_REQUEST}));
```
Mardak: This used to be dispatched after first render of base. Should we restore that.. with hooks? ;) https://github.com/mozilla/activity-stream/commit/5e613f4060564cb206cd28e5e2ecbeda7b1950c0#diff-1bc118e550611e1215c06a438572d1a4L26

ncloudioj:
The telemetry bit looks good to me! r=nanj.

NB: we won't drop this column in the database, the ETL job will populate it with the default value.

PR #5083, opened 2019-05-30
Bug 1485002 - Fluent migration using in-tree newtab.ftl strings

Mardak:
Not for merge yet. Just getting a couple initial string changes resulting in…

Screen Shot 2019-05-30 at 11 47 40 AM

(with [`intl.l10n.pseudo` set to `accented`](https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_tutorial.html#pseudolocalization))

Search component test still has mountWithIntl/shallowWithIntl because the component used to injectIntl, but tests are still happy with the unnecessary wrapping, so let's just clean up at the end.


locales-src/en-US/strings.properties
```diff
@@ -62,19 +60,6 @@ menu_action_copy_download_link=Copy Download Link
...
 menu_action_remove_download=Remove from History
 
-# LOCALIZATION NOTE (search_button): This is screenreader only text for the
-# search button.
-search_button=Search
-
-# LOCALIZATION NOTE (search_header): Displayed at the top of the panel
-# showing search suggestions. {search_engine_name} is replaced with the name of
-# the current default search engine. e.g. 'Google Search'
-search_header={search_engine_name} Search
```
Mardak: this was unused. not sure if we should track them specially or just remove as we go ?

jar.mn
```diff
@@ -2,6 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+# Just for local testing of in-progress fluent migration
+[localization] en-US.jar:
+  newtab.ftl (./locales-src/newtab.wip)
```
punamdahiya: Why we are adding newtab.ftl here in jar.mn? won't <link rel="localization" takes care of that if we rename newtab.wip to newtab.ftl

locales-src/newtab.wip
```diff
@@ -0,0 +1,10 @@
+newtab-page-title = New Tab
+
+newtab-search-box-search-text = Search
+newtab-search-box-search-button =
+  .title = { newtab-search-box-search-text }
+
+newtab-search-box-search-the-web-text = Search the Web
+newtab-search-box-search-the-web-input =
+  .placeholder = { newtab-search-box-search-the-web-text }
+  .title = { newtab-search-box-search-the-web-text }
```
flodolo: One thing is to do 

```
newtab-search-box-search-the-web-input =
  .placeholder = Search the Web
  .title = { newtab-search-box-search-the-web-input.placeholder }
```

Or referencing a button label from another message. This creates way too many cross-references.

If you ever end up changing the first string, we'll need to invalidate all the strings referencing it, and that might not be wanted or needed. Also, it doesn't help with translation memories or typing less.

jar.mn
Mardak: I suppose if we aren't actually exporting this branch to m-c until we're done with `newtab.ftl` then we could maybe just directly use `.ftl`. It's just that commits to m-c that modify `.ftl` need extra review but this file is in development and not actually exposed to localizers yet

locales-src/newtab.wip
Mardak: So is the suggestion to just do:
```
newtab-search-box-search-the-web-text = Search the Web
newtab-search-box-search-the-web-input =
  .placeholder = Search the Web
  .title = Search the Web
```

Or you're saying only use references among attributes of the same parent?

flodolo: Either this, or just reference within the same message. Both work for me

jar.mn
Mardak: Switched to plain `locales-src/newtab.ftl` and this made me realize if we aren't exporting this branch until we're done, then I'll just copy over locales-src/newtab.ftl into the appropriate m-c/browser/locales/en-US/browser/newtab/newtab.ftl location

locales-src/newtab.wip
Mardak: Removed the references

Mardak:
Added a fluent migration recipe that we can edit and test as we incrementally switch things to fluent. Here's the comment I added to `bug_xxx_newtab.py`:

For now while we're testing, use a recipe with slightly different paths for
testing from activity-stream instead of the usual steps:
https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_migrations.html#how-to-test-migration-recipes
One-time setup starting from activity-stream directory:
```
cd ..
git clone hg::ssh://hg.mozilla.org/l10n/fluent-migration
cd fluent-migration
pip install -e .
cd ..
hg clone https://hg.mozilla.org/l10n/gecko-strings
```
NB: gecko-strings needs to be cloned with mercurial not git-cinnabar
Testing from activity-stream directory:
```
rm -f ../gecko-strings/browser/browser/newtab/newtab.ftl
PYTHONPATH=./bin migrate-l10n bug_xxx_newtab --lang en-US --reference-dir . --localization-dir ../gecko-strings
diff -B locales-src/newtab.ftl ../gecko-strings/browser/browser/newtab/newtab.ftl
```
NB: migrate-l10n will make local commits to gecko-strings
The diff should result in no differences if the migration recipe matches the fluent file.

locales-src/en-US/strings.properties
Pike: I'd remove them when we remove the file.

content-src/components/Search/Search.jsx
```diff
@@ -119,33 +118,32 @@ export class _Search extends React.PureComponent {
...
           type="search" />
         <button
           id="searchSubmit"
           className="search-button"
-          onClick={this.onSearchClick}
-          title={this.props.intl.formatMessage({id: "search_button"})}>
-          <span className="sr-only"><FormattedMessage id="search_button" /></span>
+          data-l10n-id="newtab-search-box-search-button"
+          onClick={this.onSearchClick}>
+          <span className="sr-only" data-l10n-id="newtab-search-box-search-text" />
```
Pike: You don't need a span here, just use the value on the `..-search-button`.

Mardak: Not quite.. or not exactly. This `sr-only` class makes the text invisible except to screen readers, so either we make the fluent string value `…-search-button = <span data-l10n-name="sr-only">Search</span>` which seems unnecessarily complex when... we now have https://bugzilla.mozilla.org/show_bug.cgi?id=1556039 to clean up sr-only usage. In this specific case, I believe we can just get rid of the child span as the `title` attribute already makes it show up to screen readers and doesn't need a different `aria-label`:
image

locales-src/en-US/strings.properties
Mardak: How does removal affect migration recipes? Seems like the recipes need to access the original property value, but I suppose given the cross-channel nature of gecko-strings, mozilla-beta still having the properties would make them available ?

Mardak:
> Do you also want to copy the migration recipe to python/l10n/fluent_migrations/ on buildmc? Might be easier if they were in their own directory?

The plan is to incrementally build out this newtab migration recipe for newtab.ftl and only expose it when we're done with the page's strings. This won't cover all the properties strings as there's some usages outside of the page, e.g., about:preferences, asrouter, others (??) but those will probably have their own migration recipe to their own fluent file.

The process to expose the fluent file and migration.. if it's too confusing to do as part of an export, we can just have a separate bug/phabricator patch dedicated to changing properties/ftl/py files. Then do an export that starts using the fluent strings.

bin/bug_xxx_newtab.py
```diff
@@ -0,0 +1,98 @@
...
+newtab-topsites-edit-header = { COPY(from_path, "topsites_form_edit_header") }
+newtab-topsites-title-label = { COPY(from_path, "topsites_form_title_label") }
+newtab-topsites-title-input =
+    .placeholder = { COPY(from_path, "topsites_form_title_placeholder") }
+
+newtab-topsites-url-label = { COPY(from_path, "topsites_form_url_label") }
+newtab-topsites-url-input =
+    .placeholder = { COPY(from_path, "topsites_form_url_placeholder") }
+newtab-topsites-url-validation = { COPY(from_path, "topsites_form_url_validation") }
+newtab-topsites-delete-history-button = { COPY(from_path, "menu_action_delete") }
```
punamdahiya: newtab-topsites-delete-history-button is used in highlights as well. We can remove topsites from 'newtab-topsites-delete-history-button'

JeaneC: Followup, something like "Delete From History" is used in both a menu and as a button in a topsites confirmation dialog. Previously, we used the id "menu_action_delete" for both. 

Should we try to create a general id like "newtab-delete-from-history-text" or seperate it like "newtab-menu-delete-from-history" and "newtab-topsites-delete-from-history-button", where both ids have the same string "Delete From History"

Mardak: The usage in highlights is for the context menu text? It might be reasonable to keep these separate as arguably one is used as button text and the other as a menu text. But we'll just migrate from the existing multiply-used menu_action_delete

(?)

Mardak: Oh, only just saw your comment after I commented. (I thought github live updates…)

@flodolo is it okay to migrate one source properties string into 2 fluent strings where the string probably should have had 2 entries to begin with?

flodolo: Yes, this is OK. The only case where it wouldn't work is when you copy an existing string in a completely different context.

Mardak:
Rebased `fluent` branch on latest master fixing merge conflicts where #5086 added `aria-haspopup="true"` and #5099 also added them.

Mardak:
@flodolo This PR/recipe is ready to review with a note that we'll manually copy `bug_1485002_newtab.py` to `python/l10n/fluent_migrations` as part of the export containing this merged.

locales-src/onboarding.ftl
```diff
@@ -0,0 +1,127 @@
...
+onboarding-lockwise-passwords-title = Take Your Passwords Everywhere
+onboarding-lockwise-passwords-text2 = Keep the passwords you save secure and easily log in to your accounts with { -lockwise-brand-name }.
+onboarding-lockwise-passwords-button2 = Get the App
+
+onboarding-facebook-container-title = Set Boundaries with Facebook
+onboarding-facebook-container-text2 = { -facebook-container-brand-name } keeps your profile separate from everything else, making it harder for Facebook to target you with ads.
+onboarding-facebook-container-button = Add the Extension
+
+
+## Message strings belonging to the Return to AMO flow
```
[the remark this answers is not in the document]
Mardak: The onboarding.ftl is just a plain copy from mozilla-central, but I included it so it gets updated as part of exporting `locales-src/*.ftl`

Return to AMO is using the string but untouched here as migrations to onboarding.ftl (instead of newtab.ftl) will be handled in https://bugzilla.mozilla.org/show_bug.cgi?id=1556895

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -172,9 +179,7 @@ export class _CollapsibleSection extends React.PureComponent {
               <span className="learn-more-link-wrapper">
                 {learnMore &&
                   <span className="learn-more-link">
-                    <a href={learnMore.link.href}>
-                      <FormattedMessage id={learnMore.link.id} />
-                    </a>
+                    <a href={learnMore.link.href} data-l10n-id={learnMore.link.id}>{learnMore.link.text}</a>
```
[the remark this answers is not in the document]
Mardak: This comes from section definitions:
https://github.com/mozilla/activity-stream/blob/1139bfc682ff544697ba5289a58b7623f3bc72d0/lib/SectionsManager.jsm#L34-L37

This is getting cleaned up in #5127 to take a `message` that can either be raw string or fluent id instead of two separate link properties.

locales-src/newtab.ftl
```diff
@@ -0,0 +1,159 @@
...
+newtab-topsites-preview-button = Preview
+newtab-topsites-add-button = Add
+
+## Top Sites - Delete history confirmation dialog. newtab-confirm-delete-history-p1 appears in the same dialog
+## as newtab-confirm-delete-history-p2.
+newtab-confirm-delete-history-p1 = Are you sure you want to delete every instance of this page from your history?
+# Note: "This action" refers to deleting a page from history.
+newtab-confirm-delete-history-p2 = This action cannot be undone.
+
+newtab-topsites-add-search-engine = Add Search Engine
```
punamdahiya: @Pike  By group comment you mean comments with '##' which will show up in pontoon for all ongoing entries till next group comment is encountered. Reviewed newtab.ftl for comments and looks good except above string which needs its own group comment

punamdahiya: @JeaneC As discussed, lets move this string with top sites dialog and have a new string for 'add search engine' as menu option

locales-src/newtab.ftl
```diff
@@ -0,0 +1,159 @@
...
+
+## Context Menu - Download Menu: These are platform specific strings found in the context menu of an item that has
+## been downloaded. The intention behind "this action" is that it will show where the downloaded file exists on the file
+## system for each operating system.
+newtab-menu-show-file =
+    { PLATFORM() ->
+        [macos] Show in Finder
+        [windows] Open Containing Folder
+        [linux] Open Containing Folder
+       *[other] Show File
```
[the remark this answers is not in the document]
Mardak: The spec https://mozilla.invisionapp.com/share/J6GEU4EHTYC#/screens and comments in the original bug https://bugzilla.mozilla.org/show_bug.cgi?id=1433230 don't seem to see any mentions now (spec might have been updated since then?) of "Show File". So yeah combining them for special mac and everything else seems good.

locales-src/newtab.ftl
```diff
@@ -0,0 +1,159 @@
...
+newtab-section-menu-privacy-notice = Privacy Notice
+
+## Section Headers
+newtab-section-header-topsites = Top Sites
+newtab-section-header-highlights = Highlights
+# Variables:
+#  $provider (String): This is followed by the name of the corresponding content provider.
+newtab-section-header-pocket = Recommended by { $provider }
+
+## Empty Section States: These show when there are no more items in a section. Ex. When there are no more Pocket story reccomendations, in the space where there would have been stories, this is shown instead.
```
Mardak: code spell linting (on m-c) is complaining about "rec**c**ommendations"

locales-src/newtab.ftl
```diff
@@ -0,0 +1,159 @@
...
+
+## Top Sites - Delete history confirmation dialog. newtab-confirm-delete-history-p1 appears in the same dialog
+## as newtab-confirm-delete-history-p2.
+newtab-confirm-delete-history-p1 = Are you sure you want to delete every instance of this page from your history?
+# Note: "This action" refers to deleting a page from history.
+newtab-confirm-delete-history-p2 = This action cannot be undone.
+
+newtab-topsites-add-search-engine = Add Search Engine
+
+## Context Menu - Action Tooltips.
```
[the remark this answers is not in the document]
JeaneC: The period will be good to act as a separation for single-hash comments since pontoon doesn't go to a new line and just places it immediately after.

Mardak:
Rebased on latest master (includes codespell check) and renamed some commits.

locales-src/newtab.ftl
```diff
@@ -0,0 +1,164 @@
...
+
+## Top Sites - Delete history confirmation dialog. 
+
+newtab-confirm-delete-history-p1 = Are you sure you want to delete every instance of this page from your history?
+# "This action" refers to deleting a page from history.
+newtab-confirm-delete-history-p2 = This action cannot be undone.
+
+## Context Menu - Action Tooltips.
+
+# General tool tip for context menus.
```
flodolo: nit: tooltip

Mardak: fixed.

PR #5084, opened 2019-05-30
Start using Prettier (WIP)

dmose:


lib/PlacesFeed.jsm
```diff
@@ -90,21 +108,26 @@ class BookmarksObserver extends Observer {
...
-  onItemRemoved(id, folderId, index, type, uri, guid, parentGuid, source) { // eslint-disable-line max-params
-    if (type === PlacesUtils.bookmarks.TYPE_BOOKMARK &&
-        source !== PlacesUtils.bookmarks.SOURCES.IMPORT &&
-        source !== PlacesUtils.bookmarks.SOURCES.RESTORE &&
-        source !== PlacesUtils.bookmarks.SOURCES.RESTORE_ON_STARTUP &&
-        source !== PlacesUtils.bookmarks.SOURCES.SYNC) {
-      this.dispatch({type: at.PLACES_LINKS_CHANGED});
+
+  onItemRemoved(id, folderId, index, type, uri, guid, parentGuid, source) {
+    // eslint-disable-line max-params
```
Mardak: This also happened for bugzy. Seems like we need to switch to `eslint-disable-next-line` and move the comment.. Although prettier seems to want to insert blank spaces around comments, so moving to the "previous" line might not work either

.eslintrc.js
```diff
@@ -128,14 +121,19 @@ module.exports = {
...
+      "afterBlockComment": true,
+      "beforeLineComment": true,
+      "afterLineComment": true,
+      "allowBlockStart": true,
+      "allowBlockEnd": true,
       "allowObjectStart": true,
-      "beforeBlockComment": true
+      "allowObjectEnd": true,
+      "allowArrayStart": true,
+      "allowArrayEnd": true
```
Mardak: Oh, it's probably not prettier adding the extra lines around comments. What was this change for?

Mardak: I reverted this "lines-around-comment" change from before the `--fix` commit and I don't see any conflicts with prettier for this rule ?

dmose: https://github.com/prettier/eslint-config-prettier#lines-around-comment was the motivator here.  The lines-around-comment stuff still isn't working entirely correctly, and still needs debugging.

lib/PlacesFeed.jsm
dmose: Yeah, I've locally fixed a number of the remaining issues locally by switching to `eslint-disable-next-line`, but there are some conflicts with the `lines-around-comment` stuff yet to be debugged.

.eslintrc.js
Mardak: The note there says to avoid conflicting configuration if `lines-around-comment` requires a blank line inside a block (before a comment). Our configuration doesn't do that for `// line comment`s but does for `/* block comment */` but our commenting style seems to only use block comments at top level or within classes (e.g., for methods), so our usage shouldn't require changes to this rule.

We could just add `"allowBlockStart": true` to be safe (and leave all the other `lines-around-comment` options the same) if someone did try to add a block comment at the beginning of a code block.

dmose: In order to make things work, I ended up with this:

```
    "lines-around-comment": ["error", {
      "allowClassStart": true,
      "beforeBlockComment": false,
      "allowBlockStart": true,
    }],
```

.eslintrc.js
```diff
@@ -128,14 +121,12 @@ module.exports = {
...
-    "indent-legacy": ["error", 2, {"SwitchCase": 1}],
     "init-declarations": 0,
-    "jsx-quotes": [2, "prefer-double"],
     "line-comment-position": 0,
     "lines-around-comment": ["error", {
       "allowClassStart": true,
-      "allowObjectStart": true,
-      "beforeBlockComment": true
+      "beforeBlockComment": false,
+      "allowBlockStart": true,
```
Mardak: I still don't understand why you can't just add the one `"allowBlockStart": true` as suggested in https://github.com/mozilla/activity-stream/pull/5084#discussion_r289472340 instead of deleting `"allowObjectStart": true,` and switching `beforeBlockComment` to `false`.

If I checkout "591bbf5f - Install eslint/prettier package versions matching mozilla-central", add `"allowBlockStart": true,`, run `--fix` and commit, apply "e2e22a0d - Rejigger various eslint-disable-* to play nice with prettier"…

`npm run lint` is happy and there's no differences (other than the `.eslintrc.js` with the one added line).

To be clear, `npm run eslint-check` exits failure 2 listing `- lines-around-comment` but that happens on the latest version of the PR as well.

Mardak: Oh oops. I see where I did differently. I see one difference:
```diff
diff --git a/test/unit/lib/Tokenize.test.js b/test/unit/lib/Tokenize.test.js
--- a/test/unit/lib/Tokenize.test.js
+++ b/test/unit/lib/Tokenize.test.js
@@ -76,16 +76,17 @@ describe("TF-IDF Term Vectorizer", () => {
         expected: {
           easy: [269, 0.6786999710383944],
           care: [153, 0.7344156515982504],
         },
       },
       {
         input: "这个空间故意留空。",
         expected: {
+
           /* This space is left intentionally blank. */
         },
       },
     ];
     let checkTokenGeneration = tc => {
       describe(`${tc.input} should have only vocabulary tokens`, () => {
         let actual = toksToTfIdfVector(tokenize(tc.input), vocab_idfs);
 
```

Mardak: Oh no nevermind again. This is with your deleting `"allowObjectStart": true`

dmose: So I just talked to Victor, and it sounds like he's going to be disabling all rules that eslint-config-check flags as "possibly conflicting" by eslint-config-prettier at the top-level, which sounds reasonable to me.  So I'm inclined to just get rid of `lines-around-comment` entirely to keep us in line with that.  Any thoughts?

Mardak: To be clear, you've already effectively turned off the rule by by setting `"beforeBlockComment": false,`

If we don't want any conflicts showing up, then removing the rule is the only thing we can do then.

dmose: Right; I guess what I was really trying to ask was "is there any compelling reason I should push back on this proposed policy"?

Mardak: Looking at when we enabled lines-around-comment https://github.com/mozilla/activity-stream/commit/b330de028ff62d1a4e6794f9e09a6c5cacfa51fd The main change was making our comment types more consistent -- basically block comments above blocks/functions and inline comments inline.

We can always just `nit` if someone doesn't add a line before block comments or uses the different type of comment.

dmose: The irony being that prettier is all about getting rid of nits.  Another alternative would be just not to worry about comment-type consistency.

Anyway, I'll go ahead and get rid of lines-around-comment.

.eslintrc.js
```diff
@@ -32,7 +33,10 @@ module.exports = {
     "plugin:mozilla/recommended", // require("eslint-plugin-mozilla")
     "plugin:mozilla/browser-test",
     "plugin:mozilla/mochitest-test",
-    "plugin:mozilla/xpcshell-test"
+    "plugin:mozilla/xpcshell-test",
+    "prettier",
```
dmose: Note to self: I think this is superfluous and wants to go away.

dmose: Although maybe not: https://github.com/prettier/eslint-config-prettier#installation

Mardak:
Some changes here won't be necessary when rebased on #5105 updating eslint-plugin-mozilla@1.3.0 which requires `prettier` but we'll want to remove the `.prettierignore` here to enable.

dmose:
Closed in favor of #5114 and https://github.com/dmose/activity-stream/pull/34

PR #5085, opened 2019-06-03
Bug 1542863 - Adding feed load fail fallback.

ScottDowne:
To test:

1. You need to clone this https://github.com/ScottDowne/fetch-delay locally, run npm install, then node start.js
2. Set your browser.newtabpage.activity-stream.discoverystream.endpoints pref to "https,http"
3. Set your browser.newtabpage.activity-stream.discoverystream.config pref to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":false,"hardcoded_layout":false,"personalized":false,"adserver":false,"layout_endpoint":"https://gist.githubusercontent.com/ScottDowne/2c5d6774ed644c3f8bcbb00719b8f2b5/raw/f3bb1b513284622d66fd6c4ccdb9b98fb9c3124f/layout-failures" }`
4. load a new tab.

The first feed in this tab should always fail, so you can keep testing the failure state over and over.
Ever feed after that fails on a 20% less likely chance, so the second feed fails 80% of the time, third, 60%, etc.
It may pass on the first fetch. If you want to retry, you need to reset your browser.newtabpage.activity-stream.discoverystream.config off and on again.


gvn:
Note: this needs a rebase on latest master for at least OSX

content-src/components/DiscoveryStreamComponents/DSEmptyState/DSEmptyState.jsx
```diff
@@ -1,12 +1,84 @@
...
+      }, 300);
+
+      this.props.dispatch(ac.OnlyToMain({type: at.DISCOVERY_STREAM_RETRY_FEED, data: {feed}}));
+    }
+  }
+
+  renderButton() {
+    if (this.props.status === "waiting" || this.state.waiting) {
+      return (
+        <button className="try-again-button waiting">Loading...</button>
```
gvn: Do these strings need to be variables for localization?

ScottDowne: There are more strings in this file that are hard coded, I think if this is an issue, we can punt it to another bug.

Mardak: The other english string was landed for 68 with a comment of "english only for 68 but localized for 69+"
https://github.com/mozilla/activity-stream/pull/4877#discussion_r273171475

So we definitely want a P1 bug if we expect this to be more than just english for 69 (although I'm not sure if that timeline matches up since the original comment).

Mardak: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1556862

PR #5087, opened 2019-06-03
Bug 1552932 - [Top Sites] Focus and Hover UX updates

gvn:
Testing: Compare build & changes requested to [spec](https://www.figma.com/file/0xZkmT86rSXbGRB4eIvrn0ts/New-Tab-Spec-for-Fx-v68?node-id=383%3A1)

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/TopSites/_TopSites.scss
+++ b/content-src/components/DiscoveryStreamComponents/TopSites/_TopSites.scss
@@ -11,6 +11,18 @@
 
       .top-site-inner > a:-moz-any(.active, :focus) .tile {
         @include ds-fade-in;
+
+        @include dark-theme-only {
+          @include ds-fade-in($blue-40-40);
+        }
+      }
+
+      .top-site-inner > a:-moz-any(:hover) .tile {
+        @include ds-fade-in($grey-30);
+
+        @include dark-theme-only {
+          @include ds-fade-in($grey-60);
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


content-src/components/TopSites/_TopSites.scss
```diff
@@ -110,6 +110,10 @@ $hover-transition-duration: 150ms;
   padding: 0 $half-base-gutter;
   display: inline-block;
 
+  .top-site-icon.rich-icon {
+    border-radius: 4px;
```
punamdahiya: Any reason why we are changing corder-radius here and not in variable $top-sites-border-radius used in line 137 below

content-src/components/DiscoveryStreamComponents/TopSites/_TopSites.scss
```diff
@@ -11,6 +11,18 @@
...
       .top-site-inner > a:-moz-any(.active, :focus) .tile {
         @include ds-fade-in;
+
+        @include dark-theme-only {
+          @include ds-fade-in($blue-40-40);
+        }
+      }
+
+      .top-site-inner > a:-moz-any(:hover) .tile {
+        @include ds-fade-in($grey-30);
```
punamdahiya: With this there is slight flicker  because of border-radius changing from 6px to 4px on hover. Changing variable $top-sites-border-radius to 4px should fix it

content-src/components/TopSites/_TopSites.scss
gvn: $top-sites-border-radius is 6px

gvn:
Tweaked!

PR #5088, opened 2019-06-04
Bug 1554734 - Also support context-menu arrow key navigation 

AllegroFox:
added ArrowUp and ArrowDown listeners to the onKeyDown event on context menus; handled focus with DOM targeting. May try to switch to React refs to make the function a little more resilient.

### DIFF
--- a/content-src/components/ContextMenu/ContextMenu.jsx
+++ b/content-src/components/ContextMenu/ContextMenu.jsx
@@ -62,6 +62,17 @@ export class ContextMenuItem extends React.PureComponent {
     this.props.option.onClick();
   }
 
+  // This selects the correct node based on the key pressed
+  focusSibling(target, key) {
+    const parent = target.parentNode;
+    const closestSiblingSelector = (key === "ArrowUp") ? "previousSibling" : "nextSibling";
+    if (parent[closestSiblingSelector].firstElementChild) {
+      parent[closestSiblingSelector].firstElementChild.focus();
+    } else {
+      parent[closestSiblingSelector][closestSiblingSelector].firstElementChild.focus();
+    }
+  }
+
   onKeyDown(event) {
     const {option} = this.props;
     switch (event.key) {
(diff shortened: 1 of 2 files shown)


content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -84,7 +100,7 @@ export class ContextMenuItem extends React.PureComponent {
     const {option} = this.props;
     return (
       <li role="menuitem" className="context-menu-item" >
-        <button className={option.disabled ? "disabled" : ""} onClick={this.onClick} onKeyDown={this.onKeyDown} tabIndex="0" >
+        <button className={option.disabled ? "disabled" : ""} tabIndex="0" onClick={this.onClick} onKeyDown={this.onKeyDown}>
```
k88hudson: what's the reason for this change?

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -62,6 +62,17 @@ export class ContextMenuItem extends React.PureComponent {
     this.props.option.onClick();
   }
 
+  // This selects the correct node based on the key pressed
+  focusSibling(target, key) {
+    const parent = target.parentNode;
+    const closestSiblingSelector = (key === "ArrowUp") ? "previousSibling" : "nextSibling";
+    if (parent[closestSiblingSelector].firstElementChild) {
```
k88hudson: I think you also need to return early if `parent[closestSiblingSelector]` is falsey for the first/last child case

content-src/components/ContextMenu/ContextMenu.jsx
AllegroFox: I'm not sure, I may have removed the tabIndex to test something and then put it back :)

content-src/components/ContextMenu/ContextMenu.jsx
AllegroFox: `if (!parent[closestSiblingSelector]) {
      return;
    } else if (parent[closestSiblingSelector].firstElementChild) {
      parent[closestSiblingSelector].firstElementChild.focus();
    } else {
      parent[closestSiblingSelector][closestSiblingSelector].firstElementChild.focus();
    }`

Does that look about right?

AllegroFox: The behaviour seems to be the same

k88hudson: yep that's good, I see an error being thrown otherwise

PR #5090, opened 2019-06-05
Bug 1523403 - Test coverage for DS front-end

punamdahiya:


test/unit/content-src/components/DiscoveryStreamComponents/DSLinkMenu.test.jsx
```diff
@@ -13,6 +13,28 @@ describe("<DSLinkMenu>", () => {
...
   });
 
+  it("on Menu Update", () => {
+    const dsLinkMenu = mountWithIntl(<DSLinkMenu />);
+    dsLinkMenu.setState({showContextMenu: true});
+
+    dsLinkMenu.instance().onMenuUpdate(false);
+    dsLinkMenu.update();
+
+    // Should check grand parent element is missing active, how?
```
punamdahiya: @Mardak is there any way to get div (e.g.  .ds-card) containing DSLinkMenu here. Thought should ask you before moving test to check for "active" and "last-item" class in DSCard component

Mardak: The behavior of DSLinkMenu modifying ancestors does look strange, but ignoring that, you can use `dsLinkMenu.getDOMNode().parentNode.classList` to check if "active" is in there (for the "onMenuShow" test). For here "onMenuUpdate", you can do the opposite by adding the class name after mounting then checking it's removed.

test/unit/content-src/components/DiscoveryStreamComponents/TopSites.test.jsx
```diff
@@ -23,3 +27,20 @@ describe("Discovery Stream <TopSites>", () => {
...
+describe.only("DSCard DS Top Sites", () => {
+  let wrapper;
+  let DEFAULT_PROPS = {
+    header: "test",
+    isFixed: true,
+  };
+
+  beforeEach(() => {
+    const store = createStore(combineReducers(reducers), INITIAL_STATE);
+    wrapper = mountWithIntl(<Provider store={store}><TopSites {...DEFAULT_PROPS} /></Provider>);
```
punamdahiya: @Mardak Discovery stream component has 75% test coverage calling out connect as uncovered block

Mounting Topsites above is giving error here https://github.com/mozilla/activity-stream/blob/master/content-src/components/TopSites/TopSite.jsx#L418
with this.props.TopSitesRows as undefined. Looks like its set  https://github.com/mozilla/activity-stream/blob/master/content-src/components/TopSites/TopSites.jsx#L172

Is there any way to set state.Prefs.values.topSitesRows from tests to override this error? Thanks

top-site-code-coverage

punamdahiya:
@Mardak @ScottDowne PR increases test coverage for DS components to 90%. Please review. Thanks!

punamdahiya:
@gvn if you have bandwidth, please review. Thanks!

test/unit/content-src/components/DiscoveryStreamComponents/TopSites.test.jsx
Mardak: Just to note out conversation from slack, we need to import the 2 differently exported components to test the connected version vs plain component.

punamdahiya: @Mardak quick clarification on the feedback, do we want separate unit tests for connected (TopSites) and plain (_TopSites) component. My understanding mounting connected component loads and run tests on plain component, and showing test coverage 100% for TopSites component

punamdahiya:
Synched up with @Mardak and TopSites testing of the connected version also tests the wrapped unconnected version and looks good.

PR #5091, opened 2019-06-06
(nobug) - Open the bookmark panel when previewing the FxA message

piatra:
This uses internal StarUI properties but I think this is ok since it's for internal use anyway.


piatra:
Going to merge this into another follow up that refactors some of the `BookmarkPanelHub`.

PR #5095, opened 2019-06-06
Bug 1556887 - Convert strings in Top Sites dialog to use fluent (header and body)

JeaneC:
Made changes in the TopSites dialog to move to Fluent. Did not touch footer components because they get reused in another component.

### DIFF
--- a/content-src/components/CollapsibleSection/_CollapsibleSection.scss
+++ b/content-src/components/CollapsibleSection/_CollapsibleSection.scss
@@ -9,6 +9,7 @@
     font-weight: bold;
     margin: 0;
 
+    &.grey-title,
     span {
       color: var(--newtab-section-header-text-color);
       display: inline-block;
--- a/content-src/components/TopSites/TopSiteForm.jsx
+++ b/content-src/components/TopSites/TopSiteForm.jsx
@@ -164,13 +164,12 @@ export class TopSiteForm extends React.PureComponent {
       customScreenshotUrl && this.props.previewUrl === this.cleanUrl(customScreenshotUrl);
 
     if (!this.state.showCustomScreenshotForm) {
-      return (<A11yLinkButton onClick={this.onEnableScreenshotUrlForm} className="enable-custom-image-input">
-                <FormattedMessage id="topsites_form_use_image_link" />
-              </A11yLinkButton>);
+      return (<A11yLinkButton onClick={this.onEnableScreenshotUrlForm} className="enable-custom-image-input"
+          data-l10n-id="newtab_topsites_use_image_link" />);
     }
     return (<div className="custom-image-input-container">
       <TopSiteFormInput
-        errorMessageId={requestFailed ? "topsites_form_image_validation" : "topsites_form_url_validation"}
+        errorMessageId={requestFailed ? "newtab_topsites_image_validation" : "newtab_topsites_url_validation"}
         loading={isLoading}
         onChange={this.onCustomScreenshotUrlChange}
         onClear={this.onClearScreenshotInput}
@@ -178,9 +177,8 @@ export class TopSiteForm extends React.PureComponent {
         typeUrl={true}
         value={customScreenshotUrl}
         validationError={validationError}
-        titleId="topsites_form_image_url_label"
-        placeholderId="topsites_form_url_placeholder"
-        intl={this.props.intl} />
+        titleId="newtab_topsites_image_url_label"
+        placeholderId="newtab_topsites_url_placeholder" />
     </div>);
   }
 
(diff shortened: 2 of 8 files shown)


locales-src/newtab.ftl
```diff
@@ -19,3 +19,20 @@ newtab-search-box-search-the-web-text = Search the Web
 newtab-search-box-search-the-web-input =
     .placeholder = Search the Web
     .title = Search the Web
+
+## Top Sites form dialog
+
+newtab_topsites_add_header = New Top Site
+newtab_topsites_edit_header = Edit Top Site
+newtab_topsites_title_label = Title
+newtab_topsites_title_placeholder =
```
Mardak: nit: fluent convention seems to be `-` dashes instead of `_` underscores

flodolo: Indeed. Also, [this doc](https://docs.google.com/document/d/12QdVuKgJpZ39YZYt5sOZWqAGOXYyw_gfl4hbgY9AX0A/edit#) might be a useful pointer (I'm starting to share it around before landing it in-tree)

content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -205,25 +203,23 @@ export class TopSiteForm extends React.PureComponent {
       <form className="topsite-form" onSubmit={onSubmit}>
         <div className="form-input-container">
           <h3 className="section-title">
-            <FormattedMessage id={showAsAdd ? "topsites_form_add_header" : "topsites_form_edit_header"} />
+            <span data-l10n-id={showAsAdd ? "newtab_topsites_add_header" : "newtab_topsites_edit_header"} />
```
Mardak: We shouldn't need additional `<span>` children for text parents, so the `data-l10n-id` should be able to go directly on the `<h3>` here

locales-src/newtab.ftl
```diff
@@ -19,3 +19,20 @@ newtab-search-box-search-the-web-text = Search the Web
...
+    .placeholder = Enter a title
+
+newtab_topsites_url_label = URL
+newtab_topsites_url_placeholder =
+    .placeholder = Type or paste a URL
```
Mardak: Not sure if these "-placeholder" parent ids would be better as…

```ftl
newtab-topsites-url-input =
  .placeholder = …
```

(I do see both patterns (and more 😛 ) if you click into various files and look at the parent id https://searchfox.org/mozilla-central/search?q=.placeholder&path=.ftl )

content-src/components/TopSites/TopSiteForm.jsx
JeaneC: Normally true but it would remove the css selector `.collapsible-section .section-title span` ([here](https://searchfox.org/mozilla-central/source/browser/components/newtab/css/activity-stream-windows.css#1577)) that makes this gray instead of black so I chose to keep it here to preserve that.

locales-src/newtab.ftl
JeaneC: Ah.... Definitely sounds redundant now. I'll switch to the input suggestion.

content-src/components/TopSites/TopSiteForm.jsx
punamdahiya: @Mardak where ever possible may be we should updates css to stop using span

content-src/components/TopSites/TopSiteFormInput.jsx
```diff
@@ -60,22 +59,22 @@ export class TopSiteFormInput extends React.PureComponent {
     const {typeUrl} = this.props;
     const {validationError} = this.state;
 
-    return (<label><FormattedMessage id={this.props.titleId} />
+    return (<label>
+      <span data-l10n-id={this.props.titleId} />
```
punamdahiya: Changing here to <label data-l10n-id={this.props.titleId}> breaks the layout. @Mardak feel free to chime if there is a way to  remove span tag here.

punamdahiya:
Task cluster failure is not related to the PR

content-src/components/TopSites/TopSiteFormInput.jsx
Mardak: The way this is done is setting a `data-l10n-name` on the *inline* children and putting empty nodes in the fluent string:
https://github.com/mozilla/activity-stream/blob/8b0e7e99f392bf4bbbfe2fe03878ef205befb050/content-src/asrouter/templates/Trailhead/Trailhead.jsx#L280-L284
along with
```ini
onboarding-join-form-legal = By proceeding, you agree to the <a data-l10n-name="terms">Terms of Service</a> and <a data-l10n-name="privacy">Privacy Notice</a>.
```

So in this case,
```ini
newtab-topsites-title-label = Title <div data-l10n-name="input"></div>
```
```jsx
<label>
  <div data-l10n-name="input" className=…>
    <input … />
  </div>
</label>
```
 
However, the child here is actually a block and the text spacing doesn't actually matter, so we don't actually want to do the above. 😉 So we're good with the extra span in this case. I suppose we could make the `<label for=…>` have no children, but then we would need to reference the `<input>` by id.

Some more details here https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_migrations.html#concatenating-strings

content-src/components/TopSites/TopSiteForm.jsx
```diff
@@ -164,23 +164,21 @@ export class TopSiteForm extends React.PureComponent {
       customScreenshotUrl && this.props.previewUrl === this.cleanUrl(customScreenshotUrl);
 
     if (!this.state.showCustomScreenshotForm) {
-      return (<A11yLinkButton onClick={this.onEnableScreenshotUrlForm} className="enable-custom-image-input">
-                <FormattedMessage id="topsites_form_use_image_link" />
-              </A11yLinkButton>);
+      return (<A11yLinkButton onClick={this.onEnableScreenshotUrlForm} className="enable-custom-image-input"
+          data-l10n-id="newtab-topsites-use-image-link" />);
```
Mardak: This is a custom react component, so we'll most likely need to specially handle this prop. It was working before because `<A11yLinkButton>` nests the children.

Mardak: https://github.com/mozilla/activity-stream/blob/baaac1f8301678104e5b364fde465fe1982d56e5/content-src/components/A11yLinkButton/A11yLinkButton.jsx#L10-L11

Mardak: https://github.com/mozilla/activity-stream/blob/baaac1f8301678104e5b364fde465fe1982d56e5/content-src/components/A11yLinkButton/A11yLinkButton.jsx#L10-L11

Oh. I guess it just spreads the `props` which.. I believe we wanted to avoid on plain html elements? @AllegroFox @k88hudson ?

Mardak: Although……… ha. I suppose in this case, fluent-dom will just override the children or any dangerously set html… But safer to not even set in the first place !

content-src/components/CollapsibleSection/_CollapsibleSection.scss
```diff
@@ -9,6 +9,7 @@
     font-weight: bold;
     margin: 0;
 
+    &.grey-title,
     span {
```
Mardak: Is the span selector used for anything now?

JeaneC: This is used once more in the Search Dialog component. Deleting it in the next PR

Mardak: Ah right, this is collapsible section styles not just top sites.

PR #5096, opened 2019-06-06
Bug 1556039 - Clean up sr-only usage by removing or converting to aria-label

AllegroFox:
I'll probably write some better tests in future, but these solve the coverage issue that arose from removing the FormattedMessage component in a few places.

### DIFF
--- a/content-src/components/Card/Card.jsx
+++ b/content-src/components/Card/Card.jsx
@@ -245,10 +245,8 @@ export class _Card extends React.PureComponent {
           </div>
         </div>
       </a>
-      {!props.placeholder && <button aria-haspopup="true" className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
-        onClick={this.onMenuButtonClick}>
-        <span aria-haspopup="true" className="sr-only">{`Open context menu for ${link.title}`}</span>
-      </button>}
+      {!props.placeholder && <button aria-haspopup="true" aria-label={`Open context menu for ${link.title}`} className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
+        onClick={this.onMenuButtonClick} />}
       {isContextMenuOpen &&
         <LinkMenu
           dispatch={dispatch}
--- a/content-src/components/Search/Search.jsx
+++ b/content-src/components/Search/Search.jsx
@@ -119,6 +119,7 @@ export class _Search extends React.PureComponent {
       {!this.props.handoffEnabled &&
       <div className="search-inner-wrapper">
         <label htmlFor="newtab-search-text" className="search-label">
+          {/* Remove this sr-only span? */}
           <span className="sr-only"><FormattedMessage id="search_web_placeholder" /></span>
         </label>
         <input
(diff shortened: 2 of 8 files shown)


Mardak:
I believe @k88hudson's intent was that these changes would happen on the `fluent` branch https://github.com/mozilla/activity-stream/tree/fluent

This would avoid adding new `formatMessage` calls

AllegroFox:
> I believe @k88hudson's intent was that these changes would happen on the `fluent` branch https://github.com/mozilla/activity-stream/tree/fluent
> 
> This would avoid adding new `formatMessage` calls

I'll check with her! thanks :)

content-src/components/Card/Card.jsx
```diff
@@ -245,10 +245,8 @@ export class _Card extends React.PureComponent {
           </div>
         </div>
       </a>
-      {!props.placeholder && <button aria-haspopup="true" className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
-        onClick={this.onMenuButtonClick}>
-        <span aria-haspopup="true" className="sr-only">{`Open context menu for ${link.title}`}</span>
-      </button>}
+      {!props.placeholder && <button aria-haspopup="true" aria-label={`Open context menu for ${link.title}`} className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
```
Mardak: We should fix this up to use `context_menu_button_sr=Open context menu for {title}` (but migrated to fluent 😉 ) instead of hardcoded.

content-src/components/Search/Search.jsx
```diff
@@ -119,6 +119,7 @@ export class _Search extends React.PureComponent {
       {!this.props.handoffEnabled &&
       <div className="search-inner-wrapper">
         <label htmlFor="newtab-search-text" className="search-label">
+          {/* Remove this sr-only span? */}
```
Mardak: Yeah, I believe this whole `label` can be removed. devtools accessibility seems to be happy enough with the actual `input`'s `title`

content-src/components/Search/Search.jsx
```diff
@@ -132,9 +133,8 @@ export class _Search extends React.PureComponent {
           id="searchSubmit"
           className="search-button"
           onClick={this.onSearchClick}
-          title={this.props.intl.formatMessage({id: "search_button"})}>
-          <span className="sr-only"><FormattedMessage id="search_button" /></span>
-        </button>
+          title={this.props.intl.formatMessage({id: "search_button"})}
+          aria-label={this.props.intl.formatMessage({id: "search_button"})} />
```
Mardak: The `button` with `title` seems to be good in devtools, so no `aria-label` needed here

Mardak:
At a glance, it seems like most (all?) places we're touching here already have a `title` attribute, so if those all look okay in devtools accessibility, maybe we just need to remove the extra span ??

content-src/components/Search/Search.jsx
AllegroFox: I chose to double up title and aria-label wherever there was already a title because a11y docs seem to show that different screen readers will use different things - and it doesn't seem like having them both duplicates anything in the DOM. I can always change it though!

https://www.deque.com/blog/text-links-practices-screen-readers/

Mardak: Thanks. The comparison table:

D. Both Aria-label and title present besides anchor text
Aria-label + title
Anchor text + title
Anchor text + aria-label
Aria-label + title

So with a duplicate title and aria-label, some screen readers will just say the same thing twice? Although arguably maybe we should be setting them to be different things ?

Although looks like VO on OSX just doesn't read `title`s at all so it alone wouldn't be sufficient anyway.

AllegroFox: Yeah - I wasn't sure which case we wanted to optimize for, so I went for overkill :P The tricky thing is the cases where we have buttons that are icons only, or are hidden until focused, so we have no anchor text to be read - I wanted to make sure there was always something that would be picked up.

PR #5098, opened 2019-06-07
Bug 1556887 - Fixes in other topsites dialogs (Add Site, Add Search Engine, Delete From History)

JeaneC:
Made the migration in the "Delete From History Dialog" and the "Add Search Engine" Dialog

The "Cancel" string is reused in all 3 topsite dialogs.


locales-src/newtab.ftl
```diff
@@ -36,3 +36,15 @@ newtab-topsites-url-validation = Valid URL required
...
+
+newtab-topsites-cancel-button = Cancel
+newtab-topsites-delete-history-button = Delete from History
+newtab-topsites-save-button = Save
+newtab-topsites-preview-button = Preview
+newtab-topsites-add-button = Add
+
+newtab-confirm-delete-history-p1 = Are you sure you want to delete every instance of this page from your history?
+# p1 and p2 are in the same delete dialog. "This action" refers to deleting a page from history.
+newtab-confirm-delete-history-p2 = This action cannot be undone.
```
Mardak: Fluent allows `<br/>`s. @flodolo should we just make this one string "…from your history?<br/><br/>This action…" ?

Doesn't look like the text is styled differently for the first vs second `<p>`:
image

Mardak: Although it does seem a bit odd to expose the double `<br/>`…

flodolo: I'd keep them separate (as they currently are).

locales-src/newtab.ftl
```diff
@@ -36,3 +36,15 @@ newtab-topsites-url-validation = Valid URL required
...
 newtab-topsites-image-validation = Image failed to load. Try a different URL.
+
+newtab-topsites-cancel-button = Cancel
+newtab-topsites-delete-history-button = Delete from History
+newtab-topsites-save-button = Save
+newtab-topsites-preview-button = Preview
+newtab-topsites-add-button = Add
+
+newtab-confirm-delete-history-p1 = Are you sure you want to delete every instance of this page from your history?
+# p1 and p2 are in the same delete dialog. "This action" refers to deleting a page from history.
```
punamdahiya: p1 and p2 can be confusing,  I recommend updating comment here to 
# LOCALIZATION NOTE (newtab-confirm-delete-history-p2): this string is displayed in
# the same dialog as newtab-confirm-delete-history-p1. "This action" refers to deleting a page from history.

punamdahiya: It seems we are flexible and not using [LOCALIZATION NOTE (<string id>)](https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/preferences/preferences.ftl#85) in fluent files and should be ok changing comment to use complete string id instead of p1 and p2

test/unit/content-src/components/ConfirmDialog.test.jsx
```diff
@@ -1,7 +1,7 @@
 import {actionCreators as ac, actionTypes as at} from "common/Actions.jsm";
 import {_ConfirmDialog as ConfirmDialog} from "content-src/components/ConfirmDialog/ConfirmDialog";
-import {FormattedMessage} from "react-intl";
 import React from "react";
+import {shallow} from "enzyme";
 import {shallowWithIntl} from "test/unit/utils";
```
punamdahiya: We can remove shallowWithIntl from remaining ConfirmDialog tests.

punamdahiya:
Mochitest passing on local. Task cluster failures doesn't look related.

PR #5099, opened 2019-06-07
Bug-1556039 - Removed sr-only spans; updated fluent strings to include aria-label where needed.

AllegroFox:


AllegroFox:
I'm grabbing @JeaneC on this because he helped me unbreak my Fluent build! +1 successful Zoom meeting :)

content-src/components/Card/Card.jsx
```diff
@@ -245,10 +245,8 @@ export class _Card extends React.PureComponent {
           </div>
         </div>
       </a>
-      {!props.placeholder && <button className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
-        onClick={this.onMenuButtonClick}>
-        <span className="sr-only">{`Open context menu for ${link.title}`}</span>
-      </button>}
+      {!props.placeholder && <button aria-haspopup="true" aria-label={`Open context menu for ${link.title}`} className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
```
Mardak: Might as well fix this with `context_menu_button_sr`

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -184,12 +184,9 @@ export class _CollapsibleSection extends React.PureComponent {
             <button
               className="context-menu-button icon"
               title={this.props.intl.formatMessage({id: "context_menu_title"})}
+              aria-label={this.props.intl.formatMessage({id: "context_menu_title"})}
               onClick={this.onMenuButtonClick}
-              ref={this.setContextMenuButtonRef}>
-              <span className="sr-only">
-                <FormattedMessage id="section_context_menu_button_sr" />
```
Mardak: Is this an intended change from `section_context_menu_button_sr` to `context_menu_title` ?

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
```diff
@@ -43,18 +43,14 @@ export class _DSLinkMenu extends React.PureComponent {
...
     return (<div>
       <button ref={this.contextMenuButtonRef}
               className="context-menu-button icon"
               title={this.props.intl.formatMessage({id: "context_menu_title"})}
-              onClick={this.onMenuButtonClick}>
-        <span className="sr-only">
-          <FormattedMessage id="context_menu_button_sr" values={{title}} />
-        </span>
-      </button>
+              aria-label={this.props.intl.formatMessage({id: "context_menu_title"})}
```
Mardak: Is this an intended change from `context_menu_button_sr` to `context_menu_title` ?

content-src/components/Search/Search.jsx
```diff
@@ -130,9 +130,7 @@ export class _Search extends React.PureComponent {
           id="searchSubmit"
```
Mardak: Are we intending to keep the `<label><span sr-only newtab-search-box-search-the-web-text>` above? Looks like it's the last sr-only, so if it's removed, we can also remove the sr-only style.

content-src/components/Search/Search.jsx
```diff
@@ -144,7 +142,7 @@ export class _Search extends React.PureComponent {
             onClick={this.onSearchHandoffClick}
             tabIndex="-1">
             <div className="fake-textbox" data-l10n-id="newtab-search-box-search-the-web-text" />
-            <input type="search" className="fake-editable" tabIndex="-1" aria-hidden="true" onDrop={this.onSearchHandoffDrop} onPaste={this.onSearchHandoffPaste} />
+              <input type="search" className="fake-editable" tabIndex="-1" aria-hidden="true" onDrop={this.onSearchHandoffDrop} onPaste={this.onSearchHandoffPaste} />
```
Mardak: What's this intent for? This is actually a sibling of the `<div fake-textbox>`

content-src/components/TopSites/TopSite.jsx
```diff
@@ -281,11 +281,7 @@ export class TopSite extends React.PureComponent {
     const title = link.label || link.hostname;
     return (<TopSiteLink {...props} onClick={this.onLinkClick} onDragEvent={this.props.onDragEvent} className={`${props.className || ""}${isContextMenuOpen ? " active" : ""}`} title={title}>
         <div>
-          <button className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})} onClick={this.onMenuButtonClick}>
-            <span className="sr-only">
-              <FormattedMessage id="context_menu_button_sr" values={{title}} />
-            </span>
-          </button>
+          <button aria-haspopup="true" className="context-menu-button icon" aria-label={this.props.intl.formatMessage({id: "context_menu_title"})} title={this.props.intl.formatMessage({id: "context_menu_title"})} onClick={this.onMenuButtonClick} />
```
Mardak: This removes the `title` from the screenreader -- intended?

test/unit/content-src/components/DiscoveryStreamComponents/DSCard.test.jsx
```diff
@@ -64,4 +64,8 @@ describe("<DSCard>", () => {
   it("should render DSLinkMenu", () => {
     assert.equal(wrapper.children().at(1).type(), DSLinkMenu);
   });
+
+  it("should start with no .active class", () => {
+    assert.equal(wrapper.find(".active").length, 0);
```
Mardak: Are the test changes intentional or accidental rebasing ?

content-src/components/CollapsibleSection/CollapsibleSection.jsx
AllegroFox: I had wanted the titles and aria-labels to have consistent content, but `section_context_menu_button_sr` is actually probably a better choice for both of them, since the message is more detailed. I'll switch it over in all these cases.

content-src/components/Search/Search.jsx
AllegroFox: I wasn't sure if it should be removed since it was the only instance of <label> instead of <span>, but I'll try a few things and see if I can safely remove it.

content-src/components/Search/Search.jsx
AllegroFox: A typo!

test/unit/content-src/components/DiscoveryStreamComponents/DSCard.test.jsx
AllegroFox: The test changes are intentional, just to bring up the coverage since the <FormattedMessage> component was removed. I'll probably open a new bug to write some better tests soon.

content-src/components/CollapsibleSection/CollapsibleSection.jsx
AllegroFox: So `section_context_menu_button_sr` doesn't seem to pull the {title} prop through correctly - I get a lot of "open context menu for {title}." , even when I give the component a hardcoded prop. Is there somewhere else it needs to be set as well?

content-src/components/Search/Search.jsx
AllegroFox: Looks like that label was actually redundant - I'll go ahead and remove it, as well as the sr-only style.

content-src/components/TopSites/TopSite.jsx
AllegroFox: My understanding is that the screen reader will pick up either `title` or `aria-label` as long as they're both present, so as long as we're seeing one or the other it should be fine

Mardak: Sorry. `title` was ambiguous. I was referring to the js value `values={{title}}` not the html attribute.

AllegroFox: ah okay. Is `values={{title}}` able to be interpreted by anything besides the `<FormattedMessage>` component? @k88hudson and I had gone through and taken them out.

Mardak: Should be able to pass in a values object https://github.com/formatjs/react-intl/blob/master/docs/API.md#formatmessage

AllegroFox: ahh that makes sense!

AllegroFox:
> Why the comment instead of just destructuring?

I tried destructuring, and it seemed to break the message interpolation (the variable would just become blank.) Not sure why that's the case here.

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
```diff
@@ -43,18 +43,15 @@ export class _DSLinkMenu extends React.PureComponent {
     const {index, dispatch} = this.props;
     const isContextMenuOpen = this.state.showContextMenu && this.state.activeCard === index;
     const TOP_STORIES_CONTEXT_MENU_OPTIONS = ["CheckBookmarkOrArchive", "CheckSavedToPocket", "Separator", "OpenInNewWindow", "OpenInPrivateWindow", "Separator", "BlockUrl"];
-    const title = this.props.title || this.props.source;
     const type = this.props.type || "DISCOVERY_STREAM";
+    const title = this.props.title; // eslint-disable-line prefer-destructuring
```
Mardak: How did you write the destructuring? Although, why not just keep the original `title || source` ? But then again, seems unlikely the title will require a fallback…

AllegroFox: You know, I don't know! I'm not sure when I removed the `title || source` but I don't see why it wouldn't work. I'll try switching it back.

AllegroFox: That seems to work fine and the tests pass, so that's that!

bin/bug_xxx_newtab.py
```diff
@@ -59,14 +59,15 @@ def migrate(ctx):
 newtab-settings-button =
     .title = { COPY(from_path, "settings_pane_button_label") }
 
-newtab-search-box-search-text = { COPY(from_path, "search_button") }
 newtab-search-box-search-button =
     .title = { COPY(from_path, "search_button") }
+    .aria-label = { COPY(from_path, "search_button") }
```
Mardak: In general, should we always be putting aria-label for anything that has a title attribute? At minimum it's the same string, but potentially aria-label can be different?

Mardak: E.g., newtab-settings-button a few lines above has just `.title`

AllegroFox: Yep! If they're duplicated, we make sure that screen readers are all getting the same information no matter which prop they use, especially if there's no anchor text. The aria-label could potentially be different, like in a case where it's describing special behaviour (I think that'll be the case for the drag-and-drop of the Top Sites, when I build an accessible container for them.)

content-src/components/CollapsibleSection/CollapsibleSection.jsx
punamdahiya: Getting @wolasi feedback as this is a user facing change and could be tricky with localized long text as below:

tooltip

wolasi: I think the previous default of "Open menu" is fine for the tooltip

AllegroFox: okay, I'll make that change :)

content-src/components/DiscoveryStreamComponents/DSLinkMenu/DSLinkMenu.jsx
```diff
@@ -49,8 +49,8 @@ export class _DSLinkMenu extends React.PureComponent {
     return (<div>
       <button ref={this.contextMenuButtonRef}
               className="context-menu-button icon"
-              title={this.props.intl.formatMessage({id: "context_menu_button_sr"}, {title})}
-              aria-label={this.props.intl.formatMessage({id: "context_menu_button_sr"}, {title})}
+              title={this.props.intl.formatMessage({id: "context_menu_title"}, {title})}
+              aria-label={this.props.intl.formatMessage({id: "context_menu_title"}, {title})}
```
punamdahiya: For aria-label detailed string with title makes sense and it should be good with id as  'context_menu_button_sr'.

Looks like we need to change back to context_menu_title in Card.jsx, CollapsibleSection.jsx and TopSite.jsx. Thanks!

AllegroFox: Got it!

AllegroFox: Ah hold on - do we want aria-label to have `context_menu_button_sr` and the titles to be `context_menu_title`? I figured we'd want title and aria-label to be the same since screen readers will usually pick one or the other, so that way everyone gets the same info

punamdahiya: yes, my take we should keep titles to be context_menu_title and aria-label to have context_menu_button_sr , this way we will be providing detailed information where screen reader picks aria-label

AllegroFox: okay!

content-src/components/Card/Card.jsx
```diff
@@ -245,10 +246,8 @@ export class _Card extends React.PureComponent {
           </div>
         </div>
       </a>
-      {!props.placeholder && <button className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"})}
-        onClick={this.onMenuButtonClick}>
-        <span className="sr-only">{`Open context menu for ${link.title}`}</span>
-      </button>}
+      {!props.placeholder && <button aria-haspopup="true" aria-label={this.props.intl.formatMessage({id: "context_menu_button_sr"}, {title})} className="context-menu-button icon" title={this.props.intl.formatMessage({id: "context_menu_title"}, {title})}
```
punamdahiya: @AllegroFox In title={this.props.intl.formatMessage({id: "context_menu_title"}, {title})} , we can omit title and should be title={this.props.intl.formatMessage({id: "context_menu_title"})}

content-src/components/Card/Card.jsx
```diff
@@ -200,6 +200,7 @@ export class _Card extends React.PureComponent {
   render() {
     const {index, className, link, dispatch, contextMenuOptions, eventSource, shouldSendImpressionStats} = this.props;
     const {props} = this;
+    const title = link.label || link.hostname;
```
punamdahiya: @AllegroFox just noticed, it seems  we don't have link.label in Card, it should be link.title || link.hostname;

PR #5101, opened 2019-06-10
Bug 1552937 - Preserve hover state on cards when hovering over context menu

gvn:
Testing: Confirm that hovering over context menu preserves the hover state on the rest of the card.


wolasi:
There should be no hover or focus effect on the placeholder tiles in Recommended by Pocket section since they don't contain content.

I think we should maintain the hover and focus effect for the Top Sites placeholders since users can edit them i.e. Add a site or search engine.

wolasi:
Also, the hero layout shouldn't be the model for the empty Recommended by Pocket sections. It should be the 3 card layout.

gvn:
@punamdahiya just pushed a fix for that placeholder regression. Good catch!

PR #5103, opened 2019-06-11
Bug 1553933 - Ensure we delete any remote spoc data when user opts ou…

ScottDowne:
…t of spocs.

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -247,6 +247,10 @@ const PREFS_CONFIG = new Map([
     skipBroadcast: true,
     value: "{}",
   }],
+  ["discoverystream.endpointSpocsClear", {
+    title: "Endpoint for when a user opts-out of sponsored content to delete the user's data from the ad server.",
+    value: "",
+  }],
   ["discoverystream.rec.impressions", {
     title: "Track rec impressions",
     skipBroadcast: true,
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -31,6 +31,8 @@ const FETCH_TIMEOUT = 45 * 1000;
 const PREF_CONFIG = "discoverystream.config";
 const PREF_ENDPOINTS = "discoverystream.endpoints";
 const PREF_IMPRESSION_ID = "browser.newtabpage.activity-stream.impressionId";
+const PREF_TOPSTORIES = "feeds.section.topstories";
+const PREF_SPOCS_CLEAR_ENDPOINT = "discoverystream.endpointSpocsClear";
 const PREF_SHOW_SPONSORED = "showSponsored";
 const PREF_SPOC_IMPRESSIONS = "discoverystream.spoc.impressions";
 const PREF_REC_IMPRESSIONS = "discoverystream.rec.impressions";
@@ -453,6 +455,23 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       },
     });
     this._sendSpocsFill({...filtered, frequency_cap: frequencyCapped}, true);
+  }
+
+  async clearSpocs() {
+    const endpoint = this.store.getState().Prefs.values[PREF_SPOCS_CLEAR_ENDPOINT];
+    if (!endpoint) {
+      return;
+    }
+    const headers = new Headers();
+    headers.append("content-type", "application/json");
+
+    await this.fetchFromEndpoint(endpoint, {
+      method: "DELETE",
+      headers,
+      body: JSON.stringify({
+        pocket_id: this._impressionId,
+      }),
+    });
   }
 
   async loadAffinityScoresCache() {
@@ -1033,8 +1052,18 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
               data: this.config,
             }));
             break;
+          case PREF_TOPSTORIES:
+            if (!action.data.value) {
+              // Ensure we delete any remote data potentially related to spocs.
+              this.clearSpocs();
+            }
+            break;
           // Check if spocs was disabled. Remove them if they were.
           case PREF_SHOW_SPONSORED:
+            if (!action.data.value) {
+              // Ensure we delete any remote data potentially related to spocs.
+              this.clearSpocs();
+            }
             await this.loadSpocs(update => this.store.dispatch(ac.BroadcastToContent(update)));
             break;
         }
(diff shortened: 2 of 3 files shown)


lib/ActivityStream.jsm
```diff
@@ -251,6 +251,10 @@ const PREFS_CONFIG = new Map([
     skipBroadcast: true,
     value: "{}",
   }],
+  ["discoverystream.endpointSpocsClear", {
+    title: "Endpoint for when a user opts-out of sponsored content to delete the user's data from the ad server.",
+    value: "https://spocs-dev.getpocket.dev/user",
```
Mardak: Any reason for a separate pref (vs as part of `.config`? Or a pref that's specific to spocs clear vs generally some server endpoint that we append `/user`?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1033,8 +1052,18 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
+              this.clearSpocs();
+            }
+            break;
           // Check if spocs was disabled. Remove them if they were.
           case PREF_SHOW_SPONSORED:
+            if (!action.data.value) {
+              // Ensure we delete any remote data potentially related to spocs.
+              this.clearSpocs();
+            }
             await this.loadSpocs(update => this.store.dispatch(ac.BroadcastToContent(update)));
```
Mardak: Should we be doing this for both spocs pref turning on and off? And this is making me wonder if we should be doing this as well for `PREF_TOPSTORIES` (i.e., have the two `case` statements use the same body ??)

ScottDowne: It actually already works this way I think.

I asked some of these questions this morning too.

From what I can tell, loadSpocs with spocs turned off successfully loads no spocs, so if you went from on to off without calling loadSpocs again, it would require a refresh of the page or restart to remove spocs.

On top of that, changing PREF_TOPSTORIES fires a chain of events that also ends up ref retching spocs because it clears cache and fetches fresh if that pref changes.

ScottDowne: If this is not actually the case though, I do think what you expected to happen should one way or another.

Mardak: I don't see `loadSpocs` getting called when enabling/disabling topstories pref without this PR. The original comment "Remove them if they were [enabled]" makes it seem like it needs to clean up stuff, but looks like the removing is for tab content and not necessarily cached state. I'm not sure if we need to clean up the disk cache in addition to cleaning up server data. Will the local cached data have personalized data based on the server response going forwards?

ScottDowne: Hm loadSpocs not being called, it should go into here: https://github.com/mozilla/activity-stream/blob/master/lib/DiscoveryStreamFeed.jsm#L879

Which should reset everything, and then re enable everything, which calls `enable`, then `refreshAll`, then `loadSpocs`

lib/ActivityStream.jsm
ScottDowne: I kept it separate to try to reduce the noise in the main `.config` There is already a lot in `.config` and I felt like this isn't something we would be changing based on a experiment.

I think moving it into the config would work, it would just make that config even bigger, which I thought would be nice to try to avoid.

lib/ActivityStream.jsm
```diff
@@ -247,6 +247,10 @@ const PREFS_CONFIG = new Map([
     skipBroadcast: true,
     value: "{}",
   }],
+  ["discoverystream.endpointSpocsClear", {
+    title: "Endpoint for when a user opts-out of sponsored content to delete the user's data from the ad server.",
+    value: "",
```
Mardak: When will we know what value to put here?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -455,6 +457,23 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
 
+  async clearSpocs() {
+    const endpoint = this.store.getState().Prefs.values[PREF_SPOCS_CLEAR_ENDPOINT];
+    if (!endpoint) {
+      return;
+    }
+    const headers = new Headers();
+    headers.append("content-type", "application/json");
+
+    await this.fetchFromEndpoint(endpoint, {
```
Mardak: To be clear, `fetchFromEndpoint` goes through the allowed endpoints check which I think is desired but affects what endpoints we use or might need to update that pref at the same time as setting the clear endpoint

test/unit/lib/DiscoveryStreamFeed.test.js
```diff
@@ -569,6 +569,36 @@ describe("DiscoveryStreamFeed", () => { // eslint-disable-line max-statements
...
+
+      assert.notCalled(feed.fetchFromEndpoint);
+    });
+    it("should call DELETE with endpoint", async () => {
+      sandbox.stub(feed.store, "getState").returns({
+        Prefs: {
+          values: {"discoverystream.endpointSpocsClear": "https://spocs/user"},
+        },
+      });
+      sandbox.stub(feed, "fetchFromEndpoint").resolves(null);
```
Mardak: Not sure if you were intentionally stubbing `fetchFromEndpoint` instead of `fetch` to bypass the security check

lib/ActivityStream.jsm
ScottDowne: Mathijs is setting that, he might be done, not sure.

In the end, there is a handful of turn it on related thing we need to do once all the pieces are ready and tested, so I'm guessing we have 1 patch either way to set that all up.

lib/DiscoveryStreamFeed.jsm
ScottDowne: The endpoint to clear the data is intended to be on the same endpoint that gets spocs, so it's, for example "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs" with a POST and "https://getpocket.cdn.mozilla.net/v3/firefox/user" with a DELETE request. So whatever the endpoint ends up being, we'll need to update it in a patch anyway, and it'll match an existing endpoint so we probably won't need to update the allowed endpoints any more than we already need to, which is once at most.

test/unit/lib/DiscoveryStreamFeed.test.js
ScottDowne: Yeah, there is a lot happening in fetch that I didn't want to test.

I think either or would have worked, but by the time this patch fires `fetchFromEndpoint` I wasn't noticing anything unique about it after that point.

I also didn't change anything inside `fetchFromEnpoint` so I figured keep the test constrained to this spot.

PR #5105, opened 2019-06-11
(nobug) - Update eslint-plugin-mozilla to 1.3.0 and ignore prettier for now

Mardak:
r? @dmose Ignore like most of m-c https://hg.mozilla.org/mozilla-central/rev/894a44c472727258338f2d013243c9b821a62686


.eslintrc.js
```diff
@@ -38,7 +38,6 @@ module.exports = {
     // Remove this when m-c updates their eslint: See https://github.com/mozilla/activity-stream/pull/4219
     "RPMSendAsyncMessage": true,
     "NewTabPagePreloading": true,
-    "DOMLocalization": true,
```
dmose: Why did this disappear?

dmose:
Why did you add eslint-plugin-prettier, but not the other new peer dependencies added in the m-c commit you referenced?

.eslintrc.js
Mardak: It was added to eslint-plugin-mozilla as part of 1.3.0, so it's redundant here

Mardak:
Why didn't I add `"eslint-config-prettier": "4.2.0",` ? The commit doesn't seem to require it (tests pass, no complaints from `npm i`) and I didn't add a script to use `eslint-config-prettier-check`

Mardak:
`eslint-plugin-prettier` was added because otherwise eslint would complain that the plugin doesn't exist. And without`prettier`, the plugin would complain

PR #5108, opened 2019-06-13
Bug 1559007 - Trailhead/onboarding is randomly triggered when RTAMO attributionData is present

rlr:
Is this the best way to fix this? I tried using the `order` parameter, but I guess that's for ordering bundles of cards.

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -147,7 +147,7 @@ const ONBOARDING_MESSAGES = async () => ([
   {
     id: "TRAILHEAD_1",
     template: "trailhead",
-    targeting: "trailheadInterrupt == 'join'",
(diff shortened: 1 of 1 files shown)


rlr:
looks like some of the old onboarding cards already had this in their targeting #oops 
https://github.com/mozilla/activity-stream/blob/72418c1edee1d1fe3632fa2682a1ba5e11d2ca6b/lib/OnboardingMessageProvider.jsm#L104

Mardak:
We probably want to fix this before / at the same time as we change the default for https://bugzilla.mozilla.org/show_bug.cgi?id=1552280

The current behavior works on release because there's no override pref set so asrouter knows not to set branches if it should show RTAMO instead.

Mardak:
Fixing as part of https://phabricator.services.mozilla.com/D35287

PR #5109, opened 2019-06-13
Bug 1558905 - Handle two new types of shims passed in two existing ev…

ScottDowne:
…ents, save and dismiss


ScottDowne:
I was trying to keep the data super clean, so no shim at all on the object, eve if it's undefined.

Was probably unnecessary, but was being careful. We can probably revisit, but I think I'll land this as is.

ScottDowne:
Looks like data review is good to go. Merging.

PR #5110, opened 2019-06-13
no bug - Fixing impression issue

ScottDowne:
To test:

1. Fresh profile
2. Open about:config
3. search for impressionId

expected: You should only see "browser.newtabpage.activity-stream.impressionId"
actual: You may see "impressionId" too, which is broken.


ScottDowne:
@ncloudioj This should be a quick review.

I think my pref isn't properly prepending all the activity stream pref stuff, and I think this fixes it.

PR #5111, opened 2019-06-13
Bug 1558312 - Optimistic loading for ds images

gvn:
Testing: 

You can add a console log [here](https://github.com/mozilla/activity-stream/blob/e04c656f6c22d3f62284ef931136bbac6e2df7f1/content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx#L30) to see that images are being "intersected" and thus fetched before they are fully in view.

You may still see blinks if there is latency with the network and/or caching service, but that's expected.


content-src/components/DiscoveryStreamComponents/DSImage/DSImage.jsx
```diff
@@ -42,7 +42,13 @@ export class DSImage extends React.PureComponent {
   }
 
   componentDidMount() {
-    this.observer = new IntersectionObserver(this.onSeen.bind(this));
+    this.observer = new IntersectionObserver(this.onSeen.bind(this), {
+      // Assume an image will be eventually seen if it is within 520px of the viewport
+      // This is half the average Desktop vertical screen size:
+      // http://gs.statcounter.com/screen-resolution-stats/desktop/north-america
+      rootMargin: `520px`,
```
punamdahiya: nit: why string literal

gvn: I always use them because there's no reason to use any other string syntax anymore AFAIK.

punamdahiya: With this patch, we are increasing images we are loading on first run for higher resolutions. Not sure about 520px number here. Looking at stat counter it shows 1920x1080 and 1366x768 as the most used. Changing Mac to 1680x1050 shows hard coded layout loading 9 images.
@ncloudioj do we have telemetry numbers of whats most favored screen resolution for Firefox users

ncloudioj: Yes, we do. Check out the `Display resolution` on [this page](https://data.firefox.com/dashboard/hardware).

gvn: "Two display resolutions, 1366x768px and 1920x1080px, stand out as the most highly-used, with the latter showing an upward trend"

@punamdahiya So it seems like `540px` would be slightly more accurate given our numbers. Is that cool?

punamdahiya: @gvn I am good with 540px, patch definitely optimize loading images early, I am not sure if jump from 3 to 9 images on first load has any performance number shift. Only way to overrule that is run some recordings and verify which I am ok doing in a separate bug about running performance numbers for 69 release

PR #5112, opened 2019-06-13
Bug 1559143 - Checkbox UI for prefs

ScottDowne:


lib/AboutPreferences.jsm
```diff
@@ -279,6 +279,11 @@ this.AboutPreferences = class AboutPreferences {
         subcheck.classList.add("indent");
         subcheck.setAttribute("label", formatString(nested.titleString));
         linkPref(subcheck, nested.name, "bool");
+
+        subcheck.disabled = !sectionData.enabled;
+        checkbox.addEventListener("click", () => {
+          subcheck.disabled = !checkbox.getAttribute("checked");
```
Mardak: Just use `checkbox.checked`. Getting the attribute is error prone because the value is `"true"` when checked.

https://searchfox.org/mozilla-central/rev/1133b6716d9a8131c09754f3f29288484896b8b6/toolkit/content/widgets/checkbox.js#80

Mardak: Technically, you should be watching for the pref value change as someone with about:preferences#home open and changing the value from about:config should result in the subcheck getting disabled… 😉

Mardak: E.g., Preferences.get(…).on("change", disableEnable)

https://searchfox.org/mozilla-central/rev/1133b6716d9a8131c09754f3f29288484896b8b6/browser/components/preferences/in-content/search.js#68-69

ScottDowne: I'll see if I can get these in.

ScottDowne:
Fixed up, mind taking another look?

PR #5116, opened 2019-06-14
Bug 1559243 - [AS] Re-enable pocket section after disable from about:…

punamdahiya:
…preferences

### DIFF
--- a/lib/TopStoriesFeed.jsm
+++ b/lib/TopStoriesFeed.jsm
@@ -36,7 +36,7 @@ this.TopStoriesFeed = class TopStoriesFeed {
   constructor(ds) {
     // Use discoverystream config pref default values for fast path and
     // if needed lazy load activity stream top stories feed based on
-    // actual user preference when PREFS_INITIAL_VALUES and PREF_CHANGED is invoked
+    // actual user preference when INIT and PREF_CHANGED is invoked
     this.discoveryStreamEnabled = ds && ds.value && JSON.parse(ds.value).enabled;
     if (!this.discoveryStreamEnabled) {
       this.initializeProperties();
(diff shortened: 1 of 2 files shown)


lib/TopStoriesFeed.jsm
```diff
@@ -705,10 +705,8 @@ this.TopStoriesFeed = class TopStoriesFeed {
       return;
     }
     switch (action.type) {
-      // Check for pref initial values to lazy load activity stream top stories
-      // Here we are not using usual INIT and relying on PREFS_INITIAL_VALUES
-      // to check discoverystream pref and load activity stream top stories only if needed.
-      case at.PREFS_INITIAL_VALUES:
+      // Check discoverystream pref and load activity stream top stories only if needed.
+      case at.INIT:
         this.lazyLoadTopStories(action.data[DISCOVERY_STREAM_PREF]);
```
Mardak: We could probably make lazyLoadTopStories use `getState().Prefs` if we don't pass in a value here or above handleDisabled.

lib/TopStoriesFeed.jsm
```diff
@@ -705,11 +705,9 @@ this.TopStoriesFeed = class TopStoriesFeed {
...
     }
     switch (action.type) {
-      // Check for pref initial values to lazy load activity stream top stories
-      // Here we are not using usual INIT and relying on PREFS_INITIAL_VALUES
-      // to check discoverystream pref and load activity stream top stories only if needed.
-      case at.PREFS_INITIAL_VALUES:
-        this.lazyLoadTopStories(action.data[DISCOVERY_STREAM_PREF]);
+      // Check discoverystream pref and load activity stream top stories only if needed.
+      case at.INIT:
+        this.lazyLoadTopStories(action.data[DISCOVERY_STREAM_PREF] || this.store.getState().Prefs.values[DISCOVERY_STREAM_PREF]);
```
Mardak: These should not be accessing action.data. INIT doesn't pass them. My suggestion for `getState()` is from `lazyLoadTopStories` side to avoid duplicating the code here and above.

PR #5118, opened 2019-06-14
Bug 1536285 - Add Highlights as the final discovery stream section

Mardak:
r?@ScottDowne Wraps similar to what we did for TopSites and adjusting the styling to fit with discoverystream cards:

Screen Shot 2019-06-14 at 5 32 50 PM

Screen Shot 2019-06-14 at 2 29 28 PM

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -145,7 +145,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["section.highlights.rows", {
     title: "Number of rows of Highlights to display",
-    value: 2,
+    value: 1,
   }],
   ["section.topstories.rows", {
     title: "Number of rows of Top Stories to display",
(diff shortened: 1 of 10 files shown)


wolasi:
@Mardak  Any chance we can make some of the minor adjust outlined in the spec e.g. making the source Title case instead of uppercase? That alone would be great for consistency.

Mardak:
Updated some styling from https://www.figma.com/file/eGdZVYlZ5evvePHUCtKnYg7u/New-Tab-Spec-for-Fx-70 (edited images in original comment)

wolasi:
@Mardak  can we make the following changes?
- even up the top and bottom padding on the text block
- make the font-size of the bold text 14px with line-height 20px

Mardak:
Are you saying the screenshot doesn't have those sizes or that's what they should be?

lib/ActivityStream.jsm
```diff
@@ -145,7 +145,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["section.highlights.rows", {
     title: "Number of rows of Highlights to display",
-    value: 2,
+    value: 1,
```
punamdahiya: why this change?

Mardak: for a single row:
Screen Shot 2019-06-14 at 7 04 51 PM

Mardak:
Filed download dismissing https://bugzilla.mozilla.org/show_bug.cgi?id=1559548
and new bookmarks https://bugzilla.mozilla.org/show_bug.cgi?id=1559552

PR #5121, opened 2019-06-20
Bug 1560235 - Fixing alignment issue and regression.

ScottDowne:
To test:

1. Set your `browser.newtabpage.activity-stream.discoverystream.config` pref to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":true,"personalized":false,"layout_endpoint":"https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=basic"}`
2. Load a new tab until you see a spoc in the 3rd spot.
3. Inspect the middle stories body text in the firefox inspector. It should look something like `<p class="excerpt clamp"></p>`
4. Remove the `clamp` from the class list.
5. Make the `p` text much longer, until you end up with something like this screenshot: 
text-align

ensure the sponsored by text is aligned on the bottom like the screen shot, and that it's aligned with the text in the middle story.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
@@ -56,6 +56,8 @@ $excerpt-line-height: 20;
 
   .ds-card-link {
     height: 100%;
+    display: flex;
+    flex-direction: column;
 
     &:focus {
       @include ds-fade-in;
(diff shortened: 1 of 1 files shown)


content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -88,10 +90,14 @@ $excerpt-line-height: 20;
...
     padding: 12px 16px;
+    flex-grow: 1;
 
     .info-wrap {
       flex-grow: 1;
-      margin: 0 0 12px;
+    }
+
+    .context {
+      margin: 12px 0 0;
```
Mardak: Why does this margin need to move to the 2nd sibling's top margin instead of he original 1st sibling's bottom margin?

PR #5122, opened 2019-06-20
Bug 1560352 - Send version number to spoc request

ScottDowne:


ScottDowne:
This was something me and mathijs noticed we should be doing while in whistler.

To test:

Right now, it should do the same thing, and not break anything.

ScottDowne:
It'll also be required with the change to the ad server.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -416,6 +416,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           headers,
           body: JSON.stringify({
             pocket_id: this._impressionId,
+            version: 1,
```
punamdahiya: why explicit version number in the code, is it ad server requirement, can it be set in spocs_endpoint

ScottDowne: It is required for the new proxy because a decision was made that it should be required, the current live spoc endpoint supports it but it isn't required (it probably should have been from the start)

setting it in spocs endpoint would probably be weird to both send data in POST body and the query string. I feel better keeping it all in 1 spot. If we were to move it to the endpoint we would still need to hard code it for at least the hard coded layout. I'm fine with it hard coded in general though.

PR #5123, opened 2019-06-20
Bug-1556039 - Fluent changes to 1556039 (removal of sr-only) spans

JeaneC:
Adding on fluent migrations to @AllegroFox  work in #5099. 

Tooltips on
- Sections
- Top Sites and empty Top Site boxes 
- Pocket Cards
- Highlights (downloaded items and websites)

### DIFF
--- a/bin/bug_xxx_newtab.py
+++ b/bin/bug_xxx_newtab.py
@@ -7,7 +7,7 @@ from __future__ import absolute_import
 import fluent.syntax.ast as FTL
 from fluent.migrate.helpers import transforms_from
 from fluent.migrate.helpers import TERM_REFERENCE, VARIABLE_REFERENCE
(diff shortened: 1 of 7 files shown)


bin/bug_xxx_newtab.py
```diff
@@ -179,6 +184,47 @@ def migrate(ctx):
...
+                    ),
+                    FTL.Attribute(
+                        id=FTL.Identifier("aria-label"),
+                        value=COPY(
+                            "browser/chrome/browser/activity-stream/newtab.properties",
+                            "edit_topsites_edit_button"
+                        )
+                    ),
+                ]
+            ),
```
punamdahiya: we can use plain copy here similar to newtab-menu-section-tooltip above

PR #5124, opened 2019-06-21
Bug 1560448 - Fixing campaign id in hero spocs

ScottDowne:


content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx
```diff
@@ -92,7 +92,7 @@ export class Hero extends React.PureComponent {
               </div>
             </div>
             <ImpressionStats
-              campaignId={heroRec.campaignId}
+              campaignId={heroRec.campaign_id}
```
punamdahiya: Good catch, it should be same as <DSCard campaignId={rec.campaign_id} above which is recording impressions correct.

PR #5125, opened 2019-06-21
Bug 1558550 - Convert strings in newtab content for topstories, highl…

JeaneC:
Fluent migrations
- Header Labels (“Top Sites”, “Highlights”, “Recommended by Pocket”)
- Pocket header and footer strings
- Error strings that appear when the tab fails to load or when a section fails to load (https://imgur.com/a/edjJCgy)

- Removed dead Strings
    - section_disclaimer_topstories, section_disclaimer_topstories_linktext, section_disclaimer_topstories_buttontext
    - pocket_read_even_more

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -50,6 +50,8 @@ module.exports = {
         "content-src/asrouter/templates/OnboardingMessage/**",
         "content-src/asrouter/templates/Trailhead/**",
         "content-src/components/TopSites/**",
+        "content-src/components/MoreRecommendations/MoreRecommendations.jsx",
+        "content-src/components/CollapsibleSection/CollapsibleSection.jsx"
       ],
       "rules": {
         "jsx-a11y/anchor-has-content": 0,
--- a/bin/bug_xxx_newtab.py
+++ b/bin/bug_xxx_newtab.py
@@ -149,6 +149,7 @@ newtab-empty-section-highlights = { COPY(from_path, "highlights_empty_state") }
 
 newtab-pocket-read-more = { COPY(from_path, "pocket_read_more") }
 newtab-pocket-more-recommendations = { COPY(from_path, "pocket_more_reccommendations") }
+newtab-pocket-how-it-works = { COPY(from_path, "pocket_how_it_works") }
 
 newtab-error-fallback-info = { COPY(from_path, "error_fallback_default_info") }
 newtab-error-fallback-refresh-link =
@@ -240,16 +241,6 @@ newtab-error-fallback-refresh-link =
                     "topstories_empty_state",
                     {
                         "{provider}": VARIABLE_REFERENCE("provider")
-                    },
-                )
-            ),
-            FTL.Message(
-                id=FTL.Identifier("newtab-pocket-how-it-works"),
-                value=REPLACE(
-                    "browser/chrome/browser/activity-stream/newtab.properties",
-                    "pocket_how_it_works",
-                    {
-                        "How it works": FTL.TextElement('<a data-l10n-name="learn-more">How it works</a>')
                     },
                 )
             ),
--- a/bin/bug_xxx_newtab.py
+++ b/bin/bug_xxx_newtab.py
@@ -143,7 +143,16 @@ newtab-section-menu-add-search-engine =
 newtab-section-menu-move-up = { COPY(from_path, "section_menu_action_move_up") }
 newtab-section-menu-move-down = { COPY(from_path, "section_menu_action_move_down") }
 newtab-section-menu-privacy-notice = { COPY(from_path, "section_menu_action_privacy_notice") }
+newtab-section-header-topsites = { COPY(from_path, "header_top_sites") }
+newtab-section-header-highlights = { COPY(from_path, "header_highlights") }
 newtab-empty-section-highlights = { COPY(from_path, "highlights_empty_state") }
+
+newtab-pocket-read-more = { COPY(from_path, "pocket_read_more") }
+newtab-pocket-more-recommendations = { COPY(from_path, "pocket_more_reccommendations") }
+newtab-pocket-how-it-works = { COPY(from_path, "pocket_how_it_works") }
+
+newtab-error-fallback-info = { COPY(from_path, "error_fallback_default_info") }
+newtab-error-fallback-refresh-link = { COPY(from_path, "error_fallback_default_refresh_suggestion") }
 
         """, from_path='browser/chrome/browser/activity-stream/newtab.properties')
     )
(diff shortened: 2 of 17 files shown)


content-src/components/MoreRecommendations/MoreRecommendations.jsx
```diff
@@ -7,7 +6,7 @@ export class MoreRecommendations extends React.PureComponent {
     if (read_more_endpoint) {
       return (
         <a className="more-recommendations" href={read_more_endpoint}>
-          <FormattedMessage id="pocket_more_reccommendations" />
+          <span data-l10n-id="newtab-pocket-more-recommendations" />
```
JeaneC: Can't do standalone anchor or need to ignore (eslint(jsx-a11y/anchor-has-content) - Anchor nodes can't be empty and need screen-reader accessible content)

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -173,7 +180,7 @@ export class _CollapsibleSection extends React.PureComponent {
                 {learnMore &&
                   <span className="learn-more-link">
                     <a href={learnMore.link.href}>
-                      <FormattedMessage id={learnMore.link.id} />
+                      <span data-l10n-id={learnMore.link.id} />
```
JeaneC: Ah... Can switch to data-l10n-name here

bin/bug_xxx_newtab.py
```diff
@@ -143,8 +143,17 @@ def migrate(ctx):
...
+newtab-section-header-topsites = { COPY(from_path, "header_top_sites") }
+newtab-section-header-highlights = { COPY(from_path, "header_highlights") }
 newtab-empty-section-highlights = { COPY(from_path, "highlights_empty_state") }
 
+newtab-pocket-read-more = { COPY(from_path, "pocket_read_more") }
+newtab-pocket-more-recommendations = { COPY(from_path, "pocket_more_reccommendations") }
+newtab-pocket-how-it-works = { COPY(from_path, "pocket_how_it_works") }
+
+newtab-error-fallback-info = { COPY(from_path, "error_fallback_default_info") }
+newtab-error-fallback-refresh-link = { COPY(from_path, "error_fallback_default_refresh_suggestion") }
```
punamdahiya: I think test is failing because of too long line here, can break this in two lines

content-src/components/CollapsibleSection/CollapsibleSection.jsx
punamdahiya: This might work 

```
<span className="learn-more-link" data-l10n-id={learnMore.link.id}><a data-l10n-name=“learn-more” href={learnMore.link.href}>How it works</a>
</span>
```
where newtab-pocket-how-it-works in ftl is
`newtab-pocket-how-it-works = <a data-l10n-name=“learn-more”> How it works </a>`

Will be good to check with other locales or[ intl.l10n.pseudo set to accented](https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_tutorial.html#pseudolocalization)

content-src/components/MoreRecommendations/MoreRecommendations.jsx
punamdahiya: @Mardak is it ok to specify default text to make eslint happy e.g. <a className="more-recommendations" href={read_more_endpoint} data-l10n-id="newtab-pocket-more-recommendations">More recommendations</a>

Mardak: We can add it to https://github.com/mozilla/activity-stream/blob/master/.eslintrc.js#L48-L55

Mardak: We can add it to https://github.com/mozilla/activity-stream/blob/1139bfc682ff544697ba5289a58b7623f3bc72d0/.eslintrc.js#L48-L55

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -171,10 +178,8 @@ export class _CollapsibleSection extends React.PureComponent {
               </span>
               <span className="learn-more-link-wrapper">
                 {learnMore &&
-                  <span className="learn-more-link">
-                    <a href={learnMore.link.href}>
-                      <FormattedMessage id={learnMore.link.id} />
-                    </a>
+                  <span className="learn-more-link" data-l10n-id={learnMore.link.id}>
+                    <a data-l10n-name="learn-more" href={learnMore.link.href} />
```
Mardak: Sorry for the late comment, this doesn't seem right ? If the whole string is a link, then the html doesn't need to be exposed to localizers. So put the `data-l10n-id` on the `<a>` like what you did for "more-recommendations"

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -189,6 +189,8 @@ export class _DiscoveryStreamBase extends React.PureComponent {
     }
 
     // Find the first component of a type and remove it from layout
+    // Note: Do not use the return component for additional FluentDOM use. Otherwise,
+    // the fluent IDs will be stripped and replaced with copy/replace strings.
```
JeaneC: "How it works" wasn't loading in Discovery Stream layout, but was working in regular view. Looks like its because we load Top Sites and Recommended by Pocket section twice (one for regular and one for DS), but choose to extract the first instance and copy over the properties from the already inserted DOM node.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -239,8 +241,8 @@ export class _DiscoveryStreamBase extends React.PureComponent {
           isFixed={true}
           learnMore={{
             link: {
-              href: message.header.link_url,
-              id: message.header.link_text,
+              href: topStories.learnMore.link.href,
+              id: topStories.learnMore.link.id,
```
Mardak: This is changing the behavior where learnMore link and text could be set by the server. I don't think this is desired. E.g., a few lines lower the title text is from the server's message response.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
Mardak: This wasn't working because the strings from layout are translated not ids. The DOM nodes are not already inserted -- the extraction happens from the layout definition. If you inspect the link, you'll see it tried to do:

```html
<a href="https://getpocket.com/firefox/new_tab_learn_more">
  <span data-l10n-id="How it works"></span>
</a>```

Comment (JeaneC) [reply content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx]: I didn't properly read what was going on here. Thanks for clarifying here and in the other comment. I tried to revert the changes in the next commit (a bit of a hacky fix though).

File: content-src/components/CollapsibleSection/CollapsibleSection.jsx [hunk 7]
```diff
@@ -172,9 +179,7 @@ export class _CollapsibleSection extends React.PureComponent {
               <span className="learn-more-link-wrapper">
                 {learnMore &&
                   <span className="learn-more-link">
-                    <a href={learnMore.link.href}>
-                      <FormattedMessage id={learnMore.link.id} />
-                    </a>
+                    <a href={learnMore.link.href} data-l10n-id={learnMore.link.id}>{learnMore.link.text}</a>
```
Mardak: I see we have a `getFormattedMessage(message)` helper here as well as in at least Sections.jsx. Seems like we might want to refactor this to take a `message: string | {id: …, values: …}` and perhaps make it a higher order component to allow setting the data-l10n-id attribute on the desired element vs just setting plain text child? This could be a followup

.eslintrc.js
```diff
@@ -50,6 +50,8 @@ module.exports = {
         "content-src/asrouter/templates/OnboardingMessage/**",
         "content-src/asrouter/templates/Trailhead/**",
         "content-src/components/TopSites/**",
+        "content-src/components/MoreRecommendations/MoreRecommendations.jsx",
+        "content-src/components/CollapsibleSection/CollapsibleSection.jsx"
```
punamdahiya: With updated PR, we can remove  "content-src/components/CollapsibleSection/CollapsibleSection.jsx"

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -217,7 +217,7 @@ export class _DiscoveryStreamBase extends React.PureComponent {
     const topSites = extractComponent("TopSites");
     const message = extractComponent("Message") || {
       header: {
-        link_text: topStories.learnMore.link.id,
+        link_id: topStories.learnMore.link.id,
```
Mardak: A DSMessage component only has link_text and not link_id, so this isn't quite right, but I'll fix it up with my getFormattedMessage change

.eslintrc.js
punamdahiya: never mind, noticed getFormattedMessage refactor PR needs eslint to ignore CollapsibleSection

PR #5126, opened 2019-06-24
Bug 1561000 - Open linkmenu to left for last item in layout

punamdahiya:


punamdahiya:
@Mardak waiting for next animation frame allows fluent dom to settle and appears to be simplest to fix this issue

PR #5127, opened 2019-06-24
 Bug 1561091 - Refactor getFormattedMessage to set fluent attributes or use plain text

Mardak:
r?@JeaneC This is on top of #5125, so I'll rebase when that gets merged. So look at the last commit for the changes for the refactoring.

### DIFF
--- a/content-src/components/FluentOrText/FluentOrText.jsx
+++ b/content-src/components/FluentOrText/FluentOrText.jsx
@@ -1,0 +1,32 @@
+import React from "react";
+
+/**
+ * Set text on a child element/component depending on if the message is already
+ * translated plain text or a fluent id with optional args.
+ */
+export class FluentOrText extends React.PureComponent {
+  render() {
+    // Ensure we have a single child to attach attributes
+    const { children, message } = this.props;
+    const child = children ? React.Children.only(children) : <span />;
+
+    // For a string message, just use it as the child's text
+    let grandChildren = message;
+    let extraProps;
+
+    // Convert a message object to set desired fluent-dom attributes
+    if (typeof message === "object") {
+      const args = message.args || message.values;
+      extraProps = {
+        "data-l10n-args": args && JSON.stringify(args),
+        "data-l10n-id": message.id || message.string_id,
+      };
+
+      // Use original children potentially with data-l10n-name attributes
+      grandChildren = child.props.children;
+    }
+
+    // Add the message to the child via fluent attributes or text node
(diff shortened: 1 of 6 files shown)


zbraniecki:
Drive by feedback. We have helpers similar to this in vanilla JS. One change I'd recommend is that since by default you want to carry l10n-id around, and carrying actual raw text is an exception, you can swap the default for better ergonomics:

*) localizeElement("key1")
*) localizeElement({raw: "My Text"})
*) localizeElement({id: "key1", args: {}})

Mardak:
I suppose in the common case it should just be a plain fluent id with no args. The current usage of `FluentOrText` is only for places that currently need to allow for either localized or raw, but I guess down the line, we could make everything use `FluentOrText` even when a plain html element could have just set `data-l10n-id` itself…? (Hopefully not causing too much overhead)

Mardak:
Actually, thinking about this a bit more, the server is currently sending plain text strings, so if we wanted to use the `{raw: string}` format, we would need to explicitly convert the server format into an object. Whereas if we just plain use the string value from the server, down the line the server could start giving a `{id: string}` response and automatically start using a localized version of the string instead.

Most of our uses of fluent are with plain `data-l10n-id` attributes already, so there's no need to additionally wrap with a `FluentOrText`. In the common case, we don't need to support either fluent translations or pre-translated strings, so we probably shouldn't unnecessarily wrap all uses.

content-src/components/FluentOrText/FluentOrText.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+  // For a string message, just use it as the child's text
+  let grandChildren = message;
+  let extraProps;
+
+  // Convert a message object to set desired fluent-dom attributes
+  if (typeof message === "object") {
+    extraProps = {
+      "data-l10n-id": message.id,
+      "data-l10n-args": message.values && JSON.stringify(message.values),
```
Mardak: I suppose we could make the `message` object provide `args` instead of `values` but somehow our existing messages use the `values` property that we would need to fix up all instances:
https://github.com/mozilla/activity-stream/blob/db21a369dd5cccff3c9cfa0bf9e8931ed03f7b2a/lib/SectionsManager.jsm#L33

Mardak: I would guess `values` comes from our previous usage of react-intl FormattedMessage https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedmessage

PR #5129, opened 2019-06-25
Bug 1561414 - Fixing unintended changes to About:Prefs that uses same strings in newtab pocket section

JeaneC:
@Mardak . Following #5127 (checkout last commit until it gets merged). 

Fortunately, the "header_recommended_by" was originally used in two different places, one for about prefs and one for newtab pocket section. But, "pocket_how_it_works" was shared from one source. Made changes ontop of #5127 to keep newtab fluent changes.

### DIFF
--- a/lib/SectionsManager.jsm
+++ b/lib/SectionsManager.jsm
@@ -19,7 +19,7 @@ const BUILT_IN_SECTIONS = {
   "feeds.section.topstories": options => ({
     id: "topstories",
     pref: {
-      titleString: {id: "newtab-section-header-pocket", values: {provider: options.provider_name}},
+      titleString: {id: "header_recommended_by", values: {provider: options.provider_name}},
       descString: {id: "prefs_topstories_description2"},
       nestedPrefs: options.show_spocs ? [{
         name: "showSponsored",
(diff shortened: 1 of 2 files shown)


lib/SectionsManager.jsm
```diff
@@ -35,6 +35,7 @@ const BUILT_IN_SECTIONS = {
       link: {
         href: "https://getpocket.com/firefox/new_tab_learn_more",
         message: {id: "newtab-pocket-how-it-works"},
+        id: "pocket_how_it_works"
```
punamdahiya: nit: need to end with comma to make lint happy

Mardak: Hmm.. So this link object is used in both newtab and about:preferences… I guess restoring `id` happens to fix it now, but I wonder how we'll want to fix it when migrating preferences strings.

Mardak: I suppose instead of just reusing the link object, a separate link object could be set in the `pref` object above.

punamdahiya: so this should be good now, when migrating preferences strings, we can create a separate link object inside Pref

Mardak:
The commit message refers to bug 1558550 but this is actually bug 1561414.

PR #5130, opened 2019-06-25
Bug 1561694 - Switching hardcoded layout to `3-col-7-row-octr`

gvn:
Testing:

- Open FF and confirm that default layout matches [winning layout](https://www.figma.com/file/o1Ep3tU1pCsCgM0gWNe7Iu/2-Branch-Study-Firefox-v67)

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -1183,10 +1183,7 @@ defaultLayoutResp = {
         }
       ]
     }
-  ],
-  "feeds": {},
-  "error": 0,
-  "status": 1
+  ]
 };
 
 const EXPORTED_SYMBOLS = ["DiscoveryStreamFeed"];
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -1072,10 +1072,11 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   }
 };
 
+// Hardcoded version of layout_variant `3-col-7-row-octr`
 defaultLayoutResp = {
   "spocs": {
-    "url": "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs",
-    "spocs_per_domain": 1,
+    "url": "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs?consumer_key=$apiKey",
+    "spocs_per_domain": 1
   },
   "layout": [
     {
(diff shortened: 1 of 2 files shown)


gvn:
@Mardak is this basically all that needs to be done for [this ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1551273)?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1072,10 +1072,11 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   }
 };
 
+// Hardcoded version of layout_variant `3-col-7-row-octr`
 defaultLayoutResp = {
   "spocs": {
-    "url": "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs",
-    "spocs_per_domain": 1,
+    "url": "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs?consumer_key=$apiKey",
```
Mardak: Did this API change on the server to reject `consumer_key`-less requests? Any users already seeing discovery stream will be getting the `400 bad request` response -- expected?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1100,204 +1101,92 @@ defaultLayoutResp = {
...
-    },
+            ".ds-navigation": "margin-top: -10px;"
+          }
+        }
+      ]
+    }
   ],
+  "feeds": {},
+  "error": 0,
+  "status": 1
```
Mardak: Are these fields intentionally added now? I believe @ScottDowne explicitly removed them before

gvn: This is just a copy of the layout JSON from the endpoint, so if it works without them I can remove them!

gvn:
> The bug was filed to turn on discovery stream for all channels, but this PR seems to only change the layout. Are we not expecting to turn on the feature via normandy or some other way?

Ok, yeah this is what was confusing me. I changed the baked layout for this ticket based on our conversation in stand, but that didn't seem to be what the ticket was for exactly. Should this actually be 2 separate tickets? I'm unfamiliar with Normandy.

lib/DiscoveryStreamFeed.jsm
Mardak: Actually, looks like spocs are currently requested via consumer_key as part of POST body. What is the server actually wanting? Both consumer_key in the url params and the post body?

gvn: @mmiermans Can you answer this Q?

lib/DiscoveryStreamFeed.jsm
gvn: Works without them...removing.

lib/DiscoveryStreamFeed.jsm
mmiermans: It hasn't changed; `consumer_key` has always been required. Without it, the response will be `400 Bad Request` and the `X-Error` header will say `Missing API key. Get an API key at http:\/\/getpocket.com\/api\/.`.

Good thing to call our in review, and thanks for fixing it @gvn!

PR #5132, opened 2019-06-26
Bug 1557413 - make eslint-fix (and therefore prettier) part of the build

dmose:
This makes eslint fix part of the bundle step, which means that it gets run in buildmc and at the beginning of startmc.   It also adds caching to all of the eslint invocations, making them lots faster.

It doesn't handle the ongoing watcher modes of startmc or tddmc.

That said, it gets us to a place where I think we can land the reformat after a bit of strategy discussion.


yamscripts.yml
```diff
@@ -58,7 +59,7 @@ scripts:
 
   testmc:
     lint: =>lint
-    build: =>bundle:webpack && =>bundle:locales
+    build: =>bundle:eslint-fix && =>bundle:webpack && =>bundle:locales
```
Mardak: This runs lint before running fix, so `npm test` / `npm run testmc` will fail before fixing. Separately, travis does `npm test` which would fix up somebody's bad PR that didn't prettier before push.

dmose:
OK, this is now the first PR you described.  I kept the eslint caching, since this seems like a generally useful optimization / feedback loop tightener.

yamscripts.yml
dmose: Splitting apart as suggested by Mardak, so this issue is no longer a thing.

.gitignore
```diff
@@ -15,3 +15,5 @@ prerendered/
 aboutlibrary/content/aboutlibrary.bundle.js
 aboutlibrary/content/*.map
 aboutlibrary/content/*.css
+.eslintcache
+
```
Mardak: extra newline. Probably move up just under `.DS_Store`

dmose: Extra?  There is exactly 1 newline at the end of the file now.  Before, there were 0.

dmose: That said, I've made the change you suggest, so it's back at 0.

PR #5133, opened 2019-06-26
Bug 1561536 - Add new message schema and template type for feature callouts

piatra:


lib/OnboardingMessageProvider.jsm
```diff
@@ -442,6 +442,16 @@ const ONBOARDING_MESSAGES = async () => ([
...
     trigger: {id: "firstRun"},
   },
+  {
+    id: "FXA_ACCOUNTS_BADGE",
+    template: "badge",
+    content: {
+      target: "fxa-toolbar-menu-button",
+    },
+    // Never accessed the FxA panel && doesn't use Firefox sync & has FxA enabled
+    targeting: `!hasAccessedFxAPanel && !usesFirefoxSync && isFxAEnabled == true`,
```
piatra: Will land this set to `false` so we don't risk triggering it somehow.

lib/ToolbarBadgeHub.jsm
```diff
@@ -0,0 +1,98 @@
...
+      win => {
+        if (notificationsByWindow.has(win)) {
+          // nothing to do
+          return;
+        }
+        const el = this.addToolbarNotification(win, message);
+        notificationsByWindow.set(win, el);
+      },
+      win => {
+        const el = notificationsByWindow.get(win);
```
rlr: Do we want to make sure an `el` was returned? Or is it impossible to get into that state where a window isn't in `notificationsByWindow`?

piatra: You're right. The methods are more or less placeholders right now until we have the proper badges (I expect changes).

piatra: The button should be there unless you remove it yourself.

lib/OnboardingMessageProvider.jsm
```diff
@@ -442,6 +442,16 @@ const ONBOARDING_MESSAGES = async () => ([
     targeting: "attributionData.campaign == 'non-fx-button' && attributionData.source == 'addons.mozilla.org'",
     trigger: {id: "firstRun"},
   },
+  {
+    id: "FXA_ACCOUNTS_BADGE",
+    template: "badge",
```
k88hudson: maybe `toolbar-badge`?

lib/ToolbarBadgeHub.jsm
```diff
@@ -0,0 +1,98 @@
...
+    this._blockMessageById = null;
+  }
+
+  async init(waitForInitialized, {handleMessageRequest, addImpression, blockMessageById}) {
+    this._handleMessageRequest = handleMessageRequest;
+    this._blockMessageById = blockMessageById;
+    this._addImpression = addImpression;
+    // Need to wait for ASRouter to initialize before trying to fetch messages
+    await waitForInitialized;
+    this.messageRequest("firstRunFxAccounts");
```
piatra: rename to: `toolbarBadgeUpdate`

rlr:
I am trying to test locally but the badge isn't showing up. And I tried to trigger it from devtools but I get an error `Invariant Violation: "Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings. "`

lib/OnboardingMessageProvider.jsm
rlr: Just a reminder to turn this off for now.

rlr:
works great. thanks! I'll remove the prototype branch and just keep a copy in my own fork so then you can just create a new branch with this and we'll add to it.

lib/ASRouterTargeting.jsm
```diff
@@ -308,6 +308,9 @@ const TargetingGetters = {
 
     return false;
   },
+  get hasAccessedFxAPanel() {
+    return Services.prefs.getBoolPref("identity.fxaccounts.toolbar.accessed", true);
```
k88hudson: What's your rationale behind returning `true` by default here? To prevent false positives? I think that makes sense for the for this case but it might not be for other usages

lib/ToolbarBadgeHub.jsm
```diff
@@ -0,0 +1,99 @@
...
+      return toolbarbutton;
+    }
+
+    return null;
+  }
+
+  registerBadgeNotificationListener(message) {
+    this._addImpression(message);
+
+    EveryWindow.registerCallback(
```
k88hudson: This is cool!

k88hudson:
Looks good to me :+1:

PR #5136, opened 2019-06-26
(nobug) - Run codespell lint for fluent files

Mardak:
Just testing against fluent branch as it should have a failure. But we can merge into master as we already have fluent files there.


Mardak:
`npm run mochitest` runs this:
```
  mochitest: (cd $npm_package_config_mc_dir && ./mach mochitest browser/components/newtab/test/browser --headless)
```
So it doesn't cover m-c lint. The mochitest.sh file is actually only run as part of `.taskcluster.yml`:
```
          git checkout {{event.head.sha}} && bash ./mochitest.sh
```

PR #5137, opened 2019-06-27
Bug 1551273 - Turning on DS for non-nightly users

gvn:
~ON HOLD: Pending final decision on regions.~

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -228,7 +228,7 @@ const PREFS_CONFIG = new Map([
 
       // Enable for US/en-US in all channels.
       // Enable for specific geos and locales for Nightly.
-      const isEnabled = (geo === `US` && locale === `en-US`) || (IS_NIGHTLY_OR_UNBRANDED_BUILD && locales.includes(locale));
+      const isEnabled = (geo === `US` && locale === `en-US`) || (IS_NIGHTLY_OR_UNBRANDED_BUILD && locales && locales.includes(locale));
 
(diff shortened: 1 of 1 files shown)


gvn:
@Mardak Any idea what the Taskcluster error is caused by?

lib/ActivityStream.jsm
```diff
@@ -222,10 +219,11 @@ const PREFS_CONFIG = new Map([
     title: "Configuration for the new pocket new tab",
     getValue: ({geo, locale}) => {
       const locales = ({
-        "US": ["en-CA", "en-GB", "en-US", "en-ZA"],
-        "CA": ["en-CA", "en-GB", "en-US", "en-ZA"],
+        "US": ["en-US"],
       })[geo];
-      const isEnabled = IS_NIGHTLY_OR_UNBRANDED_BUILD && locales && locales.includes(locale);
+
+      const isEnabled = locales && locales.includes(locale);
```
ScottDowne: Small nit, I wonder if we keep some of the existing logic, we could keep this on for nightly in en-CA.

Something like?

```
const isEnabled = (locale === "en-US") || (locales && locales.includes(locale));
```

gvn: I don't think that's the intention though. See [Slack thread](https://mozilla.slack.com/archives/CEZEF1AHX/p1561660281201100)

gvn: Nevermind...that is what's wanted.

PR #5138, opened 2019-06-27
Bug 1559382 - Dismissing the 2nd to last card will wrongly dismiss th…

punamdahiya:
…e last sponsored card

### DIFF
--- a/content-src/lib/selectLayoutRender.js
+++ b/content-src/lib/selectLayoutRender.js
@@ -91,7 +91,8 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
     }
 
     // Ensure we have recs available for this feed.
-    const hasRecs = data && data.recommendations && data.recommendations.length;
+    const hasRecs = data && data.recommendations;
+
     // Do we ever expect to possibly have a spoc.
     if (hasRecs && component.spocs && component.spocs.positions && component.spocs.positions.length) {
       // We expect a spoc, spocs are loaded, and the server returned spocs.


punamdahiya:
Including @wolasi for UX review, with this PR dismissing spoc item below will show next spoc item till all spoc stories are dismissed after which 'You are caught up ....' message is shown.

spoc-item

wolasi:
Here's what I think should happen
1. Dismiss second to last card as described in ticket
2. SPOC is still show if it's the last card
3. Dismiss SPOC and section is emptied

I think cycling through all the SPOCS as you describe above might prove annoying for the user.

punamdahiya:
> Here's what I think should happen
> 
>     1. Dismiss second to last card as described in ticket
> 
>     2. SPOC is still show if it's the last card
> 
>     3. Dismiss SPOC and section is emptied
> 
> 
> I think cycling through all the SPOCS as you describe above might prove annoying for the user.

Please note as of now when there are 2 or more stories, click of spoc dismiss shows the next spoc story. I can look into how much effort is customizing for special handling of dismiss when spoc story is last story shown.

punamdahiya:
@Mardak @ScottDowne @gvn I believe its complex to [check if a render](https://github.com/mozilla/activity-stream/blob/master/content-src/lib/selectLayoutRender.js#L98) is following a spoc dismiss with zero recommendations left. 

My understanding that check has to be in[ DiscoveryStreamFeed]( https://github.com/mozilla/activity-stream/blob/master/lib/DiscoveryStreamFeed.jsm#L1029)
while handling PLACES_LINK_BLOCKED  where we look for zero recommendations and available spocs and pass that as action.data  while dispatching event DISCOVERY_STREAM_LINK_BLOCKED to update spocs state with a flag say lastStory and use that flag in next render to show empty section state.

Looking for feedback if there is a simpler fix in absence of which weigh in if its worth handling this edge case.

punamdahiya:
Spocs can be displayed at multiple positions, this adds further complexity to check if dismissed story is only spoc and last item shown for a component

wolasi:
What about replacing a dismissed story with a spoc if there is one? I think this would be a better experience overall.

punamdahiya:
> What about replacing a dismissed story with a spoc if there is one? I think this would be a better experience overall.

@wolasi as discussed over slack, dismissed spoc will be replaced with another spoc if available.

 Will open a bug to revisit spoc dismissal in discovery stream to evaluate options such as
a) Alternate between showing a story and an ad
b) Display spoc story after a time delay
c) Not display spoc story in that session

punamdahiya:
Opened bug https://bugzilla.mozilla.org/show_bug.cgi?id=1562277

ScottDowne:
I agree with the complexity in dismissing a spoc and making it do anything other than display another spoc.

I think I just need to give the code here a run and if it works should be good to go.

PR #5139, opened 2019-06-27
Bug 1556895 - Convert first run strings in about:welcome and startup overlay to use fluent

JeaneC:
Migrating about:welcome strings to fluent
- Migrate old about:welcome (StartupOverlay.jsx) to fluent (Done)
- Update now broken files using about:welcome strings (Done 7/1)

### DIFF
--- a/bin/bug_1556895_newtab.py
+++ b/bin/bug_1556895_newtab.py
@@ -44,8 +44,6 @@ NB: migrate-l10n will make local commits to gecko-strings
 The diff should result in no differences if the migration recipe matches the
 fluent file.
 """
-SOURCE_FILE = 'locales-src/onboarding.ftl'
-
 
 def migrate(ctx):
     """Bug 1556895 - Migrate newtab.properties to onboarding.ftl, part {index}"""
@@ -56,13 +54,14 @@ def migrate(ctx):
         transforms_from("""
 
 onboarding-sync-welcome-content = { COPY(from_path, "firstrun_content") }
+onboarding-sync-welcome-learn-more-link = { COPY(from_path, "firstrun_learn_more_link") }
 onboarding-sync-form-invalid-input = { COPY(from_path, "firstrun_invalid_input") }
 onboarding-sync-form-header = { COPY(from_path, "firstrun_form_header") }
 onboarding-sync-form-input =
     .placeholder = { COPY(from_path, "firstrun_email_input_placeholder") }
 onboarding-sync-form-continue-button = { COPY(from_path, "firstrun_continue_to_login") }
 onboarding-sync-form-skip-login-button = { COPY(from_path, "firstrun_skip_login") }
-onboarding-menu-dismiss =
+onboarding-cards-dismiss =
     .title = { COPY(from_path, "menu_action_dismiss") }
     .aria-label = { COPY(from_path, "menu_action_dismiss") }
 
@@ -78,16 +77,6 @@ onboarding-menu-dismiss =
                 value=REPLACE(
                     "browser/chrome/browser/activity-stream/newtab.properties",
                     "firstrun_title",
-                    {
-                        "Firefox": TERM_REFERENCE("brand-product-name")
-                    },
-                )
-            ),
-            FTL.Message(
-                id=FTL.Identifier("onboarding-sync-welcome-learn-more-link"),
-                value=REPLACE(
-                    "browser/chrome/browser/activity-stream/newtab.properties",
-                    "firstrun_learn_more_link",
                     {
                         "Firefox": TERM_REFERENCE("brand-product-name")
                     },
--- a/bin/bug_1556895_newtab.py
+++ b/bin/bug_1556895_newtab.py
@@ -45,6 +45,7 @@ The diff should result in no differences if the migration recipe matches the
 fluent file.
 """
 
+
 def migrate(ctx):
     """Bug 1556895 - Migrate newtab.properties to onboarding.ftl, part {index}"""
 
--- a/bin/bug_1556895_newtab.py
+++ b/bin/bug_1556895_newtab.py
@@ -55,14 +55,16 @@ def migrate(ctx):
         SOURCE_FILE,
         transforms_from("""
 
-onboarding-control-form-privacy-notice = { COPY(from_path, "firstrun_privacy_notice") }
-onboarding-control-welcome-content = { COPY(from_path, "firstrun_content") }
-onboarding-control-form-invalid-input = { COPY(from_path, "firstrun_invalid_input") }
-onboarding-control-form-header = { COPY(from_path, "firstrun_form_header") }
-onboarding-control-form-input =
+onboarding-sync-welcome-content = { COPY(from_path, "firstrun_content") }
+onboarding-sync-form-invalid-input = { COPY(from_path, "firstrun_invalid_input") }
+onboarding-sync-form-header = { COPY(from_path, "firstrun_form_header") }
+onboarding-sync-form-input =
     .placeholder = { COPY(from_path, "firstrun_email_input_placeholder") }
-onboarding-control-form-continue-button = { COPY(from_path, "firstrun_continue_to_login") }
-onboarding-control-form-skip-login-button = { COPY(from_path, "firstrun_skip_login") }
+onboarding-sync-form-continue-button = { COPY(from_path, "firstrun_continue_to_login") }
+onboarding-sync-form-skip-login-button = { COPY(from_path, "firstrun_skip_login") }
+onboarding-menu-dismiss =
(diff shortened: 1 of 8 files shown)


locales-src/onboarding.ftl
```diff
@@ -30,6 +30,29 @@ onboarding-join-form-continue = Continue
...
+onboarding-control-form-input =
+    .placeholder = Email
+
+onboarding-control-form-continue-button = Continue
+onboarding-control-form-skip-login-button = Skip this step
+
+## Welcome modal dialog strings for pre-trailhead about welcome. This is part of the line
+## "Enter your email to continue to Firefox Sync"
+onboarding-control-form-header = Enter your email
+onboarding-control-form-sub-header = to continue to { -brand-product-name } Sync
```
Mardak: "Firefox Sync" should use `{ -sync-brand-short-name }`

locales-src/onboarding.ftl
```diff
@@ -30,6 +30,29 @@ onboarding-join-form-continue = Continue
 
 onboarding-start-browsing-button-label = Start Browsing
 
+## Welcome modal dialog strings for pre-trailhead about welcome.
+
+onboarding-control-welcome-header = Take { -brand-product-name } with You
+onboarding-control-welcome-content = Get your bookmarks, history, passwords and other settings on all your devices.
+onboarding-control-welcome-learn-more-link = Learn more about { -brand-product-name } Accounts
+
+onboarding-control-form-privacy-notice = Privacy Notice
```
Mardak: Probably won't need this separately just like Terms of Service deletion ?

locales-src/onboarding.ftl
```diff
@@ -30,6 +30,29 @@ onboarding-join-form-continue = Continue
 
 onboarding-start-browsing-button-label = Start Browsing
 
+## Welcome modal dialog strings for pre-trailhead about welcome.
+
+onboarding-control-welcome-header = Take { -brand-product-name } with You
+onboarding-control-welcome-content = Get your bookmarks, history, passwords and other settings on all your devices.
+onboarding-control-welcome-learn-more-link = Learn more about { -brand-product-name } Accounts
```
Mardak: "Firefox Account" is a brand.. although this one is plural… Then again, perhaps it shouldn't be? @flodolo ?

https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/branding/sync-brand.ftl#13

Mardak: I guess we could do something like `{ -fxaccount-brand-name(case: plural) }` and update sync-brand ?

flodolo: Touching the branding file is not an option. You would update English, and locales won't even know you did, unless you break half of Firefox.

I believe this shouldn't be plural in the first place.

```
onboarding-control-welcome-learn-more-link = Learn more about { -fxaccount-brand-name }
```

Mardak: Let's just use the original string as is to avoid needing new translations. Seems likely we won't be using this string soon anyway if we end up using the new join modal messaging.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -233,30 +226,29 @@ export class _Trailhead extends React.PureComponent {
...
       content && content.className,
     ].filter(v => v).join(" ");
+
     return (<>
     {this.state.isModalOpen && content ? <ModalOverlayWrapper innerClassName={innerClassName} onClose={this.closeModal} id="trailheadDialog" headerId="trailheadHeader">
       <div className="trailheadInner">
         <div className="trailheadContent">
           <h1 data-l10n-id={content.title.string_id}
-            id="trailheadHeader">{this.getStringValue(content.title)}</h1>
+            id="trailheadHeader" />
```
JeaneC: @Mardak instead of using FluentOrText can we just change property_id => string_id in OnboardingMessageProvider.jsm and use normal fluent?

Mardak: Ah yup plain fluent should be good as we aren't passing in raw strings here.

punamdahiya:
@JeaneC As discussed, dismiss tooltip and aria-label is not rendering in nightly because of "menu_action_dismiss" string id deleted from strings.properties

https://github.com/mozilla/activity-stream/blob/master/content-src/asrouter/templates/Trailhead/Trailhead.jsx#L315

To fix this issue we have two options
a) Include newtab.ftl in trailhead.jsx and update string id to use newtab-menu-dismiss defined in newtab.ftl
b) Create a new fluent string id but this needs to be handled in migration recipe by copying string from newtab.ftl?

@Mardak Not sure about option b as we have original string "menu_action_dismiss" deleted from newtab.properties

Mardak:
You can still refer to menu_action_dismiss even though it's not in the properties file. It still exists https://hg.mozilla.org/l10n/gecko-strings/file/d0131144d6c4b4ed04736582da2b70b98a5c39ce/browser/chrome/browser/activity-stream/newtab.properties#l35

punamdahiya:
> You can still refer to menu_action_dismiss even though it's not in the local properties file. It still exists https://hg.mozilla.org/l10n/gecko-strings/file/d0131144d6c4b4ed04736582da2b70b98a5c39ce/browser/chrome/browser/activity-stream/newtab.properties#l35
> 
> You'll need to test the migration recipe using the normal way not our custom testing overrides.

cool, does these legacy strings stays till two release cycles similar to migration recipes before its removed in [gecko-strings repo ](https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_migrations.html#how-migrations-are-run-on-l10n-repositories)

Mardak:
gecko-strings is a combination of strings available in mozilla-central/beta/release. We've deleted it in central, but it still exists in beta/release, so yeah 2 releases

JeaneC:
Trailhead.jsx now fluent. Also, fixed the dismiss tooltip (previously displayed "menu_action_dismiss", now just "Dismiss")

bin/bug_1556895_newtab.py
```diff
@@ -0,0 +1,132 @@
...
+PYTHONPATH=./bin migrate-l10n bug_1556895_newtab --lang en-US --reference-dir . \
+  --localization-dir ../gecko-strings
+diff -B -w locales-src/onboarding.ftl ../gecko-strings/browser/browser/newtab/onboarding.ftl
+```
+NB: migrate-l10n will make local commits to gecko-strings
+
+The diff should result in no differences if the migration recipe matches the
+fluent file.
+"""
+SOURCE_FILE = 'locales-src/onboarding.ftl'
```
Mardak: Oh, and we'll need to remove this line before we actually merge.

locales-src/onboarding.ftl
```diff
@@ -29,6 +29,29 @@ onboarding-join-form-legal = By proceeding, you agree to the <a data-l10n-name="
...
 onboarding-start-browsing-button-label = Start Browsing
+onboarding-menu-dismiss =
+    .title = Dismiss
+    .aria-label = Dismiss
+
+## Firefox Sync modal dialog strings.
+
+onboarding-sync-welcome-header = Take { -brand-product-name } with You
+onboarding-sync-welcome-content = Get your bookmarks, history, passwords and other settings on all your devices.
+onboarding-sync-welcome-learn-more-link = Learn more about { -brand-product-name } Accounts
```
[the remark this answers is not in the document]
flodolo: We did. Also, why do we have two PRs adding the same strings??
https://github.com/mozilla/activity-stream/pull/5144

flodolo:
> @flodolo any thoughts on duplicate strings? I guess because these are migrating anyway, it doesn't really matter. And makes them easier to delete when we can remove a branch. Notably, "valid email" and the legal "by proceeding" and "continue"

In this case it should be OK to have duplicates, especially if we plan to get rid of some of the branches at some point.

content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -312,13 +300,10 @@ export class _Trailhead extends React.PureComponent {
         {this.state.showCardPanel &&
           <button
             className="icon icon-dismiss" onClick={this.hideCardPanel}
-            title={props.intl.formatMessage({id: "menu_action_dismiss"})}
-            aria-label={props.intl.formatMessage({id: "menu_action_dismiss"})} />
+            data-l10n-id="onboarding-menu-dismiss" />
```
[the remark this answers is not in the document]
JeaneC: Sounds good

locales-src/onboarding.ftl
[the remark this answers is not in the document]
JeaneC: Right.

PR #5143, opened 2019-06-28
Bug 1562009 - Turn on ad server

ScottDowne:
To test:

1. Set `browser.newtabpage.activity-stream.discoverystream.config` to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":true,"personalized":false,"layout_endpoint":"https://getpocket.cdn.mozilla.net/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=basic"}`
2. Load a new tab

Ensure you see spocs.
Also try opting out of spocs, ensure that works.

For the most part you shouldn't notice anything changing.

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -240,7 +240,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["discoverystream.endpoints", {
     title: "Endpoint prefixes (comma-separated) that are allowed to be requested",
-    value: "https://getpocket.cdn.mozilla.net/",
+    value: "https://getpocket.cdn.mozilla.net/,https://spocs.getpocket.com",
   }],
   ["discoverystream.spoc.impressions", {
     title: "Track spoc impressions",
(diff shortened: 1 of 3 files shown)


gvn:
Repro steps work for me.

lib/ActivityStream.jsm
```diff
@@ -240,7 +240,7 @@ const PREFS_CONFIG = new Map([
   }],
   ["discoverystream.endpoints", {
     title: "Endpoint prefixes (comma-separated) that are allowed to be requested",
-    value: "https://getpocket.cdn.mozilla.net/",
+    value: "https://getpocket.cdn.mozilla.net/,https://spocs.getpocket.com",
```
Mardak: This probably should have a trailing slash, otherwise someone could register something like `https://spocs.getpocket.comdummydomain.com`

ScottDowne: :+1:

ScottDowne: Updated, thanks for catching this, was a big one.

PR #5145, opened 2019-07-02
Bug 1562954 - fixing link colour

ScottDowne:


content-src/components/DiscoveryStreamComponents/Navigation/_Navigation.scss
```diff
@@ -33,7 +33,7 @@
     a {
       &:hover {
         // text-decoration: underline; didn't quite match comps.
-        border-bottom: 1px solid $blue-40;
+        border-bottom: 1px solid var(--newtab-link-primary-color);
```
gvn: Is there a reason you wouldn't use a direct color var from `content-src/styles/_variables.scss` like we're doing for the bulk of DS?

ScottDowne: Because the links colour is set to `color: var(--newtab-link-primary-color);` already.

gvn: Hmm. Ok, that probably should be a Sass var as well, but since we're already inconsistent I guess it doesn't hurt.

Mardak: What color var would you use from _variables.scss ?

ScottDowne: We need to resolve that mismatch, but yeah, I don't think it's a fight for this bug.

gvn: Hopefully it can be a discussion and not a fight, but agree it's something we need to figure out a consistent approach toward.

PR #5148, opened 2019-07-03
Bug 1560689 - Remove react-intl dependencies and script loading

Mardak:
r?@dmose Delete delete delete ! Should be no more react-intl references in the repository!


Mardak:
Moved to phabricator to remove toolkit/content/license.html entry https://phabricator.services.mozilla.com/D36835

PR #5149, opened 2019-07-03
Bug 1563127 - Fixing invalid JSON from manual string injection

JeaneC:
With the way we were creating JSON strings we were getting an incorrect JS string value.

```
<button aria-haspopup="true" data-l10n-id="newtab-menu-content-tooltip" data-l10n-args="{ &quot;title&quot;: &quot;Resolving Error: &quot;Objects are not valid as a React child&quot;&quot; }" class="context-menu-button icon"></button>

{ "title": "Resolving Error: "Objects are not valid as a React child"" }
```

Fixing this in other places we use data-l10n-args


JeaneC:
Can replicate by creating a Top Site bookmark with double quotes. "Amazon" instead of Amazon will show parsing errors in the console.

Disappears after this patch.

PR #5150, opened 2019-07-03
Bug 1555450 - Remove locale-specific behavior when fully fluent (no longer using properties/json)

Mardak:
r?@k88hudson So much simpler! We might need to do this directly to m-c as we might need changes to `browser/base/content/test/static/browser_all_files_referenced.js` and `tools/lint/codespell.yml`


Mardak:
Moved to phabricator needing other m-c changes at the same time:
https://phabricator.services.mozilla.com/D36769

PR #5152, opened 2019-07-03
Bug 1559548 - Dismissing a download highlight makes it go away and reappear

Mardak:
r?@ScottDowne or @gvn Turns out we broke this in #4459 when trying to avoid high CPU issues for people with data uri downloads. So here restore the blocked links check but specially ignore long urls.


ScottDowne:
Looking

ScottDowne:
Getting this error, dunno if it's your patch or not, but my initial test with master doesn't have this error:
```
ReferenceError: Localization is not defined
```

Does this look familiar or related?

Also, it doesn't render a new tab anymore after that error.

Mardak:
Probably need to update your mozilla-central that includes `1558602 - Allow DocumentL10n to use LocalizationSync`

ScottDowne:
That worked.

PR #5153, opened 2019-07-03
Bug 1563319 - Enable the What's New UI when pref is enabled

rlr:
This depends on https://phabricator.services.mozilla.com/D36548


locales-src/asrouter.ftl
```diff
@@ -56,6 +56,7 @@ cfr-doorhanger-extension-total-users =
 cfr-doorhanger-pintab-description = Get easy access to your most-used sites. Keep sites open in a tab (even when you restart).
 
 ## These messages are steps on how to use the feature and are shown together.
+
```
piatra: Thanks for fixing the formatting.

PR #5155, opened 2019-07-03
Bug 1561660 - Tweaked CSS for welcome modal on smaller resolutions

AllegroFox:


Mardak:
Probably should check with Aaron on how the join modal should behave. Pretty sure when it was originally implemented, it was desired to not have the modal scroll but to have the modal be a fixed size allowing the page the scroll to see the rest if necessary.

Aaron also wanted the modal to be up higher, but I'm not sure what that means in terms of hiding the "Welcome to Firefox" text

AllegroFox:
Request ALL the reviews :)

PR #5156, opened 2019-07-04
Bug 1561540 - CFR messages should have a priority field that allows sorting

piatra:


content-src/asrouter/docs/targeting-attributes.md
```diff
@@ -485,3 +486,13 @@ Boolean pref that gets set the first time the user opens the FxA toolbar panel
 ```ts
 declare const hasAccessedFxAPanel: boolean;
 ```
+
+### `hasWhatsNewPanelEnabled`
```
rlr: `hasWhatsNewPanelEnabled` looks funny. Should it be `isWhatsNewPanelEnabled` similar to `isFxAEnabled`?

rlr: I say this now and realize we already reviewed that patch... This PR needs to be rebased I guess

lib/ASRouterTargeting.jsm
```diff
@@ -172,6 +172,28 @@ function sortMessagesByTargeting(messages) {
...
+    }
+    if (!isNaN(a.priority) && isNaN(b.priority)) {
+      return -1;
+    }
+    if (isNaN(a.priority) && !isNaN(b.priority)) {
+      return 1;
+    }
+
+    // Descending order; higher priority comes first
+    return a.priority > b.priority ? -1 : 1;
```
rlr: If they are equal, this will return 1. Technically, it should return 0 but I guess the end result will be the same

rlr:
this will need to be rebased again and have the formatting fixed with prettier

content-src/asrouter/docs/targeting-attributes.md
piatra: Renamed

lib/ASRouterTargeting.jsm
piatra: updated. Now the order of messages with same priority is left unchanged.

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -141,7 +134,6 @@ export class ContextMenuItem extends React.PureComponent {
           tabIndex="0"
           onClick={this.onClick}
           onKeyDown={this.onKeyDown}
-          ref={option.first ? this.focusFirst : null}
```
piatra: I rebased against a-s/master first thinking it would make it easier to update the prettier changes, I didn't realize it would add the extra commits that landed on master. I'll try to add this separately.

test/unit/ping-centre/PingCentre.test.js
```diff
@@ -340,6 +340,7 @@ describe("PingCentre", () => {
           locale: FAKE_LOCALE,
           topic: "activity-stream",
           client_id: FAKE_TELEMETRY_ID,
+          version: "69.0a1",
```
piatra: @ncloudioj this change was required because previously `AppConstants.MOZ_APP_VERSION` was `undefined` and the key would be dropped when creating a ping in the tests. Just want to confirm sending `version` is expected (I see it defined in our data_dictionary).

ncloudioj: Yes, `version` is a required field in every single ping. Thanks for the fix!

lib/ToolbarBadgeHub.jsm
```diff
@@ -37,6 +42,14 @@ class _ToolbarBadgeHub {
     this.messageRequest("toolbarBadgeUpdate");
   }
 
+  executeAction({ id }) {
+    switch (id) {
+      case "show-whatsnew-button":
+        ToolbarPanelHub.enableToolbarButton();
```
rlr: I guess we need to implement and call `ToolbarPanelHub.disableToolbarButton()` when the badge message is dismissed 🤔  Right now the toolbar button stays around. Can be a followup.

content-src/components/ContextMenu/ContextMenu.jsx
rlr: you can also rebase the `whatsnew-panel` branch on master and force push.

lib/ASRouter.jsm
```diff
@@ -771,6 +780,10 @@ class _ASRouter {
 
   uninit() {
     this._storage.set("previousSessionEnd", Date.now());
+    this._storage.set(
+      "previousSessionFirefoxVersion",
+      ASRouterTargeting.Environment.firefoxVersion
```
rlr: I think we said we might want to uplift this to 69 right? Or will we handle the special case of `undefined` and assume that means an upgrade happened?

rlr:
oh, looks like an aboutwelcome mochitest failed

lib/ASRouter.jsm
piatra: I'm not sure anymore this is the best way to do it. For example it will not cover users that upgrade from 68 to 70.
I think a better targeting expression would be
```json
{
  "id": "WHATS_NEW_BADGE_${FIREFOX_VERSION}",
  "template": "toolbar_badge",
  ...
  "targeting": "firefoxVersion >= 70 && ..."
```
I think this will work:
* We know the feature will be in 70 and we want to show it for everyone starting with 70
* It will work for 70+
* Because of the priority field we know it will be picked before other badges
* `WHATS_NEW_BADGE_70` won't be shown after 4 days because of targeting + impressions
* `WHATS_NEW_BADGE_71` will match again because of the different impression key
  * And we don't need to update the targeting for this or publish a new message
* It's independent of the what's new messages (they can have different targeting expressions that decide what Fx 70 users see vs Fx 71)

Let me know what you think, am I missing any case here?

piatra: Ah, but we want to show `whats's new` to users that upgrade to 70 not users that first run 70.

piatra: We can use `TelemetryReportingPolicy.isFirstRun()` to detect first runs in 70 (new users). We can skip `what's new` notifications for a whole release for these users by setting some pref/indexedDB value. This should work around the `need to land before 70` constraint.

lib/ToolbarBadgeHub.jsm
piatra: Shouldn't that be the concern of the Panel? We know we want to dismiss it after the very first click. I would instead register a `panel:hidden` event listener in `_showToolbarButton`.

Since the badge disappears when you click the button we would have a situation where an event listener/action is associated with a badge but that badge does not exist/has been dismissed which I find a bit weird.

What do you think?

rlr: Yeah that's a good point. I'll file a followup to have `ToolbarPanelHub` handle that.

PR #5157, opened 2019-07-05
Bug 1562894 - Patched out focus behaviour on ContextMenu

AllegroFox:
This temporarily removes the focus behavior introduced by 1559383; which created a bug with :hover and :focus styles. I'll reintroduce it in a later patch, with some refactoring that will prevent the bug.


content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -72,19 +72,23 @@ export class ContextMenuItem extends React.PureComponent {
...
     this.props.hideContext();
     this.props.option.onClick();
   }
 
-  focusFirst(button) {
-    if (button) {
-      button.focus();
-    }
-  }
+  // This function focuses the first option when the menu is rendered.
```
k88hudson: Can you remove this instead of commenting it out? You can reference the code in the bug to restore/fix the isssue if you want to save it

AllegroFox: okay, no problem!

PR #5158, opened 2019-07-08
Bug 1557304 - “About:welcome” cards overlapping issue with the browser resized to minimum width

AllegroFox:
Removed max-height css rule for trailhead cards to prevent the cards from overlapping with the main page body.

### DIFF
--- a/content-src/asrouter/templates/Trailhead/_Trailhead.scss
+++ b/content-src/asrouter/templates/Trailhead/_Trailhead.scss
@@ -264,7 +264,6 @@
 
 .trailheadCards {
   background: var(--trailhead-cards-background-color);
-  max-height: 1000px;
   overflow: hidden;
   text-align: center;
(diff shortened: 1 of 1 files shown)


content-src/asrouter/templates/Trailhead/_Trailhead.scss
```diff
@@ -264,7 +264,6 @@
 
 .trailheadCards {
   background: var(--trailhead-cards-background-color);
-  max-height: 1000px;
```
k88hudson: I can't see any reason why we couldn't remove this, @rlr do you remember why this was added?

AllegroFox:
okie doke!

Mardak:
https://github.com/mozilla/activity-stream/commit/5951c4947bea1a4eeed45f974ff9a51e33641c9f has no bug number in the commit message. Looks like github doesn't grab the PR's title when there's only one commit when requesting squash.

PR #5159, opened 2019-07-08
Bug 1561349 - Implement message templates for What's New panel

rlr:
This requires come changes to panelUI in m-c: https://phabricator.services.mozilla.com/D37320

The styling is bare bones, I'll be working with Aaron on styling tomorrow (Tuesday).

I think there is more that needs to be done to get messages with the right targeting, but thinking that should be a followup as this is getting big enough.

### DIFF
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -494,48 +494,76 @@ this.ASRouterTargeting = {
     return result;
   },
 
+  _getSortedMessages(messages) {
+    const weightSortedMessages = sortMessagesByWeightedRank([...messages]);
+    let sortedMessages = sortMessagesByTargeting(weightSortedMessages);
+    return sortMessagesByPriority(sortedMessages);
+  },
+
+  _getCombinedContext(trigger, context) {
+    const triggerContext = trigger ? trigger.context : {};
+    return this.combineContexts(context, triggerContext);
+  },
+
+  _isMessageMatch(message, trigger, context, onError) {
+    return (
+      message &&
+      (trigger
+        ? this.isTriggerMatch(trigger, message.trigger)
+        : !message.trigger) &&
+      // If a trigger expression was passed to this function, the message should match it.
+      // Otherwise, we should choose a message with no trigger property (i.e. a message that can show up at any time)
+      this.checkMessageTargeting(message, context, onError)
+    );
+  },
+
   /**
    * findMatchingMessage - Given an array of messages, returns one message
    *                       whos targeting expression evaluates to true
    *
    * @param {Array} messages An array of AS router messages
-   * @param {obj} impressions An object containing impressions, where keys are message ids
    * @param {trigger} string A trigger expression if a message for that trigger is desired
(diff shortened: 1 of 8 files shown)


test/unit/lib/ToolbarPanelHub.test.js
```diff
@@ -1,13 +1,31 @@
 import { _ToolbarPanelHub } from "lib/ToolbarPanelHub.jsm";
 import { GlobalOverrider } from "test/unit/utils";
 
+const FAKE_MESSAGE = {
```
piatra: Can we import this from `OnboardingProvider`? Easier to keep in sync & validate schema this way.

lib/ToolbarPanelHub.jsm
```diff
@@ -70,6 +71,75 @@ class _ToolbarPanelHub {
...
+        container.appendChild(messageEl);
+      }
+    }
+
+    // TODO: TELEMETRY
+  }
+
+  // If `string_id` is present it means we are relying on fluent for translations.
+  // Otherwise, we have a vanilla string.
+  _setString(doc, el, stringObj) {
```
piatra: :+1:

rlr:
^ rebased on latest whatsnew-panel branch

lib/ToolbarPanelHub.jsm
```diff
@@ -34,6 +34,9 @@ class _ToolbarPanelHub {
     this._getMessages = getMessages;
     if (this.whatsNewPanelEnabled) {
       this.enableAppmenuButton();
+      // TODO: this will eventually be called when the badge message gets triggered
+      // instead of here in init.
+      this.enableToolbarButton();
```
piatra: Was this added back by accident? It should now be trigger by the badge PR that landed in the `whatsnew-panel` branch

rlr: gah!

lib/ASRouterTargeting.jsm
```diff
@@ -502,14 +502,22 @@ this.ASRouterTargeting = {
    * @param {obj} impressions An object containing impressions, where keys are message ids
    * @param {trigger} string A trigger expression if a message for that trigger is desired
    * @param {obj|null} context A FilterExpression context. Defaults to TargetingGetters above.
+   * @param {boolean} returnAll If true, returns a list of all matching messages.
```
rlr: @piatra what do you think of this? Should I rename method to `findMatchingMessages()`? or?

rlr:
OK, now targeting is applied properly. My local m-c was still on 69 so I was only getting one message back. After I updated, I got all 3 messages.

lib/ASRouterTargeting.jsm
piatra: I'm not sure about this one. Personally I would even consider making a new method (even though 99% identical) that fetches all messages that match targeting:
It would make it more obvious when using one or the other and we wouldn't have to pass down `returnAll` through so many methods. Feels less hacky to me.
Not sure, cc @k88hudson ?

rlr:
I'll follow up with Aaron on the height of the panel and whether clicking CTA should close the panel or not.

PR #5160, opened 2019-07-09
Bug 1564146 - add fix:eslint commit hooks for prettier + install script

dmose:
Adds pre- and post-commit hooks for running prettier, and a bootstrap script to install them in one go.

The pre-commit hook aborts the commit if the `eslint --fix` fails.

I have been unable to make `git rebase -i` invoke the hooks, although there are some reports of this happening.  If we start seeing this issue, we can add something like https://blog.swwomm.com/2018/06/skip-pre-commit-hook-on-git-rebase-or.html

It'll probably be easiest to simply review the files-changed  tab rather than the individual commits due to commit interleaving issues.  Before I land, I'll squash stuff and remove the debugging output.

To test:

```bash
./bin/bootstrap # from the top-level dir

echo >> content-src/components/Base/Base.jsx
git status # verify that the file has been changed

git commit . -m 'test commit'
# verify the debugging output from the scripts looks right

git status
# verify that the index is clean

git diff HEAD^
# verify that the hooks have reverted the change in the last commit

# To test that the commit aborts if eslint fails, type some garbage into one of the JS
# files in an editor in a similar workflow to the above.
```


hooks/pre-commit
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+#
+# Recommended pre-commit git hook for activity-stream github repo
+#
+# Based on the example script in the prettier docs at
+# https://prettier.io/docs/en/precommit.html
+#
+FILES=$(git diff --cached --name-only --diff-filter=ACM "*.js" "*.jsx" | sed 's| |\\ |g')
```
Mardak: We'll want to include .jsm and not sure if we really need to fix up files with spaces as we don't have any

dmose: Yep, will fix the JSM thing. 

I'm inclined to leave the space fixer since it makes the script more robust in case someone adds a file with space.

dmose:
I've asked @Mardak for the review, but CC @k88hudson in case she's interesting in looking over as well.

hooks/pre-commit
```diff
@@ -0,0 +1,32 @@
...
+#
+# Runs `eslint --fix` on all selected files, which, given our current
+# prettier configuration, means prettifying these files as well.  The
+# commit will be aborted if eslint exits with a failure code.
+#
+# Based on the example script in the prettier docs at
+# https://prettier.io/docs/en/precommit.html
+
+FILES=$(git diff --cached --name-only --diff-filter=ACM "*.js" "*.jsx" "*.jsm" | sed 's| |\\ |g')
+echo "files to check: $FILES"
```
Mardak: Is this just your debugging to remove?

hooks/pre-commit
```diff
@@ -0,0 +1,32 @@
...
+[ -z "$FILES" ] && exit 0
+
+echo "$FILES" | xargs -t ./node_modules/.bin/eslint --fix
```
Mardak: Does --cache here help? Theoretically people might have already run test/lint/fix before trying to commit which should have updated .eslintcache ?

hooks/pre-commit
```diff
@@ -0,0 +1,32 @@
...
+if [ $? -ne 0 ]
+then
+  echo "eslint failed; aborting commit.  Please fix and retry."
+  exit -1
```
Mardak: I thought we said we wanted it to just continue even on failure? Or was that only in the context of rebases? The flow for a regular commit wouldn't really be that much different even if we continue on failure. Instead of plain retry commit, it would be commit amend or a new commit to squash later.

hooks/pre-commit
dmose: Yes

hooks/pre-commit
dmose: Shouldn't hurt; I'll add it.

hooks/pre-commit
dmose: As per the standup, I'm going to leave the abort in place.  If this really turns out to be a problem, we'll change it.

hooks/pre-commit
dmose: Done

bin/bootstrap
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -x
+
+# this hook is no longer needed, so we're removing it
+rm -f .git/hooks/pre-push
```
k88hudson: This seems a bit aggressive if someone has their own hooks, no? I guess based on feedback it probably doesn't matter much either way

hooks/pre-commit
```diff
@@ -0,0 +1,32 @@
...
+# https://prettier.io/docs/en/precommit.html
+
+FILES=$(git diff --cached --name-only --diff-filter=ACM "*.js" "*.jsx" "*.jsm" | sed 's| |\\ |g')
+echo "files to check: $FILES"
+[ -z "$FILES" ] && exit 0
+
+echo "$FILES" | xargs -t ./node_modules/.bin/eslint --cache --fix
+if [ $? -ne 0 ]
+then
+  echo "eslint failed; aborting commit.  Please fix and retry."
```
k88hudson: You might want to say something like "eslint found errors but was unable to fix them all with --fix. Please check the output and attempt to manually fix" or along those lines

bin/bootstrap
dmose: Fair point; I'll remove it.

hooks/pre-commit
dmose: Definitely better; will tweak.

PR #5161, opened 2019-07-09
Bug 1559383 - Added keyboardAccess prop to all components with contex…

AllegroFox:
…t menus.
The context menu now uses this prop to detect whether it was accessed via the keyboard or the mouse, and focus the first menu option appropriately.


Mardak:
This seems a bit verbose needing every parent component to do something. Is it possible instead to check if the menu button has focus to conditionally move focus from the button to the menu?

AllegroFox:
> This seems a bit verbose needing every parent component to do something. Is it possible instead to check if the menu button has focus to conditionally move focus from the button to the menu?

That's what I had tried initially, but the issue was that the menu isn't actually a child of the button, so the button can't pass its props down to the menu. Additionally, since the menu hasn't been rendered to the DOM at the time that the menu button event is triggered, using the event to trigger some state change on the component is really unreliable.

Mardak:
Not using react props but DOM state `document.activeElement` should indicate if the focus is currently on the button:
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement

Although it isn't very encapsulated to check global document state.

AllegroFox:
> Not using react props but DOM state `document.activeElement` should indicate if the focus is currently on the button:
> https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement
> 
> Although it isn't very encapsulated to check global document state.

I also had some issues with React losing track on re-render, or not being able to locate the correct element (since the context menus are arrays of the same element) - the parent component's state is a little heavy-handed, but it seems safest.

content-src/components/Card/Card.jsx
```diff
@@ -326,10 +340,10 @@ export class _Card extends React.PureComponent {
         </a>
         {!props.placeholder && (
           <button
-            aria-haspopup="true"
```
k88hudson: Did you mean to remove this?

AllegroFox: I did not!  Fixed it.

k88hudson:
I tried refactoring some of this into a single component in [this PR](https://github.com/AllegroFox/activity-stream/pull/1), what do you think @AllegroFox @Mardak? The usage looks like this:

```jsx
<ContextMenuButton {..someProps}>
  <SomeLinkMenuComponent {...someOtherProps}>
</ContextMenuButton>
```
Internally, it clones the `SomeLinkMenuComponent` component and adds some additional props (like `onUpdate`)

AllegroFox:
This looks good to me with Kate's refactor - ready to merge?

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -254,29 +250,26 @@ export class CollapsibleSection extends React.PureComponent {
...
                 showPrefName={showPrefName}
                 privacyNoticeURL={privacyNoticeURL}
                 collapsed={collapsed}
-                onUpdate={this.onMenuUpdate}
                 isFixed={isFixed}
                 isFirst={isFirst}
                 isLast={isLast}
                 dispatch={dispatch}
                 isWebExtension={isWebExtension}
+                keyboardAccess={this.state.contextMenuKeyboard}
```
Mardak: Is the contextMenuKeyboard state on the parent components still being used? Seems like the prop keyboardAccess prop gets ignored in favor of the ContextMenuButton's state ?

k88hudson: no, that should be deleted

AllegroFox: done!

AllegroFox:
Okay, I think I caught them all!

Mardak:
Code coverage with this PR:

content-src/components/ContextMenu ContextMenu.jsx
65.91% Statements 29/44
51.43% Branches 18/35
86.67% Functions 13/15
65.91% Lines 29/44

content-src/components/ContextMenu ContextMenuButton.jsx
68.42% Statements 13/19
50% Branches 5/10
66.67% Functions 4/6
68.42% Lines 13/19

We should at least make sure the new functionality added here gets coverage. A separate bug would be good for the other untested stuff, e.g., focusSibling, onKeyDown -> Tab / Escape.

AllegroFox:
I'll try and write some tests today.

AllegroFox:
I think you may be right! I was primarily testing that the new prop was passed down correctly since the focusFirst function was originally a different PR - but my last test should have been to test that it was firing correctly when the prop was present. I'll put it in there.

Mardak:
The commit message in https://github.com/mozilla/activity-stream/commit/0379c988b17bcdef56bd8ac880368b0e636a2035 refers to the old bug number

AllegroFox:
shoot, sorry. Is it worth force-pushing to change?

PR #5162, opened 2019-07-09
Bug 1563741 - Keeping body scroll static, but welcome modal centered …

JeaneC:
Description: When the about:welcome is opened the user can scroll past the modal.
Changes: Disabled scrolling for about:welcome and fixed modal to stay centered. If there's an overflow, then the actual modal itself is scrollable.

For the cases of low display resolution or the user is zoomed in, pending design feedback from @aaronrbenson. Would it be better to have the page overflow instead of the modal (scrollbar on the page and not modal).

Original
https://drive.google.com/open?id=1rDg9O7iouAgigjxRpHUdfR97N6iGUBZx

New Changes
https://drive.google.com/open?id=1AtQF14SAYbGwUwwP3nBe1os5kd3yn6JI

### DIFF
--- a/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
+++ b/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
@@ -19,21 +19,26 @@ export class ModalOverlayWrapper extends React.PureComponent {
   componentWillMount() {
     this.props.document.addEventListener("keydown", this.onKeyDown);
     this.props.document.body.classList.add("modal-open");
+    this.header = this.props.document.getElementById(
+      "header-asrouter-container"
+    );
 
-    this.props.document
-      .getElementById("header-asrouter-container")
-      .classList.add("modal-scroll");
-    this.props.document.getElementById("root").classList.add("modal-height");
+    if (this.header) {
+      this.header.classList.add("modal-scroll");
+      this.props.document.getElementById("root").classList.add("modal-height");
+    }
   }
 
   componentWillUnmount() {
     this.props.document.removeEventListener("keydown", this.onKeyDown);
     this.props.document.body.classList.remove("modal-open");
 
-    this.props.document
-      .getElementById("header-asrouter-container")
-      .classList.remove("modal-scroll");
-    this.props.document.getElementById("root").classList.remove("modal-height");
+    if (this.header) {
+      this.header.classList.remove("modal-scroll");
+      this.props.document
+        .getElementById("root")
+        .classList.remove("modal-height");
+    }
   }
 
   render() {
--- a/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
+++ b/content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
@@ -32,7 +32,7 @@ export class ModalOverlayWrapper extends React.PureComponent {
 
     this.props.document
       .getElementById("header-asrouter-container")
-      .classList.remove(".modal-scroll");
+      .classList.remove("modal-scroll");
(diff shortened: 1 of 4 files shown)


AllegroFox:
I feel like we had decided the page should scroll rather than the modal content when the window is too small for the modal - but this may be better if we're trying to make sure that the modal stays in view until it's closed.

k88hudson:
Hey the design of this kind of hard to describe, so I made this pen of it: https://codepen.io/k88hudson/pen/PrVmXp

Basically we want the outer overlay to be scrollable (also `auto` is better to make sure they don't show up unnecessarily)  but the content of the page shouldn't be

JeaneC:
Thanks for the pen! I tried to do something similar by making the container holding trailhead "absolute". This had the effect of shifting normal Newtab content up (Search bar, Topsites) and we compensated for that by adding a padding because it overlaps otherwise with "Welcome to Firefox". 

Trying to figure out why the test cases are failing. Fixed some by mocking out the document.getElementById we're now using but it's not working for asrouter-content.test.jsx.

Alternatively, it was suggested to toggle TrailheadCards to none or "#root" (containing all Newtab) to none. Thoughts?

content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -15,11 +15,21 @@ export class ModalOverlayWrapper extends React.PureComponent {
   componentWillMount() {
     this.props.document.addEventListener("keydown", this.onKeyDown);
     this.props.document.body.classList.add("modal-open");
+
+    this.props.document
+      .getElementById("header-asrouter-container")
```
punamdahiya: Can move getElementById("header-asrouter-container") in constructor as this.header = this.props.document.getElementById("header-asrouter-container"); and use it to set and unset css

punamdahiya:
@k88hudson do you know why tests are failing here? Thanks

test/unit/asrouter/ModalOverlay.test.jsx
```diff
@@ -11,6 +11,9 @@ describe("ModalOverlayWrapper", () => {
       addEventListener: sandbox.stub(),
       removeEventListener: sandbox.stub(),
       body: { classList: { add: sandbox.stub(), remove: sandbox.stub() } },
+      getElementById() {
+        return { classList: { add: sandbox.stub(), remove: sandbox.stub() } };
```
k88hudson: this can just be `document.createElement("div")` if you want

content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
k88hudson: In order to fix the tests, I think the right thing to do here is to check the existence of the element before adding the class, i.e.
``` js
this.header = this.props.document.getElementById("header-asrouter-container"); 

...
if (this.header) {
  ...
}

```

k88hudson: The other thing you could do is pass `props.document` further down the chain, but because some of this code is being removed soon i don't think it's worth it.

k88hudson:
This looks great, thank you! UI+

content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
k88hudson: (The reason the tests are failing is because `props.document` is not passed down to `ModalOverlayWrapper` from the outer asrouter-content, so it's just using the default `global.document` instead)

PR #5163, opened 2019-07-09
Bug 1561349 - UI polish for What's New messages

rlr:
This requires the latest in https://phabricator.services.mozilla.com/D37320

Screen Shot 2019-07-09 at 4 22 01 PM

Some notes from discussion with Aaron:
* When user clicks on a CTA, the panel will go away... At least for now.
* We might need to tweak the height of the panel depending on the final content.


lib/ToolbarPanelHub.jsm
```diff
@@ -78,43 +78,44 @@ class _ToolbarPanelHub {
     const container = doc.getElementById(containerId);
 
     if (messages && !container.querySelector(".whatsNew-message")) {
+      let previousDate = 0;
```
piatra: Should messages be sorted by date to ensure this will work?

lib/OnboardingMessageProvider.jsm
```diff
@@ -534,8 +537,9 @@ const ONBOARDING_MESSAGES = async () => [
       body:
         "The New Enhanced Tracking Protection, gives you the best level of protection and performance. Discover how this version is the safest version of firefox ever made.",
       link_text: "Learn more on our blog",
-      link_url: "https://blog.mozilla.org/",
+      cta_url: "https://blog.mozilla.org/",
     },
+    priority: 1,
```
piatra: Why was this needed?

rlr: oh right. I should've sorted by date instead of adding this 😝

lib/ToolbarPanelHub.jsm
```diff
@@ -126,11 +136,29 @@ class _ToolbarPanelHub {
       // TODO: TELEMETRY
     });
 
-    messageEl.appendChild(dateEl);
-    messageEl.appendChild(wrapperEl);
+    if (content.icon_url) {
```
piatra: Can you combine this with the previous `if` statement on line `123`?

test/unit/lib/ToolbarPanelHub.test.js
```diff
@@ -7,7 +7,6 @@ describe("ToolbarPanelHub", () => {
   let sandbox;
   let instance;
   let everyWindowStub;
-  let message;
```
piatra: Can you please also add a test for checking that the elements are rendered in sorted order?
I would stub `toLocaleDateString` and check it's only called `X` number of times. Or that the content rendered has only `X` `dateEl` elements.

rlr:
correct. All messages will have a CTA URL but not necessarily an explicit link/button visible. And you can click anywhere on any message to trigger it.

PR #5165, opened 2019-07-10
Bug 1564898 - Use UpdateManager to get the earliest Firefox version used

piatra:
You can see the UpdateManager in action on `about:support` with a real profile 
image
image

`previousFirefoxVersion` will return the very first firefox version used with the current profile (69.0a1 as seen in my screenshot) so it could be named better, I'm open to suggestions.

Testing this might be tricky so I'm thinking we merge `whatsnew-panel` branch into master to get this into Nightly sometime this week. Then QA can verify/validate the method.

### DIFF
--- a/content-src/asrouter/docs/targeting-attributes.md
+++ b/content-src/asrouter/docs/targeting-attributes.md
@@ -36,7 +36,7 @@ Please note that some targeting attributes require stricter controls on the tele
 * [hasPinnedTabs](#haspinnedtabs)
 * [hasAccessedFxAPanel](#hasaccessedfxapanel)
 * [isWhatsNewPanelEnabled](#iswhatsnewpanelenabled)
-* [previousFirefoxVersion](#previousfirefoxversion)
+* [earliestFirefoxVersion](#earliestfirefoxversion)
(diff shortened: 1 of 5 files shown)


content-src/asrouter/docs/targeting-attributes.md
```diff
@@ -36,6 +36,7 @@ Please note that some targeting attributes require stricter controls on the tele
 * [hasPinnedTabs](#haspinnedtabs)
 * [hasAccessedFxAPanel](#hasaccessedfxapanel)
 * [isWhatsNewPanelEnabled](#iswhatsnewpanelenabled)
+* [previousFirefoxVersion](#previousfirefoxversion)
```
rlr: maybe `firstfirefoxversion`? or `oldestfirefoxversion`?

rlr:
hmm, I am thinking it only keeps the last N upgrades? Because I've had my profile forever but it is returning `69.0a1`. Was this implemented in 69 maybe?

rlr:
I did do a profile refresh recently, so maybe that cleared it out.

PR #5167, opened 2019-07-10
Bug 1564188 - Use remote Fluent files for ASRouter

ncloudioj:
This allows ASRouter to use remote Fluent files for message localization.

Since the [new bucket](https://bugzilla.mozilla.org/show_bug.cgi?id=1564189) has not yet created on Remote Settings, we need to use the RS dev environment for local test. [This doc](https://github.com/mozilla/activity-stream/blob/master/docs/v2-system-addon/remote_cfr.md#using-a-dev-server-for-remote-cfr) shows how to set up the dev environment for RS.


ncloudioj:
@leplatrem Could you please take a look at the Kinto bit of this patch [here](https://github.com/mozilla/activity-stream/pull/5167/files#diff-6201bd23e10848d056ecc572cecd3e18)?

@Pike Could you please take a look at the l10n bit [here](https://github.com/mozilla/activity-stream/pull/5167/files#diff-7b912b00bf1c976eb6572f76c816ecfe)?

lib/ASRouter.jsm
```diff
@@ -289,6 +305,28 @@ const MessageLoaderUtils = {
             provider.id,
             options.dispatchToAS
           );
+        } else if (["cfr", "cfr-fxa"].includes(provider.id)) {
```
piatra: Is there anything `provider` specific about the file we download? Are we downloading the same file twice?

Pike:
One concern I have about saving these files to the profile is around Avast. They tend to freak out about stuff we don't normally do.

Also, the files on disk don't have a language associated. This will result in weird behavior when a user switches languages.

Should we do language fallback on remote settings? That is, if we don't have a German message, show the most recent English one?

I'd also love to understand a bit more on what we're storing on RS, and how we got to the design. Aka, what attachments give us over plain entries.

ncloudioj:
> One concern I have about saving these files to the profile is around Avast. They tend to freak out about stuff we don't normally do.
> 

We've already used this feature of Remote Settings for other part of Activity Stream (the attachment was persisted in the same directory), we haven't heard about any report around antivirus interference by far. @leplatrem do you have any take on this?

> Also, the files on disk don't have a language associated. This will result in weird behavior when a user switches languages.
> 
> Should we do language fallback on remote settings? That is, if we don't have a German message, show the most recent English one?
> 

Good point! I was actually thinking about having a locale change observer here, so that we can re-download the Fluent file when one switches to another locale. @Pike would love to hear your thoughts on this. Also, do we have such notification available?

> I'd also love to understand a bit more on what we're storing on RS, and how we got to the design. Aka, what attachments give us over plain entries.

On RS, we store the exact same Fluent files imported from "l10n-central", we have scripts to sync those files from the "l10n-central" repo, and then upload them to RS. Currently, we'd publish all the new content to RS stage first, then get QA to verify the new changes before we publish it to production. In the future, we'd like to improve this publishing process by integrating things similar to the "l10n merge" check to further assure the quality of remote l10n content.   

On the client side, Firefox will fetch the Fluent file from RS based on its current locale. Once the download succeeds, our code utilizes the low level l10n API to create the `FluentBundle` generator and initialize the `DOMLocalization` accordingly.

ncloudioj:
> This works great and I really like the solution. I don't think it solves the problem for the bookmark panel messages, those are still inserting the [local .ftl file](https://searchfox.org/mozilla-central/rev/da855d65d1fbdd714190cab2c46130f7422f3699/browser/components/newtab/lib/BookmarkPanelHub.jsm#110) right?
> The bookmark panel messages also reference `brand.ftl` terms so `_createDOML10n` will have to account for that.

Correct, this solution currently doesn't support the BookmarkPanel messages. we'd have to investigate further to see if there is any workaround. Also, once this [feature](https://bugzilla.mozilla.org/show_bug.cgi?id=1564268) gets implemented in l10n, I think it would not be a problem any more.

lib/ASRouter.jsm
ncloudioj: RS downloader will compare the local file checksum against the remote one before sending out the download request. Nonetheless, you're right, it should just ignore the download if the previous one is already completed.

lib/CFRPageActions.jsm
```diff
@@ -30,6 +37,12 @@ const CATEGORY_ICONS = {
   cfrAddons: "webextensions-icon",
   cfrFeatures: "recommendations-icon",
 };
+const L10N_FLUENT_DIR = OS.Path.join(
```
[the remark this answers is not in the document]
ncloudioj: Good idea!

lib/ASRouter.jsm
ncloudioj: I've added some basic mechanism to avoid the duplicate attachment downloads (even for attempts). Note that it only does so for one message load cycle, i.e. a call of `ASRouter.loadMessagesFromAllProviders()`, so that it will always check the freshness of the attachment for every message load.

Pike:
I think we should make this multi-locale ready, and have the locale code in local paths.

For language switches, we're sending `intl:app-locales-changed`, https://firefox-source-docs.mozilla.org/intl/locale.html#events.

lib/ASRouter.jsm
```diff
@@ -116,6 +126,12 @@ const LOCAL_MESSAGE_PROVIDERS = {
 };
 const STARTPAGE_VERSION = "6";
 
+// Remote Settings
+const RS_SERVER_PREF = "services.settings.server";
+const RS_MAIN_BUCKET = "main";
+const RS_COLLECTION_L10N = "ms-language-packs"; // "ms" stands for Messaging System
```
leplatrem: was the `ms` prefix really necessary then? It's sad that it requires a comment ;)

lib/ASRouter.jsm
```diff
@@ -289,6 +320,38 @@ const MessageLoaderUtils = {
             provider.id,
             options.dispatchToAS
           );
+        } else if (["cfr", "cfr-fxa"].includes(provider.id)) {
```
leplatrem: nit: shouldn't these be defined in a constant?

lib/ASRouter.jsm
```diff
@@ -289,6 +320,38 @@ const MessageLoaderUtils = {
...
+          // Both "cfr" and "cfr-fxa" require the Fluent file for l10n, currently
+          // they share the same attachment of the record "cfr-${version}-${locale}",
+          // e.g. for "en-US" with version "v1", the record ID is "cfr-v1-en-US".
+
+          const locale = Services.locale.appLocaleAsLangTag;
+          const recordId = `cfr-${RS_FLUENT_VERSION}-${locale}`;
```
leplatrem: Is it because cfr and cfr-fxa share the same record id that it's not a module constant?

lib/ASRouter.jsm
```diff
@@ -289,6 +320,38 @@ const MessageLoaderUtils = {
...
+          }
+
+          const kinto = new KintoHttpClient(
+            Services.prefs.getStringPref(RS_SERVER_PREF)
+          );
+          const record = await kinto
+            .bucket(RS_MAIN_BUCKET)
+            .collection(RS_COLLECTION_L10N)
+            .getRecord(recordId);
+          if (record && record.data) {
```
leplatrem: I would invert this test in order to add a warning or throw...
```js
if (!record || !record.data) {
   console.warn("No language package found");
}
```

lib/ASRouter.jsm
```diff
@@ -289,6 +320,38 @@ const MessageLoaderUtils = {
...
+            const downloader = new Downloader(
+              RS_MAIN_BUCKET,
+              RS_COLLECTION_L10N
+            );
+            // Await here in order to capture the exceptions for reporting.
+            await downloader.download(record.data, { retries: 2 });
```
leplatrem: nit: Number of retries as global constant?

lib/ASRouter.jsm
```diff
@@ -134,6 +150,21 @@ const MessageLoaderUtils = {
...
+   * the same record can be simply skipped.
+   *
+   * Once all the providers complete the message load, one should reset this array
+   * via `this.resetRemoteAttachmentFetched()`.
+   */
+  _remoteAttachmentFetched: [],
+
+  resetRemoteAttachmentFetched() {
+    this._remoteAttachmentFetched = [];
+  },
```
leplatrem: I'm not sure this is necessary.

The attachment downloader is already a no-op if the file already exists locally. So basically this allows to save the HTTP requests that pull the record metadata. But if I'm not mistaken, those requests will be cached.

lib/ASRouter.jsm
```diff
@@ -289,6 +320,38 @@ const MessageLoaderUtils = {
...
+        } else if (["cfr", "cfr-fxa"].includes(provider.id)) {
+          // Both "cfr" and "cfr-fxa" require the Fluent file for l10n, currently
+          // they share the same attachment of the record "cfr-${version}-${locale}",
+          // e.g. for "en-US" with version "v1", the record ID is "cfr-v1-en-US".
+
+          const locale = Services.locale.appLocaleAsLangTag;
+          const recordId = `cfr-${RS_FLUENT_VERSION}-${locale}`;
+
+          // Skip the downloaded if it's already fetched.
+          if (MessageLoaderUtils._remoteAttachmentFetched.includes(recordId)) {
```
leplatrem: nit: using a Map could be more appropriate

lib/CFRPageActions.jsm
```diff
@@ -68,12 +84,46 @@ class PageAction {
...
+      const appLocale = Services.locale.appLocaleAsBCP47;
+      const appLocales = Services.locale.appLocalesAsBCP47;
+      // The downloaded Fluent file is located in
+      // `/${OS.Constants.Path.localProfileDir}/settings/main/ms-language-packs/`
+      const l10nFluentDir = OS.Path.join(
+        OS.Constants.Path.localProfileDir,
+        "settings",
+        "main",
+        "ms-language-packs"
+      );
```
leplatrem: This could be defined as a constant. 
Also, the downloader helper returns the file URI already. And since it's a no-op when the file exists, you can use it to obtain the URI

leplatrem:
>> One concern I have about saving these files to the profile is around Avast. They tend to freak out about stuff we don't normally do.
>>
> We've already used this feature of Remote Settings for other part of Activity Stream (the attachment was persisted in the same directory), we haven't heard about any report around antivirus interference by far. @leplatrem do you have any take on this?

Good question. I have no idea. Shall we ask our QAs?

lib/ASRouter.jsm
ncloudioj: I figured this collection was a "Messaging System" specific at the moment. I don't mind making it go as "language-packs" though :), I believe it's fine to rename a collection at this point, correct?

Pike: I'd prefer if we kept this to be messaging system specific for now.

When we add actual language packs in remote settings, we'll likely want things to change, also based on what we're learning here.

lib/ASRouter.jsm
ncloudioj: Yep, I think we can make a constant for that.

lib/ASRouter.jsm
ncloudioj: Good point.

lib/ASRouter.jsm
ncloudioj: > The attachment downloader is already a no-op if the file already exists locally.

I assume that downloader will also check the checksum against the remote apart from the file existence, correct?

If that's the case, yeah, this check is unnecessary then, I can drop it for sure.

lib/CFRPageActions.jsm
ncloudioj: Yep, I've noticed that downloader would return the file URI upon successful downloading. However, it's kinda cumbersome for us to pass or persist that URI so that it can be used later by this module. Also, given that the file URI is located in a well-known directory of RS, so I've chosen to assemble it over here.

ncloudioj:
> > > One concern I have about saving these files to the profile is around Avast. They tend to freak out about stuff we don't normally do.
> > 
> > 
> > We've already used this feature of Remote Settings for other part of Activity Stream (the attachment was persisted in the same directory), we haven't heard about any report around antivirus interference by far. @leplatrem do you have any take on this?
> 
> Good question. I have no idea. Shall we ask our QAs?

Per the offline conversion around this topic, we'd like to get QA to confirm this as part of the QA take for this feature.

ncloudioj:
@Pike Is there a preferred way to switch locale without using addons? I've tried the methods from this [page](https://blog.mozilla.org/l10n/2019/04/02/changing-the-language-of-firefox-directly-from-the-browser/) by flipping the pref `intl.multilingual.enabled` and `intl.multilingual.downloadEnabled`. Though `search for more languages` couldn't find anything.

Also, does locale change always require a browser restart to take effect? If that's the case, the re-download will be done automatically by the RS client, as it will find out the local file is different from the remote one. @leplatrem Can you confirm?

Mardak:
I believe those multilingual prefs only really work not on nightly. Here's a screenshot of about:preferences on 69 beta after selecting/downloading Italian:

Screen Shot 2019-07-23 at 7 06 08 PM

And here's the output on console after restart:
```
Services.locale.availableLocales
Array [ "en-US", "it" ]

Services.locale.appLocalesAsLangTags
Array [ "it", "en-US" ]

Services.locale.requestedLocales
Array [ "it", "en-US" ]

Services.locale.appLocaleAsBCP47
"it"
```

leplatrem:
> Also, does locale change always require a browser restart to take effect? If that's the case, the re-download will be done automatically by the RS client, as it will find out the local file is different from the remote one. @leplatrem Can you confirm?

Indeed, even if the filename will be the same, the file size/hash will be different so it will be re-downloaded. It's worth having a QA test for this scenario though :)

Pike:
> Also, does locale change always require a browser restart to take effect?

Once our UI is in Fluent, it won't require a restart. The requirement for a restart exists right now, though. I'd prefer to not introduce more code paths that require it.

Regarding switching languages on Nightly, for tests you can download language packs from http://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central-l10n/mac/xpi/ or a dated copy from http://archive.mozilla.org/pub/firefox/nightly/2019/07/ to match your local firefox. Then the preferences ui allows you to switch between them.

ncloudioj:
Thank you all for the clarification!

Agreed, let's make this feature locale switching friendly then. Will make a patch for that in this PR.

ncloudioj:
The locale switch handling has been implemented in 52e5b96.

Note that since the `intl:app-locales-changed` gets notified after one clicks the `Apply and Restart` button, the handler can only respond in a very tight window, chances are it never gets to perform the re-download before the restart takes place. Hence this will work as expected when the restart-free locale switching comes out.

@Pike Could you take another look?

ncloudioj:
> I can't reason about what happens if new content comes down, how language switching will time. In particular with the caching of parsed resources inside `FileSource`.

We periodically poll RS server for the new messages as well as the fluent files, once the new content gets synced to local, all the subsequent new CFR messages will use the latest fluent for l10n.

It's worth noting that all the CFR messages are created on-the-fly in an on-demand manner. I.e. it gets created *when* we find one meets the targeting criteria of a certain message, and the RS backed fluent file will be attached at that point via `FileSource.fetchFile`. So we won't reuse the existing `FileSource`, do we still need to handle cache invalidation here?

lib/CFRPageActions.jsm
```diff
@@ -68,17 +90,49 @@ class PageAction {
...
+        RS_DOWNLOADED_FILE_SUBDIR
+      );
+      const fs = new FileSource("cfr", [appLocale], `file://${l10nFluentDir}/`);
+      // In the case that the Fluent file has not been downloaded from Remote Settings,
+      // `fetchFile` will return `false` and fall back to the packaged Fluent file.
+      const resource = await fs.fetchFile(appLocale, "asrouter.ftl");
+      if (resource) {
+        // Use a single locale array for the bundle from Remote Settings.
+        const bundle = new FluentBundle(appLocale);
+        bundle.addResource(resource);
```
Pike: As we talked about on the phone, we'll need to get the brands stuff into this bundle, probably like so:

```javascript
if (resource) {
  for (let bundle of L10nRegistry.generateBundles([appLocale], resourceIds.slice(1))) {
    // override, as we're the last instead of the first resource
    bundle.addResource(resource, true);
    yield bundle;
  }
}
```

ncloudioj: Nice catch! I've updated the patch to also include other resources into the bundle.

Mardak:
I thought I left a comment re: versioning, but it's a general topic for us to think about our usage of remote settings + desktop firefox. What's our plan for supporting older versions of firefox that are expecting certain remote settings messages? Seems like naively, we would need to keep remote settings bucket with all messages or all versions of firefox?

Specifically for l10n, as long as the remote l10n is only used by a matching remote message, then I guess we're more okay. But if there's a desktop firefox version that relies on a remote string, then things get tricky.

ncloudioj:
> I thought I left a comment re: versioning, but it's a general topic for us to think about our usage of remote settings + desktop firefox. What's our plan for supporting older versions of firefox that are expecting certain remote settings messages? Seems like naively, we would need to keep remote settings bucket with all messages or all versions of firefox?

Those are very great points! I believe this patch is really just the beginning of this new workflow, various tougher challenges are still open for us to tackle next.

Firstly, the current implementation already employes the concept of versioning. Specifically, all message providers in Messaging System are sharing the same Remote Settings collection `ms-language-packs` for l10n, and each record in this collection is named by `{providor-id}-{version}-{locale}`. For example, the record `cfr-v1-en-US` will provide the `en-US` Fluent file (as a RS attachment) for the `cfr` of version `v1`. A new version will be added to if we're going to introduce incompatible changes to the Fluent file. Obviously, this requires some extra work on version management, which presumably could be done offline (perhaps by maintaining our own repo of all published fluent files). More thoughts are needed here to find the best solution.

Another thing, I think, we need to consider more is around fault tolerance. Particularly, as Fluent files and other messages are stored separately in different RS collections. Our code should gracefully handle all scenarios it would possibly encounter, for instance, what if only one gets synced with RS, and the other one is not? Is it always safe to use the local Fluent as the fallback? What happens when one switches among different Firefox versions with the same profile? Those are all potential problems once we switch everything to RS.

ncloudioj:
@Pike @leplatrem @piatra Could you take another quick look at this patch as all the review comments have been marked as resolved? Thanks!

lib/CFRPageActions.jsm
```diff
@@ -8,12 +8,23 @@ const { XPCOMUtils } = ChromeUtils.import(
 );
 const { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyGlobalGetters(this, ["fetch"]);
+ChromeUtils.defineModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm");
```
piatra: nit: could you please group all `moduleGetter` calls under 1 `XPCOMUtils.defineLazyModuleGetters` call instead?

lib/CFRPageActions.jsm
```diff
@@ -68,17 +85,58 @@ class PageAction {
...
+   */
+  _createDOML10n() {
+    async function* generateBundles(resourceIds) {
+      const appLocale = Services.locale.appLocaleAsBCP47;
+      const appLocales = Services.locale.appLocalesAsBCP47;
+      const l10nFluentDir = OS.Path.join(
+        OS.Constants.Path.localProfileDir,
+        RS_DOWNLOADED_FILE_SUBDIR
+      );
+      const fs = new FileSource("cfr", [appLocale], `file://${l10nFluentDir}/`);
```
piatra: I'm surprised this doesn't include `RS_FLUENT_VERSION` somewhere in the filename path.

test/unit/asrouter/ASRouter.test.js
```diff
@@ -3048,4 +3132,68 @@ describe("ASRouter", () => {
...
+      sandbox.stub(global.Services.locale, "appLocaleAsLangTag").get(getter);
+      const provider = {
+        id: "cfr",
+        enabled: true,
+        type: "remote-settings",
+        bucket: "cfr",
+      };
+      await createRouterAndInit([provider]);
+      sandbox.spy(Router, "setState");
+      sandbox.stub(Router, "loadMessagesFromAllProviders").resolves(null);
```
piatra: nit: I would stub/spy on the methods at the beginning of the test and use `resetHistory` instead.

lib/CFRPageActions.jsm
ncloudioj: Yeah, I think it makes sense to have one copy of attachment here. Since the attachments (Fluent files) do not know anything about the versions, rather, its version is completely decided by its corresponding record (e.g. `cfr-v1-en-US`). Upon locale or version changes, it will re-download the attachment accordingly and overwrite the old one.

test/unit/asrouter/ASRouter.test.js
ncloudioj: I agree with you that stubbing/spying should be done in the beginning in general, though I think the ordering does matter in this case. Because it had to call `createRouterAndInit()` with a special provider (the default fake provider for Remote Settings doesn't have `id` as `cfr` or `cfr-fxa`, so it won't trigger the attachment downloading work), spying `Router` prior to `createRouteAndInit()` won't work as it will mutate `Router`.

I did simplify this test a bit though.

ncloudioj:
Thanks again for the review, everyone!

Will land it once we get the following things prepared:

* Upload l10n to RS
* Finish QA

PR #5171, opened 2019-07-11
(nobug) Move the What's New and Badge test messages to PanelTestProvider.jsm

rlr:
With this patch you can still trigger the badging from devtools, but the actual messages don't show up in the panel. I guess because they aren't really loaded into ASRouter (?).


rlr:
I think we need some special devtools feature/code to send those messages into the panel. The panel calls `ASRouter.handleMessageRequest(...)` when opening to get the matching messages. But these local test messages aren't loaded into `ASRouter` AFAICT.

There is [Bug 1561942](https://bugzilla.mozilla.org/show_bug.cgi?id=1561942) for improving the devtools support of this feature, so it could be looked into as part of that.

PR #5179, opened 2019-07-12
Bug 1556862 - Localizing empty discovery stream messages

JeaneC:
Thinking about calling fluent headers for Discovery Stream like "Empty Section (Content Discovery Experience)", since the old experience of activity stream is still here with similar strings.

### DIFF
--- a/locales-src/newtab.ftl
+++ b/locales-src/newtab.ftl
@@ -149,13 +149,14 @@ newtab-empty-section-highlights = Start browsing, and we’ll show some of the g
 newtab-empty-section-topstories = You’ve caught up. Check back later for more top stories from { $provider }. Can’t wait? Select a popular topic to find more great stories from around the web.
 
 
-## Empty Section (Content Discovery Experience). These show when there are no more pocket stories or when some stories fail to load.
+## Empty Section (Content Discovery Experience). These show when there are no more stories or when some stories fail to load.
 
 newtab-discovery-empty-section-topstories-header = You are caught up!
 newtab-discovery-empty-section-topstories-content = Check back later for more stories.
 newtab-discovery-empty-section-topstories-try-again-button = Try Again
-newtab-discovery-empty-section-topstories-loading = Loading...
-newtab-discovery-empty-section-topstories-unfinished = Oops! We almost loaded this section, but not quite.
+newtab-discovery-empty-section-topstories-loading = Loading....
+# Displays when a layout in a section took too long to fetch articles.
+newtab-discovery-empty-section-topstories-timed-out = Oops! We almost loaded this section, but not quite.
 
 ## Pocket Content Section.
 
--- a/locales-src/newtab.ftl
+++ b/locales-src/newtab.ftl
@@ -154,7 +154,7 @@ newtab-empty-section-topstories = You’ve caught up. Check back later for more 
 newtab-discovery-empty-section-topstories-header = You are caught up!
 newtab-discovery-empty-section-topstories-content = Check back later for more stories.
 newtab-discovery-empty-section-topstories-try-again-button = Try Again
-newtab-discovery-empty-section-topstories-loading = Loading....
+newtab-discovery-empty-section-topstories-loading = Loading…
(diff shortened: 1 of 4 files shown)


locales-src/newtab.ftl
```diff
@@ -148,6 +148,15 @@ newtab-empty-section-highlights = Start browsing, and we’ll show some of the g
 #  $provider (String): Name of the content provider for this section, e.g "Pocket".
 newtab-empty-section-topstories = You’ve caught up. Check back later for more top stories from { $provider }. Can’t wait? Select a popular topic to find more great stories from around the web.
 
+
+## Empty Section (Content Discovery Experience). These show when there are no more pocket stories or when some stories fail to load.
```
flodolo: nit: Pocket

locales-src/newtab.ftl
```diff
@@ -148,6 +148,15 @@ newtab-empty-section-highlights = Start browsing, and we’ll show some of the g
...
+
+newtab-discovery-empty-section-topstories-header = You are caught up!
+newtab-discovery-empty-section-topstories-content = Check back later for more stories.
+newtab-discovery-empty-section-topstories-try-again-button = Try Again
+newtab-discovery-empty-section-topstories-loading = Loading...
```
flodolo: Don't use 3 dots for ellipsis.

locales-src/newtab.ftl
```diff
@@ -148,6 +148,15 @@ newtab-empty-section-highlights = Start browsing, and we’ll show some of the g
...
+newtab-discovery-empty-section-topstories-unfinished = Oops! We almost loaded this section, but not quite.
```
flodolo: When is this one displayed?

JeaneC: We're unsure how to mention that we have two experiences for a section (very similar and minor differences).

We put a group header here called "Empty Section (Content Discovery Experience)" and there's a group just above this called "Empty Section States". 

I can copy the other section over similarly
## Empty Section States (Content Discovery Experience): These show when there are no more items in a section. Ex. When there are no more Pocket story recommendations, in the space where there would have been stories, this is shown instead.

flodolo: That doesn't really answer my question. Is this displayed when there's an error? Is this displayed together with which strings? Why "unfinished"?

Mardak: This landed as part of #5085 to show a message when we have a layout for a section but took too long / timed out to get the articles.
Screen Shot 2019-07-13 at 10 54 53 PM

flodolo: Then just need a comment to explain that, no need to add it to the section comment.

locales-src/newtab.ftl
JeaneC: Did you mean use 4 dots or use a style property to add ellipsis?

locales-src/newtab.ftl
JeaneC: Ah I misunderstood! Sounds good - I'll change from "unfinished" to "timed-out" and I'll add a comment.

locales-src/newtab.ftl
```diff
@@ -148,6 +148,16 @@ newtab-empty-section-highlights = Start browsing, and we’ll show some of the g
 #  $provider (String): Name of the content provider for this section, e.g "Pocket".
 newtab-empty-section-topstories = You’ve caught up. Check back later for more top stories from { $provider }. Can’t wait? Select a popular topic to find more great stories from around the web.
 
+
+## Empty Section (Content Discovery Experience). These show when there are no more stories or when some stories fail to load.
+
+newtab-discovery-empty-section-topstories-header = You are caught up!
+newtab-discovery-empty-section-topstories-content = Check back later for more stories.
+newtab-discovery-empty-section-topstories-try-again-button = Try Again
+newtab-discovery-empty-section-topstories-loading = Loading....
```
flodolo: I meant proper ellipsis, not four dots. If you try to land this in Firefox, it will fail tests.

PR #5183, opened 2019-07-15
Bug 1564568 - Hide What's New toolbar button after the user dismisses the panel

piatra:
The branch is on top of #5180 that's why the extra part, only changes in `ib/ToolbarPanelHub.jsm` should be considered.


piatra:
Rebased now that the other PR landed.

PR #5184, opened 2019-07-15
Bug 1566142 - Pref change for DS

ScottDowne:


Mardak:
Do we want to make this change for nightly too or beta only?

ScottDowne:
Just beta, is my understanding. Just stopping it from "riding the trains"

Mardak:
This PR should be against the `beta` branch then. Someone will need to land a patch against mozilla-beta too

PR #5186, opened 2019-07-15
Bug 1554961 - Adding a Top Site/Search Engine in a collapsed section is buggy.

JeaneC:
Original bug
https://bug1554961.bmoattachments.org/attachment.cgi?id=9067993

Dispatching an additional action to expand if the current action is adding a top site or search engine.
https://drive.google.com/file/d/1GpSV2pAJ2sMCBbHaqoGXqN1J5v_M29Fs/view?usp=sharing

### DIFF
--- a/content-src/components/SectionMenu/SectionMenu.jsx
+++ b/content-src/components/SectionMenu/SectionMenu.jsx
@@ -26,6 +26,18 @@ const WEBEXT_SECTION_MENU_OPTIONS = [
 ];
 
 export class _SectionMenu extends React.PureComponent {
+  dispatchAction(action, userEvent) {
+    this.props.dispatch(action);
+    if (userEvent) {
+      this.props.dispatch(
+        ac.UserEvent({
+          event: userEvent,
+          source: this.props.source,
+        })
+      );
+    }
+  }
+
   getOptions() {
     const { props } = this;
 
@@ -51,15 +63,19 @@ export class _SectionMenu extends React.PureComponent {
         const { action, id, type, userEvent } = option;
         if (!type && id) {
           option.onClick = () => {
-            props.dispatch(action);
-            if (userEvent) {
-              props.dispatch(
-                ac.UserEvent({
-                  event: userEvent,
-                  source: props.source,
-                })
-              );
+            const addingElements =
+              userEvent === "MENU_ADD_TOPSITE" ||
+              userEvent === "MENU_ADD_SEARCH";
+
+            if (props.collapsed && addingElements) {
+              const {
+                action: expandAction,
+                userEvent: expandEvent,
+              } = SectionMenuOptions.ExpandSection(props);
+              this.dispatchAction(expandAction, expandEvent);
             }
+
+            this.dispatchAction(action, userEvent);
           };
         }
         return option;
--- a/content-src/components/SectionMenu/SectionMenu.jsx
+++ b/content-src/components/SectionMenu/SectionMenu.jsx
@@ -26,7 +26,8 @@ const WEBEXT_SECTION_MENU_OPTIONS = [
 ];
 
 export class _SectionMenu extends React.PureComponent {
-  dispatchAction(action, userEvent) {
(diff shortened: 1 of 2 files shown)


JeaneC:
Looks like we also don't support animation easing of expanding or closing a section when done in the ContextMenu compared to clicking the section header.

content-src/components/SectionMenu/SectionMenu.jsx
```diff
@@ -51,15 +63,19 @@ export class _SectionMenu extends React.PureComponent {
...
+            const addingElements =
+              userEvent === "MENU_ADD_TOPSITE" ||
+              userEvent === "MENU_ADD_SEARCH";
+
+            if (props.collapsed && addingElements) {
+              const {
+                action: expandAction,
+                userEvent: expandEvent,
+              } = SectionMenuOptions.ExpandSection(props);
+              this.dispatchAction(expandAction, expandEvent);
```
punamdahiya: How about taking out L71 -L75 outside in a separate handleAddWhileCollapsed function. That way we don't need to write a separate dispatchAction

content-src/components/SectionMenu/SectionMenu.jsx
```diff
@@ -51,15 +63,19 @@ export class _SectionMenu extends React.PureComponent {
...
           option.onClick = () => {
-            props.dispatch(action);
-            if (userEvent) {
-              props.dispatch(
-                ac.UserEvent({
-                  event: userEvent,
-                  source: props.source,
-                })
-              );
+            const addingElements =
```
punamdahiya: nit: more readable if boolean value variable name starting with is, has... We can rename to something like hasAddEvent

test/unit/content-src/components/SectionMenu.test.jsx
```diff
@@ -164,17 +164,6 @@ describe("<SectionMenu>", () => {
...
-      "Separator",
-      "MoveUp",
-      "MoveDown",
-      "RemoveSection",
-      "CollapseSection",
-      "ExpandSection",
-      "ManageSection",
-      "AddTopSite",
-      "PrivacyNotice",
-    ];
```
punamdahiya: Are we removing this because its not used or its breaking test?

test/unit/content-src/components/SectionMenu.test.jsx
```diff
@@ -238,4 +223,40 @@ describe("<SectionMenu>", () => {
...
+          option.onClick();
+
+          const [expandCallArgs] = dispatch.firstCall.args;
+          const { type, data } = expandCallArgs;
+          assert.ok(type === "UPDATE_SECTION_PREFS");
+          assert.deepEqual(data, expectedExpandData);
+        });
+        it(`should dispatch the add topsite action after`, () => {
+          option.onClick();
+          assert.ok(dispatch.thirdCall.calledWith(option.action));
```
punamdahiya: we can assert dispatch is called twice assert.calledTwice(dispatch); and check first and second call args similar to dispatch.secondCall.args in L219 above

content-src/components/SectionMenu/SectionMenu.jsx
JeaneC: Sounds good!

test/unit/content-src/components/SectionMenu.test.jsx
JeaneC: Removing because its not being used. In section menu we don't have a prop for propOptions but just base the default options on the webextension and call it propOptions in the component

https://github.com/mozilla/activity-stream/blob/0379c988b17bcdef56bd8ac880368b0e636a2035/content-src/components/SectionMenu/SectionMenu.jsx#L29-L34

test/unit/content-src/components/SectionMenu.test.jsx
JeaneC: Yeah we can do something similar. Dispatch here will get called four times so I'll check each pair of two for the action and then the userEvent

PR #5187, opened 2019-07-16
Bug 1561625 - Add telemetry events for What's New panel

piatra:
Adds pings for:
* Showing the panel 
  * Either from the application menu
  * or from the toolbar button
* Clicking on a message
* Seeing a message (impression)
  * the idea here is that messages could be user targeted (not everyone sees the same set)
  * my current implementation sends 1 ping per message with the `message.id` and the `IMPRESSION` event the very first time you open the panel (no mater how many times you open the panel)
  * @ncloudioj is there a different way to report on this that would make it easier for data reporting?

### DIFF
--- a/lib/ToolbarPanelHub.jsm
+++ b/lib/ToolbarPanelHub.jsm
@@ -152,7 +152,12 @@ class _ToolbarPanelHub {
     // Panel impressions are not associated with one particular message
     // but with a set of messages. We concatenate message ids and send them
     // back for every impression.
-    const eventId = { id: messages.map(({ id }) => id).join(",") };
+    const eventId = {
+      id: messages
+        .map(({ id }) => id)
+        .sort()
+        .join(","),
+    };
     // Check `mainview` attribute to determine if the panel is shown as a
     // subview (inside the application menu) or as a toolbar dropdown.
     // https://searchfox.org/mozilla-central/rev/07f7390618692fa4f2a674a96b9b677df3a13450/browser/components/customizableui/PanelMultiView.jsm#1268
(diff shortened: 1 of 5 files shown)


piatra:
We should:
* Use a ping id made up of the comma joined message ids seen by the user
* Report every single impression not just the first one

piatra:
Preview of the impression ping:
```
"message_id":"WHATS_NEW_70_2,WHATS_NEW_70_1,WHATS_NEW_69_1",
"event":"IMPRESSION",
"value":{"view":"application_menu"}
```

lib/ToolbarPanelHub.jsm
```diff
@@ -132,19 +139,31 @@ class _ToolbarPanelHub {
...
       }
     }
 
-    // TODO: TELEMETRY
     this._onPanelHidden(win);
+
+    // Panel impressions are not associated with one particular message
+    // but with a set of messages. We concatenate message ids and send them
+    // back for every impression.
+    const eventId = { id: messages.map(({ id }) => id).join(",") };
```
ncloudioj: Could you sort all the message IDs before joining them together? That'll make the analysis (such `SELECT distinct(message_id)...`) easier.

lib/ToolbarPanelHub.jsm
```diff
@@ -261,6 +280,30 @@ class _ToolbarPanelHub {
...
+      win &&
+      !PrivateBrowsingUtils.isBrowserPrivate(
+        win.ownerGlobal.gBrowser.selectedBrowser
+      )
+    ) {
+      this._sendTelemetry({
+        message_id: message.id,
+        bucket_id: message.id,
+        event,
+        value: options.value,
```
ncloudioj: Let's send an empty Object for `value` if `optins.value` is absent. I.e. s/`option.value`/`option.value || {}`/.

ncloudioj: Turns out, JSON.stringify automatically omits the properties with `undefined` values, this is no longer required.

PR #5188, opened 2019-07-16
Bug 1562277 - Spoc dismiss flow updates.

ScottDowne:
To test:

1. Ensure your block list is empty.
2. Load a new tab with spocs.
3. Dismiss a spoc, it should be replaced with the next, non spoc card.
4. Refresh, a new, spoc, but not the one you dismissed should be back.


content-src/lib/selectLayoutRender.js
```diff
@@ -30,8 +30,10 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
 
       if (rickRoll <= spocsConfig.probability) {
         spocIndex++;
-        recommendations.splice(position.index, 0, spoc);
-        chosenSpocs.add(spoc);
+        if (!spocs.blocked.includes(spoc.url)) {
+          recommendations.splice(position.index, 0, spoc);
+          chosenSpocs.add(spoc);
+        }
```
punamdahiya: Including @ncloudioj to verify change here of increment spocIndex with possibility of not updating chosenSpocs impact outOfPositionSpocsFill telemetry on line 203 below

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1182,6 +1182,21 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           const filtered = spocsList.filter(s => s.url === action.data.url);
           if (filtered.length) {
             this._sendSpocsFill({ blocked_by_user: filtered }, false);
+
+            // If we're blocking a spoc, we want a slightly different treatment for open tabs.
+            this.store.dispatch(
+              ac.AlsoToPreloaded({
```
punamdahiya: why we are not using BroadcastToContent for spoc dismissal? How AlsoToPreloaded differs from BroadcastToContent

test/unit/common/Reducers.test.js
```diff
@@ -980,6 +981,21 @@ describe("Reducers", () => {
       });
       assert.deepEqual(state.spocs, INITIAL_STATE.DiscoveryStream.spocs);
     });
+    it("should add blocked spocs to blocked array with DISCOVERY_STREAM_SPOC_BLOCKED", () => {
+      const firstState = DiscoveryStream(undefined, {
+        type: at.DISCOVERY_STREAM_SPOC_BLOCKED,
+        data: { url: "https://foo.com" },
```
punamdahiya: what will happen if action.data is empty or null

content-src/lib/selectLayoutRender.js
ncloudioj: Confirmed that `spocIndex++` is still necessary here regardless whether or not this spoc was blocked.  See [this comment](https://github.com/mozilla/activity-stream/pull/5188/files#diff-91e34b3436d635509e2017c193c45ec1R185) for more details.

lib/DiscoveryStreamFeed.jsm
ScottDowne: AlsoToPreloaded got to just main and preloaded, BroadcastToContent go to everything.

So in this case, everything gets the `DISCOVERY_STREAM_SPOC_BLOCKED` event, but existing tabs do not get the `DISCOVERY_STREAM_LINK_BLOCKED` event.

test/unit/common/Reducers.test.js
ScottDowne: It does the same thing as `DISCOVERY_STREAM_LINK_BLOCKED`

They both assume action.data.url exists, iirc.

lib/DiscoveryStreamFeed.jsm
ScottDowne: To be more clear, if action.data is undefined, both events fail in the same way, when they try to access `action.data.url`, and hit a js error. If action.data is there, but no url, they both fail at doing their job but shouldn't throw a js error.

Both cases are currently unexpected to happen.

ScottDowne: Actually, it would fail even sooner on `const filtered = spocsList.filter(s => s.url === action.data.url);` above in this function.

Also just noticed I commented to the wrong comment.

test/unit/common/Reducers.test.js
ScottDowne: To be more clear, if action.data is undefined, both events fail in the same way, when they try to access `action.data.url`, and hit a js error. If action.data is there, but no url, they both fail at doing their job but shouldn't throw a js error.

Both cases are currently unexpected to happen.

Actually, it would fail even sooner on `const filtered = spocsList.filter(s => s.url === action.data.url);` from `case at.PLACES_LINK_BLOCKED:`

lib/DiscoveryStreamFeed.jsm
punamdahiya: cool, and thats because we don't want blocked spoc to be filtered out from prevState.spocs.data.spocs , so that we can use spocs state to find blocked spoc position in selectlayoutrender.

content-src/lib/selectLayoutRender.js
punamdahiya: Thanks @ncloudioj for clarifying, so its safe to assume spocsFill telemetry is not impacted with above change and we should increment and keep spocIndex++

lib/DiscoveryStreamFeed.jsm
ScottDowne: Correct.

punamdahiya:
Steps for QA to verifiy spocs pings are getting sent can be found here:

https://github.com/mozilla/activity-stream/pull/4928#issue-273142891

PR #5189, opened 2019-07-16
(nobug) - Removing unused subtitle props for DS Message

gvn:


gvn:
Whoever has time to review...thx!

gvn:
Subtitle may be coming back...holding for now.

PR #5191, opened 2019-07-17
Bug 1566372 - Listen for pref changes as a way to trigger notifications

piatra:
To test for this have

Pref | Value
--- | ---
browser.newtabpage.activity-stream.asrouter.providers.panel_local_testing | {"id":"panel_local_testing", "cohort": "SHOW_TEST"}
browser.messaging-system.whatsNewPanel.enabled | false

Then when changing `browser.messaging-system.whatsNewPanel.enabled` to `true` the Whats New gift icon should appear.

### DIFF
--- a/lib/ToolbarBadgeHub.jsm
+++ b/lib/ToolbarBadgeHub.jsm
@@ -90,6 +90,7 @@ class _ToolbarBadgeHub {
     switch (id) {
       case "show-whatsnew-button":
         ToolbarPanelHub.enableToolbarButton();
+        ToolbarPanelHub.enableAppmenuButton();
         break;
     }
   }
--- a/lib/ToolbarBadgeHub.jsm
+++ b/lib/ToolbarBadgeHub.jsm
@@ -15,6 +15,11 @@ ChromeUtils.defineModuleGetter(
 );
 ChromeUtils.defineModuleGetter(
   this,
+  "Services",
+  "resource://gre/modules/Services.jsm"
+);
+ChromeUtils.defineModuleGetter(
+  this,
   "setTimeout",
   "resource://gre/modules/Timer.jsm"
 );
@@ -41,6 +46,9 @@ class _ToolbarBadgeHub {
     this.id = "toolbar-badge-hub";
     this.template = "toolbar_badge";
     this.state = null;
+    this.prefs = {
+      WHATSNEW_TOOLBAR_PANEL: "browser.messaging-system.whatsNewPanel.enabled",
+    };
     this.removeAllNotifications = this.removeAllNotifications.bind(this);
     this.removeToolbarNotification = this.removeToolbarNotification.bind(this);
     this.addToolbarNotification = this.addToolbarNotification.bind(this);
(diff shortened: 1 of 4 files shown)


rlr:
I haven't tested it yet but I am pretty sure this needs to also call `ToolbarPanelHub.enableAppmenuButton()` somewhere/somehow.

piatra:
Good point this will only enable the toolbar button.

piatra:
@rlr updated, now we cleanup when the pref is turned off and the notification will enable both buttons.

lib/ToolbarPanelHub.jsm
```diff
@@ -34,12 +34,25 @@ class _ToolbarPanelHub {
...
     EveryWindow.unregisterCallback(TOOLBAR_BUTTON_ID);
     EveryWindow.unregisterCallback(APPMENU_BUTTON_ID);
+    Services.prefs.removeObserver(WHATSNEW_ENABLED_PREF, this);
+  }
+
+  observe(aSubject, aTopic, aPrefName) {
+    switch (aPrefName) {
+      case WHATSNEW_ENABLED_PREF:
+        if (!this.whatsNewPanelEnabled) {
+          this.uninit();
```
rlr: `uninit()` also removes the pref observer. So it would miss if the pref gets re-enabled after.

piatra: Fixing this case is a little more tricky because `init` happens in `ASRouter` where the `getMessages` method is passed down. In the end I decided it's not worth covering this scenario because it doesn't seem like something we'll run into.
You can either:
* have the pref enabled -> it gets disabled remotely and whats new will cleanup
* have the pref disabled -> it gets enabled remotely 

Since this is not a user facing pref going through these states multiple times is not possible.

lib/ToolbarBadgeHub.jsm
```diff
@@ -52,12 +60,23 @@ class _ToolbarBadgeHub {
...
+        this.messageRequest("toolbarBadgeUpdate");
+        break;
+    }
   }
 
   executeAction({ id }) {
     switch (id) {
       case "show-whatsnew-button":
         ToolbarPanelHub.enableToolbarButton();
+        ToolbarPanelHub.enableAppmenuButton();
```
rlr: This behaves oddly when using the local panel messages because their targeting is overriden such that the `isWhatsNewPanelEnabled` isn't checked in the targetting anymore. And neither of these two functions are checking the pref. So I am currently in a situation where I have the pref disabled but all the buttons are appearing. This is how I launched:

```
./mach run --setpref 'browser.newtabpage.activity-stream.asrouter.providers.panel_local_testing={"id":"panel_local_testing", "cohort": "SHOW_TEST"}' --setpref 'browser.messaging-system.whatsNewPanel.enabled=false' --setpref 'browser.newtabpage.activity-stream.asrouter.devtoolsEnabled=true'
```

lib/ToolbarPanelHub.jsm
```diff
@@ -34,12 +34,25 @@ class _ToolbarPanelHub {
     this._getMessages = getMessages;
     if (this.whatsNewPanelEnabled) {
       this.enableAppmenuButton();
+      // Listen for pref changes that could turn off the feature
+      Services.prefs.addObserver(WHATSNEW_ENABLED_PREF, this);
```
piatra: This should happen outside of the `if` statement as the pref will be `false` by default.

PR #5193, opened 2019-07-18
Bug 1547337 - Accessibility Audit

AllegroFox:


AllegroFox:
@MarcoZehe

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -217,20 +218,13 @@ export class CollapsibleSection extends React.PureComponent {
               <span
                 className="click-target"
                 role="button"
+                label={collapsed ? "Expand Section" : "Collapse Section"}
```
punamdahiya: should this be aria-label? Strings here should be coming from ftl file

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -129,6 +126,8 @@ export class ContextMenuItem extends React.PureComponent {
         this.focusSibling(event.target, event.key);
         break;
       case "Enter":
+      case " ":
```
punamdahiya: should this be "Space" https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

AllegroFox: I'll change that! I think it's in a few other places too so I'll do a quick check for them.

content-src/components/CollapsibleSection/CollapsibleSection.jsx
AllegroFox: I'm not sure if it should be aria-label; I made it `label` based on @MarcoZehe 's review [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1547337) - but I'll try it and test the behaviour. I don't think it could hurt to have it be aria-label!

content-src/components/ContextMenu/ContextMenu.jsx
AllegroFox: So I ran into an issue - event.code causes the listener to run into issues when checking for the Shift key, which we use to determine if a `Tab` press should close the menu or not. It seems like it's more reliable to use " " and event.key, but let me know what you think!

content-src/components/CollapsibleSection/CollapsibleSection.jsx
JeaneC: Second the part on changing hard-coding strings to fluent. Looks like you'll have to create a new fluent ID since using any of the current ones will override/delete all your children with the fluent string.
https://github.com/mozilla/activity-stream/blob/8dccdc7c295bc583fb76d10e399d2f438a502d42/locales-src/newtab.ftl#L123-L125

New string with property assignment instead like
```
newtab-section-collapse-section-label =
    .aria-label = Collapse Section
```

AllegroFox: I was just looking for that, thanks!

content-src/components/ContextMenu/ContextMenu.jsx
punamdahiya: ah ok, thats because these are React [synthetic events](https://reactjs.org/docs/events.html#keyboard-events) that doesn't support event.code

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -1,7 +1,3 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this file,
- * You can obtain one at http://mozilla.org/MPL/2.0/. */
-
```
punamdahiya: Did this got deleted by mistake?

AllegroFox: Must have, I'll add it back!

punamdahiya:
Running Accessibility check on devtools shows below feedback
h4 card-title

In Highlights, its complaining because of missing heading when title is blank. This is fixable by rendering h4 only when link.title has value.

AllegroFox:
The Fluent aria-label/title seems to correct the missing title, let me know!

content-src/components/CollapsibleSection/CollapsibleSection.jsx
punamdahiya: Reading the feedback above, should we add "Expand Section" or "Collapse Section " label to the arrow button only. My only concern is will this change make all section title read as expand or collapse section.
I am ok with this change if we can verify in screen reader or with @MarcoZehe that we are correctly reading titles of section e.g. TopSites, Recommended by Pocket ...

AllegroFox: Adding the label to the arrow button makes the final heading title/label "Top Sites Collapse Section", for example - I don't know if that's exactly what we want, but it does cover all the information at least.

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -126,7 +130,7 @@ export class ContextMenuItem extends React.PureComponent {
         this.focusSibling(event.target, event.key);
         break;
       case "Enter":
-      case " ":
+      case "Space":
```
punamdahiya: Is this working or a typo?

AllegroFox: typo :)

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -138,15 +137,22 @@ export class ContextMenuItem extends React.PureComponent {
     }
   }
 
+  onKeyUp(event) {
+    if (event.key === " ") {
+      event.preventDefault();
+    }
+  }
+
```
punamdahiya: @AllegroFox can you please help clarify why we need onKeyUp handling here. Thanks

AllegroFox:
There’s a default event listener for space onKeyUp that triggers buttons/scrolls the page - it ended up interfering with the contextMenu listener, causing the menus to open and then immediately close when triggered with space.

content-src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -77,6 +73,7 @@ export class ContextMenuItem extends React.PureComponent {
     super(props);
     this.onClick = this.onClick.bind(this);
     this.onKeyDown = this.onKeyDown.bind(this);
+    this.onKeyUp = this.onKeyUp.bind(this);
```
punamdahiya: Do we need tabIndex on ContextMenuItem L60 above?

content-src/components/CollapsibleSection/CollapsibleSection.jsx
punamdahiya: That should work, do we need title for that or should aria-label suffice? With title when mouse over alt text shows as 'Expand Section' not sure if thats desired.

punamdahiya:
> There’s a default event listener for space onKeyUp that triggers buttons/scrolls the page - it ended up interfering with the contextMenu listener, causing the menus to open and then immediately close when triggered with space.

ah I see it now, it will help if we can add comment inside to explain that. Thanks

content-src/components/CollapsibleSection/CollapsibleSection.jsx
AllegroFox: I'll leave it as just aria-label for now, we can always tweak it further.

content-src/components/ContextMenu/ContextMenu.jsx
AllegroFox: Doesn't look like we do. Removed!

AllegroFox:
That should be everything!

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -119,6 +119,7 @@ export class CollapsibleSection extends React.PureComponent {
 
   onKeyPress(event) {
     if (event.key === "Enter" || event.key === " ") {
+      event.preventDefault();
```
punamdahiya: I think this prevent default is fixing scroll of page

content-src/components/CollapsibleSection/CollapsibleSection.jsx
```diff
@@ -223,16 +224,13 @@ export class CollapsibleSection extends React.PureComponent {
...
-                onKeyPress={this.onKeyPress}
-                onClick={this.onHeaderClick}
-              >
                 {isCollapsible && (
                   <span
+                    data-l10n-id={
+                      collapsed
+                        ? "newtab-section-expand-section-label"
+                        : "newtab-section-collapse-section-label"
+                    }
```
punamdahiya: @AllegroFox Has this change been verified with screen reader i.e moving aria-label here instead of span with className click-target correctly reads heading and aria-label attached to child span e.g. when tabbed to TopSites as 'TopSites Collapse Section'

AllegroFox: That's what happens for me with VoiceOver - not sure about other screen readers.

locales-src/newtab.ftl
```diff
@@ -131,6 +131,13 @@ newtab-section-menu-move-up = Move Up
 newtab-section-menu-move-down = Move Down
 newtab-section-menu-privacy-notice = Privacy Notice
 
+## Section aria-labels
+
+newtab-section-collapse-section-label =
+    .aria-label = Collapse Section
+newtab-section-expand-section-label =
+    .aria-label = Expand Section
+
```
punamdahiya: Sorry missed Including @flodolo for strings review. @Mardak do you know why this PR didn't add @flodolo for review by default for newtab.ftl?

Mardak: Only strings.properties is listed:
https://github.com/mozilla/activity-stream/blob/68a8907fd1f8c0f27b3983d0ff1b969859e43392/CODEOWNERS#L1-L2

Probably should just add the whole directory if we can.

punamdahiya: Created https://github.com/mozilla/activity-stream/pull/5202

PR #5196, opened 2019-07-22
Bug 1540842 - Adding "Saved to Pocket" parity to Discovery Stream

JeaneC:
Updating Discovery Stream reducers to add a property "badge" to stories to mark whether they've been Bookmarked/Saved To Pocket before, then showing that badge in the UI.

We had this property and feature under "type" in AS Cards but "type" is currently being used in DS for components (e.g "CardGrid").

Getting Feedback from @wolasi about changing our badge indicator after we remove a bookmark.
https://drive.google.com/file/d/1XnkHEN5lINyHtdzsF8FpuLbnVOfB301_/view?usp=sharing


punamdahiya:
@wolasi question on component type support, is it ok for List and Hero component to not have this messaging?

JeaneC:
Temporarily using the SVG from the Figma file until the new one comes.

@wolasi
https://drive.google.com/file/d/1AVbvkDZ9X1bYlhBbbhAFeptEXKNGNY-y/view?usp=sharing

wolasi:
> @wolasi question on component type support, is it ok for List and Hero component to not have this messaging?

Is it difficult to add it to the other content components?

wolasi:
> Temporarily using the SVG from the Figma file until the new one comes. Did we want an animation each time a new badge show ups?
> 
> @wolasi
> https://drive.google.com/file/d/1AVbvkDZ9X1bYlhBbbhAFeptEXKNGNY-y/view?usp=sharing

Yes please!

wolasi:
@JeaneC the animation in the video doesn't look quite right, the timing looks off. Might be the render so will reserve final review 'til I can see it in the browser

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -108,15 +103,47 @@ $excerpt-line-height: 20;
...
     }
 
-    .context,
+    .story-footer {
+      color: var(--newtab-text-secondary-color);
+      display: flex;
+      height: 24px;
+      inset-inline-start: 0;
+      margin-top: 12px;
+      align-items: center;
```
wolasi: Add
```
transition: transform .3s;
transform-style: preserve-3d;
```

content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -125,6 +152,22 @@ $excerpt-line-height: 20;
       font-size: 13px;
       color: $grey-50;
     }
+
+    .status-message {
+      display: flex;
+      align-items: center;
+      animation: rotateIn 1s ease;
```
wolasi: Let's try an ease-out or linear here

punamdahiya:
> > @wolasi question on component type support, is it ok for List and Hero component to not have this messaging?
> 
> @punamdahiya Is it difficult to add it to the other content components?

It is doable, it will help if we can get updated design on how to show messaging in hero and list view below:
list

wolasi:
First step would be to move source above the title like we did for cards. Then status messaging can go below the title with the same treatment as for cards. I've updated the spec to include example for list and hero components as reference.

punamdahiya:
> First step would be to move source above the title like we did for cards. Then status messaging can go below the title with the same treatment as for cards. I've updated the spec to include example for list and hero components as reference.

I will recommend implementing List and Hero support in a separate bug to keep this patch simpler.

wolasi:
> > First step would be to move source above the title like we did for cards. Then status messaging can go below the title with the same treatment as for cards. I've updated the spec to include example for list and hero components as reference.
> 
> I will recommend implementing List and Hero support in a separate bug to keep this patch simpler.

👍

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
```diff
@@ -0,0 +1,46 @@
...
+      this.setState({ hasUpdate: false });
+    }
+  }
+
+  render() {
+    const { context, badge, hasUpdate } = this.state;
+    const { icon, fluentID } = cardContextTypes[badge] || {};
+
+    return (
+      <footer className="story-footer">
```
punamdahiya: Should change it to use div. Footer element might not make sense for [accessibility which recommends ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Contentinfo_role)having one per webpage

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
```diff
@@ -0,0 +1,46 @@
...
+  <div className="status-message">
+    <span
+      aria-haspopup="true"
+      className={`story-badge-icon icon icon-${icon}`}
+    />
+    <div className="story-context-label" data-l10n-id={fluentID} />
+  </div>
+);
+
+export class DSFooter extends React.PureComponent {
```
punamdahiya: How about renaming DSFooter component to something which is not tied to placement and more with purpose something like ContextMessage

common/Reducers.jsm
```diff
@@ -634,6 +634,7 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
           return Object.assign({}, item, {
             open_url: action.data.open_url,
             pocket_id: action.data.pocket_id,
+            badge: "pocket",
```
punamdahiya: why badge and not type as used in Sections at.PLACES_SAVED_TO_POCKET handler

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
wolasi: I'm wary of tying it to a particular usecase at this stage since we're still experimenting. I think `ds-story-footer` is fine for now.

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
wolasi: Good catch!

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.badge !== prevState.badge) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ hasUpdate: false });
+    }
+  }
+
+  render() {
+    const { context, badge, hasUpdate } = this.state;
```
punamdahiya: Do we need context in state?

content-src/components/DiscoveryStreamComponents/DSFooter/_DSFooter.scss
```diff
@@ -0,0 +1,76 @@
...
+    line-height: 24px;
+    color: $grey-50;
+  }
+
+  .story-sponsored-label {
+    color: #737373;
+    transform: translateZ(11px);
+    transition-property: opacity;
+    transition-duration: 200ms;
+  }
```
punamdahiya: Do we need transition on Sponsored label?

data/content/assets/icon-removed-bookmark.svg
```diff
@@ -0,0 +1,3 @@
...
+</svg>
```
punamdahiya: running this with svgo will give optimized version of this svg @Mardak don't see svgo in our node_modules, is that a recent change?

Mardak:
`svgo` hasn't in package.json for a while.. removed in 38cd3cf8 #3389 although I've just had a version installed that I use. Perhaps we should just reintroduce the dev dependency and add a (yam)script to run it. ?

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      hasUpdate: prevState && prevState.badge !== nextProps.badge,
+      ...nextProps,
+    };
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.badge !== prevState.badge) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ hasUpdate: false });
```
punamdahiya: @mardak Here to display transition, there is a forced re-render by adding StatusMessage thats removed in initial render with hasUpdate as true inside getDerivedStateFromProps. This looks good to me, including you for React feedback and if it can be optimized further.

Mardak: What transition are you referring to? CSS animation should happen right away if you want something to start when rendered. From a quick test of selecting "Bookmark" from the menu then hitting escape. Also, seems like the message persists to the next new tab -- expected?

Jul-24-2019 14-50-56

Mardak: Also, with preloaded newtab, the message is already shown with no animation which may or may not be expected.

punamdahiya: My understanding animation is shown to catch attention of changing context message when user click on 'Bookmark' or 'Save to Pocket' in Context Menu. There is somework to add green color to text while animating that needs to be implemented.

In all other cases that includes preloaded tab or opening a new tab, message should persist and  should behave similar to 'Sponsored by ...' text.

punamdahiya: Just saw what you mean by " From a quick test of selecting "Bookmark" from the menu then hitting escape." experience captured in image above. Yeah that could be tricky. FWIW its parity and behave similar in AS :)

JeaneC:
@wolasi For dark theme should we use the same green color?
https://drive.google.com/file/d/1eGkuVyw3Lch0OTthJ__r2FK56ySyKw9g/view?usp=sharing
https://drive.google.com/file/d/19X1fDbkMZ69Pkrr_FHoKhBLRmlk4cYkc/view?usp=sharing

common/Reducers.jsm
JeaneC: As discussed offline its because type is reserved in DS to describe the component (i.e "Hero", "CardGrid"). Should we try to do a refactor after rewriting type => component_type and badge => type?

content-src/components/DiscoveryStreamComponents/DSFooter/_DSFooter.scss
JeaneC: Nope. Looks like this is some legacy css and we had a transition there before?

common/Reducers.jsm
punamdahiya: We can leave type in reducers.jsm as is, those context menu handlers should be removed with AS experience going away. I propose renaming badge to context_type thats closer to value it stores, wdyt?

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    font-size: 13px;
+    line-height: 24px;
+    color: $grey-50;
+  }
+
+  .story-sponsored-label {
+    color: $grey-50;
+  }
```
punamdahiya: I think this is already defined above and can be removed

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -0,0 +1,100 @@
...
+
+    .story-context-label {
+      color: $grey-50;
+      flex-grow: 1;
+      font-size: 13px;
+      line-height: 24px;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+      animation: color 2s ease-out;
```
punamdahiya: animation property for  .story-badge-icon and .story-context-label  can be moved below similar to dark-theme handling

content-src/components/DiscoveryStreamComponents/DSFooter/DSFooter.jsx
punamdahiya: To support not animating text in all other cases such as refresh and opening newtab  may be we should maintain [class animate](https://gist.github.com/punamdahiya/db6f603cff063bb977b778ce0388eab5) that should be applied to status-message when prevState inside getDerivedStateFromProps is not undefined.

JeaneC: Sounds like a good idea. Though, using prevState as undefined will prevent the first animation load from not working (from no badge to badge). Adding an extra case 

```
    return {
      // Only time we dont animate is when our first initial render (prevState == null)
      // has a context_type cached from the same browser session.
      shouldNotAnimate: !prevState && nextProps.context_type,
      hasUpdate: prevState && prevState.context_type !== nextProps.context_type,
      context_type: nextProps && nextProps.context_type,
    };
```

Might be confusing to have hasUpdate and shouldNotAnimate but I need hasUpdate to toggle the component (in the case of "Bookmark" to "Remove bookmark" to refresh the animation) and shouldNotAnimate to toggle the class (in the case of opening a new tab with the "Bookmark" badge already there from a previous session)

JeaneC:
@wolasi . Thoughts? Also, badges won't animate on refresh or newtab.

Dark Theme
https://drive.google.com/file/d/17VDhNX7pVBfyBm8j4Umeaig7081QDuGm/view?usp=sharing

Light Theme
https://drive.google.com/file/d/1y04zMkuEURTbGCOMWvUY6HYlU0zsP3bI/view?usp=sharing

test/unit/content-src/components/DiscoveryStreamComponents/DSContextFooter.test.jsx
```diff
@@ -0,0 +1,81 @@
...
+    assert.isOk(bookmarkStatusMessage);
+    assert.equal(
+      bookmarkStatusMessage.find("story-context-label").prop("data-l10n-id"),
+      bookmarkFluentID
+    );
+
+    const removeBookmarkBadge = "removedBookmark";
+    wrapper.setProps({ badge: removeBookmarkBadge });
+
+    const removedBookmarkStatusMessage = wrapper.find(".status-message");
```
punamdahiya: typo here '.status-message'

test/unit/content-src/components/DiscoveryStreamComponents/DSContextFooter.test.jsx
```diff
@@ -0,0 +1,81 @@
...
+  it("should render", () => {
+    assert.isTrue(wrapper.exists());
+    assert.isOk(wrapper.find(".story-footer"));
+  });
+
+  it("should render a badge if a proper badge prop is passed", () => {
+    wrapper = mount(<DSContextFooter context_type={bookmarkBadge} />);
+    const { fluentID } = cardContextTypes[bookmarkBadge];
+
+    const statusMessage = wrapper.find(".status-messag");
```
punamdahiya: typo here '.status-message'

JeaneC: 😵. That did it....

wolasi:
> @wolasi . Thoughts? Also, badges won't animate on refresh or newtab.
> 
> Dark Theme
> https://drive.google.com/file/d/17VDhNX7pVBfyBm8j4Umeaig7081QDuGm/view?usp=sharing
> 
> Light Theme
> https://drive.google.com/file/d/1y04zMkuEURTbGCOMWvUY6HYlU0zsP3bI/view?usp=sharing

in terms of transitioning between statuses, would it be possible to the do a continuous roll
e.g. first status rolls in, then when a new status is due, the first status rolls up and the new status rolls in
in the current implementation, the first status disappears then the new status rolls in

JeaneC:
@wolasi Here's an attempt at it. I continued the rotation (to enter your codepen had it at -90deg to 0 and to leave I made it 0 to 90deg).

https://drive.google.com/file/d/1vv9XyRAHyzj_vk5QehPNLdq0qEq_x4d3/view?usp=sharing

JeaneC:
@wolasi with the Fade In we discussed offline
Main - https://drive.google.com/open?id=1KMupeqZo2xPjOu8aI6zGnpv-0A6amz6R
Main w/ View Count - https://drive.google.com/file/d/151j-ykvDe-0vkLYKDRkj0i9lTT91xIiz

Others....
Hero - https://drive.google.com/open?id=13OLAkYK308tUTr3iOXD7Brq21FJB3FzD
List - https://drive.google.com/file/d/16ccUntKJ_jPb7HZDyOzgLjOZBLWDCsFC/
Hero w/ View Count - https://drive.google.com/open?id=1JCEz5iiPT02WWJ1bcYzzy8cUts8x0SHF

wolasi:
@JeaneC the timing looks a little fast, but it might be the video render.

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -0,0 +1,111 @@
...
+$status-dark-green: #7C6;
+
+.story-footer {
+  color: var(--newtab-text-secondary-color);
+  display: flex;
+  height: 24px;
+  inset-inline-start: 0;
+  margin-top: 12px;
+  align-items: center;
+  transition: transform 0.3s;
```
wolasi: We can lose line 11 and 12 since we aren't using the rotation effect

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+.story-animate-enter {
+  opacity: 0;
+}
+
+.story-animate-enter-active {
+  opacity: 1;
+  transition: opacity 150ms cubic-bezier(0.07, 0.95, 0, 1) 300ms;
```
wolasi: Let's use the ease-in timing function here, the cubic bezier is too harsh.

JeaneC: Ease-in looks more noticeable here (also switched to recording in Quicktime)
https://drive.google.com/open?id=1ZMBgZJIGvR9e-D_sTJ_lAjkG2YnTuW7V

wolasi: Looking good!

package.json
```diff
@@ -12,6 +12,7 @@
     "react": "16.8.6",
     "react-dom": "16.8.6",
     "react-redux": "7.0.3",
+    "react-transition-group": "^4.2.1",
```
punamdahiya: Including @dmose to help review the new package added in the build. Thanks!

JeaneC: Yes! It looks like this is officially endorsed by React as its mentioned in the official docs https://reactjs.org/docs/animation.html. 

It is under community support at https://github.com/reactjs/react-transition-group with a BSD-3 license. The addon itself significantly simplifies animations that are based on state/prop changes especially when it comes to mounting/unmounting.

```
Exposes simple components useful for defining entering and exiting transitions. React Transition Group is not an animation library like React-Motion, it does not animate styles by itself. Instead it exposes transition stages, manages classes and group elements and manipulates the DOM in useful ways, making the implementation of actual visual transitions much easier.
```

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -0,0 +1,109 @@
+$status-green: #058B00;
+$status-dark-green: #7C6;
+
+.story-footer {
+  color: var(--newtab-text-secondary-color);
+  display: flex;
+  height: 24px;
```
punamdahiya: Can we get rid of height, this is keeping extra 24px blank space when we don't have sponsored context or messaging

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -0,0 +1,109 @@
...
+      @include dark-theme-only {
+        fill: $grey-40;
+      }
+
+      fill: $grey-50;
+      height: 16px;
+      margin-inline-end: 6px;
+
+      &.icon-bookmark-removed {
+        background-image: url('../data/content/assets/icon-removed-bookmark.svg');
```
punamdahiya: can update path to use url('#{$image-path}

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
punamdahiya: Also, I think its safe to remove display: flex and align-items: center as we are making status-message flex container

content-src/components/DiscoveryStreamComponents/DSContextFooter/DSContextFooter.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+export class DSContextFooter extends React.PureComponent {
+  render() {
+    const { context, context_type } = this.props;
+    const { icon, fluentID } = cardContextTypes[context_type] || {};
+
+    return (
+      <div className="story-footer">
+        {context && <p className="story-sponsored-label">{context}</p>}
+        <TransitionGroup component={null}>
```
punamdahiya: we can move TransitionGroup inside !context check as we don't need it for sponsored stories

JeaneC: This is true but we're somehow losing the first badge animation by doing this. My hunch is that putting this in the !context check makes React render TransitionGroup and CSSTransition at the sametime, where instead we would need TransitionGroup to preexist to render incoming and outgoing CSS transitions.

This doesn't render an extra div though if we set component={null}

JeaneC: Alternatively... We can move sponsored context out of here (bring it back to DSCard) since in the Hero component the Sponsored Text is not in the footer but in the header.

https://github.com/mozilla/activity-stream/blob/20eb9c15127a2185d75fdc4358acee80a372f54b/content-src/components/DiscoveryStreamComponents/Hero/Hero.jsx#L104-L114

Later PR w/ Hero & List
https://github.com/mozilla/activity-stream/pull/5206/commits/f90ef43060abebd8709550c8e456921d6842d041 using DSFooter

content-src/components/DiscoveryStreamComponents/DSContextFooter/_DSContextFooter.scss
```diff
@@ -34,7 +33,7 @@ $status-dark-green: #7C6;
       margin-inline-end: 6px;
 
       &.icon-bookmark-removed {
-        background-image: url('../data/content/assets/icon-removed-bookmark.svg');
+        background-image: url(#{$removed-bookmark-icon});
```
punamdahiya: oh, what I meant was use global variable $image-path, something like url('#{$image-path}icon-removed-bookmark.svg')

package.json
dmose: In general, this seems reasonable.  Since it seems to have close ties to the React community, decent test coverage, and what looks to be a compatible license.  @k88hudson, any thoughts?
 
Assuming that Kate's ok with it, there are a few things that would need to happen to land this change:

- [ ] file a license bug as described at https://www.mozilla.org/en-US/MPL/license-policy/ to get formal approval from the licensing team
- [ ] split the vendoring changes into their own commit (and don't squash it, perhaps by putting it into its own PR)
- [ ] for some reason, it's not listed in package-lock.json, which makes me think it's not installed correctly.  Pin it to an exact version ("4.2.1") rather than the less specific "^4.2.1", probably by using the `--save-exact` switch to npm.  Make sure you're using npm 6.9.0 to do the install.
- [ ] post a patch to [`toolkit/content/license.html](https://searchfox.org/mozilla-central/source/toolkit/content/license.html) add the license in appropriately.  This patch could be posted to its own bug, or to the bug for the upcoming export, if there is one.

Does the CSS transition group stuff in this patch, as written, work correctly in the running browser?  Since it doesn't contain a vendored copy for the runtime, I would sort of expect it to fail.  If that's true, it needs to be added to the bin/vendor.js script and that needs to be run once as well.

punamdahiya: @dmose Looking at activity-stream.bundle.js, react-transition-group is getting added as concatenated module similar to fluent, fluent-react and cached-iterable. How webpack is picking and bundling this module is still not clear to me.

punamdahiya:
Verified changes related to  adding the package react-transition-group looks good. @dmose including you for final review of new package related changes. Thanks

k88hudson:
I looked over this with @punamdahiya and @JeaneC, the approach looks good to me;

One comment I would add is to cross-reference the `3s` constant via comments to remind any future authors to update both files

dmose:
r=dmose for including the react-transition-group library.

PR #5197, opened 2019-07-23
Testing a sponsored section flow

ScottDowne:


common/Reducers.jsm
```diff
@@ -63,6 +63,7 @@ const INITIAL_STATE = {
     spocs: {
       spocs_endpoint: "",
       lastUpdated: null,
+      showSpocs: false,
```
ScottDowne: This is needed so the jsx gets a better more explicit handle on if spocs are on or not.

common/Reducers.jsm
```diff
@@ -606,6 +607,7 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
           ...prevState,
           spocs: {
             ...prevState.spocs,
+            showSpocs: action.data.showSpocs,
```
ScottDowne: This is just part of the above.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -301,11 +302,8 @@ export class _DiscoveryStreamBase extends React.PureComponent {
     const styles = [];
     return (
       <div className="discovery-stream ds-layout">
-        {layoutRender.map((row, rowIndex) => (
-          <div
-            key={`row-${rowIndex}`}
-            className={`ds-column ds-column-${row.width}`}
-          >
+        {layoutRender.map((row, rowIndex) => {
+          const contents = (
```
ScottDowne: Changes in this file are wrapping a collection or component in DSDismiss stub.

content-src/lib/selectLayoutRender.js
```diff
@@ -139,19 +139,29 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
     return { ...component, data };
   };
 
+  const filterComponent = (c) => {
```
ScottDowne: Changes in this file are removing any components or collections that are sponsored if spocs are disabled.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1286,10 +1287,90 @@ defaultLayoutResp = {
     {
       width: 12,
       components: [
+        {
```
ScottDowne: This is all test data.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -527,6 +527,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     sendUpdate({
       type: at.DISCOVERY_STREAM_SPOCS_UPDATE,
       data: {
+        showSpocs: this.showSpocs,
```
ScottDowne: This is needed to make the jsx better understand if spocs are on or off.

content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+
+export class DSDismiss extends React.PureComponent {
+  render() {
+    // TODO:
+    // This needs an x button that dismisses the campaign_id passed to it.
```
punamdahiya: Clarifying campaign_id, are we saying any LIST, GRID, HERO, COLLECTION TEXT BANNER and TOP_SITE that has campaign_id is promotion? 
And if any of these above component has paid record/item (even one) it becomes paid promotion or do we need all items to be sponsored ?
May be @wolasi question, are non-paid promotion dismissable?

content-src/lib/selectLayoutRender.js
```diff
@@ -139,19 +139,29 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
...
+    return (
+      !filterArray.includes(c.type) &&
+      (!c.sponsored || spocs.showSpocs)
```
punamdahiya: ah I see you have  sponsored property to indicate a campaign is paid promotion.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1286,10 +1287,90 @@ defaultLayoutResp = {
...
+          properties: {},
+        },
+      ],
+    },
+    {
+      width: 12,
+      components: [
+        {
+          type: "Message",
+          campaign_id: "123456",
```
punamdahiya: Is Message component has special handling by assigning campaign_id. Clarifying how campaign_id here is different from campaign_id in Line 1356 below

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
punamdahiya: clarifying collection or component , Only possible component allowed here are LIST, GRID, HERO, COLLECTION ,TEXT BANNER and TOP_SITE or there can be more?

punamdahiya:
Overall this should work, I guess the biggest assumption this approach has is there is no way to individually identify a record is a promotion content or publisher content. All the items in a collection if it has campaign_id are promotion content and we are not allowing a collection e.g. with CardGrid have both promotion content and publisher content in it which definitely simplifies complexity.

content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
wolasi: At this stage a sponsored collection can either have paid items or non-paid items. We should be able to test both options.

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
ScottDowne: Any component can be in a collection, and would be something that if the collection is dismissed, it would take everything with it, so for example, if a collection contains a component that's not something that can be sponsored, for the sake of that collection, it's sponsored because it's used along with a collection that is only sponsored.

In the case of a single component that's sponsored, any component that's got a campaign id (is a promo) would become dismiss-able. I don't think this is something we need to enforce on our end.

A component or collection can be dismiss-able via a campaign id (or some other id) it then wraps it in a dsdismiss component. It's up to the server side data to use these components in a way that creates a logical experience to the user.

content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
ScottDowne: I think the work involved in making non paid dismiss-able is the same, and we should just do it as well.

I think there may be confusion around the naming of campaign_id. From what I'm seeing, we don't really care if a campaign is paid or not, maybe it makes more sense to change some of the existing names in the code, but how I see it without changing the names in the code.

campaign === promo
sponsored === paid

So if something is sponsored and campaign, it's a paid promo.
If something is campaign and not sponsored, it's a non paid promo.
If something is not a campaign and not sponsored, it's an organic rec.
If something is not a campaign and sponsored, it's a bug and shouldn't happen.

ScottDowne: To clarify what I meant by "we don't really care if a campaign is paid or not"

If something is a campaign, we make it dismiss-able, at this point we're done, and if it's paid or not, that's different functionality. It's logically split up.

If something is sponsored, we turn it on or off based on if spocs are enabled or not.

We only really need campaign_id, so we can block it based on something.

content-src/lib/selectLayoutRender.js
ScottDowne: Yeah, I think one bit that's confusing is anything that comes from the old spocs endpoint, doesn't have the `sponsored: true` because it explicitly came from a sponsored endpoint so we can assume it's all spocs.

same is true for any feed data coming from a card grid in a sponsored collection. We can assume that's all sponsored content.

lib/DiscoveryStreamFeed.jsm
ScottDowne: Line 1356 is to make a whole collection dissmiss-able, and everything it wraps, line 1337 is to make just a single component dismiss-able.

ScottDowne: In the above example of a component being dismiss-able, and that component being a message component, we basically have a wrapper for a promo text component.

content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
punamdahiya: > If something is not a campaign and sponsored, it's a bug and shouldn't happen.

if something is not a campaign aka promo and sponsored , isn't that organic sponsored record or same as sponsored story we show now in nightly?

punamdahiya: > At this stage a sponsored collection can house content components that either have paid items or non-paid items. We should be able to test both options.

@wolasi clarifying your above comment with examples. By house content components you mean e.g. 2 Card Grids, one CardGrid has all paid promo items and second CardGrid has all non-paid promo items.  Is there a possibility where one CardGrid will have to display both paid promo and non-paid promo item?

content-src/lib/selectLayoutRender.js
punamdahiya: is it safe to assume any feed data coming from cardgrid in a sponsored collection, will have context property set , thats used in DSCard UI to display 'Sponsored by ...' label. Just want to highlight that my WIP Card UI variant patch looks for context property against an item and doesn't handle sponsored:true set against a component

content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
wolasi: No not quite, here's an example of some different way a collection could be constructed https://www.figma.com/file/eGdZVYlZ5evvePHUCtKnYg7u/New-Tab-Spec-for-Fx-70?node-id=550%3A0

The basic idea is a collection can contain a single content component or a combination of them, with the ability to have paid promo item in the position we want it.

Happy to hop on a call to talk this thru if needed.

punamdahiya: that helps, is it possible to update above collection promo figma with spoc UI variants from https://www.figma.com/file/eGdZVYlZ5evvePHUCtKnYg7u/New-Tab-Spec-for-Fx-70?node-id=76%3A167

wolasi: Done! Is one example enough?

punamdahiya: thanks @wolasi yes it suffice. @ScottDowne I think current patch doesn't support mixing promo and organic records in a CardGrid component as shown in collection figma. I think we assume all content inside a promo CardGrid component will be promotion only.

PR #5199, opened 2019-07-24
Bug 1568262 - Removed unused onboarding strings

AllegroFox:


JeaneC:
I think some of these strings are still in the test cases

https://searchfox.org/mozilla-central/search?q=onboarding-button-label-try-now&path=

PR #5204, opened 2019-07-26
Bug 1568932 - Use idle callback for measurements

gvn:


ScottDowne:
I actually think the best way to test this is via the ff profiler, I can probably do that.

A few concerns I have though.

1. Doing `requestIdleCallback` for the state set for all images might not help, generally above the fold loads need to be asap to get to settled asap, and below the fold can then be via idle. This does it for all cases which would potentially cause above the fold images to wait for other below the fold non idle tasks.
2. In my previous profile tests, the delay was mostly from setState and `clientWidth/Height` not from `findDOMNode` but a profile would quickly confirm that, which I'll do momentarily.

gvn:
Closing for now and will try a different method in the future.

PR #5207, opened 2019-07-29
Bug 1569020 - Add a new badge action for setting homepage_override pref

piatra:
Adds a new action id `"moments-wnp"` that can set the value of `.homepage_override.once` pref.
Values to be set are defined through the `data` field of the `content.action` payload in the message.

Not sure if for messages such as `profile age == ONE_YEAR` where `expire` would not be defined in the message we have to set it when the message is selected.

Do we have other/more scenarios we want to handle?

### DIFF
--- a/lib/PanelTestProvider.jsm
+++ b/lib/PanelTestProvider.jsm
@@ -63,7 +63,7 @@ const MESSAGES = () => [
   },
   {
     id: "WNP_PROFILE_AGE",
-    template: "toolbar_badge",
+    template: "update_action",
     content: {
       action: {
         id: "moments-wnp",
@@ -74,8 +74,6 @@ const MESSAGES = () => [
         },
       },
     },
-    // Never accessed the FxA panel && doesn't use Firefox sync & has FxA enabled
-    targeting: `!hasAccessedFxAPanel && !usesFirefoxSync && isFxAEnabled == true`,
     trigger: { id: "momentsUpdate" },
   },
   {
--- a/lib/PanelTestProvider.jsm
+++ b/lib/PanelTestProvider.jsm
@@ -69,8 +69,8 @@ const MESSAGES = () => [
         id: "moments-wnp",
         data: {
           url:
-            "https://www.mozilla.org/en-US/firefox/80.0a1/whatsnew/?oldversion=50.0a2",
-          expire: Date.now() + TWO_DAYS,
+            "https://www.mozilla.org/%LOCALE%/etc/firefox/retention/thank-you-a/",
+          expireDelta: TWO_DAYS,
         },
       },
     },
--- a/lib/PanelTestProvider.jsm
+++ b/lib/PanelTestProvider.jsm
@@ -68,13 +68,13 @@ const MESSAGES = () => [
         id: "moments-wnp",
         data: {
           url: "foo.com",
-          expire: 0,
+          expire: 100,
         },
       },
     },
     // Never accessed the FxA panel && doesn't use Firefox sync & has FxA enabled
     targeting: `!hasAccessedFxAPanel && !usesFirefoxSync && isFxAEnabled == true`,
-    trigger: { id: "toolbarBadgeUpdate" },
+    trigger: { id: "momentsUpdate" },
   },
   {
     id: `WHATS_NEW_BADGE_${FIREFOX_VERSION}`,
--- a/lib/PanelTestProvider.jsm
+++ b/lib/PanelTestProvider.jsm
@@ -61,6 +61,22 @@ const MESSAGES = () => [
     trigger: { id: "toolbarBadgeUpdate" },
   },
   {
+    id: "WNP_PROFILE_AGE",
+    template: "toolbar_badge",
+    content: {
+      action: {
+        id: "moments-wnp",
+        data: {
+          url: "foo.com",
+          expire: 0,
+        },
+      },
+    },
+    // Never accessed the FxA panel && doesn't use Firefox sync & has FxA enabled
+    targeting: `!hasAccessedFxAPanel && !usesFirefoxSync && isFxAEnabled == true`,
+    trigger: { id: "toolbarBadgeUpdate" },
+  },
+  {
     id: `WHATS_NEW_BADGE_${FIREFOX_VERSION}`,
     template: "toolbar_badge",
     content: {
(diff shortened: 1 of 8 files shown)


Mardak:
My initial guess of how to set `expire` for anniversary, e.g., "show within 1 week of anniversary" would be to have the message have a "expireDelta: 1 week" then when we trigger the message, calculate "expire = now + expireDelta"

Vs a specific date message would just have a "expire: exact time" (Or specific date could also just have a "expireDelta: 1 day" instead ??)

lib/ToolbarBadgeHub.jsm
```diff
@@ -86,12 +103,19 @@ class _ToolbarBadgeHub {
...
+  executeAction({ id, data }) {
     switch (id) {
       case "show-whatsnew-button":
         ToolbarPanelHub.enableToolbarButton();
         ToolbarPanelHub.enableAppmenuButton();
         break;
+      case "moments-wnp":
+        const { url, expire } = data;
+        Services.prefs.setStringPref(
+          HOMEPAGE_OVERRIDE_PREF,
```
Mardak: Not sure if we should only set if pref isn't already set (i.e., a previous pending `once` / `moments-wnp` action)? Could be additional `targeting` condition or something all `moments-wnp` automatically check or … just allow overriding an existing one?

piatra: Good question. Maybe we should always write if the message is selected and control this via targeting. This way we could ship an `undo` message by setting targeting to `true`. What do you think?

piatra:
I like the idea of adding a `expireDelta` where we do `now + ${expireDelta}` when the message is selected. This would cover the fixed date scenario and the profile age scenario where we don't exactly know.

lib/PanelTestProvider.jsm
```diff
@@ -60,6 +60,22 @@ const MESSAGES = () => [
...
     trigger: { id: "toolbarBadgeUpdate" },
   },
+  {
+    id: "WNP_PROFILE_AGE",
+    template: "toolbar_badge",
+    content: {
+      action: {
+        id: "moments-wnp",
+        data: {
+          url: "foo.com",
```
k88hudson: Just FYI it's likely we're going to need to run this through the URL formatter like `%LOCALE%`etc

piatra: Ed is already doing that in his PR https://phabricator.services.mozilla.com/D39286
I could add a check for valid URL and I would update that dummy data with something that matches.

lib/ToolbarBadgeHub.jsm
k88hudson: Overwriting in some cases definitely makes sense, like if you decide to remove the message from remote settings or add a more important one.

Mardak:
Oh… I was just chatting with Jim and realized this is on top of the badge stuff that we've previously said isn't uplifting to 69. Do we want to uplift the what's new panel? Or perhaps uplift the functionality but not trigger the panel? Or find a different way to implement this, so it can be uplifted?

lib/ToolbarBadgeHub.jsm
```diff
@@ -70,12 +83,16 @@ class _ToolbarBadgeHub {
...
     this._addImpression = addImpression;
     this._dispatch = dispatch;
-    this.state = {};
     // Need to wait for ASRouter to initialize before trying to fetch messages
     await waitForInitialized;
     this.messageRequest("toolbarBadgeUpdate");
     // Listen for pref changes that could trigger new badges
     Services.prefs.addObserver(this.prefs.WHATSNEW_TOOLBAR_PANEL, this);
+    const _intervalId = setInterval(
+      () => this.messageRequest("toolbarBadgeUpdate"),
```
k88hudson: You could add some logic here to clear / set the pref

k88hudson: See https://docs.google.com/document/d/1r_Zx75budKb4ZgBKEyIuXSR47vm-OkW0FE50PwS53Q0/edit#heading=h.28r6948w8f9h re:our discussion

lib/ToolbarBadgeHub.jsm
```diff
@@ -64,18 +78,56 @@ class _ToolbarBadgeHub {
...
+    );
+    this.state = { _intervalId };
+  }
+
+  /**
+   * Pref is set via Remote Settings message. We want to continously
+   * monitor new messages that come in to ensure the one with the
+   * highest priority is set.
+   */
+  checkHomepageOverridePref() {
```
k88hudson: Yep, this looks right!

piatra:
I've updated the patch. The test plan I was thinking was having a local message and another message published on stage.
With a fresh profile you trigger the local message, check pref is set.
Switch to staging, wait 5 minutes, check that the pref was replaced and message #1 was unblocked.

lib/PanelTestProvider.jsm
```diff
@@ -60,6 +61,23 @@ const MESSAGES = () => [
...
+        id: "moments-wnp",
+        data: {
+          url:
+            "https://www.mozilla.org/en-US/firefox/80.0a1/whatsnew/?oldversion=50.0a2",
+          expire: Date.now() + TWO_DAYS,
+        },
+      },
+    },
+    // Never accessed the FxA panel && doesn't use Firefox sync & has FxA enabled
+    targeting: `!hasAccessedFxAPanel && !usesFirefoxSync && isFxAEnabled == true`,
```
k88hudson: This looks like it's not relevant to this targeting

lib/PanelTestProvider.jsm
```diff
@@ -60,6 +61,21 @@ const MESSAGES = () => [
...
+  {
+    id: "WNP_PROFILE_AGE",
+    template: "update_action",
+    content: {
+      action: {
+        id: "moments-wnp",
+        data: {
+          url:
+            "https://www.mozilla.org/en-US/firefox/80.0a1/whatsnew/?oldversion=50.0a2",
+          expire: Date.now() + TWO_DAYS,
```
k88hudson: In order for this to work with remote settings, you probably want to make this an expire delta and add the `Date.now()` when the pref is set

lib/PanelTestProvider.jsm
```diff
@@ -60,6 +61,21 @@ const MESSAGES = () => [
...
   },
+  {
+    id: "WNP_PROFILE_AGE",
+    template: "update_action",
+    content: {
+      action: {
+        id: "moments-wnp",
+        data: {
+          url:
+            "https://www.mozilla.org/en-US/firefox/80.0a1/whatsnew/?oldversion=50.0a2",
```
k88hudson: You could go with `https://www.mozilla.org/en-US/etc/%LOCALE%/retention/thank-you-a/` for something a little different looking

PR #5208, opened 2019-07-29
Bug 1569597 - Updated schema & template, added test snippet with corr…

AllegroFox:
…ect content

### DIFF
--- a/content-src/asrouter/templates/SimpleBelowSearchSnippet/SimpleBelowSearchSnippet.schema.json
+++ b/content-src/asrouter/templates/SimpleBelowSearchSnippet/SimpleBelowSearchSnippet.schema.json
@@ -15,6 +15,12 @@
     }
   },
   "properties": {
+    "title": {
+      "allOf": [
+        {"$ref": "#/definitions/plainText"},
+        {"description": "Snippet title displayed before snippet text"}
+      ]
+    },
     "text": {
       "allOf": [
         {"$ref": "#/definitions/richText"},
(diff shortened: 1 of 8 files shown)


k88hudson:
Looking good so far, a few things to note: 
- [ ] The hover area should line up with the rest of the new tab content
- [ ] The bottom of the hover area looks like it needs a bit more padding
- [ ] The button needs a hover state and should be centered relative to the content (check out the `SIMPLE_TEST_BUTTON_URL_1 ` for an example of this

k88hudson:
Much better, a few more things:
* [ ] I think you need to get rid of that `padding-inline-end` on the `innerWrapper` for the button variant
* [ ] Can you check the icon size? I think it looks smaller than what was on the spec

content-src/asrouter/templates/SimpleBelowSearchSnippet/SimpleBelowSearchSnippet.schema.json
```diff
@@ -39,6 +45,34 @@
...
+        {"description": "A url, button_label links to this"}
+      ]
+    },
+    "button_action_args": {
+      "type": "string",
+      "description": "Additional parameters for button action, example which specific menu the button should open"
+    },
+    "button_label": {
+      "allOf": [
+        {"$ref": "#/definitions/plainText"},
```
k88hudson: You need to add `plainText` to the `definitions` object at the top so this will work

k88hudson: You might want to also add dependencies at the bottom:
```
"dependencies": {
    "button_action": ["button_label"],
    "button_url": ["button_label"],
    "button_color": ["button_label"],
    "button_background_color": ["button_label"],
  }
```

PR #5209, opened 2019-07-29
(nobug) - Add mochitest for toolbar badging behavior

piatra:
Some basic mochitest coverage for the `ToolbarBadgeHub`. It badges the fxa button (with and without delay) and checks that clicking the button dismisses the badge.


test/browser/browser_asrouter_bookmarkpanel.js
```diff
@@ -16,11 +16,6 @@ add_task(async function test_fxa_message_shown() {
   const testURL = "data:text/plain,test cfr fxa bookmark panel message";
   const browser = gBrowser.selectedBrowser;
 
-  await SpecialPowers.pushPrefEnv({
```
piatra: This wasn't needed anymore. For a short while we had `devtoolsEnabled` set in our `browser.ini` but we reverted that after some issues.

PR #5217, opened 2019-07-31
WIP Partial first cut at the inital doc; more to come

dmose:
Still a bunch of the main content to write; but this is a start.

### DIFF
--- a/docs/v2-system-addon/vendor.md
+++ b/docs/v2-system-addon/vendor.md
@@ -1,0 +1,30 @@
+# Vendoring in npm libraries
+
+#### Build system basics
+
+#### Vendored libraries as externals
+
+#### Is this library a good candidate to depend on and vendor?
+
+#### How to vendor in a library
+
+* [ ] file a vendoring bug in Bugzilla
+
+* [ ] Get license approval from the licensing team
+  + [ ] review the mozilla-central policy at https://www.mozilla.org/en-US/MPL/license-policy/
+  + [ ] file a license bug as described in the policy doc to request approval
+  + [ ] make it block the next upcoming export bug (if that bug is not already
+
+    filed, file one)
+
+  + [ ] post a patch to add the license to
+
+    [ `toolkit/content/license.html` ](https://searchfox.org/mozilla-central/source/toolkit/content/license.html)
+
+  + [ ] get the patch reviewed
+* [ ] Install the package using npm
+  + [ ] Check that you're using npm 6.9.0 using `npm --version` 
+  + [ ] Be sure to save the exact version by passing `--save-exact` to `npm install` 
+
+* [ ] XXX more to come
+


docs/v2-system-addon/vendor.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### Is this library a good candidate to depend on and vendor?
+
+#### How to vendor in a library
+
+- [ ] file a bug for vendoring in the library itself
+
+- [ ] Get license approval from the licensing team
+
+  - [ ] review the mozilla-central policy at
```
JeaneC: Are these steps for new licenses that weren't previously used or should we follow this step with any new package? 

A note on BSD 3-Clause since its not explicitly mentioned on the license page but rather that "MIT, New BSD, and similar permissive licenses" are allowed.

Another source of truth we can look (suggested by @mhoye) at is https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=87365351 or https://docs.google.com/document/d/1Oguqp43W4_ChyroJ9AJAzG1jSwkUWfKvBKVvrDxVsMg/edit.

There's a specific bullet there mentioning 
```
The following licences are compatible with the MPL, and code under these licenses can be integrated into Mozilla products
* ...
* The New BSD and 2- and 3-clause BSD licenses
```

And I'll still need to do the remaining steps to create a patch for license.html to add this new package usage there.

docs/v2-system-addon/vendor.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### Build system basics
+
+#### Vendored libraries as externals
+
+#### Is this library a good candidate to depend on and vendor?
+
+#### How to vendor in a library
+
+- [ ] file a bug for vendoring in the library itself
```
JeaneC: Can this also be at minimum its own commit in a PR.

docs/v2-system-addon/vendor.md
dmose: The licensing page says

> Remember, always consult the licensing team before importing Third Party Code.

so I think it's always necessary to get their approval, and the other information is just there to tell you what to expect and which licenses are known to be compatible.

According to OSI, "New BSD" == "BSD 3 Clause"; I'll add some clarifying text related to that too.

I believe these steps need to be followed for all libraries.

docs/v2-system-addon/vendor.md
dmose: I think this wants to be its own bug, mostly for documentation reasons.  If someone wants to figure out why something landed, they should be able to easily get details from searching Bugzilla.  That said, I suppose that bug could just point to an existing PR, as long as the thing is in its own commit.

docs/v2-system-addon/vendor.md
JeaneC: Ah makes sense to always consult and follow these steps

docs/v2-system-addon/vendor.md
JeaneC: Do you have an example bug? This is different than the license bug right?

punamdahiya: Filed one https://bugzilla.mozilla.org/show_bug.cgi?id=1570766, @dmose is that good?

docs/v2-system-addon/vendor.md
```diff
@@ -0,0 +1,44 @@
...
+
+- [ ] Get license approval from the licensing team
+
+  - [ ] review the mozilla-central policy at
+        https://www.mozilla.org/en-US/MPL/license-policy/
+  - [ ] file a license bug as described in the policy doc to request approval
```
punamdahiya: License approval bug for BSD 3-Clause, https://bugzilla.mozilla.org/show_bug.cgi?id=1570697

docs/v2-system-addon/vendor.md
dmose: Looks good to me.

PR #5218, opened 2019-08-01
Bug 1568067 - Adding in more to DS devtools

ScottDowne:
To test:
1. Enable DS devtools with this pref: browser.newtabpage.activity-stream.asrouter.devtoolsEnabled set to true
2. Open a new tab
3. Click the little wrench in the top left.
4. In the top right open the Discovery Stream tab.
5. Test that you can reset the pref to default
6. Test that you can clear cache, this includes frequency capping.
7. Test that you can enable and disable all the checkboxes and they do what's expected.


content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -35,6 +35,7 @@ function relativeTime(timestamp) {
 
 const LAYOUT_VARIANTS = {
   basic: "Basic default layout (on by default in nightly)",
+  spocs: "A layout with all spocs shown",
```
ScottDowne: Still waiting on the layout for this.

ScottDowne: So it won't actually work. Also not sure what the endpoint ends up as, and "spocs" is only a guess.

ScottDowne:
There is one blocker on landing this, but I do think a review is probably ok to start now?

ScottDowne:
The layout got updated, probably ready for review now.

I suspect what the endpoint is returning might change, but that's the endpoint we'll end up using for the time being.

punamdahiya:
looks like travis is failing because of prettier

content-src/components/ASRouterAdmin/ASRouterAdmin.scss
```diff
@@ -209,7 +209,7 @@
     }
   }
 
-  .dsEnabled {
+  .dsCheckbox {
```
punamdahiya: is this used?

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -229,23 +270,43 @@ export class DiscoveryStreamAdmin extends React.PureComponent {
...
+        </table>
         <h3>Endpoint variant</h3>
         <p>
           You can also change this manually by changing this pref:{" "}
           <code>browser.newtabpage.activity-stream.discoverystream.config</code>
         </p>
-        <table style={config.enabled ? null : { opacity: 0.5 }}>
+        <table
+          style={
+            config.enabled && !config.hardcoded_layout ? null : { opacity: 0.5 }
```
punamdahiya: in false condition we are just reducing opacity but the radio button stays clickable and clears cache though layout stays same.

punamdahiya: Also noticed, we are replacing $apikey with key value in layout_endpoint when switching between layouts, not sure if it matters as its dev tool

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -33,8 +33,11 @@ function relativeTime(timestamp) {
   return new Date(timestamp).toLocaleString();
 }
 
+
+//https://spocs-staging.web.readitlater.com/v3/newtab/layout?version=1&consumer_key=$apiKey&layout_variant=staging_spocs
 const LAYOUT_VARIANTS = {
   basic: "Basic default layout (on by default in nightly)",
+  staging_spocs: "A layout with all spocs shown",
```
punamdahiya: This layout is hard to test as no spocs are getting returned from spoc_endpoint https://spocs.getpocket.com/spocs?site=1084367, I assume it should show spocs in all positions once we have spoc data

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -229,23 +272,43 @@ export class DiscoveryStreamAdmin extends React.PureComponent {
...
-            onChange={this.onEnableToggle}
-          />{" "}
-          enabled{" "}
-        </div>
+        <button className="button" onClick={this.restorePrefDefaults}>
+          Restore Pref Defaults
+        </button>{" "}
+        <button className="button" onClick={this.expireCache}>
+          Expire Cache
+        </button>
```
punamdahiya: It will be good if get some CTA feedback after click of Restore Defaults and Expire Cache may be a label below or next to Discovery Stream that shows action completed

content-src/components/ASRouterAdmin/ASRouterAdmin.scss
ScottDowne: Not anymore, good catch.

(I was initially using it but renaming it until I decided on a different approach, but this change got left in)

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
ScottDowne: Yeah in general the layout is still in flux, but I didn't want it to block review.

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
ScottDowne: I'm mostly just following the pattern used in about:newtab#devtools-targeting and about:newtab#devtools

Those links have similar buttons.

I agree though, but probably not worth doing now.

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
ScottDowne: I think I'm ok with "in false condition we are just reducing opacity but the radio button stays clickable and clears cache though layout stays same." it's also just reusing the existing way we disabled endpoint if enable/disable is toggled. I think as long as it's clear that changing the layout if it's grey has no effect, that's probably fine. Also technically you CAN change your endpoint if layout is disabled in the pref.

I'm not sure what you mean by "we are replacing $apikey with key value in layout_endpoint when switching between layouts, not sure if it matters as its dev tool" When I switch between layouts the pref is keeping the proper value in the url?

ScottDowne: Oh I see what you mean by "we are replacing $apikey with key value in layout_endpoint when switching between layouts, not sure if it matters as its dev tool"

Hm, I think that's probably fine? It's also not something I'm adding so probably if it's not fine, it's another issue.

PR #5220, opened 2019-08-01
Bug 1570026 - Monitor action for snippets

k88hudson:
This adds a new user action to ASRouter that allows users to directly authenticate with Firefox Monitor via a URL, similar to how the What's New Page does it.

I have a few questions about the best way to implement this, noted below in comments, before I add tests.

### DIFF
--- a/lib/SnippetsTestMessageProvider.jsm
+++ b/lib/SnippetsTestMessageProvider.jsm
@@ -91,6 +91,19 @@ const MESSAGES = () => [
     },
   },
   {
+    id: "SIMPLE_TEST_BUTTON_ACTION_1",
+    template: "simple_snippet",
+    content: {
+      icon: TEST_ICON,
+      icon_dark_theme: TEST_ICON_BW,
+      button_label: "Open about:config",
+      button_action: "OPEN_ABOUT_PAGE",
+      button_action_args: "config",
+      text: "Testing the OPEN_ABOUT_PAGE action",
+      block_button_text: "Block",
+    },
+  },
+  {
     id: "SIMPLE_WITH_TITLE_TEST_1",
     template: "simple_snippet",
     content: {
@@ -399,6 +412,27 @@ const MESSAGES = () => [
       button_url: "https://www.mozilla.org/en-US/firefox/accounts",
       title: "See if you've been part of an online data breach.",
       text: "Firefox Monitor tells you what hackers already know about you.",
+      block_button_text: "Block",
+    },
+  },
+  {
+    id: "SPECIAL_SNIPPET_MONITOR",
+    template: "simple_below_search_snippet",
+    content: {
+      icon: TEST_ICON,
+      title: "See if you've been part of an online data breach.",
+      text: "Firefox Monitor tells you what hackers already know about you.",
+      button_label: "Get monitor",
+      button_action: "ENABLE_FIREFOX_MONITOR",
+      button_action_args: {
+        url:
+          "https://monitor.firefox.com/oauth/init?utm_source=snippets&utm_campaign=monitor-snippet-test&form_type=email&entrypoint=newtab",
+        flowRequestParams: {
+          entrypoint: "snippets",
+          utm_term: "monitor",
+          form_type: "email",
+        },
+      },
       block_button_text: "Block",
     },
   },
--- a/common/Actions.jsm
+++ b/common/Actions.jsm
@@ -159,6 +159,7 @@ for (const type of [
   "OPEN_PREFERENCES_PAGE",
   "SHOW_FIREFOX_ACCOUNTS",
   "PIN_CURRENT_TAB",
+  "ENABLE_FIREFOX_MONITOR",
 ]) {
   ASRouterActions[type] = type;
 }
--- a/content-src/asrouter/asrouter-content.jsx
+++ b/content-src/asrouter/asrouter-content.jsx
@@ -2,7 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { actionCreators as ac } from "common/Actions.jsm";
+import {
+  actionCreators as ac,
+  actionTypes as at,
+  ASRouterActions as ra,
+} from "common/Actions.jsm";
 import { OUTGOING_MESSAGE_NAME as AS_GENERAL_OUTGOING_MESSAGE_NAME } from "content-src/lib/init-store";
 import { generateBundles } from "./rich-text-strings";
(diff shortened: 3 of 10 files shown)


lib/ASRouter.jsm
```diff
@@ -1767,7 +1767,62 @@ class _ASRouter {
...
+          ),
+          csp: null,
+        });
+        break;
+    }
+  }
+
+  async fetchFlowParams({ utm_term, entrypoint, form_type }) {
+    let result = {};
+    const fxaEndpoint = Services.prefs.getStringPref(
```
k88hudson: Do we want to use this Firefox accounts endpoint or rely on the snippets service to provide one?

lib/SnippetsTestMessageProvider.jsm
```diff
@@ -90,6 +90,25 @@ const MESSAGES = () => [
...
+  {
+    id: "SIMPLE_TEST_MONITOR",
+    template: "simple_snippet",
+    content: {
+      icon: TEST_ICON,
+      icon_dark_theme: TEST_ICON_BW,
+      button_label: "Get monitor",
+      button_action: "ENABLE_FIREFOX_MONITOR",
+      button_action_args: {
+        utm_term: "monitor-snippet-test",
```
k88hudson: Not sure what the correct values for any of these should be; should they all just come from the snippets service?

lib/ASRouter.jsm
```diff
@@ -1767,7 +1767,62 @@ class _ASRouter {
...
+          utm_campaign,
+          entrypoint,
+          form_type,
+        } = action.data;
+        const { flowBeginTime, flowId } = await this.fetchFlowParams({
+          entrypoint,
+          utm_term,
+          form_type,
+        });
+        const monitorURL = `https://monitor.firefox.com/oauth/init?utm_source=${utm_source}&utm_campaign=${utm_campaign}&form_type=${form_type}&entrypoint=${entrypoint}&flowBeginTime=${flowBeginTime}&flowId=${flowId}`;
```
k88hudson: I'm curious what you think about passing the whole URL with templates for `flowBeginTime` and `flowId` (which have to come from Firefox) v.s. passing variables:

With templated URL: 
```json
{
  "button_action": "ENABLE_FIREFOX_MONITOR",
  "button_action_args": "https://monitor.firefox.com/oauth/init?utm_source=snippets&utm_campaign=foo&${flowBeginTime}&flowId=${flowId}"
}
```

With variables: 
```json
{
  "button_action": "ENABLE_FIREFOX_MONITOR",
  "button_action_args": {
    "utm_source": "snippets",
    "utm_campaign": "foo"
  }
}
```

I think passing the whole URL could give you more flexibility in changing different parts of the URL remotely, but variables might be less error-prone.

glogiotatidis: I prefer URL for the flexibility. Either way works fine for me. Variables or URL will be handled in the backend and will not be directly editable by Snippet Editors so errors will be limited. Thanks for reaching out! :cake:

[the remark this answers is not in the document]
piatra: Note that the two ways of constructing the string _can_ result in different outputs:
For a `utm_campaign` value of `Firefox Monitor`
* template literals result in `&utm_campaign=Firefox Monitor` which then becomes `&utm_campaign=Firefox%20Monitor`
* URLSearchParams will do `&utm_campaign=Firefox+Monitor`

lib/SnippetsTestMessageProvider.jsm
piatra: They're all strings so it seems ok.

lib/ASRouter.jsm
```diff
@@ -1767,7 +1767,62 @@ class _ASRouter {
...
         });
         break;
+      case ra.ENABLE_FIREFOX_MONITOR:
+        const {
+          utm_term,
+          utm_source,
+          utm_campaign,
+          entrypoint,
+          form_type,
+        } = action.data;
```
piatra: This should be `action.data.args`.

lib/ASRouter.jsm
```diff
@@ -1767,7 +1767,62 @@ class _ASRouter {
...
+    const fxaEndpoint = Services.prefs.getStringPref(
+      "browser.newtabpage.activity-stream.fxaccounts.endpoint",
+      "https://accounts.firefox.com"
+    );
+    try {
+      const url = new URL(
+        `${fxaEndpoint}/metrics-flow?entrypoint=${entrypoint}&form_type=${form_type}&utm_term=${utm_term}`
+      );
+      const response = await fetch(url, { credentials: "omit" });
+      if (response.status === 200) {
```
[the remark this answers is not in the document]
k88hudson: Good question, this is what we're doing for the what's new page

lib/ASRouter.jsm
shane-tomlinson: Is this targeting Firefox 70? Over in https://bugzilla.mozilla.org/show_bug.cgi?id=1568949 and https://bugzilla.mozilla.org/show_bug.cgi?id=1556276, there is talk of having newtab automatically respect the `identity.fxaccounts.remote.root` values, it seems like this should use that same code, which I imagine is bbased off of `identity.fxaccounts.remote.root`?

lib/ASRouter.jsm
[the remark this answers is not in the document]
shane-tomlinson: /metrics-flow isn't cached and will generate a new response every time, 304 isn't possible, checking for a 200 should be sufficient.

lib/ASRouter.jsm
piatra: This will be uplifted to 69.

lib/SnippetsTestMessageProvider.jsm
```diff
@@ -402,6 +415,27 @@ const MESSAGES = () => [
...
+    template: "simple_below_search_snippet",
+    content: {
+      icon: TEST_ICON,
+      title: "See if you've been part of an online data breach.",
+      text: "Firefox Monitor tells you what hackers already know about you.",
+      button_label: "Get monitor",
+      button_action: "ENABLE_FIREFOX_MONITOR",
+      button_action_args: {
+        url:
+          "https://monitor.firefox.com/oauth/init?utm_source=snippets&utm_campaign=monitor-snippet-test&form_type=email&entrypoint=newtab",
```
glogiotatidis: I see both url and flowRequestParams. Can you drop a line about when to use what? @k88hudson

k88hudson: Yep, I will add some docs. Essentially the URL is everything snippets would be able to provide, the `flowParams` is a few properties needed to pass to a Firefox metrics endpoint (defined via a pref) in order to get the right `flowId`, `flowBeginTime` and `deviceId` back.

PR #5227, opened 2019-08-06
Bug 1571441 - Fixed snippet width and block button position on smaller screens

AllegroFox:
Later we should hide the snippets past the smallest width.

### DIFF
--- a/content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
+++ b/content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
@@ -19,6 +19,11 @@
 
         .blockButton {
           display: block;
+
+          // larger inset if discovery stream is enabled.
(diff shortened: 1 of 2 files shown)


k88hudson:
Looks awesome. I added the smallest size as a known issue in the doc. You just need to rebase to get rid of that failure

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -125,8 +154,16 @@
...
-      margin-inline-end: 12px;
       flex-shrink: 0;
+
+      @media (max-width: 865px) {
+        display: none;
+      }
+
+      // Disable breakpoints for now if discovery stream is enabled.
+      .ds-outer-wrapper-breakpoint-override & {
+        display: auto;
```
k88hudson: I don't think this is a valid value for `display`

PR #5228, opened 2019-08-07
Bug 1572059 - Need ASRouterTriggerListeners to trigger on bookmarked pages

piatra:


lib/ASRouter.jsm
```diff
@@ -129,6 +129,17 @@ async function chooseBranch(seed, branches) {
   return branches[await Sampling.ratioSample(seed, ratios)][0];
 }
 
+// istanbul ignore next
+function additionalParamsForTrigger(id) {
+  if (id === "openURL") {
+    return ASRouterTargeting.Environment.recentBookmarks.then(bookmarks =>
```
k88hudson: It seems kind of wasteful to have to do this on every page load if the bookmarks-related message isn't in play.

I wonder if instead of modifying all usages of`openURL` for for this use case, what do you think about a new trigger (`openBookmarkedURL`?) that skips the URL check and instead compares against `recentBookmarks` whenever a page is loaded?.

lib/ASRouterTargeting.jsm
```diff
@@ -353,6 +354,9 @@ const TargetingGetters = {
       }))
     );
   },
+  get recentBookmarks() {
```
k88hudson: This is useful :+1:

piatra:
I'm merging this. TC is failing bc of a missing backport. Doing that too.

PR #5230, opened 2019-08-07
Bug 1560260 - fixing cards showing with nothing after system_tick

ScottDowne:


lib/DiscoveryStreamFeed.jsm
```diff
@@ -851,9 +851,10 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
    */
   async refreshAll(options = {}) {
     const { updateOpenTabs, isStartup } = options;
+    const dispatchQueue = [];
     const dispatch = updateOpenTabs
       ? action => this.store.dispatch(ac.BroadcastToContent(action))
-      : this.store.dispatch;
+      : e => dispatchQueue.push(e);
```
Mardak: Let's keep the naming consistent, so `e` -> `action` and could probably just move it to a single function, e.g.,
```js
const dispatch = action => updateOpenTabs ? store…(action) : queue…(action);
```

lib/DiscoveryStreamFeed.jsm
```diff
@@ -851,9 +851,10 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
    */
   async refreshAll(options = {}) {
     const { updateOpenTabs, isStartup } = options;
+    const dispatchQueue = [];
```
Mardak: Add a comment why we have a queue (probably something along the lines of simulating a transaction)

lib/DiscoveryStreamFeed.jsm
```diff
@@ -865,6 +866,9 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
         Cu.reportError(`Error trying to load component feeds: ${error}`)
       ),
     ]);
+    if (!updateOpenTabs) {
+      dispatchQueue.forEach(e => this.store.dispatch(e));
```
Mardak: No need for the `if` as the queue is empty otherwise. Also consistent naming of `e` -> `action`

ScottDowne: Ah I see, yes, that makes sense.

lib/DiscoveryStreamFeed.jsm

lib/DiscoveryStreamFeed.jsm

Mardak:
The failing test is checking if dispatch returns the action, but it should probably check if this.store.dispatch was called with the appropriate non-broadcast action values.

lib/DiscoveryStreamFeed.jsm
ScottDowne: You have a huge opinion if I keep it as two functions?

```
    const dispatch = updateOpenTabs
      ? action => this.store.dispatch(ac.BroadcastToContent(action))
      : action => dispatchQueue.push(action);
``

Mardak: Either is fine. One saves an `action =>` 9 characters 😛 whatever you think is easier to read

Mardak:
Do we want a different approach with a narrow fix to uplift to 69 beta?

ScottDowne:
Closing for now in favour of https://github.com/mozilla/activity-stream/pull/5246

PR #5231, opened 2019-08-07
Bug 1571023 - Added new strings for Sync, Send Tab and Firefox Send.

AllegroFox:


locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
   .tooltiptext = What’s New
 
 cfr-whatsnew-panel-header = What’s New
+
+## Bookmark Sync
+
+cfr-sync-bookmarks-type = Recommended Feature
```
flodolo: I think this should be one string used for all this type of recommendations, and not repeated for each one of them?

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
...
+cfr-sync-bookmarks-header = Get this bookmark on your phone
+cfr-sync-bookmarks-body = Take your bookmarks, passwords, history and more everywhere you're signed into Firefox.
+cfr-sync-bookmarks-cta-button = Turn on Sync
+
+## Send Tab
+
+cfr-send-tab-type = Recommended Feature
+cfr-send-tab-header-1 = Read this on the go
```
flodolo: I don't see this mentioned in the bug?
https://bugzilla.mozilla.org/show_bug.cgi?id=1569696

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
...
+cfr-send-tab-header-2 = Take this recipe to the kitchen
+cfr-send-tab-body = Send Tab lets you easily share this link to your phone or anywhere you're signed in to Firefox.
+cfr-send-tab-cta-button = Try Send Tab
+
+## Firefox Send
+
+cfr-firefox-send-type = Recommend Feature
```
flodolo: See comment above, but in case it should be Recommend**ed**

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
   .tooltiptext = What’s New
 
 cfr-whatsnew-panel-header = What’s New
+
+## Bookmark Sync
+
+cfr-sync-bookmarks-type = Recommended Feature
+cfr-sync-bookmarks-header = Get this bookmark on your phone
+cfr-sync-bookmarks-body = Take your bookmarks, passwords, history and more everywhere you're signed into Firefox.
```
flodolo: Use proper apostrophes (’), you should use the term for Firefox (`-brand-product-name` in this case)

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
...
+cfr-sync-bookmarks-cta-button = Turn on Sync
```
flodolo: Use the term for Sync

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
...
+
+## Send Tab
+
+cfr-send-tab-type = Recommended Feature
+cfr-send-tab-header-1 = Read this on the go
+cfr-send-tab-header-2 = Take this recipe to the kitchen
+cfr-send-tab-body = Send Tab lets you easily share this link to your phone or anywhere you're signed in to Firefox.
```
flodolo: Apostrophe, term for Firefox

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
...
+cfr-send-tab-cta-button = Try Send Tab
+
+## Firefox Send
+
+cfr-firefox-send-type = Recommend Feature
+cfr-firefox-send-header = Share this PDF securely
+cfr-firefox-send-body = Keep your sensitive documents safe from prying eyes with end-to-end encryption and a link that disappears when you're done.
+cfr-firefox-send-cta-button = Try Firefox Send
```
flodolo: Term for Firefox Send
https://hg.mozilla.org/l10n/gecko-strings/file/default/browser/browser/branding/brandings.ftl

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,25 @@ cfr-whatsnew-button =
...
+cfr-send-tab-header-1 = Read this on the go
+cfr-send-tab-header-2 = Take this recipe to the kitchen
+cfr-send-tab-body = Send Tab lets you easily share this link to your phone or anywhere you're signed in to Firefox.
+cfr-send-tab-cta-button = Try Send Tab
+
+## Firefox Send
+
+cfr-firefox-send-type = Recommend Feature
+cfr-firefox-send-header = Share this PDF securely
+cfr-firefox-send-body = Keep your sensitive documents safe from prying eyes with end-to-end encryption and a link that disappears when you're done.
```
flodolo: Apostrophe

flodolo:
Looking at the other FTL files, I wonder if we have access to brandings.ftl, or need to repeat brandings here? That would be really unfortunate, and I'd like to avoid it.

flodolo:
> Looking at the other FTL files, I wonder if we have access to brandings.ftl, or need to repeat brandings here? That would be really unfortunate, and I'd like to avoid it.

Sorry, rumbling here. I see references in First Run and Activity Stream, I hope it can be added to ASRouter too (if needed). @mardak will know that for sure

locales-src/asrouter.ftl
AllegroFox: Done! I've created a cfr-feature-type-recommended variable just above the section.

flodolo: Please make sure to add a comment to explain in which content the string is used.

locales-src/asrouter.ftl
AllegroFox: This one is from https://bugzilla.mozilla.org/show_bug.cgi?id=1569692; the other strings overlap (send tab) so I put them together.

flodolo: Thanks. That bug wasn't mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1571023, had no idea there was a fifth one

locales-src/asrouter.ftl
AllegroFox: I've fixed all the apostrophes, I'll ping @Mardak and ask if we can use the variables from brandings.ftl :)

AllegroFox: I do see the branding strings in our other .ftl files, so it might be fine!

Mardak: `cfr-doorhanger-bookmark-fxa-body` already uses `{ -fxaccount-brand-name }` but that's specially adding the ftl for the bookmark panel:
https://github.com/mozilla/activity-stream/blob/70b6323072f2c801635e453e2e00f2bdb6e1830a/lib/BookmarkPanelHub.jsm#L111

Specifically for CFR page action, I believe additional ftl should be added here:
https://github.com/mozilla/activity-stream/blob/70b6323072f2c801635e453e2e00f2bdb6e1830a/lib/CFRPageActions.jsm#L71

AllegroFox:
Hmm, I don't know if I included the .ftl files correctly - taskcluster is failing to resolve the promise there. Any ideas?

lib/CFRPageActions.jsm
```diff
@@ -68,7 +68,12 @@ class PageAction {
     this._showPopupOnClick = this._showPopupOnClick.bind(this);
     this.dispatchUserAction = this.dispatchUserAction.bind(this);
 
-    this._l10n = new DOMLocalization(["browser/newtab/asrouter.ftl"]);
+    this._l10n = new DOMLocalization([
+      "browser/newtab/asrouter.ftl",
+      "browser/branding/brandings.ftl",
+      "browser/branding/sync-brand.ftl",
+      "branding/official/brand.ftl",
```
Mardak: You should be able to test this by triggering a cfr message from newtab devtools. Pretty sure this will fail because the packaged brand.ftl has no `official/`

AllegroFox: oh hmm - that was the only place I saw `-brand-product-name` defined; I can see what happens without it!

Mardak: You'll need to include brand.ftl. Just don't have the `official/` in the path. The source directory is to track different versions of the strings but the packaged-into-firefox version only picks one. See other uses:

https://searchfox.org/mozilla-central/search?q=%2Fbrand.ftl&path=.html

AllegroFox: Unfortunately triggering CFRs from DevTools is a little hit and miss...

Mardak: hit and miss how?

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,26 @@ cfr-whatsnew-button =
...
+
+## Recommended features
+
+cfr-feature-type-recommended = Recommended Feature
+
+## Bookmark Sync
+
+cfr-sync-bookmarks-header = Get this bookmark on your phone
+cfr-sync-bookmarks-body = Take your bookmarks, passwords, history and more everywhere you’re signed into { -brand-product-name }.
+cfr-sync-bookmarks-cta-button = Turn on { -sync-brand-short-name }
```
Mardak: Looks like CFR buttons need accesskey attributes

AllegroFox: should the reference just be branding/brand.ftl? Or something else? It shows up with /official in the gecko-strings: 
https://hg.mozilla.org/l10n/gecko-strings/file/default/browser/branding/official/brand.ftl

AllegroFox: I'll add .accesskey :)

Mardak: https://github.com/mozilla/activity-stream/pull/5231#discussion_r311748899

AllegroFox: Shoot sorry, I didn't refresh and didn't have that comment 👍

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,29 @@ cfr-whatsnew-button =
...
+  .accesskey = T
+
+## Send Tab
+
+cfr-send-tab-header-1 = Read this on the go
+cfr-send-tab-header-2 = Take this recipe to the kitchen
```
Mardak: A more descriptive id might be useful to clarify they're different headers not part 1 and part 2 of a related header. Or maybe just `cfr-send-tab-header` vs `cfr-send-tab-recipe-header` ?

AllegroFox: Second option sounds much clearer, I'll do that! Thanks for all the help on this one :)

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,29 @@ cfr-whatsnew-button =
   .tooltiptext = What’s New
 
 cfr-whatsnew-panel-header = What’s New
+
+## Recommended features
+
+cfr-feature-type-recommended = Recommended Feature
```
Mardak: I believe flod wanted a comment describing how this is used: https://github.com/mozilla/activity-stream/pull/5231#discussion_r311714464

These existing strings seem to be used the same way as headers:
https://github.com/mozilla/activity-stream/blob/3b224ad27ca1846d701d500d988f7ee8bfc151b9/locales-src/asrouter.ftl#L5-L6

So the string id probably should be more like the existing ones instead of some "feature type"

locales-src/asrouter.ftl
```diff
@@ -81,3 +81,29 @@ cfr-whatsnew-button =
...
+
+## Bookmark Sync
+
+cfr-sync-bookmarks-header = Get this bookmark on your phone
+cfr-sync-bookmarks-body = Take your bookmarks, passwords, history and more everywhere you’re signed into { -brand-product-name }.
+cfr-sync-bookmarks-cta-button = Turn on { -sync-brand-short-name }
```
Mardak: The existing ids are `cfr-doorhanger-extension-ok-button` and `cfr-doorhanger-pintab-ok-button`, so these all should probably start with `cfr-doorhanger-` and the buttons should be `ok` instead of `cta`

locales-src/asrouter.ftl
flodolo: Given we already have these strings, it would be a good idea to move this new one there (and change the ID), and add the comment as a section comment.

```
## These messages are used as headings in the recommendation doorhanger

cfr-doorhanger-extension-heading = Recommended Extension
cfr-doorhanger-feature-heading = Recommended Feature
cfr-doorhanger-pintab-heading = Try This: Pin Tab
##
```

The `##` after the strings resets the section comment.

AllegroFox: will do! :)

PR #5232, opened 2019-08-07
Bug 1565043 - New TextPromo Component

gvn:
Testing:

- Change `browser.newtabpage.activity-stream.discoverystream.config` to `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"https://gist.githubusercontent.com/gvn/0d988c266468ab17c3eeeeab1bed9bef/raw/7976030636ef193878a97f1359ca8c21ba355d4c/text-promo.json"} `
- Change `browser.newtabpage.activity-stream.discoverystream.endpoints` to `https://getpocket.cdn.mozilla.net/,https://spocs.getpocket.com/,file:///,https://gist.githubusercontent.com`
- Confirm layout looks like this:

Screen Shot 2019-08-13 at 1 31 57 PM


punamdahiya:
There are differences in the layout proposed by back-end
https://docs.google.com/document/d/1vLweiJXqMQ6iW4IFBmOzPkot6VaadIHUXKNUlOP5hVw/edit#
and in JSON used in patch that should be clarified
https://gist.githubusercontent.com/gvn/0d988c266468ab17c3eeeeab1bed9bef/raw/86d46d46229c7520deedf02745785117457466a6/text-promo.json

content-src/styles/_activity-stream.scss
```diff
@@ -162,6 +162,8 @@ input {
 @import '../components/DiscoveryStreamComponents/DSMessage/DSMessage';
 @import '../components/DiscoveryStreamImpressionStats/ImpressionStats';
 @import '../components/DiscoveryStreamComponents/DSEmptyState/DSEmptyState';
+@import '../components/DiscoveryStreamComponents/DSTextPromo/DSTextPromo';
+@import '../components/DiscoveryStreamComponents/DSChevronLink/DSChevronLink';
```
punamdahiya: DSChevronLink has scss only  and no corresponding jsx. We should either move these in _DSTextPromo.scss or move in common styles under content-src/styles

data/content/assets/glyph-caret-right.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="5" height="24" viewBox="0 0 5 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M0.247788 8.20425C0.560345 7.91574 1.04761 7.93523 1.33612 8.24779L4.7997 12L1.33612 15.7522C1.04761 16.0648 0.560345 16.0843 0.247788 15.7957C-0.0647694 15.5072 -0.08426 15.02 0.204254 14.7074L2.7034 12L0.204254 9.29259C-0.08426 8.98003 -0.0647694 8.49277 0.247788 8.20425Z" fill="#0C0C0D"/>
+</svg>
```
punamdahiya: Running it by node module svgo will give an optimized version of this svg.

gvn:
> There are differences in the layout proposed by back-end
https://docs.google.com/document/d/1vLweiJXqMQ6iW4IFBmOzPkot6VaadIHUXKNUlOP5hVw/edit#
and in JSON used in patch that should be clarified
https://gist.githubusercontent.com/gvn/0d988c266468ab17c3eeeeab1bed9bef/raw/86d46d46229c7520deedf02745785117457466a6/text-promo.json

@punamdahiya Are you saying I should document this new component schema in that doc? I was hoping that after this is approved I'd work with BE on adding directly to the API docs along with the other components.

content-src/styles/_activity-stream.scss
gvn: In terms of organization I think it makes more sense to put all the components in a parallel structure even if they don't require JSX siblings. DSChevronLink is broken out because it can be used in other places than just DSTextPromo.

`content-src/styles` has no organizational concept of components that I can tell (we could start one, but I'm not sure what the benefit is of having component SCSS in 2 different places?).

data/content/assets/glyph-caret-right.svg
gvn: Done!

gvn:
@punamdahiya I used the same variable names available from the existing spoc schema where I could. Also I consolidated the CSS as we discussed...

content-src/components/DiscoveryStreamComponents/DSMessage/_DSMessage.scss
```diff
@@ -43,3 +43,53 @@
     }
   }
 }
+
+.ds-chevron-link {
```
punamdahiya: Any reason css is combined with DSMessage and not as _DSTextPromo.scss? I believe this assumes DSMessage loaded when showing TextPromo

content-src/components/DiscoveryStreamComponents/DSTextPromo/DSTextPromo.jsx
```diff
@@ -0,0 +1,25 @@
...
+import React from "react";
+import { SafeAnchor } from "../SafeAnchor/SafeAnchor";
+
+export class DSTextPromo extends React.PureComponent {
+  render() {
+    return (
+      <div className="ds-text-promo">
+        <img src={this.props.image} alt={this.props.alt_text} />
+        <div className="text">
+          <header>
```
punamdahiya: nit: as per accessibility we should using headings in favor of header which should be only one in a section https://www.w3.org/WAI/tutorials/page-structure/regions/

content-src/components/DiscoveryStreamComponents/DSTextPromo/_DSTextPromo.scss
```diff
@@ -0,0 +1,86 @@
...
+    border-radius: 2px;
+  }
+
+  &::after {
+    @include dark-theme-only {
+      background-color: $blue-40;
+    }
+
+    content: ' ';
+    mask: url('resource://activity-stream/data/content/assets/glyph-caret-right.svg') 0 -8px no-repeat;
```
punamdahiya: nit: should consider global variable [$image-path] here, help keep assets reference at one place (https://github.com/mozilla/activity-stream/blob/master/content-src/styles/_variables.scss#L153)

PR #5233, opened 2019-08-07
Bug 1571817 - Hold back rollout test for Extended Triplets

ncloudioj:


lib/ASRouter.jsm
```diff
@@ -988,6 +995,40 @@ class _ASRouter {
...
+        TRAILHEAD_CONFIG.EXTENDED_TRIPLETS_EXPERIMENT_PREF,
+        branch
+      );
+    }
+
+    // In order for ping centre to pick this up, it MUST contain a substring activity-stream
+    const experimentName = `activity-stream-extended-triplets`;
+    TelemetryEnvironment.setExperimentActive(experimentName, branch);
+
+    const state = { extendedTripletsInitialized: true };
```
k88hudson: why not just this? 
```js
await this.setState({
  // Disable the extended triplets for the "test" group.
  extendedTripletsInitialized: branch !== "test"
});
```

lib/ASRouter.jsm
```diff
@@ -988,6 +995,40 @@ class _ASRouter {
...
+    // Disable the extended triplets for the "test" group.
+    if (branch === "test") {
```
k88hudson: I'm not sure if it matters but maybe naming this something slightly more descriptive like `"holdback"` or something might reduce confusion for testing etc.

lib/ASRouter.jsm
```diff
@@ -1823,11 +1864,22 @@ class _ASRouter {
...
+      });
+
+      // Set up the experiment for extended triplets. It's done here because we
+      // only want to enroll users (both for control and test) if they meet the
+      // targeting criteria.
+      if (message) {
+        await this.setupExtendedTriplets();
+      }
+
+      if (!(message && this.state.showExtendedTriplets)) {
```
k88hudson: even though it's equivalent personally I think `(!message || !this.state.showExtendedTriplets)` is more clear, or maybe put a comment as to what this is checking

k88hudson: Also I'd like to see a test for this if you don't mind in the `"#onMessage: NEWTAB_MESSAGE_REQUEST"` block

ncloudioj: Good point. More tests were added.

PR #5236, opened 2019-08-08
Bug 1570754 - Add delay to clearing triplets and always block extended

k88hudson:


content-src/asrouter/asrouter-content.jsx
```diff
@@ -377,7 +377,7 @@ export class ASRouterUISurface extends React.PureComponent {
             sendUserActionTelemetry={this.sendUserActionTelemetry}
             executeAction={ASRouterUtils.executeAction}
             dispatch={this.props.dispatch}
-            onBlock={this.onBlockById(this.state.message.id)}
+            onBlockById={ASRouterUtils.blockById}
```
Mardak: Hmm.... I do see that ASRouterUtils typically uses `ASRouterUtils` instead of `this` to avoid binding. Although there is a `theme: this.getPreviewTheme()` but this call is okay.

content-src/asrouter/templates/FirstRun/FirstRun.jsx
```diff
@@ -7,6 +7,9 @@ import { Interrupt } from "./Interrupt";
 import { Triplets } from "./Triplets";
 import { BASE_PARAMS } from "./addUtmParams";
 
+// Note: should match the transition time on .trailheadCards in _Trailhead.scss
+const TRANSITION_LENGTH = 500;
```
Mardak: Probably good to add a matching comment in the scss:
https://github.com/mozilla/activity-stream/blob/469c7ee75d8fb4ddd64784ef058d79284584acf7/content-src/asrouter/templates/Trailhead/_Trailhead.scss#L268

content-src/asrouter/templates/Trailhead/_Trailhead.scss
```diff
@@ -267,6 +267,10 @@
   text-align: center;
   transition: max-height 0.5s $photon-easing;
 
+  // This is needed for the transition to work, but will cut off content at the smallest breakpoint
+  @media (min-width: $break-point-medium) {
+    max-height: 1000px;
```
Mardak: Ah. @AllegroFox removed this in #5158, so I guess no transition for narrow but this is quite the corner case? We could just increase the max-height to say 1100px with no media query ?

content-src/asrouter/templates/FirstRun/FirstRun.jsx
```diff
@@ -140,10 +143,11 @@ export class FirstRun extends React.PureComponent {
...
     this.setState({ isTripletsContainerVisible: false });
-    // TODO: Needs to block ALL extended triplets as well
-    if (this.props.message.template === "extended_triplets") {
-      this.props.onBlock();
-    }
+
+    // Closing triplets should prevent any future extended triplets from showing up
+    setTimeout(() => {
+      this.props.onBlockById("EXTENDED_TRIPLETS_1");
+    }, TRANSITION_LENGTH);
```
Mardak: Did you try adding a "transitionend" event listener? I suppose it might be trickier to test ?

content-src/asrouter/templates/Trailhead/_Trailhead.scss
k88hudson: That's right. I think the transition would probably look really weird at the smallest screen anyway so I think the media query makes sense

PR #5239, opened 2019-08-09
Bug 1572441 - Localize Protection panel message

Mardak:
r? @k88hudson or @piatra Just adds asrouter.ftl if necessary and convert raw strings to `string_id`s

Before (hardcoded "Firefox"):
Screen Shot 2019-08-08 at 10 55 02 PM

After:
Screen Shot 2019-08-08 at 10 54 07 PM


Mardak:
I merged central to master with latest ports and added a test-only fix to meet 100% code coverage. So this PR is just the 1 localization commit now.

PR #5240, opened 2019-08-09
Bug 1571442 - Search box shouldn't move down with extended Triplets on

k88hudson:
To test this:
- Turn discovery stream off
- Collapse all your sections
- Open devtools, trigger EXTENDED_TRIPLETS_1  (or just launch with `--new-profile`)
- Make your window vertically large
- Ensure you don't see a large gap between the triplets and the search


k88hudson:
@Mardak it is definitely weird that discovery stream doesn't center but AS does, i'd prefer just to keep the existing functionality for now and bring it up with Pocket?

PR #5241, opened 2019-08-09
Bug 1572556 - Mitigation strategy for 69 DS and Adzerk

ScottDowne:
Adding steps to test to bugzilla

### DIFF
--- a/test/unit/lib/DiscoveryStreamFeed.test.js
+++ b/test/unit/lib/DiscoveryStreamFeed.test.js
@@ -294,6 +294,29 @@ describe("DiscoveryStreamFeed", () => {
         "https://spocs.getpocket.com/spocs"
       );
     });
+    it("should use local basic layout with hardcoded_layout and hardcoded_basic_layout being true", async () => {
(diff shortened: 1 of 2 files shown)


test/unit/lib/DiscoveryStreamFeed.test.js
```diff
@@ -294,6 +294,29 @@ describe("DiscoveryStreamFeed", () => {
...
       );
     });
+    it("should use local basic layout with hardcoded_layout and hardcoded_basic_layout being true", async () => {
+      feed.config.hardcoded_layout = true;
+      feed.config.hardcoded_basic_layout = true;
+      sandbox.stub(feed, "fetchLayout").returns(Promise.resolve(""));
+
+      await feed.loadLayout(feed.store.dispatch);
+
+      assert.notCalled(feed.fetchLayout);
```
punamdahiya: This test is similar to one above it, we should add a check to ensure its basic layout may be by checking layout[0].components[2].properties.items is 3

PR #5242, opened 2019-08-09
Bug 1572841 - Avoid loading JSMs eagerly in Discovery Stream

ScottDowne:
To test, this is mostly a tiny refactor, so regression testing is probably the best bet, things should work as they always did.


ScottDowne:
These are all I could find.

PR #5243, opened 2019-08-12
Bug 1571450 - Added a focus highlight to Snippet CTA and dismiss buttons

AllegroFox:
Let me know if a different highlight color would be better :)

### DIFF
--- a/content-src/asrouter/components/Button/_Button.scss
+++ b/content-src/asrouter/components/Button/_Button.scss
@@ -40,7 +40,7 @@
     }
 
     &:focus {
-      box-shadow: 0 0 0 3px $blue-40;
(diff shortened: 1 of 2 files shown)


content-src/asrouter/components/Button/_Button.scss
```diff
@@ -38,6 +38,10 @@
     &:active {
       background-color: $grey-90-30;
     }
+
```
k88hudson: We tend to just use the hover state for stuff on new tab rather than a blue box shadow, but can you ask @aaronrbenson what he thinks?

AllegroFox: sure :)

AllegroFox: So the focus state is okay, I just made it match [these](https://design.firefox.com/photon/components/buttons.html)l; for the dismiss button I modified it so it can be focused by the keyboard without hovering the div (it just un-hides, basically).

PR #5247, opened 2019-08-12
Bug 1573204 - Adding an extra line to pocket stories

ScottDowne:
To test:

1. Open a new tab
2. Ensure we see pocket stories with at max 4 lines of text.
3. Ensure cards in a single row are all the same height, based on the height of the tallest card.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
+++ b/content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
@@ -68,7 +68,7 @@ $col4-header-font-size: 14;
       }
 
       .excerpt {
-        @include limit-visibile-lines(3, 24, 15);
+        @include limit-visibile-lines(4, 24, 15);
       }
     }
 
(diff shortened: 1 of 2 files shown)


content-src/components/DiscoveryStreamComponents/CardGrid/_CardGrid.scss
```diff
@@ -68,7 +68,7 @@ $col4-header-font-size: 14;
       }
 
       .excerpt {
-        @include limit-visibile-lines(3, 24, 15);
+        @include limit-visibile-lines(4, 24, 15);
```
punamdahiya: This is just going to change excerpt length to 4 only for 9, 10, 11, 12 column card grid layout. Is that the ask?

To change it everywhere for DSCard including card list in Hero, we should be [changing here](https://github.com/mozilla/activity-stream/blob/master/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss#L107)

ScottDowne: Ah I see.

PR #5248, opened 2019-08-12
Bug 1573209 - Installing addon from ReturntoAMO shouldnt hide triplets

k88hudson:
I don't love this solution but I couldn't really think of anything else that works better at the moment. What do you think?

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -1644,10 +1644,8 @@ class _ASRouter {
         "webextension-install-notify"
       );
       this.messageChannel.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {
-        type: "CLEAR_MESSAGE",
-        data: { id: "RETURN_TO_AMO_1" },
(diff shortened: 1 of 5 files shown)


lib/ASRouter.jsm
```diff
@@ -1644,10 +1644,8 @@ class _ASRouter {
         "webextension-install-notify"
       );
       this.messageChannel.sendAsyncMessage(OUTGOING_MESSAGE_NAME, {
-        type: "CLEAR_MESSAGE",
-        data: { id: "RETURN_TO_AMO_1" },
+        type: "CLEAR_INTERRUPT",
       });
-      this.blockMessageById("RETURN_TO_AMO_1");
```
Mardak: Removing the block changes the behavior outside of the current page. Is that desired? I guess a subsequent opening of about:welcome is not common, but does this has other effects?

Is there a way to SET_MESSAGE something that would render the desired triplets to replace the CLEAR_MESSAGE ? This would avoid exposing interrupt-specific messages

PR #5249, opened 2019-08-12
Bug 1573206 - Testing a render change

ScottDowne:
To test:

1. Go to newtab.
2. Inspect a card, and look for these: `<div class="ds-card">`
3. Scroll down a bit.

Expected: Initially, any cards above the fold should be `<div class="ds-card">` and any below the fold should be  `<div class="ds-card placeholder">` and be empty (so they render fast) . As you scroll, they should update with their content.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
@@ -13,14 +13,7 @@ $excerpt-line-height: 20;
     background: transparent;
     box-shadow: inset $inner-box-shadow;
     border-radius: 4px;
-
-    .ds-card-link {
-      cursor: default;
-    }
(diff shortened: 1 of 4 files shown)


content-src/components/DiscoveryStreamComponents/DSCard/_DSCard.scss
```diff
@@ -13,14 +13,7 @@ $excerpt-line-height: 20;
...
     border-radius: 4px;
-
-    .ds-card-link {
-      cursor: default;
-    }
-
-    .img-wrapper {
-      opacity: 0;
-    }
+    min-height: 300px;
```
punamdahiya: setting min-height breaks placeholder cards shown in Hero component. Also seeing a jank due to 300px placeholder card when a card is dimissed

PR #5253, opened 2019-08-14
WIP thinking about FF prefs to control DS

ScottDowne:


ScottDowne:
@Mardak this is basically what I'm thinking about for enabling rollouts and other further risk mitigation strategies. I think the main thing this is missing is listening to pref changes.

I made the decision to use a ff pref for ds enabled, that's defaulted to false unless we turn it on. I would imagine we would need to turn it on in nightly/beta in a FF patch.

ScottDowne:
Closing for https://github.com/mozilla/activity-stream/pull/5254

PR #5254, opened 2019-08-14
Bug 1573930 - FF prefs to control DS

ScottDowne:
To test:

First pref:
1. Set `browser.newtabpage.activity-stream.discoverystream.enabled` to `true`
2. Restart, and ensure new tab shows 7 rows of story cards.
3. Set it to false, restart, and ensure we now see the old AS.

Next pref:
1. Set `browser.newtabpage.activity-stream.discoverystream.hardcoded-basic-layout` to true.
2. Set `browser.newtabpage.activity-stream.discoverystream.enabled` to `true`
3. Restart, ensure you see DS enabled with 3 cards.
4. Set it back to false, and restart, ensure you see the old AS with 3 cards.
5. Attached screen shots for each version of 3 cards to help verify. (DS 3 card experience and AS 3 card experience). Notice the slight difference in the popular topics section under the cards.

DS 3 card experience:
DS-3-cards

AS 3 card experience:
AS-3-cards

Third pref:
1. Ensure you can see sponsored content.
2. Set `browser.newtabpage.activity-stream.discoverystream.enabled` to true.
3. Set `browser.newtabpage.activity-stream.discoverystream.spocs-endpoint` to "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs"
4. Restart and watch the browser toolbox network tab for "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs" and ensure you see sponsored content.
5. Set it to "" or clear it. Restart.
6. On the next load, watch browser toolbox network tab for "https://spocs.getpocket.com/spocs" and ensure you see sponsored content.

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -55,6 +55,9 @@ const FETCH_TIMEOUT = 45 * 1000;
 const PREF_CONFIG = "discoverystream.config";
 const PREF_ENDPOINTS = "discoverystream.endpoints";
 const PREF_IMPRESSION_ID = "browser.newtabpage.activity-stream.impressionId";
+const PREF_ENABLED = "discoverystream.enabled";
+const PREF_HARDCODED_BASIC_LAYOUT = "discoverystream.hardcoded-basic-layout";
+const PREF_SPOCS_ENDPOINT = "discoverystream.spocs-endpoint";
 const PREF_TOPSTORIES = "feeds.section.topstories";
 const PREF_SPOCS_CLEAR_ENDPOINT = "discoverystream.endpointSpocsClear";
 const PREF_SHOW_SPONSORED = "showSponsored";
@@ -148,6 +151,10 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
         `Could not parse preference. Try resetting ${PREF_CONFIG} in about:config. ${e}`
       );
     }
+    this._prefCache.config.enabled =
+      this._prefCache.config.enabled &&
+      this.store.getState().Prefs.values[PREF_ENABLED];
+
     return this._prefCache.config;
   }
 
@@ -331,15 +338,24 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
 
     if (!layout || !layout.layout) {
-      if (this.config.hardcoded_basic_layout) {
+      if (
+        this.config.hardcoded_basic_layout ||
+        this.store.getState().Prefs.values[PREF_HARDCODED_BASIC_LAYOUT]
+      ) {
         layout = { lastUpdate: Date.now(), ...basicLayoutResp };
       } else {
         layout = { lastUpdate: Date.now(), ...defaultLayoutResp };
       }
     }
 
-    if (layout.spocs && this.config.spocs_endpoint) {
-      layout.spocs.url = this.config.spocs_endpoint;
+    if (
+      layout.spocs &&
+      (this.store.getState().Prefs.values[PREF_SPOCS_ENDPOINT] ||
+        this.config.spocs_endpoint)
+    ) {
+      layout.spocs.url =
+        this.store.getState().Prefs.values[PREF_SPOCS_ENDPOINT] ||
+        this.config.spocs_endpoint;
     }
 
     sendUpdate({
@@ -1247,6 +1263,9 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       case at.PREF_CHANGED:
         switch (action.data.name) {
           case PREF_CONFIG:
+          case PREF_ENABLED:
+          case PREF_HARDCODED_BASIC_LAYOUT:
+          case PREF_SPOCS_ENDPOINT:
             // Clear the cached config and broadcast the newly computed value
             this._prefCache.config = null;
             this.store.dispatch(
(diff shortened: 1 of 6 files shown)


lib/DiscoveryStreamFeed.jsm
```diff
@@ -331,15 +340,24 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
 
     if (!layout || !layout.layout) {
-      if (this.config.hardcoded_basic_layout) {
+      if (
+        this.config.hardcoded_basic_layout ||
+        Services.prefs.getBoolPref(PREF_HARDCODED_BASIC_LAYOUT, false)
```
Mardak: Why check both instead of only the standalone pref?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -331,15 +340,24 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
         layout = { lastUpdate: Date.now(), ...defaultLayoutResp };
       }
     }
 
-    if (layout.spocs && this.config.spocs_endpoint) {
-      layout.spocs.url = this.config.spocs_endpoint;
+    if (
+      layout.spocs &&
+      (this.config.spocs_endpoint ||
+        Services.prefs.getCharPref(PREF_SPOCS_ENDPOINT, ""))
```
Mardak: Looks like this ordering always prefers the value from hardcoded object.

ScottDowne: Looking, but looks like a good catch.

lib/DiscoveryStreamFeed.jsm
ScottDowne: I figured if either of them were set, we would be pretty safe to enable the basic layout, right?

lib/DiscoveryStreamFeed.jsm
ScottDowne: Ah, I see what's happening, both of these prefs are optional. So `this.config.spocs_endpoint` is also empty by default. Ideally we would change 1 with code and the other with FF pref switch. I think the order is likely irrelevant, but I do think switching it is less confusing and more intuitive so I think I'll switch it.

test/browser/browser_discovery_render.js
```diff
@@ -8,6 +8,9 @@ async function before({ pushPrefs }) {
       enabled: true,
       hardcoded_layout: true,
     }),
+  ], [
+    "browser.newtabpage.activity-stream.discoverystream.enabled",
+    true,
```
Mardak: Probably could set this in browser.ini as individual tests determine if config.enabled is true or unset:
https://github.com/mozilla/activity-stream/blob/04220818451a4695f32298db08d325a1862892cf/test/browser/browser.ini#L6-L8

ScottDowne: Trying!

ScottDowne: Pushing up a second attempt that should get this going.

lib/TopStoriesFeed.jsm
```diff
@@ -65,7 +67,10 @@ this.TopStoriesFeed = class TopStoriesFeed {
     // if needed lazy load activity stream top stories feed based on
     // actual user preference when INIT and PREF_CHANGED is invoked
     this.discoveryStreamEnabled =
-      ds && ds.value && JSON.parse(ds.value).enabled;
+      ds &&
+      ds.value &&
+      JSON.parse(ds.value).enabled &&
+      Services.prefs.getBoolPref(DISCOVERY_STREAM_PREF_ENABLED, false);
```
ScottDowne: @Mardak So this is defaulted to `false/undefined`.

What this means is as soon as we ship this patch as is, beta and nightly revert to AS.

You think it's worth doing a check here to see if we're beta/nightly/unbranded and still turning this on? But then we couldn't test that this pref turns things off.

punamdahiya: may be we should check here if [prefHasuserValue](https://searchfox.org/mozilla-central/source/modules/libpref/nsIPrefBranch.idl#) that is an explicit set value, if not we ignore

punamdahiya:
QA steps for testing  pref "browser.newtabpage.activity-stream.discoverystream.hardcoded-basic-layout" can be changed to not include "browser.newtabpage.activity-stream.discoverystream.enabled" to true step in it as it should assume DS enabled with AS or FF pref


Similarly QA steps for "browser.newtabpage.activity-stream.discoverystream.spocs-endpoint" should not assume "browser.newtabpage.activity-stream.discoverystream.hardcoded-basic-layout" to true and work irrespective of other prefs. Noticed in browser toolbox network tab we don’t need restart browser. Disable and Enable DS from about:newtab#devtools-ds clears cache and loads spoc using new endpoint

lib/TopStoriesFeed.jsm
Mardak: Rollout uses default pref branch, so checking "user" prefs will not work.

ScottDowne:
@punamdahiya It needs both to be set to true to work.

ScottDowne:
OK tests and pref observer should both be good now.

lib/PrefsFeed.jsm
```diff
@@ -84,6 +84,31 @@ this.PrefsFeed = class PrefsFeed {
       value: handoffToAwesomebarPrefValue,
     });
 
+    const discoveryStreamEnabled = Services.prefs.getBoolPref(
+      "browser.newtabpage.activity-stream.discoverystream.enabled"
+    );
```
punamdahiya: errors when pref not found, need a fall back value for both discoverystream.enabled and discoverystream.hardcoded-basic-layout e.g. Services.prefs.getBoolPref("browser.newtabpage.activity-stream.discoverystream.enabled", false); This still leaves enabling in nigthly and beta when pref not found unresolved. Should we set discoverystream.enabled pref in nightly and beta with value as true?

ScottDowne: Ah so that's what's wrong.

punamdahiya: and may be setting[ discoverystream.enabled true here](https://github.com/mozilla/activity-stream/blob/master/lib/ActivityStream.jsm#L140) as DS is on by default in nightly and beta

ScottDowne: Also I think we're fixing the prefs for nightly/beta and default prefs in the uplift. That said, I remember hearing that it needs to be off in beta for the rollout to work properly.

PR #5255, opened 2019-08-14
WIP - Adding topsites spoc component

JeaneC:
Screen Shot 2019-08-14 at 11 23 16 AM
Adding a spoc to TopSites.

Given that TopSites currently doesn't take in a data/feed prop and just gets the data from the store directly, thinking just to inject the TopSite spoc in the TopSites component. 

Expecting a spocData object to get passed into TopSites in DiscoveryStreamBase.
```
spocData = {
   position: 7,
   topSiteSpoc: {
     // normal spoc info
    image: "",
    sponsor: "",
    url: "",
    ...
  }
}
```


JeaneC:
Drag and Drop fixed by adding tippyTopIcon as a parameter for Pinning new top sites (previously only accepted customScreenshots).

Screen Recording 2019-08-14 at 4 58 33 PM

JeaneC:
@wolasi thoughts?
Including a try build to test out these changes - [link here](https://queue.taskcluster.net/v1/task/DAQbJ36DQju7kRNfhgzNTA/runs/0/artifacts/public/build/target.dmg ).

Config row count of topsites in "about:preferences#home"
To "Reset' the Spoc (after dragging and dropping the spoc somewhere)
1. go to about:config
2. searched "pinned"  and edit "browser.newtabpage.pinned" to equal "[]"
3. Hard restart the browser (quit)

wolasi:
@JeaneC feedback on the build…

**Context Menu Button**
Noticed an odd issue with the context menu button for the sponsored tile.
Context: sponsored top site in 8th position, not pinned
Steps to Reproduce
1. open context menu for sponsored tile
2. click outside context menu to close
3. context menu button stays open

**Dark Mode**
"Sponsored" should be gray 40 (#b1b1b3) in dark mode

**Pinning UX**
Pinning sponsored Top Site in 8th position sends it to the first position, image disappears, and title changes to URL. After a few seconds a screenshot is show in place of the original image.

**Dismiss UX**
When a sponsored top site is dismissed another sponsored top site is shown. What should happen is the user should not see another sponsored top site for the rest of that session. A new sponsored top site should be shown in the next session.

JeaneC:
@wolasi new build w/ fixes - [link](https://queue.taskcluster.net/v1/task/XNkLmRoITXOAFMVA8zTU4A/runs/0/artifacts/public/build/target.dmg)

- The item will now render on the lastmost position in the row (so no more gaps)
- It also won't render if the user has no TopSites (failsafe against the default empty initial state on any load, but can change this)
- Won't be able to address dismiss or new SPOC behavior in this patch. Those will likely be from "backend" changes

wolasi:
New build looks great! 👍

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -31,8 +31,16 @@ export class _TopSites extends React.PureComponent {
       url: topSiteSpoc.url,
     };
 
-    const position = Math.min(topSites.length, spocPosition);
-    topSites[position] = link;
+    if (spocPosition === 0) {
```
punamdahiya: seeing a bug while pinning a non-spoc topsite when there is a spoc at position 0, looks like we need to bump up insert index to account for a spoc at zero position

JeaneC: Ahhh. Good catch, but just to verify do we want this extra behavior? Because then there's an opportunity for SPOCs to appear in the middle of the row. @wolasi

Bug
Bug

Fix
Bug Fix

JeaneC: Just realized too that this will happen in the end position also.... It would be easier if we have Pinned things as the highest priority then, and so the SPOC would go to the first available position that's not pinned.

Screen Recording 2019-08-19 at 9 22 16 AM

wolasi: Good catch, let's fix both bugs if possible.

lib/TopSitesFeed.jsm
```diff
@@ -599,6 +602,9 @@ this.TopSitesFeed = class TopSitesFeed {
     if (searchTopSite) {
       toPin.searchTopSite = searchTopSite;
     }
+    if (tippyTopIcon) {
+      toPin.tippyTopIcon = tippyTopIcon;
```
punamdahiya: To support pinning of sponsored top sites, we are making spoc site as orgainc site and saving its third party image url  in browser.newtab.pinned pref. This raises concern as we shouldn't be supporting pinned image url for-ever. Options here are:

a) Treat pinned spoc site as any other organic site that shows image as screenshot image url. This might look as weird as after pinning image icon will change to screenshot image.

b) Not support pin and drag and drop of spoc item and have it show up fixed in topsites for experiments

@wolasi including for help clarify this flow. Thanks

JeaneC: https://bugzilla.mozilla.org/show_bug.cgi?id=1575063

We have a delay behavior that prevents us from using custom images (exists in master already). The bug only shows custom images as icons after a hard browser reset. Digging into this bug and if we can fix this then we can keep the original expected behavior instead of the alternative treatment above. Hopefully can have a response end of day.

wolasi: I think a) is the better option. Why can't we use the image from adzerk as the image when it becomes organic?

content-src/components/TopSites/TopSite.jsx
```diff
@@ -172,12 +173,16 @@ export class TopSiteLink extends React.PureComponent {
...
+      // TopSite spoc experiment only
+      const spocImgURL = link.type === "SPOC" ? link.customScreenshotURL : "";
+
       imageClassName = "top-site-icon rich-icon";
       imageStyle = {
         backgroundColor: link.backgroundColor,
         backgroundImage: hasScreenshotImage
           ? `url(${this.state.screenshotImage.url})`
-          : "none",
+          : `url(${spocImgURL})`,
```
punamdahiya: @Mardak I believe concern [raised here](https://github.com/mozilla/activity-stream/pull/5265#issuecomment-523109143), doesn't apply in above case when we specifically use customScreenshotURL to display a non-pinned spoc tile. Is that correct?

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -113,7 +113,9 @@ export class _DiscoveryStreamBase extends React.PureComponent {
       case "Highlights":
         return <Highlights />;
       case "TopSites":
-        return <TopSites header={component.header} />;
+        return (
+          <TopSites header={component.header} spocData={component.spocData} />
```
punamdahiya: lets change property here to match back-end data={component.data}

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -4,14 +4,74 @@
...
+      !TopSites.rows ||
+      TopSites.rows.length === 0 ||
+      !data.spocs ||
+      data.spocs.length === 0
+    ) {
+      return null;
+    }
+
+    const topSites = [...TopSites.rows];
+    const [topSiteSpoc] = data.spocs;
```
punamdahiya: My understanding spocs data passed to TopSites can have more than one spocs @ScottDowne is that correct?

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -4,14 +4,74 @@
...
+    const organicTopSiteExists = topSites.some(
+      topSite =>
+        topSite && topSite.url === topSiteSpoc.url && topSite.type !== "SPOC"
+    );
+
+    if (organicTopSiteExists) {
+      return null;
```
punamdahiya: instead of returning should we should next spoc if available. This will help avoid an extra filterPinned check in DiscoveryStreamFeed.jsm

content-src/lib/link-menu-options.js
```diff
@@ -161,6 +161,18 @@ export const LinkMenuOptions = {
...
     }),
   }),
+  PinSpocTopSite: (site, index) => ({
+    id: "newtab-menu-pin",
+    icon: "pin",
+    action: ac.AlsoToMain({
+      type: at.TOP_SITES_PIN,
+      data: {
+        site,
+        index,
```
punamdahiya: may be in a followup bug we send some data that shows its a spoc tile that's pinned, similar to search topsite below

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
JeaneC: Ah, I'm destructuring the first element here, but to your other comment I should find the first non organic SPOC

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -4,14 +4,104 @@
 
 import { connect } from "react-redux";
 import { TopSites as OldTopSites } from "content-src/components/TopSites/TopSites";
+import { TOP_SITES_MAX_SITES_PER_ROW } from "common/Reducers.jsm";
 import React from "react";
 
 export class _TopSites extends React.PureComponent {
+  getNonOrganicSpoc(topSites, data) {
```
punamdahiya: non-organic spoc could be confusing, my understanding this function filters and help retrieve first available spoc thats not already an organic topsite. Consider renaming to getFirstAvailableSpoc with a comment explaining what it does

test/unit/content-src/components/DiscoveryStreamComponents/TopSites.test.jsx
```diff
@@ -49,11 +61,126 @@ describe("Discovery Stream <TopSites>", () => {
...
+
+    it("Should return null if an organic SPOC topsite exists", () => {
+      const topSitesWithOrganicSpoc = {
+        rows: [...defaultTopSiteRows, topSiteSpoc],
+      };
+
+      assert.isNull(insertSpocContent(topSitesWithOrganicSpoc, data, "right"));
+    });
+
+    it("Should return the next available non organic spoc if the first SPOC is organic", () => {
```
punamdahiya: nit: as commented above instead of non organic spoc something like "Should return next spoc if the first SPOC is an existing organic top site"

test/unit/content-src/components/DiscoveryStreamComponents/TopSites.test.jsx
```diff
@@ -49,11 +61,126 @@ describe("Discovery Stream <TopSites>", () => {
...
+        ],
+      };
+
+      const result = insertSpocContent(
+        topSitesWithOrganicSpoc,
+        extraSpocData,
+        "right"
+      );
+
+      const nonOrganicResultSpoc = {
```
punamdahiya: nit: and may be rename to 'availableSpoc' here?

content-src/components/TopSites/TopSite.jsx
```diff
@@ -493,6 +506,19 @@ export class TopSiteList extends React.PureComponent {
         break;
       case "drop":
         if (index !== this.state.draggedIndex) {
+          if (this.state.draggedSite.type === "SPOC") {
```
JeaneC: @Mardak @punamdahiya - This will be deleted! I initially had this because I was storing image URLs in tippytopicon and we didn't handle that in insert.

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -4,14 +4,138 @@
...
+      topSites.map(topSite => topSite && topSite.url)
+    );
+
+    // We "clean urls" with http in TopSiteForm.jsx
+    // Spoc domains are in the format 'sponsorname.com'
+    return spocs.find(
+      spoc =>
+        !userTopSites.has(spoc.url) &&
+        !userTopSites.has(`http://${spoc.domain}`) &&
+        !userTopSites.has(`http://www.${spoc.domain}`)
```
punamdahiya: This failed for https://www.netflix.com/ with spoc.url as https://netflix.com returned from adzerk server
dupe-netflix

punamdahiya: For left alignment, organic sites menu option edit is broken

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
```diff
@@ -4,14 +4,138 @@
...
+    );
+  }
+
+  // Find the first empty or unpinned index we can place the SPOC in.
+  // Return -1 if no available index and we should push it at the end.
+  getFirstAvailableIndex(topSites, promoAlignment) {
+    if (promoAlignment === "left") {
```
punamdahiya: For left alignment, organic sites menu option edit is broken

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
JeaneC: Should we just add two extra checks here for https:// and https://www, that should cover almost all the cases

punamdahiya: +1 considering its still an experiment and should cover bulk of the cases

content-src/components/DiscoveryStreamComponents/TopSites/TopSites.jsx
JeaneC: 😅, nice bug catch. 

It looks like for the Edit Option we pass in the site data all the way from the parent TopSites in AS, which will use the original TopSites.

https://github.com/mozilla/activity-stream/blob/d9ef50f2db4d0be7b7d31a757687586b78e061c1/content-src/components/TopSites/TopSites.jsx#L180-L185

content-src/components/TopSites/TopSites.jsx
```diff
@@ -71,7 +71,9 @@ export class _TopSites extends React.PureComponent {
    * Dispatch session statistics about the quality of TopSites icons and pinned count.
    */
   _dispatchTopSitesStats() {
-    const topSites = this._getVisibleTopSites();
+    const topSites = this._getVisibleTopSites().filter(
+      topSite => topSite !== null && topSite !== undefined
+    );
     const topSitesIconsStats = countTopSitesIconsTypes(topSites);
```
JeaneC: SPOC will currently count as a "custom_screenshot", increasing that count by 1. We can add a property here called `spoc or spoc_count`

https://github.com/mozilla/activity-stream/blob/d9ef50f2db4d0be7b7d31a757687586b78e061c1/content-src/components/TopSites/TopSites.jsx#L51-L59

And add a check for link.type === 'SPOC'
https://github.com/mozilla/activity-stream/blob/d9ef50f2db4d0be7b7d31a757687586b78e061c1/content-src/components/TopSites/TopSites.jsx#L21-L24

JeaneC:
Can we create a property here (card_type = 'spoc') by checking link.type === 'spoc'. This would cover the drop, pinning statistic conversion.
https://github.com/mozilla/activity-stream/blob/0379c988b17bcdef56bd8ac880368b0e636a2035/content-src/components/TopSites/TopSite.jsx#L283-L288

content-src/components/TopSites/TopSites.jsx
punamdahiya: +1 to fix in followup patch to update spoc topsites telemetry

punamdahiya:
> Can we create a property here (card_type = 'spoc') by checking link.type === 'spoc'. This would cover the drop, pinning statistic conversion.
> 
> https://github.com/mozilla/activity-stream/blob/0379c988b17bcdef56bd8ac880368b0e636a2035/content-src/components/TopSites/TopSite.jsx#L283-L288

yes, should do that in telemtry patch for spoc topsites

PR #5257, opened 2019-08-15
Bug 1544527 - Tabbable Search box and pref button

AllegroFox:


AllegroFox:
The tabIndex=1 on the pref button isn't great, but given that it's rendered at the bottom of the document it's default tab position is right at the end, which doesn't make sense for the layout. If there's a preferred way of fixing that let me know!

content-src/components/Base/Base.jsx
```diff
@@ -19,6 +19,8 @@ const PrefsButton = props => (
       className="icon icon-settings"
       onClick={props.onClick}
       data-l10n-id="newtab-settings-button"
+      // eslint-disable-next-line jsx-a11y/tabindex-no-positive
+      tabIndex="1"
```
Mardak: If the button should appear earlier in tab order, why not move the button to be earlier?

content-src/components/Search/Search.jsx
```diff
@@ -154,7 +154,7 @@ export class _Search extends React.PureComponent {
               data-l10n-id="newtab-search-box-search-the-web-input"
               ref={this.onSearchHandoffButtonMount}
               onClick={this.onSearchHandoffClick}
-              tabIndex="-1"
+              tabIndex="0"
```
Mardak: Was this set to -1 to begin with to avoid some problems of it getting focus?

AllegroFox: The focus was turned off for an experiment, I think.

content-src/components/Base/Base.jsx
AllegroFox: I can do that - I wasn't sure if it was rendered at the bottom of the body for a reason. (it's a component placed at the end of the body wrapper, after sections.)

Mardak: Does the pref button need to appear earlier in tab order? It should be tabble right now as the last item.

AllegroFox: It is tabbable as the last item, but I don't know if that's what we want given that it's positioned to be a the top of the page - logically to me you'd be able to tab to it before reaching the rest of the content. Moving the component doesn't seem to interfere with anything else, but is there someone else we should check with before deciding?

AllegroFox: I've moved it to the top for now, I can always move it back if it turns out there was a reason :)

AllegroFox: According to Wolasi this should work fine with the prefs button rendered first, so we should be good to go!

content-src/components/Search/Search.jsx
Mardak: If it were turned off for an experiment, then why? If it's because there are bugs, then we should fix them before changing it from -1. Does this even work correctly with tabindex=0?

AllegroFox: It was turned off while the search handoff to the awesomebar was an experiment, I believe - I got the go-ahead from the folks that worked on it that it should be ready to turn back on. It just puts it back into the regular tab order for the page :)

https://bugzilla.mozilla.org/show_bug.cgi?id=1544527

Mardak: Where in the bug does it say tabindex=-1 was for the experiment and it's good to set to tabindex=0?

PR #5259, opened 2019-08-16
Bug 1570983 - The What's New Panel is wrongly dismissed

piatra:


piatra:
As far as I can tell the panel has an event listener attached that captures any type of interaction. It uses it to check if there are any actions it could take as a result of the interaction (like `doCommand`) and since it can't find anything it calls to hide the panel. 
I realized the bug in this PR is that 
```javascript
    panelContainer.addEventListener("click", removeEventListener, {
      once: true, // This shouldn't be here, only captures 1 wrong click.
    });
```
So we can attach an event listener that prevents the default one from running.
Ideally the click targets would be big enough that a wrong click is not possible. I'll try to explore this a little more.

piatra:
I got a simpler solution based on top of #5267 where I increase the click area of the message.

Mardak:
Just making sure, this PR should just be the last commit? https://github.com/mozilla/activity-stream/pull/5259/commits/af9f07a3ad81f04d2aa0f3dc766c220feae34c04

Looks like this might be a reasonable simpler approach. However, we probably want to adjust the css to take advantage of the larger click targets. Here's setting the background to red for the `.whatsNew-message`:
Screen Shot 2019-08-23 at 6 00 51 PM

Notice there's the non-message space on the sides and in between. The sides are probably padding from the parent and the in-between border line area is probably a margin. Clicking those areas with this PR causes everything to get dismissed.

piatra:
@Mardak Is this with the attached phabricator patch?

PR #5260, opened 2019-08-16
Bug 1570332 - Expose user prefs values to targeting expressions

piatra:
I'm exposing the `cfrFeatures` and `cfrAddons` user prefs to be used in targeting expressions. We can then update targeting on required messages to prevent showing them.

I've combined all `defineModuleGetter` calls into 1 `defineLazyModuleGetters` call and converted all preference accesses to use `defineLazyPreferenceGetter`. 
It's not very thorough but I am seeing some small improvements comparing two profiles with/without (compared 3 separate runs and results seems in favor of profile 2 every time).

image

Profile 1 is without the patch Profile 2 is with.
https://profiler.firefox.com/compare/calltree/?globalTrackOrder=0-1-2&implementation=js&localTrackOrderByPid=88621%20from%20profile%201-0~89339%20from%20profile%202-0~&profiles[]=https%3A%2F%2Fperfht.ml%2F3066ph6&profiles[]=https%3A%2F%2Fperfht.ml%2F306Fkdu&search=targeting&thread=2&v=4

### DIFF
--- a/lib/ASRouterTargeting.jsm
+++ b/lib/ASRouterTargeting.jsm
@@ -2,6 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const SEARCH_REGION_PREF = "browser.search.region";
+const FXA_ENABLED_PREF = "identity.fxaccounts.enabled";
+
 const { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
 const { XPCOMUtils } = ChromeUtils.import(
   "resource://gre/modules/XPCOMUtils.jsm"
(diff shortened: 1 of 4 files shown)


lib/ASRouterTargeting.jsm
```diff
@@ -417,10 +450,13 @@ const TargetingGetters = {
     return null;
   },
   get isFxABadgeEnabled() {
-    return Services.prefs.getBoolPref(
-      "browser.messaging-system.fxatoolbarbadge.enabled",
-      false
-    );
+    return isFxABadgeEnabled;
+  },
+  get userPrefs() {
```
k88hudson: This will need some refactoring after groups, but works for now. Do you want to also add the snippets pref in here?

PR #5262, opened 2019-08-16
Bug 1574549 - Small bug with pref change, not critical

ScottDowne:
To test:

1.    Ensure you can see sponsored content.
2.    Set browser.newtabpage.activity-stream.discoverystream.enabled to true.
3.    Set browser.newtabpage.activity-stream.discoverystream.spocs-endpoint to "https://getpocket.cdn.mozilla.net/v3/firefox/unique-spocs"
4.    Set browser.newtabpage.activity-stream.discoverystream.spocs-endpoint to "" or clear it.
5.    Go to "about:home#devtools-ds" and scroll down until you see "spocs_endpoint https://spocs.getpocket.com/spocs" and ensure you see sponsored content.

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -65,8 +65,7 @@ const PREF_SPOC_IMPRESSIONS = "discoverystream.spoc.impressions";
 const PREF_CAMPAIGN_BLOCKS = "discoverystream.campaign.blocks";
 const PREF_REC_IMPRESSIONS = "discoverystream.rec.impressions";
 
-let defaultLayoutResp;
-let basicLayoutResp;
+let getHardcodedLayout;
 
 this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
   constructor() {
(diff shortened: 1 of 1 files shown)


lib/DiscoveryStreamFeed.jsm
```diff
@@ -353,9 +353,15 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       (this.store.getState().Prefs.values[PREF_SPOCS_ENDPOINT] ||
         this.config.spocs_endpoint)
     ) {
-      layout.spocs.url =
-        this.store.getState().Prefs.values[PREF_SPOCS_ENDPOINT] ||
-        this.config.spocs_endpoint;
```
Mardak: Alternatively, you can leave this code untouched and instead remove the global `basicLayoutResp` and `defaultLayoutResp` and do something like:
```js
 layout = { lastUpdate: Date.now(), ...getHardcoded("basic") };

function getHardcoded(which) {
  return which === "basic" ? { …the definition for basic layout goes here… } : { … };
}
```

So `getHardcoded()` always returns a new js object similar to how `fetchLayout` returns a new js object (from JSON.parse). Or I suppose maybe another alternative is make basicLayoutResp as a json string that gets parsed to create a new js object?

These other approaches might be safer if some future change always accidentally wants to change the original hardcoded layout object.

ScottDowne: Hm, actually now that I think about it.

1. We don't need to change the `spocs.url`, just when we fire `DISCOVERY_STREAM_SPOCS_ENDPOINT`we send the right thing, the rest of the code interacting with layout in this file should be looking at what's in redux, not the layout object.
2. With the above, further changes to the hard coded object like I did should be avoided. I think maybe a comment about not modifying the layouts is probably enough at that point.

ScottDowne: Trying a much simpler solution form my first one, but not sure how you think it compares to your ideas, thoughts?

ScottDowne:
I see your point, and I think that makes sense.

I think having to rely on a test/comment alone is probably not great. So I think I'm a fan of your initial proposal now.

ScottDowne:
I also noticed I never did add a test to this original patch, so I need to add a test anyway.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -357,47 +356,43 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
...
-      layoutResp.spocs.url =
-        this.store.getState().Prefs.values[PREF_SPOCS_ENDPOINT] ||
-        this.config.spocs_endpoint;
+          this.store.getState().Prefs.values[PREF_HARDCODED_BASIC_LAYOUT]
+      );
     }
 
     sendUpdate({
       type: at.DISCOVERY_STREAM_LAYOUT_UPDATE,
       data: layoutResp,
```
ScottDowne: The reducer in this dispatch actually ends up just ignoring the value in layoutResp.spocs, so changing it directly was also pointless. So I ended up optimizing with a function that returns a new object each time, but also, not modifying the original value.

ScottDowne:
@Mardak Did changes mostly around your nits, and added a test.

PR #5267, opened 2019-08-20
Bug 1571763 - Message template for tracking protection report in what's new panel

piatra:
I've modified `_createElement` to take optional arguments to add `classList` and set the strings for the element and return it.
```js
// Allows to replace
const el = this._createElement...
el.classList.add...
this._setString..
wrapper.appendChild(el)

// To now
wrapper.appendChild(
  this._createElement("h2", { classList, strings })
)
```

image
image

To do:
- [x] add image asset
- [x] tests


flodolo:
This won't work. See for reference how about:protections and monitor are managing having an external number displayed associated to a string
https://searchfox.org/mozilla-central/rev/428cf94f4ddfb80eebc6028023a7d89eb277791b/browser/locales/en-US/browser/protections.ftl#103

Note that the current version fails even for English and 1 blocked tracker.

locales-src/asrouter.ftl
```diff
@@ -90,6 +90,13 @@ cfr-whatsnew-button =
   .label = What’s New
   .tooltiptext = What’s New
 
+cfr-whatsnew-tracking-protections-title =
```
flodolo: Localizers need the comment, otherwise they won't know what's happening here, and to not add the variable again. You can copy and adapt the one I linked to.

lib/ToolbarPanelHub.jsm
```diff
@@ -181,6 +182,29 @@ class _ToolbarPanelHub {
...
+   * Attach click event listener defined in message payload
+   */
+  _attachClickListener(win, element, message) {
+    element.addEventListener("click", () => {
+      switch (message.content.layout) {
+        case "tracking-protections":
+          win.switchToTabHavingURI(message.content.cta_url, true);
+          break;
+        default:
+          win.ownerGlobal.openLinkIn(message.content.cta_url, "tabshifted", {
```
k88hudson: You might want to consider exposing access to `handleUserAction` in ASRouter since it would give you access to opening about pages, about:preferences, etc.

lib/ASRouterTargeting.jsm
```diff
@@ -422,6 +428,9 @@ const TargetingGetters = {
       false
     );
   },
+  get totalBlockedCount() {
```
piatra: For a very new profile this could be `0` so the message doesn't make much sense. The date is also messed up and returns the `0` timestamp for `Since ...` subtitle.

locales-src/asrouter.ftl
Mardak: @flodolo I'm not sure if the existing pattern works for this case of a title that goes before the number. Or at least it might want some extra clarification to localizers. The protections page example has numbers that appear immediately before the text:
Screen Shot 2019-08-22 at 12 37 04 AM

Here's the romanian strings for "Passwords exposed across all breaches.":
https://pontoon.mozilla.org/ro/firefox/browser/browser/protections.ftl/?string=201515
1 parolă …
2 parole …
20 de parole …

@piatra says a title starting with "De …" doesn't make sense, e.g.,
```
De parole …
20
```
while this is okay:
```
Parole …
20
```

I suppose romanian localization could just repeat an entry without "De ":
```
{ $blockedCount ->
  [one] Parolă …
  [few] Parole …
  [other] Parole …
```

The localization note already says "displayed **before** a large numeral…" so I maybe that might be enough information to decide to drop the "de " ??

Here's an example without the leading "de" that would normally go between the "other" number and the noun:
Screen Shot 2019-08-22 at 1 00 58 AM

I think even for english "Trackers Blocked: 1" makes sense. E.g., in a full sentence: "Here's how many trackers Firefox blocked: 1"

flodolo: > I'm not sure if the existing pattern works for this case of a title that goes before the number.

The comment clearly says that the numeral is before the string. Each locale will have the chance to decide if they need to decline the noun or use a generic plural and singular.

Mardak:
I'll rebase this on top of #5272 which includes the strings that were originally created in this PR, so this shouldn't have any ftl changes anymore.

Mardak:
Rebased and squashed in #5277

Screen Shot 2019-08-22 at 12 52 04 PM

lib/ToolbarPanelHub.jsm
```diff
@@ -137,23 +137,45 @@ class _ToolbarPanelHub {
...
+    const messages = (await this.messages).sort(this._sortWhatsNewMessages);
     const container = doc.getElementById(containerId);
 
     if (messages && !container.querySelector(".whatsNew-message")) {
       let previousDate = 0;
+      // Get and store any variable part of the message content
+      this.state.contentArguments = await this._contentArguments();
       for (let message of messages) {
+        // Only render date if it is different from the one rendered before.
+        if (message.content.published_date !== previousDate) {
```
Mardak: I see this code was just moved, but I guess we can just be careful to make sure to use the exact same date value down to the millisecond for messages that should share the same date header. Otherwise, we could do some math to truncate to the day.. although not sure if that has issues with time zones and daylight saving… let's just leave this as is 😛

lib/ToolbarPanelHub.jsm
```diff
@@ -137,23 +137,45 @@ class _ToolbarPanelHub {
     });
   }
 
+  // Newer messages first and use `order` field to decide between messages
+  // with the same timestamp
+  _sortWhatsNewMessages(m1, m2) {
+    // Sort by published_date in descending order.
+    if (m1.content.published_date === m2.content.published_date) {
+      // Ascending order
+      return m1.order - m2.order;
```
Mardak: Do we want to make `order` required? Otherwise some undefined/NaN sorting behavior

lib/PanelTestProvider.jsm
```diff
@@ -101,38 +102,63 @@ const MESSAGES = () => [
...
+      subtitle: { string_id: "cfr-whatsnew-tracking-blocked-subtitle" },
+      icon_url:
+        "resource://activity-stream/data/content/assets/protection-report-icon.png",
+      icon_alt: "Protection Report icon",
+      body: { string_id: "cfr-whatsnew-tracking-protect-body" },
+      link_text: { string_id: "cfr-whatsnew-tracking-blocked-link-text" },
+      cta_url: "protections",
+      cta_type: "OPEN_ABOUT_PAGE",
+    },
+    targeting: `firefoxVersion > 69 && totalBlockedCount > 0`,
```
Mardak: Not sure if it's because devtools doing something, but I ended up seeing this message showing 0 blocked on a new profile (with appropriate testing prefs set). So some reason the targeting `totalBlockedCount > 0` isn't being honored ??

lib/ToolbarPanelHub.jsm
```diff
@@ -219,81 +243,142 @@ class _ToolbarPanelHub {
...
+      // Keys need to match variable names used in asrouter.ftl
+      // `earliestDate` will be either 6 weeks ago or when tracking recording
+      // started. Whichever is more recent.
+      earliestDate: new Date(
+        Math.max(
+          new Date(await TrackingDBService.getEarliestRecordedDate()),
+          dateFrom
+        )
+      ).getTime(),
+      blockedCount: nf.format(totalEvents),
```
Mardak: Is this the new / expected way to do things? Otherwise plain JS `totalEvents.toLocaleString()` should work ?

piatra: Good point this is straightforward.
I knew Fluent formats numbers and I just looked into using their API.

lib/PanelTestProvider.jsm
piatra: The testing is confusing: [local test providers override message targeting](https://github.com/mozilla/activity-stream/blob/master/lib/PanelTestProvider.jsm#L116). This is ok when you click `Show` to force a message to appear but makes it weird for interactions that have multiple steps like What's New (force show the badge -> then open the dropdown).
I'll try to find a way to improve this but for now it's expected.

lib/ToolbarPanelHub.jsm
piatra: I'm counting on the staging environment + QA to catch any issues, we haven't seen any timezone weirdness yet. Probably because it's rounded down to a day not something more precise.

lib/ToolbarPanelHub.jsm
piatra: I think it's simpler to make it required. I'll update the PR.

Mardak:
Rebased on master with the changed number of test messages. And restored the icon.

PR #5271, opened 2019-08-21
Bug 1575523 - Fixed snippet resizing at smallest breakpoint.

AllegroFox:


AllegroFox:
Screen-Recording-2019-08-21-at-5 01 32-PM

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -91,7 +91,11 @@
     }
 
     @media (max-width: 865px) {
-      margin: 0 60px 0 0;
+      inset-inline-end: 60px;
```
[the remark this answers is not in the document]
Mardak: This probably could just be removed if there's nothing special to handle at this breakpoint?

Mardak:
Looks like there's some oddness at exactly 610px:
Screen Shot 2019-08-21 at 3 04 34 PM

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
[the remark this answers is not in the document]
AllegroFox: You're right that that's not the right rule, I checked again; but the rule isn't actually the same as the above, since what we want at that breakpoint is to push the snippet in line with the search bar, by having the margin only at the end of the line (so 0 60  0 0 instead of 0 60, which would put 60px at both the start and end). 

EDIT: I think what we need is `margin-inline-end`, plus a margin of 0 to keep the next breakpoint from kicking in.

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -174,6 +182,10 @@
         margin-top: 10px;
       }
 
+      @media (max-width: 610px) {
+        margin: auto;
```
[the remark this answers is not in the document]
AllegroFox: I think that's an artifact from when we had it in the top corner instead of centered. Removed!

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -91,7 +91,11 @@
     }
 
     @media (max-width: 865px) {
-      margin: 0 60px 0 0;
+      inset-inline-end: 60px;
+    }
+
+    @media (max-width: 610px) {
```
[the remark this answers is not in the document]
AllegroFox: done!

AllegroFox:
> Looks like there's some oddness at exactly 610px:
> Screen Shot 2019-08-21 at 3 04 34 PM

Looks like that's caused by the `margin: auto` being applied to the snippet slightly before all the other page styles kick in - is there a better way to fix this than just applying `margin: auto` just below the breakpoint instead? (I tried that, and it works, but...ew.)

Mardak:
Did you mean to push an update to your branch? The PR is still showing the old changes without the fixes

AllegroFox:
> Did you mean to push an update to your branch? The PR is still showing the old changes without the fixes

Yes. >:(

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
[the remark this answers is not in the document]
Mardak: This margin isn't necessary if there's no `margin-top` to override. It's already set to `auto`

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -91,7 +91,12 @@
     }
 
     @media (max-width: 865px) {
-      margin: 0 60px 0 0;
+      margin: 0;
+      margin-inline-end: 60px;
```
Mardak: Do we really want the icon to be on the starting edge of the hover background? If so, you can simply do `margin-inline-start: 0` as the normal behavior otherwise would be to have 60px margin on the sides anyway.

Screen Shot 2019-08-22 at 2 21 26 PM

However, not doing anything special for 865px could be fine too ??
Screen Shot 2019-08-22 at 2 25 00 PM

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -91,7 +91,12 @@
...
+    }
+
+    @media (max-width: 609px) {
```
Mardak: Still better to have a reference variable than an seemingly arbitrary number. You should be able to do `$var - 1px` ? and a comment about the off-by-1 difference between min-width and max-width ?

Mardak: Or I suppose at least have a comment

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
AllegroFox: Aaron did ask for the icon to be lined up with the search box, so the line-end margin is just to make space for the dismiss button on the other side.

AllegroFox: makes sense to change it to margin-inline-start, though

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
AllegroFox: I'll set it that way with a comment - I don't know if it would be worth refactoring the breakpoints in future so that they overlap a little, to prevent those weird gaps?

AllegroFox:
Awesome, thanks! :)

PR #5272, opened 2019-08-22
Bug 1561370 - Land strings and images for Firefox 70 What's New messages

Mardak:
This is still WIP but probably good enough to get a first pass making sure we've got the desired strings. This is on top of #5267 to rename the ftl message ids for the blocked related items.

Theoretically https://docs.google.com/presentation/d/17BCPTuDWVSnLcLuFx-HFa4ESCpIH8A-Ss9V7F2m8rFQ/edit#slide=id.g5e3431b199_0_0 and https://docs.google.com/presentation/d/17BCPTuDWVSnLcLuFx-HFa4ESCpIH8A-Ss9V7F2m8rFQ/edit#slide=id.g5e44344997_3_0 are "finalized" but sounds like usage of "Protection Report" has some questions.

From those 2 pages, there's:
1) firefox 70 hero
2) protection report 
  a. for >= 100 blocked items
  b. fewer than 100 blocked items
3) lockwise
  a. password sync not on
  b. password sync on
4) firefox 70 release notes

### DIFF
--- a/locales-src/asrouter.ftl
+++ b/locales-src/asrouter.ftl
@@ -92,6 +92,42 @@ cfr-whatsnew-button =
 
 cfr-whatsnew-panel-header = What’s New
 
+cfr-whatsnew-release-notes-link-text = Read the release notes
+
+cfr-whatsnew-fx70-title = { -brand-short-name } now fights harder for your privacy
+cfr-whatsnew-fx70-body =
+   The latest update enhances the Tracking Protection feature and makes it
+   easier than ever to create secure passwords for every site.
+
+cfr-whatsnew-tracking-protect-title = Protect yourself from trackers
+cfr-whatsnew-tracking-protect-body =
+   { -brand-short-name } blocks many common social and cross-site trackers that
+   follow what you do online.
+cfr-whatsnew-tracking-protect-link-text = View your protection report
+
+# This string is displayed before a large numeral that indicates the total
+# number of tracking elements blocked. Don’t add $blockedCount to your
+# localization, because it would result in the number showing twice.
+cfr-whatsnew-tracking-blocked-title =
+  { $blockedCount ->
+    [one] Tracker blocked
+   *[other] Trackers blocked
+  }
+cfr-whatsnew-tracking-blocked-subtitle =
+   Since { DATETIME($earliestDate, month: "long", year: "numeric") }
+cfr-whatsnew-tracking-blocked-link-text = View Protection Report
+
+cfr-whatsnew-lockwise-backup-title = Back up your passwords
+cfr-whatsnew-lockwise-backup-body =
+   Now generate secure passwords you can access anywhere you sign in.
+cfr-whatsnew-lockwise-backup-link-text = Turn on backups
+
+cfr-whatsnew-lockwise-take-title = Take your passwords with you
+cfr-whatsnew-lockwise-take-body =
+   The { -lockwise-brand-short-name } mobile app lets you securely access your
+   backed up passwords from anywhere.
+cfr-whatsnew-lockwise-take-link-text = Get the app
+
 ## Bookmark Sync
 
 cfr-doorhanger-sync-bookmarks-header = Get this bookmark on your phone


locales-src/asrouter.ftl
```diff
@@ -90,17 +90,42 @@ cfr-whatsnew-button =
   .label = What’s New
   .tooltiptext = What’s New
 
+cfr-whatsnew-fx70-hero-title = { -brand-short-name } now fights harder for your privacy
+cfr-whatsnew-fx70-hero-body = The latest update enhances the Tracking Protection feature and makes it easier than ever to create secure passwords for every site.
+
+cfr-whatsnew-fx70-notes-body = Protection reports, Secure Proxy for public Wifi, improved account controls, and many more safety and security tools are included in the latest update.
```
Mardak: Looks like "Wi-Fi" is the correct way: https://searchfox.org/mozilla-central/search?q=wifi&path=dtd

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,42 @@ cfr-whatsnew-button =
...
+cfr-whatsnew-fx70-notes-link-text = Read the release notes
```
Mardak: Not all locales have release notes… I suppose we should make sure to remotely target this panel message to the appropriate locales. @piatra can we optionally not have a link and just have the body as a release-notes 1-liner?

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,42 @@ cfr-whatsnew-button =
...
+
+cfr-whatsnew-tracking-protect-title = Protect yourself from trackers
+cfr-whatsnew-tracking-protect-body = { -brand-short-name } blocks many common social and cross-site trackers that follow what you do online.
+cfr-whatsnew-tracking-protect-link-text = View your protection report
+
 # This string is displayed before a large numeral that indicates the total number
 # of tracking elements blocked. Don’t add $blockedCount to
 # your localization, because it would result in the number showing twice.
-cfr-whatsnew-tracking-protections-title =
+cfr-whatsnew-tracking-blocked-title =
```
Mardak: Renamed to `cfr-whatsnew-tracking-blocked-*` to differentiate slightly from the generic numberless version that I named `cfr-whatsnew-tracking-protect-*`

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,42 @@ cfr-whatsnew-button =
   .label = What’s New
   .tooltiptext = What’s New
 
+cfr-whatsnew-fx70-hero-title = { -brand-short-name } now fights harder for your privacy
+cfr-whatsnew-fx70-hero-body = The latest update enhances the Tracking Protection feature and makes it easier than ever to create secure passwords for every site.
```
flodolo: Side note: Fluent supports multiline strings, in case you don't want to have super long lines of text.

```
 cfr-whatsnew-fx70-hero-body = 
    The latest update enhances the Tracking Protection feature and makes it 
    easier than ever to create secure passwords for every site.
```

locales-src/asrouter.ftl
flodolo: Secure Proxy is US only, why are we exposing this to all locales?

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,42 @@ cfr-whatsnew-button =
...
+cfr-whatsnew-tracking-blocked-title =
   { $blockedCount ->
     [one] Tracker blocked
    *[other] Trackers blocked
   }
-cfr-whatsnew-tracking-protections-subtitle = Since { DATETIME($earliestDate, month: "long", year: "numeric") }
+cfr-whatsnew-tracking-blocked-subtitle = Since { DATETIME($earliestDate, month: "long", year: "numeric") }
+cfr-whatsnew-tracking-blocked-link-text = View Protection Report
+
+cfr-whatsnew-lockwise-backup-title = Back up your passwords
```
Mardak: Does this need a note? Interestingly in english "backup" is a noun while "back up" is the verb…

locales-src/asrouter.ftl
flodolo: This needs to land in m-c together with the patch adding the `-protections-` strings

locales-src/asrouter.ftl
flodolo: I don't thing so, it's the same as "log in" vs "login"

locales-src/asrouter.ftl
flodolo: All locales but en-US don't have release notes ;-)

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,42 @@ cfr-whatsnew-button =
...
+cfr-whatsnew-lockwise-backup-body = Now generate secure passwords you can access anywhere you sign in.
+cfr-whatsnew-lockwise-backup-link-text = Turn on backups
+
+cfr-whatsnew-lockwise-take-title = Take your passwords with you
+cfr-whatsnew-lockwise-take-body = The { -lockwise-brand-short-name } mobile app lets you securely access your backed up passwords from anywhere.
```
Mardak: Is this a correct usage of another message id as it's not quite simply a plain noun… ?

flodolo: This is not a normal message, it's a term, and it's designed to be only referenced in other messages.

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,41 @@ cfr-whatsnew-button =
...
+cfr-whatsnew-lockwise-take-link-text = Get the app
 
 cfr-whatsnew-panel-header = What’s New
+cfr-whatsnew-dismiss-button = Dismiss
+
+# Variables:
+#  $version (String) - Firefox browser version
+cfr-whatsnew-firefox-notes-title = { -brand-short-name } Version { $version }
+cfr-whatsnew-firefox-notes-link-text = Read the release notes
```
Mardak: Not all locales have release notes… I suppose we should make sure to remotely target this panel message to the appropriate locales. @piatra can we optionally not have a link and just have the body as a release-notes 1-liner?

locales-src/asrouter.ftl
```diff
@@ -90,17 +90,41 @@ cfr-whatsnew-button =
...
+cfr-whatsnew-lockwise-take-title = Take your passwords with you
+cfr-whatsnew-lockwise-take-body = The { -lockwise-brand-short-name } mobile app lets you securely access your backed up passwords from anywhere.
+cfr-whatsnew-lockwise-take-link-text = Get the app
 
 cfr-whatsnew-panel-header = What’s New
+cfr-whatsnew-dismiss-button = Dismiss
+
+# Variables:
+#  $version (String) - Firefox browser version
+cfr-whatsnew-firefox-notes-title = { -brand-short-name } Version { $version }
```
Mardak: This could just be a plain "{ -brand-short-name } Version 70" as otherwise the remote message would need to somehow specify args? `{string_id: "cfr_whatsnew-firefox-notes-title", args: { version: "70" } }` ?? @piatra I see #5267 just always passes in a common object for all strings. That's an interesting approach that could work too but perhaps we should name the args more explicitly if it's a shared "namespace", so all strings can use any args, `… { $firefoxVersion }` ?

locales-src/asrouter.ftl
Mardak: Ah good point. If this is supposed to be en-US/US only, we can maybe just provide the raw string in the remote message. I'll check generally what messages are supposed to be for all locales vs subset / regions.

locales-src/asrouter.ftl
Mardak: Nod. We'll do the export with the commits together (or maybe just rename the id in #5267)

locales-src/asrouter.ftl
Mardak: I was more concerned about it being used in the context of both "the" article and modifying "mobile app," but I suppose that's the point of fluent parameters to allow localizations to decide if there should be cases on the term ?

locales-src/asrouter.ftl
Mardak: Oh, I saw flod's comment on my previous line that I removed of only en-US has release notes. So I suppose this is another candidate for maybe not exposing to all locales and just do it from the remote en-US targeted message.

locales-src/asrouter.ftl
flodolo: Hardly any language can use that structure, so it will likely end up with something like "the mobile app (of) Lockwise)". Anyhow, that's expected and parameterized terms can help work around it.

locales-src/asrouter.ftl
flodolo: That's a product decision, it could be still worth sending people to release notes, even if they're not translated.

locales-src/asrouter.ftl
piatra: Yes I like this approach.

locales-src/asrouter.ftl
piatra: If it makes it easier: I see that string was removed from the designs and we were not planning on implementing it either. We can defer for later?

locales-src/asrouter.ftl
Mardak: Indented to 3 spaces and wrapped 80 characters

locales-src/asrouter.ftl
Mardak: The "release notes" is gone with the "hero" becoming a regular message with a link to unlocalized release notes.

locales-src/asrouter.ftl
Mardak: Moved to the plain message "hero"

Mardak: Jim says we'll keep this generic "Read the release notes" link to unlocalized and see how people interact which could be a signal in how we do release notes going forwards.

flodolo:
What's the status of this? I still see Wi-Fi, or the string that is supposed to be US only.

locales-src/asrouter.ftl
Mardak: Removing this particular string as we don't have "release notes" message anymore (but do have the "hero")

Mardak:
Sorry, I started commenting before actually pushing to the PR.

PR #5274, opened 2019-08-22
Bug 1575889 - Fixing a bit about how we flip the pref from AS to DS

ScottDowne:
To test:

1. Set `browser.newtabpage.activity-stream.discoverystream.enabled` to `false`
2. Open a new tab, keep it open.
3. Flip `browser.newtabpage.activity-stream.discoverystream.enabled` from `false` to `true`
4. Look at the tab you opened, expected, it shouldn't have changed.
5. Open a new tab
6. expected, it should be different

### DIFF
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -1263,13 +1263,41 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
       case at.PREF_CHANGED:
         switch (action.data.name) {
           case PREF_CONFIG:
-          case PREF_ENABLED:
-          case PREF_HARDCODED_BASIC_LAYOUT:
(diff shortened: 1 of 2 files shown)


Mardak:
Will strange things happen if the config doesn't exist in the feed and someone tries to interact with the "stale" page? E.g., scrolls down to a spoc which then tries to trigger an impression ?

gvn:
Taskcluster failed but I think it's not because of your code and related to: https://bugzilla.mozilla.org/show_bug.cgi?id=1575774

gvn:
With a fresh profile I'm not seeing a change at step 6.

ScottDowne:
@Mardak Should be good. I tested a bunch of stuff to be sure.

@gvn I just tested it with a new profile and it worked fine. I think maybe we can test this over zoom in a bit?

PR #5275, opened 2019-08-22
Test to see if TC builds work after removing `objdir`

piatra:
I've hit this and seems like TC is having the same issues 
See `dev-platform` email "Watch out for build issues on non-Unicode systems"


piatra:
The objdir is part of the Docker image to make the artifact download smaller. I might need to re-build it to work around this issue if this PR makes TC green.

PR #5277, opened 2019-08-22
Bug 1575931 - Add icon asset for Protection Report message

piatra:
Based on #5267


Mardak:
Ah I'll just squash this into #5267

PR #5278, opened 2019-08-23
Bug 1575887 - The CRF Pin Tabs recommendations are no longer displayed

piatra:


piatra:
I'll attempt to refactor that code.

PR #5279, opened 2019-08-23
Bug 1576207 - Fixed break point overlap at 1121px

AllegroFox:


AllegroFox:
> Should be using the variable math

> Should be using the variable math

Done! I didn't realize there was a breakpoint-widest because it's not used anywhere else.

PR #5282, opened 2019-08-23
Bug 1575797 - Something prevents structured clone of redux store resulting in 'Top Sites' and 'Highlights' empty: 'Oops, something went wrong loading this content. Refresh page to try again.'

Mardak:
r?@ScottDowne Here's a screenshot of directly putting a function and Promise into the store:
Screen Shot 2019-08-23 at 11 33 10 AM

### DIFF
--- a/lib/NewTabInit.jsm
+++ b/lib/NewTabInit.jsm
@@ -26,6 +26,34 @@ this.NewTabInit = class NewTabInit {
       type: at.NEW_TAB_INITIAL_STATE,
       data: this.store.getState(),
     };
+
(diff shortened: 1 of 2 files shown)


Mardak:
Potentially we can gate this behind `browser.newtabpage.activity-stream.debug` which happens to be readily available because we've already gotten the redux store Prefs.debug right there 😉

Mardak:
Or some other pref. That one seems to be a firefox.js pref for something else…

Mardak:
Favoring #5283 instead

PR #5283, opened 2019-08-23
Bug 1575797 - After downloading a file, 'Top Sites' and 'Highlights' empty: 'Oops, something went wrong loading this content. Refresh page to try again.'

Mardak:
r? @punamdahiya or @ScottDowne or @ncloudioj Save the referrer as a string and not a nsIURI.


Mardak:
Ah hold on. this originalReferrer might be null ;)

Mardak:
STR:
1) add top site url https://ftp.mozilla.org/
2) let the thumbnail load
3) open that top site (or just load https://ftp.mozilla.org/)
4) right-click favicon.ico and save link as and save it
5) go to about:newtab
6) refresh about:newtab

top sites should still show a thumbnail after step 6

PR #5284, opened 2019-08-24
test mainthreadio

Mardak:


Mardak:
Hmm.. not really sure how to run startup_mainthreadio on taskcluster here… I suppose I can't even get it running locally either………

Mardak:
Although then again, looks like on try, this test only failed on windows 7 and 10 while passing for linux and mac

piatra:
I'm trying to delay when the targeting getter is accessed by waiting on `browser-delayed-startup-finished`. [So far unsuccessful](https://treeherder.mozilla.org/#/jobs?repo=try&revision=2ea20d5502c50de738fb0de87eb2d333161cd11d), still looking into if that's the right event to wait on.

PR #5285, opened 2019-08-28
Bug 1577279 - mach lint complains about newtab due to eslint version mismatch

Mardak:
r+@dmose All of these changes were required to get eslint@6.2.2 working:
- eslint-plugin-mozilla@2.1.0 turns off some stuff from eslint@6.2.2, e.g., jsm builtinGlobals
- eslint@6.2.2 requires a file and not a path for `--print-config` so I used dummy file
- babel-eslint@10.0.3 is needed to get eslint@6.2.2 parsing correctly
- eslint-plugin-html has been a recommended plugin from eslint-plugin-mozilla for a while, but looks like it never got loaded 😛 https://eslint.org/docs/user-guide/migrating-to-6.0.0#-plugins-and-shareable-configs-are-no-longer-affected-by-eslints-location


yamscripts.yml
```diff
@@ -65,7 +65,7 @@ scripts:
 
 # lint: Run eslint and sass-lint
   lint:
-    eslint-check: eslint --cache --print-config . | eslint-config-prettier-check
+    eslint-check: eslint --cache --print-config .eslintrc.js | eslint-config-prettier-check
```
dmose: What's this change about?  I suspect that whatever we do, we want a comment describing why we've chosen the argument we have, if it's not a file that's part of the app, as suggested by the eslint-config-prettier readme.

package.json
```diff
@@ -21,7 +21,7 @@
     "@babel/plugin-proposal-async-generator-functions": "7.2.0",
     "@babel/preset-react": "7.0.0",
     "acorn": "6.1.1",
-    "babel-eslint": "10.0.1",
+    "babel-eslint": "10.0.3",
```
dmose: https://docs.google.com/document/d/1Oguqp43W4_ChyroJ9AJAzG1jSwkUWfKvBKVvrDxVsMg/edit#bookmark=id.81i7mkprevgg makes me think that for each of the packages that we're changing in this PR, we're going to have to check those parts of the dependency tree and mention that in the bug. The simplest way to do this at the moment would be to put the packages into broofa, like [this](http://npm.broofa.com/?q=eslint@6.2.2) to verify that those licenses match the runbook policies (yeah, we'll want to automate this soon, and I'll be filing a bug for that).

If @Standard8 beats us to the punch and annotates https://bugzilla.mozilla.org/show_bug.cgi?id=1540982 before you get to this, we can presumably just state that this has already been taken care of for mozilla-central.

yamscripts.yml
dmose: Ok, I missed the original description at the top of the PR, but I suspect we still want to just use a file from the app as suggested by that README, or else comment it.

package.json
Mardak: eslint@5.16.0:
Apache-2.0(4)
BSD-2-Clause(7)
BSD-3-Clause(3)
ISC(15)
MIT(82)
WTFPL OR MIT

and upgrading to eslint@6.2.2:
Apache-2.0(4)
BSD-2-Clause(7)
BSD-3-Clause(3)
ISC(16)
MIT(85)

so fewer types of licenses, so I guess we're okay here for eslint (and that m-c uses eslint anyway)

---

babel-eslint@10.0.1:
Apache-2.0
BSD-2-Clause(4)
BSD-3-Clause
MIT(26)

babel-eslint@10.0.3:
Apache-2.0
BSD-2-Clause
BSD-3-Clause
MIT(28)

so same set

---

adding eslint-plugin-html@6.0.0:
BSD-2-Clause(4)
ISC(2)
MIT(6)

subset of allowed licenses

---

eslint-plugin-mozilla@1.3.0:
MPL-2.0
Unspecified
BSD-2-Clause(4)
ISC(2)
MIT(6)

to eslint-plugin-mozilla@2.1.0:
MPL-2.0
Unspecified
BSD-2-Clause(4)
ISC(2)
MIT(6)

same set… including null/undefined but I suppose this plugin is from mozilla-central anyway

yamscripts.yml
Mardak: Switched to AboutNewTabService.jsm

package.json
dmose: Resolved as OK after discussion in the bugzilla bug.

PR #5286, opened 2019-08-28
Bug 1577345 - Intersection observer prop update bug

ScottDowne:
To test:

2. Update `browser.newtabpage.activity-stream.discoverystream.config` with `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"https://5ad1b408-b40a-4a49-b6ec-27473d192df0.mock.pstmn.io"}`

3. Open a new tab, should see a sponsored topsite.
4. Open the console.
5. Dismiss the sponsored topsite.

expected. no errors in the console.

### DIFF
--- a/content-src/components/DiscoveryStreamImpressionStats/ImpressionStats.jsx
+++ b/content-src/components/DiscoveryStreamImpressionStats/ImpressionStats.jsx
@@ -193,12 +193,6 @@ export class ImpressionStats extends React.PureComponent {
     }
   }
 
-  componentDidUpdate(prevProps) {
-    if (this.props.rows.length && this.props.rows !== prevProps.rows) {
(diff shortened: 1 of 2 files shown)


ncloudioj:
LGTM!

This impression wrapper was originally designed to be used to wrap the whole section (such as "Top Story", "Hero", "Grid" etc), hence properly handle `componentDidUpdate` was necessary to capture the state updates triggered by the users, e.g. dismissing a top story.

Now that this impression wrapper is only used for each individual card, I think we can have a refactor to simplify it. @ScottDowne Could you file a followup bug to track this?

PR #5287, opened 2019-08-28
Bug 1577348 - Use dsimage in dstextpromo; add alt text and rounding to dsimage

gvn:
To test:

2. Update `browser.newtabpage.activity-stream.discoverystream.config` with `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"https://5ad1b408-b40a-4a49-b6ec-27473d192df0.mock.pstmn.io"}`

expected: Should see a Netflix logo in the TextPromo with actual pixel size of 40x40 (or 80x80 depending on screen density)

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -143,6 +143,7 @@ export class _DiscoveryStreamBase extends React.PureComponent {
         const [spoc] = component.data.spocs;
         const {
           image_src,
+          raw_image_src,
           alt_text,
           title,
           url,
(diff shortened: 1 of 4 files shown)


punamdahiya:
@gvn when should img shown will be rendered in 80x80, not sure how to simulate screen density here?

punamdahiya:
Notice dismiss button is missing tooltip, we can add  title="Dismiss" as part of this patch which can be later picked from ftl file when we localize DSDismiss. This will help resolve accessibility issue around missing label on dismiss button.

https://github.com/mozilla/activity-stream/blob/master/content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx#L67

gvn:
@punamdahiya You're probably already seeing the 80x80px version (this is actual pixels, not measurement pixels used by the DOM) if you have a retina display. Only low pixel density displays will show the 40x40px (actual pixels) image source.

gvn:
Making follow-on ticket for tooltip: https://bugzilla.mozilla.org/show_bug.cgi?id=1577917

PR #5293, opened 2019-09-03
Bug 1561880 - Refactor browser window listeners to use EveryWindow.jsm

piatra:
There was a lot of code duplication between the two triggers so I'm happy we're able to remove most of it.
For testing I enabled the local CFRMessageProvider and set targeting to `"true"` to ensure everything triggers instantly.

### DIFF
--- a/lib/ASRouterTriggerListeners.jsm
+++ b/lib/ASRouterTriggerListeners.jsm
@@ -10,34 +10,14 @@ ChromeUtils.defineModuleGetter(
   "PrivateBrowsingUtils",
   "resource://gre/modules/PrivateBrowsingUtils.jsm"
 );
+ChromeUtils.defineModuleGetter(
+  this,
+  "EveryWindow",
+  "resource:///modules/EveryWindow.jsm"
+);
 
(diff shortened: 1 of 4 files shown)


lib/ASRouterTriggerListeners.jsm
```diff
@@ -10,35 +10,15 @@ ChromeUtils.defineModuleGetter(
...
+);
 
 const FEW_MINUTES = 15 * 60 * 1000; // 15 mins
 const MATCH_PATTERN_OPTIONS = { ignorePath: true };
 
-/**
- * Wait for browser startup to finish to avoid accessing uninitialized
- * properties
- */
-async function checkStartupFinished(win) {
```
k88hudson: Does `EveryWindow.registerCallback`remove the need to have to do this essentially?

piatra: Yes. The callback is [called only after delayedStartup](https://searchfox.org/mozilla-central/rev/588814f2edddf0e132d77d326ddae50911e8bad1/browser/modules/EveryWindow.jsm#39-41) finished.

PR #5295, opened 2019-09-04
Bug 1578236 - Add back the CFR message after navigating between tabs …

piatra:
…when a host is not defined

Issue was that upon coming back to a tab with the SAVE_LOGIN CFR, `showAddressNotifier` would not get called because the recommendation doesn't have a host defined.

Initially I also considered adding a "fallback" host when calling `addRecommendation` to be whatever website the browser was on when the message appeared. I don't think this would be a good solution because we might want to keep showing messages that don't have a host defined even when navigating to a different tab (the recommendation isn't tied to a particular website instead it advertieses a ffx feature).

TODO
- [x] backports
- [x] tests

I need to do some backports first before I can add tests.

### DIFF
--- a/test/unit/asrouter/CFRPageActions.test.js
+++ b/test/unit/asrouter/CFRPageActions.test.js
@@ -801,7 +801,7 @@ describe("CFRPageActions", () => {
         );
       });
       it("should show the pageAction if a recommendation exists and it doesn't have a host defined", () => {
-        const recNoHost = { ...savedRec, host: null };
+        const recNoHost = { ...savedRec, host: undefined };
         CFRPageActions.RecommendationMap.set(fakeBrowser, recNoHost);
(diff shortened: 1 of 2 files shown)


punamdahiya:
@piatra I am following STR from here https://bugzilla.mozilla.org/show_bug.cgi?id=1578236#c0 and still the issue with the patch, that is navigating back to tab doesn't show recommendation

test/unit/asrouter/CFRPageActions.test.js
```diff
@@ -800,6 +800,16 @@ describe("CFRPageActions", () => {
           savedRec
         );
       });
+      it("should show the pageAction if a recommendation exists and it doesn't have a host defined", () => {
+        const recNoHost = { ...savedRec, host: null };
```
punamdahiya: nit: should host value be undefined here instead of null

PR #5296, opened 2019-09-05
Bug 1575820 - Show extended triplets on new tab after sync login

punamdahiya:
Issue here is logging into sync causes services.sync.username pref to change which triggers [onPrefChange](https://github.com/mozilla/activity-stream/blob/master/lib/ASRouter.jsm#L523) adding Extended Triplets message  to list of invalid messages. 

Invalid messages gets cleared by [AS_ROUTER_TARGETING_UPDATE handler](https://github.com/mozilla/activity-stream/blob/master/content-src/asrouter/asrouter-content.jsx#L257)

Fix in patch assumes we want to show extended triplets if its visible while sync pref change.

### DIFF
--- a/lib/ASRouter.jsm
+++ b/lib/ASRouter.jsm
@@ -514,11 +514,13 @@ class _ASRouter {
     if (TARGETING_PREFERENCES.includes(prefName)) {
       // Notify all tabs of messages that have become invalid after pref change
       const invalidMessages = [];
+      const context = this._getMessagesContext();
+
       for (const msg of this._getUnblockedMessages()) {
         if (!msg.targeting) {
           continue;
         }
-        const isMatch = await ASRouterTargeting.isMatch(msg.targeting);
+        const isMatch = await ASRouterTargeting.isMatch(msg.targeting, context);
         if (!isMatch) {
           invalidMessages.push(msg.id);
         }


lib/ASRouter.jsm
```diff
@@ -510,12 +511,19 @@ class _ASRouter {
...
   async onPrefChange(prefName) {
     if (TARGETING_PREFERENCES.includes(prefName)) {
       // Notify all tabs of messages that have become invalid after pref change
       const invalidMessages = [];
       for (const msg of this._getUnblockedMessages()) {
-        if (!msg.targeting) {
+        if (!msg.targeting || this._shouldShowExtendedTriplet(msg)) {
           continue;
         }
         const isMatch = await ASRouterTargeting.isMatch(msg.targeting);
```
Mardak: Hm.. why is the extended triplets message no longer matching after a pref change? Its targeting doesn't explicitly check for logged in ??
https://github.com/mozilla/activity-stream/blob/2df1c58f59b34658d7eba6518b16ab6f51a2f11a/lib/OnboardingMessageProvider.jsm#L111-L112

k88hudson: yeah why is `isMatch` failing?

punamdahiya: not sure about https://searchfox.org/mozilla-central/source/toolkit/components/utils/FilterExpressions.jsm#54 code here which is what isMatch uses?

Mardak: Looks like this `isMatch` is missing a 2nd argument for `context` ?

Mardak: `evaluateExpression` passes in a context while `findAllMessages` just uses a helper method:
https://github.com/mozilla/activity-stream/blob/2df1c58f59b34658d7eba6518b16ab6f51a2f11a/lib/ASRouter.jsm#L1132-L1142

punamdahiya: Noticed context used with FilterExpressions on PrefChange is missing 
```
messageImpressions: {…}
previousSessionEnd: 0
trailheadInterrupt: "join"
trailheadTriplet: "supercharge"
```

will try passing customContext using helper method and see if that helps
context

punamdahiya:
@Mardak Updated patch by passing in custom context while checking targeting for all messages. It's much better one line fix that should be an easy uplift. Thanks!

lib/ASRouter.jsm
```diff
@@ -518,7 +518,8 @@ class _ASRouter {
         if (!msg.targeting) {
           continue;
         }
-        const isMatch = await ASRouterTargeting.isMatch(msg.targeting);
+        const context = this._getMessagesContext();
```
punamdahiya: On second thought this context should be independent of messages and should be safe to call once, updating patch with the change

punamdahiya: Done

PR #5298, opened 2019-09-05
Bug 1561942 - Messaging System devtools support for What's New panel

piatra:
So far the `Show` button in our devtools didn't work for what's new messages.
With this PR clicking show will
* set the panel content to that specific message
* restore the original panel content after the panel is hidden

Works best with the `Disable Pop-up Autohide` feature of the Browser Toolbox but it's not required (clicking show and then opening the panel also works).

No support for multiple messages at once.

## STR

0. Have

Pref name | Prev Value
--- | ---
browser.newtabpage.activity-stream.asrouter.providers.panel_local_testing | {"id":"panel_local_testing", "cohort": "SHOW_TEST"}
browser.newtabpage.activity-stream.asrouter.devtoolsEnabled | true
1. Confirm the panel is visible and populated with messages (might require a restart)

image
image

1. In ASRDevtools (`about:newtab#asrouter`)
2. Select the `panel_local_test` provider
3. Force a message to show by clicking the button
image

## Expected

1. Opening the panel again only shows that message
image
 
2. After closing the panel and opening it again the messages from STR step 1 are restored.

### DIFF
--- a/lib/ToolbarPanelHub.jsm
+++ b/lib/ToolbarPanelHub.jsm
@@ -27,6 +27,7 @@ const APPMENU_BUTTON_ID = "appMenu-whatsnew-button";
 const PANEL_HEADER_SELECTOR = "#PanelUI-whatsNew-title > label";
 
 const BUTTON_STRING_ID = "cfr-whatsnew-button";
+const WHATS_NEW_PANEL_SELECTOR = "PanelUI-whatsNew-message-container";
 
 class _ToolbarPanelHub {
   constructor() {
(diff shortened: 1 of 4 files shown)


lib/ToolbarPanelHub.jsm
```diff
@@ -190,6 +193,16 @@ class _ToolbarPanelHub {
     });
   }
 
+  removeMessages(win, containerId) {
+    const doc = win.document;
+    const messageNodes = doc
+      .getElementById(containerId)
+      .querySelectorAll(".whatsNew-message");
```
punamdahiya: nit: Might help to pass class selector '.whatsNew-message' here so that this method is truly generic as name suggest 'removeMessages' or consider renaming it to removeWhatsNewMessages, wdyt?

Mardak: Well, `renderMessages` also has hardcoded `.whatsNew-message` and this .jsm handling panel messages only supports whatsnew messages, so I guess this is fine for now. We might support non "whatsnew" messages in the panel later, but we still probably want to remove all messages (and maybe rename the classname at that point).

punamdahiya: Sounds good, lets get it in master, hitting merge

Issue #862, opened 2016-06-20
some keystrokes end up missing when I type in the search box


wlusink:
So when I type slowly in the search box, all is well.
However, when I type faster, some keystrokes are missing. I'm pretty sure this has to do with the search suggestions. Apparently the search suggestions change the text I'm typing, as I'm typing.

This only happens to me in the 'New Tab screen' that is presented to me by the Activity Stream 1.1.0 add-on.
I have no other add-ons active in FireFox.

I run windows 8.1 and firefox 47.0

If other info is needed, please ask.

ncloudioj:
Thanks for reporting this!
dup #848

wlusink:
ayyy sorry!

I did do a search but didn't find anything.

ahum.. must have been because my search box didn't work correctly hehehe 
;-)

thanks!

Nan Jiang schreef op 2016-06-20 09:24:

PR #5299, opened 2019-09-05
[fix snippets-service #862] Add SubmitForm error button label.

jpetto:
Fixes mozmeao/snippets-service#862

Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1510331


jpetto:
@piatra Added the corresponding key/value to the `NEWSLETTER_TEST_DEFAULTS` test snippet. However, I'm not sure how to force an error on submission, so I wasn't able to manually verify the `error_button_label` is showing (though the tests should be handling this).

Is there a way to force a form submit failure on a test snippet?

piatra:
@jpetto sorry there's no easy way to do it because the URLs are hardcoded. To test I made the snippet show and then using the browser devtools edit the `form url`. This would force it to fail.

piatra:
lgtm 

image

PR #5303, opened 2019-09-06
Bug 1579480 - newtab DS cards not removing idle callback

ScottDowne:
Hard to test, in general making sure all rows of cards appear is probably good enough for regression testing.

1. Set `browser.newtabpage.activity-stream.discoverystream.enabled` to true
2. Load a new tab.
3. wait a few seconds.
4. Scroll from top to bottom
Expected: cards should be already loaded as you scroll, or fade in.
Not Expected: If you see no cards at all, or if cards don't fade in/are not already loaded.

To test the actual bug, I've only reproduced it a few times.

1. Set `browser.newtabpage.activity-stream.discoverystream.enabled` to true
2. Open the page's console.
3. toggle `browser.newtabpage.activity-stream.discoverystream.enabled` a few times.
Expected: no console errors.
Not expected: console error for "Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method"

But for the second set of steps, it's hard to reproduce and possibly not a worry if you don't see the error without the patch.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
@@ -141,7 +141,7 @@ export class DSCard extends React.PureComponent {
   }
 
   componentDidMount() {
-    this.idleCallbackId = window.requestIdleCallback(
+    this.idleCallbackId = this.props.windowObj.requestIdleCallback(
       this.onIdleCallback.bind(this)
     );
     if (this.placholderElement) {
@@ -154,6 +154,9 @@ export class DSCard extends React.PureComponent {
     // Remove observer on unmount
     if (this.observer && this.placholderElement) {
       this.observer.unobserve(this.placholderElement);
+    }
+    if (this.idleCallbackId) {
+      this.props.windowObj.cancelIdleCallback(this.idleCallbackId);
     }
   }
 
@@ -234,4 +237,9 @@ export class DSCard extends React.PureComponent {
     );
   }
 }
+
+DSCard.defaultProps = {
+  windowObj: window, // Added to support unit tests
+};
+
 export const PlaceholderDSCard = props => <DSCard placeholder={true} />;
--- a/content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
@@ -76,7 +76,7 @@ export class DSCard extends React.PureComponent {
 
     this.onLinkClick = this.onLinkClick.bind(this);
     this.setPlaceholderRef = element => {
-      this.placholderElement = element;
+      this.placeholderElement = element;
     };
 
     this.state = {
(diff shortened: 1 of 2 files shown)


content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -155,6 +155,9 @@ export class DSCard extends React.PureComponent {
     if (this.observer && this.placholderElement) {
       this.observer.unobserve(this.placholderElement);
     }
+    if (this.idleCallbackId) {
```
gvn: Won't this always be true? Why are you immediately cancelling the idle callback?

gvn: Also, why does the idle callback need to be removed if it's only called once?

gvn:
Also, minor nit, and it's pre-existing, but I noticed that `placholderElement` is spelled wrong.

content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
ScottDowne: For removing the callback, a component can be mounted, and then unmounted, before the browser does an idle.

For the cases where the component gets unmounted before idle happens, it calls `onIdleCallback` after the component has been unmounted. At best this throws at `setState` at worst, this kinda smells like a memory leak. 

I was initially able to reproduce it but couldn't the next day, but I'm pretty it's there, just hard to reproduce.

For `idleCallbackId` always existing, yeah, it should. I think I'm just being careful here to cover potential refactors that may clear the id, or to cover something I may not be able to predict. Now looking at it, I kinda want to manually clear the id myself inside `onIdleCallback`

ScottDowne: What are your thoughts on for `idleCallbackId` always existing?

I'm happy to:
1. Remove the if check, like you initially suggested.
2. Leave it a is, after me explaining I'm just being careful and not actually trying to check something.
3. Clear the id so the if in `componentWillUnmount` can actually be avoided if we don't need to cancel it.

gvn: I actually wasn't looking at this right...I missed that this is part of the unmount method, so I actually think this is probably fine as is. Derp.

gvn: Once `onIdleCallback` is called I don't think you have to worry about cancelling the idle callback, because AFAIK they only get fired once per ID.

ScottDowne: Yeah, I did ensure that calling cancel idle on an id that has been fired, there was no errors, so figured I could leave it as is.

ScottDowne:
Managed to get better steps to test.

1. Set `browser.newtabpage.activity-stream.debug` to true
2. Open a new tab
3. Open about:config
4. Toggle browser.newtabpage.activity-stream.discoverystream.enabled on and off, a few times.
5. Check console.

Expected, no errors.

PR #5305, opened 2019-09-10
Bug 1559185 - Use existing default prefs for dynamic prefs too

Mardak:
r?@ScottDowne or @k88hudson Similar to #4531 except handle dynamic prefs to reuse existing defaults while still allowing a temporary default (for dummy geo).

You can test with the existing `firefox.js` to dynamically set on geo change for a new profile. Then also test with an explicit `false` pref (as well as `true`).

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -720,11 +720,25 @@ this.ActivityStream = class ActivityStream {
 
     // Update the pref config of those with dynamic values
     for (const pref of PREFS_CONFIG.keys()) {
+      // Only need to process dynamic prefs
       const prefConfig = PREFS_CONFIG.get(pref);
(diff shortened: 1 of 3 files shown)


Mardak:
Looks like mochitests aren't happy…

Mardak:
Locally passes with try/catch around getting default pref

Mardak:
I just copied over the existing comment for the try/catch:
https://github.com/mozilla/activity-stream/blob/3854bd503f9b471704978a64855386fdb044f833/lib/ActivityStreamPrefs.jsm#L72-L74

Seems strange to add unit test behavior for a bug although I suppose I'll file a bug on this…
https://bugzilla.mozilla.org/show_bug.cgi?id=1580610

PR #5306, opened 2019-09-11
Bug 1577655 - Dismiss wrapper design tweaks

gvn:
To test:

2. Update `browser.newtabpage.activity-stream.discoverystream.config` with `{"api_key_pref":"extensions.pocket.oAuthConsumerKey","collapsible":true,"enabled":true,"show_spocs":true,"hardcoded_layout":false,"personalized":false,"layout_endpoint":"https://5ad1b408-b40a-4a49-b6ec-27473d192df0.mock.pstmn.io/layout-1577654"} `

3. Verify requested changes [per ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1577655)

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
+++ b/content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
@@ -69,10 +69,9 @@ export class DSDismiss extends React.PureComponent {
           onClick={this.onDismissClick}
           onMouseEnter={this.onHover}
           onMouseLeave={this.offHover}
+          aria-label="dismiss"
         >
-          <span className="icon icon-dismiss">
-            <span className="sr-only">Dismiss</span>
-          </span>
+          <span className="icon icon-dismiss" />
         </button>
       </div>
     );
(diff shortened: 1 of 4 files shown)


content-src/components/DiscoveryStreamComponents/DSDismiss/DSDismiss.jsx
```diff
@@ -70,7 +70,9 @@ export class DSDismiss extends React.PureComponent {
           onMouseEnter={this.onHover}
           onMouseLeave={this.offHover}
         >
-          <span className="icon icon-dismiss" />
+          <span className="icon icon-dismiss">
+            <span className="sr-only">Dismiss</span>
```
ScottDowne: What does sr-only mean?

gvn: I added a utility class+mixin (used Bootstrap's approach) for hiding text content offscreen so that only screenreaders can pick it up.

Mardak: This seems to be re-adding an anti-pattern that was explicitly removed for accessibility reasons https://github.com/mozilla/activity-stream/commit/b2a197b8c4a772d4642f5f1b08a0a40a5b1d16a3

ScottDowne: What does sr-only mean?

ScottDowne: Yeah I don't think I knew the whole story here, but if sr-only is an issue we should figure out how to get this done the right way.

PR #5309, opened 2019-09-16
Bug 1578754 - Consider implementing "groups" for messages

piatra:
Use the `dev.kinto` server
```
Services.prefs.setStringPref("services.settings.server", "https://kinto.dev.mozaws.net/v1");

// Disable signature verification
const { RemoteSettings } = ChromeUtils.import("resource://services-settings/remote-settings.js", {});

RemoteSettings("message-groups").verifySignature = false;
```

Firefox is a little slow to switch. It helps to restart the browser and also run 
```
RemoteSettings("message-groups").get().then(console.log)
```
to confirm that the message is fetched.
If it is fetched it should show up in ASRAdmin
image
And in the new Message Groups section in ASRAdmin
image
Add it to one of the existing messages from `PanelTestProvider`
image
* Toggle off the pref associated with the group.
* Notice the group shows up as disabled in ASRAdmin
* Notice that the message is filtered out


lib/ASRouter.jsm
```diff
@@ -1487,17 +1559,30 @@ class _ASRouter {
...
+            time
+          );
+        }
+        return { messageImpressions, groupImpressions };
       });
     }
   }
 
+  // Easier to get all impressions from
+  _addImpressionForGroups(groupImpressions, groupId, time) {
```
piatra: I think this can be refactored to use `_addImpressionForItem` instead.

lib/ASRouter.jsm
```diff
@@ -2081,9 +2172,11 @@ class _ASRouter {
         QueryCache.expireAll();
         break;
       case "ENABLE_PROVIDER":
+        await this.unblockProviderById(action.data);
```
piatra: I added these in in order to have a way to clear provider impressions through ASRouter admin. It would not cover the cases where there are impressions from a group that is different from an existing provider though.

lib/ASRouter.jsm
```diff
@@ -2095,6 +2185,9 @@ class _ASRouter {
           action.data.value
         );
         break;
+      case "SET_GROUP_STATE":
```
k88hudson: Can we also add a user action to block a group? It should apply to future messages in the group as well.

piatra:
@k88hudson I was trying to decide what to do about `"categories":["cfrAddons","cfrFeatures"]`
The current behavior is that `providers.cfr` is enabled if any of the two are `enabled`.
This doesn't really map well to the concept for groups, it should instead be two different groups:
```
{
  id: "cfr-addons-recommendations",
  userPreferences: ["cfrAddons"]
}
{
  id: "cfr-features-recommendations",
  userPreferences: ["cfrFeatures"]
}
```
Because the groups implementation of `userPreferences` is that every pref should be enabled.

k88hudson:
> @k88hudson I was trying to decide what to do about `"categories":["cfrAddons","cfrFeatures"]`
> The current behavior is that `providers.cfr` is enabled if any of the two are `enabled`.
> This doesn't really map well to the concept for groups, it should instead be two different groups:
> 
> ```
> {
>   id: "cfr-addons-recommendations",
>   userPreferences: ["cfrAddons"]
> }
> {
>   id: "cfr-features-recommendations",
>   userPreferences: ["cfrFeatures"]
> }
> ```
> 
> Because the groups implementation of `userPreferences` is that every pref should be enabled.

I do think it makes sense to have each of these be separate groups; it's not necessary to track impressions really, but applying the user pref that way makes sense.

lib/ASRouter.jsm
```diff
@@ -1480,10 +1562,12 @@ class _ASRouter {
   }
 
   async addImpression(message) {
-    const provider = this.state.providers.find(p => p.id === message.provider);
+    const groupsWithFrequency = this.state.groups.filter(
```
k88hudson: This is good because it means we can use groups just for prefs if we want (like in the case of CFR features v.s. addons) without adding any extra unnecessary impression data.

lib/ASRouter.jsm
```diff
@@ -664,6 +670,61 @@ class _ASRouter {
     });
   }
 
+  hasGroupsEnabled(groups = []) {
+    return this.state.groups
+      .filter(({ id }) => groups.includes(id))
+      .every(({ enabled }) => enabled);
```
k88hudson: Do we want this to return `true` if a message is part of no groups (I guess that's not really possible... but maybe if there's a failure case. Probably more efficient to do `every({ id, enabled }) => !groups.includes(id) || enabled`

piatra: Using `!groups.includes(id)` predicate for `.every` We‘ll also cover the case for `[]`.

lib/ASRouter.jsm
```diff
@@ -696,11 +760,8 @@ class _ASRouter {
             storage: this._storage,
             dispatchToAS: this.dispatchToAS,
           });
-          messages = messages.filter(
-            ({ content }) =>
-              !content ||
-              !content.category ||
-              ASRouterPreferences.getUserPreference(content.category)
+          messages = messages.filter(({ groups }) =>
+            this.hasGroupsEnabled(groups)
```
piatra: The consequence of this is that a message that is part of a disabled group will be filtered out. This might be confusing when using ASRAdmin to debug messages.

content-src/components/ASRouterAdmin/ASRouterAdmin.jsx
```diff
@@ -1142,6 +1159,35 @@ export class ASRouterAdminInner extends React.PureComponent {
...
+                  <td>Enabled</td>
+                  <td>Impressions count</td>
+                  <td>Custom frequency</td>
+                </tr>
+              </thead>
+              {this.state.groups &&
+                this.state.groups.map(({ id, enabled, frequency }, index) => (
+                  <Row key={id}>
+                    <td>
+                      <TogglePrefCheckbox
```
piatra: The checkbox is meant to show if the groups is enabled and disabled. Trying to toggle it off/on will not change the `disabled` state of the group.
I've considered adding `disabled` attribute to the checkbox but it is also used as a way to clear impressions right now.
Perhaps it can be improved with additional UI (follow up).

lib/ASRouter.jsm
piatra: Don't filter out here, do it in getUnblockedMessages so we keep the messages around in ASRAdmin and we can explain the state.

piatra: Move the `exclude` provider configuration in there as well.

lib/ASRouter.jsm
```diff
@@ -674,6 +680,61 @@ class _ASRouter {
...
+
+  /**
+   * Fetch all message groups and update Router.state.groups
+   * @param provider RS messages provider for message groups
+   */
+  async loadAllMessageGroups(provider) {
+    if (!provider) {
+      return;
+    }
+    let { messages } = await MessageLoaderUtils.loadMessagesForProvider(
```
k88hudson: It's a bit confusing that you're loading groups via `loadMessagesForProvider` I would probably have a generic function to like load data i.e. either messages or groups from the provider, and then have anything specific to messages be handled separately

lib/ASRouter.jsm
```diff
@@ -1569,12 +1656,17 @@ class _ASRouter {
         state.messages,
         "messageImpressions"
       );
-      const providerImpressions = this._cleanupImpressionsForItems(
+      const groupImpressions = this._cleanupImpressionsForItems(
+        state,
+        state.providers,
```
k88hudson: this is supposed to be groups, not providers here

piatra:
Moved to https://phabricator.services.mozilla.com/D56545

PR #5310, opened 2019-09-16
Bug 1581570 - Spocs placement bug impacting non placements fixed

ScottDowne:
Steps to test in bugzilla

### DIFF
--- a/common/Reducers.jsm
+++ b/common/Reducers.jsm
@@ -526,7 +526,7 @@ function DiscoveryStream(prevState = INITIAL_STATE.DiscoveryStream, action) {
     const { data, placements } = prevState.spocs;
     const result = {};
 
-    placements.forEach(placement => {
+    const forPlacement = placement => {
       const placementSpocs = data[placement.name];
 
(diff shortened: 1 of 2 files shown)


ScottDowne:
So right now we have this concept of "if no placements assume spocs as the only placement"

It was the cause of this bug because I implemented the fallback in the usage of placements inside DS Feed and not actually defaulting the placements inside the reducer.

This can be moved into the reducer's `DISCOVERY_STREAM_SPOCS_PLACEMENTS` in the future, but I hesitated to do that because doing it the way I have it in the patch is easier to uplift because it impacts less.

gvn:
Failing Taskcluster

gvn:
When I follow STR I get a new spoc where the old one was dismissed after I refresh. Is that expected?

ScottDowne:
Yeah, that's expected. If you refresh after dismiss, it does a fresh load including spoc placement.

We only keep the spoc out of that spot for that specific tab instance.

ScottDowne:
@gvn should be ready for another look.

PR #5311, opened 2019-09-16
Bug 1581209 - enable gzip compression for Structure Ingestion payload

ncloudioj:
Test instructions:

* Flipping `browser.newtabpage.activity-stream.telemetry` and `browser.newtabpage.activity-stream.telemetry.structuredIngestion`
* Open Browser Toolbox from the Tools menu and switch to the `Network` tab
* Open a newtab, and scroll down until you see some Pocket stories, and then close the tab
* Check the `Domain` column, and look for `incoming.telemetry.mozilla.org`
* Click on any entry, make sure the `Requeset headers` from the `Headers` section in the side view contains `"Content-Encoding": gzip`, and the value in the `Params` section is in binary format

### DIFF
--- a/ping-centre/PingCentre.jsm
+++ b/ping-centre/PingCentre.jsm
@@ -365,8 +365,41 @@ class PingCentre {
       });
   }
 
+  static _gzipCompressString(string) {
+    let observer = {
+      buffer: "",
+      onStreamComplete(loader, context, status, length, result) {
+        this.buffer += String.fromCharCode(...result);
+      },
+    };
+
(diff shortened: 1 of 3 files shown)


ncloudioj:
Turns out the Fetch API couldn't handle the gzip payload properly, I had to switch back to the good old XMLHttpRequest (a.k.a. ServiceRequest).

The compression ratio looks pretty promising with our telemetry payloads, I'm getting a 50% reduction on average.

@ScottDowne mind taking another look?

ScottDowne:
Still looks good to me. Once taskcluster/travis is green, you get another r+.

PR #5314, opened 2019-09-17
Fix Bug 1582002 - make sync optional in the about:welcome signin flow

vladikoff:
Requires https://github.com/mozilla/fxa/pull/2534 in production

This is for 71 at some point

### DIFF
--- a/content-src/asrouter/components/FxASignupForm/FxASignupForm.jsx
+++ b/content-src/asrouter/components/FxASignupForm/FxASignupForm.jsx
@@ -90,7 +90,6 @@ export class FxASignupForm extends React.PureComponent {
           rel="noopener noreferrer"
           onSubmit={this.onSubmit}
         >
-          <input name="service" type="hidden" value="sync" />
           <input name="action" type="hidden" value="email" />
           <input name="context" type="hidden" value="fx_desktop_v3" />
           <input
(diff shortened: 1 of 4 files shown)


content-src/asrouter/templates/Trailhead/Trailhead.jsx
```diff
@@ -179,7 +179,6 @@ export class Trailhead extends React.PureComponent {
               rel="noopener noreferrer"
               onSubmit={this.onSubmit}
             >
-              <input name="service" type="hidden" value="sync" />
```
vladikoff: I'm not sure if the Trailhead template is affected

vladikoff:
@dmose @k88hudson could someone look at this? 

* The FxA server and Firefox Desktop for this has landed
* We would like this to ride the trains with Firefox Desktop (making sure this patch **does NOT** get uplifted with some other changes, it requires something that just landed in Nightly)

vladikoff:
to test this:

* open about:welcome
* fill out the FxA email, you should get to `https://accounts.firefox.com/signup?action=email&context=fx_desktop_v3&......`

(notice no `service=sync` parameter in the URL)

* If it is a new account you will be presented with a new screen:

image

* if you use the new "Do not sync" option, the about:prefences now would reflect that:

image

vladikoff:
this was rebased

vladikoff:
Thank you!

PR #5316, opened 2019-09-18
Bug 1581618 - Remove old about:welcome control variant

k88hudson:
Removes the "blue" old onboarding experience and related assets. I think I got all the strings that weren't used elsewhere, but might be worth double checking.

I removed the targeting expression from the `"join"` interrupt so that it would become the default experience.


lib/ASRouter.jsm
```diff
@@ -1039,8 +1039,9 @@ class _ASRouter {
         );
       }
     } else {
-      // If the user is not in a trailhead-compabtible locale, return the control experience and no experiment.
-      interrupt = "control";
+      // If the user is not in a trailhead-compabtible locale, return the join + supercharge (default) experience and no experiment.
```
Mardak: Just noting for others reviewing, this "locale-specific" behavior/code is turned off because `trailhead.firstrun.branches` has a default value.

punamdahiya:
Need to remove StartupOverlay.jsx and its containing folder, thatswhy travis failures

content-src/asrouter/templates/StartupOverlay/_StartupOverlay.scss
```diff
@@ -1,343 +0,0 @@
...
-.firstrun-link {
-  opacity: 0;
-  transform: translateY(-5px);
-}
-
-.error {
-  display: none;
-}
-
-.error.active {
```
punamdahiya: Looks like we are reusing these styles in trailhead interrupts. Invalid email error shows incorrect without these.

lib/ASRouter.jsm
```diff
@@ -995,11 +995,11 @@ class _ASRouter {
       [interrupt, triplet] = overrideValue.split("-");
     }
 
-    // Use control Trailhead Branch (for cards) if we are showing RTAMO.
+    // Use join Trailhead Branch (for cards) if we are showing RTAMO.
     if (await this._hasAddonAttributionData()) {
       return {
         experiment,
-        interrupt: "control",
+        interrupt: "join",
```
punamdahiya: how can we test this?

test/unit/asrouter/ASRouterTargeting.test.js
```diff
@@ -80,29 +80,10 @@ describe("#CachedTargetingGetter", () => {
...
-    assert.equal(stub.callCount, 6);
+    assert.equal(stub.callCount, 5);
     const calls = stub.getCalls().map(call => call.args[0]);
     const lastCall = calls[calls.length - 1];
-    assert.equal(lastCall.id, "FXA_1");
-  });
-  it("should return FxA message (is fallback)", async () => {
-    const messages = (await OnboardingMessageProvider.getUntranslatedMessages()).filter(
-      m => m.id !== "RETURN_TO_AMO_1"
-    );
```
punamdahiya: makes sense to remove, isn't this test same as above, wonder why we were filtering out RTAMO messages here?

punamdahiya:
Looks good, with test fixes and invalid email styles in place looks good to merge. Thanks!

punamdahiya:
Patch looks good. Should we hold off merging to master till https://phabricator.services.mozilla.com/D46677 lands which should help fix tests breaking due to removing chrome://branding/content/about-logo.png reference.

PR #5250, opened 2019-08-13
Bug 1566653 - Topsites promo back end.

ScottDowne:

PR #5318, opened 2019-09-19
Bug 1576680 - Telemetry additions for SPOC Topsites

ncloudioj:
This adds a new card_type `spoc` to identify a SPOC top site. When one clicks or dismisses a SPOC top site, we will include this value into the telemetry ping so that click-through rate and dismissal rate could be calculated accordingly.

Note that the icon type `custom-screenshot` is not a new addition but a document miss when we implemented the custom top site feature, added it back in for the completeness.

This needs a data review.

@punamdahiya Scott is on PTO now, could you take a look at this one?

### DIFF
--- a/content-src/components/TopSites/TopSite.jsx
+++ b/content-src/components/TopSites/TopSite.jsx
@@ -316,6 +316,9 @@ export class TopSite extends React.PureComponent {
       value.card_type = "search";
       value.search_vendor = this.props.link.hostname;
     }
+    if (this.props.link.type === SPOC_TYPE) {
+      value.card_type = "spoc";
+    }
     return { value };
   }
 
--- a/docs/v2-system-addon/data_dictionary.md
+++ b/docs/v2-system-addon/data_dictionary.md
@@ -208,7 +208,7 @@ Schema definitions/validations that can be used for tests can be found in `syste
 | `action` | [Required] Either `activity_stream_event`, `activity_stream_session`, or `activity_stream_performance`. | :one:
 | `addon_version` | [Required] Firefox build ID, i.e. `Services.appinfo.appBuildID`. | :one:
 | `client_id` | [Required] An identifier for this client. | :one:
-| `card_type` | [Optional] ("bookmark", "pocket", "trending", "pinned", "search") | :one:
+| `card_type` | [Optional] ("bookmark", "pocket", "trending", "pinned", "search", "spoc") | :one:
 | `search_vendor` | [Optional] the vendor of the search shortcut, one of ("google", "amazon", "wikipedia", "duckduckgo", "bing", etc.). This field only exists when `card_type = "search"` | :one:
 | `date` | [Auto populated by Onyx] The date in YYYY-MM-DD format. | :three:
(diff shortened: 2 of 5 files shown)


punamdahiya:
Looks good! Verified by using QA steps from https://github.com/mozilla/activity-stream/pull/5250#issue-307018371 and logging event_object [here in console](https://searchfox.org/mozilla-central/source/browser/components/newtab/lib/TelemetryFeed.jsm#620). Will be good to add steps in the bug for QA to verify pings in about:telemetry

Click of spocs link menu option (Pin, open windows, dismiss) and click on spoc topsite is successfully sending user event ping with value as 
card_type: "spoc" and icon_type: "custom_screenshot".


For click and dismiss event of spoc topsites we are additionally capturing impression stats, not sure if we need to do for pin spoc topsites. If needed this can be [done here ](https://searchfox.org/mozilla-central/source/browser/components/newtab/content-src/lib/link-menu-options.js#174) similar to blockurl.

It will be good to note [may be here](https://searchfox.org/mozilla-central/source/browser/components/newtab/content-src/components/TopSites/TopSites.jsx#41) that we are assigning icon_type as custom_screenshot for spocs.  Ideally we  should specifiy a new icon_type 'spoc' so that custom screen shot can be correctly counted. This can be taken up as a separate fix. Thanks!

ncloudioj:
> Looks good! Verified by using QA steps from [#5250 (comment)](https://github.com/mozilla/activity-stream/pull/5250#issue-307018371) and logging event_object [here in console](https://searchfox.org/mozilla-central/source/browser/components/newtab/lib/TelemetryFeed.jsm#620). Will be good to add steps in the bug for QA to verify pings in about:telemetry
> 

Thanks for verifying it. I was unable to test this feature by following the QA steps commented in #5250. Will try it again.

> For click and dismiss event of spoc topsites we are additionally capturing impression stats, not sure if we should do same for pin spoc topsites. If needed this can be [done here ](https://searchfox.org/mozilla-central/source/browser/components/newtab/content-src/lib/link-menu-options.js#174) similar to blockurl.

> It will be good to note [may be here](https://searchfox.org/mozilla-central/source/browser/components/newtab/content-src/components/TopSites/TopSites.jsx#41) that we are assigning icon_type as custom_screenshot for spocs. Ideally we should specifiy a new icon_type 'spoc' so that custom screen shot can be correctly counted. This can be taken up as a separate fix. Thanks!

Those are all good points, and I did notice them in the code comments of #5250 

Given that we would mostly focus on the click-through & dismissal rates as well as the total volume of traffic that SPOC top sites could generate, I left them out somewhat intentionally to keep this patch as scoped as possible. After all, pinning & unpinning is much less used by the users in comparison with click & dismiss.

As for the `icon_type`, we use it to measure the icon coverage for Tippytop and rich icons, so `custom_screenshot` won't affect those counters at all, though we should definitely assign a new value to it if we were to ship this feature to release. 

@punamdahiya Thanks again for the review!

Will merge it once our data steward finishes the data review on Bugzilla.

ncloudioj:
Data review granted on Bugzilla.

PR #5321, opened 2019-09-23
Bug 1569696 - CFR: recommend send tab on popular recipe sites

piatra:


lib/ASRouterTriggerListeners.jsm
```diff
@@ -429,14 +487,7 @@ this.ASRouterTriggerListeners = new Map([
...
           );
-
-          for (let win of Services.wm.getEnumerator("navigator:browser")) {
-            if (isPrivateWindow(win)) {
-              continue;
-            }
-            win.gBrowser.removeTabsProgressListener(this);
-          }
-
+          EveryWindow.unregisterCallback(this.id);
```
piatra: This fixes an issue from my recent `EveryWindow` refactoring, the mochitest associated with this trigger would fail with `--verify`.

lib/PanelTestProvider.jsm
```diff
@@ -240,6 +240,110 @@ const MESSAGES = () => [
       patterns: ["*://*/*.pdf"],
     },
   },
+  {
+    id: "SEND_TAB_CFR",
```
k88hudson: Maybe something about recipes specifically since there's likely to be multiple variants of this? I guess this is for testing anyway so it doesn't really matter

lib/ASRouterTriggerListeners.jsm
```diff
@@ -77,6 +77,64 @@ function checkURLMatch(aLocationURI, { hosts, matchPatternSet }, aRequest) {
...
+    "openArticleURL",
+    {
+      id: "openArticleURL",
+      _initialized: false,
+      _triggerHandler: null,
+      _hosts: new Set(),
+      _matchPatternSet: null,
+      readerModeEvent: "Reader:UpdateReaderButton",
+
+      init(triggerHandler, hosts, patterns) {
```
k88hudson: A lot of this logic is looking pretty similar to other triggers, I wonder if we want another abstraction on top of it at some point. For now this is fine though.

lib/ASRouterTriggerListeners.jsm
```diff
@@ -77,6 +77,64 @@ function checkURLMatch(aLocationURI, { hosts, matchPatternSet }, aRequest) {
...
 this.ASRouterTriggerListeners = new Map([
+  [
+    "openArticleURL",
+    {
+      id: "openArticleURL",
+      _initialized: false,
+      _triggerHandler: null,
+      _hosts: new Set(),
+      _matchPatternSet: null,
+      readerModeEvent: "Reader:UpdateReaderButton",
```
k88hudson: Will this also trigger when the reader button changes to be inactive? I guess that's accounted for by checking `data.isArticle`?

lib/ASRouter.jsm
```diff
@@ -1879,6 +1879,15 @@ class _ASRouter {
       case ra.OPEN_APPLICATIONS_MENU:
         UITour.showMenu(target.browser.ownerGlobal, action.data.args);
         break;
+      case ra.HIGHLIGHT_FEATURE:
+        const highlight = await UITour.getTarget(
```
k88hudson: Whoa this is useful... nice find

k88hudson:
Looks good, that highlight feature is going to be useful I think. One thing I am noticing from manual testing is that the menu with Send Tab to Device seems to stick instead of hiding after clicking outside the bounds of the doorhanger; is that an intentional design choice I wonder? It seems a bit annoying if you decide not to try it...

piatra:
@k88hudson agreed it is annoying. [This seems intentional](https://searchfox.org/mozilla-central/rev/23f836a71cfe961373c8bd0d0219ec60a64b3c8f/layout/xul/nsMenuPopupFrame.cpp#212-214), (the attribute [is set in showMenu](https://searchfox.org/mozilla-central/rev/23f836a71cfe961373c8bd0d0219ec60a64b3c8f/browser/components/uitour/UITour.jsm#1486)) probably makes sense in the UITour context where you show some features and then also hide the panel as part of the tour.

I worked around it by passing an `options` argument from the `ASRouter.showHighlight` call to `showMenu` in `UITour.jsm`. There's a phabricator patch with the changes to UITour https://phabricator.services.mozilla.com/D47777

I had to do this because after the panel widget is shown ([the `noautohide` attribute is set before `show` is called](https://searchfox.org/mozilla-central/rev/23f836a71cfe961373c8bd0d0219ec60a64b3c8f/browser/components/uitour/UITour.jsm#1486-1497)) changes to `noautohide` value no longer influence the behavior of the panel (and this happens ["by design"](https://searchfox.org/mozilla-central/rev/23f836a71cfe961373c8bd0d0219ec60a64b3c8f/browser/components/uitour/UITour.jsm#1646-1658)).

lib/ASRouterTriggerListeners.jsm
piatra: Yes the event triggers on every page, even multiple times (if you have sub frames for example) and it could be `false` every time.

lib/ASRouter.jsm
```diff
@@ -1880,6 +1880,21 @@ class _ASRouter {
...
+          target.browser.ownerGlobal,
+          action.data.args
+        );
+        if (highlight) {
+          await UITour.showHighlight(
+            target.browser.ownerGlobal,
+            highlight,
+            // aEffect
```
k88hudson: Forgot to remove this or intentional?

piatra: Intentional. I thought „none“ as a string was kind of random, it’s just the argument name as a comment.

Mardak:
Added extra commit to uninit the test listener to avoid linux64 debug leak https://github.com/mozilla/activity-stream/commit/da1c2f184d9f2e450aa35d428f993f9a9bfc15dc

PR #5322, opened 2019-09-24
Bug 1582713 - When What's New toolbar button is active the Overflow M…

piatra:
…enu position is changed in Customize Firefox

Add listeners for `customization(ready|ending)` to show/hide the What's New button that otherwise breaks the Customize view.


gijsk:
Instead of this, can we use CSS to display: none the button while the window/toolbar has a `[customizing]` attribute? That might be simpler...

piatra:
https://phabricator.services.mozilla.com/D46964

PR #5323, opened 2019-09-24
Bug 1582242 - Modal should not close when clicked outside the dialog

punamdahiya:


content-src/asrouter/components/ModalOverlay/ModalOverlay.jsx
```diff
@@ -51,7 +51,6 @@ export class ModalOverlayWrapper extends React.PureComponent {
       <React.Fragment>
         <div
           className="modalOverlayOuter active"
-          onClick={props.onClose}
```
k88hudson: Can you add a comment that mentions "The intended behaviour is to listen for an escape key but not for a click; see Bug 1582242"

punamdahiya: done

punamdahiya:
Patch updated by adding dismiss icon in modal

modal-with-dismiss

PR #5324, opened 2019-09-26
Bug 1550720 - Expose FxA services associated with profile to targetin…

piatra:
…g expression

Here are some example expressions to validate that the current data format is usable in JEXL.

* Number of times a certain service has been used
    * `listAttachedOAuthClients[.name == "Pocket"]|length`
* Uses a certain service
    * `"Pocket" in listAttachedOAuthClients|mapToProperty("name")`
* Services used in the past week
    * `listAttachedOAuthClients[.lastAccessTime >= (currentDate|date - 3600 * 24 * 7 * 1000)]|mapToProperty("name")`

### DIFF
--- a/content-src/asrouter/docs/targeting-attributes.md
+++ b/content-src/asrouter/docs/targeting-attributes.md
@@ -40,7 +40,7 @@ Please note that some targeting attributes require stricter controls on the tele
 * [totalBlockedCount](#totalblockedcount)
 * [recentBookmarks](#recentbookmarks)
 * [userPrefs](#userprefs)
-* [listAttachedOAuthClients](#listattachedoauthclients)
+* [attachedFxAOAuthClients](#attachedfxaoauthclients)
 
 ## Detailed usage
 
@@ -557,7 +557,7 @@ declare const userPrefs: {
 }
 ```
 
-### `listAttachedOAuthClients`
+### `attachedFxAOAuthClients`
 
 Information about connected services associated with the FxA Account.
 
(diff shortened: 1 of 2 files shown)


piatra:
I ran into an issue on my main profile where the values returned by the FxA endpoint showed multiple entries. I created [bug 1547120](https://bugzilla.mozilla.org/show_bug.cgi?id=1547120) but that shouldn't block this.

glogiotatidis:
Will we be able to query if a specific service has been used in the past week?

piatra:
> Will we be able to query if a specific service has been used in the past week?

Yes. That last JEXL example retrieves the list of all services used in the past week, we can do `"service name" in list` to check for something in particular.

glogiotatidis:
> > Will we be able to query if a specific service has been used in the past week?
> 
> Yes. That last JEXL example retrieves the list of all services used in the past week, we can do `"service name" in list` to check for something in particular.

Got it, thanks! Do we plan to have this in Fx71?

content-src/asrouter/docs/targeting-attributes.md
```diff
@@ -41,6 +41,7 @@ Please note that some targeting attributes require stricter controls on the tele
 * [totalBlockedCount](#totalblockedcount)
 * [recentBookmarks](#recentbookmarks)
 * [userPrefs](#userprefs)
+* [listAttachedOAuthClients](#listattachedoauthclients)
```
k88hudson: Personally I would have the word FXA in there somewhere but up to you

content-src/asrouter/docs/targeting-attributes.md
```diff
@@ -566,3 +567,20 @@ declare const userPrefs: {
   snippets: boolean;
 }
 ```
+
+### `listAttachedOAuthClients`
+
+Information about connected services associated with the FxA Account.
```
k88hudson: Might be worth adding an example of available clients (or actually add them to the type info, since users will need that)

PR #5325, opened 2019-09-26
Bug 1578575 - Ad Privacy Info

gvn:
Test steps: See https://bugzilla.mozilla.org/show_bug.cgi?id=1578575#c0


gvn:
@ncloudioj Looking into the Travis failure, but not a blocker to start review. Thanks!

flodolo:
Who can see sponsored stories? In other words, should this content be localized at all?

gvn:
> Who can see sponsored stories? In other words, should this content be localized at all?

Currently only English locales are getting the new-new tab experience, but eventually more geo/locale combos will be seeing it.

locales-src/newtab.ftl
```diff
@@ -83,6 +83,13 @@ newtab-menu-delete-history = Delete from History
 newtab-menu-save-to-pocket = Save to { -pocket-brand-name }
 newtab-menu-delete-pocket = Delete from { -pocket-brand-name }
 newtab-menu-archive-pocket = Archive in { -pocket-brand-name }
+newtab-menu-show-privacy-info = Our sponsors & your privacy
+
+## newtab privacy info modal
```
flodolo: Leave an empty line after a section comment. Also, these comments are supposed to be read by localizers, can it be a bit more explanatory?

locales-src/newtab.ftl
```diff
@@ -83,6 +83,13 @@ newtab-menu-delete-history = Delete from History
...
+newtab-privacy-modal-done = Done
```
flodolo: How is this used? Is it a button, a link? It would be good to have the role in the message ID.

locales-src/newtab.ftl
```diff
@@ -83,6 +83,13 @@ newtab-menu-delete-history = Delete from History
...
+newtab-privacy-modal-header = Your privacy matters.
+newtab-privacy-modal-paragraph = In addition to dishing up captivating stories, we also show you relevant, highly-vetted content from select sponsors. Rest assured, your browsing data never leaves your personal copy of Firefox—we don’t see it, and our sponsors don’t either.
```
flodolo: You can use multiline strings (indent with spaces) when the text gets this long.

```
 newtab-privacy-modal-paragraph =     
    In addition to dishing up captivating stories, we also show you relevant,
    highly-vetted content from select sponsors. Rest assured, your browsing
    data never leaves your personal copy of Firefox—we don’t see it, and our
    sponsors don’t either.
```

Never hard-code Firefox in strings, there are brand terms that can be used instead (`{ -brand-product-name }`).

The lack of spaces around the hyphen doesn't look correct?
https://design.firefox.com/photon/copy/punctuation.html

locales-src/newtab.ftl
```diff
@@ -83,6 +83,13 @@ newtab-menu-delete-history = Delete from History
...
+newtab-privacy-modal-link = Learn how privacy works on the new tab
```
flodolo: Reset the section comment after the last string by adding a `##`.

locales-src/newtab.ftl
gvn: Can you give me an example of what you mean?

locales-src/newtab.ftl
gvn: GTK. Will use a spaced em dash per style guide.

locales-src/newtab.ftl
gvn: Like this?

```
newtab-privacy-modal-link = Learn how privacy works on the new tab

##

```

flodolo: Exactly 👍

locales-src/newtab.ftl
gvn: Ah, ok, thanks!

gvn:
@flodolo I believe I've addressed all your feedback. Thanks for the review!

gvn:
@flodolo Yup, gotcha! I have Nan reviewing as well. I think you were automatically flagged due to being a code owner.

PR #5326, opened 2019-09-27
Bug 1576510 - Use aria-labelledby to add additional description information to badged toolbar buttons

piatra:
This is one idea for fixing the notification description, posting this for discussion.

I've used the OSX VoiceOver software to test this. The screenshot shows the button description as "What's New New feature": a combination of the button's own description and the element referred to via `aria-labelledby`.

Screenshot 2019-09-27 at 13 05 35

The idea is to use the hidden elements associated with any `toolbarbutton` [that has a badge](https://searchfox.org/mozilla-central/rev/f1e99da78fe6c3c68696358dac06aed90f8112d3/toolkit/content/widgets/toolbarbutton.js#110) to hold the button description.
[One issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1576510#c1) described in the bug comment was related to the string needed to describe the notification. I think we have more flexibility here because we can have a set of strings describing various types of badges: new feature, feature callout, hints etc. We can reference any one of these strings in the badge payload and set that description to the `toolbarbutton`.

### DIFF
--- a/content-src/asrouter/templates/OnboardingMessage/ToolbarBadgeMessage.schema.json
+++ b/content-src/asrouter/templates/OnboardingMessage/ToolbarBadgeMessage.schema.json
@@ -22,8 +22,9 @@
       "type": "number",
       "description": "Optional delay in ms after which to show the notification"
     },
-    "aria-label": {
+    "badgeDescription": {
       "type": "object",
+      "description": "This is used in combination with the badged button to offer a text based alternative to the visual badging. Example 'New Feature: What's New'",
       "properties": {
         "string_id": {
           "type": "string",
--- a/content-src/asrouter/templates/OnboardingMessage/ToolbarBadgeMessage.schema.json
+++ b/content-src/asrouter/templates/OnboardingMessage/ToolbarBadgeMessage.schema.json
@@ -1,7 +1,7 @@
 {
   "title": "ToolbarBadgeMessage",
   "description": "A template that specifies to which element in the browser toolbar to add a notification.",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "type": "object",
   "properties": {
     "target": {
@@ -21,6 +21,16 @@
     "delay": {
       "type": "number",
       "description": "Optional delay in ms after which to show the notification"
+    },
+    "aria-label": {
+      "type": "object",
+      "properties": {
+        "string_id": {
+          "type": "string",
+          "description": "Fluent string id"
+        }
+      },
+      "required": ["string_id"]
     }
   },
   "additionalProperties": false,
--- a/lib/ToolbarBadgeHub.jsm
+++ b/lib/ToolbarBadgeHub.jsm
@@ -3,56 +3,24 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
-ChromeUtils.defineModuleGetter(
-  this,
-  "EveryWindow",
-  "resource:///modules/EveryWindow.jsm"
+const { XPCOMUtils } = ChromeUtils.import(
+  "resource://gre/modules/XPCOMUtils.jsm"
 );
-ChromeUtils.defineModuleGetter(
-  this,
-  "ToolbarPanelHub",
-  "resource://activity-stream/lib/ToolbarPanelHub.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "Services",
-  "resource://gre/modules/Services.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "setTimeout",
-  "resource://gre/modules/Timer.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "clearTimeout",
-  "resource://gre/modules/Timer.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "Services",
-  "resource://gre/modules/Services.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "PrivateBrowsingUtils",
-  "resource://gre/modules/PrivateBrowsingUtils.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "setInterval",
-  "resource://gre/modules/Timer.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "clearInterval",
-  "resource://gre/modules/Timer.jsm"
-);
-ChromeUtils.defineModuleGetter(
-  this,
-  "requestIdleCallback",
-  "resource://gre/modules/Timer.jsm"
-);
+
+XPCOMUtils.defineLazyModuleGetters(this, {
+  EveryWindow: "resource:///modules/EveryWindow.jsm",
+  ToolbarPanelHub: "resource://activity-stream/lib/ToolbarPanelHub.jsm",
+  Services: "resource://gre/modules/Services.jsm",
+  PrivateBrowsingUtils: "resource://gre/modules/PrivateBrowsingUtils.jsm",
+});
+
+const {
+  setInterval,
+  clearInterval,
+  requestIdleCallback,
+  setTimeout,
+  clearTimeout,
+} = ChromeUtils.import("resource://gre/modules/Timer.jsm");
 
 // Frequency at which to check for new messages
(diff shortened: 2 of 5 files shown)


piatra:
@Mardak, what do you think?

lib/ToolbarBadgeHub.jsm
```diff
@@ -243,10 +248,26 @@ class _ToolbarBadgeHub {
...
-        .querySelector(".toolbarbutton-badge")
-        .classList.add("feature-callout");
+      // If we have additional aria-label information for the notification
+      // we add this content to the hidden `toolbarbutton-text` node.
+      // We then use `aria-labelledby` to link this description to the button
+      // that received the notification badge.
+      if (message.content["aria-label"]) {
+        toolbarbutton.setAttribute(
+          "aria-labelledby",
+          "toolbarbutton-notification-description"
```
jcsteh: As I understand it, #toolbarbutton-notification-description will just contain "New feature", not "What's New". Thus, I would've thought setting aria-labelledby like this would override "What's new" completely. However, you noted you'd tested this with VoiceOver and you saw both? That's good, but I'm a little concerned because I don't understand *how* that's working.

Does the Accessibility Inspector in the Browser Toolbox also show the concatenated name?

lib/ToolbarBadgeHub.jsm
```diff
@@ -243,10 +248,26 @@ class _ToolbarBadgeHub {
...
+        );
+        toolbarbutton
+          .querySelector(".toolbarbutton-text")
+          .setAttribute("id", "toolbarbutton-notification-description");
```
jcsteh: Can there only be one badged button at a time? If there can be multiple, this will be a problem because this id is global to the document and is not unique per badged button.

piatra: Yes, [There can be only one](https://www.youtube.com/watch?v=sqcLjcSloXs).

lib/ToolbarBadgeHub.jsm
piatra: From the Browser Toolbox:
Screenshot 2019-09-30 at 11 39 54

piatra: [From MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute#Possible_effects_on_user_agents_and_assistive_technology) it looks like `aria-labelledby` takes precedence over `aria-label` but this button only has a `label` defined.
If this behavior seems strange we could change it so that only the `labelledby` content is shown (moving `aria-label` responsibility completely on the badge, when active) and then revert when the badge is removed.

jcsteh:
I'm blind and can't read screenshots reliably. That said, if OCR is
accurate here, I see it does indeed include both names. I'm still puzzled
as to how exactly that's working. aria-labelledby should completely
override the accessible name with the name from the target id. I'll try to
dig into this at some point.

piatra:
You're right about something not being right. I tried to get a reduced test case of the What's new button: a button with `label` attribute and a `aria-labelledby` attribute, to see if I get the same result and in browser content the result is different. Only the `labelledby` content is shown.

Pasted my example below, it replicates the What's New button markup. The result I get is "bar".

data:text/html,<button id="foo" label="foo" aria-labelledby="bar">foo</button><label id="bar" value="baz" hidden>bar</div>

piatra:
I've created [a phabricator patch](https://phabricator.services.mozilla.com/D47581) if you want to try this locally. It's modified so that on a fresh profile the What's New button should show up immediately.

jcsteh:
Okay. So I figured out what's going on here.

You're pointing aria-labelledby to the hidden label element inside the toolbarbutton. That label contains the text you want: "New feature". However, that label element also has a value attribute containing "What's new". Since the value also participates in name computation for an XUL label element, you get value attribute + text content = "What's New New feature".

I think we really need a separate hidden element here. For example:
```
<toolbarbutton id="whats-new-menu-button" class="toolbarbutton-1" badged="true" aria-labelledby="toolbarbutton-notification-description whats-new-menu-button" data-l10n-id="cfr-whatsnew-button" label="What’s New" tooltiptext="What’s New">
...
<span id="toolbarbutton-notification-description" hidden>New feature:</span>
...
```

One other problem I'm seeing is that because tooltiptext is different to the label, it gets duplicated as the description. So with the current patch, we get:
"What's New New feature  button  What's New"
With my changes above, we'll get:
"New feature: What's New  button  What's New"
We can work around this by setting aria-describedby to the same value as aria-labelledby; in the above example, "toolbarbutton-notification-description whats-new-menu-button". The a11y code will see that the name and description are duplicated and will kill the description. That's a bit of a hack, but it's less ugly than users getting duplication.

content-src/asrouter/templates/OnboardingMessage/ToolbarBadgeMessage.schema.json
```diff
@@ -21,6 +21,16 @@
     "delay": {
       "type": "number",
       "description": "Optional delay in ms after which to show the notification"
+    },
+    "aria-label": {
```
jcsteh: I'd rather we didn't use the name aria-label here. While aria-label is about a11y labelling, aria-label refers to something more specific: an attribute which specifies the text of an entire label (not just part of the label) for an element. Also, people might get confused and think this will replace the label for the button. Perhaps we could just call this "description" or "badgeDescription"; i.e. it describes the badge. If it's essential that we make it clear this is related to a11y, perhaps we could go with "a11yDescription".

piatra: `badgeDescription` sounds good.

MarcoZehe: > 
> 
> r+ with comments addressed. I tested and this works as expected; bravo and thank you. However, before you land, I'd like @MarcoZehe to take this for a spin to confirm he's happy with the UX now that he can actually try it out.
> 
> Marco, [here's a try build](https://treeherder.mozilla.org/#/jobs?repo=try&revision=1b165d7c25af2e7966f970f2eb6efe531c68faaf). If you test it with a new profile, the What's New button will appear on the toolbar automatically and it will be badged.

So I created a new profile, and the What's New button did appear, but it did not say anything other than "What's New button" when I arrowed to it via the keyboard. I also inspected it, and it didn't contain anything other than that label, no description. So there seems to be something missing that's in your local build, but not the try build.

piatra: I've been able to reproduce the issue , not sure why we did not hit this earlier. I will update the PR and post a new try build link.

piatra: Update: It looks like for fresh profiles there's a race condition between showing the notification and inserting the `asrouter.ftl` strings file. The badge is added first and Fluent doesn't know where to get the strings (that's why we don't get show `New feature:`).
It doesn't seem to reproduce for 2nd profile runs (and onward) so that's why we probably missed this.

piatra: Here is the new try build for windows that attempts to fix the issue: https://treeherder.mozilla.org/#/jobs?repo=try&revision=efd6d3fd07394713578c304e5495fb8926d2356d

locales-src/asrouter.ftl
```diff
@@ -86,6 +86,10 @@ cfr-protections-panel-link-text = Learn more
 
 ## What's New toolbar button and panel
 
+# This string is used by screen readers to offer a text based alternative for
+# the notification icon
+cfr-badge-reader-label-newfeature = New feature:
```
flodolo: I'm confused: if it's read by screen readers, is the colon relevant, or actually detrimental?

MarcoZehe:
This new try server build worked for me. It is also good that the colon and space are there, so that it says "New feature: What's New button". The colon makes the synthesizer pause for a moment and doesn't run the words together without taking a slight virtual breath. Makes it a little more distinct and noticeable. CC @flodolo

jcsteh:
I did actually see this issue where the new feature text was missing on
first run when I tried your initial patch, but when I tried again with a
clean profile yesterday, it worked. So I thought it must have been
something I did or something you fixed in the second revision, hence my not
mentioning it. Glad you were able to fix it. :)

PR #5273, opened 2019-08-22
Bug 1575888 - Enable the What's New toolbar badge

piatra:
There's no risk of showing up accidentally because there are no what's new messages yet.

PR #5327, opened 2019-09-27
(nobug) - Remove earliestFirefoxVersion attribute for causing perform…

piatra:
…ance issues

And more generally because detecting version numbers is not a good way to target upgrades.
[Link to backouts when we tried to use this attribute](https://github.com/mozilla/activity-stream/pull/5273#issuecomment-524525961).


piatra:
Yes, we never used it.

PR #5329, opened 2019-09-30
Bug 1584625 - browser_aboutwelcome.js mochitest fails when run twice

piatra:
The issue is caused by `browser_onboarding_rtamo.js` that doesn't properly cleanup the attribution data. Mochitests use a fresh profile on every run but attribution data is persisted to disk.
Cleanup is different between mac and windows because the APIs are platform specific and `AttributionCode.jsm` doesn't expose a unified API to restore everything.

### DIFF
--- a/test/browser/browser_onboarding_rtamo.js
+++ b/test/browser/browser_onboarding_rtamo.js
@@ -33,8 +33,21 @@ async function setRTAMOOnboarding() {
   ASRouter._updateMessageProviders();
   await ASRouter.loadMessagesFromAllProviders();
 
-  registerCleanupFunction(() => {
+  registerCleanupFunction(async () => {
+    // Separate cleanup methods between mac and windows
+    const { path } = Services.dirsvc.get("GreD", Ci.nsIFile).parent.parent;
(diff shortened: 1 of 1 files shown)


Mardak:
This was reverted with https://github.com/mozilla/activity-stream/commit/1977f0f6b93b35dbe21622bd80c9c3751b8fd81f for failing windows bc tests because it tries to access `Cc["@mozilla.org/mac-attribution;1"]` on windows:

```
[task 2019-10-04T14:47:07.779Z] 14:47:07     INFO - TEST-UNEXPECTED-FAIL | browser/components/newtab/test/browser/browser_onboarding_rtamo.js | Cleanup function threw an exception - at chrome://mochitests/content/browser/browser/components/newtab/test/browser/browser_onboarding_rtamo.js:39 - TypeError: Cc['@mozilla.org/mac-attribution;1'] is undefined
```

dmose:
I wonder why this didn't cause the taskcluster runs to fail.  Which is to say, shouldn't this failure happen on Linux too?

Mardak:
The test skips linux and runs only windows and mac.

PR #5332, opened 2019-10-01
Backports

ncloudioj:
@piatra Looks like `Bug 1576601` caused a coverage regression in `ToolbarPanelHub.test.js`, can you take a look at that?


piatra:
Sure. This can go into central now and I'll fix up the tests in master.

PR #5340, opened 2019-10-07
Bug 1584625 - browser_aboutwelcome.js mochitest fails when run twice

piatra:
Updated to detect OSX. And here is a try run:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=6143cc3837cde919722f6a8645440722ba122cd1


piatra:
@punamdahiya try push is looking ok, do you mind having another look? thanks.

PR #5342, opened 2019-10-08
Bug 1585434 - Put ASRouter remote l10n behind the pref

ncloudioj:
This, coupled with this [patch](https://phabricator.services.mozilla.com/D48412), allows us to enable/disable the remote l10n for ASRouter.

What's changed?
* Add a pref observer for `ASRouter`, upon pref changes, it delegates the l10n reload work to `CFRPageActions`. 
* Add a function `reloadL10n` to `PageAction`, by which `CFRPageActions` can reload the l10n file for all the existing pageActions.

### DIFF
--- a/lib/CFRPageActions.jsm
+++ b/lib/CFRPageActions.jsm
@@ -47,6 +47,8 @@ const CATEGORY_ICONS = {
  * profile directory.
  */
 const RS_DOWNLOADED_FILE_SUBDIR = "settings/main/ms-language-packs";
+const USE_REMOTE_L10N_PREF =
+  "browser.newtabpage.activity-stream.asrouter.useRemoteL10n";
 
 /**
  * A WeakMap from browsers to {host, recommendation} pairs. Recommendations are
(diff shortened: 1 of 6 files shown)


lib/CFRPageActions.jsm
```diff
@@ -118,10 +120,21 @@ class PageAction {
...
+   *   * the remote Fluent file is not available
+   *   * it was told to use the local Fluent file
    */
   _createDOML10n() {
+    if (!Services.prefs.getBoolPref(USE_REMOTE_L10N_PREF, true)) {
+      return new DOMLocalization([
+        "browser/newtab/asrouter.ftl",
+        "browser/branding/brandings.ftl",
+        "browser/branding/sync-brand.ftl",
+        "branding/brand.ftl",
```
Mardak: Let's pull out this array to share in the `return new DOMLocalization([…` below.

lib/CFRPageActions.jsm
```diff
@@ -914,6 +931,18 @@ const CFRPageActions = {
     this.PageActionMap = PageActionMap;
     this.RecommendationMap = RecommendationMap;
   },
+
+  /**
+   * Reload the l10n Fluent files for all PageActions
+   */
+  reloadL10n() {
+    for (const win of Services.wm.getEnumerator("navigator:browser")) {
```
Mardak: We were using EveryWindow elsewhere which seems to be the new way to enumerate browser windows?

ncloudioj: OK, I will swap that in.

ncloudioj: Hmm, looks like there are a few other modules that are still calling `getEnumerator`, wonder if we should have a separate patch to refactor it?

Mardak: Ah for `clearRecommendations` sure those can be done together separately

lib/CFRPageActions.jsm
```diff
@@ -118,10 +120,21 @@ class PageAction {
...
-   * Note that it still uses the packaged Fluent file as the fallback if the remote one
-   * is not available.
+   *
+   * Note: it will use the local Fluent file in any of following cases:
+   *   * the remote Fluent file is not available
+   *   * it was told to use the local Fluent file
    */
   _createDOML10n() {
+    if (!Services.prefs.getBoolPref(USE_REMOTE_L10N_PREF, true)) {
+      return new DOMLocalization([
```
Mardak: Oh, I just realized, this probably could have been simplified to:
```js
_createDOML10n() {
  async function* generateBundles() { … }

  return new DOMLocalization(["ftl stuff"], getBoolPref() ? generateBundles : null);
```

ncloudioj: Doh, I've pushed it to autoland. Will fix it in bug 1588199.

ncloudioj:
Close it as I have ported it directly into m-c.

PR #5346, opened 2019-10-11
Bug 1585147 - Migrate AS Router telemetry to Structured Ingestion (GCP)

ncloudioj:
Gear switch from AWS to GCP for ASRouter telemetry.

Steps to test:
* Enable PingCentre logging by flipping `browser.ping-centre.log`
* Enable telemetry for Activity Stream by flipping `browser.newtabpage.activity-stream.telemetry`
* Enable Structured Ingestion by flipping `browser.newtabpage.activity-stream.telemetry.structuredIngestion` 

Then you should be able to see ping logging in the browser console.


ncloudioj:
The new schema was added in [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/430).

Change highlights:
* We have four topics for ASRouter: onboarding, cfr, snippets, and undesired-events. As such, we don't need the `action` field in the ping
* impresion_id and client_id are both included in the cfr related pings, we make sure only one of them present in the ping
* We only use impression_id for ASRouter undesired events
* Snippets and onboarding pings still use client_id, but we don't store it in the impression_id field any more
* Rename the `value` field to `event_context`, and JSON.stringify it if it's of `Object` type

lib/TelemetryFeed.jsm
```diff
@@ -577,32 +577,28 @@ this.TelemetryFeed = class TelemetryFeed {
    * 3). In shield experiments conducted in release, it collects client_id and message_id
    */
   applyCFRPolicy(ping) {
+    let excludeClientID;
```
piatra: I'm not sure about these changes that add `excludeClientID`, the logic is simple but it's difficult to follow through the code.
The behavior of `apply*Policy` now changed, but the name hasn't and it's only half applying the policy and half deferring for later in `_createStructuredIngestionPing`.
Could we change things around so that in `handleASRouterUserEvent` we don't immediately call `createASRouterEvent` and instead do it later in `_createStructuredIngestionPing`?
We could instead of passing `excludeClientID`, pass the policy name that should be applied (or the `event.action`).

lib/TelemetryFeed.jsm
```diff
@@ -634,33 +635,40 @@ this.TelemetryFeed = class TelemetryFeed {
...
       );
     }
   }
 
-  sendASRouterEvent(event_object) {
-    if (this.telemetryEnabled) {
-      this.pingCentreForASRouter.sendPing(event_object, {
-        filter: ACTIVITY_STREAM_ID,
-      });
+  sendASRouterEvent(event_object, pingType, version, excludeClientID) {
```
piatra: I think cleaning up the `excludeClientID` would also help merging this method with `sendStructuredIngestionEvent`.

lib/TelemetryFeed.jsm
ncloudioj: Thanks for pointing it out! Agreed, this `excludeClientID` thing complicated the code substantially. I've refactored it out so that the logic of "shall we use `client_id` or `impression_id`?" was completely decided by those `applyXYZPolicy` functions.

lib/TelemetryFeed.jsm
ncloudioj: I've also merged these two send functions as they basically did the same thing.

docs/v2-system-addon/data_events.md
```diff
@@ -1099,7 +1099,7 @@ This reports a failure in the Remote Settings loader to load messages for Activi
   "user_prefs": 7,
   "event": ["ASR_RS_NO_MESSAGES" | "ASR_RS_ERROR"],
   // The value is set to the ID of the message provider. For example: remote-cfr, remote-onboarding, etc.
-  "value": "REMOTE_PROVIDER_ID"
+  "event_context": "REMOTE_PROVIDER_ID"
```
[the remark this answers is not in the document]
ncloudioj: Good catch. I will update the description.

lib/ToolbarPanelHub.jsm
```diff
@@ -477,7 +477,7 @@ class _ToolbarPanelHub {
         message_id: message.id,
         bucket_id: message.id,
         event,
-        value: options.value,
+        event_context: JSON.stringify(options.value),
```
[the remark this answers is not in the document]
ncloudioj: Hmm, could you point me out any potential broken cases? Based on the contract of `option.value`, its value should be of one of the following types: undefined, string, and object. I believe `JSON.stringify` should generate the string representation of all of them as expected (note: `JSON.stringify(undefined) -> undefined`).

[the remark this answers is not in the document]
ncloudioj: Ah, I think i see what your concern is now. Agreed, this stringification would better be done at later time. Will fix it up.

jgmize:
Hi @ncloudioj, nice work on this. Now that this is merged, how soon would you expect us to see data from real clients (as opposed to  [test data](https://sql.telemetry.mozilla.org/queries/65754/source)) in BQ?

ncloudioj:
Thanks, @jgmize! this should be merged to m-c soon, we usually do that at the end of the week, so we should be able to see traffic from Nightly early next week.

PR #5350, opened 2019-10-11
Update data_events.md

punamdahiya:


punamdahiya:
@Mardak Updated doc with data review feedback.  Looks like don't have permission to change directly in master and needs review before committing. Thanks

PR #5355, opened 2019-10-17
Bug 1588524 - The 'What's New' toolbar button is redisplayed after br…

piatra:
…owser restart if it was previously accessed using keyboard navigation

I was able to verify the same behavior as in the bug report: the keypress events are different between mac and windows. Our `click` event listener was enough on mac to handle keyboard access but on windows this never triggered. Added a specific `keypress` event listener.


piatra:
On second thought: we shouldn't need the `click` event listener since `keypress` should handle both platforms.

PR #5359, opened 2019-10-18
Bug 1589659 - Enabling hover state for privacy modal

gvn:
Test steps: https://bugzilla.mozilla.org/show_bug.cgi?id=1589659#c0


content-src/components/DiscoveryStreamComponents/DSPrivacyModal/DSPrivacyModal.jsx
```diff
@@ -40,6 +40,7 @@ export class DSPrivacyModal extends React.PureComponent {
           <h3 data-l10n-id="newtab-privacy-modal-header" />
           <p data-l10n-id="newtab-privacy-modal-paragraph" />
           <SafeAnchor
+            className="a11y-link-button"
```
Mardak: The SafeAnchor isn't really an accessibility link-button. Seems like the issue is that we're passing in a `<span>` as a child instead of setting the text directly as the child of the resulting `<a>`. Probably can just pass the data-l10n-id to anchor ?

Similarly, seems like SafeAnchor was intended for remote article urls, so it has custom referrer overrides and url checks. If it doesn't need that custom behavior, a plain `<a>` would probably work too.

PR #5361, opened 2019-10-21
Bug 1589653 - Provide bookmark object to targeting expression

piatra:
The issue is that `bookmark-icon-updated` is also triggered when adding a bookmark not only when visiting a bookmarked URL and this results in showing the message at an undesired time.
I didn't want to make a super specific "opened-already-bookmarked-url" trigger so instead I'm passing in through the `context` parameter of the trigger the `bookmark` object associated with the current url.
This contains:
```
title,
url,
_lastModified_,
...
```
We can then use this information inside of the targeting expression `currentDate|date - bookmark.lastModified > 60 * 60`.
It can potentially allow us to do other interesting messages `bookmark.url in [list of web developer resources]` => show some new devtools feature?


piatra:
Moved to https://phabricator.services.mozilla.com/D51881.

piatra:
Moved to https://phabricator.services.mozilla.com/D51881.

PR #5362, opened 2019-10-21
Bug 1589025 - Displaying privacy modal in correct position

gvn:
~~@k88hudson As far as I can tell & observe this isn't causing any regressions, but since the modal is used by trailhead I figured you might be able to give a 2nd opinion there. Positioning is `fixed` by default and this was overriding to `absolute` – unsure why. 🤔~~

Test steps: https://bugzilla.mozilla.org/show_bug.cgi?id=1589025#c0

### DIFF
--- a/content-src/asrouter/templates/Trailhead/_Trailhead.scss
+++ b/content-src/asrouter/templates/Trailhead/_Trailhead.scss
@@ -332,7 +332,7 @@
     overflow-y: hidden;
   }
 
-  .modalOverlayInner {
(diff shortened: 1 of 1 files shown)


Mardak:
Without the `position: absolute`, users can't scroll down to "Start Browsing"

gvn:
> Without the position: absolute, users can't scroll down to "Start Browsing"

Ok, making the selector more specific then so it doesn't target every modal.

PR #5363, opened 2019-10-22
Bug 1589503 - Prevent bad remote triggering data from breaking all me…

piatra:
…ssages


piatra:
Moved this to phabricator: https://phabricator.services.mozilla.com/D51664.

PR #5365, opened 2019-10-29
Bug 1587495 - Document all DS prefs

gvn:


ScottDowne:
Only 1 pref I found that I can see being shared `browser.newtabpage.blocked` which works the same as the campaign block pref.

PR #5366, opened 2019-10-30
Bug 1589731 - Update targeting for What's New and Firefox accounts badge

piatra:
The requirement was that 
> For new profiles (<= 28 days) the Firefox accounts badge should show first and then the What's New button badge.

We now have (in `priority` order):
  1. FXA_ACCOUNTS_BADGE_NEW_USER - same targeting as FXA_ACCOUNTS_BADGE + profile has to be newer than 28 days so it captures new users
  2. WHATS_NEW_BADGE
  3. FXA_ACCOUNTS_BADGE - everyone that never opened the FxA panel, get triggered if other 2 options are exhausted

### DIFF
--- a/lib/OnboardingMessageProvider.jsm
+++ b/lib/OnboardingMessageProvider.jsm
@@ -465,6 +465,20 @@ const ONBOARDING_MESSAGES = () => [
     trigger: { id: "firstRun" },
   },
   {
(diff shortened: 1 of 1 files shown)


piatra:
Moved to https://phabricator.services.mozilla.com/D51876.

PR #5367, opened 2019-10-30
Bug 1567273 - Enable DE using Discovery Stream

ScottDowne:
A note, this relies on https://phabricator.services.mozilla.com/D53236 in MC to properly enabled 7 rows in English based locales. So if you happen to be using this patch locally without the above MC patch, you need to create this pref `browser.newtabpage.activity-stream.discoverystream.lang-layout-config` with the value `en`

To test:

1. Setup a profile enabled with DE. You can find help with how to do that here: https://github.com/mozilla/activity-stream/blob/master/docs/v2-system-addon/geo_locale.md
2. Enable Discovery Stream.
3. Load a new tab.

Expected: no spocs, no english strings, no navigation component under cards, and complete new content that's de based. Also check that en-US/en-CA has not changed. Also, DE should only be 1 row of content. You can change the value in `browser.newtabpage.activity-stream.discoverystream.lang-layout-config` to `en,de` to see 7 rows in DE.

### DIFF
--- a/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
+++ b/content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
@@ -26,7 +26,7 @@ const ALLOWED_CSS_URL_PREFIXES = [
   "https://img-getpocket.cdn.mozilla.net/",
 ];
 const DUMMY_CSS_SELECTOR = "DUMMY#CSS.SELECTOR";
-let rickRollCache = []; // Cache of random probability values for a spoc position
+let rollCache = []; // Cache of random probability values for a spoc position
 
 /**
  * Validate a CSS declaration. The values are assumed to be normalized by CSSOM.
@@ -264,17 +264,18 @@ export class _DiscoveryStreamBase extends React.PureComponent {
 
   componentWillReceiveProps(oldProps) {
     if (this.props.DiscoveryStream.layout !== oldProps.DiscoveryStream.layout) {
-      rickRollCache = [];
+      rollCache = [];
     }
   }
 
   render() {
     // Select layout render data by adding spocs and position to recommendations
-    const { layoutRender, spocsFill } = selectLayoutRender(
-      this.props.DiscoveryStream,
-      this.props.Prefs.values,
-      rickRollCache
-    );
+    const { layoutRender, spocsFill } = selectLayoutRender({
+      state: this.props.DiscoveryStream,
+      prefs: this.props.Prefs.values,
+      rollCache,
+      lang: this.props.document.documentElement.lang,
+    });
     const { config, spocs, feeds } = this.props.DiscoveryStream;
 
     // Send SPOCS Fill if any. Note that it should not send it again if the same
@@ -412,4 +413,5 @@ export const DiscoveryStreamBase = connect(state => ({
   DiscoveryStream: state.DiscoveryStream,
   Prefs: state.Prefs,
   Sections: state.Sections,
+  document: global.document,
 }))(_DiscoveryStreamBase);
--- a/content-src/components/DiscoveryStreamComponents/DSMessage/DSMessage.jsx
+++ b/content-src/components/DiscoveryStreamComponents/DSMessage/DSMessage.jsx
@@ -4,6 +4,7 @@
 
 import React from "react";
 import { SafeAnchor } from "../SafeAnchor/SafeAnchor";
+import { FluentOrText } from "content-src/components/FluentOrText/FluentOrText";
 
 export class DSMessage extends React.PureComponent {
   render() {
@@ -17,11 +18,13 @@ export class DSMessage extends React.PureComponent {
             />
           )}
           {this.props.title && (
-            <span className="title-text">{this.props.title}</span>
+            <span className="title-text">
+              <FluentOrText message={this.props.title} />
+            </span>
           )}
           {this.props.link_text && this.props.link_url && (
             <SafeAnchor className="link" url={this.props.link_url}>
-              {this.props.link_text}
+              <FluentOrText message={this.props.link_text} />
             </SafeAnchor>
           )}
         </header>
--- a/content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
+++ b/content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
@@ -4,6 +4,7 @@
 
 import React from "react";
 import { SafeAnchor } from "../SafeAnchor/SafeAnchor";
+import { FluentOrText } from "content-src/components/FluentOrText/FluentOrText";
 
 export class Topic extends React.PureComponent {
   render() {
@@ -25,7 +26,11 @@ export class Navigation extends React.PureComponent {
     const header = this.props.header || {};
     return (
       <div className={`ds-navigation ds-navigation-${alignment}`}>
-        {header.title ? <div className="ds-header">{header.title}</div> : null}
+        {header.title ? (
+          <FluentOrText message={header.title}>
+            <div className="ds-header" />
+          </FluentOrText>
+        ) : null}
         <div>
           <ul>
             {links &&
--- a/content-src/lib/selectLayoutRender.js
+++ b/content-src/lib/selectLayoutRender.js
@@ -2,7 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export const selectLayoutRender = (state, prefs, rickRollCache) => {
+export const selectLayoutRender = ({
+  state = {},
+  prefs = {},
+  rollCache = [],
+  lang = "",
+}) => {
   const { layout, feeds, spocs } = state;
   let spocIndexMap = {};
   let bufferRollCache = [];
@@ -23,11 +28,11 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
 
       // Cache random number for a position
       let rickRoll;
-      if (!rickRollCache.length) {
+      if (!rollCache.length) {
         rickRoll = Math.random();
         bufferRollCache.push(rickRoll);
       } else {
-        rickRoll = rickRollCache.shift();
+        rickRoll = rollCache.shift();
         bufferRollCache.push(rickRoll);
       }
 
@@ -61,6 +66,10 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
 
   if (!prefs["feeds.topsites"]) {
     filterArray.push("TopSites");
+  }
+
+  if (!lang.startsWith("en-")) {
+    filterArray.push("Navigation");
   }
 
   if (!prefs["feeds.section.topstories"]) {
@@ -213,9 +222,9 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
 
   const layoutRender = renderLayout();
 
-  // If empty, fill rickRollCache with random probability values from bufferRollCache
-  if (!rickRollCache.length) {
-    rickRollCache.push(...bufferRollCache);
+  // If empty, fill rollCache with random probability values from bufferRollCache
+  if (!rollCache.length) {
+    rollCache.push(...bufferRollCache);
   }
 
   // Generate the payload for the SPOCS Fill ping. Note that a SPOC could be rejected
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -469,10 +469,11 @@ const PREFS_CONFIG = new Map([
       title: "Configuration for the new pocket new tab",
       getValue: ({ geo, locale }) => {
         // PLEASE NOTE:
-        // hardcoded_layout in `lib/DiscoveryStreamFeed.jsm` only works for en-* and requires refactoring for non english locales
+        // hardcoded_layout in `lib/DiscoveryStreamFeed.jsm` only works for en-* and DE and requires refactoring for other locales
         const dsEnablementMatrix = {
           US: ["en-CA", "en-GB", "en-US"],
           CA: ["en-CA", "en-GB", "en-US"],
+          DE: ["de", "de-DE", "de-AT", "de-CH"],
         };
 
         // Verify that the current geo & locale combination is enabled
--- a/lib/DiscoveryStreamFeed.jsm
+++ b/lib/DiscoveryStreamFeed.jsm
@@ -58,6 +58,7 @@ const PREF_IMPRESSION_ID = "browser.newtabpage.activity-stream.impressionId";
 const PREF_ENABLED = "discoverystream.enabled";
 const PREF_HARDCODED_BASIC_LAYOUT = "discoverystream.hardcoded-basic-layout";
 const PREF_SPOCS_ENDPOINT = "discoverystream.spocs-endpoint";
+const PREF_LANG_LAYOUT_CONFIG = "discoverystream.lang-layout-config";
 const PREF_TOPSTORIES = "feeds.section.topstories";
 const PREF_SPOCS_CLEAR_ENDPOINT = "discoverystream.endpointSpocsClear";
 const PREF_SHOW_SPONSORED = "showSponsored";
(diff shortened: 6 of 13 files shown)


content-src/components/DiscoveryStreamComponents/DSMessage/DSMessage.jsx
```diff
@@ -17,7 +18,7 @@ export class DSMessage extends React.PureComponent {
             />
           )}
           {this.props.title && (
-            <span className="title-text">{this.props.title}</span>
```
ScottDowne: I probably didn't need to make this change.

content-src/lib/selectLayoutRender.js
```diff
@@ -63,6 +68,10 @@ export const selectLayoutRender = (state, prefs, rickRollCache) => {
     filterArray.push("TopSites");
   }
 
+  if (!lang.startsWith("en-")) {
+    filterArray.push("Navigation");
```
ScottDowne: Test for this.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -207,7 +208,9 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     // The server somtimes returns this value already replaced, but we try this for two reasons:
     // 1. Layout endpoints are not from the server.
     // 2. Hardcoded layouts don't have this already done for us.
-    const endpoint = rawEndpoint.replace("$apiKey", apiKey);
+    const endpoint = rawEndpoint
```
ScottDowne: Need a test for this.

lib/ActivityStream.jsm
```diff
@@ -476,10 +476,11 @@ const PREFS_CONFIG = new Map([
       title: "Configuration for the new pocket new tab",
       getValue: ({ geo, locale }) => {
         // PLEASE NOTE:
-        // hardcoded_layout in `lib/DiscoveryStreamFeed.jsm` only works for en-* and requires refactoring for non english locales
+        // hardcoded_layout in `lib/DiscoveryStreamFeed.jsm` only works for en-* and DE and requires refactoring for other locales
         const dsEnablementMatrix = {
           US: ["en-CA", "en-GB", "en-US"],
           CA: ["en-CA", "en-GB", "en-US"],
+          DE: ["de", "de-DE", "de-AT", "de-CH"],
```
ScottDowne: Can probably get tests for geo and regions.

content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
```diff
@@ -25,7 +26,11 @@ export class Navigation extends React.PureComponent {
     const header = this.props.header || {};
     return (
       <div className={`ds-navigation ds-navigation-${alignment}`}>
-        {header.title ? <div className="ds-header">{header.title}</div> : null}
+        {header.title ? (
+          <div className="ds-header">
+            <FluentOrText message={header.title} />
```
ScottDowne: Test this

content-src/components/DiscoveryStreamComponents/DSMessage/DSMessage.jsx
```diff
@@ -17,11 +18,13 @@ export class DSMessage extends React.PureComponent {
...
+            <span className="title-text">
+              <FluentOrText message={this.props.title} />
+            </span>
           )}
           {this.props.link_text && this.props.link_url && (
             <SafeAnchor className="link" url={this.props.link_url}>
-              {this.props.link_text}
+              <FluentOrText message={this.props.link_text} />
```
ScottDowne: Test this

content-src/components/DiscoveryStreamBase/DiscoveryStreamBase.jsx
```diff
@@ -26,7 +26,7 @@ const ALLOWED_CSS_URL_PREFIXES = [
   "https://img-getpocket.cdn.mozilla.net/",
 ];
 const DUMMY_CSS_SELECTOR = "DUMMY#CSS.SELECTOR";
-let rickRollCache = []; // Cache of random probability values for a spoc position
+let rollCache = []; // Cache of random probability values for a spoc position
```
ScottDowne: Just changing this little bit of whimsical naming, mostly because now it's exposed as an actual option value in function call.

content-src/lib/selectLayoutRender.js
```diff
@@ -2,7 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export const selectLayoutRender = (state, prefs, rickRollCache) => {
+export const selectLayoutRender = ({
+  state = {},
+  prefs = {},
+  rollCache = [],
```
ScottDowne: Just changing this little bit of whimsical naming, mostly because now it's exposed as an actual option value in function call.

content-src/components/DiscoveryStreamComponents/DSMessage/DSMessage.jsx
ScottDowne: These accept a text value, and it'll then just display the text value, or a Fluent object which then does the proper locale value lookup.

content-src/lib/selectLayoutRender.js
ScottDowne: Right now this is hard coded in here, as the component doesn't support non english right now. For the time being it's a 1 off.

lib/ActivityStream.jsm
ScottDowne: I got these values from the previous values used for Activity Stream topstories.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1547,7 +1555,9 @@ getHardcodedLayout = basic => {
           {
             type: "TopSites",
             header: {
-              title: "Top Sites",
+              title: {
```
ScottDowne: Title in the layout can be either a localized object or a string. If it's a hard coded string, we must be mindful that it'll show up in de. This is probably fine as hard coded layouts are under our control, and remote layouts are the likely candidates for hard coded strings, which is for testing purposes, and mostly we do test on a per locale basis.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1586,7 +1601,7 @@ getHardcodedLayout = basic => {
             feed: {
               embed_reference: null,
               url:
-                "https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=en-US&count=30",
+                "https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=$locale&count=30",
```
ScottDowne: This is what makes the feeds update their content returned based on locale.

content-src/components/DiscoveryStreamComponents/Navigation/Navigation.jsx
Mardak: This results in a div with a span inside. You can have `FluentOrText` set the data-l10n-id on the div instead by putting it as the child instead.

ScottDowne: Thanks, not sure if I follow, is there an example I can see?

Mardak: https://github.com/mozilla/activity-stream/blob/f5ff7c676dccbf94b660f397b5b0b09e8d43a992/content-src/components/Sections/Sections.jsx#L303-L305

ScottDowne: That's pretty wild. Updated.

gvn:
@Mardak is this good to land now?

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1481,7 +1489,7 @@ getHardcodedLayout = basic => {
               feed: {
                 embed_reference: null,
                 url:
-                  "https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=en-US&feed_variant=default_spocs_on",
+                  "https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=$locale&feed_variant=default_spocs_on",
```
Mardak: Is the server expecting `locale_lang=en-GB` as well as `locale_lang=de-CH` etc?

Mardak: Also, given the naming of `locale_lang` instead of just `locale` is the server actually expecting just "en" and "de" which are the "language" portion of a locale while the "GB" and "CH" portion is technically the "region" portion of a locale.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -74,6 +74,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
 
     // Persistent cache for remote endpoint data.
     this.cache = new PersistentCache(CACHE_KEY, true);
+    this.locale = Services.locale.appLocaleAsLangTag;
```
Mardak: This should have an observer as the app's primary locale can change, but probably just file a followup bug.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1451,14 +1454,19 @@ getHardcodedLayout = basic => {
...
             {
               type: "Message",
               header: {
-                title: "Recommended by Pocket",
+                title: {
+                  id: "newtab-section-header-pocket",
+                  values: { provider: "pocket" },
+                },
                 subtitle: "",
                 link_text: "How it works",
```
Mardak: This probably wants to have a string id like the other "How it works" elsewhere

lib/DiscoveryStreamFeed.jsm
ScottDowne: This should just work, because it's using the same pattern as the old AS which wasn't a problem, but I def think it's worth double checking.

ScottDowne: Also, when testing I saw german feeds, but I can see a case where the locale code of de, is being interpreted as DE and en-US is being considered incorrect, and thus the server defaults to US, which would give us what we expect, but in a very incorrect way. Which if the case, has been a bug on the previous AS system too.

ScottDowne: Also going to see what the server does with "en-GB/en-CA/de-CH/etc"

Also question, the way you said "has only received "en-US" even for en-GB users" makes me ask the question of, are en-GB users primarily using the en-US version?

lib/DiscoveryStreamFeed.jsm
ScottDowne: Did the old AS have a observer for this, if so, probably easiest if I just connect to that observer here, instead of making a new one.

ScottDowne:
@Mardak "Separately, there's some more hardcoded english strings in the layout as well as various DS components, e.g., https://bugzilla.mozilla.org/show_bug.cgi?id=1573905 that might want to be correctly localized for DE."

Spocs are not currently supported outside of US region. The mvp of this patch was to enable enough localized content to get the non spoc hard coded layout translated, but happy to start opening up some new strings. I'm worried though if we start using strings that are not yet translated, and not sure if we're going to be using anytime soon, that we're exposing ourselves and the l10n community to extra work.

lib/DiscoveryStreamFeed.jsm
ScottDowne: I verified this in a test, at the very least, de-CH also pinged this question to back end, but I don't think we should wait on it.

I think I have a good idea how the back end logic works, because I tried bogus locales like:
https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=en-AB&count=30
https://getpocket.cdn.mozilla.net/v3/firefox/global-recs?version=3&consumer_key=$apiKey&locale_lang=de-AB&count=30 

And both returned the proper language content. So pretty sure the server is looking for starts with en or de. Or maybe starts with de or default to en.

ScottDowne:
I'll see what hard coded strings we can turn on that already have localized values for them.

ScottDowne:
From what I can tell, https://bugzilla.mozilla.org/show_bug.cgi?id=1573905 covers the cases I could find that are most concerning, and are both either behind localized feed responses, or something we're currently limited to en-US based experiments. Bug 1573905 is a p2, so we can sort it out in 73.

lib/ActivityStream.jsm
```diff
@@ -483,17 +483,25 @@ const PREFS_CONFIG = new Map([
...
+          !!dsBasicLayoutMatrix[geo] && dsBasicLayoutMatrix[geo].includes(locale);
+
         return JSON.stringify({
           api_key_pref: "extensions.pocket.oAuthConsumerKey",
           collapsible: true,
           enabled: isEnabled,
           show_spocs: showSpocs({ geo }),
           hardcoded_layout: true,
           personalized: true,
+          ...(isBasicLayout ? { hardcoded_basic_layout: true } : {}),
```
gvn: So, is the idea here to make Germany get the 1 row basic layout by default and then rollout a pref change to switch some of the population to the 7 row layout?

ScottDowne: Yes, but given this patch was still in review, I think I was able to get my solution into this patch.

lib/ActivityStream.jsm
```diff
@@ -483,25 +483,17 @@ const PREFS_CONFIG = new Map([
           DE: ["de", "de-DE", "de-AT", "de-CH"],
         };
 
-        const dsBasicLayoutMatrix = {
```
ScottDowne: Removing this logic now, it's basically controlled by a pref.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -58,6 +58,7 @@ const PREF_IMPRESSION_ID = "browser.newtabpage.activity-stream.impressionId";
 const PREF_ENABLED = "discoverystream.enabled";
 const PREF_HARDCODED_BASIC_LAYOUT = "discoverystream.hardcoded-basic-layout";
 const PREF_SPOCS_ENDPOINT = "discoverystream.spocs-endpoint";
+const PREF_LANG_LAYOUT_CONFIG = "discoverystream.lang-layout-config";
```
ScottDowne: This pref needs to exist in a patch with MC which I have not created yet.

If you want to test it now, you can add this string pref:
`browser.newtabpage.activity-stream.discoverystream.lang-layout-config` and give the value of "en" or "en,de"

lib/PrefsFeed.jsm
```diff
@@ -102,6 +102,10 @@ this.PrefsFeed = class PrefsFeed {
       "browser.newtabpage.activity-stream.discoverystream.spocs-endpoint",
       ""
     );
+    let discoveryStreamLangLayoutConfig = Services.prefs.getStringPref(
```
ScottDowne: The changes in this file just keep the page in sync with the pref value in case it changes.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -1405,6 +1409,7 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
           case PREF_ENABLED:
           case PREF_HARDCODED_BASIC_LAYOUT:
           case PREF_SPOCS_ENDPOINT:
+          case PREF_LANG_LAYOUT_CONFIG:
```
ScottDowne: This ensures if the pref value changes, we update newtab.

lib/DiscoveryStreamFeed.jsm
```diff
@@ -365,12 +366,15 @@ this.DiscoveryStreamFeed = class DiscoveryStreamFeed {
     }
 
     if (!layoutResp || !layoutResp.layout) {
+
+      const langLayoutConfig = this.store.getState().Prefs.values[PREF_LANG_LAYOUT_CONFIG].split(",");
+      const isBasic = this.config.hardcoded_basic_layout ||
+          this.store.getState().Prefs.values[PREF_HARDCODED_BASIC_LAYOUT] ||
+          !langLayoutConfig.find(lang => this.locale.startsWith(lang.trim()));
```
ScottDowne: If the lang does not exist in the array, we use the basic layout.

PR #5368, opened 2019-10-31
Bug 1593018 - Placements bug after layout update

ScottDowne:
Steps to test in bug 1593018


ScottDowne:
This just got accidentally removed during a refactor, and the thing that got removed didn't get a test.

This puts it back, and also adds a test.

PR #4864, opened 2019-03-22
Bug 1520890 - Spocs need to use v1 personaliation at the least

ScottDowne:

PR #5369, opened 2019-11-01
Bug 1593290 - Personalization switch for Discovery Stream

ScottDowne:
To test, I think we can try using these steps again, should work: https://github.com/mozilla/activity-stream/pull/4864#issuecomment-479962039

And also, ensure basic spoc functionality works.


gvn:
Code seems fine, but not seeing this on step 11... `Scroll to the bottom of that page until you see a spoc in "spoc data" with the id 36296 (there should be only one spoc)`

ScottDowne:
Updated the steps to test in the old bug, working for me now.

gvn:
Confirmed repro steps are working now.

PR #5373, opened 2019-11-12
Bug 1595058 - add a new ping type for pre-trailhead

ncloudioj:


lib/TelemetryFeed.jsm
```diff
@@ -579,6 +579,8 @@ this.TelemetryFeed = class TelemetryFeed {
       case "snippets_user_event":
         event = await this.applySnippetsPolicy(event);
         break;
+      // Bug 1594125 added a new onboarding-like provider called `whats-new-panel`.
+      case "whats-new-panel_user_event":
       case "onboarding_user_event":
         event = await this.applyOnboardingPolicy(event);
```
Mardak: Is it important that the `type` is set to "onboarding" vs something different for what's new panel?

ncloudioj: It's not important here since we only use it to decide which policy to be applied to the payload. We have other fields in the payload to idenfity the message source ('whats-new-panel').

ncloudioj:
Close here to favor https://phabricator.services.mozilla.com/D52755

PR #5376, opened 2019-11-13
Bug 1574297 - "How it works" -> "What's Pocket?"

gvn:
Test Steps:

- Open browser & look for "What's Pocket?" (See pic)

Screen Shot 2019-11-13 at 10 58 44 AM

### DIFF
--- a/locales-src/newtab.ftl
+++ b/locales-src/newtab.ftl
@@ -185,7 +185,7 @@ newtab-discovery-empty-section-topstories-timed-out = Oops! We almost loaded thi
 # This is shown at the bottom of the trending stories section and precedes a list of links to popular topics.
 newtab-pocket-read-more = Popular Topics:
 newtab-pocket-more-recommendations = More Recommendations
-newtab-pocket-how-it-works = How it works
+newtab-pocket-how-it-works = What's Pocket?
 newtab-pocket-cta-button = Get { -pocket-brand-name }
 newtab-pocket-cta-text = Save the stories you love in { -pocket-brand-name }, and fuel your mind with fascinating reads.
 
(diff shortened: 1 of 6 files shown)


locales-src/newtab.ftl
```diff
@@ -185,7 +185,7 @@ newtab-discovery-empty-section-topstories-timed-out = Oops! We almost loaded thi
 # This is shown at the bottom of the trending stories section and precedes a list of links to popular topics.
 newtab-pocket-read-more = Popular Topics:
 newtab-pocket-more-recommendations = More Recommendations
-newtab-pocket-how-it-works = How it works
+newtab-pocket-how-it-works = What's Pocket?
```
flodolo: String changes require new IDs
https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_review.html#changes-to-existing-messages

Also, `’` instead of `'`

locales-src/newtab.ftl
```diff
@@ -185,7 +185,7 @@ newtab-discovery-empty-section-topstories-timed-out = Oops! We almost loaded thi
 # This is shown at the bottom of the trending stories section and precedes a list of links to popular topics.
 newtab-pocket-read-more = Popular Topics:
 newtab-pocket-more-recommendations = More Recommendations
-newtab-pocket-how-it-works = How it works
+newtab-pocket-whats-pocket = What’s Pocket?
 newtab-pocket-cta-button = Get { -pocket-brand-name }
```
Mardak: If the new string is used in the same context as this one, you should be able to use `{ -pocket-brand-name }`

gvn:
@ScottDowne This will probably have some overlap with the [DE work](https://github.com/mozilla/activity-stream/pull/5367). Do you think I should wait until that lands to merge this?

gvn:
@flodolo Is there anything I need to do to get this string translated, or is landing it sufficient to trigger the process? Thanks!

flodolo:
This needs to be ported to mozilla-central. Once it lands there, localization will start.

Mardak:
@gvn did you get review from scott out of github? I would think flod only reviewed/approved the .ftl changes

gvn:
@Mardak Yes.

ScottDowne:
Yeah this looked good on my end. I've already rebased my DE patch against it and everything's working as intended.

PR #5377, opened 2019-11-15
(nobug) - Follow up to 1595090 fix test coverage

piatra:


piatra:
TC failure is related to the aws gcp migration that results in the artifacts used in the build process to go missing.

PR #5379, opened 2019-11-19
Backports nov 19

ScottDowne:


ScottDowne:
I did something wrong and something doesn't look right. Closing for now.

ScottDowne:
Ah, I didn't do anything wrong. Commit 1 added changes and commit 2 just removed them. Technically this is correct, it just shows no changes in the files once both are applied.

PR #5382, opened 2019-11-19
Bug 1597697 - Stop sending pings to Tiles data pipeline

ncloudioj:
This is the part 2 of https://phabricator.services.mozilla.com/D53875, which shuts off all the telemetry in Activity/Discovery Stream to Tiles data pipeline.

### DIFF
--- a/lib/ActivityStream.jsm
+++ b/lib/ActivityStream.jsm
@@ -314,13 +314,6 @@ const PREFS_CONFIG = new Map([
     },
   ],
   [
-    "telemetry.ping.endpoint",
-    {
-      title: "Telemetry server endpoint",
-      value: "https://tiles.services.mozilla.com/v4/links/activity-stream",
-    },
-  ],
-  [
     "section.highlights.includeVisited",
     {
       title:
--- a/lib/TelemetryFeed.jsm
+++ b/lib/TelemetryFeed.jsm
@@ -74,8 +74,6 @@ XPCOMUtils.defineLazyServiceGetters(this, {
 });
 
 const ACTIVITY_STREAM_ID = "activity-stream";
-const ACTIVITY_STREAM_ENDPOINT_PREF =
-  "browser.newtabpage.activity-stream.telemetry.ping.endpoint";
 const DOMWINDOW_OPENED_TOPIC = "domwindowopened";
 const DOMWINDOW_UNLOAD_TOPIC = "unload";
 const TAB_PINNED_EVENT = "TabPinned";
@@ -257,10 +255,7 @@ this.TelemetryFeed = class TelemetryFeed {
    */
   get pingCentre() {
     Object.defineProperty(this, "pingCentre", {
-      value: new PingCentre({
-        topic: ACTIVITY_STREAM_ID,
-        overrideEndpointPref: ACTIVITY_STREAM_ENDPOINT_PREF,
-      }),
+      value: new PingCentre({ topic: ACTIVITY_STREAM_ID }),
     });
     return this.pingCentre;
   }
@@ -420,7 +415,6 @@ this.TelemetryFeed = class TelemetryFeed {
         source,
         tiles: impressionSets[source],
       });
-      this.sendEvent(payload);
       this.sendStructuredIngestionEvent(
         payload,
         STRUCTURED_INGESTION_NAMESPACE_AS,
@@ -453,7 +447,6 @@ this.TelemetryFeed = class TelemetryFeed {
         tiles,
         loaded: tiles.length,
       });
-      this.sendEvent(payload);
       this.sendStructuredIngestionEvent(
         payload,
         STRUCTURED_INGESTION_NAMESPACE_AS,
@@ -641,9 +634,9 @@ this.TelemetryFeed = class TelemetryFeed {
   }
 
   sendEvent(event_object) {
-    if (this.telemetryEnabled) {
-      this.pingCentre.sendPing(event_object, { filter: ACTIVITY_STREAM_ID });
-    }
+    // As per Bug 1597697, all pings to Tiles are simply stopped here. Though we
+    // leave these pings around so that we could send to another destination or
+    // drop them if needed in the future.
   }
 
   sendUTEvent(event_object, eventFunction) {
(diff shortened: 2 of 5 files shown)


lib/ActivityStream.jsm
```diff
@@ -313,13 +313,6 @@ const PREFS_CONFIG = new Map([
       value: "https://incoming.telemetry.mozilla.org/submit",
     },
   ],
-  [
-    "telemetry.ping.endpoint",
```
ScottDowne: Curious, does the browser clean up this pref if you delete it here like you are?

Or does it stick around in the user's profile?

lib/TelemetryFeed.jsm
```diff
@@ -420,7 +415,6 @@ this.TelemetryFeed = class TelemetryFeed {
         source,
         tiles: impressionSets[source],
       });
-      this.sendEvent(payload);
```
ScottDowne: Technically, if I understand what's happening, it's probably harmless to leave this line here because `sendEvent` is now a "black hole" you're just removing this line as a little cleanup, right?

ScottDowne: Ah, so given that this is an event that's already been migrated, there is no need to ever have `sendEvent` here.

ScottDowne: `sendStructuredIngestionEvent` is the new pipeline now, right?

lib/TelemetryFeed.jsm
```diff
@@ -257,10 +255,7 @@ this.TelemetryFeed = class TelemetryFeed {
    */
   get pingCentre() {
     Object.defineProperty(this, "pingCentre", {
-      value: new PingCentre({
-        topic: ACTIVITY_STREAM_ID,
-        overrideEndpointPref: ACTIVITY_STREAM_ENDPOINT_PREF,
```
ScottDowne: I noticed pingCentre is used in both `sendEvent` and `sendStructuredIngestionEvent` in the case of `sendEvent` it's being removed. In the case of `sendStructuredIngestionEvent` it's still using `pingCenter` So it's safe to remove `overrideEndpointPref` because it was only needed for the `sendEvent` usage, and it's however NOT needed in `sendStructuredIngestionEvent`'s usage of `pingCenter`?

lib/TelemetryFeed.jsm
```diff
@@ -641,9 +634,9 @@ this.TelemetryFeed = class TelemetryFeed {
   }
 
   sendEvent(event_object) {
```
ScottDowne: So if I got this, the events we're losing are `countPinnedTab`, `handleUndesiredEvent`, `sendPageTakeoverData` and `TELEMETRY_PERFORMANCE_EVENT`.

We're keeping because they are already migrated are: `handleUserEvent` `handleImpressionStats` `sendDiscoveryStreamLoadedContent` `sendDiscoveryStreamImpressions` `endSession`. I don't see where you remove the `sendEvent` call in `endSession` and `handleUserEvent`, could we, or is there a reason to keep it around for now?

ping-centre/PingCentre.jsm
```diff
@@ -38,7 +34,6 @@ const PREF_BRANCH = "browser.ping-centre.";
 
 const TELEMETRY_PREF = `${PREF_BRANCH}telemetry`;
 const LOGGING_PREF = `${PREF_BRANCH}log`;
-const PRODUCTION_ENDPOINT_PREF = `${PREF_BRANCH}production.endpoint`;
```
ScottDowne: Ah, I think I see now an answer to one of my previous questions.

ping-centre/PingCentre.jsm
```diff
@@ -201,8 +196,6 @@ class PingCentre {
     this._topic = options.topic;
     this._prefs = Services.prefs.getBranch("");
 
-    this._setPingEndpoint(options.topic, options.overrideEndpointPref);
```
ScottDowne: There is a comment above with "@param {string} options.overrideEndpointPref - optional pref for URL where the POST is sent."

We can probably delete that, right?

ping-centre/PingCentre.jsm
```diff
@@ -336,39 +319,6 @@ class PingCentre {
     return payload;
   }
 
-  async sendPing(data, options) {
-    if (!this.enabled) {
-      return Promise.resolve();
-    }
-
-    const payload = await this._createPing(data, options);
```
ScottDowne: Is `_createPing` dead code now and can we delete it too?

ScottDowne: Or are we going to possibly use it in the upcoming patches?

lib/ActivityStream.jsm
ncloudioj: I think the browser will delete it if this pref no longer exists.

lib/TelemetryFeed.jsm
ncloudioj: Correct, `overrideEndpointPref` is only used by PingCentre for Tiles. For the new data pipeline (`StructuredIngestion`), its endpoint will be specified by the function argument of `sendStructuredIngestionEvent`.

lib/TelemetryFeed.jsm
ncloudioj: Exactly, we used to send it to both sides, now there is no point to do so as it's already been migrated.

lib/TelemetryFeed.jsm
ncloudioj: Actually, `sendEvent` takes care of all pings routed to Tiles. So this patch will make us lose all those pings.

Like I mentioned earlier, we're going to add them back one by one in the following releases. As planned, the event pings will be done first in 72.

ping-centre/PingCentre.jsm
ncloudioj: Good point.

ping-centre/PingCentre.jsm
ncloudioj: Yes, it's still useful for the upcoming patches, such as the event pings.

lib/ActivityStream.jsm
ScottDowne: Good to know!

lib/TelemetryFeed.jsm
ScottDowne: Hm, I think I'm still missing something.

It looks like to me that `handleUserEvent` and `endSession` have already been migrated, and thus can have the `sendEvent` removed like you do in `handleImpressionStats` `sendDiscoveryStreamLoadedContent`, etc, or am I missing something?

ScottDowne: We're good because `sendUTEvent` is not the same as the stuff we do in handleImpressionStats sendDiscoveryStreamLoadedContent, etc

PR #5384, opened 2019-11-20
Bug 1597301 - Greening up dev deps

gvn:


gvn:
Looks like Prettier update blew up, perhaps due to new rules despite being a minor bump? Should I separate that out into its own PR with files prettified to pass?

piatra:
@gvn our `package.json` specifies `npm 6.9`. I don't think it's a big difference in the end result but it would mean that everyone else doing npm install would have a different `package-lock.json`.

gvn:
I accidentally put the wrong nvm profile in when I was writing that comment. Correct versions are now listed!

piatra:
Hm strage
https://www.npmjs.com/package/karma-sinon
There is no 1.8.0 release.

package.json
```diff
@@ -17,61 +17,61 @@
...
     "chai-json-schema": "1.5.1",
     "cpx": "1.5.0",
-    "enzyme": "3.9.0",
-    "enzyme-adapter-react-16": "1.13.2",
-    "eslint": "6.2.2",
-    "eslint-config-prettier": "4.2.0",
+    "enzyme": "3.10.0",
+    "enzyme-adapter-react-16": "1.15.1",
+    "eslint": "6.6.0",
+    "eslint-config-prettier": "6.7.0",
```
piatra: `Prettier` is safe to update but we should keep the configs in sync with [m-c](https://searchfox.org/mozilla-central/source/package.json).

package.json
```diff
@@ -17,61 +17,61 @@
...
-    "eslint-plugin-jsx-a11y": "6.2.1",
-    "eslint-plugin-mozilla": "2.1.0",
+    "eslint-plugin-import": "2.18.2",
+    "eslint-plugin-jsx-a11y": "6.2.3",
+    "eslint-plugin-mozilla": "2.3.0",
     "eslint-plugin-no-unsanitized": "3.0.2",
-    "eslint-plugin-prettier": "3.0.1",
-    "eslint-plugin-react": "7.13.0",
-    "eslint-plugin-react-hooks": "1.6.0",
+    "eslint-plugin-prettier": "3.1.1",
```
piatra: This is ok too.

dmose:
Sorry for the long delay getting to this!

@gvn, I had not see `next-update` before, that's very cool!

Prettier makes changes with updated versions (see https://prettier.io/docs/en/install.html). This means that we need to be using the same version as mozilla-central, so it can't be upgraded.  We should probably add a package.json comment to that effect to avoid future confusion:

```json
    "//": "Prettier must be the same version as the one in mozilla-central's top-level package.json",
    "prettier": "1.17.0"
```

I suppose we could add (meta-data) [https://github.com/bahmutov/next-update#ignoring-or-skipping-some-modules] to make `next-update` do the right thing also.

PR #5385, opened 2019-11-21
Bug 1594537 - Make DS Responsive

gvn:
Test steps:

- With DS enabled, resize the window and scroll around. Nothing should look broken.
- With DS disabled, resize the window and scroll around. Nothing should look broken.


content-src/components/Base/_Base.scss
```diff
@@ -5,6 +5,14 @@
   min-height: 100vh;
   padding: ($section-spacing + $section-vertical-padding) $base-gutter $base-gutter;
 
+  &.ds-outer-wrapper-breakpoint-override {
```
ScottDowne: I see another instance of this class further down the file, is it intended to have it twice?

content-src/components/Search/_Search.scss
```diff
@@ -64,14 +64,14 @@ $glyph-forward: url('chrome://browser/skin/forward.svg');
     height: $search-height;
     margin: 0 auto;
     position: relative;
-    width: $searchbar-width-small;
```
ScottDowne: These changes just reducing complexity because these variables are only used in 1 place? Or is there something else you're trying to adjust, where the values in the variables are off?

content-src/components/Base/_Base.scss
gvn: Yeah, they're for different selectors.

content-src/components/Search/_Search.scss
gvn: This is now removed...wasn't necessary as it was a single use var.

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -99,19 +74,13 @@
...
-      margin: auto;
-    }
   }
 
   .blockButton {
     display: block;
     inset-inline-end: 20px;
     opacity: 1;
-    top: 50%;
+    top: 20px;
```
ScottDowne: Does this apply to AS and DS, is that intended?

content-src/asrouter/templates/SimpleBelowSearchSnippet/_SimpleBelowSearchSnippet.scss
```diff
@@ -53,10 +32,6 @@
   margin: auto;
   z-index: auto;
 
-  @media (min-width: $break-point-large) {
```
ScottDowne: Does this apply to AS and DS, is that intended?

content-src/components/DiscoveryStreamComponents/DSTextPromo/_DSTextPromo.scss
```diff
@@ -1,18 +1,27 @@
 .ds-dismiss-ds-text-promo {
-  width: 744px;
```
ScottDowne: If you add this back in, but change it from `width: 744px;` to `max-width: 744-x;` You can get both the old size at max breakpoint, and the smaller breakpoints work how you have them. Thoughts?

gvn:
Moving to Phabricator:

https://phabricator.services.mozilla.com/D56784

PR #5386, opened 2019-11-21
Bug 1597697 - Migrate events ping to the new data pipeline

ncloudioj:
This migrates all Event pings in AS/DS to the Structured Ingestion pipeline.

Steps to test:
* Turn on PingCentre logging
* Turn on AS Telemetry as well as Structured Ingestion by flipping "activity-stream.telemetry" and "activity-stream.telemetry.structuredIngestion"
* Click any card on the newtab page, or search in the search box
* You should see events ping logged in the browser console
* You can further check if those pings made to the database by using this [query](https://sql.telemetry.mozilla.org/queries/65254/source), note that you might need to change the date in it.

### DIFF
--- a/lib/TelemetryFeed.jsm
+++ b/lib/TelemetryFeed.jsm
@@ -634,9 +634,25 @@ this.TelemetryFeed = class TelemetryFeed {
   }
 
   sendEvent(event_object) {
-    // As per Bug 1597697, all pings to Tiles are simply stopped here. Though we
-    // leave these pings around so that we could send to another destination or
-    // drop them if needed in the future.
+    switch (event_object.action) {
+      case "activity_stream_user_event":
+        this.sendEventPing(event_object);
+        break;
+    }
+  }
+
+  async sendEventPing(ping) {
+    delete ping.action;
+    ping.client_id = await this.telemetryClientId;
+    if (ping.value && typeof ping.value === "object") {
+      ping.value = JSON.stringify(ping.value);
+    }
+    this.sendStructuredIngestionEvent(
+      ping,
+      STRUCTURED_INGESTION_NAMESPACE_AS,
+      "events",
+      1
+    );
   }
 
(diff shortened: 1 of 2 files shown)


lib/TelemetryFeed.jsm
```diff
@@ -634,9 +634,25 @@ this.TelemetryFeed = class TelemetryFeed {
...
+    }
+  }
+
+  async sendEventPing(ping) {
+    delete ping.action;
+    ping.client_id = await this.telemetryClientId;
+    if (ping.value && typeof ping.value === "object") {
+      ping.value = JSON.stringify(ping.value);
+    }
+    this.sendStructuredIngestionEvent(
```
ScottDowne: Is this at some point still looking to ensure `telemetryEnabled` is true before sending it?

lib/TelemetryFeed.jsm
```diff
@@ -634,9 +634,25 @@ this.TelemetryFeed = class TelemetryFeed {
...
+      case "activity_stream_user_event":
+        this.sendEventPing(event_object);
+        break;
+    }
+  }
+
+  async sendEventPing(ping) {
+    delete ping.action;
+    ping.client_id = await this.telemetryClientId;
+    if (ping.value && typeof ping.value === "object") {
```
ScottDowne: `ping` is always an object and can never be undefined?

ScottDowne: Oh, I see it now, yeah `ping` is safe and always an object.

lib/TelemetryFeed.jsm
ScottDowne: Actually I think the other uses of `sendStructuredIngestionEvent` example: handleImpressionStats don't have a check for `telemetryEnabled` because it's inside `telemetryEnabled` so this is good here too.

PR #5388, opened 2019-11-25
Bug 1597492 - Add Pocket tile type to Top Story click ping

ncloudioj:
This adds a `card_type` field to the `CLICK` ping for Top Stories. Note that this doesn't change the Pocket telemetry (i.e. with `impression_id` and `tile_id`), but the user event ping, which contains `client_id` without `tile_id`. Per bug 1597492, this will be useful for us to study the user engagement of Top Story related experiments. 

This requires a data review.

### DIFF
--- a/content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
+++ b/content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
@@ -95,6 +95,7 @@ export class DSCard extends React.PureComponent {
           event: "CLICK",
           source: this.props.type.toUpperCase(),
           action_position: this.props.pos,
+          value: { card_type: this.props.flightId ? "spoc" : "organic" },
         })
       );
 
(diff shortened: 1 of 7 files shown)


content-src/components/DiscoveryStreamComponents/DSCard/DSCard.jsx
```diff
@@ -95,6 +95,7 @@ export class DSCard extends React.PureComponent {
           event: "CLICK",
           source: this.props.type.toUpperCase(),
           action_position: this.props.pos,
+          value: { card_type: this.props.flightId ? "spoc" : "organic" },
```
ScottDowne: Is value new?

ncloudioj: No, `value` is not new to this ping. The new part is the `card_type` for Top Stories events.

ScottDowne: So previously we were not sending a value in this ping but it's optional?

(I'm just confused because it looks like you're adding it here)

ncloudioj: Correct, it's an optional field, we didn't send it for Top Stories `CLICK` ping, but we did for other `CLICK` pings in Top Sites and Highlights. So yeah, in a sense, it's a new field to this particular ping.

ncloudioj:
Data review granted on bugzilla.

TravisCI was broken due to unable to locate ESLint, will try to fix it separately.

PR #5389, opened 2019-11-26
(no bug) - Fix TravisCI

ncloudioj:


ncloudioj:
This makes Travis happy again. Somehow, it could not find those dependent executables such as eslint and webpack located in `./node_modules/.bin`.

PR #5390, opened 2019-11-27
Bug 1598876 - Fix intermittent link test failure

gvn:
Run `mach test discovery_render` and confirm exit code of 0. (Try this several times.)


ScottDowne:
I also wonder if we can instead of waiting, just check for the href, and not the text content, thus it wouldn't need to be waiting for fluent. (we're not testing fluent anyway)

thoughts?

ScottDowne:
Was thinking about this more. I think maybe we have a test layout for this, so we don't have fluent strings being used here at all. (assuming that's the issue, which I'm only about 80% sure of that)

PR #5391, opened 2019-12-03
Backport 12 03

ncloudioj:


Mardak:
Gah what timing again. We just discussed with kate and dan that we would just do batch port/import now. https://github.com/mozilla/activity-stream/commit/a0a94918a8873dd7f1c8ced1750127fee1dd96d4

PR #5393, opened 2019-12-06
Bug 1597492 - Add Pocket tile type to Top Story click ping (#5388)

ncloudioj:
* Bug 1597492 - Add Pocket tile type to Top Story click ping


ncloudioj:
Already uplifted in beta 72.

